PR #35, opened 2017-08-18
Favicon is at root

stefanv:


stefanv:
Without this patch, when switching to `/source`, the browser may request favicon from `/source/favicon.png`.

PR #40, opened 2017-08-24
Create fixture data for tests

bnaul:
Also copied over the default Docker configuration from `cesium_web`; don't think anything about that needs to change for now.


stefanv:
This looks good; what's the purpose of the Docker file here?

PR #44, opened 2017-09-16
Initial fixture data and tests

bnaul:
Right now only tests basic functionality:
- Front page renders
- Source page renders
- Plots are present
- Can post a comment

Worth looking closely at the way fixtures are setup before merging, but once we're happy we should merge and add additional tests as we go along.


bnaul:
Let's go ahead and merge this since it's better than nothing.

PR #51, opened 2017-10-10
Env loader

stefanv:


config.yaml.example
```diff
@@ -28,7 +27,6 @@ server:
     # You need to have Google+ API enabled; it takes a few minutes to activate.
 
     auth:
-        debug_login: False
+        debug_login: True
```
bnaul: Was this intentional? I always check this in by mistake which might suggest that it should stay this way...

stefanv:
Good question. Either way is probably fine, although I think this is always 
the initial state you want?

PR #59, opened 2017-10-20
Add thumbnails

bnaul:
Currently just serves files from local storage, but should be very easy to adapt to use S3. Would appreciate some feedback though on the general architecture / JavaScript style.

EDIT: After changing the data setup around a bit, the handling of S3 access is going to happen at data-insert time; the set of changes here can be data location-agnostic, so it should be good to go as-is.


skyportal/app_server.py
```diff
@@ -45,13 +46,19 @@ def make_app(cfg, baselayer_handlers, baselayer_settings):
...
 
+
         # User-facing pages
         (r'/.*', MainPageHandler)  # Route all frontend pages, such as
                                    # `/source/g647ba`, through the main page.
                                    #
                                    # Refer to Main.jsx for routing info.
     ]
 
+    if os.path.exists(cfg['paths:thumbnails']):
```
bnaul: TODO serve `/thumbnails/` via nginx instead

requirements.txt
```diff
@@ -7,3 +7,4 @@ bokeh==0.12.9
 pytest-randomly
 factory-boy==2.8.1
 pytest-factoryboy==1.3.1
+cloudstorage
```
stefanv: Is this used anywhere?

skyportal/app_server.py
```diff
@@ -1,3 +1,4 @@
+import os.path
```
stefanv: Revert changes to this file.

skyportal/models.py
```diff
@@ -170,3 +221,30 @@ def from_ascii(cls, filename, source_id, instrument_id, observed_at):
         return cls(wavelengths=data[:, 0], fluxes=data[:, 1],
                    source_id=source_id, instrument_id=instrument_id,
                    observed_at=observed_at)
+
+
+#def format_public_url(context):
```
stefanv: What should happen to this?

skyportal/plot.py
```diff
@@ -145,6 +145,7 @@ def _plot_to_json(plot):
     return docs_json, render_items, custom_model_js
 
 
+# TODO async
```
stefanv: What was our decision here?

skyportal/models.py
bnaul: Just leave it there for now, there's nowhere better for it at the moment.

skyportal/plot.py
bnaul: Not related to this PR, just a reminder. I clarified the TODO text a bit.

PR #68, opened 2017-12-04
Plot caching

acrellin:


static/js/containers/PlotContainer.jsx
```diff
@@ -1,12 +1,70 @@
...
+      error: false,
+      fetchingPlotIDs: []
+    };
+  }
+  componentWillMount() {
+    this.fetchPlotDataIfNotCached();
+  }
+  componentWillReceiveProps() {
+    this.fetchPlotDataIfNotCached();
+  }
```
stefanv: This is not correct; should be replaced by `componentDidMount`.

static/js/containers/PlotContainer.jsx
```diff
@@ -1,12 +1,70 @@
...
+  fetchPlotDataIfNotCached() {
+    if (!this.props.plots.plotIDList.includes(this.props.url) &&
```
stefanv: This if check may be better refactored into a function, like `needsFetching`, which can be reused below.  Actually, the part below can probably just go in an `else` clause?

static/js/containers/PlotContainer.jsx
```diff
@@ -1,12 +1,70 @@
...
+          Actions.FETCH_SOURCE_PLOT
+        )
+      );
+      this.setState(
+        { fetchingPlotIDs: this.state.fetchingPlotIDs.concat([this.props.url]) }
+      );
+    }
+    if (this.props.plots.plotData[this.props.url] &&
+        this.state.fetchingPlotIDs.includes(this.props.url)) {
+      const fetchingPlotIDs = this.state.fetchingPlotIDs.slice();
```
stefanv: This is weird; if we really can't find a cleaner way to express this, a comment may be helpful.

static/js/reducers.js
```diff
@@ -71,6 +71,33 @@ export function profileReducer(state={ username: '' }, action) {
   }
 }
 
+export function plotsReducer(state={ plotData: {}, plotIDList: [] }, action) {
+  switch (action.type) {
+    case Action.FETCH_SOURCE_PLOT_OK: {
+      const plotData = Object.assign({}, state.plotData);
```
stefanv: Use a spread operator?

static/js/reducers.js
```diff
@@ -71,6 +71,33 @@ export function profileReducer(state={ username: '' }, action) {
...
+      const plotIDList = state.plotIDList.slice();
+
+      const { url, ...incomingData } = action.data;
+      plotIDList.unshift(url);
+      plotData[url] = incomingData;
+      if (plotIDList.length >= 40) {
+        plotIDList.length = 40;
+        Object.keys(plotData).forEach((ID) => {
```
stefanv: What is happening here?

acrellin:
Thanks for the feedback Stéfan, I'll update accordingly.

static/js/containers/PlotContainer.jsx
acrellin: @stefanv we want to check every time the component either a) mounts, or b) receives new props. Can you please explain what you're thinking here?

static/js/containers/PlotContainer.jsx
acrellin: No, they're totally different checks. One is checking if the data needs to be fetched, and the other is checking if the fetch is complete and the ID needs to be removed from the list of in-process fetches (that is, it's complete).

static/js/containers/PlotContainer.jsx
acrellin: `splice` simply removes the arrray element(s) at the specified index. Does knowing that clarify this enough?

acrellin:
@stefanv what else would you like to see changed before this gets merged?

static/js/containers/PlotContainer.jsx
stefanv: They've now renamed `componentWillReceiveProps` to `UNSAFE_componentWillReceiveProps`, so that's a first indication :)  `componentDidMount` isn't enough in isolation, I imagine, but it looks like `getDerivedStateFromProps` can work in the newest React.  See https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

static/js/containers/PlotContainer.jsx
stefanv: I see, you're right.

static/js/containers/PlotContainer.jsx
stefanv: Yes, I think a comment could help here, since `slice` has no argument.

PR #77, opened 2018-03-07
[WIP] Add barebones helm chart

yuvipanda:
Needs more work before this can be merged, but is a good start!

Result of a hack session with @stefanv!


stefanv:
Thank you for getting me started on Kubernetes, Yuvi!  We now have a working deployment in our `skyportal-deploy` repo, so I'll close this issue.

PR #86, opened 2018-08-27
push ubuntu/node versions; fix Dockerfile

profjsb:
Updated the README to show docker-compose usage. Python3.6 is the default python in ubuntu:17.10 so no need to use 3rd party ppa repos here.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -66,3 +66,4 @@ rules:
   lines-between-class-members: 1
   react/no-access-state-in-setstate: 1
   implicit-arrow-linebreak: 1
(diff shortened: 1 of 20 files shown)


profjsb:
Not sure why the travis builds are failing:
`The command ".travis/travis_script.sh" exited with 2.`
Perhaps a separate issue than the docker changes?

acrellin:
@profjsb Updating baselayer to `42f55cc` should fix the linter error (I just pushed some ESLint rule updates)

profjsb:
Tests passing on mac and linux. Travis tests to be fixed in another branch.

PR #89, opened 2018-08-29
Fix Travis issues

acrellin:


stefanv:
Requested activation of debug builds from Travis-CI, so that we can investigate the nginx failure.

stefanv:
:+1: :champagne:

PR #93, opened 2018-09-03
ZTF upload

profjsb:
- database changes 
- calculate variable star flux (using difference imaging)
- plot updates, with errorbars


profjsb:
Current SourceList page:
skyportal

Current Source page:
skyportal

profjsb:
This should be close to ready to go. Note:  I left in the work I did to parallelize the reading of AVRO files, but there's some complaints via `concurrent.futures` about pickling of lambda functions (related to the User table). Could not debug this.

static/js/components/SourceList.jsx
```diff
@@ -8,17 +8,31 @@ const SourceList = ({ sources }) => (
...
+              </td>
+              <td>{source.detect_photometry_count}</td>
+              <td>{source.simbad_class}</td>
+          </tr>
         ))
       }
-    </ul>
+    </table>
+    <script>
+    sorttable.makeSortable(document.getElementById('tab'));
```
stefanv: What does this do?

acrellin:
Looks good to me so far, but I'm not seeing the associated changes to the Source and Photometry handlers - shouldn't that be included here, or am I missing something?

static/js/components/SourceList.jsx
profjsb: I was trying to add a sortable table, but failed. Removed this.

stefanv:
@acrellin Scripts that use load_env should allow flags such as `--config` and `--debug`, otherwise there is no way to specify custom configuration.

stefanv:
`--workers` argument broken; does not cast string to int

acrellin:
Closing and moving changes to https://github.com/skyportal/skyportal/pull/104

PR #97, opened 2018-09-11
Add comment attachments

### LINKED ISSUE #78: Ability to add attachments to comments
A commenter should be able to upload an attachment (.PDF, .DOC, .MD, .JPG, .GIF, .HTML) to the comments section. This file should get stored locally (and/or S3). Others who would like to see the attached file will need to click on a small icon that appears after the attachment is uploaded. Only the user who uploaded the attachment can delete it (which for now means it's no longer viewable, but isn't permanently deleted on disk).

acrellin:
Closes #78


profjsb:
Can you post some screenshots of what this looks like?

acrellin:
comment_attachment

acrellin:
They're download links

profjsb:
Looking good. A few comments:

  - Would be useful to limit the file size, say to 10MB to avoid jamming up the filesystem.
  - Add comment attachments to the test suite

acrellin:
@profjsb  There are now tests for uploading and downloading comment attachments.  As for limiting the file size, it doesn't look like there's a way to configure that per endpoint and not just the whole application (it's an nginx config issue).  Let me know if you'd like to impose a 10MB file size limit application-wide.

profjsb:
Yes, I think it makes sense to have nginx limit to 10MB. If we have a new use case that calls for me, we can reevaluate it. Please make the change.

PR #99, opened 2018-09-13
Add PS1 thumbnail crosshair overlay

acrellin:


acrellin:
ps1crosshairsscreenshot

PR #104, opened 2018-10-19
WIP:  Source filtering and updated tests

acrellin:


tools/ztf_upload_avro.py
```diff
@@ -660,6 +662,7 @@ def update_to(self, b=1, bsize=1, tsize=None):
     parser.add_argument('-c', '--clobber', action='store_true', default=False)
     parser.add_argument('--datadir', default='/tmp/ZTF')
     parser.add_argument('--workers', default=1)
+    parser.add_argument('--only_pure', action='store_true', default=False)
```
profjsb: This should be True by default. We only want `pure` detections.

acrellin: With `only_pure=True`, every single source gets skipped on my machine.

dannygoldstein:
requesting this be merged

stefanv:
Moved to #132

Issue #110, opened 2018-12-25
use q3c


dannygoldstein:
Source positions should be indexed with a quad tree cube for fast spatial cross-matching. 

q3c is the standard tool for this in psql.

https://github.com/segasai/q3c

Danny

dannygoldstein:
fixed in my branch of skyportal

stefanv:
See discussion on https://github.com/cesium-ml/baselayer/pull/59

acrellin:
Danny's spatial backend PR covers this

PR #111, opened 2018-12-26
integrate q3c

dannygoldstein:
This PR addresses #110 by building q3c into the database docker image and indexing the source table on RA Dec via q3c.


dannygoldstein:
superseded by cesium-ml/baselayer#59

PR #137, opened 2019-03-12
Improve Make help message

stefanv:
- Display the SkyPortal version with `make help`
- `make help` is now the default target
- Use a different mechanism for calling out to the baselayer Makefile


acrellin:
This needs a rebase, then I'll go ahead and merge.

PR #141, opened 2019-03-22
Display empty DB message in SourceList component

acrellin:


static/js/components/SourceList.jsx
```diff
@@ -3,24 +3,39 @@ import PropTypes from 'prop-types';
...
-              {source.id}
-            </Link>
-          </li>
-        ))
-      }
-    </ul>
-  </div>
-);
+const SourceList = ({ sources }) => {
+  if (typeof sources === 'string' || sources instanceof String) {
```
stefanv: I think we want to avoid returning varying types from our API.  Better would be to do a check for an empty list here, and then display something like "There are no sources to display."

acrellin: @stefanv this is per our conversation about needing to distinguish between the case in which there are no sources that a user has access to versus the case in which the sources table is empty. Returning an empty array doesn't help to distinguish between these two cases. Any thoughts?

stefanv:
I have a plan to address this. Let's chat about it tomorrow.

acrellin:
Moved to https://github.com/skyportal/skyportal/pull/142

PR #142, opened 2019-03-25
Provide system information to frontend

stefanv:


static/js/components/Footer.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import styles from './Footer.css';
+
+
+const Footer = props => (
+  <div className={styles.footer}>
+    <div className={styles.footerContent}>
+      SkyPortal version {props.sysinfo.skyportal_version}. Please file issues at&nbsp;
```
stefanv: Let's make this v{...}.

static/js/components/Footer.jsx
```diff
@@ -0,0 +1,22 @@
...
+      <a href="https://github.com/skyportal/skyportal">
+        https://github.com/skyportal/skyportal
+      </a>
+      .
+    </div>
+  </div>
+);
+Footer.propTypes = {
+  sysinfo: PropTypes.object.isRequired
```
stefanv: Map out only version here.

static/js/components/Footer.jsx
stefanv: Components should not know anything about the app state.  So this should simply be something like

```
const Footer = ({version}) => (
...
```

static/js/containers/Main.css
```diff
@@ -5,6 +5,7 @@
     top: 0;
     left: 0;
     min-height: 100%;
+    padding-bottom: 5em;
```
stefanv: :+1:

static/js/containers/SourceListContainer.jsx
```diff
@@ -15,6 +15,9 @@ class SourceListContainer extends React.Component {
   }
 
   render() {
+    if (this.props.sourcesTableEmpty) {
+      return "The sources table is empty.";
```
stefanv: Here, we should render a standard messages when the source list has length zero.

The message about the source table being uninitialized should go into another component.  That component will render some text that points to the manual, how to ingest data, etc.

skyportal/handlers/sysinfo.py
```diff
@@ -0,0 +1,15 @@
...
+from ..models import Source, DBSession
+import skyportal
+
+import tornado.web
+
+
+class SysInfoHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        info = {}
```
stefanv:
```
info = {
    'db_empty': DBSession.query(Source).first() is None,
    'version': skyportal.__version__
}
```

static/js/components/Footer.jsx
```diff
@@ -0,0 +1,25 @@
...
+      </a>
+      .
+    </div>
+  </div>
+);
+Footer.propTypes = {
+  version: PropTypes.string
+};
+Footer.defaultProps = {
+  version: '_unknown_'
```
stefanv: Maybe make this `0.0.0`, in case something goes wrong.

static/js/components/UninitializedDBMessage.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+
+const UninitializedDBMessage = () => (
+  <div>
+    Welcome to SkyPortal. The Sources table is currently empty. <br />
+    For help with initializing the database, see the&nbsp;
+    <a href="https://skyportal.io/docs/setup.html">getting started documentation</a>.
```
stefanv: Great, and let's double check that the documentation shows how to ingest data.  (`make data`)

skyportal/tests/frontend/test_remote.py
```diff
@@ -3,10 +3,14 @@
 # fixture data directly into the database). Used for testing broadly that a
 # server or Docker image was started successfully.
 
+from skyportal.models import Source, DBSession
+
 
 def test_remote(driver):
     # TODO expand to cover the basics of all site functionality
     # (c.f. `test_pipeline_sequentially` from `cesium_web`)
+    Source.query.delete()
```
stefanv: Still necessary?

stefanv:
Looks good, Ari, thanks. Go ahead and merge as soon as tests pass.

PR #150, opened 2019-04-08
Move tokens, plot, and profile to /api/internal/<x>

acrellin:


stefanv:
Can we group the internal API items at the bottom and the external API ones at the top?  Otherwise looks good; you can merge after that change.

PR #126, opened 2019-02-02
Add user info endpoint

stefanv:
Needs a test still—@acrellin is our expert there.

PR #152, opened 2019-04-09
Add user info endpoint

acrellin:
Takes over https://github.com/skyportal/skyportal/pull/126


acrellin:
Danny wrote: "Only one change I would make:

The new route for users is /user/id but in Group.jsx the Link is to /users/id, which is an invalid route.

Danny"

stefanv:
Looks good; just needs a test!

acrellin:
@stefanv tests in place - ready to go!

PR #162, opened 2019-04-30
Return skyportal version with each API response

### LINKED ISSUE #143: Return the version of `skyportal` with each API result
We don't version the JSON responses as such, but this would at least allow a robot to work around any issue that we introduced in a specific version, on purpose or by accident.

acrellin:
This PR introduces a change to how we call the `success` and `error` handler methods: when passing in `data`, this must always be an explicit keyword argument (e.g. `self.success(data=data)`).

Closes #143.

### DIFF
--- a/skyportal/handlers/base.py
+++ b/skyportal/handlers/base.py
@@ -1,0 +1,16 @@
+from baselayer.app.handlers.base import BaseHandler as BaselayerHandler
+import skyportal
+
+
+class BaseHandler(BaselayerHandler):
+    def success(self, *args, **kwargs):
+        data = kwargs.get('data', {})
+        data['version'] = skyportal.__version__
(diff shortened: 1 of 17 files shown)


skyportal/handlers/base.py
```diff
@@ -0,0 +1,16 @@
+from baselayer.app.handlers.base import BaseHandler as BaselayerHandler
+import skyportal
```
stefanv: I'm worried a bit about circular dependencies here.  Maybe `from ..version import __version__`?

acrellin: Hmm, that doesn't work - there's no `version` module - are you suggesting we add it?

acrellin: @stefanv - changed

PR #163, opened 2019-05-02
Expand API tests

### LINKED ISSUE #160: Expand API tests
- Ensure a token can only be used within the scope of its permissions

acrellin:
- Adds `author` column to `Comment` table which stores either username or token name as a string instead of `user` and `user_id`.

Closes #160


stefanv:
Very nice work, Ari.

One question: when a comment gets filed by a token, what name is displayed?  This is something the user should be able to specify, but I don't see those changes here yet.  We should also make sure those token names cannot overlap with existing usernames, and they probably should not contain '@' characters.

acrellin:
@stefanv - currently, when a token creates a comment, the token ID will be displayed as the author. These cannot overlap with valid user names, as they are UUID4 strings, which only contain alphanumeric characters and "-". Do you want the token "name"/description (optional) to be displayed instead?

stefanv:
Yes, let's give tokens a name. That way you can have, eg, "classifier-bot" 
comment.

acrellin:
@stefanv - updated to use the token name instead of ID as the comment author.

skyportal/tests/api/test_comments.py
```diff
@@ -12,3 +12,11 @@ def test_add_and_retrieve_comment(comment_token, public_source):
 
     assert status == 200
     assert data['data']['comment']['text'] == 'Comment text'
+
+
+def test_cannot_add_comment_without_permission(view_only_token, public_source):
+    status, data = api('POST', 'comment', data={'source_id': public_source.id,
+                                                'text': 'Comment text'},
+                       token=view_only_token)
+    assert status == 400
```
stefanv: Minor issue for later: would be nice if we could return a 403 instead.  See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

Issue #154, opened 2019-04-12
More general source table columns


acrellin:
From Danny Goldstein:

The new columns in source are fairly general, but I worry that eventually the Source table will grow to so many columns that it will be difficult to keep track of the schema. To simplify things, I wonder if it would be better to have a single metadata column with the JSONB datatype, and columns such as the following could be fields in a JSONB document (no-sql style):

simbad_class = sa.Column(sa.Unicode, nullable=True, )
simbad_info = sa.Column(JSONB, nullable=True)
gaia_info = sa.Column(JSONB, nullable=True)
tns_info = sa.Column(JSONB, nullable=True)
tns_name = sa.Column(sa.Unicode, nullable=True)

profjsb:
As it's clear that we are now pushing non-generic fields to metadata JSONB, we should give some thought to how we might keep track of what's available for query and for display. Some possibilities:

1) Do nothing. Leave it up to the power user to determine what they want to query on in the metadata. Configured display would have to known which metadata fields are available if they want to show them (e.g., a Gaia ID or SIMBAD link).

2) Provide a look up function on available fields (that the user can see). For example, scan ~1000 random sources and assemble the union of available fields. Could be a materialized view with a long refresh time to avoid making this call often.

3) Save the available fields in a separate table as the data is loaded and check as new sources are added if that table needs to be updated. 

One idea would be that this could be part of a data/schema exploration widget.

I can see pluses and minuses of each of the above. Curious to get @stefanv and @acrellin's thoughts.

stefanv:
I was thinking of having the custom fields be specified in the configuration file, and then exposing those fields to the API and rendering them to users.  So, I guess a pre-specified option 2.

PR #166, opened 2019-05-16
Collapse some Source table columns into single JSONB column

### LINKED ISSUE #128: Add metadata storage column to source

acrellin:
Closes #128 & #154

### DIFF
--- a/skyportal/handlers/source.py
+++ b/skyportal/handlers/source.py
@@ -206,10 +206,10 @@ class FilterSourcesHandler(BaseHandler):
             end_date = arrow.get(data['endDate'].strip())
             q = q.filter(Source.last_detected <= end_date)
         if data['simbadClass']:
-            q = q.filter(func.lower(Source.other_metadata['simbad_class'].astext)
+            q = q.filter(func.lower(Source.external_metadata['simbad']['class'].astext)
                          == data['simbadClass'].lower())
         if data['hasTNSname']:
-            q = q.filter(Source.tns_name.isnot(None))
+            q = q.filter(Source.external_metadata['tns']['name'].isnot(None))
 
         all_matches = list(q)
         info['totalMatches'] = len(all_matches)
--- a/skyportal/handlers/source.py
+++ b/skyportal/handlers/source.py
@@ -206,8 +206,8 @@ class FilterSourcesHandler(BaseHandler):
             end_date = arrow.get(data['endDate'].strip())
             q = q.filter(Source.last_detected <= end_date)
         if data['simbadClass']:
-            q = q.filter(func.lower(Source.simbad_class) ==
-                         data['simbadClass'].lower())
+            q = q.filter(func.lower(Source.other_metadata['simbad_class'].astext)
+                         == data['simbadClass'].lower())
         if data['hasTNSname']:
             q = q.filter(Source.tns_name.isnot(None))
 
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -100,8 +100,12 @@ class Source(Base):
     altdata = sa.Column(JSONB, nullable=True)
     created = sa.Column(ArrowType, nullable=False,
                         server_default=sa.func.now())
-
+    modified = sa.Column(sa.DateTime, nullable=False,
+                         server_default=sa.func.now(),
+                         server_onupdate=sa.func.now())
     last_detected = sa.Column(ArrowType, nullable=True)
+    origin = sa.Column(sa.String, nullable=True)
+
     dist_nearest_source = sa.Column(sa.Float, nullable=True)
     mag_nearest_source = sa.Column(sa.Float, nullable=True)
     e_mag_nearest_source = sa.Column(sa.Float, nullable=True)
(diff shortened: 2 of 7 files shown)


skyportal/models.py
```diff
@@ -112,24 +116,7 @@ class Source(Base):
 
     score = sa.Column(sa.Float, nullable=True)
 
-    ## pan-starrs
-    sgmag1 = sa.Column(sa.Float, nullable=True)
```
acrellin: @profjsb these are the columns we're moving to the single `other_metadata` JSONB column. Any thoughts/suggestions?

skyportal/models.py
```diff
@@ -112,24 +116,7 @@ class Source(Base):
...
-    srmag1 = sa.Column(sa.Float, nullable=True)
-    simag1 = sa.Column(sa.Float, nullable=True)
-    objectidps1 = sa.Column(sa.BigInteger, nullable=True)
-    sgscore1 = sa.Column(sa.Float, nullable=True)
-    distpsnr1 = sa.Column(sa.Float, nullable=True)
-
-    origin = sa.Column(sa.String, nullable=True)
-    modified = sa.Column(sa.DateTime, nullable=False,
```
profjsb: should keep `modified` as it's a generic record for when the row is changed.

skyportal/models.py
```diff
@@ -112,24 +116,7 @@ class Source(Base):
...
 
-    ## pan-starrs
-    sgmag1 = sa.Column(sa.Float, nullable=True)
-    srmag1 = sa.Column(sa.Float, nullable=True)
-    simag1 = sa.Column(sa.Float, nullable=True)
-    objectidps1 = sa.Column(sa.BigInteger, nullable=True)
-    sgscore1 = sa.Column(sa.Float, nullable=True)
-    distpsnr1 = sa.Column(sa.Float, nullable=True)
-
-    origin = sa.Column(sa.String, nullable=True)
```
profjsb: `origin` is a good column to keep for provenance -- in fact, I'd suggest making this not nullable.

acrellin: @profjsb these lines were actually just moved up a little bit, not removed

acrellin:
@profjsb the columns that have been collapsed into the new matadata JSONB column are: 
```    sgmag1 = sa.Column(sa.Float, nullable=True)
    srmag1 = sa.Column(sa.Float, nullable=True)
    simag1 = sa.Column(sa.Float, nullable=True)
    objectidps1 = sa.Column(sa.BigInteger, nullable=True)
    sgscore1 = sa.Column(sa.Float, nullable=True)
    distpsnr1 = sa.Column(sa.Float, nullable=True)
    simbad_class = sa.Column(sa.Unicode, nullable=True, )
    simbad_info = sa.Column(JSONB, nullable=True)
    gaia_info = sa.Column(JSONB, nullable=True)
    tns_info = sa.Column(JSONB, nullable=True)
    tns_name = sa.Column(sa.Unicode, nullable=True)
```
Any feedback?

skyportal/models.py
profjsb: ok

profjsb:
It's to move those ZTF-specific data to `other_metadata` JSONB field. But we should move the remainder of the data to `external_metadata` JSONB column. This can hold externally available metadata about that source that is not found in the survey from which that source originated. It can be structured like:

```json
external_metadata = 
 {"simbad": {"class": "RRL", "info": {...}},
  "TNS": {"name": "SN1987a", "info": {...}},
  "Gaia": {"info": {...}}
 }
``` 
So the lookups in the ZTF version of skyportal would be to `external_metadata.Gaia.info.Id` etc.

acrellin:
> It's to move those ZTF-specific data to `other_metadata` JSONB field. But we should move the remainder of the data to `external_metadata` JSONB column. This can hold externally available metadata about that source that is not found in the survey from which that source originated. It can be structured like:
> 
> ```json
> external_metadata = 
>  {"simbad": {"class": "RRL", "info": {...}},
>   "TNS": {"name": "SN1987a", "info": {...}},
>   "Gaia": {"info": {...}}
>  }
> ```
> 
> So the lookups in the ZTF version of skyportal would be to `external_metadata.Gaia.info.Id` etc.

Thanks for the feedback - I'll change it accordingly.

stefanv:
For `external_metadata`, those fields should be defined in the configuration file so that we can check such entries as they are uploaded.

Easiest way for now may be to use the same description as [OpenAPI](https://swagger.io/docs/specification/describing-parameters/).

profjsb:
I'm definitely in favor of having application-specific/source-specific configs here.

skyportal/handlers/source.py
```diff
@@ -206,10 +206,10 @@ def post(self):
             end_date = arrow.get(data['endDate'].strip())
             q = q.filter(Source.last_detected <= end_date)
         if data['simbadClass']:
-            q = q.filter(func.lower(Source.simbad_class) ==
-                         data['simbadClass'].lower())
+            q = q.filter(func.lower(Source.external_metadata['simbad']['class'].astext)
```
stefanv: Does this work if `simbad` is not in `external_metadata`?

acrellin: @stefanv yes, this works even when `simbad` is not in `external_metadata` for some or all sources.

acrellin:
@profjsb do we still want to collapse ZTF-, Gaia-, Simbad- and TNS-specific fields into a JSONB column?

dannygoldstein:
@acrellin yes, reviving this conversation.

acrellin:
Closing in favor of https://github.com/skyportal/skyportal/pull/446

PR #176, opened 2019-05-22
Comment timestamps displayed with local timezone

denniscfeng:
Using moment-timezone, comment timestamps should be displayed with local timezone instead of UTC.


stefanv:
Hi @denniscfeng, this is great, thanks!

Please run `make lint-install` so that you get automated style checking of Javascript code before commits (you only need to do this once).

Then, this looks fine, but a test would be good: could you add a check to one of the comment tests to make sure the timestamp looks reasonable?

skyportal/tests/frontend/test_sources.py
```diff
@@ -33,6 +34,16 @@ def test_comments(driver, user, public_source):
     comment_box.send_keys(comment_text)
     driver.find_element_by_css_selector('[type=submit]').click()
     driver.wait_for_xpath(f'//div[text()="{comment_text}"]')
+    driver.wait_for_xpath('//span[contains(@class,"commentTime")]')
+    timestamp_text = driver.find_element(By.XPATH, 
+                        '//span[contains(@class,"commentTime")]').text
+    day_text, time_text = timestamp_text.split(" at ", 2)
+    timestamp = datetime.strptime(time_text, '%I:%M %p')
+    day = datetime.today()
+    if day_text == 'Yesterday':
```
stefanv: What happens when the timestamp is more than one day in the past?

package.json
```diff
@@ -7,12 +7,13 @@
...
   "dependencies": {
     "bokehjs": "0.12.9",
-    "check-dependencies": "1.1.0",
+    "check-dependencies": "^1.1.0",
     "classnames": "2.2.6",
     "css-loader": "0.28.4",
     "exports-loader": "0.6.4",
     "imports-loader": "0.7.1",
     "moment": "2.20.1",
+    "moment-timezone": "^0.5.25",
```
stefanv: Does this pass the linting test for you?  I thought we force pinned versions.

skyportal/tests/frontend/test_sources.py
```diff
@@ -33,6 +34,16 @@ def test_comments(driver, user, public_source):
...
+        day = day.AddDays(-1)
+    timestamp = datetime.combine(day, timestamp.time())
+    assert(abs((datetime.now() - timestamp).seconds) < 300)
```
acrellin: Will this ever be longer than a few seconds?  How about setting this to 10 seconds?

package.json
acrellin: Yeah, the linter should have caught this, but just remove the caret to pin the versions.

denniscfeng: Fixed!

skyportal/tests/frontend/test_sources.py
denniscfeng: Okay, I readjusted it to 30 seconds (since it seems to take slightly longer on my VM--which is on my laptop--and I wanted to make sure it would pass)

PR #179, opened 2019-06-05
PS1 reticle adjustment

denniscfeng:
Made new centered crosshair image to resolve PS1 reticle being off by a little bit


stefanv:
It looks like this image is 199x199, with the crosshair centered on 99, instead of on 100?

stefanv:
What is the status of this PR, @denniscfeng?

denniscfeng:
This pull request should be ready to merge. The current version of the crosshair--when rescaled on the page--is centered, despite not using any centering using CSS
image

Issue #121, opened 2019-01-31
DECam link goes to the wrong coordinates


dannygoldstein:
They appear to be random

stefanv:
Should be fixed by #180.

PR #180, opened 2019-06-06
Fix survey link URL formatting issue

denniscfeng:
Removed '{ }s' formatting from survey urls, addressing https://github.com/skyportal/skyportal/issues/121


stefanv:
@denniscfeng Did you get a chance to wrap this up?  If not, perhaps @acrellin could take a look as well.

denniscfeng:
I think the formatting for the URL's should be fixed. However, a lot of the survey links are broken/don't lead to anything, so we should look at if some of the links should be updated/removed.

acrellin:
@denniscfeng can you rebase on master so the tests here will pass? Otherwise looks good

Issue #187, opened 2019-06-19
photometry data model  is ambiguous


dannygoldstein:
The current photometry data model is ambiguous as it does not specify a magnitude system (eg., AB, Vega, etc.)

I recommend adopting Kyle Barbary's model for photometric data:

https://sncosmo.readthedocs.io/en/v2.0.x/photdata.html

PR #188, opened 2019-06-19
Photometry

dannygoldstein:
Addresses #187

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -242,6 +242,10 @@ class Photometry(Base):
     mag = sa.Column(sa.Float)
     e_mag = sa.Column(sa.Float)
     lim_mag = sa.Column(sa.Float)
+    
+    ra = sa.Column(sa.Float)
+    dec = sa.Column(sa.Float)
+
     filter = sa.Column(sa.String)  # TODO Enum?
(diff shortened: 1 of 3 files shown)


acrellin:
@dannygoldstein looks like this PR and #182 have some overlap -- what's the status of these? Should we close and try to get the other in? thanks!

skyportal/models.py
```diff
@@ -239,9 +245,18 @@ class Photometry(Base):
     mjd = sa.Column(sa.Float)  # mjd date
     time_format = sa.Column(sa.String, default='iso')
     time_scale = sa.Column(sa.String, default='utc')
-    mag = sa.Column(sa.Float)
-    e_mag = sa.Column(sa.Float)
+
```
dannygoldstein: I think the only things that need to be merged here are the change to this file, but I would also keep mag + e_mag as fields.

dannygoldstein: ZTF alerts do not provide flux/fluxerr, but offline forced photometry does. Best to have all of these as fields, and require either mag or flux, but not both.

Danny

acrellin:
Closing, as we're not going with q3c, and moving updates to photometry data model to https://github.com/skyportal/skyportal/pull/430

PR #212, opened 2019-08-05
Infer source group(s) from user/token group membership 

acrellin:
Infer source group(s) from user/token group membership if IDs not explicitly specified in POST handler

### DIFF
--- a/skyportal/handlers/source.py
+++ b/skyportal/handlers/source.py
@@ -91,7 +91,7 @@ class SourceHandler(BaseHandler):
     def post(self):
         """
         ---
-        description: Upload a source
+        description: Upload a source. If group_ids is not specified, the user or token's groups will be used.
         parameters:
           - in: path
             name: source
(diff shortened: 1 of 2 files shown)


skyportal/handlers/source.py
```diff
@@ -114,7 +114,12 @@ def post(self):
         try:
             group_ids = data.pop('group_ids')
```
stefanv: Is there any verification done to ensure that a user doesn't add sources to a group they don't belong to?

skyportal/handlers/source.py
```diff
@@ -114,7 +114,12 @@ def post(self):
...
         except KeyError:
-            return self.error("Missing required fields: group_ids is a required field.")
+            if self.current_user.groups:
+                group_ids = [g.id for g in self.current_user.groups]
+            else:
+                return self.error("group_ids is a required field when user/token "
```
stefanv: I can't quite parse this error message

acrellin:
@stefanv thanks for the feedback. I've added a check to ensure a user cannot add sources to groups to which they don't belong, and have also changed the error message.

skyportal/handlers/source.py
```diff
@@ -111,10 +111,14 @@ def post(self):
         """
         data = self.get_json()
         schema = Source.__schema__()
+        user_group_ids = [g.id for g in self.current_user.groups]
+        if not user_group_ids:
+            return self.error("You must belong to one or more groups before "
+                              "you can add sources.")
         try:
-            group_ids = data.pop('group_ids')
+            group_ids = [id for id in data.pop('group_ids') if id in user_group_ids]
```
stefanv: What happens when `group_ids` is empty?

If the group is added to all user groups by default, we should document that.

acrellin: Okay -- document that where?  In the docstring description?

acrellin: Do you mean what happens if a user specifies `group_ids = []` in the request data?  In that case, it returns with an error message.

stefanv: No, more the situation where the user specifies group IDs that either don't exist or where the user doesn't have permission to post to, so that we end up with no valid IDs.

Documentation should go with the endpoint docs, so it appears in our API docs.

acrellin: In that case, we return with the error message "Invalid group_ids field. Please specify at least one valid group ID that you belong to."

acrellin:
@stefanv good to go here?

PR #219, opened 2019-08-22
Replace PS1 cutout image with DESI DR8

### LINKED ISSUE #217: Use DESI DR8 for postage stamps
Instead of PS1, the default postage stamp to grab should be DESI DR8. E.g.

curl -o "name.jpeg" "http://legacysurvey.org/viewer/jpeg-cutout?ra=334.62567&dec=0.20588&size=512&layer=dr8&pixscale=0.262&bands=grz"

where the (pixel) size should be chosen to correspond the size of the discovery image postage stamp area.

acrellin:
Closes #217

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -149,10 +149,10 @@ class Source(Base):
         sdss_thumb = Thumbnail(photometry=self.photometry[0],
                                public_url=self.get_sdss_url(),
                                type='sdss')
-        ps1_thumb = Thumbnail(photometry=self.photometry[0],
-                              public_url=self.get_panstarrs_url(),
-                              type='ps1')
-        DBSession().add_all([sdss_thumb, ps1_thumb])
+        dr8_thumb = Thumbnail(photometry=self.photometry[0],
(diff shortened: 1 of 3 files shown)


acrellin:
@stefanv this should be good to go!

skyportal/models.py
```diff
@@ -147,36 +147,24 @@ class Source(Base):
...
-                              type='ps1')
-        DBSession().add_all([sdss_thumb, ps1_thumb])
+        dr8_thumb = Thumbnail(photometry=self.photometry[0],
+                              public_url=self.desi_dr8_url(),
+                              type='dr8')
+        DBSession().add_all([sdss_thumb, dr8_thumb])
         DBSession().commit()
 
-    def get_sdss_url(self):
+    def sdss_url(self):
```
stefanv: These still need an `@property` above.  And then removal of the `()` in all calls elsewhere.

acrellin: Sure, done.

stefanv:
In it goes!

PR #230, opened 2019-09-26
Clear auth token cookie on become_user and logout; update frontend tests

### LINKED ISSUE #227: Investigate why websocket messages are not reaching front-end during tests
When resolved, remove instances of `driver.refresh()` in tests that act as temporary workarounds.

acrellin:
Closes #227

### DIFF
--- a/skyportal/tests/frontend/test_tokens.py
+++ b/skyportal/tests/frontend/test_tokens.py
@@ -14,19 +14,20 @@ def test_add_token(driver, user, public_group):
     group_select.select_by_value(str(public_group.id))
     driver.wait_for_xpath('//input[@name="name"]').send_keys(token_name)
     driver.wait_for_xpath('//input[@value="Generate Token"]').click()
-    driver.refresh()
     driver.wait_for_xpath(f'//td[contains(.,"{token_name}")]')
 
 
 def test_delete_token(driver, user, public_group,
                       view_only_token_created_by_fulluser):
+    token_id = view_only_token_created_by_fulluser
     driver.get(f'/become_user/{user.id}')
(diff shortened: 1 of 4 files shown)


skyportal/tests/frontend/test_tokens.py
```diff
@@ -14,19 +14,20 @@ def test_add_token(driver, user, public_group):
...
+    token_id = view_only_token_created_by_fulluser
     driver.get(f'/become_user/{user.id}')
     driver.get('/profile')
     driver.wait_for_xpath(f'//input[@value="{view_only_token_created_by_fulluser}"]')
     driver.wait_for_xpath('//a[contains(text(),"Delete")]').click()
-    driver.refresh()
+    driver.wait_for_xpath(f'//div[contains(text(),"Token {token_id} deleted.")]')
     try:
-        driver.wait_for_xpath(f'//input[@value="{view_only_token_created_by_fulluser}"]')
+        driver.wait_for_xpath(f'//input[@value="{token_id}"]',
```
stefanv: @acrellin We'd prefer not to have a timeout-based test here.  Grab the table and ensure the token id is missing.

acrellin: @stefanv this is what we decided was the best approach when we were looking at this together, remember? Note that the timeout is set to 0, so it doesn't wait

acrellin: @stefanv it waits for the notification to display then immediately (without a wait period) makes sure this element doesn't exist

stefanv:
Missed the `timeout=0`, all good!

Issue #207, opened 2019-07-19
permissions errors come back as server errors


dannygoldstein:
When I try to comment as a user without the "Comment" ACL I get a 400 response saying 

TypeError: get() missing 1 required positional argument: 'comment_id'
! App Error: get() missing 1 required positional argument: 'comment_id'
WARNING:tornado.access:400 GET /api/comment (127.0.0.1) 12.35ms

This should return a more informative error that says something like "permission denied" or something like that

PR #231, opened 2019-09-26
Only render add comment component for users who have comment permissions; add view-only user role; add test

acrellin:
Closes https://github.com/skyportal/skyportal/issues/207

### DIFF
--- a/skyportal/tests/frontend/test_sources.py
+++ b/skyportal/tests/frontend/test_sources.py
@@ -84,3 +84,16 @@ def test_download_comment_attachment(driver, user, public_source):
         assert l.split('\n')[0] == 'wavelength,flux,instrument_id'
     finally:
         os.remove(fpath)
+
+
+
+def test_view_only_user_cannot_comment(driver, view_only_user, public_source):
+    driver.get(f"/become_user/{view_only_user.id}")
+    driver.get(f"/source/{public_source.id}")
+    driver.wait_for_xpath(f'//div[text()="{public_source.id}"]')
+    try:
+        driver.wait_for_xpath('//input[@name="comment"]', timeout=0)
+    except:
+        pass
+    else:
+        raise Exception('Comment form is visible to view-only user, but should be hidden.')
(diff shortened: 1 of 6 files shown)


skyportal/tests/frontend/test_sources.py
```diff
@@ -84,3 +84,16 @@ def test_download_comment_attachment(driver, user, public_source):
...
+def test_view_only_user_cannot_comment(driver, view_only_user, public_source):
+    driver.get(f"/become_user/{view_only_user.id}")
+    driver.get(f"/source/{public_source.id}")
+    driver.wait_for_xpath(f'//div[text()="{public_source.id}"]')
+    try:
+        driver.wait_for_xpath('//input[@name="comment"]', timeout=0)
+    except:
+        pass
+    else:
+        raise Exception('Comment form is visible to view-only user, but should be hidden.')
```
stefanv: Nicely done.

This seems to become a pattern, so we can also add a little shim: `driver.assert_missing_xpath(XPATH, error message)`?  Will leave that for another PR.

static/js/components/CommentList.jsx
```diff
@@ -50,7 +51,10 @@ const CommentList = () => {
     <div className={styles.comments}>
       {items}
       <br />
-      <CommentEntry addComment={addComment} />
+      {
```
stefanv: Curious about logic; would a user not sometimes have comment rights *only on certain sources*?

acrellin: Hmm, good question... We don't currently have that concept, but it's something we should consider adding. Currently the endpoint handler just checks whether the user has the "Comment" ACL: `@permissions(['Comment'])` - I'll open an issue to look into adding per-source (or per-group) permissions

skyportal/tests/frontend/test_sources.py
acrellin: Good call - after this one gets merged, I'll add that to baselayer and update all the skyportal tests (including this one) accordingly.

PR #234, opened 2019-09-26
Use driver.wait_for_xpath_missing in front-end tests

acrellin:


stefanv:
Does this wait until the element disappears then?

acrellin:
@stefanv by default, this checks every 0.5 seconds until the specified element no longer exists, up to (configurable) 5 seconds. But if the element is already gone when this is called, it returns immediately.

acrellin:
https://seleniumhq.github.io/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.wait.html

stefanv:
The naming is just awkward.  Would something like `wait_for_xpath_to_disappear`, e.g., not read more logically?

acrellin:
Sure, I'll change it to that!

stefanv:
We may have to check that we don't break Cesium as well.  Thanks!

acrellin:
Done! I'll update cesium as well.

acrellin:
@stefanv turns out we were never using that in `cesium_web`, so no issues there

Issue #193, opened 2019-06-20
`sources_table_empty` is for frontend only


stefanv:
Remove from `/api/sysinfo` into private API.

acrellin:
@stefanv do you want there to be a separate endpoint for this? `/api/internal/sources_table_empty`?

stefanv:
Sure, anything `/internal/` works.  But, do we really need to know this?  We can also check if the user has access to >= 1 source, and that should accomplish the same.  Not sure whether there is a *cheap* way of doing that in the database, though, although this seems to be exactly what we'd need:

https://stackoverflow.com/questions/6587879/how-to-elegantly-check-the-existence-of-an-object-instance-variable-and-simultan

PR #236, opened 2019-09-30
Move source_table_empty check to separate handler; update front-end and tests

acrellin:
Closes https://github.com/skyportal/skyportal/issues/193 
Also updates ESLint-related package versions & de-lints JS sources


stefanv:
Let's rename API endpoint to dbinfo and include postgres version. Feel free to merge if it passes then.

PR #238, opened 2019-10-02
Allow POSTing thumbnails along with photometry data in /api/photometry handler

### LINKED ISSUE #215: Allow uploading of thumbnails with photometry
Currently, there is no way to add a thumbnail to a photometry measurement.

acrellin:
Closes #215

### DIFF
--- a/skyportal/handlers/photometry.py
+++ b/skyportal/handlers/photometry.py
@@ -1,6 +1,7 @@
 import os
 import io
 import base64
+from pathlib import Path
 import tornado.web
 from astropy.time import Time
 from sqlalchemy.orm import joinedload
@@ -73,8 +74,13 @@ class PhotometryHandler(BaseHandler):
             ids.append(p.id)
         if 'thumbnails' in data:
             for thumb in data['thumbnails']:
+                basedir = Path(os.path.dirname(__file__))/'..'/'..'
+                if os.path.abspath(basedir).endswith('skyportal/skyportal'):
+                    basedir = basedir/'..'
                 file_uri = os.path.abspath(
-                    f'static/thumbnails/{source.id}_{thumb["type"]}.png')
+                    basedir/f'static/thumbnails/{source.id}_{thumb["type"]}.png')
+                if not os.path.exists(os.path.dirname(file_uri)):
+                    (basedir/'static/thumbnails').mkdir(parents=True)
                 file_bytes = base64.b64decode(thumb['data'])
                 im = Image.open(io.BytesIO(file_bytes))
                 if im.format != 'PNG':
@@ -83,11 +89,15 @@ class PhotometryHandler(BaseHandler):
                     return self.error('Invalid thumbnail size. Only 200x200 px allowed.')
                 with open(file_uri, 'wb') as f:
                     f.write(file_bytes)
-                t = Thumbnail(type=thumb['type'],
-                              photometry_id=ids[0],
-                              file_uri=file_uri,
-                              public_url=f'/static/thumbnails/{thumb["fname"]}')
-                DBSession.add(t)
+                try:
+                    t = Thumbnail(type=thumb['type'],
+                                  photometry_id=ids[0],
+                                  file_uri=file_uri,
+                                  public_url=f'/static/thumbnails/{thumb["fname"]}')
+                    DBSession.add(t)
+                except TypeError:
+                    return self.error('Invalid thumbnail type. Please refer to '
+                                      'API docs for a list of allowed types.')
         DBSession().commit()
 
         return self.success(data={"ids": ids})
--- a/skyportal/handlers/photometry.py
+++ b/skyportal/handlers/photometry.py
@@ -20,8 +20,41 @@ class PhotometryHandler(BaseHandler):
         description: Upload photometry
         parameters:
           - in: path
-            name: photometry
-            schema: Photometry
+            name: time_format
+            schema:
+              type: string
+          - in: path
+            name: time_scale
+            schema:
+              type: string
+          - in: path
+            name: source_id
+            schema:
+              type: string
+          - in: path
+            name: instrument_id
+            schema:
+              type: integer
+          - in: path
+            name: time
+            schema:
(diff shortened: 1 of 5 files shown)


skyportal/handlers/photometry.py
```diff
@@ -1,10 +1,14 @@
+import base64
 import tornado.web
 from astropy.time import Time
 from sqlalchemy.orm import joinedload
 from marshmallow.exceptions import ValidationError
+import os
```
stefanv: Nitpick, but lets keep the Python core imports together at the top.

skyportal/handlers/photometry.py
```diff
@@ -68,6 +71,20 @@ def post(self):
             DBSession().add(p)
             DBSession().flush()
             ids.append(p.id)
+        if 'thumbnails' in data:
```
stefanv: I think we need to check a bit more carefully here that only thumbnails of allowed types are provided.  Also check size of resulting images (we should only allow within a certain range).  Do not allow the user to specify output filenames under any circumstances!

We should also note somewhere that this attached thumbnails to photometry 0.

acrellin: We _do_ check the thumbnail type - see line 79 below.

stefanv: No, not the file type, I mean the 'NEW', 'SD1', etc. — the 'type' dictionary entry.

acrellin: Ah I see - that's automatically taken care of by sqlalchemy.  If we try to create a thumbnail with a type not listed in the class definition, it errors.

acrellin: Just pushed up changes to address your comments - wanna take another look?

stefanv: I guess those errors are fine, although they may not provide the user with very sensible error messages. Thoughts?

skyportal/handlers/photometry.py
```diff
@@ -68,6 +71,23 @@ def post(self):
...
+            for thumb in data['thumbnails']:
+                file_uri = os.path.abspath(
+                    f'static/thumbnails/{source.id}_{thumb["type"]}.png')
```
stefanv: Again, `type` is provided by the user, so I think we should just generate our own filename here.  If you want to use type in there, I would first verify the type.

skyportal/handlers/photometry.py
```diff
@@ -68,6 +71,23 @@ def post(self):
...
+                file_bytes = base64.b64decode(thumb['data'])
+                im = Image.open(io.BytesIO(file_bytes))
+                if im.format != 'PNG':
+                    return self.error('Invalid thumbnail image type. Only PNG are supported.')
+                if im.size != (200, 200):
```
stefanv: Genuine question: how restrictive do we want to be with thumbnails?  Perhaps a range check here of between 100 and 500?

stefanv: This should also be documented in the API docs.

acrellin:
@stefanv The thumbnail size can now be between (100, 100) and (500, 500) - and we first validate the type before writing to disk.

acrellin:
@stefanv the docs are now updated, and we also now catch the sqlalchemy error and return our own error response if the type is incorrect. Should be good to go now.

skyportal/handlers/photometry.py
```diff
@@ -15,8 +20,58 @@ def post(self):
         description: Upload photometry
         parameters:
           - in: path
-            name: photometry
-            schema: Photometry
+            name: time_format
```
stefanv: These fields still need descriptions.  No idea what are valid values for "time_format"!

skyportal/handlers/photometry.py
```diff
@@ -68,6 +122,33 @@ def post(self):
             DBSession().add(p)
             DBSession().flush()
             ids.append(p.id)
+        if 'thumbnails' in data:
+            for thumb in data['thumbnails']:
+                basedir = Path(os.path.dirname(__file__))/'..'/'..'
+                if os.path.abspath(basedir).endswith('skyportal/skyportal'):
```
stefanv: What's happening here?

skyportal/handlers/photometry.py
```diff
@@ -68,6 +122,33 @@ def post(self):
...
+                try:
+                    t = Thumbnail(type=thumb['ttype'],
+                                  photometry_id=ids[0],
+                                  file_uri=file_uri,
+                                  public_url=f'/static/thumbnails/{source.id}_{thumb["ttype"]}.png')
+                    DBSession.add(t)
+                except TypeError:
+                    return self.error('Invalid thumbnail type. Please refer to '
+                                      'API docs for a list of allowed types.')
+                with open(file_uri, 'wb') as f:
```
stefanv: This logic is incorrect: do not write file if DB call fails.

acrellin: No, if the DB call fails, it returns immediately.

stefanv: Well, if it returns a TypeError.  But who knows what all else can go wrong.  So we should probably include a catch-all as well.

skyportal/handlers/photometry.py
acrellin: This is no longer used

skyportal/handlers/photometry.py
acrellin: On Travis, for some reason `Path(os.path.dirname(__file__))/'..'/'..'` returns something like `/home/user/skyportal/skyportal` instead of the expected `/home/user/skyportal` (which is what it returns locally), so I added this check

skyportal/handlers/photometry.py
acrellin: Sure thing -- done!

skyportal/handlers/photometry.py
acrellin: Done

acrellin:
@stefanv I have addressed all of your comments -- anything else you'd like to see here before it goes in?

PR #240, opened 2019-10-04
Implement Thumbnail API

acrellin:
Allows thumbnails to be added via new `/thumbnail` API endpoint.  Either a source ID or a photometry ID can be supplied in the request.  If the former, the first photometry point associated with the source is used.


skyportal/handlers/thumbnail.py
```diff
@@ -0,0 +1,129 @@
...
+                          public_url=f'/static/thumbnails/{source.id}_{data["ttype"]}.png')
+            DBSession.add(t)
+        except TypeError:
+            return self.error('Invalid thumbnail type. Please refer to '
+                              'API docs for a list of allowed types.')
+        except Exception as e:
+            return self.error(f'Error creating thumbnail: {str(e)}. Please check '
+                              'submitted values against API docs.')
+        with open(file_uri, 'wb') as f:
+            f.write(file_bytes)
```
acrellin: @stefanv  do you mean factoring this part out so it can be called in both handlers?

stefanv: Yes, I think ideally we do not want duplicated logic like this floating around.

stefanv: But there may be a way to call the one post method from another.

skyportal/handlers/thumbnail.py
```diff
@@ -127,3 +100,35 @@ def delete(self, thumbnail_id):
...
+    file_uri = os.path.abspath(
+        basedir/f'static/thumbnails/{source_id}_{thumbnail_type}.png')
+    if not os.path.exists(os.path.dirname(file_uri)):
+        (basedir/'static/thumbnails').mkdir(parents=True)
+    file_bytes = base64.b64decode(thumbnail_data)
+    im = Image.open(io.BytesIO(file_bytes))
+    if im.format != 'PNG':
+        return self.error('Invalid thumbnail image type. Only PNG are supported.')
+    if not (100, 100) <= im.size <= (500, 500):
+        return self.error('Invalid thumbnail size. Only thumbnails '
```
stefanv: I don't think this will work.

skyportal/handlers/thumbnail.py
```diff
@@ -113,22 +115,23 @@ def create_thumbnail(thumbnail_data, thumbnail_type, source_id, photometry_obj):
     file_bytes = base64.b64decode(thumbnail_data)
     im = Image.open(io.BytesIO(file_bytes))
     if im.format != 'PNG':
-        return self.error('Invalid thumbnail image type. Only PNG are supported.')
+        raise TypeError('Invalid thumbnail image type. Only PNG are supported.')
```
stefanv: This should be `ValueError`.

skyportal/handlers/thumbnail.py
```diff
@@ -113,22 +115,23 @@ def create_thumbnail(thumbnail_data, thumbnail_type, source_id, photometry_obj):
...
-    except TypeError:
-        return self.error('Invalid thumbnail type. Please refer to '
+        DBSession.flush()
+    except LookupError:
+        raise LookupError('Invalid thumbnail type. Please refer to '
                           'API docs for a list of allowed types.')
     except Exception as e:
-        return self.error(f'Error creating thumbnail: {str(e)}. Please check '
-                          'submitted values against API docs.')
+        raise Exception(f'Error creating thumbnail: {str(e)}. Please check '
```
stefanv: Perhaps better to just re-raise this exception.

skyportal/handlers/thumbnail.py
acrellin: Ah, right!

skyportal/handlers/thumbnail.py
```diff
@@ -0,0 +1,136 @@
...
+            try:
+                phot = source.photometry[0]
+            except IndexError:
+                return self.error('Specified source does not yet have any photometry data.')
+        else:
+            return self.error('One of either source_id or photometry_id are required.')
+        try:
+            t = create_thumbnail(data['data'], data['ttype'], source.id, phot)
+        except Exception as e:
+            return self.error(str(e))
```
stefanv: Heh, now that I think about it: what happens when exceptions get raised from handlers?  Perhaps we already explicitly call `error` on them.  Either way, this is fine, but would be good to know!

acrellin:
@stefanv thanks for the feedback; I made all of the changes you suggested, and fixed the failing test. Merging.

skyportal/handlers/thumbnail.py
```diff
@@ -0,0 +1,136 @@
...
+        t = Thumbnail(type=thumbnail_type,
+                      photometry=photometry_obj,
+                      file_uri=file_uri,
+                      public_url=f'/static/thumbnails/{source_id}_{thumbnail_type}.png')
+        DBSession.add(t)
+        DBSession.flush()
+    except LookupError:
+        raise LookupError('Invalid thumbnail type. Please refer to '
+                          'API docs for a list of allowed types.')
+    except Exception as e:
```
stefanv: Hrm, I guess we could also just remove the "except" clause here with the same effect.

acrellin:
@stefanv good point. Done -- see https://github.com/skyportal/skyportal/commit/57dde9521c33a11f3f332525dd84354b12b65e83

stefanv:
Beautiful, thanks Ari!

Issue #247, opened 2019-10-09
"photometry" not in data structure


AlexGKim:
When I "get" a source, the structure does not have an element with key "photometry".

PR #248, opened 2019-10-09
Return photometry in GET /api/sources handler; add test

acrellin:
Closes https://github.com/skyportal/skyportal/issues/247


stefanv:
Should we serve this as part of the source, or under `/api/source/14gqr/photometry`?

acrellin:
@stefanv are you leaning towards the latter?

stefanv:
Probably, since the photometry data might be quite large, and when you query the source you may only be interested in meta-data like position etc.  Also, if we're going to embed photometry, we also have to do the spectra, and then it feels like the right call to factor them both out to a new endpoint.

skyportal/handlers/source.py
```diff
@@ -237,3 +237,34 @@ def post(self):
...
+    def get(self, source_id):
+        """
+        ---
+        description: Retrieve a source's photometry
+        parameters:
+        - in: path
+          name: source_id
+          required: false
+          schema:
+            type: integer
```
stefanv: Probably not an integer?

skyportal/handlers/source.py
```diff
@@ -237,3 +237,34 @@ def post(self):
...
+            content:
+              application/json:
+                schema: Error
+        """
+        info = {}
+        source = Source.get_if_owned_by(source_id, self.current_user)
+        if not source:
+            return self.error('Invalid source ID, or inadequate permissions to '
+                              'access source.')
+        info['photometry'] = source.photometry
```
stefanv: No big deal, but could probably just be `info = {'photometry': source.photometry}` instead of instantiating dict separately, or could even just put this inside the return statement.

acrellin: Done!

skyportal/handlers/source.py
acrellin: Right - fixed!

Issue #243, opened 2019-10-09
Instrument / Telescope through http request


AlexGKim:
I understand that being able to add instrument/telescopes through http requests is an intended feature.  I do not see this feature in the API docs.   Is this a feature that is implemented but is not yet documented?  Otherwise, I would request that this gets implemented sooner than later.

acrellin:
I'll take care of this today

PR #252, opened 2019-10-09
Implement Telescope and Instrument API

acrellin:
Closes https://github.com/skyportal/skyportal/issues/243

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -9,7 +9,7 @@ from skyportal.handlers import (SourceHandler, CommentHandler, GroupHandler,
                                 PhotometryHandler, TokenHandler,
                                 FilterSourcesHandler, SysInfoHandler,
                                 UserHandler, SpectrumHandler, ThumbnailHandler,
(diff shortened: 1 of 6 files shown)


profjsb:
How is the permissioning handled for POST/GET with respect to groups? That is, can anybody with write permission on the DB create a new telescope/instrument? And can they see other people's telescopes/instruments?

stefanv:
@profjsb Currently, instruments are accessible by all.  Should these be tied to a group?

profjsb:
> @profjsb Currently, instruments are accessible by all. Should these be tied to a group?

I think so -- it would be a mild form of IP leakage if one (rival) group could know what instruments the other group is using to observe. The default could be that everyone belongs to a "global" group and that their write/read permissions apply to "global" as well as any specific group(s) they are in. But this would still need to enforce the per-group permissioning logic discussed above.

acrellin:
We're doing a big API security check/overhaul soon - shall we include any changes to the telescope permissions there so we can get this in?

PR #253, opened 2019-10-10
Set FLAGS to point to config.yaml

stefanv:
See also https://github.com/cesium-ml/baselayer/pull/66


acrellin:
@stefanv does this need to be pinned to the newer baselayer?

stefanv:
Yes, let's do that to make sure.

Issue #255, opened 2019-10-10
post source with group_id not assigning groups


AlexGKim:
I uploaded a new source using the `group_id=[3,4]` parameter but only the group of my token is assigned.

acrellin:
It should be `group_ids`, not `group_id`

acrellin:
e.g. https://github.com/skyportal/skyportal/blob/master/skyportal/tests/api/test_sources.py#L70

acrellin:
But yes, what you describe is actually the expected behavior. Tokens cannot add sources to groups to which they do not already belong. The token needs to be associated with both groups if it is to add a source to each of them.

AlexGKim:
I see.  How do I upload a source associated with multiple groups?  Is it possible to submit multiple tokens?

acrellin:
@AlexGKim we're going to push an update soon to allow a user to create a token that belongs to any number of the groups that the user currently belongs to.

AlexGKim:
Got it.  I can wail until then.

AlexGKim:
From the documentation it seems that the intent is for groups to be associated with sources, say based on their observing program.  Then the user must be assigned which sources she can see.  Call this Scenario A.

On the other hand, groups could be associated with users, say by the collaboration that they belong to.  Then the source must be assigned which users it can be seen by.  This is Scenario B.

Suppose a new set of sources are introduced:
Scenario A: All users that have access to the new source must be modified to add the new group.
Scenario B: New entries of this kind of source are initialized with the users that have access.

Suppose a new collaboration is introduced:
Scenario A: Every new user is initialized with the sources to which they have access.
Scenario B: All sources that can be seen by the new users must be modified to add the new group.

Although this seems symmetric, I think that Scenario A, modifying users to add the new group, will be the logistically difficult.  Determining the provenance of a user is harder than a source.

I don't have a strong opinion on which scenario is better, but since but users AND sources can belong to multiple groups, it seems that the design is such that either scenario is possible.

stefanv:
@AlexGKim The relationship is roughly as you describe, but in the other direction: each user belongs to one or more groups, and sources are visible by members from certain groups.

So, e.g., you can have a group for one collaboration, and all sources that get added to that group will be visible to all members of the group.

A group is a proxy for a subset of users, so that we do not have to do individual access control to sources.

PR #258, opened 2019-10-14
Add groups property to tokens

acrellin:
`Token.groups` now points to `Token.created_by.groups` for simplified token group membership logic.

Tokens now belong to all groups that the creating user is a member of.

Closes https://github.com/skyportal/skyportal/issues/255


acrellin:
@stefanv where are we on this?

stefanv:
@acrellin So, did the investigation into the usage of a key relation fail?

acrellin:
@stefanv Yes, I couldn't find a better way to access the creating user's group list via the token.

PR #261, opened 2019-10-15
Style searchbox and table

jialin-wu-02:
1. Using css flex box to style the search box.
2. Update the style of the table.
3. Move the "# of source displaying" into the `SourceList` component.


stefanv:
Thanks, Alex.  The errors are related: you can check your code formatting by doing `make lint`.

acrellin:
@jialin-wu-02  Thanks!  Please go ahead and just modify the original `SearchBox` component (and css file, if relevant) instead of adding a new `NewSearchBox` component.

jialin-wu-02:
Just passed all checks, and there is no error in the local server either.

stefanv:
Thanks, Alex, this is looking great!

I see that, currently, you have one Custom component that itself switches out to, e.g., a time formatted box if format == UTC.  We have to write the JSX anyway, so how about making each of those separate components: TimeFormatInput, etc. and then explicitly invoking them in the JSX?

jialin-wu-02:
Fixed.

static/js/components/SearchBox.jsx
```diff
@@ -4,6 +4,13 @@ import { useDispatch } from 'react-redux';
 
 import * as Actions from '../ducks/sources';
 
+import styles from "./SearchBox.css";
+
+// Custom Components:
```
acrellin: Unnecessary comment

static/js/components/CheckBox.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import styles from './CheckBox.css';
+
+
+function CheckBox({ label, name, checked, onChange, size }) {
```
acrellin: For consistency, this could be an arrow function

jialin-wu-02:
- [ ] Thumbnail jump to another row while scaling.

stefanv:
Will be re-done using Material UI now

acrellin:
Closed in favor of https://github.com/skyportal/skyportal/pull/375

PR #279, opened 2019-10-22
Changed thumbnail storage from disk to database.

glenn2763:
Updated models.py and thumbnail.py to be able to handle thumbnails as .png files under the image attribute, which replaced the old file_uri attribute. Additionally, in ThumbnailList.jsx I made it so that the "new", "ref", and "sub" thumbnails got their image data from the new pngs on database instead of reading them from a filepath. If we ever want to display other types of thumbnails the only thing that needs to be changed is their src must be set to the png image not the url or filepath (line 71).

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   },
   "dependencies": {
     "bokehjs": "0.12.9",
-    "check-dependencies": "^1.1.0",
+    "check-dependencies": "1.1.0",
     "classnames": "2.2.6",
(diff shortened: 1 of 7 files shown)


glenn2763:
load demo data and load ztf data were also changed to handle the new format of using png images instead of filepaths

glenn2763:
messed around with stuff and broke some stuff, going to submit a new PR once more ready to fix

Issue #244, opened 2019-10-09
Able to "put" photometry to a source for which I don't have group access


AlexGKim:
I was able to "put" photometry to a source for which I don't have group access though I know its id.

acrellin:
Fixed in #293

PR #293, opened 2019-10-31
Add API security checks

### LINKED ISSUE #155: Security review
- [ ] Check access / deletion of all objects for proper access rights
  - [ ] https://github.com/skyportal/skyportal/issues/244
- [ ] Launch on 127.0.0.1 instead of 0.0.0.0 if debug mode is enabled

acrellin:
Ensures that tokens/users initiating requests have requisite permissions (group membership and ACLs) before creating/reading/updating/deleting entries for the following endpoints:
- `source`
- `photometry`
- `spectrum`
- `thumbnail`
- `comment`
- `group`
- `telescope`
- `instrument`

Also adds group membership for telescopes (and thereby their child instruments) for access control.

Closes #155 and #244

### DIFF
--- a/skyportal/handlers/comment.py
+++ b/skyportal/handlers/comment.py
@@ -25,6 +25,10 @@ class CommentHandler(BaseHandler):
                 schema: SingleComment
         """
         comment = Comment.query.get(comment_id)
+        if comment is None:
+            return self.error('Invalid comment ID.')
+        # Ensure user/token has access to parent source
+        s = Source.get_if_owned_by(comment.source.id, self.current_user)
         if action == 'download_attachment':
             self.set_header(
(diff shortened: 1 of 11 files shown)


stefanv:
This is good, Ari, and you should merge whenever you feel it is ready.

It again makes me realize that we would benefit from a generic DatabaseMapper that allows you to do the default GET/PUT/UPDATE etc. directly to the database.  And then we can customize that class wherever more specific functionality is needed.

I think you responded positively when I mentioned this idea last time; thoughts?

acrellin:
@stefanv Yes, I also think that's a good idea, as it would avoid a lot repeated code.  However, the security checks in this PR might make it harder to customize generic handlers.  I'll merge this and open a new issue for addressing that going forward.

Issue #246, opened 2019-10-09
Refactor load_demo_data to use API calls


stefanv:
This will also ensure that we have all the necessary API endpoints

acrellin:
@stefanv Currently we have no way of creating Super admin users via the users API - how should we handle permissions when there aren't any users in the DB yet?  Or should initial user creation still be done via direct `sqlalchemy` calls in `load_demo_data`, and then all subsequent additions be done via API calls?

stefanv:
In order to use the API, you need a token, and you can't make a token without a user.  So, you have to start somewhere!  I think the current way we handle it—initializing the first user via a script—is reasonable.

PR #295, opened 2019-11-02
Use API calls in load_demo_data whenever possible

acrellin:
Closes https://github.com/skyportal/skyportal/issues/246


tools/load_demo_data.py
```diff
@@ -34,80 +37,191 @@
...
-                              public_url=f'/static/thumbnails/{fname}')
-                DBSession().add(t)
-                shutil.copy(basedir/f'skyportal/tests/data/{fname}', basedir/'static/thumbnails/')
-
-            s.add_linked_thumbnails()
+    with status("Launching web app & executing API calls"):
+        web_client = subprocess.Popen(['make', 'run'],
+                                      cwd=basedir, preexec_fn=os.setsid)
+
+        server_url = f"http://localhost:{cfg['ports.app']}"
```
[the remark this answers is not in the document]
acrellin: @stefanv this is consistent with every other instance of URL construction in our app:  `f'http://localhost:{cfg["ports.app"]}'`

acrellin:
@stefanv thanks for the feedback - I've addressed all of your comments/suggestions!

tools/load_demo_data.py
```diff
@@ -34,80 +38,168 @@
...
+
+                data = assert_post('instrument',
+                                   data={'name': 'P60 Camera',
+                                         'type': 'phot',
+                                         'band': 'optical',
+                                         'telescope_id': telescope1_id})
+                instrument1_id = data['data']['id']
+
+                data = assert_post('telescope',
+                                            data={'name': 'Nordic Optical Telescope',
```
stefanv: Indentation

tools/load_demo_data.py
```diff
@@ -34,80 +38,168 @@
...
+
+                    data = assert_post('sources', data=source_info)
+                    assert data['data']['id'] == source_info['id']
+
+                    for comment in comments:
+                        data = assert_post('comment',
+                                           data={'source_id': source_info['id'],
+                                                 'text': comment})
+
+                    phot_file = os.path.join(os.path.dirname(os.path.dirname(__file__)),
```
stefanv: I know you enjoy using `pathlib.Path` — maybe that would look nice here.  Your call.  Can also just be `skyportal/tests/data/phot.csv`.

tools/load_demo_data.py
```diff
@@ -34,80 +38,168 @@
...
+                            open(os.path.abspath(fpath), 'rb').read())
+                        data = assert_post('thumbnail',
+                                           data={'source_id': source_info['id'],
+                                                 'data': thumbnail_data,
+                                                 'ttype': ttype
+                                           })
+
+                    source = Source.query.get(source_info['id'])
+                    source.add_linked_thumbnails()
+        except Exception as e:
```
stefanv: I'd remove this except clause, since it doesn't seem to do anything useful.

tools/load_demo_data.py
```diff
@@ -34,80 +38,168 @@
...
+                                           data={'source_id': source_info['id'],
+                                                 'text': comment})
+
+                    phot_file = os.path.join(os.path.dirname(os.path.dirname(__file__)),
+                                             'skyportal', 'tests', 'data',
+                                             'phot.csv')
+                    phot_data = pd.read_csv(phot_file)
+
+                    data = assert_post('photometry',
+                                       data={'source_id': source_info['id'],
```
stefanv: I'm actually quite curious what [black](https://pypi.org/project/black/) would recommend in this case.

PR #297, opened 2019-11-05
Changed thumbnail storage from disk to database.

glenn2763:
Updated models.py and thumbnail.py to be able to handle thumbnails as .png files under the image attribute, which replaced the old file_uri attribute. Additionally, in ThumbnailList.jsx I made it so that the "new", "ref", and "sub" thumbnails got their image data from the new pngs on database instead of reading them from a filepath. If we ever want to display other types of thumbnails the only thing that needs to be changed is their src must be set to the png image not the url or file path.


dannygoldstein:
During the dev meeting today we discussed adding a `bytes` column to the `thumbnails` table and a `hybrid_property` (`render_object`?) that returns an HTML element to render:

  1. The bytes as a bytearray, if bytes is not NULL (see https://stackoverflow.com/questions/20756042/how-to-display-an-image-stored-as-byte-array-in-html-javascript)
  2. The public_url

profjsb:
While storing the postagestamps in the DB will make the DB more portable (and recoverable), I'm not a fan of having such data spinning given the size and the expense. The majority of stamps will only get looked at a few times if at all and there are zero use cases where the contents of the stamps will be queried. My vote is to have stamps saved to a local/common storage that is visible by all worker nodes. Having a `hybrid_property` or the like that allows the display of stamps from disk or public URL would be preferred.

stefanv:
Just mentioned it inline above, but wanted to emphasise: any data that is written to hard drive should be considered very temporary, since a restart of the web pods will destroy it.  This type of cache is not currently shared between web instances, but the plan is to create a persistent volume that mounts onto each pod, where the cache is stored.

We should, to start, move all currently cached artifacts (thumnails, starchart queries) to a single directory structure, such as `__cache__/thumbnails` and `__cache__/whatever`.

PR #316, opened 2020-01-21
Update API docs and doc generation script

acrellin:


skyportal/app_server.py
```diff
@@ -47,7 +47,7 @@ def make_app(cfg, baselayer_handlers, baselayer_settings):
     handlers = baselayer_handlers + [
         # API endpoints
         (r'/api/comment(/[0-9]+)?', CommentHandler),
-        (r'/api/comment(/[0-9]+)/(download_attachment)', CommentHandler),
+        (r'/api/comment(/[0-9]+)/(attachment)', CommentHandler),
```
stefanv: Should this be CommentAttachmentHandler, and `(attachment)` change to `attachment`?

skyportal/handlers/api/comment.py
```diff
@@ -159,3 +169,35 @@ def delete(self, comment_id):
         self.push_all(action='skyportal/REFRESH_SOURCE',
                       payload={'source_id': source_id})
         return self.success()
+
+
+class CommentAttachmentHandler(BaseHandler):
+    @auth_or_token
+    def get(self, comment_id, attachment):
```
stefanv: Probably don't want attachment included here

skyportal/schema.py
```diff
@@ -92,6 +99,22 @@ class Meta(object):
             setattr(sys.modules[__name__], class_.__name__,
                     schema_class())
 
+            # Create a version of the source schema w/out ID field for PUT docs
+            if class_.__name__ == 'Source':
+                class Meta2(object):
```
stefanv: Meta2 -> MetaNoID or somesuch

skyportal/handlers/api/comment.py
```diff
@@ -17,19 +17,23 @@ def get(self, comment_id, action=None):
...
+              application/json:
+                schema: Error
         """
         comment = Comment.query.get(comment_id)
         if comment is None:
             return self.error('Invalid comment ID.')
         # Ensure user/token has access to parent source
         s = Source.get_if_owned_by(comment.source.id, self.current_user)
-        if action == 'download_attachment':
+        if action == 'attachment':
```
stefanv: Is this still relevant?

skyportal/app_server.py
acrellin: Ah, yes! Good catch -- this was an unsuccessful merge conflict resolution (from your PR that re-ordered these). Looks like this is why some of the tests are failing, too.

skyportal/openapi.py
```diff
@@ -79,14 +79,17 @@ def spec_from_handlers(handlers):
...
-            path_template = re.sub('(?=[^/]{1}){}', '/{}', path_template)
+            path_template = re.sub('\(.*?\)\??', '/{}', path_template)
+            path_template = re.sub('(/)+', '/', path_template)
             path_parameters = path_template.count('{}')
 
             spec = yaml_utils.load_yaml_from_docstring(method.__doc__)
             parameters = list(inspect.signature(method).parameters.keys())[1:]
             parameters = parameters + (path_parameters - len(parameters)) * ['',]
 
+            if parameters[-1:] == [''] and path_template.endswith('/{}'):
```
stefanv: Question: doesn't this just remove the trailing slash, same as below?

skyportal/schema.py
```diff
@@ -76,21 +76,43 @@ def setup_schema():
                     "Model class ends with 'Schema'"
                 )
 
-            class Meta(object):
+            class MetaExcludeID(object):
```
stefanv: How about using a function to produce these?  Something like:

```
def schema_class_factory(excludes=None):
    ...
    return type(...)
```

skyportal/openapi.py
acrellin: That only handles the case for the "multiple" paths - when we omit these two lines, we still get paths like this in the docs: `/api/comment/`

stefanv: So, I guess the idea would be to do the trailing slash removal later (i.e., remove the .sub above, and place it after parameter substitution).

acrellin: We can do that, yes, but we'll have to do it twice -- there are two cases below. If we do it this way, we only do it once, and don't have any repeated code. Would you prefer I do it twice below, for each of the cases?

acrellin: (I removed the later sub, as it's actually only needed here)

skyportal/schema.py
acrellin: Done! Check it out...

skyportal/handlers/api/comment.py
```diff
@@ -159,3 +162,35 @@ def delete(self, comment_id):
...
+        responses:
+          200:
+            content:
+              application:
+                schema:
+                  type: string
+                  format: base64
+                  description: base64-encoded contents of attachment
+        """
+        comment = Comment.query.get(comment_id)
```
stefanv: Shouldn't there be a check here to make sure the user should be able to view the comment?

skyportal/handlers/api/comment.py
```diff
@@ -159,3 +162,35 @@ def delete(self, comment_id):
...
+        if comment is None:
+            return self.error('Invalid comment ID.')
+        # Ensure user/token has access to parent source
+        s = Source.get_if_owned_by(comment.source.id, self.current_user)
```
stefanv: This source is fetched, but not used?

skyportal/schema.py
```diff
@@ -76,21 +76,39 @@ def setup_schema():
...
+            if hasattr(class_, 'id') and class_.id is not None:
+                MetaExcludeID.exclude.append('id')
 
-            schema_class_name = '%s' % class_.__name__
+            schema_with_id_class_name = f'{class_.__name__}'
+            schema_no_id_class_name = f'{class_.__name__}NoID'
 
-            schema_class = type(
-                schema_class_name,
+            schema_with_id_class = type(
```
stefanv: This isn't exactly what I had in mind; if we change the utility function slightly, we would not need the class name specification on line 93, 94 or these type instantiations.  We can also have the factory insert the type into the namespace.

skyportal/handlers/api/comment.py
acrellin: Yes, did you see the comment above? It's for permissions checking

skyportal/handlers/api/comment.py
acrellin: See comment below - we do a permissions check

skyportal/schema.py
acrellin: What do you propose?

skyportal/handlers/api/comment.py
stefanv: Hrm, does "get_if_owned_by" throw an error if it fails, or how do you report that back to the user?

skyportal/schema.py
stefanv: A factory method that takes the name of the class and the schema and, perhaps, a list of excludes, and then does all the logic.

The main goal is to avoid the duplication that is currently there.

skyportal/handlers/api/comment.py
acrellin: Yeah, it returns with an informative error message

skyportal/schema.py
acrellin: I see what you mean. Sure, will do!

acrellin: @stefanv done!

skyportal/schema.py
```diff
@@ -76,21 +77,28 @@ def setup_schema():
...
+                for exclude_attr in exclude:
+                    if hasattr(class_, exclude_attr) and getattr(class_, exclude_attr) is not None:
+                        schema_class_meta.exclude.append(exclude_attr)
+
+                schema_class = type(schema_class_name, (_ModelSchema,),
+                                    {'Meta': schema_class_meta}
+                )
 
-            schema_class_name = '%s' % class_.__name__
+                if 'id' not in exclude:
```
stefanv: Overall, this now looks great!

This if-statement is a bit magical.  Perhaps a keyword like `install_schema` or `add_to_model` would make things more explicit.

Then, let's add a docstring to explain this to our future selves, and I think we're done!

acrellin: Good call on just adding that as a parameter. That's in now, as is the docstring.

stefanv:
Nice work, Ari! Thank you.

PR #320, opened 2020-01-31
Add section to docs outlining how to add new features

acrellin:
[ch471]


doc/contributing.md
```diff
@@ -19,14 +12,225 @@ To make a code contribution to the project:
...
 
-For a more detailed explanation, see
-the
-[scikit-image contributors' guide](http://scikit-image.org/docs/stable/contribute.html).
-We follow a very similar process.
+For a more detailed explanation, see the [scikit-image contributors' guide](http://scikit-image.org/docs/stable/contribute.html). We follow a very similar process.
+
+### Adding new features
+
+This section will help you get started adding new features to SkyPortal. We'll walk through adding a new feature: a button that triggers an API call to the back-end, which generates and returns a random string that is stored in the front-end redux state, triggering an automatic re-render of the React component that displays that value in our example component.
```
stefanv: This initial description is a bit too detailed.  Can we start with a very high level description, and explain the details later?

doc/contributing.md
```diff
@@ -19,14 +12,225 @@ To make a code contribution to the project:
...
+
+#### Front-end: React & Redux
+
+The SkyPortal front-end uses React in conjunction with Redux for state management. See the following links for introductions to these libraries:
+
+https://reactjs.org/tutorial/tutorial.html
+https://reactjs.org/docs/hooks-intro.html
+https://redux.js.org/basics/basic-tutorial/
+
+We'll start with defining our new `RandomString` component, which accesses (reads, not modifies) the Redux state via the `useSelector` hook and dispatches action creators with the `useDispatch` hook (see tutorials links) in a new file in the components directory (`static/js/components/RandomString.js`):
```
stefanv: I think for someone who has never seen React / Redux before, this is all fairly overwhelming.  We'll have to start with an overview of how information flows through the system.  We can use the sketches I made for the presentation for that, if helpful, or grab some from online.

doc/contributing.md
```diff
@@ -19,14 +12,225 @@ To make a code contribution to the project:
...
+    }
+    default:
+      return state;
+  }
+}
+```
+
+And we'll render our new `RandomString` component at the bottom of our profile page. The `Profile` component is defined in `static/js/components/Profile.jsx`:
+
+``` javascript
```
stefanv: Maybe just show the relevant sections of that file, using `...` to hide other lines that the user shouldn't care about.

doc/contributing.md
```diff
@@ -19,14 +12,225 @@ To make a code contribution to the project:
...
+      <RandomString />
+    </div>
+  );
+};
+
+export default Profile;
+```
+Note that we've only added a single line (`<RandomString />`), plus a few line breaks, to render our new component within the `Profile` component.
+
+Then we add two lines to our main reducer file to import our new reducer from `randomString.js` and add it to our root reducer (`static/js/reducers.js`):
```
stefanv: Once we merge my reducer branch (I still need to make a PR) this step can go away!

doc/contributing.md
```diff
@@ -19,14 +12,225 @@ To make a code contribution to the project:
...
+``` python
+import uuid
+from baselayer.app.access import permissions, auth_or_token
+from ..base import BaseHandler
+
+
+class RandomStringHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        """
```
stefanv: Maybe this docstring is too much to show initially.  I'd add a separate section for that later on.

skyportal/handlers/api/random_string.py
```diff
@@ -0,0 +1,25 @@
+import uuid
```
stefanv: This is kind of a weird thing to have hanging around SkyPortal.  Any thoughts on how we can test it without adding it to the "official" application?

static/js/components/Profile.jsx
```diff
@@ -26,6 +27,11 @@ User roles:
       />
       <br />
       <TokenList tokens={profile.tokens} />
+      <br />
```
stefanv: Definitely don't want *this* checked in by default :)

acrellin:
@stefanv I could remove the changes to the actual application code, but keep it all in the docs, so that a new contributor could copy/paste the code into the relevant files and thereby get a sense of the code organization -- thoughts?

stefanv:
Perhaps you want to write this document in RST, so that you can include the source directly, without having to paste it into the docs.  But I'm a bit stumped on how to test it.  Maybe add a new page to SkyPortal, called "developer", that gives links to docs, show this example component, etc.?

acrellin:
@stefanv I don't think it actually needs to be included in the application -- we can just detail what you'd need to do to implement it, and then as an exercise to familiarize themselves with the code base, new contributors can create/edit the necessary files and paste in the code from the doc. That way we don't have to add it to the app. I'll pull this into a new Contributing New Features document.

acrellin:
@stefanv I've made some changes, and would love any more feedback you have.

doc/adding_features.md
```diff
@@ -0,0 +1,161 @@
...
+import re
+from selenium.webdriver.support.ui import Select
+
+def test_random_string_generator_button(driver, user):
+    driver.get(f'/become_user/{user.id}')
+    driver.get('/profile')
+
+    orig_display_xpath = '//div[contains(.,"Current value is: null")]'
+    display_div = driver.wait_for_xpath(orig_display_xpath)
+    assert not re.search(r'[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}', display_div.text)
```
stefanv: I feel like this function is overly complicated; can we show them a simpler test?

doc/adding_features.md
```diff
@@ -0,0 +1,161 @@
...
+
+`skyportal/tests/frontend/test_random_string.py`:
+
+``` python
+import pytest
+import re
+from selenium.webdriver.support.ui import Select
+
+def test_random_string_generator_button(driver, user):
+    driver.get(f'/become_user/{user.id}')
```
stefanv: You explained this to me today, but I may need another round of explanation.  Can we not make the token belong to the same user, so that it is unnecessary to become the user?  Otherwise, we could probably automatically log the correct user in anyway.

doc/adding_features.md
```diff
@@ -0,0 +1,161 @@
...
+
+We see that our tests pass (`make test`) and we're ready to file a pull request.
+
+## File creation/edit bookkeeping
+
+Let's recap the files we've created or edited to implement this new feature:
+
+- component definition file `static/js/components/RandomString.jsx` _(new)_
+- Redux-related code (action creators, reducer, etc.) `static/js/ducks/randomString.js` _(new)_
+- importing our new reducer and adding it to the root reducer in `static/js/reducers.js` _(2 lines of code)_
```
stefanv: Gone, yay!

acrellin:
This is very much WIP, thanks - I'll let you know when it's ready for
review.

doc/adding_features.md
acrellin: Is it the regex part that's too complicate, or the logic?

doc/adding_features.md
```diff
@@ -0,0 +1,541 @@
...
+In order for our components to access the application state, the `react-redux` library provides the `useSelector` hook:
+
+``` jsx
+import { useSelector } from 'react-redux';
+
+// Accessing the application state
+const { sourceList } = useSelector((state) => state.sources);
+```
+
+In terms of code organization, we bundle all of our Redux-related code (action types, action creators, reducer) associated with a particular branch of the application state together in a file in the `static/js/ducks` directory ([read more about "ducks" modules here](https://github.com/erikras/ducks-modular-redux)). Each component definition typically goes in its own file in `static/js/components`.
```
stefanv: Maybe mention this the first time that ducks is mentioned above?

doc/adding_features.md
```diff
@@ -0,0 +1,541 @@
...
+// Action type
+export const FETCH_SOURCES = 'skyportal/FETCH_SOURCES';
+
+// Action creator that returns a thunk
+export function fetchSources() {
+  return API.GET('/api/sources', FETCH_SOURCES);
+}
+```
+
+When we call `dispatch(fetchSources())`, the thunk returned by `API.GET` is called, which dispatches an action of type `FETCH_SOURCES`, then makes an asynchronous HTTP GET request to the back-end. If the request returns with a successful response, the response data is then packaged into another action of type `FETCH_SOURCES + "_OK"` which is then dispatched to the store. The first action dispatched, of type `FETCH_SOURCES`, may or may not need to be handled in our reducers. In the case of SkyPortal, we do indeed handle this action and update the front-end state object `sources.queryInProgress` in our reducer. The final action dispatched, of type `FETCH_SOURCES + "_OK"`, which contains the HTTP request response data, is also handled in our reducer, updating other portions of the `sources` branch of the app state.
```
stefanv: Just `FETCH_SOURCES_OK`?

doc/adding_features.md
```diff
@@ -0,0 +1,541 @@
...
+          Generate new random string on the back-end.
+        </button>
+      </div>
+    </div>
+  );
+};
+
+export default RandomString;
+```
+Our new component can see a portion of the app state tree (specifically the `randomString` branch) via the `useSelector` hook, and can dispatch action creators via the `useDispatch` hook. The value of `state.randomString` (which is intialized to the string "null" in its associated reducer, which we'll see shortly) is displayed, and a button is shown. We've supplied a handler to the button's `onClick` event, which when called, dispatches an action creator called `fetchRandomString`. Let's look at that action creator defined belo.
```
stefanv: below

stefanv:
A lot of writing; this should be very helpful to new devs starting up.  Thank you, Ari!

acrellin:
Thanks for the suggestions / edits, @stefanv! Shall I just push those little changes up or file another PR?

stefanv:
Go ahead and push them up, thanks Ari.

PR #326, opened 2020-02-11
Return new source info in source update response

acrellin:


stefanv:
I just read [this SO answer](https://stackoverflow.com/questions/797834/should-a-restful-put-operation-return-something) and feel quite conflicted.  What do you think after reading that?

stefanv:
Just learned that we are using the wrong HTTP status codes for successful POST as well :)

stefanv:
For our use case, perhaps we should rather *not* include the modified body.  In Danny's application, we can then untoggle the button if the request comes back a failure.  That seems even simpler to implement for now.  Sorry for going back and forth on this.

acrellin:
@stefanv I just looked over that standard, and we definitely do not follow it throughout the application, both in terms of status codes and response content -- e.g. we return the new record's ID in the response body for POST requests. But if you want to move towards complying with that spec, sure, I'll scrap this and we can take care Danny's update score issue with the other approach we discussed.

acrellin:
@stefanv actually, if we want to conform with that standard, we'll have to do away with returning the SkyPortal version in the response body as we currently do for all requests.

acrellin:
@stefanv actually, after a more careful reading, this standard just specifies the status codes that should be used, but doesn't say that the body should be empty.

acrellin:
@stefanv ah, never mind - this doc does indeed say a PUT response should have no body: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

acrellin:
(a successful POST response should have a body, however)

acrellin:
@stefanv looks like we should actually be using PATCH instead of PUT for our update handlers, in which case the response SHOULD have a body :)

acrellin:
Will update the API to use PATCH where appropriate, and will return the updated record in those responses. See upcoming separate PR.

PR #328, opened 2020-02-11
Add Jump to Page functionality on source list

acrellin:
Adds a text input field and button to bottom of `SearchBox` component to allow users to jump to any page in current source filter results to avoid manually paging one at a time.

@dannygoldstein here you go


dannygoldstein:
Thanks a ton @acrellin. Any insight on the  issue where the same source is shown on multiple pages? Were you able to reproduce?

stefanv:
I think this is exactly the kind of thing JS testing would have been good for; I'll investigate. I think it'd make life a lot easier. I guess for now we need to test the long way :/

The changes look good, thanks Ari!

PR #329, opened 2020-02-12
Remove extra eslint rules; fix linting errors

stefanv:
This will require restyling the "delete" button on comments, but otherwise the app should be unaffected.


static/js/components/Group.jsx
```diff
@@ -33,18 +33,20 @@ const Group = ({ route }) => {
...
       <ul>
         {
-          group.users.map((user, idx) => (
+          group.users.map((user) => (
             <li key={user.id}>
               <Link to={`/user/${user.id}`}>
                 {user.username}
-              </Link>&nbsp;&nbsp;
+              </Link>
+&nbsp;&nbsp;
```
acrellin: Indentation

acrellin:
great, thanks @stefanv

stefanv:
@acrellin There's some actual breakage here; I fear it is because I didn't guess the types correctly anywhere.  Could you run this branch and see (in the web developer console) if you see what I got wrong there?

PR #330, opened 2020-02-12
Wrap Plot component with React.memo

acrellin:
This fixes a bug where the bokeh plots re-render whenever any part of `state.source` is updated, included when adding comments. This significantly speeds up updates to the source page.


stefanv:
Best one liner ever :)

PR #345, opened 2020-02-13
Replace global message handler with one per duck

stefanv:
Localizing the message handlers should make it much easier to develop individual components.  Of course, it would still be possible to handle messages outside of components, should that be desired.


acrellin:
This is great, thanks @stefanv! Want to pin this branch to latest baselayer master and I'll merge?

static/js/ducks/group.js
```diff
@@ -10,6 +12,21 @@ export function fetchGroup(id) {
   return API.GET(`/api/groups/${id}`, FETCH_GROUP);
 }
 
+
+// Websocket message handler
+messageHandler.add((action, payload, dispatch, getState) => {
```
acrellin: Can I suggest `action` -> `action_type` throughout?

acrellin:
This is excellent, @stefanv! Further reduces the number of files that need to be touched to add new features.

PR #351, opened 2020-02-14
Set initial state for profile and newsFeed reducers.

stefanv:
This avoids checks inside components to ensure that state is defined and
of correct form.

It would be good to review other components that do this kind of checking too.

---

The major changes here are now:

- The backend never returns undefined for profile preferences, but rather an empty dictionary.
- The newsfeed component now checks for its preferences, and if not found applies a set of default preferences.
  - TODO: The default preferences should probably be merged with whatever is provided by the user, so we can ensure that the widget always has all of its configuration parameters.
- Random cleanups, but specifically API.js, which I rewrote to give more logical error responses: a) connection error b) JSON decoding error c) actual 400 and similar errors that are reported by the backend.

### DIFF
--- a/static/js/components/GroupList.jsx
+++ b/static/js/components/GroupList.jsx
@@ -8,7 +8,7 @@ import SourceList from './SourceList';
 
 const GroupList = ({ title, listSources, groups }) => {
   if (!groups || groups.length === 0) {
-    groups = useSelector((state) => state.groups.latest);
+    groups = useSelector((state) => state.groups.user);
   }
   return (
     <div style={{ border: "1px solid #DDD", padding: "10px" }}>
--- a/static/js/ducks/profile.js
+++ b/static/js/ducks/profile.js
@@ -52,7 +52,9 @@ const initialState = {
   roles: [],
   acls: [],
   tokens: [],
-  preferences: {}
+  preferences: {
+    newsFeed: {}
+  }
 };
 
 const reducer = (state=initialState, action) => {
--- a/static/js/ducks/profile.js
+++ b/static/js/ducks/profile.js
@@ -52,9 +52,7 @@ const initialState = {
   roles: [],
   acls: [],
   tokens: [],
-  preferences: {
-    newsFeed: {}
-  }
+  preferences: {}
 };
 
 const reducer = (state=initialState, action) => {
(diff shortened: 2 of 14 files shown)


acrellin:
@stefanv the issue _wasn't_ that the default value of `profile.preferences` was an empty dict and didn't have a `newsFeed` attribute, which is why there was still an error when I removed all of the attribute checks _except_ for `Object.prototype.hasOwnProperty.call(profile.preferences, "newsFeed")` -- even with that being checked, the error was occurring. This PR doesn't address the issue -- it wasn't that the default redux state wasn't properly initialized. I don't know that we even want to bake all of the possible component config state into the redux state...

stefanv:
What was the issue that you witnessed then?  I can reproduce these failures locally, and am trying to see where things go wrong.  I see that the reducers are initialized correctly.

acrellin:
Can you reproduce them locally even with the new default reducer values?

stefanv:
Not via the browser, but via the tests.  Digging a bit deeper into baselayer now :)

static/js/components/GroupList.jsx
```diff
@@ -8,7 +8,7 @@ import SourceList from './SourceList';
 
 const GroupList = ({ title, listSources, groups }) => {
   if (!groups || groups.length === 0) {
-    groups = useSelector((state) => state.groups.latest);
+    groups = useSelector((state) => state.groups.user);
```
acrellin: Nice, I've been thinking we should rename this for a long time!

static/js/ducks/profile.js
```diff
@@ -52,7 +52,9 @@ const initialState = {
   roles: [],
   acls: [],
   tokens: [],
-  preferences: {}
+  preferences: {
+    newsFeed: {}
```
acrellin: This seems unnecessary, and I'm not sure we want to have to keep updating this default value every time we add a new configurable component/widget. The `NewsFeed` component already checks whether `preferences` has  a `newsFeed` attr, and supplies a default value if not. I think it'd simplify the process of adding new features to just have `preferences` default to `{}` here.

stefanv: Removed, let's see how the tests go!

acrellin:
Looks great! Thanks @stefanv! In it goes :rocket:

PR #353, opened 2020-02-18
Check that source list page number value is valid and return informative error message if not

### LINKED ISSUE #350: Page jumping can go out of range
Page jumping can currently lead to "showing 401-2 of 2 matching sources".

acrellin:
Closes #350


acrellin:
@stefanv the check is duplicated here, but after we merge this I'll submit another PR that combines the two different handlers and puts all filter parameters in a GET query string, so the duplication will be removed there (multiple handlers that need to be combined is a separate issue)

acrellin:
Moved to https://github.com/skyportal/skyportal/pull/355

PR #355, opened 2020-02-19
Combine source handlers; add page number checks

### LINKED ISSUE #311: Combine source and filter sources handlers

### LINKED ISSUE #350: Page jumping can go out of range

acrellin:
Removes FilterSourcesHandler and merges with SourcesHandler.
Also checks that source list page number value is valid and returns informative error message if not (displays a notification saying either "Page number out of range" or "Invalid page number value" depending on the case).
Closes #350 
Closes #311


static/js/ducks/sources.js
```diff
@@ -4,14 +4,17 @@ import store from '../store';
...
-  return API.GET(`/api/sources?page=${page}`, FETCH_SOURCES);
-}
-
-export function submitSourceFilterParams(formData) {
-  return API.POST(`/api/sources/filter`, FETCH_SOURCES, formData);
+export function fetchSources(filterParams={}) {
+  if (!Object.keys(filterParams).includes("pageNumber")) {
+    filterParams.pageNumber = 1;
+  }
+  const queryString = Object.keys(filterParams).map(
```
stefanv: I'm not sure this is safe; shouldn't there be some form of escaping?  Perhaps fetch already supports providing the query arguments in another way.

acrellin: Fetch doesn't provide that (the docs say to just put query parameters in the URL). Do you mean just URI encoding certain characters? Sure, I'll add this, which does the mapping and URI encodes any special characters: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString

acrellin: @stefanv I've updated this, take a look now. All relevant characters are now URI-encoded and all URL query strings are safe and to spec. Should be good to go.

static/js/ducks/sources.js
```diff
@@ -4,14 +4,16 @@ import store from '../store';
...
-  return API.GET(`/api/sources?page=${page}`, FETCH_SOURCES);
-}
-
-export function submitSourceFilterParams(formData) {
-  return API.POST(`/api/sources/filter`, FETCH_SOURCES, formData);
+export function fetchSources(filterParams={}) {
+  if (!Object.keys(filterParams).includes("pageNumber")) {
+    filterParams.pageNumber = 1;
+  }
+  const params = new URLSearchParams(filterParams);
```
stefanv: This looks good to me, similar to what is suggested [here](https://github.com/github/fetch/issues/256#issuecomment-170228674).

I'm going to merge, but how about we add this as a general feature of `API.GET`, so that we can utilize it in other places too?

I.e., `GET(endpoint, actionType, queryParams)`.

skyportal/handlers/api/source.py
```diff
@@ -60,18 +69,56 @@ def get(self, source_id=None):
...
+                page = int(page_number)
+            except ValueError:
+                return self.error("Invalid page number value.")
             q = Source.query.filter(Source.id.in_(DBSession.query(
                 GroupSource.source_id).filter(GroupSource.group_id.in_(
                     [g.id for g in self.current_user.groups]))))
-            info['totalMatches'] = q.count()
+            if sourceID:
+                q = q.filter(Source.id.contains(sourceID.strip()))
+            if ra and dec and radius:
```
stefanv: Silent failure if not all three specified?

static/js/ducks/sources.js
acrellin: Good call! I'll do this tomorrow.

skyportal/handlers/api/source.py
acrellin: Do you think if they provide any of these but not all three we should return an error message? Now it just omits those params from the query.

stefanv: Right, because it shows intent to do spatial filtering, just that it is done incorrectly.

PR #356, opened 2020-02-19
Return informative error message if any, but not all, of (ra, dec, radius) are provided in source filtering

acrellin:


skyportal/handlers/api/source.py
```diff
@@ -78,7 +78,10 @@ def get(self, source_id=None):
                     [g.id for g in self.current_user.groups]))))
             if sourceID:
                 q = q.filter(Source.id.contains(sourceID.strip()))
-            if ra and dec and radius:
+            if any([ra, dec, radius]):
+                if not all([ra, dec, radius]):
+                    return self.error("If any of 'ra', 'dec' or 'radius' are "
```
stefanv:
```suggestion
                    return self.error("If any of 'ra', 'dec' or 'radius' is "
```

stefanv:
Good to go, apart from typo.

skyportal/handlers/api/source.py
acrellin: It's any number of, not just any single one, so I think "are" is right here ;)

stefanv: Hah, I think the grammar breaks down either way... so guess it doesn't matter :joy:

PR #359, opened 2020-02-20
Most viewed sources widget

acrellin:
This PR adds a most viewed sources widget to the home page. The number of sources displayed, and the time range to include are customizable by the user and are persisted in the DB. The `/api/internal/source_views` endpoint and associated handlers are added (handlers for POST, for registering new source views, and GET, for fetching the list of top sources, are implemented). Front-end source views (fetching of non-cached source) and token `/api/sources/<id>` GET requests are added to the new `SourceViews` table, along with the username/token ID (and a bool indicating whether it was a token or a front-end user) and timestamp.

### DIFF
--- a/skyportal/handlers/api/internal/source_views.py
+++ b/skyportal/handlers/api/internal/source_views.py
@@ -8,20 +8,22 @@ from ....models import (
 )
 
 
+default_prefs = {
+    'maxNumSources': 10,
+    'sinceDaysAgo': 7
+}
+
+
 class SourceViewsHandler(BaseHandler):
     @auth_or_token
     def get(self):
-        prefs = (self.current_user.preferences if
-                 hasattr(self.current_user, 'preferences') and
-                 self.current_user.preferences is not None else {})
-        if 'topSources' in prefs and 'maxNumSources' in prefs['topSources']:
-            max_num_sources = int(prefs['topSources']['maxNumSources'])
-        else:
-            max_num_sources = 10
-        if 'topSources' in prefs and 'sinceDaysAgo' in prefs['topSources']:
-            since_days_ago = int(prefs['topSources']['sinceDaysAgo'])
-        else:
-            since_days_ago = 30
+        user_prefs = getattr(self.current_user, 'preferences', None) or {}
+        top_sources_prefs = user_prefs.get('topSources', {})
+        top_sources_prefs = {**default_prefs, **top_sources_prefs}
+
+        max_num_sources = int(top_sources_prefs['maxNumSources'])
+        since_days_ago = int(top_sources_prefs['sinceDaysAgo'])
+
         cutoff_day = datetime.datetime.now() - datetime.timedelta(days=since_days_ago)
         q = (DBSession.query(func.count(SourceView.source_id).label('views'),
                              SourceView.source_id).group_by(SourceView.source_id)
--- a/skyportal/handlers/api/internal/source_views.py
+++ b/skyportal/handlers/api/internal/source_views.py
@@ -1,0 +1,50 @@
+import datetime
+from sqlalchemy import func, desc
+from baselayer.app.access import permissions, auth_or_token
+from ...base import BaseHandler
+from ....models import (
+    DBSession, Source, SourceView, GroupSource
+)
+
+
+class SourceViewsHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        prefs = (self.current_user.preferences if
+                 hasattr(self.current_user, 'preferences') and
+                 self.current_user.preferences else {})
+        if 'topSources' in prefs and 'maxNumSources' in prefs['topSources']:
+            max_num_sources = int(prefs['topSources']['maxNumSources'])
+        else:
+            max_num_sources = 10
+        if 'topSources' in prefs and 'sinceDaysAgo' in prefs['topSources']:
+            since_days_ago = int(prefs['topSources']['sinceDaysAgo'])
+        else:
+            since_days_ago = 30
+        cutoff_day = datetime.datetime.now() - datetime.timedelta(days=since_days_ago)
+        q = (DBSession.query(func.count(SourceView.source_id).label('views'),
+                             SourceView.source_id).group_by(SourceView.source_id)
+             .filter(SourceView.source_id.in_(DBSession.query(
+                 GroupSource.source_id).filter(GroupSource.group_id.in_(
+                     [g.id for g in self.current_user.groups]))))
+             .filter(SourceView.created_at >= cutoff_day)
+             .order_by(desc('views')).limit(max_num_sources))
+        return self.success(data={'sourceViews': q.all()})
+
+    @auth_or_token
+    def post(self, source_id):
+        # Ensure token/user has access to source
+        s = Source.get_if_owned_by(source_id, self.current_user)
+        # This endpoint will only be hit by front-end, so this will never be a token
+        register_source_view(source_id=source_id,
+                             username_or_token_id=self.current_user.username,
+                             is_token=False)
+        return self.success()
+
+
+def register_source_view(source_id, username_or_token_id, is_token):
+    sv = SourceView(source_id=source_id,
+                    username_or_token_id=username_or_token_id,
+                    is_token=is_token)
+    DBSession.add(sv)
+    DBSession.commit()
--- a/skyportal/handlers/api/internal/source_views.py
+++ b/skyportal/handlers/api/internal/source_views.py
@@ -1,6 +1,7 @@
 import datetime
 from sqlalchemy import func, desc
-from baselayer.app.access import permissions, auth_or_token
+import tornado.web
+from baselayer.app.access import auth_or_token
 from ...base import BaseHandler
 from ....models import (
     DBSession, Source, SourceView, GroupSource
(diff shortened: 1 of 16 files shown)


acrellin:
[ch465]

stefanv:
I think it is too expensive to calculate a sum on a views table *every time* you need to know how many times an object has been viewed. But, this is probably still the correct way to store data, since updating a counter is a blocking operation.

So, I would suggest:

- Making `SourceView` more generic; perhaps `Activity`.  Each activity has a type, ID, and description associated with it.  We can use such a table to store "x comment on y" type events too.

It you think it is too much to mix all events, then perhaps `View` or something similar, just to log object views of various kinds.

- We'll need a cron-job that aggregates view counts every, say, ten minutes.  This will be a service managed by supervisor, that will trigger every so-often to perform a task.  In this case, the task will be to sum the `View` table per object, and to store the result in a `ViewCount` table.

stefanv:
(You can even make the view counts accurate, by having the cron job remove view logs from the table; then, the view count = the result from the viewcount table + sum(views logged since) --- which should be very few items.)

acrellin:
@stefanv just on my laptop, the query takes ~40ms when there are 100,000 rows (all matching the query) in the `SourceView` table. If we run a regular cron job to delete view records more than a month old (then put a cap on the configurable look-back period at 30 days), even if the marshal is getting an average of 3,000 source views per day, that is still only a less than 40ms query. With the approach you're proposing, we lose the ability to configure the look-back period, because we're losing timestamps (currently the user can select the number of days to look back). This query is only made (asynchronously) when a user loads the homepage for the first time during a browser session. Thoughts?

stefanv:
@acrellin That's convincing, thanks for doing that experiment.

stefanv:
@acrellin What do you think about making the table a bit more generic than sources, i.e. having an activity table?

acrellin:
@stefanv hmm, maybe we can ask the group today whether we'd want to explicitly track other activity besides source views

skyportal/handlers/api/internal/source_views.py
```diff
@@ -0,0 +1,50 @@
+import datetime
+from sqlalchemy import func, desc
+from baselayer.app.access import permissions, auth_or_token
```
stefanv: Looks like some unused imports; I habitually run my files through flake8 to flag this sort of annoying little thing.

skyportal/handlers/api/internal/source_views.py
```diff
@@ -0,0 +1,50 @@
...
+    DBSession, Source, SourceView, GroupSource
+)
+
+
+class SourceViewsHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        prefs = (self.current_user.preferences if
+                 hasattr(self.current_user, 'preferences') and
+                 self.current_user.preferences else {})
```
stefanv: Is this a check for None?  Should then be: "and self.current_user.preferences is not None.

But cleaner may be:

`getattr(self.current_user, 'preferences') or {}`

skyportal/handlers/api/internal/source_views.py
```diff
@@ -0,0 +1,50 @@
...
+        if 'topSources' in prefs and 'maxNumSources' in prefs['topSources']:
```
stefanv: I think we discussed the possibility of a defaults dictionary that gets updated. Thoughts?

skyportal/handlers/api/internal/source_views.py
```diff
@@ -0,0 +1,50 @@
...
+            max_num_sources = int(prefs['topSources']['maxNumSources'])
+        else:
+            max_num_sources = 10
+        if 'topSources' in prefs and 'sinceDaysAgo' in prefs['topSources']:
+            since_days_ago = int(prefs['topSources']['sinceDaysAgo'])
+        else:
+            since_days_ago = 30
+        cutoff_day = datetime.datetime.now() - datetime.timedelta(days=since_days_ago)
+        q = (DBSession.query(func.count(SourceView.source_id).label('views'),
```
stefanv: Just for my own education, what is the `label("views")` part?

skyportal/handlers/api/internal/source_views.py
```diff
@@ -0,0 +1,50 @@
...
+                             SourceView.source_id).group_by(SourceView.source_id)
+             .filter(SourceView.source_id.in_(DBSession.query(
+                 GroupSource.source_id).filter(GroupSource.group_id.in_(
+                     [g.id for g in self.current_user.groups]))))
+             .filter(SourceView.created_at >= cutoff_day)
+             .order_by(desc('views')).limit(max_num_sources))
+        return self.success(data={'sourceViews': q.all()})
+
+    @auth_or_token
```
stefanv: If this is never going to be a token, then the web app authentication decorator should be used instead.

skyportal/handlers/api/source.py
```diff
@@ -60,7 +62,13 @@ def get(self, source_id=None):
         simbad_class = self.get_query_argument('simbadClass', None)
         has_tns_name = self.get_query_argument('hasTNSname', None)
         total_matches = self.get_query_argument('totalMatches', None)
+        is_token_request = isinstance(self.current_user, Token)
         if source_id:
+            if is_token_request:
```
stefanv: So, we register both token and non-token views. Looks like we currently tally both in the view count, correct?

skyportal/handlers/api/internal/source_views.py
```diff
@@ -0,0 +1,50 @@
...
+    def post(self, source_id):
+        # Ensure token/user has access to source
+        s = Source.get_if_owned_by(source_id, self.current_user)
+        # This endpoint will only be hit by front-end, so this will never be a token
+        register_source_view(source_id=source_id,
```
stefanv: Why does post get a source view?

skyportal/handlers/api/internal/source_views.py
acrellin: Thanks

skyportal/handlers/api/internal/source_views.py
acrellin: Agreed, thanks

skyportal/handlers/api/internal/source_views.py
acrellin: That aggregate count value gets the label "views" here just so that we can refer to that below in `order_by`

acrellin: Just gives that value an alias so you can refer to it later

skyportal/handlers/api/source.py
acrellin: Yes, we tally both token and browser views. The token views are tallied here in the source post handler, but for the front-end, we explicitly hit `/api/internal/source_views` with a POST whenever a non-cached source page is loaded

skyportal/handlers/api/internal/source_views.py
acrellin: This is the post handler of the `SourceViewHandler`, which is what we hit to register a front-end view (we don't just tally ever front-end GET request, because that happens too often, e.g. when comments are added, the source is re-fetched. So we handle the logic of when to log a source view on the front-end, and from there we hit `/api/internal/source_views` with a POST, which is what this handles)

skyportal/handlers/api/internal/source_views.py
acrellin: `label("views")` -> `SELECT count(sourceviews.source_id) AS views`, then below there's `ORDER BY views DESC`

skyportal/handlers/api/internal/source_views.py
acrellin: Sure

skyportal/handlers/api/internal/source_views.py
acrellin: Can you explain what you're thinking here? Can we set up our JSONB column to be a default dict? I'm a bit unclear about what you mean

stefanv: I mean that, somewhere in the app, on frontend or backend, we should define the default values for the preferences.  And instead of having logic for each branch, just start off with the default values dictionary, and update it with whatever is set in the preferences.

skyportal/handlers/api/internal/source_views.py
stefanv: Nice, that's clever!

skyportal/handlers/api/internal/source_views.py
stefanv: Aha, thanks!

static/js/components/HomePage.jsx
```diff
@@ -16,6 +17,9 @@ const HomePage = () => {
       <div style={{ float: "left", paddingRight: "40px", paddingBottom: "40px" }}>
         <NewsFeed />
       </div>
+      <div style={{ float: "left", paddingRight: "40px", paddingBottom: "40px" }}>
```
stefanv: Styling should be defined in CSS whenever possible.

skyportal/handlers/api/internal/source_views.py
acrellin: How about adding a default value in the model definition?

static/js/components/HomePage.jsx
acrellin: Done

PR #375, opened 2020-03-04
Material ui search box & source table

jialin-wu-02:
1. Update the search filter and the source table using material-ui. 
2. Change the position of widgets to be at the top of the home page.


jialin-wu-02:
Forget to install lint... Will try to pass all the checks.

config.yaml
```diff
@@ -0,0 +1,33 @@
+app:
```
acrellin: This file shouldn't be committed to the repo

acrellin:
@jialin-wu-02 This fixes the test errors:  https://github.com/jialin-wu-02/skyportal/pull/95
cc @stefanv

acrellin:
@jialin-wu-02 green now :heavy_check_mark:

stefanv:
Try merge this with master branch, and enable dark mode from the profile page, to highlight some problems.

stefanv:
Note that, in Material UI, we should have to provide almost zero custom styles.

stefanv:
The component that still needs a bit of attention is `SearchBox.jsx`.

static/js/components/CustomInput.jsx
```diff
@@ -0,0 +1,51 @@
+import { TextField } from "@material-ui/core";
```
stefanv, line 1: We have standardized on the following input order (I'm surprised eslint doesn't warn about this!):

- React imports
- Material + all other library imports
- Own imports, such as from baselayer, and styles.

static/js/components/CustomInput.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+import React from "react";
+import PropTypes from "prop-types";
+import styles from "./Input.css";
+
+const CustomInput = ({
```
stefanv, line 7: This whole component is probably unnecessary, given that we are going to restyle anyway.

static/js/components/SearchBox.jsx
```diff
@@ -1,9 +1,20 @@
...
 
-import * as Actions from '../ducks/sources';
+import {
+  Paper,
+  Button,
+  ButtonGroup,
+  Checkbox,
+  FormControlLabel,
+  TextField,
+} from "@material-ui/core";
```
stefanv, line 12: We should use the explicit Material UI imports, as described in their documentation and used elsewhere in SkyPortal, to keep the JS bundle size down.

static/js/components/SearchBox.jsx
```diff
@@ -1,9 +1,20 @@
...
+import { makeStyles } from "@material-ui/core/styles";
 
+import * as Actions from "../ducks/sources";
+
+import styles from "./SearchBox.css";
```
stefanv, line 17: can we get rid of these?

jialin-wu-02: I think I can get rid of the styles, but I am not sure about the Actions, since we need them to handle fetching sources.

acrellin: I think he just means the styles

stefanv:
Good work, @jialin-wu-02 !

stefanv:
🚀! Well done.

Issue #377, opened 2020-03-18
docker-compose enhancements requests


dmitryduev:
- Can you add
```
version: "3.3"

services:
```
to the top of the `docker-compose.yaml` file? That will ensure the "skyportal_default" network gets created so that other dockerized services running on the same machine may connect to it and talk to skyportal.

- Please also add persistent volumes so that the data do not get wiped out every "docker-compose down".

stefanv:
Certainly! If you have the changes already, feel free to make a PR, otherwise we'll do it soon.

stefanv:
The postgres db is usually stored locally. Did it get changed to storing it inside of the container? If so, we should address that.

dmitryduev:
Created a [PR](https://github.com/skyportal/skyportal/pull/379) for the first point

dmitryduev:
Addressed the second point in #379 as well, so once that is merged, this issue can be closed.
[I am creating a persistent docker volume rather than mounting a local directory -- is that ok with you?]

stefanv:
> Addressed the second point in #379 as well, so once that is merged, this issue can be closed.
> [I am creating a persistent docker volume rather than mounting a local directory -- is that ok with you?]

Sorry, Dima, I didn't see this until I had already made the other PR!

PR #379, opened 2020-03-18
docker-compose #377

dmitryduev:
See [this issue](https://github.com/skyportal/skyportal/issues/377)


docker-compose.yaml
```diff
@@ -1,14 +1,17 @@
...
+      - "9000:5000"
+      - "9001:5001"
+    links:
+      - db
+
+  db:
+    image: postgres:9.6
+    environment:
+      - POSTGRES_USER=skyportal
+      - POSTGRES_PASSWORD=latropyks
```
stefanv: Should this be specified in this PR?

dmitryduev: Well, I just want a placeholder. This should probably be changed anyways, otherwise you'd have to modify `docker-compose.yaml` directly screwing things for the next time you do `git pull`.

dmitryduev:
Also see [here](https://github.com/skyportal/skyportal/blob/master/Dockerfile#L40). Shouldn't `docker.yaml` be changed to `config.yaml.defaults`?

acrellin:
@dmitryduev are there still outstanding questions here?

stefanv:
> Also see [here](https://github.com/skyportal/skyportal/blob/master/Dockerfile#L40). Shouldn't `docker.yaml` be changed to `config.yaml.defaults`?

This seems to be correct.

acrellin:
@dmitryduev the above comments are unclear - are you going to change where some of these configs live, or is this good to go as-is?

dmitryduev:
I propose to explicitly rename the relevant config files into *.defaults. That way it is ok to store there default pwds I think. So should be good to go. On a side note, I added relevant checks (if the config files are there with an option to copy/use the defaults) to the `fritz` utility.

stefanv:
Thanks for investigating this, @dmitryduev!  I'm going to close this for now; all relevant fixes have been made in https://github.com/skyportal/skyportal/pull/400

stefanv:
> I propose to explicitly rename the relevant config files into *.defaults. That way it is ok to store there default pwds I think. So should be good to go. On a side note, I added relevant checks (if the config files are there with an option to copy/use the defaults) to the `fritz` utility.

Let's start using pull requests on Fritz from now on, so that we can help one another review these types of changes.

The `.defaults` logic is that the app specifies all defaults there, and that the user then overrides whatever they want in `config.yaml`.

dmitryduev:
> Let's start using pull requests on Fritz from now on, so that we can help one another review these types of changes.
Sure. I should have noticed that my changes are not even committed yet :)

PR #383, opened 2020-03-27
Scanning Page #2: Add candidates request handler

acrellin:


acrellin:
@stefanv do you mean the pagination-related code? Yes, I can factor that out. Unfortunately the filtering statements probably can't be factored out, even though they're repetitive and similar (but not identical) in these two handlers.

acrellin:
[ch749]

skyportal/handlers/api/candidate.py
```diff
@@ -0,0 +1,280 @@
...
+                options=[
+                    joinedload(Source.candidate_comments),
+                    joinedload(Source.candidate_groups),
+                    joinedload(Source.thumbnails)
+                    .joinedload(Thumbnail.photometry)
+                    .joinedload(Photometry.instrument)
+                    .joinedload(Instrument.telescope),
+                ],
+            )
+            if c is None or not c.is_candidate:
```
profjsb, line 39: add parentheses for clarity: `(c is None) or (not c.is_candidate)`

skyportal/handlers/api/candidate.py
```diff
@@ -0,0 +1,280 @@
...
+    GroupCandidate,
+)
+
+
+CANDIDATES_PER_PAGE = 25
+
+
+class CandidateHandler(BaseHandler):
+    @auth_or_token
+    def get(self, candidate_id=None):
```
profjsb, line 24: Add an OpenAPI spec in the docstring?

skyportal/handlers/api/candidate.py
```diff
@@ -0,0 +1,280 @@
...
+            required: true
+            schema:
+              type: string
+        responses:
+          200:
+            content:
+              application/json:
+                schema: Success
+        """
+        _ = Source.get_if_owned_by(candidate_id, self.current_user,
```
profjsb, line 275: Will this raise an exception if not owned? Perhaps we need a `is_owned_by` which returns a bool.

skyportal/handlers/api/candidate.py
```diff
@@ -0,0 +1,280 @@
...
+        if saved_as_source_by_id is None:
+            saved_as_source_by_id = self.current_user.id
+        # TODO - create GroupSources with appropriate fields
+        DBSession.add(c)
+        DBSession().commit()
+
+        self.push_all(action="skyportal/FETCH_CANDIDATES")
+        return self.success(data={"id": c.id})
+
+    @permissions(["Manage sources"])
```
profjsb, line 185: The permissioning around deletion of a source vs deletion of a candidate will be very different. We basically will want almost no one to be able to delete a source (once it's official, it's got an official name and there's no reason to kill it even if it turns out to be bogus). Default should be only `superadmin` here. But many more roles should be able to delete candidates.

skyportal/handlers/api/candidate.py
acrellin: If not owned, the request will return with an error code and message of `"Inadequate permissions."`

skyportal/handlers/api/candidate.py
acrellin: I don't see this endpoint actually getting hit, as I *think* the only way candidates are completely wiped from the DB is when the vacuum cron job sweeps up old, unsaved candidates. Is this right? Or can a user delete a candidate so that it's no longer in the system at all?

skyportal/handlers/api/candidate.py
acrellin: Done.

skyportal/handlers/api/candidate.py
acrellin: Done.

acrellin:
@profjsb I've implemented your suggestions.

stefanv:
@acrellin The changes here look fine to me.  Question, though: is there any way we can avoid duplication between the handling of candidates and sources, given how similar they are?  E.g., can we have an abstract base handler that takes care of almost everything, with overrides where necessary?

acrellin:
@stefanv the handlers are different enough that they need to be separate. I can file an issue to factor out the pagination-related logic to use for both handlers, but I'd rather do that in a future PR so that this doesn't get held up forever.

acrellin:
@stefanv I just opened an issue for that so that it doesn't get forgotten. If this looks good to you, please go ahead and merge.

acrellin:
Closed in favor of https://github.com/skyportal/skyportal/pull/447

PR #384, opened 2020-03-27
Scanning Page #3: Update handlers to account for new candidates table

acrellin:
The changes in this PR account for the fact that thumbnails, comments, photometry, spectra, plots, etc may now have either a candidate, a source, or both as parents.

Also, allow for plot height/width to be specified in API call as query parameters (needed in a subsequent PR, but tied to the handler changes here)


acrellin:
[ch810]

skyportal/handlers/api/internal/plot.py
```diff
@@ -7,22 +7,36 @@
 class PlotPhotometryHandler(BaseHandler):
     @auth_or_token
     def get(self, source_id):
```
profjsb, line 9: Do we need an OpenAPI spec for an internal API?

skyportal/handlers/api/photometry.py
```diff
@@ -41,46 +34,59 @@ def post(self):
...
+                scale=data["time_scale"],
+            )
+            observed_at = arrow.get(t.tcb.iso)
+            converted_times.append(observed_at)
+            p = Photometry(
+                source_id=source_id,
+                observed_at=observed_at,
+                mag=data["mag"][i],
+                e_mag=data["e_mag"][i],
+                time_scale="tcb",
```
profjsb, line 68: these will be removed in #382 I believe.

skyportal/handlers/api/source.py
```diff
@@ -52,54 +57,72 @@ def get(self, source_id=None):
...
-                     .filter(Source.dec <= dec + radius)\
-                     .filter(Source.dec >= dec - radius)
+                    return self.error(
+                        "Invalid values for ra, dec or radius - could not convert to float"
+                    )
+                q = (
+                    q.filter(Source.ra <= ra + radius)
+                    .filter(Source.ra >= ra - radius)
+                    .filter(Source.dec <= dec + radius)
+                    .filter(Source.dec >= dec - radius)
```
profjsb, line 124: this is not a true cone search: it will fail near the poles and also near ra 360->0.

skyportal/handlers/api/thumbnail.py
```diff
@@ -173,30 +173,30 @@ def delete(self, thumbnail_id):
...
-        basedir/f'static/thumbnails/{source_id}_{thumbnail_type}.png')
+        basedir / f'static/thumbnails/{source_id}_{thumbnail_type}.png')
     if not os.path.exists(os.path.dirname(file_uri)):
-        (basedir/'static/thumbnails').mkdir(parents=True)
+        (basedir / 'static/thumbnails').mkdir(parents=True)
     file_bytes = base64.b64decode(thumbnail_data)
     im = Image.open(io.BytesIO(file_bytes))
     if im.format != 'PNG':
         raise ValueError('Invalid thumbnail image type. Only PNG are supported.')
     if not (100, 100) <= im.size <= (500, 500):
```
profjsb, line 197: This does not do what you think it does. When the second element is outside the range, the condition will not be met. Ie.
```
(100, 100) <= (100, 2000) <= (500, 500)  == True
```

skyportal/handlers/api/internal/plot.py
acrellin: We typically don't have them, no

skyportal/handlers/api/photometry.py
acrellin: We still have these fields in that PR - perhaps a future PR will need to update this

skyportal/handlers/api/source.py
acrellin: Yeah, that's just been a placeholder for now. I think there's already an open issue for that, but I'll open one now if not. Since this isn't new in this PR, we won't handle it here (it's only in the diff due to formatting changes).

skyportal/handlers/api/thumbnail.py
acrellin: Noted, thanks for catching that. Opening an issue and will take care of it in an upcoming PR.

acrellin:
@stefanv I just realized that some formatting changes made it into this PR accidentally, but it should be obvious that that's all those parts of the diff are

stefanv:
I have the same question over here, I guess: is the duplication, source_id + comment_id, for all our objects strictly necessary?  It doesn't feel like the correct design pattern.  Does it matter whether a comment is on a source or a candidate, if those live inside the same table?

acrellin:
@stefanv what do you mean source_id & comment_id?  Do you mean candidate_id?  Yes, for comments, we need to track whether they're associated with a candidate or a source (I believe candidate auto-annotations should not propagate to the associated source).

acrellin:
@stefanv we've discussed this multiple times before (weeks ago), and I thought we were on the same page about needing to do it this way.

acrellin:
Closing in favor of https://github.com/skyportal/skyportal/pull/447

PR #387, opened 2020-03-31
Follow-up requests

acrellin:
Sample available parameters 1:
followup_requests

Sample available parameters 2:
followup_requests_2


acrellin:
[ch759]

acrellin:
[ch758]

acrellin:
@stefanv this still needs tests and an `editRequest` handler, but 99% of the code is in now. If you could please review this in the next week, I will add the remaining pieces in a separate commit that you can review independently afterwards.

instrument_observation_params.json
```diff
@@ -0,0 +1,21 @@
+{
```
acrellin, line 1: Alternatively, this could live in a JSONB column in the instruments table.

dannygoldstein:
Is there a way to use this to request spectroscopy, e.g., from SEDM on the p60?

acrellin:
We could easily extend this to add spectroscopic followup requests

instrument_observation_params.json
profjsb: Perhaps check to see if this file exists or not (or is empty). In which case, this part of the Source page is not rendered.

skyportal/handlers/api/followup_request.py
```diff
@@ -0,0 +1,84 @@
...
+                          type: string
+                          description: New follow-up request ID
+        """
+        data = self.get_json()
+        _ = Source.get_if_owned_by(data["source_id"], self.current_user)
+        data["start_date"] = arrow.get(data["start_date"]).datetime
+        data["end_date"] = arrow.get(data["end_date"]).datetime
+        data["requester_id"] = self.current_user.id
+        if isinstance(data["filters"], str):
+            data["filters"] = [data["filters"]]
```
profjsb, line 38: Should be catching errors here -- for instance if the `start_date` is improperly formatted (ie. arrow throws an error)

static/js/components/FollowupRequestForm.jsx
```diff
@@ -0,0 +1,260 @@
...
+        };
+      }
+    }
+    setFormState({
+      ...formState,
+      ...newState
+    });
+  };
+
+  const handleSubmit = () => {
```
profjsb, line 61: here's where you can validate the dates on the client side (e.g. https://www.telerik.com/kendo-react-ui/components/dateinputs/datepicker/forms/#toc-setting-validation-requirements)

skyportal/handlers/api/followup_request.py
acrellin: @profjsb if the parse fails, it automatically returns with an error code & message such that a notification is displayed on the front-end with the arrow exception message: `Backend error: Could not match input 'mydate' to any of the following formats: YYYY-MM-DD, YYYY-M-DD, YYYY-M-D, YYYY/MM/DD, YYYY/M/DD, YYYY/M/D, YYYY.MM.DD, YYYY.M.DD, YYYY.M.D, YYYYMMDD, YYYY-DDDD, YYYYDDDD, YYYY-MM, YYYY/MM, YYYY.MM, YYYY`. Does that suffice?

instrument_observation_params.json
acrellin: Done (see the handler method)

static/js/components/FollowupRequestForm.jsx
acrellin: How does this look? https://reactdatepicker.com/

skyportal/handlers/api/followup_request.py
profjsb: Sure.

static/js/components/FollowupRequestForm.jsx
profjsb: Yes, good. For a picker this is fine. But does it also validate on the client side?

acrellin: It doesn't allow for invalid dates - if you enter some invalid text, upon element blur, it reverts back to the last valid input.

acrellin: Done.

acrellin:
Updated UI screenshot to show date picker: 
followup_requests_form

acrellin:
[ch864]

acrellin:
Now includes an edit button which, when clicked, displays a modal with a form to edit the follow-up request. Here it is in action (note that the relevant row is automatically updated via web-socket message & re-fetching of data asynchronously): 


edit_followup_request_form

acrellin:
[ch873]

skyportal/handlers/api/followup_request.py
stefanv: As noted in the meeting earlier today, we should catch all errors so they don't get thrown to the backend logs.

skyportal/handlers/api/followup_request.py
```diff
@@ -0,0 +1,129 @@
...
+                    - Success
+                    - type: object
+                      properties:
+                        id:
+                          type: string
+                          description: New follow-up request ID
+        """
+        data = self.get_json()
+        _ = Source.get_if_owned_by(data["source_id"], self.current_user)
+        data["start_date"] = arrow.get(data["start_date"]).datetime
```
stefanv, line 34: The arguments should be documented.

skyportal/handlers/api/internal/instrument_observation_params.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class InstrumentObservationParamsHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        try:
+            with open("instrument_observation_params.json") as f:
+                instrument_data = json.load(f)
+        except FileNotFoundError:
+            return self.error("instrument_observation_params.json does not exist.")
```
stefanv, line 13: This leaks implementation details to the user.  How about something like: "Instrument observation parameters unavailable."

skyportal/handlers/api/internal/instrument_observation_params.py
```diff
@@ -0,0 +1,17 @@
...
+        except json.JSONDecodeError:
+            return self.error("JSON parse error: instrument_observation_params.json "
```
stefanv, line 15: Same here:

"System misconfiguration: invalid instrument observation parameters."

static/js/components/EditFollowupRequestDialog.jsx
```diff
@@ -0,0 +1,54 @@
...
+    <span key={followupRequest.id}>
+      <button type="button" onClick={handleClickOpen}>
+        Edit
+      </button>
+      <Dialog open={open} onClose={handleClose} style={{ position: "fixed" }}>
+        <DialogContent>
+          <FollowupRequestForm
+            source_id={followupRequest.source_id}
+            action="editExisting"
+            followupRequest={followupRequest}
```
stefanv, line 30: You should be able to unpack these automatically without specifying each field.

static/js/components/FollowupRequestForm.jsx
```diff
@@ -0,0 +1,288 @@
...
+import PropTypes from 'prop-types';
+import DatePicker from 'react-datepicker';
+
+import { showNotification } from 'baselayer/components/Notifications';
+import * as Actions from '../ducks/source';
+
+import 'react-datepicker/dist/react-datepicker-cssmodules.css';
+
+
+const FollowupRequestForm = ({ source_id, action, instrumentList, instrumentObsParams, followupRequest = null, title = "Submit new follow-up request", afterSubmit = null }) => {
```
stefanv, line 12: This form could probably benefit quite a bit from React Hook Form!  Will let you investigate and see if you agree!

webpack.config.js
```diff
@@ -52,7 +52,7 @@ const config = {
       // Enable CSS Modules for Skyportal
       {
         test: /\.css$/,
-        include: /static\/js/,
+        include: [/static\/js/, /node_modules\/react-datepicker\/dist/],
```
stefanv, line 55: Why is this necessary?

skyportal/handlers/api/followup_request.py
acrellin: We specify the schema, as in our other handlers, so these fields will automatically show up in the API docs

skyportal/handlers/api/internal/instrument_observation_params.py
acrellin: Sure, done.

static/js/components/EditFollowupRequestDialog.jsx
acrellin: Can you please explain? This is just standard props-passing

static/js/components/FollowupRequestForm.jsx
acrellin: Will investigate...

webpack.config.js
acrellin: This is needed for this import line: `import 'react-datepicker/dist/react-datepicker-cssmodules.css';`

skyportal/handlers/api/followup_request.py
stefanv: The input fields do not show up automatically for get requests, do they?

acrellin: this is a POST handler...

acrellin: Look at any other POST handler -- we always use the schema to create the API docs. Look above here in this file:
```
        requestBody:
          content:
            application/json:
              schema: FollowupRequest
```

stefanv: You're right; thanks.

static/js/components/EditFollowupRequestDialog.jsx
stefanv:
```
<ChildComponent {...props} />
```

acrellin: Only 3 of the 7 props being passed into `<ChildComponent>` are also props of `<ParentComponent>` (`followupRequest, instrumentList, instrumentObsParams`), and since we follow the object destructuring pattern for props in our functional component definitions, this isn't available here.

acrellin: Sure, I changed the function signature to use a single props object and not use destructuring so that we can do `{...props}` here.

acrellin: Actually, that's an anti-pattern - reverting to what I had done originally which is the correct form. 
```
/home/travis/build/skyportal/skyportal/static/js/components/EditFollowupRequestDialog.jsx
  21:16  error  Must use destructuring props assignment  react/destructuring-assignment
  28:24  error  Must use destructuring props assignment  react/destructuring-assignment
  32:13  error  Prop spreading is forbidden              react/jsx-props-no-spreading
```

stefanv: Ooh forbidden! :fire:

stefanv:
Loving the new hook-form integration—thanks for figuring that out for us, Ari!

In it goes :rocket: !

dannygoldstein:
Hi guys, a few minor comments - 

  * Exposure time should be a free form text entry field with a default value and min and max values for each instrument

Details - 

  * P60 Rainbow Camera only has sdssu, sdssg, sdssr, and sdssi filters

  * ALFOSC has Bessell UBVRI and some other filters, I think @arielgoobar knows the full filter set

PR #400, opened 2020-04-08
Fix and improve docker-compose deployment

### LINKED ISSUE #377: docker-compose enhancements requests

stefanv:
- Update docker-compose file to v3
- Upgrade postgres, and add now-required password
- Store database data in a volume that lives outside of web container
- Add deployment docs
- Update node to 12 in Docker images
- Add config.yaml to .dockerignore so it doesn't get added to images,
  accidentally exposing tokens

Closes #377

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,11 @@
-.git
-.ipynb_checkpoints
+*~
+.*
 log
 worker*
 tmp
 node_modules
-ptf_data
+run
+config.yaml
+test_config.yaml
+github_deploy_key.enc
+dbdata
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ static/thumbnails
 *.lock
 *.dirlock
 package-lock.json
+.jobs_timestamps.yaml
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM ubuntu:18.04
 
 RUN apt-get update && \
     apt-get install -y curl build-essential software-properties-common && \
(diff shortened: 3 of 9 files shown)


profjsb:
@dmitryduev @acrellin are we ready to merge this?

stefanv:
@dmitryduev Looking at your PR, I think there is only one change that I don't have (forcing restart of postgres).  Is that one important?

dmitryduev:
Looks great! My only concern is with `docker.yaml` and `docker-compose.yaml`: if you change those, which you should, what happens the next time you `git pull`? Wouldn't `git` complain? That's why I wanted to rename them into *.defaults

dmitryduev:
As for `restart: always` -- I would add it, find it very useful. Otherwise if the db container dies for some reason, you'll have to manually restart it.

stefanv:
@dmitryduev I had no problem loading demo data locally; did you test it and experience problems?

For Fritz, we probably want to make a new docker-compose config that [extends](https://docs.docker.com/compose/extends/#multiple-compose-files) the SkyPortal and Kowalski ones and a docker.yaml (or fritz.yaml), mounting the latter into the skyportal container as config.yaml.

I've added restarting to the db container.

dmitryduev:
@stefanv yea, I tried running it inside a container on my mac and on a linux machine, hangs every time.
Agree on the fritz point.

PR #403, opened 2020-04-12
ObservingRun model [ch823] [ch824]

dannygoldstein:
Adds two database models: `ObservingRun`, which represents a single, non-robotic follow-up run with a PI, and `Assignment`, which is effectively a join table between `ObservingRun` and `Source` that keeps track of Source assignments to the `ObservingRun`.

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -33,8 +33,6 @@ def is_owned_by(self, user_or_token):
         return (user_or_token in self.users)
     else:
         raise NotImplementedError(f"{type(self).__name__} object has no owner")
-
-
 Base.is_owned_by = is_owned_by
 
 
@@ -82,12 +80,9 @@ User.group_users = relationship('GroupUser', back_populates='user',
 User.groups = relationship('Group', secondary='group_users',
                            back_populates='users')
 
-
 @property
 def token_groups(self):
     return self.created_by.groups
-
-
 Token.groups = token_groups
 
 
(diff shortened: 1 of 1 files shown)


dannygoldstein:
[ch823]

dannygoldstein:
[ch824]

acrellin:
@dannygoldstein any overlap between this and https://github.com/skyportal/skyportal/pull/387, or are followups and observing runs distinct concepts?

dannygoldstein:
There's overlap between the `FollowupRequest` of #387 and the `Assignment` of this PR. In the current (GROWTH) marshal these are separate entities: `FollowupRequest` as currently implemented would be mainly for robotic facilities that can be triggered remotely, whereas `Assignment` would be for classical nights where a human is actually required to make decisions about what will get observed and in what order. But it is worth considering whether `Assignment` and `FollowupRequest` should be merged together on the Skyportal backend as they represent very similar things.

acrellin:
> There's overlap between the `FollowupRequest` of #387 and the `Assignment` of this PR. In the current (GROWTH) marshal these are separate entities: `FollowupRequest` as currently implemented would be for robotic facilities that can be triggered remotely, whereas `Assignment` would be for classical nights where a human is actually required to make decisions about what will get observed and in what order. But it is worth considering whether `Assignment` and `FollowupRequest` should be merged together on the Skyportal backend as they represent very similar things.

@stefanv @profjsb thoughts on combining `FollowupRequest` and `Assignment` in SkyPortal? As @dannygoldstein noted, in the current marshal they represent two slightly different things (robotic vs manual observations).

acrellin:
@dannygoldstein the the followup requests PR merged, I'm happy to work with you on extending that to encapsulate any manual observation concepts from this PR.

dannygoldstein:
closed in favor of a future PR that will be based off #387

PR #407, opened 2020-04-27
Catch handler errors & return `self.error` for cleaner logs; fix image size check bug

### LINKED ISSUE #401: Properly check thumbnail image size in thumbnail handler
Currently only the first dimension is properly checked

acrellin:
Closes #401


skyportal/handlers/api/thumbnail.py
```diff
@@ -183,20 +186,20 @@ def delete(self, thumbnail_id):
...
+        (basedir / 'static/thumbnails').mkdir(parents=True)
     file_bytes = base64.b64decode(thumbnail_data)
     im = Image.open(io.BytesIO(file_bytes))
     if im.format != 'PNG':
         raise ValueError('Invalid thumbnail image type. Only PNG are supported.')
-    if not (100, 100) <= im.size <= (500, 500):
+    if not all(100 <= x <= 500 for x in im.size):
         raise ValueError('Invalid thumbnail size. Only thumbnails '
-                        'between (100, 100) and (500, 500) allowed.')
+                         'between (100, 100) and (500, 500) allowed.')
```
profjsb, line 202: Is there a reason for the lower-limit on the thumbnail size? LSST will be providing 31x31 pixel images it seems (e.g. https://github.com/lsst-dm/sample-avro-alert/tree/master/examples). Could open a separate issue or perhaps we just lower it...?

acrellin: Sure - remove it altogether or just lower it to 16?

acrellin: Done.

PR #416, opened 2020-04-30
Add initial support for dark mode

stefanv:
- Adds a UI preference section to /profile
- Injects Material UIs default CSS into the app
- Relies on the existing profile preferences saving to store the
  preferred theme to the db

This should make @dannygoldstein happy.

Next up: switch to stylized components, and illustrate how to build custom components that take the theme into consideration (such as our lightcurve plotting).


acrellin:
Looks great!

skyportal/handlers/api/internal/token.py
```diff
@@ -34,9 +34,7 @@ def post(self):
         user = (User.query.filter(User.username == self.current_user.username)
                     .first())
         user_acls = {acl.id for acl in user.acls}
-        requested_acls = {k.replace('acls_', '') for k, v in data.items() if
-                          k.startswith('acls_') and v == True}
-        token_acls = requested_acls & user_acls
+        token_acls = set(data['acls']) & user_acls
```
acrellin, line 37: Nice simplification on this end.

profjsb:
Looks great. Naive question: Do we need to think now about how pages render in mobile/tablet form or do we get a default behavior with Material?

acrellin:
@stefanv I'm seeing lots of browser warnings along the lines of `A component is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component.` How sure are you this is the correct way to use `Controller` and `FormControlLabel` with the `Checkbox` component?

static/js/components/TokenList.jsx
```diff
@@ -26,67 +35,53 @@ const TokenList = ({ tokens }) => {
...
-              </b>
-              &nbsp;&nbsp;
-            </td>
-            <td>
-              <b>
-                Delete
-              </b>
-            </td>
-          </tr>
+      <Table size="small" component={Paper}>
```
acrellin, line 38: Any way to make this part work without getting these browser warnings? `Warning: validateDOMNesting(...): <tbody> cannot appear as a child of <div>.`

acrellin: How about wrapping `Table` in `Paper` instead, as in this example? https://codesandbox.io/s/2omrn3oq30?file=/demo.js

acrellin: Can confirm wrapping Table in Paper instead does not produce any warnings.

stefanv:
> Looks great. Naive question: Do we need to think now about how pages render in mobile/tablet form or do we get a default behavior with Material?

It looks pretty good out of the box.  But I think the one improvement that can be made here is that tokens should not be displayed in a table but in a list.  Tables are awkward to render on smaller displays, because they are forced to put all elements of a row horizonally.

To see how things render on mobile, do ctrl-shift I (or switch the developer mode), then find the browser's cell-phone emulator (for firefox, it's a little cell-phone icon on the right).

static/js/components/NewTokenForm.jsx
```diff
@@ -1,119 +1,127 @@
...
-      }
-    });
-    setFormState({
-      group_id: "",
-      name: "",
-      ...acls_state
-    });
+  const defaultValues = {
+    name: '',
+    acls: Array(50).fill(false) // This should be `acls.length`, but see https://github.com/react-hook-form/react-hook-form/issues/1558
```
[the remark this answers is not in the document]
stefanv: Unfortunately, no solution for this in sight.  I have a ticket open at https://github.com/react-hook-form/react-hook-form/issues/1558

static/js/components/TokenList.jsx
stefanv: Fixed, thanks!

stefanv:
@acrellin Failure is unrelated.

acrellin:
@stefanv just restarted the build, hopefully it was a random issue. Will merge when it passes.
Looks great, thanks!

acrellin:
@stefanv I'll push up a fix for that failing test shortly - in it goes! Great work! :rocket: :sparkles:

PR #430, opened 2020-05-01
Update photometry data model

acrellin:


dannygoldstein:
See the conversation in #191 -- we need to determine if mag, magerr, flux, and fluxerr should all be allowed to be free parameters or if we require only flux and fluxerror (and neither is nullable)

stefanv:
@acrellin I'll let you and @dannygoldstein sort this one out; go ahead and merge when you're ready.

dannygoldstein:
Superseded by #191

PR #434, opened 2020-05-05
Jump to page range

Jenniferrrrrr:
- Modified the jump to page field's logic so any invalid page number entered rounds to the closest valid one(ex. -1 rounds to 1)


config.yaml
```diff
@@ -0,0 +1,33 @@
+app:
```
acrellin, line 1: This file shouldn't be committed

acrellin:
@Jenniferrrrrr can you please update the relevant tests that look for page range error notifications?

PR #437, opened 2020-05-05
Correctly reset new token form

stefanv:
See also https://github.com/react-hook-form/react-hook-form/issues/1558

### DIFF
--- a/static/js/components/NewTokenForm.jsx
+++ b/static/js/components/NewTokenForm.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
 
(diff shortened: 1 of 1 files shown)


static/js/components/NewTokenForm.jsx
```diff
@@ -24,14 +24,14 @@ import * as ProfileActions from '../ducks/profile';
...
-  const { handleSubmit, register, errors, reset, control } = useForm({
-    defaultValues
-  });
+  const { handleSubmit, register, errors, reset, control } = useForm();
+
+  useEffect(() => {
+    reset({
+      acls: Array(acls.length).fill(false)
+    });
+  }, [reset, acls]);
```
acrellin, line 34: Why is `reset` included here? Isn't that a static function? Does this work as expected without `reset` here?

acrellin: Can confirm that just `[acls]` works as expected, and I think is best here, as this hook isn't watching the value of `reset`, just `acls`.

acrellin:
```suggestion
  }, [acls]);
```

stefanv:
I had a warning in the codebox version when removing it, but I'll investigate a bit more. I also want to take a look at useFieldArray and see if that is suitable for our needs.

PR #439, opened 2020-05-06
Group management reskin

stefanv:
Please see individual commits for descriptions.


static/js/components/Source.jsx
```diff
@@ -19,129 +19,136 @@ import FollowupRequestList from './FollowupRequestList';
...
-            action="createNew"
-            instrumentList={instrumentList}
-            instrumentObsParams={instrumentObsParams}
-          />
-          <FollowupRequestList
-            followupRequests={source.followup_requests}
-            instrumentList={instrumentList}
-            instrumentObsParams={instrumentObsParams}
-          />
+        <div className={styles.name}>
```
acrellin, line 67: Is everything below supposed to be in this div with className `styles.name`? That makes it sound like it's just supposed to contain the source name text.

stefanv: I didn't change anything here, at least not intentionally. It is highlighted because the indentation changed.

acrellin: Ah, I see this is the same. My mistake

PR #440, opened 2020-05-06
Memoize reset with acls argument and use in useEffect

acrellin:
As per https://overreacted.io/a-complete-guide-to-useeffect/#tldr: 

> 🤔 Question: Do I need to specify functions as effect dependencies or not?
> 
> The recommendation is to hoist functions that don’t need props or state outside of your component, and pull the ones that are used only by an effect inside of that effect. If after that your effect still ends up using functions in the render scope (including function from props), wrap them into useCallback where they’re defined, and repeat the process. Why does it matter? Functions can “see” values from props and state — so they participate in the data flow. There’s a more detailed answer in our FAQ.

### DIFF
--- a/static/js/components/NewTokenForm.jsx
+++ b/static/js/components/NewTokenForm.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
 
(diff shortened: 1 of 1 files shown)


stefanv:
@acrellin I studied the documentation you shared carefully, and I am pretty sure the intent is to have functions define their dependencies *upon declaration* (as a way of tidying up `useEffect` dependency declaration):

> As a last resort, you can add a function to effect dependencies but wrap its definition into the `useCallback` Hook.

Since we are *given* the `reset` function, it does not make sense to add it into a `useCallback`: such a memoization would depend on two variables changing: `reset` and `acls`, the same same variables that we already had listed inside our `useEffect`.

The only way in which we could avoid re-renders, is if `reset` was defined using `useCallback`, which it is, with zero dependencies as expected.

https://github.com/react-hook-form/react-hook-form/blob/master/src/useForm.ts#L1275

So, I think we're good in this case.  Let me know if you understand it differently.

acrellin:
Ah, okay, so `reset` is already a memoized hook - that's great. Thanks for finding that. Closing.

PR #446, opened 2020-05-12
Collapse several Obj metadata columns into altdata JSONB column; update babel-related packages for ES2020

acrellin:
Closes https://github.com/skyportal/skyportal/issues/154


skyportal/handlers/api/source.py
```diff
@@ -121,9 +121,10 @@ def get(self, obj_id=None):
                 end_date = arrow.get(end_date.strip())
                 q = q.filter(Obj.last_detected <= end_date)
             if simbad_class:
-                q = q.filter(func.lower(Obj.simbad_class) == simbad_class.lower())
+                q = q.filter(func.lower(Obj.altdata['simbad']['class'].astext)
+                             == simbad_class.lower())
             if has_tns_name == 'true':
-                q = q.filter(Obj.tns_name.isnot(None))
+                q = q.filter(Obj.altdata['tns']['name'].isnot(None))
```
stefanv, line 127: What happens if tns entry is not present?

acrellin: @stefanv it just treats the resulting property (`altdata['tns']['name']`) as `None`

acrellin: that is to say, we don't need to worry about these keys not being present -- it handles it gracefully

stefanv:
Nice discovery on the ES2020 property checking!

acrellin:
This PR now uses newer babel-related packages to use ES2020-style nested property checking

PR #448, opened 2020-05-14
We don't need jquery, so don't include it

stefanv:


webpack.config.js
```diff
@@ -88,12 +87,6 @@ const config = {
     ]
   },
   plugins: [
-    // We do not use JQuery for anything in this project; but Bootstrap
```
acrellin, line 91: @stefanv we've removed bootstrap entirely?

stefanv:
We purposefully never used it, as far as I'm aware.

PR #454, opened 2020-05-20
Scanning page front-end

acrellin:


skyportal/models.py
```diff
@@ -252,6 +251,25 @@ def get_source_if_owned_by(obj_id, user_or_token, options=[]):
 Source.is_owned_by = source_is_owned_by
 Source.get_if_owned_by = get_source_if_owned_by
 
+
+def get_obj_if_owned_by(obj_id, user_or_token, options=[]):
```
acrellin, line 255: Note that this will be re-written in an upcoming ACLs PR

static/js/components/CandidateList.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+
+const CandidateList = () => {
+  const { candidates } = useSelector((state) => state.candidates);
+
+  const userGroups = useSelector((state) => state.groups.user);
+
+  const dispatch = useDispatch();
+
+  useEffect(() => {
```
stefanv, line 21: Ari, out of curiosity---is this pattern safe for loading?  What happens if the fetch fails?

acrellin: Curious what you mean by safe here - if it fails, it won't try again until the page is refreshed, if that's what you're referring to

acrellin: (because `candidates` doesn't change)

acrellin: @stefanv I just tested this by making the handler error, and the back-end error is shown as a front-end notification once (as happens automatically with all of our fetch failures), but it doesn't attempt to fetch again until the user reloads the page or clicks "Submit" on the filtering form. That all seems like what should happen to me.

skyportal/tests/frontend/test_scanning_page.py
```diff
@@ -0,0 +1,229 @@
...
+                "transient": False,
+                "ra_dis": 2.3,
+                "filter_ids": [public_filter.id],
+            },
+            token=upload_data_token,
+        )
+        assert status == 200
+        assert data["data"]["id"] == f"{candidate_id}_{i}"
+
+    status, data = api(
```
stefanv, line 113: Is this group being used anywhere?

skyportal/tests/frontend/test_scanning_page.py
```diff
@@ -0,0 +1,229 @@
...
+    driver.get("/candidates")
+    for i in range(5):
+        driver.wait_for_xpath(f'//a[text()="{candidate_id}_{i}"]')
+    start_date_input = driver.wait_for_xpath("//input[@name='startDate']")
+    start_date_input.clear()
+    start_date_input.send_keys("20001212")
+    end_date_input = driver.wait_for_xpath("//input[@name='endDate']")
+    end_date_input.clear()
+    end_date_input.send_keys("20011212")
+    time.sleep(0.1)
```
stefanv, line 189: Why do we need a sleep instead of a wait_for here?

static/js/API.js
```diff
@@ -77,6 +77,10 @@ function POST(endpoint, actionType, payload) {
   return API(endpoint, actionType, 'POST', payload);
 }
 
+function PATCH(endpoint, actionType, payload) {
```
stefanv, line 80: :+1:

stefanv:
If you think any of these changes warrant changes, go ahead and make them in a follow-up PR.

Merging for now to keep things rolling along!

skyportal/tests/frontend/test_scanning_page.py
acrellin: Because the button is already rendered, but without this, it seems to occasionally submit before RHF has updated its internal values

skyportal/tests/frontend/test_scanning_page.py
acrellin: Good catch -- removed in https://github.com/skyportal/skyportal/pull/470/commits/21cb443d48b9f427ab7e3d45a674dbac8f25b837

acrellin:
> If you think any of these changes warrant changes, go ahead and make them in a follow-up PR.
> 
> Merging for now to keep things rolling along!

Implemented your suggestions in another PR -- thanks, @stefanv!

Issue #352, opened 2020-02-14
Incorrect API documentation for `api/group.py`


stefanv:
https://skyportal.io/docs/api.html#/paths/~1api~1groups~1group_id/get

https://skyportal.io/docs/api.html#/paths/~1api~1groups/get

Returned JSON has different fields

acrellin:
Can you explain what the issue is? This isn't clear.

stefanv:
The fields returned are different from those described in the docstrings.

acrellin:
@stefanv we need to figure out a way to nest our model schemas as properties of larger dictionaries, ideally in a general way across all our API docs. Currently this issue applies to many of the API docs. In the API docs we have: 
```
          responses:
            200:
              content:
                application/json:
                  schema: SingleGroup
```
but the response body's JSON is actually of the form
```
{'status', 'success',
'data': {'group': {...} }
``` 
Let's chat about how to update our schemas accordingly

acrellin:
Also, I see that for the groups endpoint, the docs issue is threefold: a) the issue described in the previous comment, b) the schema includes more relationships than are actually shipped, and c) our GET `api/groups` endpoint is a hack-job from the early SkyPortal days that returns "user_groups" and "all_groups", which should probably be split into separate queries or moved to a private endpoint.

PR #463, opened 2020-05-23
Fix & expand API docs; update API response data structure; update front-end and tests

acrellin:
API docs previously described incorrect JSON response structures throughout and were missing many fields.
We move away from the pattern of wrapping all response data in a `dict` in API handlers (e.g. `return self.success(data={'sources': q.all()})`) to passing the relevant data directly to `data` whenever applicable (e.g. `return self.success(data=q.all())`). The API docs are updated accordingly. 
Closes https://github.com/skyportal/skyportal/issues/352

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ doc_reqs:
 
 api-docs: | doc_reqs
 	@PYTHONPATH=. python tools/openapi/build-spec.py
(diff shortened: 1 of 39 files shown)


stefanv:
Thanks, Ari, much appreciated.

stefanv:
I will look into a mechanism to automatically verify the returns.

PR #468, opened 2020-05-27
Refine Instrument data model

dannygoldstein:
This PR introduces some slight refinements of the `Instrument` data model, namely:

   * The introduction of a `filters` field, that is an `Array` of `Enum`. The allowed Enum values are the filters defined by our photometry library, `sncosmo`. The `filters` of an instrument can be `[]`, indicating no filters. This would be appropriate for, e.g., a Spectrograph without any imaging capabilities. 
   * Update of the `InstrumentHandler` to this modification and proper rendering of the allowed filter values for new instruments in the api docs:

Screen Shot 2020-05-26 at 10 39 22 PM

   * Requirement that a user be a system admin to update or delete instruments. This is because follow-up requests will assume a certain schema for instruments, and we dont want anyone messing around with the capabilities of the instruments or it could affect other users' follow-up requests.

### DIFF
--- a/skyportal/handlers/api/instrument.py
+++ b/skyportal/handlers/api/instrument.py
@@ -1,39 +1,13 @@
 from marshmallow.exceptions import ValidationError
 from baselayer.app.access import permissions, auth_or_token
 from ..base import BaseHandler
-from ...models import DBSession, Instrument, Telescope, GroupTelescope
+from ...models import (DBSession, Instrument, Telescope, GroupTelescope,
+                       ALLOWED_BANDPASSES)
 
 
 class InstrumentHandler(BaseHandler):
     @permissions(['Upload data'])
     def post(self):
-        """
-        ---
-        description: Add a new instrument
-        requestBody:
-          content:
-            application/json:
-              schema: InstrumentNoID
-        responses:
-          200:
-            content:
-              application/json:
-                schema:
-                  allOf:
-                    - $ref: '#/components/schemas/Success'
-                    - type: object
-                      properties:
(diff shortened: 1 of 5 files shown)


dannygoldstein:
failing travis only due to a transient github issue - newest commit only introduces a comment and not any substantive modifications to the code from the previous (passing) commit

dannygoldstein:
[ch941]

skyportal/tests/conftest.py
```diff
@@ -137,6 +137,18 @@ def manage_users_token(super_admin_user):
...
+    role = Role.query.get('Super admin')
+    token_id = create_token(
+        permissions=[a.id for a in role.acls],
+        created_by_id=super_admin_user.id,
+        name=str(uuid.uuid4()),
+    )
+    return token_id
+
+
+
```
acrellin, line 151: One too many blank lines

acrellin:
> @dannygoldstein `load_demo_data` will also need to be updated, as `filters` is required

Scratch that -- I see that it has a default value. My mistake. Looks good.

skyportal/handlers/api/instrument.py
```diff
@@ -170,3 +147,45 @@ def delete(self, instrument_id):
         DBSession().commit()
 
         return self.success()
+
+
+fdoc = "List of filters on the instrument. If the instrument has no filters " \
```
stefanv, line 152: Why not inline this?

stefanv:
Thanks, @dannygoldstein !

PR #470, opened 2020-05-28
Post photometry via frontend

acrellin:
This PR introduces an upload photometry page that allows users to post raw photometry data for a source.  The CSV string is validated, then parsed & displayed in tabular form before the user is able to submit it.  

Misc. metadata columns can be provided that will ultimately be part of the photometry row's `altdata` JSONB column by preceding the column names with "altdata." (e.g. "altdata.metadata1").

See below gif for a live demo, including real-time validation

upload_photometry_2


acrellin:
[ch908]

profjsb:
Very nice. I think it would be useful to assign a "bulk upload" ID to such posts (not necessarily from the just from frontend) to facilitate a "bulk delete" (eg. if you upload photometry but need to change it because you made an error with the ZP or the bandpass). User could be shown a dialog after they upload, something like "Upload complete. To delete this data `curl -i DELETE http://..../photmetry/remove?upload_id=1394`"

acrellin:
> Very nice. I think it would be useful to assign a "bulk upload" ID to such posts (not necessarily from the just from frontend) to facilitate a "bulk delete" (eg. if you upload photometry but need to change it because you made an error with the ZP or the bandpass). User could be shown a dialog after they upload, something like "Upload complete. To delete this data `curl -i DELETE http://..../photmetry/remove?upload_id=1394`"

Good idea - should we do that data model/handler update here, or in another PR?

stefanv:
I like that idea very much, @profjsb!

acrellin:
@profjsb @stefanv here's a demo of that requested feature in action:

bulk_delete_upload_phot

acrellin:
Ready for review. 

Question: which fields should be scalar inputs/selects and which should be required columns of the CSV? Currently `["mjd", "flux", "fluxerr", "zp", "zpsys", "filter"]` are required columns. Instrument is a select. What other fields should we add as required, either in the CSV or as other inputs?

acrellin:
[ch962]

acrellin:
[ch963]

config.yaml.defaults
```diff
@@ -28,6 +28,8 @@ app:
       component: SourceList
     - path: "/user/:id"
       component: UserInfo
+    - path: "/upload_photometry/:id"
```
dannygoldstein, line 31: Can we rename this to bulk_upload_photometry?

dannygoldstein: Or maybe, `upload_photometry_from_csv`? The current photometry handler already lets you upload photometry in bulk

skyportal/app_server.py
```diff
@@ -11,7 +11,7 @@
     GroupHandler, GroupUserHandler,
     InstrumentHandler,
     NewsFeedHandler,
-    PhotometryHandler,
+    PhotometryHandler, BulkDeletePhotometryHandler,
```
dannygoldstein, line 14: Any reason not to incorporate this functionality into `PhotometryHandler.delete`?

skyportal/models.py
```diff
@@ -392,6 +391,8 @@ class Photometry(Base):
 
     origin = sa.Column(sa.String, nullable=True)
 
+    bulk_upload_id = sa.Column(sa.String, nullable=True)
```
dannygoldstein, line 394: Could this be assigned by `PhotometryHandler.post`?

skyportal/tests/frontend/test_scanning_page.py
```diff
@@ -110,14 +110,6 @@ def test_candidate_unsaved_only_filtering(
         assert status == 200
         assert data["data"]["id"] == f"{candidate_id}_{i}"
 
-    status, data = api(
```
dannygoldstein, line 113: Why delete this?

config.yaml.defaults
acrellin: this isn't the API endpoint, just the path name for the page that displays the form. still want it changed?

skyportal/app_server.py
acrellin: just seems simpler -- if we combine them, we'll have to check to see if the value passed in is a valid phot ID or a bulk upload ID

config.yaml.defaults
dannygoldstein: Ah, no nevermind

skyportal/models.py
acrellin: it is created in `PhotometryHandler.post`

skyportal/tests/frontend/test_scanning_page.py
acrellin: this is totally unrelated, just cleaning up some unused extraneous code here

acrellin: probably should've just pushed this straight up to master

acrellin:
> Main comments are:
> 
>     * It would be useful to be able to upload data from multiple instruments in a single CSV, with some kind of instrument identifier (if not the specific numeric ID) as a column
> 
>     * I think some of this infrastructure can be incorporated into the existing `Photometry` handler. Bulk upload of photometry is already implemented there.

@dannygoldstein sure, so I can make one of either instrument ID or instrument name required columns. how does that sound? see my other comment about having a separate delete bulk endpoint

skyportal/models.py
dannygoldstein: 😅 misunderstanding related to the issue of `/upload_photometry` being different from the api endpoint

dannygoldstein:
> > Main comments are:
> > ```
> > * It would be useful to be able to upload data from multiple instruments in a single CSV, with some kind of instrument identifier (if not the specific numeric ID) as a column
> > 
> > * I think some of this infrastructure can be incorporated into the existing `Photometry` handler. Bulk upload of photometry is already implemented there.
> > ```
> 
> @dannygoldstein sure, so I can make one of either instrument ID or instrument name required columns. how does that sound? see my other comment about having a separate delete bulk endpoint

sounds good. thanks @acrellin

acrellin:
@dannygoldstein will do, thanks for the feedback

profjsb:
> Ready for review.
> 
> Question: which fields should be scalar inputs/selects and which should be required columns of the CSV? Currently `["mjd", "flux", "fluxerr", "zp", "zpsys", "filter"]` are required columns (this will be updated accordingly when the expansion to include mags is merged). Instrument is a select. What other fields should we add as required, either as columns in the CSV or as scalar inputs?

Take a look at the API spect for the `Photometry` upload. I believe that ra, dec, ra_unc, dec_unc may be required (but can be null).

acrellin:
> > Ready for review.
> > Question: which fields should be scalar inputs/selects and which should be required columns of the CSV? Currently `["mjd", "flux", "fluxerr", "zp", "zpsys", "filter"]` are required columns (this will be updated accordingly when the expansion to include mags is merged). Instrument is a select. What other fields should we add as required, either as columns in the CSV or as scalar inputs?
> 
> Take a look at the API spec for the `Photometry` upload. I believe that `ra`, `dec`, `ra_unc`, `dec_unc` may be required (but can be null).

These are not required fields (nullable)

acrellin:
@profjsb @dannygoldstein  See below screenshots regarding your suggested additions: 

Sample CSV placeholder text and listed required columns above textarea: 

upload_phot_helper_text

Allowing a user to select an instrument or select "multiple" in which case `instrument_id` becomes a required column, and show IDs next to names in instrument select so they have those at hand:

upload_phot_inst_select

acrellin:
@profjsb @dannygoldstein all of your suggestions/requests have been implemented. Ready for further review.

acrellin:
@dannygoldstein regarding your request, a user must now either select an instrument or an option whose text is "Multiple (requires instrument_id column below)". Also, each instrument's ID is displayed next to its name in that dropdown, so they have that information easily available

acrellin:
Thanks for the feedback, Danny. Regarding the flux/mag select, that strikes
me as a bit overkill just for changing the placeholder text to give an
alternate example, especially with the help text above describing the
different required columns in each respective case. The validation function
already auto-detects whether flux or mag is being provided.

acrellin:
@dannygoldstein what do you mean about splitting up the export button into
two?

acrellin:
@dannygoldstein here's the current layout. I actually like the required fields text above the textarea, but hear that you don't. @profjsb can you break the tie? 
upload_phot

dannygoldstein:
> @dannygoldstein here's the current layout. I actually like the required fields text above the textarea, but hear that you don't. @profjsb can you break the tie?
> upload_phot

Hi @acrellin, looking much better. I think that as is, though, the users could benefit a bit from a bit more of an explanation of how to use this important form. Some immediate questions that spring to mind that a new user might have looking at this page are:

1. Can I upload in both flux and mag space simulatneously, or just one of the two?
2. What is the `altdata` field and how do I format it?

acrellin:
@dannygoldstein I've added a help icon with a help tooltip that appears on mouseover/click. See below screenshots:

help_icon
help_icon_tooltip

acrellin:
@profjsb @dannygoldstein Your latest round of feedback has been addressed. Ready for further review.

static/js/components/UploadPhotometry.jsx
```diff
@@ -0,0 +1,382 @@
...
+                interactive
+                title={(
+                  <>
+                    <p>
+                      Use this form to upload flux- or mag-space photometry data
+                      (only one type per request, not mixed).
+                    </p>
+                    <p>
+                      Required fields (flux-space):&nbsp;
+                      <code>mjd,flux,fluxerr,zp,magsys,filter[,instrument_id]</code>
```
dannygoldstein, line 269: Maybe indicate that instrument id is only required when someone selects Multiple instruments from the drop down on the right hand side. What happens when someone selects `instrument_id = 1` from the drop down and specifies `instrument_id=2` in the CSV upload?

static/js/components/UploadPhotometry.jsx
```diff
@@ -0,0 +1,382 @@
...
+                    </p>
+                    <p>
+                      Other miscellanous metadata can be supplied by preceding the column
+                      name with
+                      {' '}
+                      <code>&quot;altdata.&quot;</code>
+                      {' '}
+                      (e.g.
+                      {' '}
+                      <code>&quot;altdata.metdata1&quot;</code>
```
dannygoldstein, line 292:
```suggestion
                      <code>&quot;altdata.metadata1&quot;</code>
```

static/js/components/UploadPhotometry.jsx
```diff
@@ -0,0 +1,382 @@
...
+                      ).
+                      Such fields will ultimately be stored in the photometry table&apos;s
+                      {' '}
+                      <code>altdata</code>
+                      &nbsp;JSONB column, e.g.
+                      {' '}
+                      <code>
+                        {"{"}
+                        &quot;metadata1&quot;: &quot;value&quot;, ...
```
dannygoldstein, line 301: Can we make this more concrete with an actual column that would be used in altdata, so people have a better sense of what to put here?

static/js/components/UploadPhotometry.jsx
acrellin: Done

static/js/components/UploadPhotometry.jsx
acrellin: Can you provide a realistic example?

static/js/components/UploadPhotometry.jsx
acrellin: Done. If an instrument is selected in the select input, and the user provides an instrument_id column, it shows a validation error message: "Invalid input: instrument_id already specified in select input"

static/js/components/UploadPhotometry.jsx
dannygoldstein: `{'calibrated_to': 'ps1', 'aperture_diameter_pixels':  6}`

static/js/components/UploadPhotometry.jsx
acrellin: (& if "Multitple (provide instrument_id in CSV)" is selected, and they don't provide that column, that also shows a validation error. Likewise if they haven't chosen an option for that select. That covers every case)

acrellin:
> Hi @acrellin ,
> 
> Sorry for the late response here. Looking really nice, just a few minor comments inline and one here, any chance of moving the photometry upload button to the bokeh plot?

@dannygoldstein that's not realistic as that is all created by bokeh. Open to other suggestions, though.

dannygoldstein:
> > Hi @acrellin ,
> > Sorry for the late response here. Looking really nice, just a few minor comments inline and one here, any chance of moving the photometry upload button to the bokeh plot?
> 
> @dannygoldstein that's not realistic as that is all created by bokeh. Open to other suggestions, though.

Yeah and that would break when there is no photometry to be rendered. Hm. OK, fine to leave it as is for now. Nice work @acrellin !

static/js/components/UploadPhotometry.jsx
acrellin: Done.

acrellin:
Thanks @dannygoldstein . All comments addressed.  If we get additional, similar feedback about the button placement on the source page, we can move it later.

acrellin:
Self-merging (I'm currently doing work that with crash head-on with these changes, and want to resolve conflicts sooner than later) as of the two reviewers, there is one approval, and all of the other reviewer's comments/requests have been addressed.

skyportal/handlers/api/photometry.py
```diff
@@ -119,22 +121,41 @@ def post(self):
                               items:
                                 type: integer
                               description: List of new photometry IDs
+                            bulk_upload_id:
```
stefanv, line 124: For simplicity, why not always return a bulk-upload-id?  Otherwise the client has to add a special check.

static/js/components/UploadPhotometry.jsx
```diff
@@ -0,0 +1,385 @@
...
+    setCsvData({});
+    setSuccessMessage(null);
+    formState = getValues();
+    if (!formState.csvData) {
+      return "Missing CSV data";
+    }
+    const delim = new RegExp(formState.delimiter);
+    let [header, ...dataRows] = formState.csvData.trim().split("\n");
+    header = header.split(delim);
+    dataRows = dataRows.map((row) => row.split(delim));
```
stefanv, line 60: [CSV files](https://tools.ietf.org/html/rfc4180) may contain commas, so we need a slightly more sophisticated splitting strategy.

Here's a [code snippet](https://gist.github.com/plbowers/7560ae793613ee839151624182133159) that we can use more-or-less as-is.

stefanv: (If we want to do it beautifully, we can even generate a parser using a PEG specification: http://jsfiddle.net/knvzk/10).  We'd only generate the parser once and then re-use it, not generate it on the fly for each user.

PR #478, opened 2020-05-29
Fix jump to page input out of range bug & update tests

acrellin:


acrellin:
Minor bug fix -- self-merging so that other PR tests can pass

PR #492, opened 2020-06-02
fix a few bugs in the photometry handler

dannygoldstein:
- Handle null mags / magerrs in api/sources/photometry
- Output correct mag systems in `serialize` 
- Handle invalid source id in SourcePhotometryHandler
- Add tests for these cases


dannygoldstein:
This should be merged to fix some important bugs

Issue #466, opened 2020-05-26
Default save behavior


dannygoldstein:
@mansikasliwal @acrellin @profjsb 

We are working on the scanning page and want to confirm the default save button behavior. 

What we are currently envisioning is the following:

<img width="468" alt="Screen Shot 2020-05-26 at 7 56 22 PM" src="https://user-images.githubusercontent.com/2769632/82961317-fc92be00-9f8a-11ea-8161-b5dee0b9300d.png">

What do you think?

acrellin:
material-ui doesn't provide split buttons like this, but bootstrap does (https://react-bootstrap.github.io/components/dropdowns/#split-button-dropdowns). @stefanv how opposed to incorporating a bootstrap button are you?

acrellin:
(deleted previous comment - it can indeed be done with material-ui)

acrellin:
Currently, there is a multi-action button (identical in behavior to github's PR merge button) where the default action is to save to the group(s) for which the candidate passed a filter:

![](https://user-images.githubusercontent.com/7230285/83804343-d512b400-a662-11ea-9914-9ec9bdc262c6.png)

Clicking the caret shows another option called "Select groups & save":

![](https://user-images.githubusercontent.com/7230285/83804413-f07dbf00-a662-11ea-9512-2f442b797b0f.png)

When this action is selected & the button is clicked, the user is shown a dialog with a list of checkboxes for all of their groups, with those groups for which the candidate passed a filter checked by default:

![](https://user-images.githubusercontent.com/7230285/83667187-df5d8100-a582-11ea-86ec-243df3643239.png)

acrellin:
Closing per our discussion where it was determined that needing to change the default target groups would be an infrequent occurrence

PR #493, opened 2020-06-03
Add split button for saving candidates; quick save & select groups

acrellin:
This PR introduces a split button for saving candidates as sources. The default save functionality is to save the source to those of the user's groups for which the candidate passed a filter; the manual group select option allows a user to manually select the groups to which the source will be saved (those groups for which the candidate passed a filter are checked by default). See screenshots below.

Screenshot from 2020-06-03 10-11-56

Screenshot from 2020-06-03 10-12-22

Screenshot from 2020-06-03 10-13-04

Screenshot from 2020-06-03 10-12-51


profjsb:
This looks great. I’d be slightly in favor of having the split dropdown show all the groups with a checkbox next to them so as to avoid the modal pop-up (simpler UX). 

You might double check that the split box works well in a mobile setting (could just be another PR if not).

acrellin:
@profjsb Danny was saying many some would potentially belong to hundreds of groups, which I think would become too cumbersome with a dropdown attached to the button

acrellin:
@dannygoldstein ready for review

static/js/components/SaveCandidateButton.jsx
```diff
@@ -0,0 +1,210 @@
...
+import Popper from "@material-ui/core/Popper";
+import MenuItem from "@material-ui/core/MenuItem";
+import MenuList from "@material-ui/core/MenuList";
+import { useForm, Controller } from "react-hook-form";
+
+import * as sourceActions from "../ducks/source";
+import FormValidationError from "./FormValidationError";
+
+
+const options = ["Save as Source", "Save (select groups)"];
```
dannygoldstein, line 24: For the first option, should say which group it will save to

dannygoldstein: For the second option, the modal dialog should come up as soon as someone selects save to groups from the dropdown, rather than having to click the button after selecting the dropdown

acrellin: That's not how this kind of button works - think the github merge button. That behavior would require a major refactor. This seems to be a common enough type of button group, right? Especially for folks familiar with github pull requests.

acrellin: What you're asking for would best be provided by two separate buttons next to each other, if you feel strongly enough about not wanting to click the button type from the drop down and then click the button

acrellin: The drop down is just to change the button's action, not for actually clicking the button

acrellin: @dannygoldstein regarding your first comment, the button text will wrap to two lines if we add the group name to the button text. How about just a tooltip describing the quick save action?

acrellin: I thought of a way to implement your suggested dropdown button behavior, but still don't think that's desirable behavior from a UX perspective. Let's let others weigh in, though.

stefanv:
There are two component types that may come in handy here:

https://material-ui.com/components/transfer-list/  --- for selecting the groups

and

https://material-ui.com/components/chips/

Another option would be to unify the interfaces: whenever the save button is clicked, pop up a dialog that shows, using chips, all groups it will be saved to.  Users can then remove groups as necessary.

@dannygoldstein Is the "hundreds of groups" scenario feasible?  In that case, the last approach described above may be too cumbersome, in which case we'll have to resort to, e.g., a transfer list.

static/js/components/SaveCandidateButton.jsx
```diff
@@ -0,0 +1,210 @@
...
+
+  const { handleSubmit, errors, reset, control, getValues } = useForm();
+
+  useEffect(() => {
+    reset({
+      group_ids: userGroups.map((userGroup) => candidate.passing_group_ids.includes(userGroup.id))
+    });
+  }, [reset, userGroups, candidate]);
+
+  const handleClickOpenDialog = () => {
```
stefanv, line 40: We're starting to see numerous dialog boxes, so would it make sense to farm a dialog box out into a separate component that only knows how to open and close itself?

static/js/components/SaveCandidateButton.jsx
```diff
@@ -0,0 +1,210 @@
...
+    const formState = getValues({ nest: true });
+    return formState.group_ids.filter((value) => Boolean(value)).length >= 1;
+  };
+
+  const onSubmit = async (data) => {
+    data.id = candidate.id;
+    const groupIDs = userGroups.map((g) => g.id);
+    const selectedGroupIDs = groupIDs.filter((ID, idx) => data.group_ids[idx]);
+    data.group_ids = selectedGroupIDs;
+    const result = await dispatch(sourceActions.saveSource(data));
```
stefanv, line 58: Should we disable save button until action has been completed?

Should a message be displayed next to the button on error, or OK to rely on notification?

static/js/components/SaveCandidateButton.jsx
```diff
@@ -0,0 +1,210 @@
...
+    if (result.status === "success") {
```
stefanv, line 59: Or error!  Otherwise you don't get to try again.

acrellin:
> A comment in general (aimed mainly at those who request features): please consider the complexity of what certain "nice to haves" may introduce in the UI. E.g., compare the new group selector to the old one; many, many more lines of code, even though it does essentially the same thing.
> 
> To save ourselves future maintenance burden, who should try and utilize the simplest solution that we can for any given problem. Skinning it nicely is one thing, but, in my mind, having complex flow between modal dialog boxes, e.g., does not warrant the added complexity.

Agreed, thanks for mentioning this, Stéfan.

static/js/components/SaveCandidateButton.jsx
acrellin: Reset on error? If they spent time carefully crafting their group list, wouldn't we want to preserve the form state so they can attempt to click submit again?

acrellin:
@stefanv by default, those groups for which the candidate passed a filter are already checked. The user can then check additional groups, or uncheck some of the default groups.

static/js/components/SaveCandidateButton.jsx
acrellin: @stefanv actually, the element whose `onClick` triggers the opening of the dialog needs to be defined within the dialog (because it needs to call the dialog's `handleOpen`), which is why I haven't factored this out already. Can you think of a (simple) way to do that?

acrellin: After reading about this exact use case in the offical React docs (https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs), they recommend handling the open/close logic outside of the dialog (if the dialog is a child component) and passing `isOpen` as a prop to the dialog component. So we could factor out the `<Dialog><DialogContent>` part, but that doesn't spare much code, as the logic needs to be implemented in each instance.

acrellin: @stefanv can we move this discussion to an issue and resolve for now? This doesn't seem like an obvious should-do considering the above, and I don't want to drag out the review process unnecessarily.

static/js/components/SaveCandidateButton.jsx
acrellin: As for your second comment, let's rely on the notification -- that's exactly what they're for! :)

acrellin: Done

acrellin:
@dannygoldstein regarding your comment about showing default target group name in save button: done! see below

image

& I renamed the other option: 

image

static/js/components/SaveCandidateButton.jsx
acrellin: @dannygoldstein I've updated https://github.com/skyportal/skyportal/issues/466 and request that we continue the conversation there and change it later if deemed necessary, resolving this for now.

static/js/components/SaveCandidateButton.jsx
stefanv: Certainly; thanks for looking into this!

static/js/components/SaveCandidateButton.jsx
stefanv: Yes, my concern was that the user would be focusing on the button, and looking at the form for feedback.  But, I agree that the notifications are prominent enough not to miss.

static/js/components/SaveCandidateButton.jsx
stefanv: Please ignore; I misread.

stefanv:
> > There are two component types that may come in handy here:
> > https://material-ui.com/components/transfer-list/ --- for selecting the groups
> > and
> > https://material-ui.com/components/chips/
> 
> @stefanv sure, we can use chips instead of regular checkboxes, if you'd like. But I don't see how the transfer list helps -- having to move all groups to one column seems like a lot of unnecessary work, and more cumbersome than simple checkboxes or chips

The list has a button for transferring all items, so it's a one-click job.  The chips may work, except it may make it hard to add new groups to the list.  Anything is fine, really.

stefanv:
> @stefanv by default, those groups for which the candidate passed a filter are already checked. The user can then check additional groups, or uncheck some of the default groups.

That's a good default.

stefanv:
I'm going to merge, since all the functionality is there.  Let's iterate on design if others want to change the button behavior.

PR #497, opened 2020-06-05
Implement photometry-level permission checking

acrellin:
This PR introduces photometry-level permissions checking. Each `Photometry` point has its own `groups` relationship, and the intersection between a user's group list and a photometry point's group list determines access. 

- The photometry POST API endpoint handler now requires a `group_ids` field;
- `Obj.get_if_owned_by` now first checks whether an associated candidate or source is part of the given user's group, and if not, then checks whether the user has access to any of the object's photometry, raising an access error if not;
- `Obj.get_photometry_owned_by_user` is added;
- `photometry_plot` has been updated to only render those photometry points that should be visible to the current user;
- `SourcePhotometryHandler` now returns only those photometry points a user should have access to
- A (required) groups multi-select input is added to the upload source photometry page (gif below)

group_select_chips

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -208,11 +208,9 @@ class PhotometryHandler(BaseHandler):
     def get(self, photometry_id):
         # The full docstring/API spec is below as an f-string
 
-        phot = Photometry.query.get(photometry_id)
+        phot = Photometry.get_if_owned_by(photometry_id, self.current_user)
         if phot is None:
             return self.error('Invalid photometry ID')
-        # Ensure user/token has access to parent source
-        _ = Source.get_if_owned_by(phot.obj_id, self.current_user)
 
         # get the desired output format
         format = self.get_query_argument('format', 'mag')
@@ -248,9 +246,7 @@ class PhotometryHandler(BaseHandler):
               application/json:
                 schema: Error
         """
-        # Ensure user/token has access to parent source
-        s = Source.get_if_owned_by(Photometry.query.get(photometry_id).obj_id,
-                                   self.current_user)
+        _ = Photometry.get_if_owned_by(photometry_id, self.current_user)
         packet = self.get_json()
         group_ids = packet.pop("group_ids", None)
 
@@ -295,9 +291,7 @@ class PhotometryHandler(BaseHandler):
               application/json:
                 schema: Error
         """
-        # Ensure user/token has access to parent source
-        s = Source.get_if_owned_by(Photometry.query.get(photometry_id).obj_id,
-                                   self.current_user)
+        _ = Photometry.get_if_owned_by(photometry_id, self.current_user)
         DBSession.query(Photometry).filter(Photometry.id == int(photometry_id)).delete()
         DBSession().commit()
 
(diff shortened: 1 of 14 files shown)


acrellin:
[ch912]

acrellin:
[ch986]

skyportal/handlers/api/photometry.py
```diff
@@ -181,6 +186,7 @@ def post(self):
 
             phot.original_user_data = packet
             phot.bulk_upload_id = bulk_upload_id
```
profjsb, line 188: It looks like `bulk_upload_id` can be `None` based on the logic above. This will make the id non-unique and so deleting by this id will cause the removal of unrelated photometry. Perhaps you could fallback to hash of the datetime string with a random value appended?

skyportal/handlers/api/photometry.py
```diff
@@ -261,6 +264,8 @@ def put(self, photometry_id):
 
         phot.original_user_data = packet
         phot.id = photometry_id
+        if group_ids is not None:
```
profjsb, line 267: When we every want to upload photometry if there are no `group_ids` associated with it?

acrellin:
@profjsb regarding your first point, you just click the group name again and it is then de-selected

skyportal/handlers/api/photometry.py
acrellin: This is the `put` handler, not the `post` handler.

skyportal/handlers/api/photometry.py
acrellin: I'm changing this to `upload_id` and it will always be non-null

acrellin:
@profjsb all of your requested changes have been made. Please re-review. 

Replies to your comments inline below:

> Looks great and is an important move towards the access control granularity we need! A few things:
> 
>     1. It's unclear from the UI demo whether someone can delete a group tag/id once it's been selected. Is there an "x" to remove a group if it's added by accident?

See comment above - clicking the group name again de-selects it (just like a regular multi-select input field).

>     2. It doesn't look like you're currently testing a variety of scenarios around groups. It would be good to add tests for:
>        
>        * when photometry is added with more than one group that a user belongs to (should pass)

Done.

>        * when photometry is added with at least one group that a user does not belong to. This shouldn't be allowed through the front end but could easily happen with the API (should fail)

Done.

>        * when no groups are added (?) -- document what should happen in docstring?

Done (test added). This is documented as a required field -- the request returns with an error code and message saying that this required field is missing. 

>        * when someone else tries to retrieve photometry (uploaded by a different test user) and belongs to a subset of that photometry's group identifications (should pass)

Done.

>        * when someone else tries to retrieve photometry (uploaded by a different test user) who does not belong to any groups of that photometry (should fail or perhaps return an empty list?)

Done.

acrellin:
Single failing test on Travis (`test_download_comment_attachment`) is totally unrelated. This has been happening a lot, will open a new ticket & investigate. All tests related to this PR are passing.

acrellin:
@stefanv it's just the single, completely unrelated [`test_download_comment_attachment`](https://travis-ci.org/github/skyportal/skyportal/builds/695762333#L2984) that's failing on Travis (I'll address that soon outside of this PR)

skyportal/handlers/api/photometry.py
```diff
@@ -262,7 +271,20 @@ def put(self, photometry_id):
...
+            groups = Group.query.filter(Group.id.in_(group_ids)).all()
+            if not groups:
+                return self.error("Invalid group_ids field. "
+                                  "Specify at least one valid group ID.")
+            if "Super admin" not in [r.id for r in self.associated_user_object.roles]:
+                if not all([group in self.current_user.groups for group in groups]):
+                    return self.error("Cannot upload photometry to groups you "
+                                      "are not a member of.")
+            photometry.groups = groups
+            DBSession().commit()
```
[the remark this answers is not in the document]
acrellin: @stefanv yes, agreed -- I was uncomfortable with that, too. Fixed! (just took a fresh look to realize a `flush()` call would do the trick). Will merge after tests pass.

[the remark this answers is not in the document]
acrellin: (updating the scalar fields with `schema.load` and updating the relationships has to happen in separate transactions, for some reason)

acrellin:
@profjsb @stefanv thank you for the quick reviews and for your helpful feedback/suggestions!

PR #498, opened 2020-06-08
Robustify selenium tests

acrellin:


skyportal/handlers/api/comment.py
```diff
@@ -214,4 +214,5 @@ def get(self, comment_id):
         self.set_header(
             "Content-Disposition", "attachment; "
             f"filename={comment.attachment_name}")
+        self.set_header("Content-type", "application/octet-stream")
```
acrellin, line 217: The omission of this line is what was causing the download comment attachment test to fail

PR #500, opened 2020-06-09
Use Vega Lite to render candidate photometry.

stefanv:
Our current Bokeh plotting is quite slow and requires significant
browser-side resources for a page containing many plots.

Vega Lite is fast and requires much fewer resources.


package.json
```diff
@@ -36,6 +36,9 @@
     "redux-thunk": "2.3.0",
     "semver": "7.3.2",
     "style-loader": "1.2.1",
+    "vega": "5.13.0",
```
acrellin, line 39: Are both `vega` and `vega-lite` necessary here? I noticed the bundle size jumps to ~12MB after adding these.

static/js/components/CandidateList.jsx
```diff
@@ -112,9 +111,8 @@ const CandidateList = () => {
                     <br />
                   </td>
                   <td>
-                    <Plot
-                      className={sourceStyles.smallPlot}
-                      url={`/api/internal/plot/photometry/${candidateObj.id}?plotHeight=200&plotWidth=300`}
+                    <VegaPlot
+                      dataUrl={`/api/sources/${candidateObj.id}/photometry`}
```
acrellin, line 115: Awesome that this just plugs right into an existing API endpoint so seamlessly

package.json
stefanv: Yes, it looks like `vega-lite` does a bunch of imports from `vega`.  However, tree shaking will get rid of what we don't use, so if you set `mode: 'production'` instead of `mode: 'development'` in `webpack.config.js`, the bundle size goes down to 2.8MB.

stefanv:
@acrellin I've now enabled code splitting in webpack, which means we will generate more javascript bundles, but they won't all be loaded at once (there's a small baselayer tweak necessary as well to make this work smoothly, which I can push if you're happy with the approach).

I enabled source maps again, so that we can easily find out where things went wrong when we have JS errors.  That does increase the bundle size quite a bit, but this is only for development time.

I modified the Docker file to strip out the maps and to "shake the tree", i.e., include only necessary JS.  This brings the bundle size down to somewhere around 2MB.  Not ideal, but not crazy either.

You'll see here that I also use `lazy` and `Suspense`, along with a dynamic import, to delay rendering of the component and to bring in the JS module as necessary.  The more dynamic imports we use, the more webpack will be able to split our code chunks.

acrellin:
Merging for now, and we can add additional plot functionality down the road.

PR #501, opened 2020-06-10
Add inline starlist display

stefanv:
Peek 2020-06-11 00-46


dmitryduev:
Once the starlist is displayed, should the button be changed to "HIDE STARLIST" (with the appropriate behavior)?

stefanv:
Good suggestion, @dmitryduev: done.

stefanv:
This is mainly to help people who have the finding chart get an easy to copy-paste copy of the list, otherwise they have to parse the JSON themselves, since the PDF does not allow copy.

Happy to have this moved to wherever.

acrellin:
Merging for now, and we can move later if/when relevant.

PR #499, opened 2020-06-09
Add sidebar

acrellin:
This PR adds a collapsible sidebar with links to the app's main pages.

The footer is removed, and SkyPortal info is moved to an About page, with links in the sidebar and profile dropdown.

image

PR #502, opened 2020-06-11
Styled components

joelachance:
Hello all,

1. Adding styled-components. I think you'll find this library easy to use and will clean up some things with styling and webpack.

2. Switches `Main.jsx` to use re-exported components. I've found this pattern useful on other projects. I've usually stuck with named exports instead of defaults, but defaults work just fine here. I think you'll find improved readability when glancing through a files' imports this way-- you know exactly where your imports are coming from.

I've broken a couple of eslint rules with this PR-- namely arrow functions w/o brackets, and the component imports in Main.jsx. These are usually OK, so curious people's thoughts. Last note on linting: Prettier is used heavily and would likely make everyone's life a touch easier. JS Standard is cool too. These use ESLInt under the hood IIRC, and just have really good defaults.


static/js/components/Logo.jsx
```diff
@@ -1,13 +1,34 @@
...
+    transform: rotate(0deg);
+  }
+`;
+
+const StyledLogo = styled.img`
+  vertical-align: middle;
+  height: 50px;
+  margin-top: -0.5em;
+  animation-name: ${props => props.rotate ? rotate : null};
+  animation-duration: ${props => props.rotate ? '4s' : null};
```
joelachance, line 23: This seems to work as expected (when `props.rotate` is undefined, these rules don't render). This is probably some kind of styled-components anti-pattern though.

static/js/components/Logo.jsx
```diff
@@ -1,13 +1,34 @@
...
+`;
 
 const Logo = () => {
-  const rotateLogo = useSelector((state) => state.logo.rotateLogo);
+  const rotateLogo = useSelector(state => state.logo.rotateLogo);
```
joelachance, line 27: the parameter parens is a really small thing (either way is 100% fine), this is just a personal preference as this feels easier to read.

static/js/components/Main.jsx
```diff
@@ -0,0 +1,151 @@
...
+  Groups,
+  Group,
+  Profile,
+  SourceList,
+  UserInfo,
+  Theme,
+  ProfileDropdown,
+  Logo,
+  Footer,
+} from '../components';
```
joelachance, line 32: ESLint doesn't like the `../components` import, but I find this a lot easier to read. You'll know exactly what imports are components vs. business logic.

stefanv:
@joelachance So good to see you here!  Thank you very much for taking a look at the styling issue.

There's some refactoring happening with the main component and sidebar in #499, just for reference.

But, I wanted to ask: together with material-ui, we've drifted into their recommended way of styling: https://github.com/skyportal/skyportal/pull/499/files#diff-b28c5ab547fbd8d190898d186d5dc23cR48

How do you think that approach differs from styled-components.  Are they different enough that you would still prefer the latter?

One SkyPortal specific: the `main.jsx` file is a template rendered from `main.jsx.template`, so changes will need to go there.

W.r.t. the re-exported modules, I see no problem with that; in fact, it reads well!  Is there an easy way to auto-generate the `index.js` file? We have a templating mechanism, but the config file is fed into it.  E.g., that is what we use to automatically generate the mapping from JavaScript component to URL:

- https://github.com/skyportal/skyportal/blob/master/config.yaml.defaults#L13
- https://github.com/skyportal/skyportal/blob/master/static/js/components/Main.jsx.template

Thanks again for taking the time to take a look at SkyPortal; your review is much appreciated.

stefanv:
W.r.t. automated code formatting, I've been close a couple of times to installing black for Python, but in the end I think there are still cases where personal judgement results in nicer looking code.  So, in this sense I would prefer something like StandardJS over PrettierJS.

I just looked at the [StandardJS](https://standardjs.com/rules.html), and to a large extent it's quite similar to what we have right now.  We've been using the AirBNB style, with a tiny handful of exceptions (we're trying to keep that number down low):

https://github.com/skyportal/skyportal/blob/master/.eslintrc.yaml#L20

Joe, in your opinion, are the StandardJS rules superior to AirBNB?

joelachance:
@stefanv 
> How do you think that approach differs from styled-components. Are they different enough that you would still prefer the latter?

I think it depends on what we're using MaterialUI for. When I've used it in the past, I haven't used it for _everything_, so it's likely we'll still need styled-components or css modules. If that's the case, I do think styled-components is the way to go-- I find it a bit easier to read. When I've worked with MUI in the past, I think my preference was to lean a bit more heavily on styled-components, but of course any direction as great, as long as the developers are happy!

> One SkyPortal specific: the main.jsx file is a template rendered from main.jsx.template, so changes will need to go there.

Good to know! Is that why `Main.jsx` is in gitignore?

> Is there an easy way to auto-generate the index.js file?

Probably, but I'm unsure about this one. I've made it a point to simply add it to the index out of habit, but something automated would be super nice.

What is the benefit of having both a `Main.jsx` and a `Main.jsx.template`? It looks like there's some HTML templating going on in the `.template`, but what's the bonus for having both, vs. rendering the React component?

> Joe, in your opinion, are the StandardJS rules superior to AirBNB?

Good question :) I think even if you like AirBnb, which has a great and widely used set of rules, there's some benefit to not having to deal with configuring linters and just use JS Standard. I think they say it well here: https://standardjs.com/#i-disagree-with-rule-x-can-you-change-it

So, to answer your question, yes, I think I prefer a linting library like JS Standard. However, every project or job I've been at has had conflicting linting standards, so I tend to just roll with the punches. I'm not sure there's a right or a wrong way (as you know), but my thought is that given this project is run by at the least _some_ volunteer contributors, the easier we can make it, the better.

stefanv:
@joelachance I'd like to move this forward.  Let's stick to styling in this PR, and discuss linting elsewhere, since our eslint configuration is reasonable for now.

Could you perhaps look into how styled components play with Material UI, and perhaps show how one of the currently styled MUI components should be done when using this pattern?

joelachance:
@stefanv yep, of course!

I'm still not clear on the purpose of the Main.jsx template, do we want to style that component and not the Main.jsx?
None of the other components have templates from what I understand, so I think we're good to go on those.

I'll pull the rest of this out into another PR, I think it'll still be good to have.

stefanv:
Main.jsx is compiled from Main.jsx.template before each launch, so that file (the template) is the definitive version. We need the templating to auto populate some components from our config file.

joelachance:
Closing this PR-- I plan to pull these into separate PRs that are a bit smaller and focused.

PR #513, opened 2020-06-23
Catch invalid thumbnail file type exception in POST handler

acrellin:


acrellin:
@stefanv there was another uncaught exception that this PR addresses.

acrellin:
Simple addendum to previous PR -- merging

PR #514, opened 2020-06-24
Implement simple status monitor

stefanv:
This adds a very simple status monitor that can be run on a different host from SkyPortal.  It probes services once in a set interval, and writes out a static html report, that can be served whichever way preferred.  The monitor is simple, because it should be easily deployable, should run with minimal support, and should be easy to maintain/run to ensure maximal uptime.

Currently served at `/static/status.html`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ package-lock.json
 .jobs_timestamps.yaml
 skyportal_image_cache
 static/js/components/Main.jsx
+static/status.html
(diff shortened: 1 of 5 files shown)


profjsb:
I believe that we are going to get a 3rd party service for this (e.g. statuspage.io). cc @dannygoldstein

stefanv:
Closed in favor of using [Atlassian's statuspage](https://www.atlassian.com/software/statuspage).

PR #517, opened 2020-06-26
Photometry API performance improvement

dannygoldstein:
This PR implements some performance improvements to  `PhotometryHandler.post`:

   * Uses `sqlalchemy.core` instead of ORM for inserts
   * Uses `executemany_mode='VALUES'` (see cesium-ml/baselayer#106)
   * Implements `Obj.last_detected` as a hybrid property

The net effect of these changes is to speed up POSTs by ~50x.


skyportal/tests/api/test_photometry.py
```diff
@@ -869,3 +865,26 @@ def test_token_user_retrieve_null_photometry(upload_data_token, public_source,
...
+                           'limiting_mag': 22.3,
+                           'magsys': 'ab',
+                           'filter': 'ztfg',
+                           'group_ids': [public_group.id]
+                       },
+                       token=upload_data_token)
+    stop = time.time()
+    assert status == 200
+    assert data['status'] == 'success'
+    assert stop - start < 25
```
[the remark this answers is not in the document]
dannygoldstein: This is about 5x what it should take on a reasonable system  (5s)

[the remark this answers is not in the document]
acrellin: So this should still pass on a Raspberry Pi?

dannygoldstein:
I have an implementation I am happy with here now based on Pre-fetching PKs...

skyportal/tests/api/test_photometry.py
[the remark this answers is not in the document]
dannygoldstein: @acrellin it might.. but good point. will remove this

dannygoldstein:
this PR depends on cesium-ml/baselayer#106

dannygoldstein:
This is no longer a WIP. Ready for review. There are some travis failures, and slow travis performance, but this should be fixed by merging cesium-ml/baselayer#106 which enables the query to take advantage of batch mode. @acrellin One of the travis failures is the same as the one happening on #519 for deleting comments.

skyportal/tests/api/test_photometry.py
```diff
@@ -869,3 +865,25 @@ def test_token_user_retrieve_null_photometry(upload_data_token, public_source,
     assert data['status'] == 'success'
     assert data['data']['mag'] is None
     assert data['data']['magerr'] is None
+
+
+def test_token_user_big_post(upload_data_token, public_source, ztf_camera,
+                             public_group):
+
+    start = time.time()
```
profjsb, line 873: remove.
```suggestion
```

skyportal/tests/api/test_photometry.py
```diff
@@ -1,9 +1,5 @@
-import os
-import datetime
-import base64
+import time
```
profjsb, line 1: remove
```suggestion
```

skyportal/tests/api/test_photometry.py
```diff
@@ -869,3 +865,25 @@ def test_token_user_retrieve_null_photometry(upload_data_token, public_source,
...
+                           'instrument_id': ztf_camera.id,
+                           'mag': np.random.uniform(low=18, high=22, size=50000).tolist(),
+                           'magerr': np.random.uniform(low=0.1, high=0.3, size=50000).tolist(),
+                           'limiting_mag': 22.3,
+                           'magsys': 'ab',
+                           'filter': 'ztfg',
+                           'group_ids': [public_group.id]
+                       },
+                       token=upload_data_token)
+    stop = time.time()
```
profjsb, line 887: remove
```suggestion
```

skyportal/handlers/api/photometry.py
```diff
@@ -171,46 +192,172 @@ def post(self):
...
+        for oid in df['obj_id'].unique():
+            obj = Obj.query.get(oid)
+            if not obj:
+                return self.error(f'Invalid object ID: {oid}')
+
+        # pre-fetch the photometry PKs. these are not guaranteed to be
+        # gapless (e.g., 1, 2, 3, 4, 5, ...) but they are guaranteed
+        # to be unique in the table and thus can be used to "reserve"
+        # PK slots for uninserted rows
+
```
profjsb, line 303:
```suggestion
```

skyportal/handlers/api/photometry.py
```diff
@@ -1,18 +1,21 @@
 import uuid
 import numpy as np
 import arrow
-from astropy.time import Time
+import time as t
```
profjsb, line 4: Not using `time` (after debugging complete)
```suggestion
```

skyportal/handlers/api/photometry.py
```diff
@@ -138,6 +140,24 @@ def post(self):
 
         if "altdata" in data and not data["altdata"]:
             del data["altdata"]
+
+        # quick validation - just to make sure things have the right fields
+        try:
```
dmitryduev, line 145: what happens if someone tries to post both fluxes and mags (by mistake)?

dmitryduev: I guess the fluxes are just ignored then

skyportal/handlers/api/photometry.py
```diff
@@ -182,46 +202,170 @@ def post(self):
...
+        # PK slots for uninserted rows
+        pkq = f"SELECT nextval('photometry_id_seq') FROM " \
+              f"generate_series(1, {len(df)})"
+
+        proxy = DBSession().execute(pkq)
+
+        # cache this as list for response
+        ids = [i[0] for i in proxy]
+        df['id'] = ids
+        rows = df.where(pd.notnull(df), None).to_dict('records')
```
dmitryduev, line 321: Do you really need to convert df to a dict? Using `df.loc[whatever].itertuples()` on L:324 should be much faster

skyportal/models.py
```diff
@@ -159,6 +161,17 @@ class Obj(Base):
 
     followup_requests = relationship('FollowupRequest', back_populates='obj')
 
+    @hybrid_property
+    def last_detected(self):
```
dmitryduev, line 165: More of a note to myself: currently, K relies on `last_detected` to figure out which prv_candidates to post to SP if an alert passes a filter and a candidate with the objectId already exists in SP. Follow-up photometry posted to the candidate will obviously break this. I'm making K track which objectId's passed which filters when (for the last time) and so try to post stuff that came later.

skyportal/models.py
```diff
@@ -453,6 +466,9 @@ class Photometry(Base):
     ra_unc = sa.Column(sa.Float, doc="Uncertainty of ra position [arcsec]")
     dec_unc = sa.Column(sa.Float, doc="Uncertainty of dec position [arcsec]")
 
+    detected = sa.Column(sa.Boolean, default=True, nullable=False,
```
dmitryduev, line 469: So anything with a `candid` in ZTF speak?

skyportal/models.py
```diff
@@ -510,6 +528,20 @@ def e_mag(cls):
             else_=None
         )
 
+    @hybrid_property
+    def jd(self):
+        return self.mjd + 2_400_000.5
```
dmitryduev, line 533: Ooh, I learned something new :joy: Must have missed PEP 515  )

skyportal/tests/api/test_photometry.py
```diff
@@ -740,68 +735,6 @@ def test_token_user_update_photometry_groups(upload_data_token_two_groups,
     assert "Insufficient permissions" in data["message"]
 
 
-def test_post_same_alertid_update_groups(upload_data_token_two_groups,
```
dmitryduev, line 743: I see your logic in wanting to pull this out of here and move to an AlertHandler -- agreed, let's do that. The same handler will then deal with filter id's, duplicate alert_id's etc. 
K would then be like "hey, AlertHandler, alert `candid` for `objectId` passed filter `filter_id` of program `program_id` (or whatever we're calling it, will refresh my memory in the coming days on this for sure :) and here are all new photometry points for that `objectId` since it passed that filter last time". 
Need to think a bit on the details not to duplicate too much data posted in api calls, but let's come back to it a bit later.

skyportal/handlers/api/photometry.py
dannygoldstein: It will fail with "unknown column"

dannygoldstein: In both cases (unknown column Mag if attempted to parse as flux, unknown column flux if attempted to parse as mag)

skyportal/models.py
dannygoldstein: Yeah. This may not be the best way to implement this. But currently I am leaving it to the user to define if something counts as a detection, rather than enforcing specific criteria. See convo in #531

skyportal/tests/api/test_photometry.py
dannygoldstein: Awesome glad you agree with this. @profjsb expressed similar views on the standup this AM

skyportal/handlers/api/photometry.py
dannygoldstein: Need dicts for L:358

skyportal/models.py
dannygoldstein: Yeah thats a good plan. It is correct that Non-alert photometry will break this

skyportal/models.py
dannygoldstein: I decided to define `detected` as 5-sigma

skyportal/handlers/api/photometry.py
dmitryduev: Sure, but you can iterate `for row in df.where(pd.notnull(df), None).itertuples()` and assemble `phot` using the named tuple `row`, e.g. L:330 would be `flux = row.standardized_flux`. It's usually much faster, but I don't think we'll notice the difference in most cases.

skyportal/tests/api/test_photometry.py
dmitryduev: Note, I would argue that AlertHandler is generic enough and should still be part of SP and only a F-specific extension.

dannygoldstein: Yes, I think `AlertHandler` would be useful for LSST

dannygoldstein:
Awesome !! Thanks @acrellin

PR #522, opened 2020-06-29
PR #2 Classification Epic: refactor data loader

profjsb:
This PR introduces a new Makefile command:
```
make load_seed_data
```
This will add tables, permissions/ACL, users, groups, telescopes, instruments, and taxonomies to the DB. It is meant to be run as part of a new SkyPortal instance after `make db_init` (when testing this out you should do a `make db_clear` beforehand). The intent is to seed the empty skyportal DB with site-wide data. Everyone logged into the Skyportal instance should be able to see/access this data via their membership in the `Sitewide Group`.

`make load_seed_data` is different than `make load_demo_data` in that the latter is meant to populate the DB with example sources, objects, candidates, etc. to facilitate testing and demoing of the basic functionality of SkyPortal. But `make load_seed_data` is meant to persist data for a production-level instance of Skyportal.

This refactor uses a common codebase for both `load_seed_data` and `load_demo_data`

TODO after this PR is merged:
  - New users should be automatically (by default) added to the `Sitewide Group` (if it exists) so that they can see the Sidewide taxonomy.


skyportal/models.py
```diff
@@ -359,6 +359,52 @@ class Instrument(Base):
                         default=[])
 
 
+class Taxonomy(Base):
+    __tablename__ = 'taxonomy'
```
dannygoldstein, line 363:
```suggestion
    __tablename__ = 'taxonomies'
```

skyportal/models.py
```diff
@@ -359,6 +359,52 @@ class Instrument(Base):
...
+                          )
+    provenance = sa.Column(sa.String, nullable=True,
+                           doc='Identifier (e.g., URL or git hash) that '
+                               'uniquely ties this taxonomy back '
+                               'to an origin or place of record'
+                           )
+    version = sa.Column(sa.String, nullable=False,
+                        doc='Semantic version of this taxonomy'
+                        )
+    isLatest = sa.Column(sa.Boolean, default=True, nullable=False,
```
dannygoldstein, line 381: What happens if multiple rows in the `taxonomies` table have this flagged?

Makefile
```diff
@@ -45,6 +45,11 @@ docs: ## Build the SkyPortal docs
 docs: | doc_reqs api-docs
 	export SPHINXOPTS=-W; make -C doc html
 
+db_seed: ## Import example dataset
```
dannygoldstein, line 48:
```suggestion
db_seed: ## Import included dataset of instruments, telescopes, etc. 
```

tools/db_seed.py
```diff
@@ -0,0 +1,138 @@
...
+
+from baselayer.app.env import load_env
+from baselayer.app.model_util import status, create_tables
+from skyportal.models import Base, init_db, DBSession, User
+from skyportal.model_util import setup_permissions, create_token
+from skyportal.tests import api
+from baselayer.tools.test_frontend import verify_server_availability
+
+
+if __name__ == "__main__":
```
dannygoldstein, line 18: The logic here is alsmot identical to that in `make load_demo_data`. Might make sense to factor out the common code into a single place, have `load_demo_data.py` and `db_seed.py` import what is needed?

skyportal/models.py
profjsb: It shouldn't happen since all older taxonomies with that name have this flagged changed upon ingestion. Still, the logic of what to show in case there's a conflict will be handled by the classification apis.

profjsb:
> My main comment is that the logic in `db_seed.py` is almost identical to that in make load_demo_data. Might make sense to factor out the common code into a single place, have load_demo_data.py and db_seed.py import what is needed?

This is now fixed. We use the same codebase, but different config files for the two commands.

skyportal/models.py
```diff
@@ -314,10 +314,10 @@ class SourceView(Base):
 class Telescope(Base):
     name = sa.Column(sa.String, nullable=False)
     nickname = sa.Column(sa.String, nullable=False)
-    lat = sa.Column(sa.Float, nullable=False)
-    lon = sa.Column(sa.Float, nullable=False)
-    elevation = sa.Column(sa.Float, nullable=False)
-    diameter = sa.Column(sa.Float, nullable=False)
+    lat = sa.Column(sa.Float, nullable=True)
+    lon = sa.Column(sa.Float, nullable=True)
+    elevation = sa.Column(sa.Float, nullable=True)
```
dannygoldstein, line 319: ^^ These three must not be null in order to do observability calculations. 

Diameter being null is fine

skyportal/models.py
```diff
@@ -362,12 +362,12 @@ class Instrument(Base):
     spectra = relationship('Spectrum', back_populates='instrument')
 
     # can be [] if an instrument is spec only
-    filters = sa.Column(ArrayOfEnum(allowed_bandpasses), nullable=False,
+    filters = sa.Column(ArrayOfEnum(allowed_bandpasses), nullable=True,
```
dannygoldstein, line 365: This needs to be nonnull for other handlers (gets set to [] if null)

configs/db_demo.yaml
```diff
@@ -0,0 +1,155 @@
...
+  - name: Nordic Optical Telescope
+    nickname: NOT
+    lat: 28.75
+    lon: 17.88
+    elevation: 2000
+    diameter: 2.56
+    skycam: http://catserver.ing.iac.es/weather/archive/concam/concam_labels.png
+    group_names:
+      - Program A
+    token: "load_demo_data token"
```
acrellin, line 68: Why is there a token field in the telescope entries?

skyportal/models.py
profjsb: I had to do this since you put in HST!

skyportal/models.py
profjsb: There are examples in the data you sent where there were no filters.

skyportal/models.py
dannygoldstein: Good point... OK. I will account for this in an upcoming observability PR

skyportal/models.py
dannygoldstein: Yeah, there they get set to the default (`[]`)

configs/db_demo.yaml
profjsb: It's the named token to use to POST the data. In principle multiple tokens can be established and different ones can be used throughout the data import process.

skyportal/models.py
profjsb: Ok, got it. I need to fix some logic in the code.

dannygoldstein:
@profjsb I think the travis failures may be related to taxonomy ACLs

profjsb:
> @profjsb I think the travis failures may be related to taxonomy ACLs

Should be fixed now. Tests are passing.

tools/data_loader.py
```diff
@@ -0,0 +1,377 @@
...
+                        for i in ins_src:
+                            if i.get("token") is not None:
+                                token_name = i["token"]
+                            else:
+                                token_name = ins_token
+
+                            if i.get("telescope_nickname") is not None:
+                                telname = i["telescope_nickname"]
+                            else:
+                                telname = i.get("telescope_id")
```
dannygoldstein, line 239: What if neither of these keys is present?

dannygoldstein: Perhaps we should be using `jsonschema` to validate the `instruments.yaml` and `telescopes.yaml` config files?

profjsb: Good idea. Make this an issue and we'll do this in another PR.

profjsb: > What if neither of these keys is present?

There's not much error checking in `data_loader.py` on purpose -- I think an error getting raised is the right response: 
   - These scripts are not for use during live, unsupervised production. Instead they are use at the start of a SkyPortal project.
   - The yaml files are unlikely to change much from deployment to deployment

dannygoldstein: Yes, but people may extend or modify the files before running `make load_seed_data`. In that case they may neglect to add some important keys such as `telescope_nickname` or similar, in which case the join against instruments will fail

profjsb: In principle we should have schemata for all YAMLs in the SkyPortal. Not sure where to draw the line here.

acrellin:
@profjsb merge conflicts

configs/db_demo.yaml
```diff
@@ -0,0 +1,155 @@
...
+    type: both
+    band: V
+    telescope_nickname: NOT
+    token: "load_demo_data token"
+sources:
+  - id: 14gqr
+    candidate:
+      candidate_filters:
+        - "example filter"
+    unsaved_candidate_copy:
```
acrellin, line 86: This is not currently working. They're showing up as saved sources

acrellin: @profjsb these are also missing thumbnails

acrellin: @profjsb actually 14gqr and 14gqr_unsaved_copy have thumbnails but 16fil* do not

profjsb: these are fixed now.

skyportal/tests/conftest.py
```diff
@@ -237,17 +237,27 @@ def comment_token(user):
...
 @pytest.fixture()
 def taxonomy_token(user):
     token_id = create_token(
         permissions=["Post Taxonomy", "Delete Taxonomy"],
         created_by_id=user.id, name=str(uuid.uuid4())
     )
+
+
+@pytest.fixture()
+def comment_token_two_groups(user_two_groups):
```
acrellin, line 250: Duplicated definition?

skyportal/tests/conftest.py
```diff
@@ -237,17 +237,27 @@ def comment_token(user):
...
+    token_id = create_token(
+        permissions=["Comment"], created_by_id=user_two_groups.id, name=str(uuid.uuid4())
+    )
     return token_id
 
+
 @pytest.fixture()
 def comment_token_two_groups(user_two_groups):
```
acrellin, line 258: Duplicate

package.json
```diff
@@ -11,7 +11,7 @@
     "@material-ui/icons": "4.9.1",
     "@material-ui/pickers": "3.2.10",
     "bokehjs": "0.12.9",
-    "check-dependencies": "1.1.0",
+    "check-dependencies": "^1.1.0",
```
acrellin, line 14:
```suggestion
    "check-dependencies": "1.1.0",
```

skyportal/models.py
acrellin: Should this be reverted to `nullable=False`, then?

tools/data_loader.py
```diff
@@ -0,0 +1,378 @@
...
+            users_dict = {u.username: u.id for u in users}
+
+        with status("Creating tokens"):
+
+            tokens = {}
+            for u in src.get('users', []):
+                for t in u.get("tokens", []):
+                    tokens[t["name"]] = create_token(
+                        t["permissions"],
+                        users_dict[u["username"]],
```
acrellin, line 84: Isn't the username different from the ID?

package.json
profjsb: Done

skyportal/tests/conftest.py
profjsb: Done.

skyportal/models.py
profjsb: Keeping this as `nullable=False`, given the default value.

tools/data_loader.py
```diff
@@ -0,0 +1,378 @@
...
+                        data = assert_post(
+                            "taxonomy",
+                            payload,
+                            tokens[tax["token"]]
+                        )
+            if src.get("telescopes") is not None:
+                with status("Creating Telescopes"):
+                    telescope_dict = {}
+                    tel_src = src.get("telescopes")
+                    if isinstance(tel_src, dict) and \
```
acrellin, line 182: Should we be catching the case where this is `False` and `group_names`, `token`, etc below are `None`? Currently it'll raise `NameError` but perhaps we should have a more descriptive error message, e.g. "This value must be specified either here or here" (where we describe the possible configurations)

tools/data_loader.py
```diff
@@ -0,0 +1,378 @@
...
+        try:
+            verify_server_availability(server_url)
+            print("App running - continuing with API calls")
+
+            if src.get("groups") is not None:
+                with status("Creating groups"):
+                    group_ids = []
+                    group_dict = {}
+                    filter_dict = {}
+                    for g in src.get('groups', []):
```
acrellin, line 121: The preponderance of single-character variable names makes this file much harder to read for me. I'd request more descriptive variable names (e.g. `group` instead of `g`) throughout this file. Otherwise, when jumping in to edit/debug, one has to hunt down the assignment statement to have any idea of what the variable is, whereas longer, more descriptive names are pretty obvious and easier to read.

tools/data_loader.py
acrellin: Same for other data types below where relevant (instruments, etc)

configs/db_demo.yaml
```diff
@@ -0,0 +1,159 @@
...
```
acrellin, line 86: Is this general enough to hard-code like this?

tools/data_loader.py
```diff
@@ -0,0 +1,378 @@
...
+                with status("Loading Source and Candidates"):
+                    (basedir / "static/thumbnails").mkdir(parents=True, exist_ok=True)
+                    for s in src.get("sources"):
+                        sinfo = [{"id": s["id"], "ra": s["ra"],
+                                  "dec": s["dec"], "save_source": True,
+                                  "redshift": s.get("redshift", 0.0),
+                                  "altdata": s.get("altdata", None),
+                                  "cand_filts": s["candidate"]["candidate_filters"],
+                                  "comments": s.get("comments", [])}]
+                        if s.get("unsaved_candidate_copy") is not None:
```
acrellin, line 264: Is this general enough to hard-code like this?

tools/data_loader.py
dannygoldstein: General comment (not a request from @profjsb) -- for refactoring PyCharm is great, it constructs an abstract syntax tree for your code and identifies all the uses of a variable throughout the module, then lets you rename them all at once

tools/data_loader.py
profjsb: yes

configs/db_demo.yaml
profjsb: yes

tools/data_loader.py
profjsb: It's fine for this to throw errors -- this is a one-time run script (at the beginning of a SkyPortal instance) and not production code. There's another issue for doing YAML schema checking which will help with malformed YAMLs.

acrellin:
This looks great, thanks a lot @profjsb !

PR #524, opened 2020-06-29
Add field type for magErr tooltip

stefanv:


stefanv:
Unrelated failures.

static/js/components/VegaPlot.jsx
```diff
@@ -59,7 +59,7 @@ const spec = (url) => ({
           type: "nominal"
         },
         tooltip: [
-          { field: "magAndErr", title: "mag" },
+          { field: "magAndErr", title: "mag", type: "nominal" },
```
dmitryduev, line 62: I see that the type is actually required, was it the reason why it was throwing errors?

PR #529, opened 2020-06-30
Provision admin/public tokens on each run of SkyPortal (if they don't exist)

stefanv:


stefanv:
@acrellin I could use your help on this one!

stefanv:
Perhaps we don't even need two tokens: the admin token can also upload public data?

acrellin:
@stefanv is the public token for users to actually use, or just for populating the DB? if the latter, then yeah we don't need it

stefanv:
@acrellin Now I generate only one token.

skyportal/model_util.py
```diff
@@ -56,6 +88,10 @@ def setup_permissions():
 
 
 def create_token(permissions, created_by_id, name):
+    """
+    permissions --- ACLs
```
acrellin, line 92: we can update the function signature to be clearer throughout

acrellin:
@stefanv what does this PR need? Anything I can help with?

stefanv:
@acrellin I gave it a shot; is this what you meant?

stefanv:
@acrellin I think it's good to go; I just wanted to get your eyes on it, in case I'm doing something crazy with the tokens!

skyportal/model_util.py
```diff
@@ -13,37 +13,69 @@
...
+    admin = add_user('admin@cesium-ml.org', roles=['Super admin'])
+    token_name = 'Initial admin token'
+
+    token = (
+        Token.query
+        .filter(Token.created_by == admin)
+        .filter(Token.name == token_name)
+    ).first()
+
+    if not token:
```
acrellin, line 56: Minor nitpick, but `if token is None`

skyportal/model_util.py
```diff
@@ -13,37 +13,69 @@
...
-        super_user.roles.append(admin_role)
-    DBSession().add(super_user)
-    DBSession().commit()
+    add_user(username, roles=['Super admin'], auth=True)
+
+
+def provision_token():
+    """Provision an initial administrative token.
+    """
+    admin = add_user('admin@cesium-ml.org', roles=['Super admin'])
```
acrellin, line 47: I think "testadmin" is what's used in the new yaml-based seed user

acrellin: Or is this not meant to map onto that?

stefanv: It can be anything, I guess---it just has to be a user with enough permissions.

stefanv:
@acrellin I *think* this change is unrelated; can you confirm?

PR #547, opened 2020-07-01
Use PapaParse for parsing CSV strings

acrellin:
This PR improves CSV parsing in the photometry upload front-end component. We use `PapaParse` for parsing CSV, which auto-detects the delimiter, simplifying our form by not requiring the delimiter to be specified. This also allows for escaped (quoted) values that contain the delimiter, as in the example screenshot below.

Screenshot from 2020-07-01 15-48-12


acrellin:
[ch1016]

PR #559, opened 2020-07-06
Add airmass and altitude calculation to objects

dannygoldstein:
This PR adds a new function to `Obj` that calculates the airmass of the object at a specified telescope and time (or array of times). 

`astroplan` is used to do the calculation, with `Telescope.lat`, `Telescope.lon`, and `Telescope.elevation` used to set the observer.


skyportal/models.py
```diff
@@ -196,6 +211,33 @@ def desi_dr8_url(self):
...
+        secz: astropy.units.Quantity
+           The airmass of the Obj at the requested times
+        """
+
+        target = astroplan.FixedTarget(name=self.id, coord=self.skycoord)
+        observer = astroplan.Observer(latitude=telescope.lat * u.deg,
+                                      longitude=telescope.lon * u.deg,
+                                      elevation=telescope.elevation * u.m)
+
+        return observer.altaz(time, target).secz
```
profjsb, line 239: Airmass near the horizon isn't well represented by secz (ie. we dont actually live in a plane-parallel atmosphere). Perhaps use Pickering's interpolation? (https://en.wikipedia.org/wiki/Air_mass_(astronomy)#Interpolative_formulas)

skyportal/tests/tools/test_sky.py
```diff
@@ -0,0 +1,26 @@
+import numpy as np
+from astropy.time import Time
+
+
+def test_airmass(public_source, ztf_camera):
+    telescope = ztf_camera.telescope
+    times = Time(np.linspace(58000, 59000), format='mjd')
+    airmass_calc = public_source.airmass(telescope, times).value
+    airmass_true = np.asarray([ -6.98036718, -23.34945066,   4.33830158,
```
profjsb, line 9: What does negative mean? We should have some `below_the_horizon` value.

skyportal/models.py
```diff
@@ -27,6 +32,18 @@
 PHOT_ZP = 23.9
 PHOT_SYS = 'ab'
 
+env, cfg = load_env()
```
acrellin, line 35: These changes exist in the spatial backend PR -- should be separated out

skyportal/tests/tools/test_sky.py
dannygoldstein: Negative means below the horizon. For below the horizon, how about `np.inf`?

skyportal/models.py
dannygoldstein: This PR is based on that one

skyportal/models.py
dannygoldstein: done

skyportal/tests/tools/test_sky.py
dannygoldstein: Implemented as np.inf in ec23e7d

dannygoldstein:
[ch822]

dannygoldstein:
@acrellin @profjsb Ready for final review

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
         return (f"http://legacysurvey.org/viewer/jpeg-cutout?ra={self.ra}"
                 f"&dec={self.dec}&size=200&layer=dr8&pixscale=0.262&bands=grz")
 
+    def airmass(self, telescope, time, below_horizon=np.inf):
```
stefanv, line 211: below_horizon undocumented

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+        """Return the airmass of the Obj at time `time` from Telescope
+        `telescope`.
+
+        Parameters
+        ----------
+
```
stefanv, line 217:
```suggestion
```

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+        telescope: skyportal.models.Telescope
+            The telescope to use for the airmass calculation
+
```
stefanv, line 220:
```suggestion
```

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+        time: astropy.time.Time
+            The time or times at which to calculate the airmass
+
+        Returns
+        -------
+
```
stefanv, line 226:
```suggestion
```

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
         return (f"http://legacysurvey.org/viewer/jpeg-cutout?ra={self.ra}"
                 f"&dec={self.dec}&size=200&layer=dr8&pixscale=0.262&bands=grz")
 
+    def airmass(self, telescope, time, below_horizon=np.inf):
+        """Return the airmass of the Obj at time `time` from Telescope
```
stefanv, line 212: We typically use a sentence without any parameters here, e.g.

"Return the airmass of an object at a given time."

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+        `telescope`.
+
+        Parameters
+        ----------
+
+        telescope: skyportal.models.Telescope
```
stefanv, line 218:
```suggestion
        telescope : skyportal.models.Telescope
```

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+            The telescope to use for the airmass calculation
+
+        time: astropy.time.Time
```
stefanv, line 221:
```suggestion
        time : astropy.time.Time
```

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+            The time or times at which to calculate the airmass
```
stefanv, line 222: If this can be times (plural), then the type needs to be astropy.time.Time or list of Time

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+
+        Returns
+        -------
+
+        secz: ndarray in the shape of `time`
```
stefanv, line 227:
```suggestion
        secz : ndarray in the shape of `time`
```

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+           The airmass of the Obj at the requested times
+        """
+
+        output_shape = np.shape(time)
+        time = np.atleast_1d(time)
+        altitude = self.altitude(telescope, time).to('degree').value
+
+        # use Pickering (2002) interpolation to calculate the airmass
+
+        with np.errstate(invalid='ignore'):
```
stefanv, line 237: Why this ignore?

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+            sinarg = altitude + 244 / (165 + 47 * altitude ** 1.1)
+            airmass = 1. / np.sin(np.deg2rad(sinarg))
```
stefanv, line 239: How is airmass defined as the denomenator approaches zero?

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+
+        # set objects below the horizon to an airmass of infinity
+        airmass[altitude < 0] = below_horizon
+        airmass = airmass.reshape(output_shape)
+
+        return airmass
+
+    def altitude(self, telescope, time):
+        """Return the altitude of the Obj at time `time` from Telescope
```
stefanv, line 248: Same formatting comments as above.

skyportal/models.py
```diff
@@ -206,6 +208,71 @@ def desi_dr8_url(self):
...
+        telescope: skyportal.models.Telescope
+            The telescope to use for the altitude calculation
+
+        time: astropy.time.Time
+            The time or times at which to calculate the altitude
+
+        Returns
+        -------
+
+        alt: astropy.units.Quantity
```
stefanv, line 263: Can you confirm that this is the type of the object being returned?

skyportal/tests/tools/test_sky.py
```diff
@@ -0,0 +1,71 @@
+import numpy as np
+from astropy.time import Time
+
+
+def test_airmass(public_source, ztf_camera):
+    telescope = ztf_camera.telescope
+    times = Time(np.linspace(58000, 59000), format='mjd')
+    airmass_calc = public_source.airmass(telescope, times)
+    airmass_true = np.asarray([np.inf, np.inf, 4.259917450786537,
```
stefanv, line 9: Where do these numbers come from?  Is it possible to only query the airmass at one or two points in time, and to hand-verify the result instead?

skyportal/models.py
```diff
@@ -206,6 +208,69 @@ def desi_dr8_url(self):
...
+
+        telescope: skyportal.models.Telescope
+            The telescope to use for the airmass calculation
+        time: astropy.time.Time
+            The time or times at which to calculate the airmass
+
+        Returns
+        -------
+        secz: ndarray in the shape of `time`
+           The airmass of the Obj at the requested times
```
profjsb, line 226: You are no longer sending `secz` but instead the Pickering value of the Rayleigh (molecular atmosphere) airmass. Just call this variable `airmass`. 
```suggestion
        airmass: ndarray in the shape of `time`
           The airmass of the Obj at the requested times
```
You should document what it means to be below the horizon (numerically). [Looks like the limit of airmass](https://www.wolframalpha.com/input/?i=lim%281%2Fsin%28%28pi%2F180%29*%28a+%2B+244+%2F+%28165+%2B+47*a**1.1%29%29%29%2C+a-%3E0%29) as h->0 is 38.7494. Might want to document that.

skyportal/models.py
```diff
@@ -206,6 +208,69 @@ def desi_dr8_url(self):
...
+        """
+
+        output_shape = np.shape(time)
+        time = np.atleast_1d(time)
+        altitude = self.altitude(telescope, time).to('degree').value
+
+        # use Pickering (2002) interpolation to calculate the airmass
+
```
profjsb, line 234:
```suggestion
```

skyportal/models.py
dannygoldstein: It tends toward ~39. I added this to the documentation

skyportal/models.py
dannygoldstein: Because below the horizon, the altitude takes on a negative value, and numpy's power ufunc causes the negative values to be turned into `nan`s. I rewrote this chunk to avoid this behavior and not rely on `errstate`.

skyportal/tests/tools/test_sky.py
dannygoldstein: Redid this using solid numbers from an independent calculation (http://www.briancasey.org/artifacts/astro/airmass.cgi?observatory=p&obslong=&obslat=&actual=&effective=&site=&zone=&zonename=&zoneabbrev=&daylight=1&year=2020&month=07&day=21&starname=pubsource&rightasc=00+00+00&declin=00+00+00&epoch=2000&altitude=-18&command=display)

dannygoldstein:
[ch822]

dannygoldstein:
@acrellin @stefanv Requesting re-review

PR #561, opened 2020-07-07
Add individual user groups & public group that all users automatically belong to

acrellin:
This PR adds individual user groups, which are automatically created whenever new users are added, to facilitate the sharing of data between individual users (PR forthcoming). This PR also adds a public group to which all users are automatically added (upon user creation).

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -94,9 +94,8 @@ class GroupHandler(BaseHandler):
                                                              False)
         info = {}
         info['user_groups'] = list(self.current_user.groups)
-        info['all_groups'] = (list(Group.query) if hasattr(self.current_user, 'roles')
-                              and 'Super admin' in
-                              [role.id for role in self.current_user.roles]
+        info['all_groups'] = (list(Group.query) if 'Super admin' in
+                              [role.id for role in self.associated_user_object.roles]
                               else None)
         if (not include_single_user_groups) or (include_single_user_groups == "false"):
             info["user_groups"] = [g for g in info["user_groups"]
@@ -272,6 +271,7 @@ class GroupUserHandler(BaseHandler):
                               description: Boolean indicating whether user is group admin
         """
         data = self.get_json()
+        admin = data.pop("admin", False)
         group = Group.query.get(group_id)
         if group.single_user_group:
             return self.error("Cannot add users to single user groups.")
@@ -282,8 +282,7 @@ class GroupUserHandler(BaseHandler):
         gu = (GroupUser.query.filter(GroupUser.group_id == group_id)
                        .filter(GroupUser.user_id == user_id).first())
         if gu is None:
-            gu = GroupUser(group_id=group_id, user_id=user_id)
-        gu.admin = data['admin']
+            gu = GroupUser(group_id=group_id, user_id=user_id, admin=admin)
         DBSession().add(gu)
         DBSession().commit()
 
@@ -316,7 +315,7 @@ class GroupUserHandler(BaseHandler):
         """
         group = Group.query.get(group_id)
         if group.single_user_group:
-            return self.error("Cannot add users to single user groups.")
+            return self.error("Cannot delete users from single user groups.")
         user_id = User.query.filter(User.username == username).first().id
         (GroupUser.query.filter(GroupUser.group_id == group_id)
          .filter(GroupUser.user_id == user_id).delete())
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -98,7 +98,7 @@ class GroupHandler(BaseHandler):
                               and 'Super admin' in
                               [role.id for role in self.current_user.roles]
                               else None)
-        if not include_single_user_groups:
+        if (not include_single_user_groups) or (include_single_user_groups == "false"):
             info["user_groups"] = [g for g in info["user_groups"]
                                    if not g.single_user_group]
             if info["all_groups"]:
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -90,14 +90,20 @@ class GroupHandler(BaseHandler):
                 return self.success(data=group)
             return self.error(f"Could not load group with ID {group_id}")
 
+        include_single_user_groups = self.get_query_argument("includeSingleUserGroups",
+                                                             False)
         info = {}
-        info['user_groups'] = [g for g in list(self.current_user.groups)
-                               if not g.single_user_group]
-        info['all_groups'] = ([g for g in list(Group.query) if not g.single_user_group]
-                              if hasattr(self.current_user, 'roles')
+        info['user_groups'] = list(self.current_user.groups)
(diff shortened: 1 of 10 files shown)


acrellin:
[ch997]

acrellin:
[ch1024]

tools/load_demo_data.py
```diff
@@ -38,6 +38,8 @@
         setup_permissions()
 
     with status(f"Creating dummy users"):
+        DBSession().add(Group(name="Public Group"))
```
dannygoldstein, line 41: See comment above

skyportal/models.py
```diff
@@ -74,12 +75,25 @@ class Group(Base):
     photometry = relationship("Photometry", secondary="group_photometry",
                               back_populates="groups",
                               cascade="save-update, merge, refresh-expire, expunge")
+    single_user_group = sa.Column(sa.Boolean, default=False)
```
dannygoldstein, line 78: I think this should be excluded from the marshmallow schema

acrellin: @dannygoldstein this is something we will likely use on the front-end, thus it's an expected part of the API call return values and should be left in the schema.

skyportal/models.py
```diff
@@ -74,12 +75,25 @@ class Group(Base):
...
 
 
 GroupUser = join_model('group_users', Group, User)
 GroupUser.admin = sa.Column(sa.Boolean, nullable=False, default=False)
 
 
+@event.listens_for(User, "after_insert")
+def create_user_group_and_add_to_public_group(mapper, connection, target):
+    object_session(target).add(
```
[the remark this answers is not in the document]
acrellin: https://docs.sqlalchemy.org/en/13/orm/session_api.html?highlight=object_session#sqlalchemy.orm.session.object_session 
See import statement above

skyportal/handlers/api/group.py
```diff
@@ -90,12 +90,20 @@ def get(self, group_id=None):
...
         info['all_groups'] = (list(Group.query) if hasattr(self.current_user, 'roles')
                               and 'Super admin' in
                               [role.id for role in self.current_user.roles]
                               else None)
+        if (not include_single_user_groups) or (include_single_user_groups == "false"):
+            info["user_groups"] = [g for g in info["user_groups"]
+                                   if not g.single_user_group]
+            if info["all_groups"]:
+                info["all_groups"] = [g for g in info["all_groups"]
+                                      if not g.single_user_group]
```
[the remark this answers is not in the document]
acrellin: @dannygoldstein there's no way for a user to create single-user groups, as they have no way off accessing that field (we don't load arbitrary json data into a schema in that handler).

skyportal/models.py
```diff
@@ -74,12 +75,25 @@ class Group(Base):
...
+        Group(name=target.username, users=[target], single_user_group=True)
+    )
+    public_group = Group.query.filter(Group.name == "Public Group").first()
```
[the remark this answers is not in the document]
acrellin: @dannygoldstein Done.

tools/load_demo_data.py
acrellin: @dannygoldstein Done.

skyportal/models.py
```diff
@@ -74,12 +75,25 @@ class Group(Base):
...
 
 
 GroupUser = join_model('group_users', Group, User)
 GroupUser.admin = sa.Column(sa.Boolean, nullable=False, default=False)
 
 
+@event.listens_for(User, "after_insert")
+def create_user_group_and_add_to_public_group(mapper, connection, target):
+    object_session(target).add(
+        Group(name=target.username, users=[target], single_user_group=True)
```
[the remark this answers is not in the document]
acrellin: I don't think it matters -- they should never need to be changed

skyportal/handlers/api/group.py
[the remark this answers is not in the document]
dannygoldstein: Yes but based on the docstring it will be listed in the API docs as an acceptable parameter of the method due to the schema including GroupNoID

skyportal/models.py
dannygoldstein: I think this is fine to include for GETs but not something we should include in POST

skyportal/handlers/api/group.py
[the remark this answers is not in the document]
acrellin: :+1: will remove from GroupNoID schema

skyportal/handlers/api/user.py
```diff
@@ -27,11 +28,87 @@ def get(self, user_id=None):
...
+                            id:
+                              type: integer
+                              description: New user ID
+        """
+        data = self.get_json()
+        roles = data.get("roles", ["Full user"])
+        group_ids_and_admin = data.get("groupIDsAndAdmin", [])
+
+        # Add user
+        user = User(username=data["username"], role_ids=roles)
```
profjsb, line 93: We should be normalizing the `username` here (and potentially elsewhere) to insure we have all lowercase, no escaped characters, allowed email formats, etc. We certainly dont want usernames like Josh!😂.

A similar checking and normalization could be done for `group.name` upon creation. Ie. only allow alphanumeric characters and underscores.

skyportal/handlers/api/user.py
```diff
@@ -27,11 +28,87 @@ def get(self, user_id=None):
...
+        user = User.query.get(int(user_id))
         if user is None:
             return self.error(f'Invalid user ID ({user_id}).')
         else:
-            return self.success(data=user)
+            user_info = user.to_dict()
+            user_info['acls'] = user.acls
+            return self.success(data=user_info)
+
+    @permissions(["Manage users"])
```
profjsb, line 39: In general, it seems like we should allow group admins to add users to the system. Super admins could add to any group, but we could restrict group admins only to add users to groups that that group admin has admin privileges to. As currently written, and as the default ACL, I worry that adding users to a Skyportal instance will be unnecessarily burdensome to the (few) super admins of a SkyPortal instance.

skyportal/model_util.py
```diff
@@ -14,6 +14,18 @@
                               Token)
 
 
+all_acl_ids = ['Become user', 'Comment', 'Manage users', 'Manage sources',
```
profjsb, line 17: See comment above about `Manage users`

skyportal/handlers/api/user.py
acrellin: Sure, I'll add "Manage users" to the group admin ACLs

skyportal/handlers/api/group.py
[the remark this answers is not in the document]
acrellin: Done

skyportal/models.py
acrellin: Done

skyportal/handlers/api/user.py
acrellin: Done

acrellin:
@dannygoldstein @profjsb all comments have been addressed. Please re-review and merge

acrellin:
This PR contains fixes for two tests that are failing on master due to changes in the baselayer database schema

PR #562, opened 2020-07-07
Update Travis CI build to send test results to BuildPulse

jasonrudolph:
@dannygoldstein: Here's the starting point for the pull request we discussed.

To get everything wired up end-to-end, we'll need to do just a few things:

- [x] In your repository settings for https://travis-ci.org/github/skyportal/skyportal, set the `BUILDPULSE_ACCESS_KEY_ID` and `BUILDPULSE_SECRET_ACCESS_KEY` environment variables [[docs]](https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings)

- [x] Configure pytest to generate an XML report for the test results. (You can see an example commit in https://github.com/Workshop64/buildpulse-example-pytest/commit/2ccd887723dcd48b44cbd18d9a9bb99de7a183f1.)

- [x] Add an `after_script` clause to `.travis.yml` to send the test results to BuildPulse for analysis

    This is done in 71c3b173bcb3a9a0698735b574d4928a8616bc75. It currently assumes that XML report(s) will reside in the `test-results` directory. If it would be helpful to use a different directory, that's totally fine. We'll just need to update the `BUILDPULSE_REPORT_PATH` in `.travis.yml` accordingly.

Please ping me when the first two items are done, and we can make sure that everything is flowing through nicely. And of course, if you run into any trouble with those items or if you have any questions at all, just let me know. :bow:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ package-lock.json
 .jobs_timestamps.yaml
 skyportal_image_cache
 static/js/components/Main.jsx
+test-results/
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,3 +82,10 @@ after_success:
   # ESLint gets called when we try to commit the docs
   - rm .git/hooks/pre-commit
   - .travis/deploy_docs.sh
+
+after_script:
+  # Upload test results to BuildPulse for flaky test detection
+  - export BUILDPULSE_ACCOUNT_ID=25833053
+  - export BUILDPULSE_REPOSITORY_ID=82240075
+  - export BUILDPULSE_REPORT_PATH=test-results
+  - /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Workshop64/buildpulse-travis-ci/main/uploader-for-ubuntu.sh)"
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ matrix:
   include:
     - python: 3.7
       env:
-        - TEST_TARGET=test
+        - TEST_TARGET=test_xml
 
 before_install:
   - ccache -s
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+junit_family=xunit2


dannygoldstein:
Awesome, will do! Many thanks @jasonrudolph

jasonrudolph:
> - [ ] Configure pytest to generate an XML report for the test results. (You can see an example commit in Workshop64/buildpulse-example-pytest@2ccd887.)

@dannygoldstein: You mentioned possibly adding a new test target to the makefile for this, and I think that was motivated by wanting to only generate the XML report on CI and not during local dev. That should work, but if it's too much trouble, then configuring `pytest` to always generate the report should be fairly non-invasive to the local dev experience.

In case it helps, I did a quick check locally, and adding the `--junitxml=test-results/junit.xml` option to `pytest` doesn't remove any of the existing output that you see during local dev. Instead, it just additionally creates the XML file. The only change what you see via `STDOUT` is one additional line like this:

```
--------------- generated xml file: /usr/src/app/test-report.xml ---------------
```

I hope that helps.

dannygoldstein:
@jasonrudolph I implemented the first two points, but it looks like BuildPulse is not finding my `BUILDPULSE_ACCESS_KEY_ID`: 

https://travis-ci.org/github/skyportal/skyportal/builds/706284991#L3345

jasonrudolph:
> I implemented the first two points, but it looks like BuildPulse is not finding my `BUILDPULSE_ACCESS_KEY_ID`:
> 
> https://travis-ci.org/github/skyportal/skyportal/builds/706284991#L3345

Thanks for the ping, @dannygoldstein. I suspect this is due to the fact that this pull request comes from my fork. I see that [Travis CI applies different security when building from a fork](https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings) (which makes sense):

> Variables defined in repository settings are the same for all builds, and when you restart an old build, it uses the latest values. These variables are not automatically available to forks.

With that in mind, I think we'll need to merge this branch into the main skyportal repository in order to resolve this issue.

Are you comfortable merging this to the `master` branch? I suspect it will Just Work™ once merged.

Alternatively, you could create another branch in this repository (perhaps call the branch `buildpulse-integration`), and we can [change this pull request](https://github.blog/2016-08-15-change-the-base-branch-of-a-pull-request/) to target that branch instead of `master.` Then, we could merge this pull request and verify that everything works correctly when Travis CI builds that branch. Assuming all is well, we can then merge that branch into the `master` branch.

What do you think?

dannygoldstein:
@jasonrudolph I'm merging this into `deploy`, then will create a new PR to merge to `master`. Thanks for the recommendations!

dannygoldstein:
@jasonrudolph Does this mean that all future PRs will not be able to send test results to buildpulse as well, as they will originate from forks?

jasonrudolph:
> Does this mean that all future PRs will not be able to send test results to buildpulse as well, as they will originate from forks?

@dannygoldstein: Good question. That sounds right to me based on what I'm seeing in the Travis CI docs. Let's see what we can figure out together to get you the functionality you need while working within the Travis CI security model.

So far, all the teams using BuildPulse are working in private repositories and have workflows where everyone opens branches in the repository (as opposed to using a fork). Given that skyportal/skyportal is open source, I imagine forks will be more common, so I want to find a solution that will work for you all. For the team that's working on skyportal/skyportal full-time, are folks doing their work in forks?

dannygoldstein:
> Given that skyportal/skyportal is open source, I imagine forks will be more common, so I want to find a solution that will work for you all. For the team that's working on skyportal/skyportal full-time, are folks doing their work in forks?

Yes that's right.

jasonrudolph:
>> For the team that's working on skyportal/skyportal full-time, are folks doing their work in forks?
>
> Yes that's right.

Thanks for that insight, @dannygoldstein.

It seems like this lots of folks are bumping up against this Travis CI behavior -- https://github.com/travis-ci/travis-ci/issues/5579#issuecomment-180805112

For the moment, this means that you'll be able to detect flaky tests for commits that live in this repository (e.g., the build that occurs when a pull request is merged to master), but I'm not yet sure how to overcome this limitation regarding forks.

I'm going to reach out to Travis CI and see what options we might have. I'll follow up with you via email tomorrow to share anything I'm able to find out.

stefanv:
I think this is fine; we don't need to detect flaky tests in forks, do we?  We want to know which of the SkyPortal tests are brittle, and user forks won't have additional tests with a long history.

stefanv:
@dannygoldstein You'll need to make a separate PR to baselayer for those changes, but I'm in favor.

PR #565, opened 2020-07-10
Data loader refactor

stefanv:
This PR makes the data loader static—in other words, it can be run from a node other than that on which SkyPortal is running, and only ingests data from YAML files.  The YAML files match the API: each entry corresponds to an API endpoint, and is a list of objects to post.  Lists can be loaded from external YAML files too.  Inside the YAML file, once an object is posted, its attributes can be saved as references.  E.g., a group can be posted as follows:

```
groups:
  - name: Program A
    group_admins:
      - testadmin@cesium-ml.org
    =id: program_A
```

Note that the `id` field of the created group is stored as `program_A`.  This can now be used when, e.g., posting a telescope:

```
telescope:
  - name: Palomar 1.5m
    nickname: P60
    lat: 33.3633675
    lon: -116.8361345
    elevation: 1870
    diameter: 1.5
    skycam_link: http://bianca.palomar.caltech.edu/images/allsky/AllSkyCurrentImage.JPG
    group_ids:
      - =program_A
```

This refactoring is necessary for adding data to SkyPortal as a Kubernetes batch job.  Previously, in order to do data loading, the same exact version of SkyPortal would need to run in the batch job.  This is often not practical, given that SkyPortal can be given by the CI, while data loading may be triggered externally by a developer.


requirements.txt
```diff
@@ -25,3 +25,4 @@ Pillow>=6
 sncosmo>=2.1.0
 tdtax>=0.1.1
 jsonschema
+pytest-rerunfailures>=9.0
```
acrellin: Looks like this could use a rebase?

skyportal/handlers/api/group.py
```diff
@@ -90,12 +90,19 @@ def get(self, group_id=None):
                 return self.success(data=group)
             return self.error(f"Could not load group with ID {group_id}")
```
acrellin: These changes shouldn't be here; looks like a rebase is needed

stefanv: @acrellin I think you reviewed an old version, perhaps.  That's no longer in the diff.

stefanv:
[ch-1121]

skyportal/handlers/api/public_group.py
```diff
@@ -0,0 +1,36 @@
...
+                      - type: object
+                        properties:
+                          id:
+                            type: integer
+                            description: ID of the public group
+        """
+        pg = Group.query.filter(Group.name == cfg['misc.public_group_name']).first()
+        if pg is None:
+            return self.error('Public group does not exist')
+        return self.success(data={"id": pg.id})
```
acrellin, line 36: I think this should return the whole group (so that name, etc is included)

skyportal/handlers/api/spectrum.py
```diff
@@ -38,6 +38,11 @@ def post(self):
         """
         data = self.get_json()
         instrument_id = data.get('instrument_id')
+        if isinstance(instrument_id, list):
```
acrellin, line 41: Wait, I thought this should always be a single inst ID??

tools/data_loader.py
```diff
@@ -1,387 +1,209 @@
...
+
+    if code not in (200, 400):
+        print(f'Error: could not connect to server (HTTP status {code})')
+        sys.exit(-1)
+
+    if data['status'] != 'success':
+        print('Error: Could not authenticate against SkyPortal; please specify a valid token.')
+        sys.exit(-1)
+
+    code, response = get('groups/public')
```
acrellin, line 116: `status`, for consistency?

skyportal/handlers/api/public_group.py
stefanv: This seems to be a pattern elsewhere as well:

```
        return self.success(data={"id": telescope.id})
```

Happy to change it, but what is our standard way of returning the object as a dict?

skyportal/handlers/api/spectrum.py
stefanv: It should; this is just a safe-guard in case someone provides many.

We could take it out, but I imagined sometimes people would split a spreadsheet with "instrument_id" as a field (i.e., in every row), and I don't want to prevent it in case this is valid input.

But, I don't feel strongly about it either way.

skyportal/handlers/api/public_group.py
acrellin:
```suggestion
        return self.success(data=pg})
```

skyportal/handlers/api/spectrum.py
acrellin: Oh, I see, a homogeneous list.. yeah this seems fine.

skyportal/handlers/api/public_group.py
acrellin: @stefanv not for GET requests, but yes sometimes for POST requests we just return the ID

stefanv: Yes, of course, that makes total sense: will update.

acrellin:
@stefanv Front-end photometry plotting seems to be broken on this branch -- we'll need to investigate before merging. To reproduce: `make db_clear && make db_init && make run`, then in another terminal when the app is up and running, `make load_demo_data`, then navigate to `localhost:5000/source/14gqr` in the browser

stefanv:
Permission errors fixed in https://github.com/skyportal/skyportal/pull/645

PR #567, opened 2020-07-13
Use `pytest-rerunfailures` to mark flaky front-end tests to be rerun up to 2 times

acrellin:
This PR adds the `pytest-rerunfailures` dependency, using it to mark flaky tests to be rerun if they fail (up to two times, and this is easily configurable). This should greatly improve the overall reliability of the test suite.

A few small test bugs are also addressed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,3 +25,4 @@ Pillow>=6
 sncosmo>=2.1.0
 tdtax>=0.1.1
 jsonschema
+pytest-rerunfailures>=9.0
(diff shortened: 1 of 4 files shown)


acrellin:
Note that the test suite will not pass until #561 is merged, as there are real bugs that that PR fixes

acrellin:
> Ari, this is a fine workaround for now; but do we have a sense of why these fail? First prize would be to fix whatever causes the problem.

Agreed! This should just help a bit in the meantime.

PR #569, opened 2020-07-14
PR #3 Classification Epic: Classification API

profjsb:
This PR added `classification` API capabilities to the backend. Similar to `Comments` the `Classification` model attached to an `Obj` and is group/user aware. A classification is also associated with one `taxonomy` ID and must be in the `allowed_classes` list of a taxonomy (which is generated when the taxonomy is saved). The classification probability (if given) must be in the range [0,1]. Many classifications can be given for a single `Obj`. It will be up to the front end (PR #4++) to decide how to display such classifications.

This PR:
  - creates the `Classification` model
  - introduces a new ACL called `Classify`
  - provides GET, POST, PUT, DELETE methods
  - contains tests


skyportal/model_util.py
```diff
@@ -79,6 +79,9 @@ def setup_permissions():
     """Create default ACLs/Roles needed by application.
 
     If a given ACL or Role already exists, it will be skipped."""
+    all_acl_ids = ['Become user', 'Comment', 'Manage users', 'Manage sources',
```
acrellin, line 82: @profjsb this is defined as a global above. Also, currently only the super admin can classify unless you add it to the ACLs for other roles above

acrellin:
@profjsb regarding the travis test failures: `create_token` has a new signature (`permissions` -> `ACLs`)

skyportal/handlers/api/classification.py
```diff
@@ -0,0 +1,239 @@
...
+                          type: object
+                          properties:
+                            comment_id:
+                              type: integer
+                              description: New classification ID
+        """
+        data = self.get_json()
+        obj_id = data['obj_id']
+        # Ensure user/token has access to parent source
+        _ = Source.get_obj_if_owned_by(obj_id, self.current_user)
```
dannygoldstein, line 96:
```suggestion
        source = Source.get_obj_if_owned_by(obj_id, self.current_user)
        if source is None:
           return self.error("Invalid source.")
```

skyportal/handlers/api/classification.py
```diff
@@ -0,0 +1,239 @@
...
+        # check the taxonomy
+        taxonomy_id = data["taxonomy_id"]
+        taxonomy = Taxonomy. \
+            get_taxonomy_usable_by_user(taxonomy_id, self.current_user)
+        if taxonomy is None:
+            return self.error(
+                'That taxonomy does not exist or is not available to user.'
+            )
+        if not isinstance(taxonomy, list):
+            return self.error('Problem retreiving taxonomy')
```
dannygoldstein, line 116: Why would this happen?

skyportal/handlers/api/classification.py
```diff
@@ -0,0 +1,239 @@
...
+        _ = Source.get_obj_if_owned_by(obj_id, self.current_user)
+        user_group_ids = [g.id for g in self.current_user.groups]
+        group_ids = data.pop("group_ids", user_group_ids)
+        group_ids = [gid for gid in group_ids if gid in user_group_ids]
+        if not group_ids:
+            return self.error(f"Invalid group IDs field ({group_ids}): "
+                              "You must provide one or more valid group IDs.")
+        groups = Group.query.filter(Group.id.in_(group_ids)).all()
+
+        author = self.associated_user_object.username
```
dannygoldstein, line 105:
```suggestion
        author = self.associated_user_object
```

skyportal/handlers/api/classification.py
```diff
@@ -0,0 +1,239 @@
...
+                               'is not in the allowed classes for the chosen '
+                               f'taxonomy (id={taxonomy_id}')
+
+        probability = data.get('probability')
+        if probability is not None:
+          if probability < 0 or probability > 1:
+            return self.error(f"That probability ({probability}) is outside "
+                               "the allowable range (0-1).")
+
+        classification = Classification(classification=data['classification'],
```
dannygoldstein, line 129: Concretely, what does it mean for a classification to be associated with a group?

skyportal/handlers/api/classification.py
```diff
@@ -0,0 +1,239 @@
...
+            required: true
+            schema:
+              type: integer
+        responses:
+          200:
+            content:
+              application/json:
+                schema: Success
+        """
+        user = self.associated_user_object.username
```
dannygoldstein, line 225:
```suggestion
        user = self.associated_user_object
```

skyportal/handlers/api/classification.py
```diff
@@ -0,0 +1,239 @@
...
+        roles = (self.current_user.roles if hasattr(self.current_user, 'roles') else [])
+        c = Classification.query.get(classification_id)
+        if c is None:
+            return self.error("Invalid classification ID")
+        obj_id = c.obj_id
+        author = c.author
+        if ("Super admin" in [role.id for role in roles]) or (user == author):
```
dannygoldstein, line 232:
```suggestion
        if ("Super admin" in [role.id for role in roles]) or (user.id == author.id):
```

skyportal/models.py
```diff
@@ -463,6 +476,25 @@ class Comment(Base):
...
 
+class Classification(Base):
+    classification = sa.Column(sa.String, nullable=False)
+    taxonomy_id = sa.Column(sa.ForeignKey('taxonomies.id'),
+                            nullable=False)
+    probability = sa.Column(sa.Float,
+                            doc='User-assigned probability of belonging '
+                            'to this class', nullable=True)
+
+    author = sa.Column(sa.String, nullable=False)
```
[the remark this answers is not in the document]
profjsb: This is fine -- but note that this is modeled after `Comments`, which uses author as a string:
```python
author = sa.Column(sa.String, nullable=False)
```

skyportal/models.py
```diff
@@ -463,6 +476,25 @@ class Comment(Base):
 GroupComment = join_model("group_comments", Group, Comment)
 
 
+class Classification(Base):
+    classification = sa.Column(sa.String, nullable=False)
```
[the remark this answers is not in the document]
profjsb: I spent awhile on this idea previously. Turns out we cannot do that since we don't know the enum apriori. The enum depends on the taxonomy loaded. The `Taxonomy` handler now adds an `allowed_classes` list during a new post and the `Classification` handler ensures that the classification is allowed in that taxonomy.

skyportal/handlers/api/classification.py
profjsb: It means that users in that group can see the classification, otherwise they cannot.

skyportal/handlers/api/classification.py
```diff
@@ -0,0 +1,241 @@
+import base64
```
acrellin, line 1:
```suggestion
```

skyportal/handlers/api/classification.py
profjsb: Classifications are competition sensitive, even more than for comments.

skyportal/models.py
```diff
@@ -463,6 +476,25 @@ class Comment(Base):
...
+                            nullable=False)
+    probability = sa.Column(sa.Float,
+                            doc='User-assigned probability of belonging '
+                            'to this class', nullable=True)
+
+    author = sa.Column(sa.String, nullable=False)
+    obj_id = sa.Column(sa.ForeignKey('objs.id', ondelete='CASCADE'),
+                       nullable=False, index=True)
+    obj = relationship('Obj', back_populates='classifications')
+    groups = relationship("Group", secondary="group_classifications",
```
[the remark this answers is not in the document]
profjsb: done.

skyportal/model_util.py
```diff
@@ -15,7 +15,7 @@
 
 all_acl_ids = ['Become user', 'Comment', 'Manage users', 'Manage sources',
                'Manage groups', 'Upload data', 'System admin', 'Post taxonomy',
-               'Delete taxonomy']
+               'Delete taxonomy', 'Classify']
 
 role_acls = {
```
acrellin, line 20: Should add "Classify" to relevant roles here

profjsb: @acrellin is making this an issue. Will address in the next PR.

PR #572, opened 2020-07-15
Auto-complete usernames & create new users via group users form

acrellin:
This PR adds auto-complete functionality to the group users front-end form. If a _new_ username is provided, that user is created, along with the associated individual user group and associated `GroupUser` row for the site-wide public group. Tests are updated.

### DIFF
--- a/skyportal/handlers/api/user.py
+++ b/skyportal/handlers/api/user.py
@@ -1,9 +1,6 @@
 from ..base import BaseHandler
 from baselayer.app.access import permissions
 from ...models import DBSession, User, Group, GroupUser, cfg
-from ...model_util import role_acls
-
-from sqlalchemy.orm import joinedload
 
 
 class UserHandler(BaseHandler):
@@ -11,30 +8,53 @@ class UserHandler(BaseHandler):
     def get(self, user_id=None):
         """
         ---
-        description: Retrieve a user
-        parameters:
-          - in: path
-            name: user_id
-            required: true
-            schema:
-              type: integer
-        responses:
-          200:
-            content:
-              application/json:
-                schema: SingleUser
-          400:
-            content:
-              application/json:
-                schema: Error
+        single:
+          description: Retrieve a user
+          parameters:
+            - in: path
+              name: user_id
+              required: true
+              schema:
+                type: integer
+          responses:
+            200:
+              content:
+                application/json:
+                  schema: SingleUser
+            400:
+              content:
+                application/json:
+                  schema: Error
+        multiple:
+          description: Retrieve all users
+          responses:
+            200:
+              content:
+                application/json:
+                  schema:
+                    allOf:
+                      - $ref: '#/components/schemas/Success'
+                      - type: object
+                        properties:
+                          data:
+                            type: array
+                            items:
+                              $ref: '#/components/schemas/User'
+                            description: List of users
+            400:
+              content:
+                application/json:
+                  schema: Error
         """
-        user = User.query.get(int(user_id))
-        if user is None:
-            return self.error(f'Invalid user ID ({user_id}).')
-        else:
+        if user_id is not None:
+            user = User.query.get(int(user_id))
+            if user is None:
+                return self.error(f'Invalid user ID ({user_id}).')
             user_info = user.to_dict()
             user_info['acls'] = user.acls
             return self.success(data=user_info)
+        users = User.query.all()
+        return self.success(data=users)
 
     @permissions(["Manage users"])
     def post(self):
--- a/skyportal/handlers/api/user.py
+++ b/skyportal/handlers/api/user.py
@@ -3,6 +3,28 @@ from baselayer.app.access import permissions
 from ...models import DBSession, User, Group, GroupUser, cfg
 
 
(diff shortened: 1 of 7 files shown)


acrellin:
Screen capture demos:

auto_suggest_user
group_user_autocomplete_add_new

acrellin:
[ch1027]

acrellin:
[ch1028]

acrellin:
@profjsb @stefanv ready for review

profjsb:
Looking good. A few UX changes:

1) Shouldn't be able to delete all the admins of a group. Probably should disable the remove from group button on the admin if there is only 1.
Jul-17-2020 11-17-15

2) If you want to grant admin privileges to an existing user in the group you have to delete the user then readd them. This should be made more simple (like an "grant admin status" button in the user list for those that are not admin).

3) Trying to add an existing user pops up a cryptic red error message which disappears quickly. Should probably catch that error and through a more descriptive error as a click-to-close modal.

acrellin:
> Looking good. A few UX changes:
> 
>     1. Shouldn't be able to delete all the admins of a group. Probably should disable the remove from group button on the admin if there is only 1.
>        Jul-17-2020 11-17-15
> 
>     2. If you want to grant admin privileges to an existing user in the group you have to delete the user then readd them. This should be made more simple (like an "grant admin status" button in the user list for those that are not admin).
> 
>     3. Trying to add an existing user pops up a cryptic red error message which disappears quickly. Should probably catch that error and through a more descriptive error as a click-to-close modal.

Thanks @profjsb . None of these issues are actually related to the changes in this PR (they're all pre-existing features/bugs ;) ) -- I'll roll them out into issues and address them in another PR soon, so as to keep the scope of this PR focused

static/js/components/NewGroupUserForm.jsx
```diff
@@ -1,52 +1,105 @@
...
+      <Autocomplete
         id="newUserEmail"
         value={formState.newUserEmail}
-        onChange={handleChange}
+        onChange={(event, newValue) => {
+          if (typeof newValue === 'string') {
+            setFormState({
+              newUserEmail: newValue,
+            });
+          } else if (newValue && newValue.inputValue) {
```
stefanv, line 57: what does this else clause handle?

acrellin:
> Great. Please add those UX comments as new issues.

Done. Thanks for catching those, @profjsb

static/js/components/NewGroupUserForm.jsx
```diff
@@ -1,52 +1,105 @@
...
+        selectOnFocus
+        clearOnBlur
+        handleHomeEndKeys
+        options={allUsers}
+        getOptionLabel={(option) => {
+          // Value selected with enter, right from the input
+          if (typeof option === 'string') {
+            return option;
+          }
+          // Add "xxx" option created dynamically
```
stefanv, line 88: What does this do?

skyportal/handlers/api/user.py
```diff
@@ -1,40 +1,82 @@
 from ..base import BaseHandler
 from baselayer.app.access import permissions
 from ...models import DBSession, User, Group, GroupUser, cfg
-from ...model_util import role_acls
 
-from sqlalchemy.orm import joinedload
+
+def add_user_and_setup_groups(username, roles, group_ids_and_admin):
```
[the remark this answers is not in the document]
acrellin: We need to allow users to specify a list of `[group_id, admin]` (`[int, bool]`) pairs in the user API so that one can add a user and add them to the specified groups (some of which they might be admins) in a single request. What would be a better approach here? A list of objects? e.g. `[{"group_id": 1, "admin": false}, ...]`

static/js/components/NewGroupUserForm.jsx
acrellin: As the comment below describes, it handles the case where a new value (not in the options list) is being provided by the user (i.e. they're specifying a _new_ username)

static/js/components/NewGroupUserForm.jsx
acrellin: This adds the visible option that says "Add 'xxx'" where 'xxx' is whatever the user typed in (see screen capture gif to see this in action). This is the standard recipe provided by mui

skyportal/handlers/api/user.py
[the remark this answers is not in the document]
stefanv: Ah, I see!  No problem, maybe just clarify that in the docstring.

acrellin:
> This looks great!
> 
> Are you planning any tests, @acrellin?

We already have tests for this component, which I've updated accordingly. I'll add another test for creating new users

static/js/components/NewGroupUserForm.jsx
stefanv: Ah, ok, I didn't know what the XXX meant---maybe just clarify that.

acrellin:
> > This looks great!
> > Are you planning any tests, @acrellin?
> 
> We already have tests for this component, which I've updated accordingly. I'll add another test for creating new users

@stefanv done.

PR #586, opened 2020-07-20
Mobile profile link

joelachance:
A very simple example of using styled-components. This replaces the Responsive component, and hopefully everyone finds its more explicit, easier to read, and a bit simpler.

We can continue down this path if people like this example.

This will keep the ProfileDropdown link on the right side of the header instead of positioned over the header on widths < 768px. 

Screen Shot 2020-07-19 at 10 19 44 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "redux-thunk": "2.3.0",
     "semver": "7.3.2",
     "style-loader": "1.2.1",
+    "styled-components": "^5.1.1",
     "vega": "5.13.0",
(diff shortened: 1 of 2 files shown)


stefanv:
I think this is a much better pattern than what we had before for mobile.

@acrellin Do you agree?  If so, shall we start using styled components and see how it goes?

acrellin:
@stefanv yes, I like this a lot, and like the idea of using this pattern more widely!

stefanv:
Thanks @joelachance!  Let's keep going with this.

PR #592, opened 2020-07-21
mark test_finder xfail as it relies on a flaky web resource (NERSC DR8)

dannygoldstein:
[ch 1043]


skyportal/tests/api/test_sources.py
```diff
@@ -125,6 +126,7 @@ def test_starlist(manage_sources_token, public_source):
     assert isinstance(data['data']["starlist_info"][2]["dec"], float)
 
 
+@pytest.mark.xfail
```
profjsb, line 129:
```suggestion
@pytest.mark.xfail(strict=False)
```
We want to insure that this test will not fail the suite if this function actually succeeds (which it normally does).

PR #596, opened 2020-07-24
Classification Epic: PR #4 Frontend on Source

profjsb:
Jul-24-2020 16-43-52
Jul-24-2020 16-44-32

Jul-23-2020 18-50-28

This PR adds the frontend interaction on the Source page for Classifications:

   - Classifications are taxonomy specific
   - Like Comments, Classifications (and Taxonomies) respect Group membership
   - probabilities are required to be [0,1] (default to 1.0)
   - Summary classification (most recent classification for each taxonomy) is shown at the top of each source

Possible extensions:
   - add fine-grained group viewing when a classification is made (currently defaults to the groups that the user/token is in)
   -  ~~only show the most recent taxonomy of each type (defaults to show all taxonomies)~~


static/js/components/ClassificationEntry.jsx
```diff
@@ -0,0 +1,98 @@
...
+import FormControlLabel from "@material-ui/core/FormControlLabel";
+import Typography from "@material-ui/core/Typography";
+import Button from "@material-ui/core/Button";
+import Box from "@material-ui/core/Box";
+
+import FormValidationError from "./FormValidationError";
+import styles from "./ClassificationEntry.css";
+
+const ClassificationEntry = ({ addClassification }) => {
+  const userGroups = useSelector((state) => state.groups.user);
```
dannygoldstein, line 16: Groups are displayed for classifications like comments, but there is no option in the comments form to set group ownership. So everything is being shared by default?

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+    dispatch({name: "probability_errored", value: false});
+    dispatch({name: "probability", value: 1.0});
+  };
+
+  const handleClasschange = (event, value) => {
+    dispatch({name: "classification", value: value});
+    dispatch({name: "probability_select_enabled", value: true});
+  };
+
+  const processProb = (event, value) => {
```
dannygoldstein, line 93: When I classify something within a single taxonomy with P=0.9, then classify it as something else within the same taxonomy with P=0.1, I'd expect both classifications to be displayed with their corresponding probabilities. If I have a total probability > 1, I'd expect an error to be thrown. Instead, what happens is the first classification (and its probability) are just overwritten by the second, so the only way the probabilities will sum to one is if I classify something with P=1. If only one classification can be supported at a time per taxonomy, the Probability selector is not really being used.

static/js/components/ClassificationList.jsx
```diff
@@ -0,0 +1,103 @@
...
+              {author_name}
+            </span>
+          </span>
+          &nbsp;
+          <span className={styles.classificationTime}>
+            {dayjs().to(dayjs(created_at))}
+          </span>
+          &nbsp;
+          <Tooltip title={groups.map((group) => group.name).join(", ")}>
+            <GroupIcon fontSize="small" style={{ paddingTop: "6px", paddingBottom: "0px" }} />
```
dannygoldstein, line 61: See comment about groups above

static/js/components/ClassificationList.jsx
```diff
@@ -0,0 +1,103 @@
...
+          </Tooltip>
+        </div>
+        <div className={styles.wrap} name={`classificationDiv${id}`}>
+          <div className={styles.classificationMessage}>
+            {classification} (P={probability})
```
dannygoldstein, line 66: Currently the classifications are displayed in two different parts of the page, header tags and the classification "comment list". I think only one of these is needed. I would opt for the header tags that aggregate all the classifications of a single object within a taxonomy into a single tooltip. The comment list has the downside of having different classifications not grouped by taxonomy, so trickier to track the classification history of an object within a single taxonomy. If the comment list is kept,

dannygoldstein: It should also display the taxonomy and version that a classification was made under. But I would advocate for consolidating everything into the header tags

static/js/components/Source.jsx
```diff
@@ -63,6 +70,47 @@ const Source = ({ route }) => {
...
+      Object.keys(groupedClasses).forEach((item, i) => sortedClasses.push(groupedClasses[item].sort(function(a,b){
+            return a.modified < b.modified ? 1 : -1;
+      })));
+
+      if (sortedClasses.length > 0) {
+         return (
+            <div>
+             <b>Classification: </b>
+             {sortedClasses.map((c, index) => (
+                    <Tooltip key={index} disableFocusListener disableTouchListener title={`taxonomy=${c[0].taxonomy_id} P=${c[0].probability}`}>
```
dannygoldstein, line 99: The taxonomy name, rather than the id, should be rendered

dannygoldstein: I would also advocate that we show who classified the thing within the taxonomy within these tooltips

static/js/components/ClassificationEntry.jsx
profjsb: The idea for this first PR is to have classifications shared with all groups that the user/token belongs to. We can add fine grained classifications in another PR.

static/js/components/ClassificationForm.jsx
profjsb: No classifications are overwritten (the most recent one is shown near the top of the Source page) and by design I think it's ok for classifications not to sum to 1. For example, if new data arrives or another ML model with the same token wants to write a classification statement which may be in conflict with the first one. If you set P=0 then that classification will not be shown in the summary at top.

I thought about having classification groupings which would allow someone to add a classification vector with the same tag/group. But for now all the classes are atomized and distinct from each other (for accounting purposes).

static/js/components/ClassificationList.jsx
profjsb: I disagree. It's helpful to see all the classifications (as they change as more data arrives). We also need a frontend way for people to delete classifications (like comments). I think it's also helpful to have a "quicklook" view at the top.

static/js/components/Source.jsx
profjsb: I disagree -- it's too much info for a quicklook summary.

configs/db_demo.yaml
```diff
@@ -148,7 +148,21 @@ sources:
       sub: skyportal/tests/data/16fil_sub.png
     token: "load_demo_data token"
 taxonomies:
-  - name: Demo taxonomy
+  - name: Simple Taxonomy
```
dannygoldstein, line 151: New taxonomy shows that taxonomy select list is not ordred by Taxonomy ID. Consider sorting this list by ID, or removing the ID from the list

Screen Shot 2020-07-23 at 8 20 31 PM

configs/db_demo.yaml
```diff
@@ -148,7 +148,21 @@ sources:
...
+    hierarchy:
+      class: Time-domain Source
```
dannygoldstein, line 153: Time-Domain source renders as a class, not a subclass, is this intended? 
Screen Shot 2020-07-23 at 8 22 33 PM

should class probabilities be shown in addition to subclass probabilities? Or maybe some kind of tree visualization?

dannygoldstein:
One more comment - why the 0.1 granularity on probabilities? Presumably ML classifiers will emit scores at higher levels of granularity?

configs/db_demo.yaml
profjsb: I think that the tree viz is outside the scope of this PR. It's a major amount of work as you can imagine. We could have a page for taxonomy viz/adding.

profjsb:
> One more comment - why the 0.1 granularity on probabilities? Presumably ML classifiers will emit scores at higher levels of granularity?

That's only a UX decision. Any float from 0 to 1 is accepted by the API. And that's a free-form box so the user can type in anything.

configs/db_demo.yaml
profjsb: Removed "Time-domain Source" as a viable class (it's needed by the schema validator for the taxonomy itself)

static/js/components/ClassificationList.jsx
profjsb: > It should also display the taxonomy and version that a classification was made under. But I would advocate for consolidating everything into the header tags

Now showing classification author and the name of the taxonomy

configs/db_demo.yaml
profjsb: Done.

configs/db_demo.yaml
profjsb: Done.

profjsb:
Updated the UI as per discussions. See the interaction demos above, which includes the ability to "see" the taxonomy path for each selection.  Also, now filtering the allowed taxonomies to include only the "islatest" from the same astronomy namespace.

dannygoldstein:
@profjsb Thanks for the changes. I marked everything from the last round of comments as "resolved" and opened up some new issues for things delegated to future PRs. 

I checked out the PR at 00d7772 and did `make load_demo_data && make run`. When I try to render the source page, I get a javascript error that prevents the page from rendering on Chrome:

Screen Shot 2020-07-25 at 1 09 21 AM

It looks like the source of the error is related to some new React jsx in Source.jsx:

Screen Shot 2020-07-25 at 1 09 29 AM

For the tree stuff, it's nice to see the classes displayed on the selector menu now. However, the text density of each node is high. I found this readymade tool which may simplify the menu by having only one "class"/"subclass" per node and displaying stuff for indentation to reduce text density: 

https://github.com/iannbing/react-simple-tree-menu
https://iannbing.github.io/react-simple-tree-menu/

This menu might be a good use case for that.

profjsb:
> @profjsb Thanks for the changes. I marked everything from the last round of comments as "resolved" and opened up some new issues for things delegated to future PRs.
> 
> I checked out the PR at [00d7772](https://github.com/skyportal/skyportal/commit/00d7772f28b518bbd2661d6d162f034aa57397eb) and did `make load_demo_data && make run`. When I try to render the source page, I get a javascript error that prevents the page from rendering on Chrome:
> 
> Screen Shot 2020-07-25 at 1 09 21 AM
> 
> It looks like the source of the error is related to some new React jsx in Source.jsx:
> 
> Screen Shot 2020-07-25 at 1 09 29 AM
> 
> For the tree stuff, it's nice to see the classes displayed on the selector menu now. However, the text density of each node is high. This readymade tool may simplify the menu by having only one "class"/"subclass" per node and displaying stuff with indentation to reduce text density:
> 
> https://github.com/iannbing/react-simple-tree-menu
> https://iannbing.github.io/react-simple-tree-menu/

Sorry, @dannygoldstein -- it looks like the linting process broke some functionality. Will fix and let you know.

profjsb:
Will leave the new classification selector idea to another issue/PR.

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const onSubmit = () => {
+    // TODO: allow fine-grained user groups in submission
+    const formData = {
+      taxonomy_id: latest_taxonomyList[state.taxonomy_index].id,
+      obj_id,
+      classification: state.classification.class,
+      probability: parseFloat(state.probability)
+    };
+    // We need to include this field in request, but it isn't in form data
```
acrellin, line 117:
```suggestion
```

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+  return (
+    <MenuItem value={index} key={index.toString()}>
+      {render_string}
+    </MenuItem>
+  );
+}
+
+
+const ClassificationForm = ({ obj_id, taxonomyList }) => {
+  const latest_taxonomyList = taxonomyList.filter((t) => t.isLatest);
```
acrellin, line 31: Either snake_case or camelCase, but not both (JS standard is camelCase, but we have occasional exceptions)

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+        };
+      default:
+        return {
+          ...state,
+          [action.name]: action.value
+        };
+    }
+  }
+
+  const submitDispatch = useDispatch();
```
acrellin, line 49:
```suggestion
  const reduxDispatch = useDispatch();
```

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+    taxonomy_index: latest_taxonomyList.length > 0 ? 0 : null,
+    classification: null,
+    probability: 1.0,
+    class_select_enabled: false,
+    probability_select_enabled: false,
+    probability_errored: false,
+    allowed_classes: latest_taxonomyList.length > 0 ?
+      latest_taxonomyList[0].allowed_classes : [null]
+  };
+  const [state, dispatch] = useReducer(reducer, initialState);
```
acrellin, line 61:
```suggestion
  const [state, localDispatch] = useReducer(reducer, initialState);
```

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+  const { handleSubmit } = useForm();
```
acrellin, line 62: react-hook-form is not actually ever being used below, so can remove this and change the form's `onSubmit` below

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+      probability: parseFloat(state.probability)
+    };
+    // We need to include this field in request, but it isn't in form data
+    submitDispatch(Actions.addClassification(formData));
+  };
+
+
+  return (
+    <div>
+      <form onSubmit={handleSubmit(onSubmit)}>
```
acrellin, line 124:
```suggestion
      <form onSubmit={onSubmit}>
```
after removing react-hook-form instantiation above, which is entirely unused except for here

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+              {latest_taxonomyList.map((taxonomy, index) => makeMenuItem(
+                taxonomy, index
+              ))}
+            </Select>
+          </FormControl>
+          <div style={{ display: state.class_select_enabled ? "block" : "none" }}>
+            <Autocomplete
+              options={state.allowed_classes}
+              id="classification"
+              getOptionSelected={(option) => {
```
acrellin, line 143: how does this ever return anything other than `true`?

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+            <TextField
+              id="probability"
+              label="Probability"
+              error={state.probability_errored}
+              type="number"
+              defaultValue="1.0"
+              helperText="[0-1]"
+              InputLabelProps={{
+                shrink: true,
+
```
acrellin, line 174:
```suggestion
```

static/js/components/ClassificationList.jsx
```diff
@@ -0,0 +1,132 @@
...
+import * as sourceActions from "../ducks/source";
+import styles from "./ClassificationList.css";
+
+dayjs.extend(relativeTime);
+
+const ClassificationList = () => {
+  const [hoverID, setHoverID] = useState(null);
+
+  const handleMouseHover = (id, userProfile, author) => {
+    if (userProfile.roles.includes("Super admin") || userProfile.username === author) {
```
acrellin, line 21: Should group admins also be able to delete?

static/js/components/ClassificationList.jsx
```diff
@@ -0,0 +1,132 @@
...
+  const handleMouseLeave = () => {
+    setHoverID(null);
+  };
+
+  const dispatch = useDispatch();
+  const taxonomyList = useSelector((state) => state.taxonomies.taxonomyList);
+  const source = useSelector((state) => state.source);
+  const obj = source;
+  const userProfile = useSelector((state) => state.profile);
+  // const acls = useSelector((state) => state.profile.acls);
```
acrellin, line 35: Should this be removed?

static/js/components/ClassificationList.jsx
```diff
@@ -0,0 +1,132 @@
...
+      setHoverID(id);
+    }
+  };
+
+  const handleMouseLeave = () => {
+    setHoverID(null);
+  };
+
+  const dispatch = useDispatch();
+  const taxonomyList = useSelector((state) => state.taxonomies.taxonomyList);
```
acrellin, line 31: Prefer object destructuring assignment in these cases

static/js/components/ClassificationList.jsx
```diff
@@ -0,0 +1,132 @@
...
+  // newest classifications on top reverse sort the classifications by created_at
+  const sorted_classifications = classifications.sort((a, b) => (
+    a.created_at > b.created_at ? -1 : 1));
+
+  const items = sorted_classifications.map(
+    ({ id, author_name, created_at, classification, probability, taxonomy_id, groups }) => {
+      let taxname = taxonomyList.filter((i) => i.id === taxonomy_id);
+      if (taxname.length > 0) {
+        taxname = taxname[0].name;
+      }
```
acrellin, line 49: Should this have an else block?

static/js/ducks/taxonomies.js
```diff
@@ -0,0 +1,25 @@
...
+export const FETCH_TAXONOMIES_OK = 'skyportal/FETCH_TAXONOMIES_OK';
+
+export const fetchTaxonomies = () => (
+  API.GET('/api/taxonomy', FETCH_TAXONOMIES)
+);
+
+const reducer = (state={ taxonomyList: [] }, action) => {
+  switch (action.type) {
+    case FETCH_TAXONOMIES_OK: {
+      const taxonomies = action.data;
```
acrellin, line 14: if we do `const taxonomyList = ...` here, then line 17 below can simply be `taxonomyList`

static/js/components/Source.jsx
```diff
@@ -63,6 +70,63 @@ const Source = ({ route }) => {
     );
   }
 
+  const groupBy = (array, key) => array.reduce((result, cv) => {
+    (result[cv[key]] = result[cv[key]] || []).push(cv);
+    return result;
+  }, {});
+  function showClassification() {
```
acrellin, line 77: We should probably make this a regular component and just pass in the required props

static/js/components/Source.jsx
```diff
@@ -73,6 +137,7 @@ const Source = ({ route }) => {
         </div>
 
         <br />
+        {showClassification()}
```
acrellin, line 140: See comment above about making this a regular React component that can be rendered in JSX tags

static/js/components/Source.jsx
```diff
@@ -63,6 +70,63 @@ const Source = ({ route }) => {
...
+    if (sortedClasses.length > 0) {
+      return (
+        <div>
+          <b>Classification: </b>
+          {sortedClasses.map((c) => {
+            let name = taxonomyList.filter((i) => i.id === c[0].taxonomy_id);
+            if (name.length > 0) {
+              name = name[0].name;
+            }
+            return (
```
acrellin, line 98: what's with the return statement inside another return statement? not sure what's going on here or if this is even valid

static/js/components/Source.jsx
```diff
@@ -63,6 +70,63 @@ const Source = ({ route }) => {
     );
   }
 
+  const groupBy = (array, key) => array.reduce((result, cv) => {
+    (result[cv[key]] = result[cv[key]] || []).push(cv);
```
acrellin, line 74: This is a bit dense; perhaps an inline comment explaining the logic would help

static/js/components/ClassificationList.jsx
profjsb: Good idea.

skyportal/handlers/api/taxonomy.py
```diff
@@ -10,40 +9,54 @@
...
-            return self.error(
-                'Taxonomy does not exist or is not available to user.'
-            )
-
-        if not isinstance(taxonomy, list):
-            return self.error('Problem retreiving taxonomy')
-
-        return self.success(data=taxonomy[0])
+        if taxonomy_id is not None:
+            taxonomy = Taxonomy. \
```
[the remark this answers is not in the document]
profjsb: done.

skyportal/handlers/api/taxonomy.py
```diff
@@ -161,10 +174,25 @@ def post(self):
         provenance = data.get('provenance', None)
 
         # compute the allowable classes
-        jsonpath_expr = parse('$..class')
-        allowed_classes = list(set(
-                               [x.value for x in jsonpath_expr.find(hierarchy)]
-                               ))
+        rez = []
+        def dict_path(path,my_dict):
```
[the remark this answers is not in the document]
profjsb: No, it a recursive function and isn't pure.

skyportal/handlers/api/taxonomy.py
```diff
@@ -161,10 +174,25 @@ def post(self):
         provenance = data.get('provenance', None)
 
         # compute the allowable classes
-        jsonpath_expr = parse('$..class')
-        allowed_classes = list(set(
-                               [x.value for x in jsonpath_expr.find(hierarchy)]
-                               ))
+        rez = []
```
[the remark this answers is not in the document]
profjsb: done.

skyportal/handlers/api/taxonomy.py
```diff
@@ -161,10 +174,25 @@ def post(self):
...
+                if cla == 'Time-domain Source':
+                    cla = ''
+                rez.append(path+"-><b>"+cla+"</b>")
+                for i, item in enumerate(my_dict.get("subclasses", [])):
+                    if isinstance(item ,dict):
+                        dict_path( path + "->" + cla, item)
+
+        dict_path("", hierarchy)
+        allowed_classes = []
+        for v in rez:
```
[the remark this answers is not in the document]
profjsb: Fixed.

skyportal/tests/frontend/test_sources.py
```diff
@@ -19,6 +20,46 @@ def test_public_source_page(driver, user, public_source):
...
+                             },
+                       token=taxonomy_token)
+    assert status == 200
+    taxonomy_id = data['data']['taxonomy_id']
+
+    driver.get(f"/become_user/{user.id}")  # TODO decorator/context manager?
+    driver.get(f"/source/{public_source.id}")
+    driver.wait_for_xpath(f'//div[text()="{public_source.id}"]')
+    taxonomy_button = driver.wait_for_xpath(f'//div[@id="tax-select"]')
+    taxonomy_button.click()
```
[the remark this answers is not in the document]
profjsb: Thank you! In another PR?

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+                return (true);
+              }}
+              value={state.classification || ""}
+              onChange={handleClasschange}
+              getOptionLabel={(option) => option.class || ""}
+              renderInput={(params) => <TextField {...params} style={{ width: '100%' }} label="Classification" fullWidth />}
+              renderOption={(option) => {
+                const val = `${option.label}`;
+                return (
+                  <div dangerouslySetInnerHTML={{ __html: val }} />
```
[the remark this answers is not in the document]
profjsb: Could use your help on this one... I want to be able to do bolded <b> and unbolded text inside a menu item. This was the only way I could figure out how to do it.

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,207 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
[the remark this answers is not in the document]
profjsb: Done.

static/js/components/ClassificationForm.jsx
profjsb: fixed.

static/js/components/ClassificationForm.jsx
profjsb: done.

static/js/components/ClassificationForm.jsx
profjsb: done

static/js/components/ClassificationForm.jsx
profjsb: done.

static/js/components/ClassificationForm.jsx
profjsb: done.

static/js/components/ClassificationForm.jsx
profjsb: Fixed to be more explicit.

static/js/components/ClassificationList.jsx
profjsb: leaving it in as a comment because we'll need it when we add group granularity.

static/js/components/Source.jsx
profjsb: It's inside a map which is inside an if/then clause. Adding some comments inline.

static/js/components/Source.jsx
profjsb: done.

static/js/components/Source.jsx
profjsb: Can we leave this for another PR?

acrellin: As someone who will have to edit/update/maintain this code, I'd rather we get it right before it gets merged so that it doesn't fall through the cracks and potentially cause headaches later.

static/js/components/Source.jsx
acrellin: Ah right, I see now.

skyportal/handlers/api/taxonomy.py
```diff
@@ -161,10 +174,28 @@ def post(self):
         provenance = data.get('provenance', None)
 
         # compute the allowable classes
-        jsonpath_expr = parse('$..class')
-        allowed_classes = list(set(
-                               [x.value for x in jsonpath_expr.find(hierarchy)]
-                               ))
+        result = []
+
+        def dict_path(path, my_dict):
```
stefanv, line 179: Could you explain what these two paragraphs do?

profjsb: This generates the list of allowed classes for this hierarchy, capturing both the leaf nodes as well as the parent nodes as possible classes. In addition, it generates a label to be show during the selection process. Eg, the output is:
```
[{"class": "SN", "label": "<b>SN</br>⇦Cataclysmic"}, {"class": "Ia", "label": "<b>Ia</br>⇦SN⇦Cataclysmic"}]
```
The Classification selection on the Source page then uses array to allow the user to find the classification more easily using <Autocomplete> from MUI.

profjsb: Adding this response as a comment inline.

static/js/components/Source.jsx
profjsb: done

static/js/components/Source.jsx
profjsb: done

profjsb:
> @profjsb Do you think `useReducer` is really necessary here? We don't use that anywhere else in the project so far, so it adds another layer of complexity for people who have to follow the code. I'm not necessarily opposed, but I would like to know whether this logic is difficult to implement using standard react component state (it seems as though there isn't much complex specialized state logic involved).

I dont believe this adds another layer of complexity, in fact I think it makes the code easier to understand. `useReducer` is a more initiative way, IMO, to handle a bunch of different/non-trivial changes to the state. The `reducer` function in `ClassificationForm.jsx` handles different actions pretty clearly I think. From the reactjs documents:

> useReducer is usually preferable to useState when you have complex state logic that involves multiple sub-values or when the next state depends on the previous one. useReducer also lets you optimize performance for components that trigger deep updates because you can pass dispatch down instead of callbacks.

static/js/components/ClassificationForm.jsx
acrellin: Looks like this still needs to be removed

static/js/components/ClassificationForm.jsx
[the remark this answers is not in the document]
acrellin: @stefanv thoughts on this? You're the more security-minded dev among us

static/js/components/Source.jsx
```diff
@@ -73,6 +81,10 @@ const Source = ({ route }) => {
         </div>
 
         <br />
+        <ShowClassification
```
acrellin, line 84: Great!

static/js/components/ClassificationList.jsx
acrellin: This could still be changed to how you've done it in Source.jsx

profjsb:
The classification functionality does not load when I remove the `useForm()`. We need to keep it in.

`dangerouslySetInnerHTML` is because the default is to santize the text to avoid cross-scripting. For this specific text to have malicious code, it would have to be uploaded as part of the taxonomy (so a Group or Super Admin) and it would need to be carefully crafted to avoid the schema checking. I'm not worried.

acrellin:
@profjsb there are a few comments from yesterday that still need to be addressed. Let me know if you want me to help pull RHF out of that form.

profjsb:
> @profjsb there are a few comments from yesterday that still need to be addressed. Let me know if you want me to help pull RHF out of that form.

Looks like there were two remaining issues:  on the dangerousHTML question I wrote back saying *"dangerouslySetInnerHTML is because the default is to santize the text to avoid cross-scripting. For this specific text to have malicious code, it would have to be uploaded as part of the taxonomy (so a Group or Super Admin) and it would need to be carefully crafted to avoid the schema checking. I'm not worried."* **So I think we should leave it as is.**

profjsb:
[ch 994]

static/js/components/ClassificationForm.jsx
acrellin: If you remove this form onSubmit part altogether, change the button type below to "button" and set its `onClick={onSubmit}`, you can remove RHF altogether as no inputs are actually using it

acrellin: Sorry to keep insisting on this, but I really don't think we should be using RHF just to trigger our submit function without registering any components, etc, with it.

acrellin: It just adds an unnecessary layer of complexity, which will slow us down in the long run

static/js/components/ClassificationList.jsx
acrellin: This was never addressed, but is a minor nitpick and can be skipped as eslint looks to be okay with it.

static/js/components/ClassificationList.jsx
acrellin: How about this? Any issues if taxname.length is zero?

static/js/components/ClassificationForm.jsx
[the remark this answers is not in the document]
acrellin: @stefanv?

static/js/components/ClassificationForm.jsx
```diff
@@ -0,0 +1,199 @@
...
+import InputLabel from "@material-ui/core/InputLabel";
+import TextField from '@material-ui/core/TextField';
+import Button from '@material-ui/core/Button';
+import MenuItem from "@material-ui/core/MenuItem";
+import Autocomplete from '@material-ui/lab/Autocomplete';
+import { showNotification } from "baselayer/components/Notifications";
+import * as Actions from '../ducks/source';
+
+
+function makeMenuItem(taxonomy, index) {
```
acrellin, line 14: This should either be refactored into a react component, or, preferably, just move the JSX down to the main return statement because it's so small.

profjsb: done.

skyportal/schema.py
```diff
@@ -217,6 +217,10 @@ class PhotBaseFlexible(object):
                            missing=None, default=None, required=False)
 
 
+    altdata = fields.Field(description='Misc. alternative metadata.',
```
acrellin, line 220: This looks like an accidental inclusion (duplicates the lines above)

stefanv:
:rocket:

PR #615, opened 2020-07-30
Add facility selector to starlist component

dannygoldstein:
Jul-29-2020 21-30-32

This PR adds a dropdown menu to the StarList and ObservingRunStarList components  that allows you to choose the facility formatting guidelines that you would like applied to your starlist. 

This PR is an extension of #711 . It changes the `ObservingRunStarList` API (to have it take an `observingRun` object as opposed to an `observingRunId`) as defined in that PR, but does not alter the single-object `StarList` API that renders on the source page, which was written by others. 

This PR also fixes a bug in `skyportal/tests/frontend/test_sources.py::test_show_starlist` that caused the test to pass without the expected text being present. 

[ch1095]

### DIFF
--- a/static/js/components/StarList.jsx
+++ b/static/js/components/StarList.jsx
@@ -2,34 +2,59 @@ import React, { useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import PropTypes from "prop-types";
 
-import { GET } from "../API";
+import Select from "@material-ui/core/Select";
+import MenuItem from "@material-ui/core/MenuItem";
+import InputLabel from "@material-ui/core/InputLabel";
 
+import { GET } from "../API";
 import styles from "./StarList.css";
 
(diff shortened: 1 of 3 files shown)


dannygoldstein:
green with `click_and_hold` fix

static/js/components/StarList.jsx
```diff
@@ -2,57 +2,91 @@ import React, { useEffect, useState } from "react";
...
+  };
+
+  return (
+    <div className={styles.starListDiv}>
+      <code className={styles.starList}>
+        <div className={styles.codeText}>
+          {starList &&
+            starList.map((item, idx) => (
+              // eslint-disable-next-line react/no-array-index-key
+              <React.Fragment key={idx}>
```
dmitryduev, line 25: Isn't `<>` is now recommended instead of `<React.Fragment>`?

PR #620, opened 2020-07-30
System admin permissions

### LINKED ISSUE #619: Allow admin (token) to do everything
Currently, trying to post a telescope, e.g., would error:

```
/telescope: You must specify at least one group of which you are a member.
```

Group membership checks have to be updated.

acrellin:
This PR improves permissions checking by introducing a new `User.accessible_groups` attribute, which returns either the normal user's group list (`User.groups`) in the case of non-system admin users, or, in the case of system admin users, returns all groups. This new attribute is used in place of `User.groups` for checking whether a user has access to the specified record(s). Some security lapses are also addressed.

Closes #619


stefanv:
Thanks very much, Ari.

PR #621, opened 2020-07-31
Profile page improvements and Avatars

profjsb:
This PR introduces avatars for users and a frontend to add basic information for a user via the front-end profile page. When the username or `contact_email` is set to a valid email, and that email is connected to gravatar.com, then the gravatar will be used as the avatar. Otherwise, the first and last name of the user is used within a colored circle. The color of the circle is derived from a hash of the username.

Kapture 2020-08-01 at 17 04 29

### DIFF
--- a/skyportal/handlers/api/internal/profile.py
+++ b/skyportal/handlers/api/internal/profile.py
@@ -82,7 +82,9 @@ class ProfileHandler(BaseHandler):
             'first_name': user.first_name,
             'last_name': user.last_name,
             'contact_email': user.contact_email,
-            'contact_phone': user.contact_phone.e164,
+            'contact_phone': (
+              None if user.contact_phone is None else user.contact_phone.e164
+            ),
             'gravatar_url': user.gravatar_url,
             'roles': user_roles,
             'acls': user_acls,
@@ -117,7 +119,6 @@ class ProfileHandler(BaseHandler):
                               '"preferences" parameter'
                              )
         preferences = data['preferences']
-
         if preferences.get("first_name"):
             self.current_user.first_name = preferences.pop("first_name")
 
@@ -126,10 +127,10 @@ class ProfileHandler(BaseHandler):
 
         if preferences.get("contact_phone"):
             phone = preferences.pop("contact_phone")
-            if phone is not None:
+            if phone not in [None, ""]:
                 try:
                     if not phonenumbers.is_possible_number(
-                            phonenumbers.parse(phone, None)):
+                            phonenumbers.parse(phone, "US")):
                         return self.error(
                                   'Phone number given is not valid'
                                  )
@@ -137,11 +138,13 @@ class ProfileHandler(BaseHandler):
                     return self.error(
                                   'Could not parse input as a phone number'
                                  )
-            self.current_user.contact_phone = phone
+                self.current_user.contact_phone = phone
+            else:
+                self.current_user.contact_phone = None
 
         if preferences.get("contact_email"):
             email = preferences.pop("contact_email")
-            if email is not None:
+            if email not in [None, ""]:
                 if not validate_email(
                        email_address=email,
                         check_regex=True,
@@ -151,7 +154,9 @@ class ProfileHandler(BaseHandler):
                     return self.error(
                               'Email does not appear to be valid'
                              )
-            self.current_user.contact_email = email
+                self.current_user.contact_email = email
+            else:
+                self.current_user.contact_email = None
 
         # Do not save blank fields (empty strings)
         for k, v in preferences.items():
--- a/skyportal/handlers/api/internal/profile.py
+++ b/skyportal/handlers/api/internal/profile.py
@@ -114,8 +114,9 @@ class ProfileHandler(BaseHandler):
         """
         data = self.get_json()
 
-        user = (User.query.filter(User.username == self.current_user.username)
-                    .first())
+        user = User.query.filter(
+            User.username == self.current_user.username
+        ).first()
 
         if 'preferences' not in data:
             return self.error(
@@ -174,7 +175,6 @@ class ProfileHandler(BaseHandler):
             user_prefs = recursive_update(user_prefs, preferences)
         user.preferences = user_prefs
 
-        DBSession.add(user)
         DBSession.commit()
         if 'newsFeed' in preferences:
             self.push(action='skyportal/FETCH_NEWSFEED')
--- a/skyportal/handlers/api/internal/profile.py
+++ b/skyportal/handlers/api/internal/profile.py
@@ -131,8 +131,6 @@ class ProfileHandler(BaseHandler):
 
         if preferences.get("contact_phone") is not None:
             phone = preferences.pop("contact_phone")
-            print("->",phone,"<-")
-            print(phone not in [None, ""])
             if phone not in [None, ""]:
                 try:
                     if not phonenumbers.is_possible_number(
--- a/skyportal/handlers/api/internal/profile.py
+++ b/skyportal/handlers/api/internal/profile.py
@@ -11,7 +11,6 @@ from ...base import BaseHandler
 from ....models import User, DBSession
 
 
-
 class ProfileHandler(BaseHandler):
     @auth_or_token
     def get(self):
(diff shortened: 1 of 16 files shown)


profjsb:
[ch 1088]

static/js/components/UpdateProfileForm.jsx
```diff
@@ -0,0 +1,151 @@
...
+                  name="firstName"
+                  id="firstName_id"
+                  error={!!errors.firstName}
+                  helperText={errors.firstName ? "Required" : ""}
+                />
+              </Grid>
+              <Grid item xs={6} sm={3}>
+                <InputLabel htmlFor="lastName_id">Last Name</InputLabel>
+                <TextField
+                  inputRef={register({ required: true })}
```
acrellin, line 84: Shouldn't we allow for users supplying only some of these fields and not have them all be required? Perhaps they want to add a first name but not last for privacy, or just want to add an email/phone number but don't want their full name displayed...

skyportal/handlers/api/internal/profile.py
```diff
@@ -89,20 +113,68 @@ def put(self):
...
+                user.contact_email = email
+            else:
+                user.contact_email = None
+
         # Do not save blank fields (empty strings)
         for k, v in preferences.items():
             if isinstance(v, dict):
-                preferences[k] = {key: val for key, val in v.items() if val != ''}
-        user_prefs = deepcopy(self.current_user.preferences)
+                preferences[k] = \
```
acrellin, line 168: 88 is the new 79/80, so this can be reverted (and using the curly braces in this expression to span two lines, or introducing parentheses in other cases, is preferred over trailing slashes)

static/js/components/ShowAvatar.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+import Avatar from '@material-ui/core/Avatar';
+
+const theme = createMuiTheme({});
+
+const ShowAvatar = ({ size, firstName, lastName, username, gravatarUrl }) => {
+  // use the hash of the username (which is in the gravatarUrl) to
+  // select a unique color for this user
+  function bgcolor() {
+    let hash = gravatarUrl.split("/");
```
stefanv, line 13: Why not just use the username that was passed in?

stefanv:
Once the preferences issue has been worked out, I'll gladly take another look.

static/js/components/ShowAvatar.jsx
```diff
@@ -0,0 +1,58 @@
...
+      // make the color string with a slight transparency
+      return `#${hash.slice(0, 6)}aa`;
+    } else {
+      return '#aaaaaaaa';
+    }
+  }
+
+  const usercolor = bgcolor();
+
+  const backUpLetters = (firstName === null ? username.slice(0, 2) :`${firstName?.charAt(0)}${lastName?.charAt(0)}`);
```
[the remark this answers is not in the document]
profjsb: Yes, that's a style choice. Also, the front end wont allow adding only a first name and not a last name.

static/js/components/ShowAvatar.jsx
profjsb: `gravatarUrl` is a hash of the `username` (by default, and of the email address if it's given), already calculated on the Python side in the `User` model:
```
    @property
    def gravatar_url(self):
        email = self.contact_email if \
                self.contact_email is not None else self.username

        digest = md5(email.lower().encode('utf-8')).hexdigest()
        # return a 404 status code if not found on gravatar
        return f'https://www.gravatar.com/avatar/{digest}?d=404'
```
I'm avoiding having to recalculate the hash on the client side.

skyportal/tests/frontend/test_profile.py
```diff
@@ -11,8 +13,72 @@ def test_token_acls_options_rendering1(driver, user):
...
+    first_name_entry.send_keys(first_name)
+    last_name_entry = driver.wait_for_xpath('//input[@name="lastName"]')
+    last_name = ""
+    last_name_entry.send_keys(last_name)
+
+    driver.scroll_to_element_and_click(
+        driver.find_element_by_xpath('//*[@id="updateProfileButton"]'))
+
+    helper = driver.wait_for_xpath('//p[@id="lastName_id-helper-text"]')
+    assert helper.text == 'Required'
```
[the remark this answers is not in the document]
acrellin: @profjsb here

static/js/components/ShowAvatar.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { createMuiTheme } from '@material-ui/core/styles';
+
+import Avatar from '@material-ui/core/Avatar';
+
+const theme = createMuiTheme({});
+
+const ShowAvatar = ({ size, firstName, lastName, username, gravatarUrl }) => {
```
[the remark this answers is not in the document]
acrellin: @profjsb here

skyportal/tests/frontend/test_profile.py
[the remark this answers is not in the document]
profjsb: This does test to see if the "Required" error is shown: path does not exist in the DOM if this field is not errored. It only is injected into the DOM on error.

static/js/components/ShowAvatar.jsx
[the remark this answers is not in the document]
profjsb: Done.

static/js/components/ShowAvatar.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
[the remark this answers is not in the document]
profjsb: Done

static/js/components/UpdateProfileForm.jsx
profjsb: now allowing just a first name. Last name is optional.

skyportal/handlers/api/internal/profile.py
```diff
@@ -52,32 +67,65 @@ def get(self):
...
+                "roles": user_roles,
+                "acls": user_acls,
+                "tokens": user_tokens,
+                "preferences": self.current_user.preferences or {},
+            }
+        )
 
-    def put(self):
+    @auth_or_token
+    def patch(self):
```
profjsb, line 100: Note: this is a patch not a put (it's an update to an existing record).

static/js/components/ProfileDropdown.jsx
```diff
@@ -33,9 +34,23 @@ const ProfileDropdown = () => {
     <Container>
       <Dropdown ref={dropdown}>
         <DropdownTrigger>
-          { profile.username }
-          {' '}
+          <div style={{ display: "flex",
```
stefanv, line 37: This is fine for now, but I'll be writing up a styling guide soon.  I recommend we do not use inline styles in general, and instead use `makeStyles` and `useStyles`.  No reason to change this for now, I will update when I review the entire code for consistent styling.

static/js/components/UserProfileInfo.jsx
```diff
@@ -0,0 +1,61 @@
...
+          alignItems: "center" }}
+        >
+          <UserAvatar
+            size={128}
+            firstName={profile.first_name}
+            lastName={profile.last_name}
+            username={profile.username}
+            gravatarUrl={profile.gravatar_url}
+          />
+            &nbsp;&nbsp;
```
stefanv, line 29: Instead of these, we can use https://material-ui.com/system/spacing/, in this case probably `mr`.

static/js/components/UserProfileInfo.jsx
```diff
@@ -0,0 +1,61 @@
...
+          <h2
+            id="userRealname"
+            style={{ visibility: !profile.first_name && !profile.last_name ? "hidden" : "visible" }}
```
stefanv, line 32:
```
!(profile.first_name || profile.last_name)
```

static/js/components/UserProfileInfo.jsx
```diff
@@ -0,0 +1,61 @@
...
+          >
+            {profile.first_name}
+            {' '}
+            {profile.last_name}
```
stefanv, line 36: These render OK when either is empty?

profjsb: > These render OK when either is empty?

Yes

static/js/components/UserProfileInfo.jsx
profjsb: will hold off on this - should make this apart of the styles revamp

Issue #639, opened 2020-08-03
Get telescopes by name


dmitryduev:
Following up on a conversation with @acrellin:
Telescope names should be unique and it should thus be possible to get a telescope by its name with, e.g. ` GET /api/telescopes/name/<telescope_name>`

acrellin:
Also, consider whether to also implement fetching by name for groups, users, and instruments (we'll need to ensure that the `name`/`username` columns are always unique for this)

acrellin:
@profjsb @dannygoldstein any reason we can't make telescope and instrument names unique so that we can add an endpoint to fetch by name?

PR #646, opened 2020-08-04
Update handlers to allow for fetching telescopes, instruments, and groups by name

acrellin:
Also ensures that the `name` columns in the relevant tables are unique.

Adds new API tests.

Closes https://github.com/skyportal/skyportal/issues/639

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -344,3 +344,35 @@ class GroupUserHandler(BaseHandler):
         self.push_all(action='skyportal/REFRESH_GROUP',
                       payload={'group_id': int(group_id)})
         return self.success()
+
+
+class GroupByNameHandler(BaseHandler):
+    @auth_or_token
+    def get(self, group_name):
+        """
+        ---
+        description: Retrieve a group by name
+        parameters:
+          - in: path
+            name: group_name
+            required: true
+            schema:
+              type: string
+        responses:
+          200:
+            content:
+              application/json:
+                schema: SingleGroup
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+        group = Group.query.filter(Group.name == group_name).first()
+        if group is None:
+            return self.error("Invalid group name.")
+        # Ensure user has access to group
+        if group not in self.current_user.accessible_groups:
+            return self.error("Insufficient permissions")
+
+        return self.success(data=group)
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -44,6 +44,12 @@ class GroupHandler(BaseHandler):
                   schema: Error
         multiple:
           description: Retrieve all groups
+          parameters:
+            - in: query
+              name: name
+              schema:
+                type: string
+              description: Fetch by name (exact match)
           responses:
             200:
               content:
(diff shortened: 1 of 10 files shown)


stefanv:
@acrellin What do you think of `byName`?  Also, any thoughts about query string vs additional path element?

stefanv:
@acrellin Actually, thinking about it, why not just use the same endpoints and try both---if ID fails, try name?

skyportal/handlers/api/group.py
```diff
@@ -344,3 +344,35 @@ def delete(self, group_id, username):
...
+          200:
+            content:
+              application/json:
+                schema: SingleGroup
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+        group = Group.query.filter(Group.name == group_name).first()
```
dmitryduev, line 371: Great, I didn't realize we had
```python
class Group(Base):
    name = sa.Column(sa.String, unique=True, nullable=False)
```
already in

skyportal/tests/api/test_telescope.py
```diff
@@ -33,6 +33,36 @@ def test_token_user_post_get_telescope(upload_data_token, public_group):
             assert data['data'][key] == post_data[key]
 
 
+def test_fetch_telescope_by_name(upload_data_token, public_group):
+    name = str(uuid.uuid4())
+    post_data = {'name': name,
```
dmitryduev, line 38: I saw `id` fields in some of the seed configs (in one of Danny's recent PRs, I think), do you know why they ended up there and were they used in any way?

stefanv:
@acrellin After chatting about it online, I think `/api/telescopes/?name=keck` is perhaps the most consistent version.

`/api/telescopes/name/keck` gets confusing, because we can have sub-hierarchies such as `/api/users/04/groups/...`—this does not match that pattern.

I could also go with `/api/telescopesByName/keck`, although I think the first is a bit cleaner.

skyportal/tests/api/test_telescope.py
acrellin: no idea!

skyportal/handlers/api/telescope.py
```diff
@@ -66,32 +66,57 @@ def post(self):
...
-        if t is not None:
+        if telescope_id is not None:
+            t = Telescope.get_if_owned_by(int(telescope_id), self.current_user)
+            if t is None:
+                return self.error(f"Could not load telescope with ID {telescope_id}")
             return self.success(data=t)
-        else:
-            return self.error(f"Could not load telescope with ID {telescope_id}")
+        tel_name = self.get_query_argument("name", None)
+        query = Telescope.query.filter(Telescope.id.in_(
```
stefanv, line 113: Maybe we should expand `get_if_owned_by` to take multiple IDs in another PR.  Thoughts?

skyportal/handlers/api/instrument.py
```diff
@@ -72,10 +78,14 @@ def get(self, instrument_id=None):
...
                                           self.current_user)
             return self.success(data=instrument)
+
+        inst_name = self.get_query_argument("name", None)
         query = Instrument.query.filter(Instrument.telescope_id.in_(
             DBSession().query(GroupTelescope.telescope_id).filter(GroupTelescope.group_id.in_(
                 [g.id for g in self.current_user.accessible_groups]
             ))))
+        if inst_name is not None:
+            query = query.filter(Instrument.name == inst_name)
```
dannygoldstein, line 88: This returns a list when someone requests an instrument by name, instead of a scalar. Is that intended?

acrellin: @dannygoldstein yes, in order to be consistent with the docs, any GET `/api/instrument` request must return an array of instruments (query parameters do not change this), while `/api/instrument/<instrument_id>` returns a scalar. We went back and forth for a while about whether to implement a new `/api/instrument/byName/<instrument_name>` or similar endpoint that would return a scalar object, but ultimately decided upon this approach.

PR #649, opened 2020-08-05
Add images to ``Top sources'' widget

mcoughlin:
This PR adds detection images to the top sources widget.

image

### DIFF
--- a/skyportal/handlers/api/internal/source_views.py
+++ b/skyportal/handlers/api/internal/source_views.py
@@ -1,5 +1,6 @@
 import datetime
 from sqlalchemy import func, desc
+from sqlalchemy.orm import joinedload
 import tornado.web
 from baselayer.app.access import auth_or_token
 from ...base import BaseHandler
@@ -32,7 +33,24 @@ class SourceViewsHandler(BaseHandler):
                      [g.id for g in self.current_user.accessible_groups]))))
              .filter(SourceView.created_at >= cutoff_day)
              .order_by(desc('views')).limit(max_num_sources))
-        return self.success(data=q.all())
+
+        thumbnail_order = ['new', 'ref', 'sub', 'sdss', 'dr8']
+        sources = []
+        for view, obj_id in q.all():
+            s = Source.get_obj_if_owned_by(  # Returns Source.obj
+                obj_id, self.current_user,
+                options=[joinedload(Source.obj)
+                         .joinedload(Obj.thumbnails)])
+
+            public_url = ""
+            for thumbtype in thumbnail_order:
+                for thumbnail in s.thumbnails:
+                    if (thumbnail.type == thumbtype) and (public_url==""):
+                        public_url = thumbnail.public_url
+            sources.append({'views': view, 'obj_id': obj_id,
+                            'public_url': public_url})    
+
+        return self.success(data=sources)
 
     @tornado.web.authenticated
     def post(self, obj_id):
--- a/skyportal/handlers/api/internal/source_views.py
+++ b/skyportal/handlers/api/internal/source_views.py
@@ -34,19 +34,13 @@ class SourceViewsHandler(BaseHandler):
              .filter(SourceView.created_at >= cutoff_day)
              .order_by(desc('views')).limit(max_num_sources))
 
-        thumbnail_order = ['new', 'ref', 'sub', 'sdss', 'dr8']
         sources = []
         for view, obj_id in q.all():
             s = Source.get_obj_if_owned_by(  # Returns Source.obj
                 obj_id, self.current_user,
                 options=[joinedload(Source.obj)
                          .joinedload(Obj.thumbnails)])
-
-            public_url = ""
-            for thumbtype in thumbnail_order:
-                for thumbnail in s.thumbnails:
-                    if (thumbnail.type == thumbtype) and (public_url==""):
-                        public_url = thumbnail.public_url
+            public_url = first_public_url(s.thumbnails)
             sources.append({'views': view, 'obj_id': obj_id,
                             'public_url': public_url})    
 
@@ -69,3 +63,11 @@ def register_source_view(obj_id, username_or_token_id, is_token):
                     is_token=is_token)
     DBSession.add(sv)
     DBSession.commit()
+
+def first_public_url(thumbnails):
+    urls = [t.public_url for t in sorted(thumbnails, key=lambda t: tIndex(t))]
+    return urls[0] if urls else ""
+
+def tIndex(t):
+    thumbnail_order = ['new', 'ref', 'sub', 'sdss', 'dr8'] 
+    return thumbnail_order.index(t) if t in thumbnail_order else len(thumbnail_order)
--- a/static/js/components/TopSources.css
+++ b/static/js/components/TopSources.css
@@ -22,12 +22,12 @@
   padding-left: 10px;
 }
 
-.ul {
-  display: block;
-  align-items: center;
-  list-style-type: none;
+.topSourceList {
+   display: block;
+   align-items: center;
+   list-style-type: none;
 }
 
 .topSource {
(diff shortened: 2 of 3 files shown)


static/js/components/TopSources.jsx
```diff
@@ -49,6 +49,9 @@ const TopSources = () => {
                   &nbsp;view(s)
                 </em>
               </span>
+              <span>
+                <img src={public_url} alt="" />
```
profjsb, line 53: The Top Sources widget is going to take up a small fraction of the homepage, so large postage stamps doesn't make sense. So the img tag should probably be set a fraction of the widget height/width. You can do this with css styling or inline.

profjsb: Probably also makes sense to have an `alt` text with the source name (which will make this more accessible and resilient).

dannygoldstein: In addition to @profjsb's suggestions, I'd suggest replacing the bullet points that are currently there with the small (~100x100px) images and letting the text float to the right of those.

mcoughlin: Yes excellent points. Should be improved.

mcoughlin:
> * [x]  Smallify images (100 x 100 px)
> * [x]  Replace `<ul>` bullets with small images

static/js/components/TopSources.jsx
```diff
@@ -32,10 +32,13 @@ const TopSources = () => {
...
         Displaying most-viewed sources
       </p>
-      <ul>
+      <ul style={{ listStyleType: "none" }}>
         {
-          sourceViews.map(({ obj_id, views }) => (
+          sourceViews.map(({ obj_id, views, public_url }) => (
             <li key={`topSources_${obj_id}_${views}`}>
+              <span>
+                <img src={public_url} alt={obj_id} width="50%" height="50%" />
```
profjsb, line 40: Make the image clickable to the source and put the styles in `TopSources.css`:
```suggestion
            <Link to={`/source/${obj_id}`}>
                <img className={styles.stamp} src={public_url} alt={obj_id} />
             </Link>
```
Here, the file `TopSources.css` can be something like:
```
.stamp {
    transition: transform 0.1s;
    width: 20%;
    height: 20%;
    margin: 0 auto;
}

.stamp:hover {
  transform: scale(2.0);
}
```
And `styles` can be added in `TopSources.jsx` like (around line 7):
```
import styles from "./TopSources.css";
```
This will look like:
Aug-04-2020 22-51-37

mcoughlin: This is really neat! Implemented.

mcoughlin:
I went with a .css flexbox based version for alignment @dannygoldstein, please see what you think.

profjsb:
Looks great. For posterity, and ease of reviewing, I'd suggest replacing the initial animated gif illustrating the functionality with a gif that captures the current approach.

mcoughlin:
review

This is with Josh's and Danny's fixes.

skyportal/handlers/api/internal/source_views.py
```diff
@@ -32,7 +33,24 @@ def get(self):
...
+            s = Source.get_obj_if_owned_by(  # Returns Source.obj
+                obj_id, self.current_user,
+                options=[joinedload(Source.obj)
+                         .joinedload(Obj.thumbnails)])
+
+            public_url = ""
+            for thumbtype in thumbnail_order:
+                for thumbnail in s.thumbnails:
+                    if (thumbnail.type == thumbtype) and (public_url==""):
+                        public_url = thumbnail.public_url
```
stefanv, line 49: If you need to sort thumbnails, you can use something like:

```
def tIndex(t): 
    return thumbnail_order.index(t) if t in thumbnail_order else len(thumbnail_order)

for thumbnail in sorted(s.thumbnails, key=lambda t: tIndex(t)):
    ...
```

But, I can't quite figure out what this is supposed to do: grab the first public URL in the order of `new`, `ref` `sub`, etc.?

If so, you could do something like:

```
def first_public_url(thumbnails):
    urls = [t.public_url for t in sorted(s.thumbnails, key=lambda t: tIndex(t))]
    return urls[0] if urls else ""
```

mcoughlin: Yes much clearer. I pushed this.

static/js/components/TopSources.jsx
```diff
@@ -32,12 +34,17 @@ const TopSources = () => {
...
-          sourceViews.map(({ obj_id, views }) => (
+          sourceViews.map(({ obj_id, views, public_url }) => (
             <li key={`topSources_${obj_id}_${views}`}>
+              <Link to={`/source/${obj_id}`}>
+                <img className={styles.stamp} src={public_url} alt={obj_id} />
+              </Link>
               <span>
                 <Link to={`/source/${obj_id}`}>
+                  &nbsp;
+                  -&nbsp;
```
dannygoldstein, line 47: Move outside link

static/js/components/TopSources.css
```diff
@@ -0,0 +1,29 @@
...
+    height: 20%;
+    margin: 0 auto;
+    vertical-align: middle;
+}
+
+.stamp:hover {
+  transform: scale(2.0);
+}
+
+.div {
```
dannygoldstein, line 13:
```suggestion
.topSourcesContainer {
```

static/js/components/TopSources.jsx
```diff
@@ -17,7 +19,7 @@ const TopSources = () => {
   ) || defaultPrefs;
 
   return (
-    <div style={{ border: "1px solid #DDD", padding: "10px" }}>
+    <div className={styles.div}>
```
dannygoldstein, line 22:
```suggestion
    <div className={styles.topSourcesContainer}>
```

static/js/components/TopSources.jsx
```diff
@@ -32,12 +34,17 @@ const TopSources = () => {
...
-      <ul>
+      <ul className={styles.ul}>
         {
-          sourceViews.map(({ obj_id, views }) => (
+          sourceViews.map(({ obj_id, views, public_url }) => (
             <li key={`topSources_${obj_id}_${views}`}>
+              <Link to={`/source/${obj_id}`}>
+                <img className={styles.stamp} src={public_url} alt={obj_id} />
+              </Link>
               <span>
```
dannygoldstein, line 44:
```suggestion
              <div>
```

static/js/components/TopSources.jsx
```diff
@@ -32,12 +34,17 @@ const TopSources = () => {
       <p>
         Displaying most-viewed sources
       </p>
-      <ul>
+      <ul className={styles.ul}>
         {
-          sourceViews.map(({ obj_id, views }) => (
+          sourceViews.map(({ obj_id, views, public_url }) => (
             <li key={`topSources_${obj_id}_${views}`}>
```
dannygoldstein, line 40:
```suggestion
            <li key={`topSources_${obj_id}_${views}`} className={styles.topSource}>
```

static/js/components/TopSources.css
```diff
@@ -0,0 +1,29 @@
...
+  padding-bottom: 10px;
+  padding-left: 10px;
+}
+
+.ul {
+   display: flex;
+   align-items: center;
+   list-style-type: none;
+}
+
```
dannygoldstein, line 29:
```suggestion
.topSource {
  margin: 10px;
}
```

mcoughlin:
danny

New look with Danny's changes.

dannygoldstein:
@mcoughlin I take back the changes I suggested here. See my PR to your branch for my newest suggested changes. What do you think of those? The list should be organized vertically as a result of that PR, similar to whats on the current marshal.

mcoughlin:
@dannygoldstein yeah i am cool with it, looks good to me!

PR #650, opened 2020-08-05
Alert streams: backend

dmitryduev:
In this PR:

- Data model and API adjustments
- API handlers to deal with alert streams
- Tests


skyportal/models.py
```diff
@@ -66,9 +66,12 @@ def process_result_value(self, value, dialect):
 class Group(Base):
     name = sa.Column(sa.String, unique=True, nullable=False)
 
-    streams = relationship('Stream', secondary='stream_groups',
+    streams = relationship('Stream', secondary='group_streams',
```
dmitryduev, line 69: Changed the order since I think in terms of adding stream access to groups rather than adding groups to streams.

dannygoldstein:
@dmitryduev Merge conflicts

dmitryduev:
All nice and green again, @dannygoldstein!

data/db_demo.yaml
```diff
@@ -12,6 +12,23 @@ user:
...
+streams:
+  - name: ZTF Public
+    altdata:
+      collection: ZTF_alerts
+      selector: [1]
+    =id: ztf_public
+  - name: ZTF Partnership
+    altdata:
+      collection: ZTF_alerts
+      selector: [1, 2]
```
dannygoldstein, line 24: Why not `2`? Technically the "partnership stream" is everything with PID=2.

data/db_demo.yaml
```diff
@@ -12,6 +12,23 @@ user:
...
+    =id: ztf_partnership
+  - name: ZTF Caltech
+    altdata:
+      collection: ZTF_alerts
+      selector: [1, 2, 3]
```
dannygoldstein, line 29: See comment on partnership stream

data/db_demo.yaml
```diff
@@ -12,6 +12,23 @@ user:
     roles:
       - View only
 
+streams:
+  - name: ZTF Public
+    altdata:
+      collection: ZTF_alerts
+      selector: [1]
```
dannygoldstein, line 19: Should this be a scalar?

data/db_demo.yaml
```diff
@@ -30,9 +47,13 @@ groups/=program_A/users:
...
     admin: false
 
+groups/=program_A/streams:
+  - stream_id: =ztf_partnership
+
 filters:
-  - query_string: sample_query_string
+  - name: Yellow transients
     group_id: =program_A
+    stream_id: =ztf_partnership
```
dannygoldstein, line 56: `stream_ids`?

data/db_demo.yaml
```diff
@@ -30,9 +47,13 @@ groups/=program_A/users:
   - username: fulluser@cesium-ml.org
     admin: false
 
+groups/=program_A/streams:
+  - stream_id: =ztf_partnership
+
 filters:
-  - query_string: sample_query_string
+  - name: Yellow transients
```
dannygoldstein, line 54: Should this be versioned?

skyportal/handlers/api/filter.py
```diff
@@ -46,6 +47,14 @@ def get(self, filter_id=None):
             f = Filter.get_if_owned_by(filter_id, self.current_user)
             if f is None:
                 return self.error("Invalid filter ID.")
+            # get stream:
+            stream = (
```
dannygoldstein, line 51:
```suggestion
            stream = DBSession().query(Stream).get(f.stream_id)
```

skyportal/handlers/api/group.py
```diff
@@ -93,6 +93,27 @@ def get(self, group_id=None):
                 # Do not include User.groups to avoid circular reference
                 group['users'] = [{'id': user.id, 'username': user.username}
                                   for user in group['users']]
+                # grab streams:
+                group_streams = (
```
dannygoldstein, line 97:
```suggestion
                streams = DBSession().query(Stream).join(GroupStream).filter(GroupStream.group_id == group_id).all()
                group['streams'] = streams
```

skyportal/handlers/api/group.py
```diff
@@ -93,6 +93,27 @@ def get(self, group_id=None):
...
+                streams = (
+                    DBSession().query(Stream)
+                        .filter(Stream.id.in_(stream_ids))
+                        .all()
+                )
+                group['streams'] = streams
+                # grab filters:
+                filters = (
+                    DBSession().query(Filter)
+                        .filter(Filter.group_id.in_([group_id]))
```
dannygoldstein, line 112:
```suggestion
                        .filter(Filter.group_id == group_id)
```

skyportal/handlers/api/group.py
```diff
@@ -358,3 +379,113 @@ def delete(self, group_id, username):
...
+        stream = Stream.query.filter(Stream.id == stream_id).first()
+        if stream is None:
+            return self.error(
+                "Specified stream_id does not exist."
+            )
+        else:
+            # Add new GroupStream
+            gs = GroupStream.query.filter(
+                GroupStream.group_id == group_id
+            ).filter(
```
dannygoldstein, line 437:
```suggestion
           ,
```

skyportal/handlers/api/group.py
```diff
@@ -358,3 +379,113 @@ def delete(self, group_id, username):
...
+        group = Group.query.get(group_id)
+        stream = Stream.query.filter(Stream.id == int(stream_id)).first()
+        stream_id = stream.id
+        if stream is None:
+            return self.error(
+                "Specified stream_id does not exist."
+            )
+        else:
+            if group.single_user_group:
+                return self.error("Cannot delete streams from single user groups.")
```
dannygoldstein, line 485: We should support this, for when people change institutions.

skyportal/handlers/api/stream.py
```diff
@@ -0,0 +1,168 @@
...
+            - in: path
+              name: filter_id
+              required: true
+              schema:
+                type: integer
+          responses:
+            200:
+              content:
+                application/json:
+                  schema: SingleFilter
```
dannygoldstein, line 28: Wrong return type

skyportal/handlers/api/stream.py
```diff
@@ -0,0 +1,168 @@
...
+              content:
+                application/json:
+                  schema: Error
+        multiple:
+          description: Retrieve all streams
+          responses:
+            200:
+              content:
+                application/json:
+                  schema: ArrayOfFilters
```
dannygoldstein, line 39: Check return type

skyportal/handlers/api/stream.py
```diff
@@ -0,0 +1,168 @@
...
+from baselayer.app.access import auth_or_token, permissions
+from ..base import BaseHandler
+from ...models import (
+    DBSession,
+    Stream,
+)
+
+
+class StreamHandler(BaseHandler):
+    @auth_or_token
```
dannygoldstein, line 12: What are the use cases for exposing this endpoint to a general user?

skyportal/handlers/api/stream.py
```diff
@@ -0,0 +1,168 @@
...
+            return self.error(
+                "Invalid/missing parameters: " f"{e.normalized_messages()}"
+            )
+        DBSession().add(stream)
+        DBSession().commit()
+
+        return self.success(data={"id": stream.id})
+
+    @permissions(["System admin"])
+    def patch(self, stream_id):
```
dannygoldstein, line 107: We have been using PUT for partial updates, although we should be using PATCH according to the HTTP standards. 

@stefanv What do you think is the appropriate path here? Use PUT for now and then open another PR changing all the PUTs to PATCHes?

skyportal/models.py
```diff
@@ -300,9 +316,11 @@ def altitude(self, telescope, time):
 
 
 class Filter(Base):
-    query_string = sa.Column(sa.String, nullable=False, unique=False)
+    name = sa.Column(sa.String, nullable=False, unique=False)
+    stream_id = sa.Column(sa.ForeignKey("streams.id"), nullable=False)
+    stream = relationship("Stream", foreign_keys=[stream_id], back_populates="filters")
     group_id = sa.Column(sa.ForeignKey("groups.id"), nullable=False)
```
dannygoldstein, line 322:
```suggestion
    group_id = sa.Column(sa.ForeignKey("groups.id"), nullable=False, ondelete='CASCADE')
```

skyportal/models.py
```diff
@@ -96,23 +98,37 @@ class Group(Base):
...
                           passive_deletes=True)
+    users = relationship('User', secondary='stream_users',
+                         back_populates='streams',
+                         passive_deletes=True)
+    filters = relationship('Filter',
+                           back_populates='stream',
+                           passive_deletes=True)
+
 
+GroupStream = join_model('group_streams', Group, Stream)
```
dannygoldstein, line 116: If a stream is deleted, will it delete all the associated users in that stream? We should ensure that doesn't happen...

skyportal/models.py
```diff
@@ -300,9 +316,11 @@ def altitude(self, telescope, time):
 
 
 class Filter(Base):
-    query_string = sa.Column(sa.String, nullable=False, unique=False)
+    name = sa.Column(sa.String, nullable=False, unique=False)
+    stream_id = sa.Column(sa.ForeignKey("streams.id"), nullable=False)
```
dannygoldstein, line 320: Shouldn't need this, as it is now specified by `group`, right? I.e., filters no longer store stream credentials.

skyportal/tests/api/test_groups.py
```diff
@@ -120,3 +120,59 @@ def test_public_group(view_only_token):
     assert status == 200
     assert response["status"] == "success"
     int(response["data"]["id"])
+
```
dannygoldstein, line 123: Can you add the following tests?

   

- [ ] Grant a stream to a user, then delete the stream, and check that the user still exists, and has access to their original streams, minus the new one

- [ ]  Grant a stream to a group, then delete the stream, and check that the group still exists, and has access to its original streams, minus the new one

- [ ] Attach a filter to a group, then delete the group, and check that the filter has also been deleted in a cascade.

skyportal/tests/conftest.py
```diff
@@ -48,8 +54,8 @@ def public_group2():
 
 
 @pytest.fixture()
-def public_filter(public_group):
-    return FilterFactory(group=public_group)
+def public_filter(public_group, public_stream):
+    return FilterFactory(group=public_group, stream=public_stream)
```
dannygoldstein, line 58: stream should no longer be needed here

skyportal/tests/fixtures.py
```diff
@@ -75,17 +75,28 @@ class Meta(BaseMeta):
     observed_at = datetime.datetime.now()
 
 
+class StreamFactory(factory.alchemy.SQLAlchemyModelFactory):
+    class Meta(BaseMeta):
+        model = Stream
+    name = factory.LazyFunction(lambda: str(uuid.uuid4()))
+    users = []
+    groups = []
+    filters = []
```
dannygoldstein, line 84: filters should no longer be needed here

data/db_demo.yaml
dmitryduev: That's why I called it a "selector". This is then used on K as in `{"candidate.programid": {"$in": <stream["altdata"]["selector"]>}}`, so that a filter defined on this stream will also get executed on `programid=1` data.

data/db_demo.yaml
dmitryduev: Again, it's meant to be used with a `$in` operator. I can add an extra type check, but since it anyways lives in `altdata`, would prefer not to bother :)

data/db_demo.yaml
dmitryduev: Ah, I see what you mean.. To my taste, it's easer to define, e.g. the Partnership stream, as Public-only+Partnership-only data as opposed to Partnership-only, so that a filter is associated with a single stream (and a single group) -- this also simplifies the book-keeping on the K side a bit. After all, alerts with programid=2 come with programid=1 stuff in prv_candidates.

data/db_demo.yaml
dmitryduev: I wouldn't, but I tend to generally not care about such things :-) Are we doing that in any other situations?

dannygoldstein: I thought we were keeping track of multiple filter versions. Is that only happening on the K side or should we do that in SP as well?

data/db_demo.yaml
dannygoldstein: This makes me think - are we going to allow streams with data from multiple instruments? E.g., a ZTF PID=2  + Gattini stream?

skyportal/models.py
```diff
@@ -96,23 +98,37 @@ class Group(Base):
 
 class Stream(Base):
     name = sa.Column(sa.String, unique=True, nullable=False)
-    url = sa.Column(sa.String, unique=True, nullable=False)
-    username = sa.Column(sa.String)
-    password = sa.Column(sa.String)
+    altdata = sa.Column(JSONB, nullable=True,
```
dannygoldstein, line 101: `altdata` doesn't seem like the right place for this. Should we also be keeping track of what instrument a stream is connected to, so we can prevent users from implementing filters that run on streams from multiple instruments / collections, if that is not supported on K?

data/db_demo.yaml
dmitryduev: In my opinion, no, it would significantly complicate the alert filtering logic (most certainly on the K side). 
This is what the groups are for (among other things): they may have access to multiple streams from different discovery engines and you would define dedicated filters on each stream.

data/db_demo.yaml
dmitryduev: Only on the K side, since that is where the action is happening -- I want to separate the concerns as much as possible and also reduce the amount of duplicated info to a bare minimum. SP really only stores `filter_id` and the corresponding `group_id`, and that is the only piece of information that is duplicated on K. That way, the AMPEL folks, for example, won't have to bother about filter versioning on the SP side either -- they would just create a filter and then post stuff to /candidates with the `filter_id`.

skyportal/handlers/api/filter.py
dmitryduev: Changed.

skyportal/handlers/api/group.py
dmitryduev: Nice, changed!

skyportal/handlers/api/group.py
dmitryduev: True, removed.

skyportal/handlers/api/stream.py
dmitryduev: Damn, fixed.

skyportal/handlers/api/stream.py
dmitryduev: Fixed

skyportal/handlers/api/stream.py
dmitryduev: Couldn't think of anything specific, but thought there might be, and there isn't much criminal in getting this information. But ok, changed for `@permissions(["System admin"])`.

skyportal/handlers/api/stream.py
dmitryduev: We use PATCH for filter and user preferences updates, so there are precedents :)

skyportal/models.py
dmitryduev: Nice catch, fixed

skyportal/models.py
dmitryduev: A group may have access to multiple streams and you define a filter in the group on a stream, so I think it should be there.

skyportal/tests/conftest.py
dmitryduev: See my comment above

skyportal/tests/fixtures.py
dmitryduev: See me comment above

skyportal/models.py
dmitryduev: I originally had two columns instead, `collection` and `selector`, but decided to move them to `altdata` (thanks to Josh for the suggestion!) to keep it general enough to be included in general SP. 

The users won't be able to do that: a filter is defined in a group for a single stream, and only the System admin can create new streams.

skyportal/tests/api/test_groups.py
dmitryduev: Added

skyportal/models.py
dannygoldstein: Also make sure to index all foreign keys with `index=True`, to ensure fast joins.

skyportal/models.py
dannygoldstein: In that case I think it makes more sense to enforce stream access on the filter level than the group level. Otherwise, you need to manually update each group when you remove a stream with, e.g., PID=4 (Gattini).

skyportal/handlers/api/group.py
dmitryduev: Fixed

skyportal/handlers/api/group.py
dmitryduev: Fixed in the commit, forgot to comment

skyportal/models.py
dmitryduev: Done.

skyportal/models.py
dmitryduev: This will not happen, `test_add_stream_to_single_user_group_delete_stream` in `test_groups.py` tests for that.

skyportal/models.py
dmitryduev: I don't see why you would delete a stream. But even if you do, the update will happen automatically, see `test_post_new_filter_delete_group_deletes_filter` and `test_post_new_filter_delete_stream_deletes_filter` in `test_groups.py`.

dmitryduev:
@dannygoldstein @acrellin all tests pass locally; on the last three commits, different frontend tests fail. Any wisdom?

dmitryduev:
@dannygoldstein @acrellin this time, `skyportal/tests/frontend/test_candidate_page.py` timed out and failed. Any suggestions?

acrellin:
> @dannygoldstein @acrellin this time, `skyportal/tests/frontend/test_candidate_page.py` timed out and failed. Any suggestions?

@dmitryduev go ahead and mark it flaky

dannygoldstein:
@dmitryduev see convo and resolution on #677 for current recommended
practice on flaky tests

data/db_demo.yaml
```diff
@@ -12,6 +12,23 @@ user:
     roles:
       - View only
 
+streams:
+  - name: ZTF Public
+    altdata:
+      collection: ZTF_alerts
```
dannygoldstein, line 18: What if I make a group that filters on streams 2 and 4 only. Will I need to make a new Stream object for 2+4 in addition to these three and grant it manually to each user and group? What if I had 2+1+4? Will I have to grant that to each user manually for them to be able to see a stream even if they have [1], and [2,4]? 

If so, I think that argues for having each stream map 1-to-1 to each PID, and granting those permissions individually to each user.

skyportal/handlers/api/group.py
```diff
@@ -358,3 +365,110 @@ def delete(self, group_id, username):
         self.push_all(action='skyportal/REFRESH_GROUP',
                       payload={'group_id': int(group_id)})
         return self.success()
+
+
+class GroupStreamHandler(BaseHandler):
+    @permissions(['System admin'])
+    def post(self, group_id, *ignored_args):
```
dannygoldstein, line 372: Do we need `*ignored_args`?

dannygoldstein:
Will re-review once green

dannygoldstein:
Also, echoing @acrellin, @dmitryduev can you please fill out a more detailed description of the PR in the original post?

data/db_demo.yaml
```diff
@@ -12,6 +12,23 @@ user:
...
+      selector: [1]
+    =id: ztf_public
+  - name: ZTF Partnership
```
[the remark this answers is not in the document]
dmitryduev: Sure, it's more explicit like this, added.

data/db_demo.yaml
```diff
@@ -12,6 +12,23 @@ user:
...
+    altdata:
+      collection: ZTF_alerts
+      selector: [1, 2]
+    =id: ztf_partnership
+  - name: ZTF Caltech
```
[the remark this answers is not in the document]
dmitryduev: Same as above, thanks

skyportal/handlers/api/group.py
dmitryduev: Yes, otherwise you'd get errors like `Error in `/api/groups/8/streams`: post() takes 2 positional arguments but 3 were given` since stream id would be None see endpoint def in `app_server.py` -- `(r'/api/groups(/[0-9]+)/streams(/[0-9]+)?', GroupStreamHandler),`

skyportal/models.py
```diff
@@ -101,23 +104,37 @@ class Group(Base):
...
+                            "`{'collection': 'ZTF_alerts', selector: [1, 2]}`")
 
-    groups = relationship('Group', secondary='stream_groups',
+    groups = relationship('Group', secondary='group_streams',
                           back_populates='streams',
                           passive_deletes=True)
+    users = relationship('User', secondary='stream_users',
+                         back_populates='streams',
+                         passive_deletes=True)
+    filters = relationship('Filter',
```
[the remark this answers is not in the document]
dmitryduev: Thanks for bringing this up! It made me think about the ACL stuff in `api/filters.py` and I rewrote it so that now all the necessary checks are in place (I think -- needs review; updated the tests accordingly).

I strongly prefer the current scheme and believe that switching to the proposed "simplified" scheme would actually complicate things in practice, and is _a lot_ of work. If you assign streams to filters then the user who does it must have the corresponding SU permissions (remember, this is a very "privileged" task), and this will have to be done for every user filter, and we will have potentially multiple filters per program. Whereas in the current scheme, as long as you are a group admin, you can create as many filters as you like on the streams the group has access to, without asking SUs for permissions -- I think it is much more user-friendly.

I also don't quite see what needs to be "manually kept in sync". It is forbidden to change both `group_id` and `stream_id` for a filter; deleting a stream from a group deletes corresponding filters as well, see `test_post_new_filter_delete_stream_deletes_filter` in `test_groups.py`.

data/db_demo.yaml
dmitryduev: I feel that this is a valid scenario abstractly speaking, but an artificial one in practice. Concretely, what would this 2+4 mean? In practice, for F we will have ztf_public, ztf_public+partnership, ztf_public+partnership+caltech, and gattini (and in the future, LSST, and potentially other surveys). So I do not quite understand what kinds of combinations you are talking about (and in general, this all PID stuff essentially only applies to ZTF). However, the implementation here in SP is quite general and none of the `altdata` stuff is required.

Next, these streams really point at a sink on the K side (even though eventually, SP may get its own stream listeners, but it is a completely different story), this is why I "aggregate" them this way, otherwise things will be nice, abstract, and clean on the SP side, but that would add unnecessary complexity on the K side of F.

Finally, yes, when a new stream is added to the system, a SU must add it individual users first, that's the scheme we converged on. Only when/if all users of a group have access to a particular (new) stream, it can be added to the group.

skyportal/handlers/api/group.py
dmitryduev: It's similar to `GroupUserHandler` in this respect.

dmitryduev:
:tada:

PR #652, opened 2020-08-05
Add ability to share current URI via QR code

stefanv:
Not-entirely-joking idea to allow users to quickly load a page they're viewing in their browser on their phone.

qr

Who can tell my server URL from the gif? :)


profjsb:
I like it. Not sure what the use cases are you had in mind, but it's a nice piece of (new functionality) sugar, which we need more of. Your server is `{ url }`, right? ;)

stefanv:
`{ url }` indeed :)

My thinking was that someone would be working on SkyPortal, then would need to run to a telescope or to catch a bus.  They'd quickly whip up the QR code and scan it, rather than having to email or text themselves the URL.

stefanv:
(Also, the sharing component is generic, so can be used on any page.)

dannygoldstein:
This is super awesome.

profjsb:
It would be useful to have a little URL and/or statement about what this is sharing. Eg.
"Link to ZTF14gqr<br />https://...source/ZTF16gqr"

I'd worry as an end user what this QR code is actually sharing.

PR #653, opened 2020-08-05
Add pre-commit checking

stefanv:


.flake8
```diff
@@ -0,0 +1,16 @@
...
+# https://github.com/PyCQA/flake8-bugbear
+#
+# for error codes.  And
+#
+# https://flake8.pycqa.org/en/latest/user/violations.html#selecting-violations-with-flake8
+#
+# for error classes selected below.
+
+[flake8]
+max-line-length = 80
```
acrellin, line 14: Can this be 88?

stefanv: Yes, this is a bit strange, but what is happening is that it is setting the line length limit to 80, and then ignoring that rule later on.  This is what the black docs recommend, IIRC.

acrellin: Ah, I see. Great, thanks!

PR #657, opened 2020-08-06
Fix bugs in Obj.last_detected, Photometry.snr and return last_detected in Source and Candidate handlers

### LINKED ISSUE #558: Fix Obj.last_detected

acrellin:
Closes #558

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -195,8 +195,8 @@ class Obj(Base, ha.Point):
 
     @hybrid_property
     def last_detected(self):
-        detecteds = [phot.iso for phot in self.photometry if phot.snr and phot.snr > 5]
-        return max(detecteds) if detecteds else None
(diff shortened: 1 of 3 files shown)


skyportal/models.py
```diff
@@ -699,7 +701,7 @@ def iso(cls):
 
     @hybrid_property
     def snr(self):
-        return self.flux / self.fluxerr
+        return self.flux / self.fluxerr if self.flux and self.fluxerr else None
```
dannygoldstein, line 704:
```suggestion
        return self.flux / self.fluxerr if self.flux and self.fluxerr else None
@snr.expression
def snr(self):
    return self.flux / self.fluxerr      
```

dannygoldstein: This ensures that the hybrid property can properly compile to SQL

PR #661, opened 2020-08-06
Render groups badges at top of source page

acrellin:
image

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -173,7 +173,7 @@ class SourceHandler(BaseHandler):
                 register_source_view(obj_id=obj_id,
                                      username_or_token_id=self.current_user.id,
                                      is_token=True)
-            s = Source.get_obj_if_owned_by(  # Returns Source.obj
+            s = Source.get_obj_if_owned_by(
                 obj_id, self.current_user,
                 options=[joinedload(Source.obj)
                          .joinedload(Obj.followup_requests)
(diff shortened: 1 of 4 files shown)


acrellin:
[ch1098]

skyportal/handlers/api/source.py
```diff
@@ -192,6 +192,13 @@ def get(self, obj_id=None):
             s.classifications = s.get_classifications_owned_by(self.current_user)
             source_info = s.to_dict()
             source_info["last_detected"] = s.last_detected
+            source_info["groups"] = Group.query.filter(
+                Group.id.in_(
+                    DBSession().query(Source.group_id).filter(
+                        Source.obj_id == obj_id
+                    )
+                )
+            ).all()
```
dannygoldstein, line 201:
```suggestion
            ).filter(Group.id.in_(self.current_user.accessible_groups)).all()
```

acrellin: Really? People can't even see which other groups it belongs to if they're not a member of those??

acrellin: @dannygoldstein Done

profjsb:
Also, we dont have a restriction on group name length. So you might want to cut off the badge size to <= 10 characters (ie. "'Fabulous Green Transie...")

acrellin:
> Looks good. The badge being there or not there (following group accessibility) should be added to the tests. Also, have you looked at the UI when the source belongs to many groups (like dozens?) ... will they wrap to another line? Will you show in a carousel?

@profjsb I've updated the tests to check that the badges render. In the case of many groups, they'll just wrap like this: 
image

Do you think there's a need to contain it in a scrollable div with max height? Not sure how many groups a typical source will be associated with...

acrellin:
@profjsb I've also truncated long group names in the badges

PR #662, opened 2020-08-06
If initialState.allowed_classes is null, return loading message in ClassificationForm

acrellin:
@profjsb @stefanv there seems to be an underlying data issue as well that this PR does not address: in `ClassificationForm`, even when `taxonomyList[0]` is not null, the `allowed_classes` attribute that the front-end expects is not present (after running `make load_demo_data`). Just a heads up. I can break that out into an issue.

### DIFF
--- a/skyportal/tests/frontend/test_sources.py
+++ b/skyportal/tests/frontend/test_sources.py
@@ -16,42 +16,58 @@ def test_public_source_page(driver, user, public_source):
     driver.get(f"/become_user/{user.id}")  # TODO decorator/context manager?
     driver.get(f"/source/{public_source.id}")
     driver.wait_for_xpath(f'//div[text()="{public_source.id}"]')
-    driver.wait_for_xpath('//label[contains(text(), "band")]')  # TODO how to check plot?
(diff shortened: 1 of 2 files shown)


stefanv:
@acrellin I think this is due to my unfamiliarity with the reducer logic.  Shouldn't the allowed classes always be set to something?

Once the length of the list is >0, the allowed classes should be computable immediately; so why can't we render the component then?

acrellin:
@stefanv my initial comment explains it: the elements of the taxonomy list fetched from the backend are expected to have an allowed_classes attribute, but it is missing.

stefanv:
@acrellin We removed that on purpose, so anything that expects that should be changed. The allowed classes are now computed on the frontend.

acrellin:
> @acrellin We removed that on purpose, so anything that expects that should be changed. The allowed classes are now computed on the frontend.

Okay -- @profjsb can you update the ClassificationForm component accordingly, then?

stefanv:
Closing since #674 was merged; thanks for this PR, Ari.

PR #664, opened 2020-08-07
Update pre-commit config to include prettier

stefanv:
Avoid having to comment on JavaScript style in review.

This will soon be enabled by default.


acrellin:
@stefanv will this just do a check or will it run `prettier --write`?

stefanv:
I think if it has to make changes it fails, but saves the changes to disk.  Can you confirm?

stefanv:
We should also remove eslint.

.pre-commit-config.yaml
```diff
@@ -1,19 +1,25 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.5.0
     hooks:
       - id: check-yaml
       - id: end-of-file-fixer
       - id: trailing-whitespace
--   repo: https://github.com/python/black
+  - repo: https://github.com/python/black
```
profjsb, line 8: Wasn't there some discussion about ignoring single- vs double- quoted strings? Should we be installing a dot file with explicit checking preferences for `black`?

stefanv: There was, in this is how you do it:

https://github.com/napari/napari/blob/master/pyproject.toml#L12

But, you cannot configure black to enforce single quotes, just to ignore them.  In the end, I wasn't sure it was worth the hassle.  Also, PrettierJS uses double quotes, so it makes things consistent.

If you'd like us to add that option, I'm happy to add the `.pyproject.toml`.

stefanv:
@acrellin Thinking about eslint, it may be that it catches unused imports and other similar errors through its plugins that PrettierJS doesn't, so let's keep it in, with https://github.com/prettier/eslint-plugin-prettier

stefanv:
OK, that's done now.

acrellin:
@stefanv great!

stefanv:
Added black configuration to ignore quoting style.

profjsb:
As discussed, we should try these pre-commits in `baselayer` first.

stefanv:
For now, I've disabled flake8, since we have a lot of work to do there.

PR #667, opened 2020-08-07
Add GitHub Actions CI

stefanv:


.github/workflows/test.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Lint
+        run: |
+          pip install pre-commit
+          if pre-commit run --from-ref origin/master --to-ref HEAD; then
+              echo "Linting errored; this will be fatal sometime in the near future"
+          fi
+
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.8'
```
profjsb, line 30: We should also be checking 3.7 (I believe we require >=3.7).

.github/workflows/test.yml
```diff
@@ -0,0 +1,105 @@
...
+on: [push]
+
+jobs:
+  test:
+    name: Test SkyPortal
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:11.8
```
profjsb, line 12: I thought we're on postgres 12.x -- please confirm.

package.json
```diff
@@ -11,21 +11,24 @@
...
-    "material-ui-phone-number": "2.2.6",
     "bokehjs": "0.12.9",
     "check-dependencies": "1.1.0",
     "clsx": "1.1.1",
+    "core-js": "3.6.5",
     "css-loader": "4.2.0",
     "dayjs": "1.8.31",
     "exports-loader": "0.7.0",
     "imports-loader": "0.8.0",
+    "material-ui-phone-number": "2.2.6",
```
acrellin, line 22: Should these changes be in here?

.github/workflows/test.yml
stefanv: Switched to 3.7 to test our minimal version of Python.

.github/workflows/test.yml
stefanv: I'm now using postgres latest

package.json
stefanv: Yes, without these extra side dependencies, npm installs fail.

stefanv:
I added the relevant CI keys.

doc/dev.md
```diff
@@ -79,3 +79,14 @@ We use ESLint to ensure that our JavaScript & JSX code is consistent and conform
 ## Docker images
 
 Run `make docker-images` to build and push to Docker hub.
+
+## Debugging GitHub Actions
+
+You can insert the following step to debug your workflow:
+
+```yaml
```
dannygoldstein, line 87: Love this

PR #668, opened 2020-08-07
First commit to showcase the kind of model changes for localizations

mcoughlin:
@dannygoldstein @stefanv @acrellin @dmitryduev  This is to highlight how we do things in the ToO Marshal (not really a pull request, but ```make load_gcn_data``` works fine). I figure that it would be good to start here and then can think about front end things.


mcoughlin:
event_list

lpsinger:
Honestly, I would like to nail down the backend stuff for storing and querying the multi-resolution probability maps in the database before migrating stuff over from the ToO marshal. I could use a hand with that.

PR #671, opened 2020-08-07
index all foreign keys to improve join performance

dannygoldstein:
There were some ForeignKeys in models that were not indexed. This should speed up joins against those relationships.


stefanv:
@acrellin Think we can remove all sleeps from the test suite with the function you introduce for waiting for elements to become clickable?

PR #672, opened 2020-08-07
remove GroupTelescopes in preparation for Allocations model

dannygoldstein:
In our current data model, `GroupTelescope`s are used to determine a user's access to instruments and telescopes. After a discussion on how access to instruments and telescopes is implemented, @acrellin and I decided that a better model would be to mediate facility access at the instrument level. Such a model more closely mirrors the astronomical proposal writing process: people apply for time on specific instruments and are granted time on them, as opposed to time on telescopes. In ch1158, @acrellin requested that the GroupTelescope model be removed and replaced with an Allocation model, which tracks InstrumentID, GroupID, and some other columns (PI, Proposal ID, Allocation ID [PK])

This PR facilitates the removal of `GroupTelescope`. A future PR will implement `Allocation`.

[ch1158]

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -44,9 +44,6 @@ telescope:
     diameter: 1.5
     robotic: true
     skycam_link: http://bianca.palomar.caltech.edu/images/allsky/AllSkyCurrentImage.JPG
-    group_ids:
-      - =public_group_id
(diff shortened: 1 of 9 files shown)


dannygoldstein:
[ch1172]

PR #673, opened 2020-08-08
Afix ProfileDropdown to the top of page

profjsb:
This PR adds `top: "1rem"` to the styling of the `ProfileDropdown` component to afix it to the top of the page. Before this PR, it was only `position: fixed` to the right of the page. When adding a new component to the header (quick searchbar, coming soon!) I found that we needed both the top and right glue for proper formatting.

Note: the styling is added inline in `HeaderContent`, consistent with how it is currently done. It, like many other inline styling commands, ought to be moved using the makeStyles paradigm (or into css files).

Also: found a flaky test `test_frontpage.py::test_jump_to_page_invalid_values` which has been marked accordingly.

### DIFF
--- a/skyportal/tests/frontend/test_frontpage.py
+++ b/skyportal/tests/frontend/test_frontpage.py
@@ -1,6 +1,8 @@
 import uuid
 import time
 
+import pytest
+
 from skyportal.tests import api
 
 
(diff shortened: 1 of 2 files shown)


skyportal/tests/frontend/test_frontpage.py
```diff
@@ -129,7 +129,7 @@ def test_source_filtering_and_pagination(driver, user, public_group, upload_data
         time.sleep(1)
         assert not next_button.is_enabled()
 
-
+@pytest.mark.flaky(reruns=2)
```
acrellin, line 132: @profjsb this will fail if pytest isn't imported above

profjsb: Fixed

PR #675, opened 2020-08-08
Quick Search Bar for Sources

profjsb:
Kapture 2020-08-08 at 11 08 35

This PR adds a quick search bar for sources. Rather than having endusers type in the full source name to an address bar to jump to a source, this quick search bar will allow them to find and navigate to that source more quickly, from within the UI. Searches are debounced to 250ms to avoid making unnecessarily too many calls rapidly to the backend. This debouncing can be seen in the screenshot above, as the network does not get called when a long string is entered quickly. 

A maximum of 25 results are returned to avoid large data transfers from the backend to the frontend.

Future:
  - allow searches for richer content, like comment strings or classification
  - memoize the searches so that we do not rely on the browser to cache results

### DIFF
--- a/static/js/components/HeaderContent.jsx
+++ b/static/js/components/HeaderContent.jsx
@@ -1,19 +1,23 @@
 import React from "react";
 import { Link } from "react-router-dom";
 
+import Box from "@material-ui/core/Box";
 import Responsive from "./Responsive";
 import ProfileDropdown from "./ProfileDropdown";
 import Logo from "./Logo";
+import QuickSearchBar from "./QuickSearchBar";
 
 import styles from "./Main.css";
 
 const HeaderContent = () => (
   <div className={styles.topBannerContent}>
-    <div style={{ display: "inline-block", float: "left" }}>
+    <div style={{ display: "inline-flex", flexDirection: "row" }}>
       <Logo className={styles.logo} />
       <Link className={styles.title} to="/">
-        SkyPortal ∝
+        SkyPortal
       </Link>
+      <Box p={2} />
+      <QuickSearchBar id="search" />
     </div>
     <div style={{ position: "fixed", right: "1rem", top: "1rem" }}>
       <Responsive desktopElement={ProfileDropdown} />
--- a/static/js/components/QuickSearchBar.jsx
+++ b/static/js/components/QuickSearchBar.jsx
@@ -34,7 +34,7 @@ function useDebouncer(value, delay) {
 
 const QuickSearchBar = () => {
   const dispatch = useDispatch();
-  const [options, setOptions] = React.useState([]);
+  const [options, setOptions] = useState([]);
   const [value, setValue] = useState(null);
   const [inputValue, setInputValue] = useState("");
   const [loading, setLoading] = useState(false);
@@ -121,6 +121,7 @@ const QuickSearchBar = () => {
       selectOnFocus
       limitTags={15}
       value={value}
+      popupIcon={null}
       renderOption={(option) => {
         const v = `/source/${option}`;
         return (
@@ -135,6 +136,7 @@ const QuickSearchBar = () => {
           {...params}
           className={classes.root}
           variant="outlined"
+          label="Source ID Search"
           InputProps={{
             ...params.InputProps,
             startAdornment: (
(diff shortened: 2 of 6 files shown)


static/js/components/ClassificationForm.jsx
```diff
@@ -153,7 +153,7 @@ const ClassificationForm = ({ obj_id, taxonomyList }) => {
           </Select>
           <div style={{ display: state.class_select_enabled ? "block" : "none" }}>
             <Autocomplete
-              options={state.allowed_classes}
```
stefanv, line 156: Leftover from other PR.

static/js/components/HeaderContent.jsx
```diff
@@ -1,21 +1,25 @@
...
       <Logo className={styles.logo} />
       <Link className={styles.title} to="/">
-        SkyPortal ∝
+        SkyPortal
       </Link>
+      <Box p={2} />
+      <QuickSearchBar id="search" />
     </div>
-    <div style={{ position: "fixed", right: "1em" }}>
+    <div style={{ position: "fixed", right: "1em", top: "1em" }}>
```
stefanv, line 22: This one too.

static/js/components/QuickSearchBar.jsx
```diff
@@ -0,0 +1,135 @@
...
+      const matchingSources = await response.data.sources;
+
+      let newOptions = [];
+
+      if (matchingSources) {
+        const rez = Object.keys(matchingSources).map((key) => matchingSources[key].id);
+        newOptions = [...newOptions, ...rez];
+      }
+      setOptions(newOptions);
+      return undefined;
```
stefanv, line 68: Is this to make the linter happy?

profjsb: yes. An an anonymous async function still needs to return something. I added this to make the linter happy.

static/js/components/HeaderContent.jsx
profjsb: Would you like me to fix the right: 1em as well?

profjsb:
> We haven't used history before, so could you explain a bit how it works? I see the pushes, but where do you pull the history back up? Is this the caching mechanism?

This is to allows us to dynamically navigate to another page. The matching options are links so if someone clicks on them then that works too. But if they click off the letters then this results in a "select-option" `reason` in the callback. It looks like `push.history` is the cleanest way to navigate as such.

> Also, it looks like the debouncer is used whether there are matches or not. But I think this is the part of the logic I wasn't following perfectly, so thanks for helping me get a grasp on it.

Yes, the debouncer just captures the raw inputValue changes and then updates a debounced version for useEffect.

static/js/components/QuickSearchBar.jsx
stefanv: Ok, you may also use single line eslint disables to keep it from complaining.

stefanv:
Re the debouncer, would it make sense to complete on cached results first and, if that doesn't yield anything, to search again? Perhaps that's how it works already.

I recommend styling the backdrop a bit lighter to make it more readable.

static/js/components/HeaderContent.jsx
profjsb: Already fixed in closed PR #673

static/js/components/QuickSearchBar.jsx
profjsb: I am ok with complying with this expectation. Leaving in the return statement.

profjsb:
> Re the debouncer, would it make sense to complete on cached results first and, if that doesn't yield anything, to search again? Perhaps that's how it works already.
> 

I'm now memoizing the API calls.

> I recommend styling the backdrop a bit lighter to make it more readable.

Done.

stefanv:
I pushed a fix for the linter error.  @profjsb You can install the pre-commit checks with `pip install pre-commit && pre-commit install`.

acrellin:
@profjsb I removed the dropdown caret icon that did nothing and added a small label (it's not obvious what one can search for without that): 
image

acrellin:
@profjsb I also added tests

profjsb:
Thanks @acrellin and @stefanv!

PR #687, opened 2020-08-10
Make pytest output easier to follow

stefanv:


acrellin:
@stefanv does it automatically change the output, or do we need to pass a flag into pytest?

PR #689, opened 2020-08-10
Timestamp tests

stefanv:


skyportal/tests/conftest.py
```diff
@@ -25,6 +27,8 @@
 )
 from skyportal.model_util import create_token
 from skyportal.models import DBSession, Source, Candidate, Role
+
+from psycopg2 import OperationalError
```
acrellin, line 31: Where is this being used?

acrellin, line 31:
```suggestion
```

stefanv: No idea, I didn't add that.

PR #690, opened 2020-08-10
only download IERS data if latest data is >=14 days old

dannygoldstein:
This PR adds a check to `conftest.py`, telling it to only download the latest IERS data if the iers data in disk cache (if any) is >= 14 days old. If there is no cached IERS data then it is downloaded.


stefanv:
Thanks @dannygoldstein, very elegant

Issue #593, opened 2020-07-22
Upload photometry page fails to render if urlFor is clicked before source page finishes rendering


dannygoldstein:
If one goes to 

http://localhost:5000/source/16fil

Then clicks the "upload photometry" button before the page finishes loading, then the upload photometry page at

http://localhost:5000/upload_photometry/16fil

fails with 

<img width="1451" alt="Screen Shot 2020-07-22 at 12 39 29 AM" src="https://user-images.githubusercontent.com/2769632/88148687-d2890100-cbb3-11ea-8e37-0a3a7e48e495.png">

dannygoldstein:
It works fine, though, if you refresh the url for the upload_photometry page

stefanv:
Is this before the page has finished loading, or before skyportal his finished starting up? Latter is expected, and gives that error. 

We can probably do a better check in some places, but api calls cannot succeed before the API is served.

dannygoldstein:
The former.

________________________________

PR #705, opened 2020-08-12
Fix upload photometry page initial render bug

acrellin:
Closes https://github.com/skyportal/skyportal/issues/593


acrellin:
[ch1159]

PR #707, opened 2020-08-12
Fetch and display single user groups in data sharing component; update handler & action creator

acrellin:
share_data_user_groups

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 from sqlalchemy.orm import joinedload
 from marshmallow.exceptions import ValidationError
 from baselayer.app.access import permissions, auth_or_token
@@ -60,6 +59,13 @@ class GroupHandler(BaseHandler):
               schema:
                 type: string
               description: Fetch by name (exact match)
+            - in: query
+              name: includeSingleUserGroups
+              schema:
+                type: boolean
+              description: |
+                Bool indicating whether to include single user groups.
+                Defaults to false.
           responses:
             200:
               content:
@@ -82,8 +88,9 @@ class GroupHandler(BaseHandler):
                                 items:
                                   $ref: '#/components/schemas/Group'
                                 description: |
-                                  List of all groups user has access to, optionally
-                                  including single user groups if specified in request.
+                                  List of all groups, optionally including single user
+                                  groups if query parameter `includeSingleUserGroups` is
+                                  `true`.
             400:
               content:
                 application/json:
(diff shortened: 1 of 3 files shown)


acrellin:
[ch1189]

skyportal/handlers/api/group.py
```diff
@@ -80,7 +81,9 @@ def get(self, group_id=None):
                                 type: array
                                 items:
                                   $ref: '#/components/schemas/Group'
-                                description: List of all groups if current user is Super admin, else None.
+                                description: |
+                                  List of all groups user has access to, optionally
+                                  including single user groups if specified in request.
```
stefanv, line 86: What does it mean to "specify in request"?

skyportal/handlers/api/group.py
```diff
@@ -138,22 +141,25 @@ def get(self, group_id=None):
...
+                for g in self.current_user.accessible_groups
+                if not g.single_user_group
             ]
-            if info["all_groups"]:
-                info["all_groups"] = [
-                    g for g in info["all_groups"] if not g.single_user_group
-                ]
+        else:
+            # `accessible_groups` already includes single user groups for sys admins
+            if "System admin" in self.current_user.permissions:
```
stefanv, line 154: If you use set logic to do the joins, as we do in some other place in SkyPortal, there is no need to handle these cases separately.

skyportal/handlers/api/group.py
```diff
@@ -138,22 +141,25 @@ def get(self, group_id=None):
         include_single_user_groups = self.get_query_argument(
             "includeSingleUserGroups", False
```
stefanv, line 142: I don't see this in the docs

skyportal/handlers/api/group.py
```diff
@@ -138,22 +141,25 @@ def get(self, group_id=None):
...
         )
-        acls = [acl.id for acl in self.current_user.acls]
         info = {}
         info['user_groups'] = list(self.current_user.groups)
-        info['all_groups'] = (
-            Group.query.all()
-            if "System admin" in acls or "Manage groups" in acls
-            else None
-        )
         if (not include_single_user_groups) or (include_single_user_groups == "false"):
```
stefanv, line 146: False by default, need to check for true.

```
include_single_user_groups.lower() == 'true'
```

Also, are you going to remove single user group if the argument is set to false for the super user?

skyportal/handlers/api/group.py
acrellin: e.g. the parameter `includeSingleUserGroups` described above equals `true`

skyportal/handlers/api/group.py
acrellin: Ah, yes, I'll add that.

skyportal/handlers/api/group.py
acrellin: That's already being done on lines 147-150

skyportal/handlers/api/group.py
acrellin: using a single query won't work because accessible_groups doesn't include single user groups for non-admins. or are you referring to something else?

acrellin: if you mean just do the query in either case, then pass through `set` to remove dupicates; yes, I could, but thought it would be more efficient to not have to execute that query unnecessarily in the case of admins

dannygoldstein:
Reverting the grouphandler will break 599, as it relies on all users being able to see the names of groups observing runs are affiliated with, in assignments and observing run pages.

skyportal/handlers/api/group.py
stefanv: I don't see the logic to coerce the variable to true/false anywhere; can you point it out again?

skyportal/handlers/api/group.py
stefanv: OK, this is fine.

skyportal/handlers/api/group.py
acrellin: @acrellin if the parameter isn't provided, it defaults to `False`, so the above condition is `True` (`not include_single_user_groups`), and single user groups are removed. If the parameter is `True` or `"true"` (both truthy), the above evaluates to `False`, and single user groups are not removed. If the parameter is `"false"` or `False`, the above evaluates to `True`, and single user groups are removed.

skyportal/handlers/api/group.py
acrellin: Updated docs

skyportal/handlers/api/group.py
acrellin: Added

skyportal/handlers/api/group.py
acrellin: See updated logic (same condition check, though)

acrellin:
@stefanv @dannygoldstein ready for re-review

skyportal/handlers/api/group.py
stefanv: My concern was the case in which include_single_user_groups is the string "False", which evaluates to True.  A `.lower()` would fix that.

acrellin: Fixed. Thanks for the suggestion

PR #709, opened 2020-08-13
Add ErrorBoundary to handle component rendering errors

stefanv:
With an ErrorBoundary in place, SkyPortal should never just become a
blank white screen.  Instead, it will render an error message inside
the main panel (i.e. with sidebar and top bar).  If the error arises
inside of the sidebar or top bar, the error message will occupy the
whole screen.

The error message will disappear when a link is clicked or the URL
location is changed.

Whenever an error occurs, an API call is made to a new internal
endpoint, logging that error and the component stack to the app log.


stefanv:
I also wanted to comment on why I did not follow the style guide here: I am trying to make the ErrorBoundary as simple as possible, hence I used inline styles and only used minimal components.  Even DayJS can potentially be removed, if that gives us any problems whatsoever.

acrellin:
Very neat. Thanks @stefanv

PR #710, opened 2020-08-13
GA: merge branch with master, and pre-commit correct range

stefanv:


stefanv:
@dannygoldstein This should ensure that we always test against a merged branch.

PR #713, opened 2020-08-13
Headless testing

stefanv:


acrellin:
@stefanv where have you updated the test target for GA?

PR #715, opened 2020-08-14
Reskin News Feed and Comments

kmshin1397:
Includes addition of avatars to News Feed and Comments items

News Feed:
Screenshot 2020-08-13 175353
Screenshot 2020-08-13 175331

Comments:
comments_light
comments_dark


acrellin:
Beautiful, thanks Kyung!

stefanv:
Love it!

stefanv:
@Ari failure on Travis is delete group user.

profjsb:
This is great. One request and a few suggestions:

Request: please make the News Feed more compact, perhaps the height of the avatar. To do this, my suggestion is to remove the display of the name of the author (making that a hoverover atop the Avatar) which you can slot the comment itself into.

Suggestions:
 1) make the news feed scrollable (see Comments/Classification)
 2) put the name of the source first in the newsfeed and make it clickable. Eg. <Link ...>16fil</Fil>Birds are not real

skyportal/models.py
```diff
@@ -483,7 +484,33 @@ def get_obj_if_owned_by(obj_id, user_or_token, options=[]):
...
+                "username": author.username,
+                "first_name": author.first_name,
+                "last_name": author.last_name,
+                "gravatar_url": author.gravatar_url,
+            }
+        else:
+            username = "unknown user"
+            digest = md5(username.lower().encode('utf-8')).hexdigest()
+            # return a 404 status code if not found on gravatar
+            gravatar_url = f'https://www.gravatar.com/avatar/{digest}?d=404'
```
profjsb, line 505: Is there a way to leverage the baselayer method in the `User` model that does this? I worry about repeat functionality here.

skyportal/models.py
```diff
@@ -725,6 +755,36 @@ class Comment(Base):
...
```
profjsb, line 778: See comment above.

skyportal/models.py
kmshin1397: Good point. I managed to do it with this [commit](https://github.com/skyportal/skyportal/pull/715/commits/6d1265b375fdcfb4ff17f32861ce1c086b9a29c2).

kmshin1397:
New look based on feed back from @profjsb:

newsfeed_light

newsfeed_dark

(Also added scrolling to the comments list but otherwise comments look the same as above)

skyportal/handlers/api/news_feed.py
```diff
@@ -61,13 +61,52 @@ def fetch_newest(model):
...
+                author = User.query.filter(User.username == comment.author).first()
+                if author:
+                    comment.author_info = {
+                        "username": author.username,
+                        "first_name": author.first_name,
+                        "last_name": author.last_name,
+                        "gravatar_url": author.gravatar_url,
+                    }
+                else:
+                    unknown_user = User(username="unknown user")
```
stefanv, line 92: Could you explain what this section is for?

kmshin1397: As far as I can tell, the unknown user parts shouldn't really effect production environments - I put them in mainly to deal with test data. Whenever we generate test sources, they get test comments added using the CommentFactory in tests/fixtures.py. But these test comments aren't associated with any users (just given the author "test factory" but that's not a real user) so I was failing all the tests related to this since the API couldn't fetch the user info for those comments. But now that I'm reflecting on this, perhaps the proper solution is to add a User with username "test factory" as part of the base testing database instead?

kmshin1397: I thought about this some more and I think it would be fine to keep this in instead of adding a "test factory" user to the tests. I can imagine cases in which a user account is deleted and their comments still alive in the database since there is no true foreign key relationship between them (which makes sense I think), so we would want to make sure rendering those orphaned comments doesn't break anything.

stefanv: @acrellin Is that a scenario we hope to support?  It adds a lot of duplicate logic to this PR.

acrellin: I agree that we shouldn't patch the back-end to cover these test-only cases, but should rather update the test fixtures to mimic real-life production DB state. I don't think we'll ever want to support unknown users (deactivated users will still have an associated record).

kmshin1397: Okay, I've just pushed a commit to remove that logic and just add in a test factory user to the test DB state instead.

skyportal/models.py
```diff
@@ -725,6 +742,24 @@ class Comment(Base):
...
+    @classmethod
+    def get_if_owned_by(cls, ident, user, options=[]):
+        comment = cls.query.options(options).get(ident)
+
+        if comment is not None and not comment.is_owned_by(user):
+            raise AccessError('Insufficient permissions.')
+
+        # Grab basic author info for the comment
+        author = User.query.filter(User.username == comment.author).first()
+        comment.author_info = {
```
stefanv, line 754: Why not then pass `author` to `author_info`, instead of duplicating the dictionary?

Otherwise, perhaps:

```
user_info = {field: getattr(user, field) for field in ('username', 'first_name', 'last_name', 'gravatar_url')}
```

Or consider `user.to_dict()`?

kmshin1397: I mainly didn't want to pass along the entire User model as that would technically expose contact_emails for every author with comments on a given source to anyone with access to that source and I wasn't sure if that was something we wanted. Also, it never hurts to lower the size of messages we're passing between the back-end and front-end. 

That said, I like the use of getattr() to make that part more succinct so I've incorporated that in.

skyportal/handlers/api/news_feed.py
```diff
@@ -61,13 +61,41 @@ def fetch_newest(model):
...
+                    )
+                )
+                .order_by(desc(Comment.created_at or Comment.saved_at))
+                .limit(n_items)
+                .all()
+            )
+
+            for comment in owned_comments:
+                author = User.query.filter(User.username == comment.author).first()
+                comment.author_info = {
```
acrellin, line 84: This should be factored out into a function as it's duplicated several times. See comment below in models.py

skyportal/tests/conftest.py
```diff
@@ -8,15 +8,13 @@
 
 from baselayer.app import models
 from baselayer.app.config import load_config
-from baselayer.app.test_util import (
+
+from baselayer.app.test_util import (  # noqa: F401
```
acrellin, line 12: Are all of these imports actually required for the tests? If not we should remove those that aren't needed

skyportal/handlers/api/news_feed.py
```diff
@@ -61,13 +61,41 @@ def fetch_newest(model):
...
+                        DBSession()
+                        .query(Source.obj_id)
+                        .filter(
+                            Source.group_id.in_(
+                                [g.id for g in self.current_user.accessible_groups]
+                            )
+                        )
+                    )
+                )
+                .order_by(desc(Comment.created_at or Comment.saved_at))
```
acrellin, line 77: I don't think `Comment` has a `saved_at` column -- this just applies to the more general function above (`Source` has a `saved_at` column)
```suggestion
                .order_by(desc(Comment.created_at))
```

skyportal/handlers/api/news_feed.py
```diff
@@ -61,13 +61,41 @@ def fetch_newest(model):
                 .all()
             )
 
+        def fetch_newest_comments():
```
acrellin, line 64: This query is a duplication of what's implemented above in `fetch_newest`, right? This should be removed, and perhaps just add a check whether `model` is `Comment` in `fetch_newest`, and if so, add the author info. It appears the author info part is the only thing added by this function, but there's a lot of duplication that can be avoided.

static/js/components/NewsFeed.jsx
```diff
@@ -16,11 +19,90 @@ const defaultPrefs = {
   numItems: "",
 };
 
+const newsFeedItem = (item) => {
+  let EntryAvatar;
+  let entryUserName;
```
acrellin, line 24:
```suggestion
  let entryTitle;
```

static/js/components/NewsFeed.jsx
```diff
@@ -16,11 +19,90 @@ const defaultPrefs = {
...
+            backgroundColor: "#38B0DE",
+            color: "white",
+            fontSize: "10px",
+          }}
+        >
+          S
+        </Avatar>
+      );
+      /* eslint-disable react/display-name */
+      entryUserName = "New source";
```
acrellin, line 59:
```suggestion
      entryTitle = "New source";
```

static/js/components/NewsFeed.jsx
```diff
@@ -16,11 +19,90 @@ const defaultPrefs = {
...
+        <UserAvatar
+          size={32}
+          firstName={item.author_info.first_name}
+          lastName={item.author_info.last_name}
+          username={item.author_info.username}
+          gravatarUrl={item.author_info.gravatar_url}
+        />
+      );
+      /* eslint-disable react/display-name */
+      entryUserName = item.author;
```
acrellin, line 39:
```suggestion
      entryTitle = item.author;
```

static/js/components/NewsFeed.jsx
```diff
@@ -16,11 +19,90 @@ const defaultPrefs = {
...
+  }
+
+  return (
+    <Paper
+      key={`newsFeedItem_${item.time}`}
+      className={styles.entry}
+      elevation={2}
+    >
+      <Tooltip
+        title={entryUserName}
```
acrellin, line 72:
```suggestion
        title={entryTitle}
```

static/js/components/NewsFeed.css
```diff
@@ -1,13 +1,63 @@
...
+.entrySourceId {
+  margin-right: 0.5em;
+  font-size: 105%;
+}
+
+.entrySourceId > a {
+  color: #9b9a9a !important;
+}
+
+.entryUserName {
```
acrellin, line 60:
```suggestion
.entryTitle {
```

static/js/components/NewsFeed.jsx
```diff
@@ -16,11 +19,90 @@ const defaultPrefs = {
...
+        arrow
+        placement="top-start"
+        classes={{ tooltip: styles.entryUserName }}
```
acrellin, line 75:
```suggestion
        classes={{ tooltip: styles.entryTitle }}
```

static/js/components/NewsFeed.jsx
```diff
@@ -16,11 +19,90 @@ const defaultPrefs = {
   numItems: "",
 };
 
+const newsFeedItem = (item) => {
```
acrellin, line 22:
```suggestion
const NewsFeedItem = (item) => {
```
b/c this is a React component

static/js/components/NewsFeed.jsx
```diff
@@ -32,21 +114,8 @@ const NewsFeed = () => {
...
-              </span>
-              &nbsp;
-              {item.type}
-              :&nbsp;
-              <span className={styles.entry}>{item.message}</span>
-            </li>
-          ))}
-        </ul>
+      <div className={styles.newsFeed} style={{ height }}>
+        {items.map((item) => newsFeedItem(item))}
```
acrellin, line 118:
```suggestion
        {items.map((item) => NewsFeedItem(item))}
```

skyportal/handlers/api/news_feed.py
acrellin: n/a if this function is removed entirely as suggested below

static/js/components/NewsFeed.jsx
acrellin: This will also need a propTypes statement

acrellin:
@kmshin1397 I updated a few of my review comments above, so don't refer to the comments in the initial GH email

skyportal/tests/conftest.py
kmshin1397: I think so. I removed all the imports that flake8 tagged as being unused, except the baselayer.app.test_util.driver as removing that broke the tests when I ran it.

kmshin1397:
Okay @acrellin I think I incorporated all your suggestions. Thanks, they were definitely good ones. Hopefully the tests will pass

skyportal/tests/conftest.py
acrellin: That's perfect, thanks

acrellin:
@kmshin1397 looks great, thanks! Sorry, one last suggestion that I should've made initially - `Comment.construct_author_info_dict` can just grab the author from itself (`self.author`) -- that should simplify that a bit. Then :+1: from me.

acrellin:
The failure here (`test_classifications`) has been failing elsewhere, too, so can be ignored and this PR can be considered green. Will address in another PR

kmshin1397:
@acrellin I did the last refactor you suggested so this should be good to go I think

acrellin:
This looks great, thanks @kmshin1397 !

PR #716, opened 2020-08-14
added galactic coordinates

guynir42:
Added galactic coordinates to the sources/candidates. 

This shows up in the scanning page as well as in the individual pages for candidates or sources.

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -305,13 +305,13 @@ class Obj(Base, ha.Point):
         return astroplan.FixedTarget(name=self.id, coord=coord)
 
     @property
-    def gal_lat(self):
+    def gal_lat_deg(self):
         """Get the galactic latitute of this object"""
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
         return coord.galactic.b.deg
 
     @property
-    def gal_lon(self):
+    def gal_lon_deg(self):
         """Get the galactic longitude of this object"""
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
         return coord.galactic.l.deg
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -304,32 +304,17 @@ class Obj(Base, ha.Point):
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit='deg')
         return astroplan.FixedTarget(name=self.id, coord=coord)
 
-    @hybrid_property
+    @property
     def gal_lat(self):
         """Get the galactic latitute of this object"""
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
         return coord.galactic.b.deg
 
-    @gal_lat.expression
-    def gal_lat(cls):
-        ra_array = DBSession.query(cls.ra).all()
-        dec_array = DBSession.query(cls.dec).all()
-        coords = ap_coord.SkyCoord(ra_array, dec_array, unit="deg")
-        return coords.galactic.b.deg
-
-    @hybrid_property
+    @property
     def gal_lon(self):
         """Get the galactic longitude of this object"""
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
         return coord.galactic.l.deg
-
-    @gal_lon.expression
-    def gal_lon(cls):
-        ra_array = DBSession.query(cls.ra).all()
-        dec_array = DBSession.query(cls.dec).all()
-        coords = ap_coord.SkyCoord(ra_array, dec_array, unit="deg")
-        # return sa.orm.query.Query(coords.galactic.l.deg).label("gal_lon")
-        return coords.galactic.l.deg
 
     def airmass(self, telescope, time, below_horizon=np.inf):
         """Return the airmass of the object at a given time. Uses the Pickering
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -304,6 +304,32 @@ class Obj(Base, ha.Point):
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit='deg')
         return astroplan.FixedTarget(name=self.id, coord=coord)
 
+    @hybrid_property
+    def gal_lat(self):
+        """Get the galactic latitute of this object"""
+        coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
+        return coord.galactic.b.deg
+
+    @gal_lat.expression
+    def gal_lat(cls):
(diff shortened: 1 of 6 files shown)


static/js/components/CandidateList.jsx
```diff
@@ -105,6 +105,10 @@ const CandidateList = () => {
                     &nbsp;
                     {candidateObj.dec}
                     <br />
+                    Gal. Coords (l,b) :&nbsp;
```
acrellin, line 108:
```suggestion
                    Gal. Coords (l,b):&nbsp;
```

skyportal/models.py
```diff
@@ -304,6 +304,32 @@ def target(self):
...
+        """Get the galactic longitude of this object"""
+        coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
+        return coord.galactic.l.deg
+
+    @gal_lon.expression
+    def gal_lon(cls):
+        ra_array = DBSession.query(cls.ra).all()
+        dec_array = DBSession.query(cls.dec).all()
+        coords = ap_coord.SkyCoord(ra_array, dec_array, unit="deg")
+        return sa.orm.query.Query(coords.galactic.l.deg).label("gal_lon")
```
dannygoldstein, line 331: I believe this is will error or silently fail as `coords.galactic.l.deg` is not an SQLalchemy entity. @guynir42 Can you add a test for this?

skyportal/models.py
```diff
@@ -304,6 +304,32 @@ def target(self):
...
+        """Get the galactic latitute of this object"""
+        coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
+        return coord.galactic.b.deg
+
+    @gal_lat.expression
+    def gal_lat(cls):
+        ra_array = DBSession.query(cls.ra).all()
+        dec_array = DBSession.query(cls.dec).all()
+        coords = ap_coord.SkyCoord(ra_array, dec_array, unit="deg")
+        return sa.orm.query.Query(coords.galactic.b.deg).label("gal_lat")
```
dannygoldstein, line 318: See comment on `@gal_lon.expression` below

skyportal/models.py
dannygoldstein: in the preceding lines, the code can pull the entire ra, dec columns of the table (potentially billions of rows) into memory, convert them to to galactic coords client side, then (if line 331 does not throw an exception) push the results back to the database as a literal array. If the conversion is happening client side anyway, maybe best to implement this as a straight up `@property` instead of a `@hybrid_property` (i.e., delete this method and re-implement the corresponding getter as `@property`)?

guynir42: First off you are right on the first comment: I can just as easily return an array and it seems to work. I am not sure what kind of test to do but the coordinates show up on all the relevant pages. The only question I am now left with is how it worked to begin with...?

As for your second comment, I think that since there are two "getters" in the hybrid property, it can choose which ever one is more useful, for pulling one item at a time or to pull the entire column. I have no idea what is more efficient or if SQLAlchemy is smart enough to do the right thing. So whatever you think is best to do.

skyportal/models.py
```diff
@@ -304,6 +304,33 @@ def target(self):
...
+        coords = ap_coord.SkyCoord(ra_array, dec_array, unit="deg")
+        return coords.galactic.b.deg
+
+    @hybrid_property
+    def gal_lon(self):
+        """Get the galactic longitude of this object"""
+        coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
+        return coord.galactic.l.deg
+
+    @gal_lon.expression
```
dannygoldstein, line 326: Delete this method

skyportal/models.py
```diff
@@ -304,6 +304,33 @@ def target(self):
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit='deg')
         return astroplan.FixedTarget(name=self.id, coord=coord)
 
+    @hybrid_property
+    def gal_lat(self):
+        """Get the galactic latitute of this object"""
+        coord = ap_coord.SkyCoord(self.ra, self.dec, unit="deg")
+        return coord.galactic.b.deg
+
+    @gal_lat.expression
```
dannygoldstein, line 313: Delete this method

skyportal/models.py
```diff
@@ -304,6 +304,33 @@ def target(self):
...
+    def gal_lat(cls):
+        ra_array = DBSession.query(cls.ra).all()
+        dec_array = DBSession.query(cls.dec).all()
+        coords = ap_coord.SkyCoord(ra_array, dec_array, unit="deg")
+        return coords.galactic.b.deg
+
+    @hybrid_property
```
dannygoldstein, line 320:
```suggestion
    @property
```

skyportal/models.py
```diff
@@ -304,6 +304,33 @@ def target(self):
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit='deg')
         return astroplan.FixedTarget(name=self.id, coord=coord)
 
+    @hybrid_property
```
dannygoldstein, line 307:
```suggestion
    @property
```

skyportal/models.py
```diff
@@ -304,33 +304,18 @@ def target(self):
         coord = ap_coord.SkyCoord(self.ra, self.dec, unit='deg')
         return astroplan.FixedTarget(name=self.id, coord=coord)
 
-    @hybrid_property
+    @property
     def gal_lat(self):
```
dannygoldstein, line 308:
```suggestion
    def gal_lat_deg(self):
```

skyportal/models.py
```diff
@@ -304,33 +304,18 @@ def target(self):
...
-    @gal_lat.expression
-    def gal_lat(cls):
-        ra_array = DBSession.query(cls.ra).all()
-        dec_array = DBSession.query(cls.dec).all()
-        coords = ap_coord.SkyCoord(ra_array, dec_array, unit="deg")
-        return coords.galactic.b.deg
-
-    @hybrid_property
+    @property
     def gal_lon(self):
```
dannygoldstein, line 314:
```suggestion
    def gal_lon_deg(self):
```

guynir42:
@dannygoldstein  I pushed the requested changes...

dannygoldstein:
Will merge when green

guynir42:
Thanks @dannygoldstein

dannygoldstein:
#733 (now merged) should fix these failures, will rerun CI.

dannygoldstein:
Thanks @guynir42

PR #717, opened 2020-08-14
updated setup page 

guynir42:
…t about load_demo_data a little more clear.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -51,7 +51,7 @@ brew install geckodriver
 1. Install dependencies
 
 ```
-sudo apt install nginx supervisor postgresql libpq-dev nodejs
+sudo apt install nginx supervisor postgresql libpq-dev npm python3-pip
 ```
 
 2. Configure your database permissions.
@@ -84,10 +84,19 @@ sudo service postgresql restart
 1. Initialize the database with `make db_init` (this only needs to
    happen once).
 2. Copy `config.yaml.defaults` to `config.yaml`.
-2. If you want some test data to play with, run `make load_demo_data`.
 3. Run `make log` to monitor the service and, in a separate window, `make run` to start the server.
+4. Direct your browser to `http://localhost:5000`.
+5. If you want some test data to play with, run `make load_demo_data` (do this while the server is running!).
+6. Change users by navigating to `http://localhost:5000/become_user/<#>` where # is a number from 1-5.
+   Different users have different privileges and can see more or less of the demo data.
 
-Direct your browser to http://localhost:5000.
+## Troubleshooting
+
+If you have trouble getting the demo data try doing
+
+```make db_clear && make db_init && make run```
+
+and then, from a different window, do `make load_demo_data`.
 
 ## Additional Configuration
 


doc/setup.md
```diff
@@ -51,7 +51,7 @@ brew install geckodriver
 1. Install dependencies
 
 ```
-sudo apt install nginx supervisor postgresql libpq-dev nodejs
```
stefanv, line 54: I think removing nodejs here is not correct.  See https://github.com/skyportal/skyportal/blob/master/.github/workflows/test.yml#L80

doc/setup.md
```diff
@@ -79,15 +79,32 @@ sudo service postgresql restart
 - Extract the binary to somewhere on your path
 - Ensure it runs with `geckodriver --version`
 
+4. make the dependencies using
```
stefanv, line 82: This is done automatically upon first launch of skyportal.

doc/setup.md
```diff
@@ -79,15 +79,32 @@ sudo service postgresql restart
...
+```
+
 ## Launch
 
 1. Initialize the database with `make db_init` (this only needs to
    happen once).
 2. Copy `config.yaml.defaults` to `config.yaml`.
-2. If you want some test data to play with, run `make load_demo_data`.
 3. Run `make log` to monitor the service and, in a separate window, `make run` to start the server.
+4. Direct your browser to http://localhost:5000. Login to SkyPortal.
```
stefanv, line 94:
```suggestion
4. Direct your browser to `http://localhost:5000`. Log in to SkyPortal (by default, you are in debug mode and no password will be required).
```

doc/setup.md
```diff
@@ -79,15 +79,32 @@ sudo service postgresql restart
...
+5. If you want some test data to play with, run `make load_demo_data` (do this while the server is running!).
+6. Change users by navigating to localhost:5000/become_user/<#> where # is a number from 1-5.
```
stefanv, line 96:
```suggestion
6. Change users by navigating to `http://localhost:5000/become_user/<#>` where # is a number from 1-5.
```

Should we just make the default user have access to the demo data?

doc/setup.md
```diff
@@ -79,15 +79,32 @@ sudo service postgresql restart
...
+   Different users have different privileges and can see more or less of the demo data.
+
+## Troubleshooting
+
+If you have trouble getting the demo data try doing
+
+```make db_clear && make db_init && make run```
+
+and then, after logging in, from a different window, do `make load_demo_data`.
```
stefanv, line 105: I don't think logging in is a prerequisite.

doc/setup.md
```diff
@@ -79,15 +79,32 @@ sudo service postgresql restart
...
 
-Direct your browser to http://localhost:5000.
+You should be able to log in as testadmin rather than testuser.
```
stefanv, line 107: Not sure what this instruction is meant to tell the reader.

doc/setup.md
guynir42: I'm not sure what the reference means, but I tested it on a new install of Ubuntu 20 and it works without specifying nodejs. 

I think what happens is that npm install it.

doc/setup.md
guynir42: Ok I will remove it.

doc/setup.md
guynir42: Ok I'm not sure either. Will remove it...

doc/setup.md
guynir42: I am not 100% sure but I think that if you don't log in at all it will complain about not having tokens and it fails to load the demo data.

doc/setup.md
guynir42: I tested this line on a fresh install of Ubuntu 20, and everything works.

doc/setup.md
stefanv: Can you confirm this?  I've never seen this issue.  The tokens are generated upon app startup.

guynir42: Actually I just checked again and you are right, there is no need to login.

package.json
```diff
@@ -12,7 +12,7 @@
     "@material-ui/lab": "4.0.0-alpha.56",
     "@material-ui/pickers": "3.2.10",
     "bokehjs": "0.12.9",
-    "check-dependencies": "1.1.0",
+    "check-dependencies": "^1.1.0",
```
stefanv, line 15:
```suggestion
    "check-dependencies": "1.1.0",
```

stefanv:
Please reset baselayer to the same version as on master so that I can merge.

stefanv:
Sorry, @guynir42, not sure why you had problems with baselayer, but I compressed all your changes into a single patch.

dannygoldstein:
green

dannygoldstein:
Is this ready to merge?

doc/setup.md
guynir42: @stefanv what do you say? can we merge this?

PR #724, opened 2020-08-18
Make Home Page responsive, customizable, and use MUI Paper

kmshin1397:
For [User Story 1078](https://app.clubhouse.io/astromarshal/story/1078/draggable-and-resizable-layout-of-home-page)

Layout reactive to screen size:
reactive_homepage

Components can be resized and dragged:
resize_homepage

Todo:
- Save layouts in the state so that when users follow a link elsewhere and then comeback, the layout is as they dragged/resized it last (assuming the window size hasn't changed drastically)

Notes:
- Still doesn't look very good on mobile-size devices (the sidebar/top bar also needs to be made more responsive to get there I think)
- Leaving this as a draft as I think several components on the page will be changing soon (i.e. my Newsfeed reskin) so may need to tweak default layouts in response
- Seems like eventually we want these layouts to be configurable through the YAML configs as well so that would also need to be implemented


acrellin:
Looking very cool!

profjsb:
Yes, very cool and heading in the right direction for us. Rather than allowing the change of size, I'd much rather focus on allowing the change of location (that is, we as developers should able to consider how big a widget is within reason) and the SAVING of that location to the user preferences JSON, in addition to state.

kmshin1397:
- Updated widgets to MUI theme.
- Widgets now only draggable and are not resizable
- On layout change, save the layout to user preferences and on component load, check for saved preferences

Screenshot (2)

acrellin:
Looking great! Thanks @kmshin1397 !

dmitryduev:
🚀 @kmshin1397 !!

dmitryduev:
GA has been consistently failing on the recent PRs (`frontend/test_sources`)

static/js/components/GroupList.jsx
```diff
@@ -1,31 +1,33 @@
...
+import Paper from "@material-ui/core/Paper";
 import List from "@material-ui/core/List";
 import ListItem from "@material-ui/core/ListItem";
 import ListItemText from "@material-ui/core/ListItemText";
 import Typography from "@material-ui/core/Typography";
 
 const GroupList = ({ title, groups }) => (
-  <Box p={1}>
-    <Typography variant="h5">{title}</Typography>
+  <Paper elevation={1} height="100%">
```
dmitryduev, line 10: Please update https://github.com/skyportal/skyportal/blob/master/static/js/components/Groups.jsx#L27 accordingly as well.

static/js/components/HomePage.css
```diff
@@ -1,5 +1,71 @@
 .homePageWidgetDiv {
-  float: left;
+  /* float: left;
```
dmitryduev, line 2: ditch the commented out section?

static/js/components/HomePage.css
```diff
@@ -1,5 +1,71 @@
...
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  -o-user-select: none;
+  user-select: none;
+}
+
+.react-grid-item > .react-resizable-handle {
+  position: absolute;
+  width: 20px;
```
dmitryduev, line 51: -> `rem` here and below?

static/js/components/HomePage.jsx
```diff
@@ -1,29 +1,108 @@
...
 import SourceList from "./SourceList";
 import GroupList from "./GroupList";
 import NewsFeed from "./NewsFeed";
 import TopSources from "./TopSources";
 import styles from "./HomePage.css";
 
+const ResponsiveGridLayout = WidthProvider(Responsive);
+
+const xlgLayout = [
+  { i: "sourceList", x: 0, y: 0, w: 9, h: 6, minW: 9, isResizable: false },
```
dmitryduev, line 20: This is great, but eventually we would probably want to move this over to a config of some sort. What say you, @acrellin?

static/js/components/HomePage.jsx
```diff
@@ -1,29 +1,108 @@
...
   const groups = useSelector((state) => state.groups.user);
+
+  const preferredLayouts = useSelector(
+    (state) => state.profile.preferences.layouts
+  );
+
+  const currentLayouts =
+    preferredLayouts == null ? defaultLayouts : preferredLayouts;
+
+  console.log(preferredLayouts);
```
dmitryduev, line 72: remove debugging stuff please

static/js/components/HomePage.jsx
```diff
@@ -1,29 +1,108 @@
...
+  console.log(currentLayouts);
```
dmitryduev, line 73: Same

static/js/components/NewsFeed.css
```diff
@@ -1,8 +1,19 @@
-.newsFeed {
+.newsFeed,
+.newsFeedDark {
   display: flex;
   flex-direction: column;
   overflow-y: scroll;
   padding-top: 5px;
+  padding-left: 5px;
```
dmitryduev, line 7: rem?

static/js/components/SearchBox.jsx
```diff
@@ -83,6 +83,7 @@ const SearchBox = ({ sources }) => {
     },
     paper: {
       padding: "1rem",
+      marginTop: "10px",
```
dmitryduev, line 86: rem

dannygoldstein:
This is nice stuff. Good work @kmshin1397

dmitryduev:
LGTM!

profjsb:
Looking great! I'd indeed agree with @dmitryduev in moving the hardcoded layout matrices into the config.yaml. This could also serve the dual purpose of setting explicitly whether a component would be shown or not on the homepage. E.g. it's possible a SP instance does not want to show the "groups" widget. If there is no entry for it in this matrix then it wont be shown in the homepage.

Minor issue is that there is large vertical space between the view sources and the other widgets in tablet/mobile views.

dannygoldstein:
> Looking great! I'd indeed agree with @dmitryduev in moving the hardcoded layout matrices into the config.yaml. This could also serve the dual purpose of setting explicitly whether a component would be shown or not on the homepage. E.g. it's possible a SP instance does not want to show the "groups" widget. If there is no entry for it in this matrix then it wont be shown in the homepage.
> 
> Minor issue is that there is large vertical space between the view sources and the other widgets in tablet/mobile views.

Two more issues: 

1. the size of each component on the drag-n-drop can be much larger than the space the component acutally takes up, leading to empty space between components (on desktop):
Screen Shot 2020-08-30 at 4 14 08 PM

2. Looks like a dark background is used even when dark mode is not enabled? This makes  text difficult to read:
Screen Shot 2020-08-30 at 4 14 29 PM


\

PR #725, opened 2020-08-19
fix bug in Photometry.mag.expression

dannygoldstein:
This fixes a bug in `Photometry.mag.expression` that caused usages of SQL hybrid of this property to error with `NotImplementedError: Operator 'getitem' is not supported on this expression`.


acrellin:
Is this just wrapping the elements in a tuple?

dannygoldstein:
Yep

PR #730, opened 2020-08-20
GA: merge with correct upstream branch

stefanv:
Earlier, I believed that GA did not do a merge before testing, but this was incorrect.  This PR therefore removes the manual steps put in place to fix it.

FWIW, the manual step was subtly incorrect, and merged against the *user's* master branch instead of against *upstream/master* (I think).


stefanv:
Unrelated failures.

PR #731, opened 2020-08-20
Fix classification test

stefanv:


acrellin:
@stefanv any luck with this?

PR #732, opened 2020-08-21
Quicksearchbar tweaks

stefanv:


profjsb:
Closing as PR #739 implements most of the quick searchbar changes that were first introduced in this PR by @stefanv, but not the test fixes and the work to try to fix the link/history issue. We may need to revisit some of the react-dom ideas herein later.

PR #733, opened 2020-08-24
Conditionally xfail test_classifications, test_data_sharing on CI builds only

acrellin:


skyportal/tests/frontend/test_data_sharing.py
```diff
@@ -9,6 +12,8 @@ def test_share_data(
     public_group,
     public_group2,
 ):
+    if IS_CI_BUILD:
```
dannygoldstein, line 15: Can we reimplement this (and subsequent usages) as a decorator?

dannygoldstein: Rescinding this comment as @acrellin's implementation follows `pytest` best practices: 

https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail

PR #736, opened 2020-08-24
Add recent sources widget

kmshin1397:
Move large sources search widget to separate page and replace with "Recently Added Sources" widget:

recentsources

Addresses [User Story 1067](https://app.clubhouse.io/astromarshal/story/1067/enhance-sources-widget)


profjsb:
Looks great. What does the preferences icon in the top right allow the user to do? Have you tested what happens to the widget size (length) if someone puts in a large number of sources to show? There should likely be a scrollbar and/or some max number to show.

profjsb:
To be clear, I think this PR should be called something like "Recently added sources widget" since that is what it does. We still have the notion of a sources display:
image

kmshin1397:
> Looks great. What does the preferences icon in the top right allow the user to do? Have you tested what happens to the widget size (length) if someone puts in a large number of sources to show? There should likely be a scrollbar and/or some max number to show.

The preferences icon has the max number of items setting. The number is set to 5 by default in the backend so the widget shouldn't get too big accidentally, but I should probably put the default 5 into the preferences widget on the frontend. Do you think a scrollbar should still be there even if the max number preference is there?

profjsb:
> > Looks great. What does the preferences icon in the top right allow the user to do? Have you tested what happens to the widget size (length) if someone puts in a large number of sources to show? There should likely be a scrollbar and/or some max number to show.
> 
> The preferences icon has the max number of items setting. The number is set to 5 by default in the backend so the widget shouldn't get too big accidentally, but I should probably put the default 5 into the preferences widget on the frontend. Do you think a scrollbar should still be there even if the max number preference is there?

Probably fine to avoid the scrollbar in this case.

static/js/components/SourceQuickView.jsx
```diff
@@ -0,0 +1,199 @@
...
+          onClick={handleClickOpen}
+          startIcon={<VisibilityIcon />}
+          style={{ marginBottom: "10px" }}
+        >
+          QUICK VIEW
+        </Button>
+        <Dialog open={open} onClose={handleClose} style={{ position: "fixed" }}>
+          <DialogTitle onClose={handleClose}>{source.id}</DialogTitle>
+          <DialogContent dividers>
+            {getDialogContentDiv(source, isCached, taxonomyList)}
```
acrellin, line 153:
```suggestion
            <DialogContentDiv source={source} isCached={isCached} taxonomyList={taxonomyList} />
```

kmshin1397:
@acrellin Suggestions addressed. Thanks for the review!

static/js/components/RecentSources.css
```diff
@@ -0,0 +1,77 @@
...
+.recentSourceItemWithButton:hover .quickViewButton {
+  visibility: visible;
+  display: block;
+}
+
+.recentSourceItemWithButton {
+  display: flex;
+  flex-flow: column nowrap;
+  justify-content: center;
+  margin-bottom: 15px;
```
dmitryduev, line 74: Please convert to rem.

static/js/components/RecentSources.css
```diff
@@ -0,0 +1,77 @@
...
+  margin-left: 10px;
```
dmitryduev, line 75: Same here

static/js/components/SourceQuickView.css
```diff
@@ -0,0 +1,34 @@
...
+.title {
+  font-size: 1.5rem;
+  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
+  font-weight: 500;
+  line-height: 1.6;
+  letter-spacing: 0.0075em;
+}
+
+.dialogContent {
+  padding: 15px;
```
dmitryduev, line 28: rem?

static/js/components/SourceQuickView.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.sourceLinkButton {
+  margin-top: 15px;
```
dmitryduev, line 32: rem?

static/js/components/SourceQuickView.jsx
```diff
@@ -0,0 +1,229 @@
+import React, { useState, useEffect } from "react";
```
dmitryduev, line 1: Really like the quick view idea!

static/js/ducks/recentSources.js
```diff
@@ -0,0 +1,32 @@
...
+messageHandler.add((actionType, payload, dispatch) => {
+  if (actionType === FETCH_RECENT_SOURCES) {
+    dispatch(fetchRecentSources());
+  }
+});
+
+const reducer = (state = { recentSources: [] }, action) => {
+  switch (action.type) {
+    case FETCH_RECENT_SOURCES_OK: {
+      const recentSources = action.data;
```
dmitryduev, line 22: Could you specify what's in `action.data` explicitly?

skyportal/tests/frontend/test_recent_sources.py
```diff
@@ -0,0 +1,54 @@
+import uuid
+from selenium.webdriver.common.action_chains import ActionChains
+
+from skyportal.tests import api
+
+
+def test_recent_sources(driver, user, public_source, public_group, upload_data_token):
```
dmitryduev, line 7: Could you please add a test where a user that is not part of the group and thus does not see the recently added source?

static/js/ducks/recentSources.js
kmshin1397: Do you mean to split `action.data` into more explicit properties? This is an array of objects instead of a single object so I'm not sure how to make this more explicit.

dmitryduev: Right, nm then

kmshin1397:
Updated to reflect merge of MUI Paper and responsive grid layout for home page.

Also implement a potential response to Issue #764 by switching from a component which increases in height as preferred number of sources displayed is increased to a fixed larger size with a scrollbar. The wasted grid space in the issue is because heights can't be made dynamic in the react-grid-layout package if we set the components to be non-resizable so I needed to set the grid item size to some maximum greater than needed to accommodate only 2 recent sources.

New look of recent sources widget:
Screenshot 2020-08-30 182933

static/js/components/HomePage.jsx
```diff
@@ -83,13 +93,12 @@ const HomePage = () => {
     <ResponsiveGridLayout
       className="layout"
       layouts={currentLayouts}
-      breakpoints={{ xlg: 1400, lg: 1150, md: 996, sm: 768, xs: 480 }}
+      breakpoints={{ xlg: 1400, lg: 1150, md: 996, sm: 650, xs: 0 }}
```
stefanv, line 96: Should these be defined somewhere and be used throughout the app?

static/js/components/RecentSources.css
```diff
@@ -0,0 +1,82 @@
+.stampContainer {
```
stefanv, line 1: Use `makeStyle`?

kmshin1397:
> Could you explain how the classification logic ties into this PR?

The classifications are fetched to be displayed with the source name and also in the quick summary view dialog (you can see this in the initial comment to this PR in the GIF).

static/js/components/RecentSources.css
kmshin1397: Done in latest commit

static/js/components/HomePage.jsx
kmshin1397: Might be a good idea to have defined globally, but I'm not sure. These breakpoint values will probably differ slightly from what we want for the global app because these widths are with respect to viewport width minus the width of the sidebar (except the small ones where the sidebar is closed by default) since that's what is covered by the grid div.

kmshin1397:
Included resizing of other widget papers to fill the "red space" for Issue #764 
Screenshot 2020-09-02 100957

kmshin1397:
@dmitryduev @acrellin Just a few styling changes (mainly to accommodate changes to master) and tweaks to unrelated tests to get this green since you last approved this. If one of you could do a quick re-review and merge, that'd be great. Thanks!

dmitryduev (approved):
👍 @acrellin will you please merge this one? (as long as @stefanv is happy)

PR #738, opened 2020-08-25
Add Allocation API (GroupInstrument)

dannygoldstein:
This PR adds a new API endpoint: `/allocation`, which represents an allocation of observing time on an instrument. This is the `GroupInstrument` endpoint discussed in #672. This model can be used downstream for controlling the instruments that users can submit follow-up requests to. 

Some implementation details: 

   * Users can only retrieve allocations that are associated with groups that they are members of.
   * Only System Admins can post or modify  allocations. We may relax this later - I am thinking of developing a hook that creates ACLs for controlling allocations on an instrument-by-instrument basis. Not a part of this PR. 

This is an API-only PR, there are no subsequent frontend PRs. Follow-up request refactor will be based on this PR.


dannygoldstein:
[ch1158]

skyportal/handlers/api/allocation.py
```diff
@@ -0,0 +1,182 @@
...
+            DBSession()
+            .query(Allocation)
+            .filter(
+                Allocation.group_id.in_(
+                    [g.id for g in self.current_user.accessible_groups]
+                )
+            )
+        )
+
+        if allocation_id is not None:
```
acrellin, line 53: Why not just have a separate query for this case? It'd simplify some of the logic below, specifically everywhere we check whether `allocation_id is not None`, and those lines can just be moved inside this block. This block could (should) then also be moved to the top with its own return statement, then everything following would be for the "multiple" case. That's a pattern we use in several other GET handlers. Trying to make a single query generalized for all cases seems to actually increase complexity, but I imagine the intent is the opposite

skyportal/tests/api/test_allocations.py
```diff
@@ -0,0 +1,97 @@
+from skyportal.tests import api
+
+
+def test_super_user_post_allocation(sedm, public_group, super_admin_token):
```
acrellin, line 4: Can we also check some of the cases you're handling regarding posting invalid instrument/group IDs?

skyportal/handlers/api/allocation.py
dannygoldstein: done

skyportal/tests/api/test_allocations.py
dannygoldstein: done

skyportal/handlers/api/allocation.py
```diff
@@ -0,0 +1,176 @@
...
+            try:
+                allocation_id = int(allocation_id)
+            except ValueError:
+                return self.error("Allocation ID must be an integer.")
+            allocations = allocations.filter(Allocation.id == allocation_id).all()
+            if len(allocations) == 0:
+                return self.error("Could not retrieve allocation.")
+            return self.success(data=allocations[0])
```
acrellin, line 61: Why doesn't this need to be passed to `schema.dump` as with the multiple case below? I think it was before. Is that really needed below? `BaseHandler.success` does convert everything to JSON already

dannygoldstein: fixed

PR #739, opened 2020-08-25
location fix + Stefan's style changes

profjsb:
Fixes a typo in `HeaderContent.jsx.template` which caused the strange formatting and adds @stefanv's style changes from #732.

### DIFF
--- a/static/js/components/HeaderContent.jsx.template
+++ b/static/js/components/HeaderContent.jsx.template
@@ -11,7 +11,7 @@ import styles from "./Main.css";
 
 const HeaderContent = () => (
   <div className={styles.topBannerContent}>
-    <div style= {%- raw -%} {{display: "inline-flx", flexDirection: "row" }} {%- endraw -%} >
+    <div style= {%- raw -%} {{display: "inline-flex", flexDirection: "row" }} {%- endraw -%} >
       <Logo className={styles.logo} />
       <Link className={styles.title} to="/">
         {{ app.title }}
--- a/static/js/components/HeaderContent.jsx.template
+++ b/static/js/components/HeaderContent.jsx.template
@@ -5,7 +5,6 @@ import Box from "@material-ui/core/Box";
 import Responsive from "./Responsive";
 import ProfileDropdown from "./ProfileDropdown";
 import Logo from "./Logo";
-import QuickSearchBar from "./QuickSearchBar";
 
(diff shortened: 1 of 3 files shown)


stefanv:
This doesn't do the moving over to the sidebar, does it? My styling changes were meant for that move.

I guess the tests on that one may have failed because the component changed places?

static/js/components/HeaderContent.jsx.template
```diff
@@ -5,19 +5,17 @@ import Box from "@material-ui/core/Box";
...
 import ProfileDropdown from "./ProfileDropdown";
 import Logo from "./Logo";
-import QuickSearchBar from "./QuickSearchBar";
 
 import styles from "./Main.css";
 
 const HeaderContent = () => (
   <div className={styles.topBannerContent}>
-    <div style= {%- raw -%} {{display: "inline-flx", flexDirection: "row" }} {%- endraw -%} >
+    <div style= {%- raw -%} {{display: "inline-flex", flexDirection: "row" }} {%- endraw -%} >
```
stefanv, line 13: No longer necessary?

profjsb: It is -- need to spell `flex` whrite!

profjsb:
> I think the following fix is quite an important one too:
> 
> https://github.com/skyportal/skyportal/pull/732/files#diff-8e6371e901a30fdd0767925dd55ba088R64
> 
> (This prevents the quicksearch from messing with filtering results.)
> 
This isn't defined in `sources.js` -- do we need to create a path for that in `sources.js`?

static/js/components/HeaderContent.jsx.template
stefanv: I meant, is that whole line still needed :)

stefanv:
> > I think the following fix is quite an important one too:
> > https://github.com/skyportal/skyportal/pull/732/files#diff-8e6371e901a30fdd0767925dd55ba088R64
> > (This prevents the quicksearch from messing with filtering results.)
> 
> This (`FETCH_AUTOCOMPLETE_SOURCES`) isn't defined in `sources.js` -- do we need to create a path for that in `sources.js`?

Yes, problem is that FETCH_SOURCES is handled, and causes the interaction with the filter I described.  By sending an unknown event, nothing happens on the frontend (correct behavior).

PR #741, opened 2020-08-25
Fix quick search box positioning in header

dmitryduev:


dmitryduev:
Addressed in #739, closing.

PR #742, opened 2020-08-25
Only run GA on pull requests or on pushes to master

stefanv:
@dmitryduev This should resolve the building on your own repo that you've been seeing.


dannygoldstein:
@stefanv Looks like, on PR, this no longer runs the test suite on GA. Want to confirm - is this the intended behavior?

stefanv:
Looks like that didn't work then.  The idea was to run only on PR or pushes to master, but not on people's own GAs.

Guess we'll revert for now and live with the multiple runs?

PR #744, opened 2020-08-26
Update photometry plot

kmshin1397:
- Add vertical bars for the first and last detection (if any)
- Update default zoom scale to use a slightly wider y-axis scale than the faintest and brightest detections
- Fix download for binned data

Screenshot 2020-08-25 191417
Screenshot 2020-08-25 191458

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -375,36 +375,39 @@ def photometry_plot(obj_id, user, width=600, height=300):
 
     # Mark the first and last detections
     detection_dates = data[data['hasflux']]['mjd']
-    first = detection_dates.min()
-    last = detection_dates.max()
-    first_color = "#34b4eb"
-    last_color = "#8992f5"
-    line_top = upper + 50
-    line_bottom = lower - 50
-    first_r = plot.vbar(
-        first,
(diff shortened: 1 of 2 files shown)


kmshin1397:
Okay @dannygoldstein I think I fixed what you suggested, let me know what you think

skyportal/plot.py
```diff
@@ -377,28 +373,73 @@ def photometry_plot(obj_id, user, width=600, height=300):
...
+        line_top = 5 * ymax - 4 * midpoint
+        line_bottom = 5 * ymin - 4 * midpoint
+        y = np.linspace(line_bottom, line_top, num=5000)
+        first_r = plot.line(
+            x=first_x, y=y, line_alpha=0.5, line_color=first_color, line_width=2,
+        )
+        plot.add_tools(
+            HoverTool(tooltips=[("First detection", f'{first}')], renderers=[first_r],)
+        )
+        last_r = plot.line(
```
stefanv, line 432: Would these be better implemented as a Span?

http://docs.bokeh.org/en/latest/docs/user_guide/annotations.html#spans

PR #762, opened 2020-08-30
update gitignore to ignore file with secrets

dannygoldstein:
the previous `.gitignore` did not ignore config.yaml, with this PR it is ignored. now it can be used to store secrets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@ geckodriver.log
 __pycache__/
 node_modules/
 *.pyc
-./config.yaml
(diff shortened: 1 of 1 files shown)


stefanv:
I think the idea of this formulation was to only ignore the root config. That said, very unlikely we'll have other files of that name.

acrellin:
This is a duplicate of a patch in my on-boarding PR

dannygoldstein:
Closed in favor of #757

PR #763, opened 2020-08-30
Group sources

guynir42:
Added a page: /group_sources/### where ### is the group id, where you can see all the sources saved to that group. 

I had to make some changes to the back end source handler (source.py) I think it is still consistent with the rest of the app, and also now you can make API calls to get all sources and filter on things like group id, even without specifying the page number and number of sources per page. 

Screenshot from 2020-09-10 16-22-02

Screenshot from 2020-09-10 16-21-52


skyportal/handlers/api/source.py
```diff
@@ -232,17 +240,25 @@ def get(self, obj_id=None):
...
-        q = Obj.query.filter(
-            Obj.id.in_(
-                DBSession()
-                .query(Source.obj_id)
-                .filter(
-                    Source.group_id.in_(
-                        [g.id for g in self.current_user.accessible_groups]
-                    )
+
+        # Instead of single source, let's try to get all of them
```
acrellin, line 244: What does this comment mean? Perhaps you misunderstood the original query - it was always querying all matching Objs

guynir42: The way I understood it, the first IF statement tests for a specific obj_id and returns a single match. Then if that doesn't happen we go to the "get all sources" branch. 

I think it is good to put a comment (maybe a differently phrased one) to show this is a completely different option that is not reached if the first IF is triggered (due to a "return" in that branch).

skyportal/handlers/api/source.py
```diff
@@ -274,13 +290,20 @@ def get(self, obj_id=None):
             )
         if has_tns_name in ['true', True]:
             q = q.filter(Obj.altdata['tns']['name'].isnot(None))
-
+        if group_id:
+            if not group_id.isdigit() or int(group_id) not in [
+                g.id for g in self.current_user.accessible_groups
+            ]:
+                return self.error(
+                    "Requested group '" + group_id + "' is inaccessible to user."
```
acrellin, line 298:
```suggestion
                    f"Requested group with ID '{group_id}' is inaccessible to user."
```

skyportal/handlers/api/source.py
```diff
@@ -274,13 +290,20 @@ def get(self, obj_id=None):
             )
         if has_tns_name in ['true', True]:
             q = q.filter(Obj.altdata['tns']['name'].isnot(None))
-
+        if group_id:
+            if not group_id.isdigit() or int(group_id) not in [
```
acrellin, line 294: We should check separately whether it's not a digit, and return a more specific error message in that case.

package.json
```diff
@@ -27,7 +27,7 @@
     "npm-check": "5.9.2",
     "papaparse": "5.2.0",
     "prop-types": "15.7.2",
-    "qrcode.react": "1.0.0",
+    "qrcode.react": "^1.0.0",
```
acrellin, line 30:
```suggestion
    "qrcode.react": "1.0.0",
```

skyportal/handlers/api/source.py
```diff
@@ -131,6 +131,13 @@ def get(self, obj_id=None):
             description: |
               Arrow-parseable date string (e.g. 2020-01-01). If provided, filter by
               last_detected <= endDate
+          - in: query
+            name: group_id
```
acrellin, line 135: This should be a list, so that one can filter on any number of groups

skyportal/handlers/api/source.py
```diff
@@ -274,13 +290,20 @@ def get(self, obj_id=None):
...
             q = q.filter(Obj.altdata['tns']['name'].isnot(None))
-
+        if group_id:
+            if not group_id.isdigit() or int(group_id) not in [
+                g.id for g in self.current_user.accessible_groups
+            ]:
+                return self.error(
+                    "Requested group '" + group_id + "' is inaccessible to user."
+                )
+            q = q.filter(Source.group_id == int(group_id))
```
acrellin, line 300: This should be a list, so the filter expression will change slightly here

skyportal/handlers/api/source.py
```diff
@@ -274,13 +290,20 @@ def get(self, obj_id=None):
...
+            # return self.error(str(self.current_user.accessible_groups))
```
acrellin, line 301:
```suggestion
```

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,250 @@
...
+        <Link href={`/api/sources/${source.id}/finder`}>
+          <PictureAsPdfIcon />
+        </Link>
+      </IconButton>
+    );
+  };
+
+  const columns = [
+    {
+      name: "Source Name",
```
acrellin, line 151:
```suggestion
      name: "Source ID",
```

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,250 @@
...
+import PropTypes from "prop-types";
+import { useSelector, useDispatch } from "react-redux";
+
+import TableCell from "@material-ui/core/TableCell";
+import TableRow from "@material-ui/core/TableRow";
+
+import Typography from "@material-ui/core/Typography";
+import IconButton from "@material-ui/core/IconButton";
+import Grid from "@material-ui/core/Grid";
+// import BuildIcon from "@material-ui/icons/Build";
```
acrellin, line 11: Remove these commented out lines here and below

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,250 @@
...
+  }, [route.id, dispatch]);
+
+  if (sources === undefined || sources === null) {
+    return "Loading Sources...";
+  }
+
+  const group_id = parseInt(route.id, 10);
+  let group_name = "";
+
+  if (groups === undefined || groups === null) {
```
acrellin, line 47: Invert the condition and remove this empty block

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,250 @@
...
+const VegaPlot = React.lazy(() => import("./VegaPlot"));
+
+const GroupSources = ({ route }) => {
+  const dispatch = useDispatch();
+  const sources = useSelector((state) => state.sources.latest);
+  const groups = useSelector((state) => state.groups.user);
+
+  // Load the group sources
+  useEffect(() => {
+    dispatch(SourcesAction.fetchSources({ group_id: route.id }));
```
acrellin, line 37: This could have side effects for other components that use `state.sources.latest` that don't re-fetch when they're rendered.

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,250 @@
...
+    return "Loading Sources...";
+  }
+
+  const group_id = parseInt(route.id, 10);
+  let group_name = "";
+
+  if (groups === undefined || groups === null) {
+    // nothing, just wait
+  } else {
+    const group_obj = groups.find((obj) => {
```
acrellin, line 50:
```suggestion
    const group = groups.find((group) => {
```
This is very confusing variable naming. Obj is a meaningful variable name, don't use that for a group instance.

skyportal/handlers/api/source.py
acrellin: Should be obvious from the indentation and return statement, or can add a comment like this 
```suggestion
        # Fetch multiple sources
```

static/js/components/GroupSources.jsx
guynir42: I thought all components update as soon as the store is changed... maybe I got that wrong. But what is the solution, if I want to re-use the existing `fetchSources`?

acrellin: Yes, what I'm saying is that other components (specific `SourceList`) might expect `state.sources.latest` to initially be an un-filtered list (e.g. see `fetchSources` in `hydrate`) -- if so, that component will need to be updated to re-fetch upon render, otherwise it will display this filtered source list. Does that make sense?

acrellin: @guynir42 actually, I mis-wrote, the fetch doesn't occur in hydrate. But look at this check in `SourceList`: 
``` jsx
    if (!sources.latest) {
      dispatch(sourcesActions.fetchSources());
    }
```
This will obviously need to be updated so that it re-fetches upon every render. Another solution is just to add a new `groupSources` branch to the front-end state that this new component uses, so it won't mess up the `state.sources` branch that other components depend on.

skyportal/handlers/api/source.py
guynir42: You and I are in an excellent place to say what is obvious to someone who already knows this code very well and to someone who is looking at it for the first time... 

That being said, I've changed the comment to what you suggested!

package.json
guynir42: Honestly I don't know why there's a change in this file, I didn't modify it...

static/js/components/GroupSources.jsx
guynir42: Yes, this is true in astronomy in general when doing any OOP and looking at celestial objects. I'll change it.

skyportal/handlers/api/source.py
guynir42: ok, formatted strings. Got it.

skyportal/handlers/api/source.py
guynir42: yes, done

skyportal/handlers/api/source.py
guynir42: ok, done

skyportal/handlers/api/source.py
guynir42: and done

skyportal/handlers/api/source.py
guynir42: done

static/js/components/GroupSources.jsx
guynir42: done

static/js/components/GroupSources.jsx
guynir42: done

static/js/components/GroupSources.jsx
guynir42: yes you are right. done

static/js/components/GroupSources.jsx
guynir42: I can see there are a few ways to go about this. We should discuss it.

acrellin: I think the best approach is the latter noted above - either add a new branch to the front end state called `groupSources`, or another leaf to the `sources` branch with the same name.

acrellin: Probably the latter (add a new leaf `state.sources.groupSources`)

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+  useEffect(() => {
+    dispatch(SourcesAction.fetchGroupSources({ group_id: route.id }));
+  }, [route.id, dispatch]);
+
+  if (sources === undefined || sources === null) {
+    return "Loading Sources...";
+  }
+
+  const group_id = parseInt(route.id, 10);
+  let group_name = "";
```
dannygoldstein, line 38:
```suggestion
```

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+      return g.id === group_id;
+    });
+
+    if (group === undefined || group === null) {
+      // couldn't find a group object
+      group_name = "";
+    } else {
+      group_name = group.name;
+    }
+  }
```
dannygoldstein, line 52:
```suggestion
   const group_name = group?.name
```

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+
+  // This is just passed to MUI datatables options -- not meant to be instantiated directly.
+  const renderPullOutRow = (rowData, rowMeta) => {
+    if (sources === undefined) {
```
dannygoldstein, line 56:
```suggestion
    if (!sources) {
```

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+    expandableRows: true,
+    renderExpandableRow: renderPullOutRow,
+    selectableRows: "none",
+  };
+
+  const data = sources.map((source) => [
+    source.id,
+    source.ra,
+    source.dec,
+    source.redshift,
```
dannygoldstein, line 188: Can we add some more info here? Groups saved to? Classifications? Recent comments?

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+              >
+                <em>Group Sources:</em>
+              </Typography>
+              <Typography
+                variant="h4"
+                gutterBottom
+                color="textSecondary"
+                align="center"
+              >
+                <b>Sources for {group_name}</b>
```
dannygoldstein, line 217:
```suggestion
                <b>Sources saved to {group_name}</b>
```

config.yaml.defaults
```diff
@@ -46,7 +46,8 @@ app:
       component: RunSummary
     - path: "/share_data/:id"
       component: ShareDataForm
-
+    - path: "/group_sources/:id"
```
dannygoldstein, line 49: Can we add a component for this page to render in the sidebar?

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+          <Grid item>
+            <div>
+              <Typography
+                variant="h4"
+                gutterBottom
+                color="textSecondary"
+                align="center"
+              >
+                <em>Group Sources:</em>
+              </Typography>
```
dannygoldstein, line 210:
```suggestion
```

skyportal/handlers/api/source.py
```diff
@@ -274,13 +292,23 @@ def get(self, obj_id=None):
             )
         if has_tns_name in ['true', True]:
             q = q.filter(Obj.altdata['tns']['name'].isnot(None))
+        if group_id:
+            if not all(g.isdigit() for g in group_id):
```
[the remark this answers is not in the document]
guynir42: You are right. Fixed to accept numbers or strings only.

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+  const dispatch = useDispatch();
+  const sources = useSelector((state) => state.sources.groupSources);
+  const groups = useSelector((state) => state.groups.user);
+
+  // Load the group sources
+  useEffect(() => {
+    dispatch(SourcesAction.fetchGroupSources({ group_id: route.id }));
+  }, [route.id, dispatch]);
+
+  if (sources === undefined || sources === null) {
```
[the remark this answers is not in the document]
guynir42: ok

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,241 @@
...
+    return "Loading Sources...";
+  }
+
+  const group_id = parseInt(route.id, 10);
+  let group_name = "";
+
+  if (groups !== undefined && groups !== null) {
```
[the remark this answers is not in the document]
guynir42: yes I see a pattern here...

static/js/components/GroupSources.jsx
guynir42: yes, indeed

guynir42: No, actually I need this in case groups is not defined.

static/js/components/GroupSources.jsx
guynir42: What kind of sorcery is this??

static/js/components/GroupSources.jsx
guynir42: ok

static/js/components/GroupSources.jsx
guynir42: yes good idea

config.yaml.defaults
guynir42: I don't know what that means...

static/js/components/GroupSources.jsx
acrellin: This still works if group is undefined. Pretty neat, eh? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

dmitryduev:
@guynir42 For PRs that implement something on the frontend, I find very useful a quick screen capture on screenshot in the PR description.

guynir42:
@acrellin Sorry I couldn't work on this in the last few days. I will add a few more features requested by danny and then I will ping you guys for a second review.

static/js/components/GroupSources.jsx
guynir42: Ok this is now added to sources.groupSources in the state tree.

static/js/components/GroupSources.jsx
guynir42: I added groups and comments (do we want to limit the number of displayed comments or just let the user scroll down?). 

I am having some trouble add classifications to sources, so I'm not sure if my addition of classifications to the table is actually working.

skyportal/handlers/api/source.py
```diff
@@ -243,17 +254,26 @@ def get(self, obj_id=None):
...
+
+        # Fetch multiple sources
+        user_group_ids = [g.id for g in self.current_user.accessible_groups]
+        q = (
+            DBSession()
+            .query(Obj)
+            .join(Source)
+            .filter(
+                Source.group_id.in_(  # only give sources the user has access to
+                    [g.id for g in self.current_user.accessible_groups]
```
dannygoldstein, line 266:
```suggestion
                    user_group_ids
```

skyportal/handlers/api/source.py
```diff
@@ -285,13 +305,25 @@ def get(self, obj_id=None):
             )
         if has_tns_name in ['true', True]:
             q = q.filter(Obj.altdata['tns']['name'].isnot(None))
+        if group_id:
+            if not all(
+                isinstance(g, numbers.Number) or isinstance(g, str) and g.isdigit()
```
dannygoldstein, line 310: I'd recommend using a `try/except` block here, something like

```
for id in group_ids:
  try:
    gid = int(id)
  except: 
    return self.error(f'Invalid group id: {id}")
```

skyportal/handlers/api/source.py
```diff
@@ -305,11 +337,28 @@ def get(self, obj_id=None):
...
+                        DBSession()
+                        .query(Source.group_id)
+                        .filter(Source.obj_id == source_list[-1]["id"])
+                    )
+                )
+                .filter(
+                    Group.id.in_([g.id for g in self.current_user.accessible_groups])
+                )
+                .all()
+            )
```
dannygoldstein, line 361:
```suggestion
            source_list[-1]["groups"] = DBSession().query(Group).join(Source).filter(Source.obj_id == source_list[-1]["id"], Group.id.in_(user_accessible_groups)).all()
```

static/js/components/GroupSources.css
```diff
@@ -0,0 +1,153 @@
...
+.comments {
+}
+
+.tableContainer {
+  margin: 0.625rem;
+}
+
+.commentsList {
+  overflow-y: scroll;
+  max-height: 350px;
```
dannygoldstein, line 53: use relative sizes everywhere and replace .css files with `makeStyles`

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,349 @@
...
+import { ra_to_hours, dec_to_hours } from "../units";
+import * as SourcesAction from "../ducks/sources";
+import styles from "./GroupSources.css";
+import ThumbnailList from "./ThumbnailList";
+import UserAvatar from "./UserAvatar";
+import ShowClassification from "./ShowClassification";
+
+const VegaPlot = React.lazy(() => import("./VegaPlot"));
+
+const useStyles = makeStyles((theme) => ({
```
dannygoldstein, line 32: incorporate CSS into this

static/js/ducks/sources.js
```diff
@@ -14,8 +18,15 @@ export function fetchSources(filterParams = {}) {
...
 
+export function fetchGroupSources(filterParams = {}) {
+  const params = new URLSearchParams(filterParams);
+  const queryString = params.toString();
+  return API.GET(`/api/sources?${queryString}`, FETCH_GROUP_SOURCES);
+}
+
 const initialState = {
   latest: null,
+  groupSources: null,
```
dannygoldstein, line 29: do you need any of the other params here? pageNumber, lastPage, etc.? are those ever used?

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,349 @@
...
+  if (groups) {
+    const group = groups.find((g) => {
+      // find the object for the group
+      return g.id === group_id;
+    });
+    group_name = group?.name;
+  }
+
+  // This is just passed to MUI datatables options -- not meant to be instantiated directly.
+  const renderPullOutRow = (rowData, rowMeta) => {
```
dannygoldstein, line 73: Can you add the light curves and spectra here? There should be a Vegaplot component for the light curves. Can you add a VegaPlot for the spectra? Should be fairly straightforward, just need to add a line plot.

dannygoldstein (changes requested):
Getting closer.

config.yaml.defaults
guynir42: I thought the natural thing is to add a link from the group page...

skyportal/handlers/api/source.py
guynir42: Yes, also replaced a similar line above that (moved the definition of `user_group_ids` up to the beginning.

skyportal/handlers/api/source.py
guynir42: Ok!

skyportal/handlers/api/source.py
guynir42: Good, and also I've changed a similar part on line 242 (same thing, only for single source).

static/js/ducks/sources.js
guynir42: I think they are used in the sources list page. We are using the same reducer with a different case for the group sources page.

config.yaml.defaults
dmitryduev: I agree with @guynir42

guynir42:
> @guynir42 For PRs that implement something on the frontend, I find very useful a quick screen capture on screenshot in the PR description.

@dmitryduev sorry this took so long, I've added a screenshot of the page when it loads and when you click each of the lines to see more info. The spectra are still missing (working on that).

dmitryduev:
Excellent, thanks, @guynir42! Looks good! One q: what are those empty photometry plots on the second screenshot?

guynir42:
> Excellent, thanks, @guynir42! Looks good! One q: what are those empty photometry plots on the second screenshot?

These will be spectroscopy plots, as soon as I get them working!

dannygoldstein:
@guynir42 Merge conflicts

static/js/components/GroupSources.jsx
guynir42: VegaPlot works by default with photometry data. To get it to show spectra will require a change to that component. I think we should do that separately and then add the spectrum plots to this page.

static/js/components/GroupSources.jsx
guynir42: Is there any other page where the comments CSS have been incorporated into the makeStyles? I could really use a working example.

guynir42:
@dannygoldstein  look in test_group_sources.py, the failure is on line 40. The source appears but I cannot find the link to the source.

dannygoldstein:
@guynir42 The test is passing for me locally. Try resolving your merge conflicts, then the CI will run and it should probably pass.

dannygoldstein:
@guynir42 I'm not sure why its failing on CI and passing locally. Can you add some screenshots into the test and upload them as artifacts on github actions? See #893. If you follow what's done there, the screenshots must be named `test*png` in order to be uploaded automatically to github actions

guynir42:
@dannygoldstein @acrellin This is finally green and everything is working. Please re-review...

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,351 @@
...
+    if (!sources) {
+      return "Loading...";
+    }
```
acrellin, line 78:
```suggestion
```
duplicate

static/js/ducks/sources.js
```diff
@@ -14,8 +18,15 @@ export function fetchSources(filterParams = {}) {
   return API.GET(`/api/sources?${queryString}`, FETCH_SOURCES);
 }
 
+export function fetchGroupSources(filterParams = {}) {
+  const params = new URLSearchParams(filterParams);
```
acrellin, line 22: Check whether group_ids is present as expected

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,351 @@
...
+  const userColorTheme = useSelector(
+    (state) => state.profile.preferences.theme
+  );
+
+  const commentStyle =
+    userColorTheme === "dark" ? styles.commentDark : styles.comment;
+
+  // Load the group sources
+  useEffect(() => {
+    dispatch(SourcesAction.fetchGroupSources({ group_id: route.id }));
```
acrellin, line 56:
```suggestion
    dispatch(SourcesAction.fetchGroupSources({ group_ids: [route.id] }));
```

acrellin (changes requested):
Getting closer @guynir42 , thanks for the great work! A few more requested changes below. Let me know if you have questions about any of my comments.

acrellin:
@guynir42 there is no link the group sources page on the group page. Currently there's no way to navigate to it. Please add a link to the group sources page somewhere on the group page.
image

static/js/components/GroupSources.jsx
dmitryduev: Let's replace it with a spinner?
Import:
`import CircularProgress from "@material-ui/core/CircularProgress";`

```suggestion
    if (!sources) {
      return (
        <div>
          <CircularProgress color="secondary" />
        </div>
      );
    }
```

skyportal/handlers/api/source.py
```diff
@@ -310,6 +341,11 @@ def get(self, obj_id=None):
 
         source_list = []
         for source in query_results["sources"]:
+            comments = source.get_comments_owned_by(self.current_user)
+            source.comments = sorted(comments, key=lambda x: x.created_at, reverse=True)
+            source.classifications = source.get_classifications_owned_by(
+                self.current_user
+            )
```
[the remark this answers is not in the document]
guynir42: Ok but I am still not sure what is the difference between the two approaches. I've seen both used in the code so I wasn't sure.

dmitryduev:
> @guynir42 there is no link the group sources page on the group page. Currently there's no way to navigate to it. Please add a link to the group sources page somewhere on the group page.
> image

I'd add another another accordion above Members, call it "Sources" and add the link there; we may later expand it do display some high-level stats for that group. Also pie charts, bc how come we don't have a single pie chart anywhere? Just kidding about the latter :)

skyportal/handlers/api/source.py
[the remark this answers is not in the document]
acrellin: We ran into some permissions issues regarding source/candidate comments last week and the fix was to simply modify the associated dict and the `Obj` instance properties. This is just safer for relationships (probably not an issue for static properties)

static/js/components/GroupSources.jsx
guynir42: Cool!

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,351 @@
...
+  const group_id = parseInt(route.id, 10);
+  let group_name = "";
+
+  if (groups) {
+    const group = groups.find((g) => {
+      // find the object for the group
+      return g.id === group_id;
+    });
+    group_name = group?.name;
+  }
```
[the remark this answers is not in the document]
guynir42: Even more sorcery :)

static/js/ducks/sources.js
guynir42: What's the right way to raise an exception if the group_id parameter is not there?

acrellin: this is actually probably fine to leave as-is

static/js/components/GroupSources.jsx
guynir42: Ok but why?

acrellin: Because group_ids should always be a list. see my comments/suggestions elsewhere

guynir42: I agree. But I think in this case it doesn't matter because the URLSearchParams turns it into a string that then gets parsed on the back-end. Otherwise, I can't understand how it worked for me until now...

acrellin: get_query_argument converts it to a list
If you were just passing a string like "2" before, it was likely only working because you were iterating over each character of the string, and each character happened to be a valid group ID.

guynir42: Yes, that was the problem I had with the test, but I think that if I send a string to URLSearchParams and then call toString() on that I will get a string with comma-separated values. And I am parsing for that on the back-end. That's why I don't think it matters much... But I certainly agree it looks more clear.

guynir42:
All comments by @acrellin have been addressed. Still waiting for second reviewer (@dannygoldstein )

skyportal/tests/frontend/test_group_sources.py
```diff
@@ -0,0 +1,60 @@
+import uuid
+
+from .. import api
+from selenium.common.exceptions import TimeoutException
+
+
+def test_add_new_source(driver, super_admin_user, public_group, upload_data_token):
```
dannygoldstein, line 7:
```suggestion
def test_add_new_source_renders_on_group_sources_page(driver, super_admin_user, public_group, upload_data_token):
```

static/js/components/Group.jsx
```diff
@@ -231,6 +237,26 @@ const Group = () => {
...
+
+      <Accordion
+        expanded={panelSourcesExpanded === "panel-sources"}
+        onChange={handlePanelSourcesChange("panel-sources")}
+      >
+        <AccordionSummary
+          expandIcon={<ExpandMoreIcon />}
+          aria-controls="panel-sources-content"
+          id="panel-sources-header"
+          style={{ borderBottom: "1px solid rgba(0, 0, 0, .125)" }}
```
dannygoldstein, line 249: Should be `makeStyles`

dannygoldstein (approved):

skyportal/handlers/api/source.py
```diff
@@ -178,6 +187,19 @@ def get(self, obj_id=None):
         start_date = self.get_query_argument('startDate', None)
         end_date = self.get_query_argument('endDate', None)
         sourceID = self.get_query_argument('sourceID', None)  # Partial ID to match
+
+        # parse the group ids:
+        group_ids = self.get_query_argument('group_id', None)
```
acrellin, line 192:
```suggestion
        group_ids = self.get_query_argument('group_ids', None)
```

static/js/components/GroupSources.jsx
```diff
@@ -0,0 +1,346 @@
...
+  const userColorTheme = useSelector(
+    (state) => state.profile.preferences.theme
+  );
+
+  const commentStyle =
+    userColorTheme === "dark" ? styles.commentDark : styles.comment;
+
+  // Load the group sources
+  useEffect(() => {
+    dispatch(SourcesAction.fetchGroupSources({ group_id: [route.id] }));
```
acrellin, line 57:
```suggestion
    dispatch(SourcesAction.fetchGroupSources({ group_ids: [route.id] }));
```

acrellin (changes requested):
@guynir42 looking great. Just couple more API field names need updating, then :+1: from me

dannygoldstein:
Congratulations @guynir42 , this was a big one!

Issue #604, opened 2020-07-27
Auto-colapse sidebar on sm breakpoint


dmitryduev:

dannygoldstein:
Fixed by #766

PR #766, opened 2020-08-31
Close sidebar by default on mobile

kmshin1397:
Issue #604

### DIFF
--- a/static/js/ducks/sidebar.js
+++ b/static/js/ducks/sidebar.js
@@ -8,8 +8,10 @@ export function toggleSidebar() {
   };
 }
 
-// TODO Set open to false by default on mobile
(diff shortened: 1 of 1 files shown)


kmshin1397:
Oops, I seem to have accidentally closed the sidebar for all breakpoints. Debugging right now

kmshin1397:
> Looks good, just a quick q: what about doing something like the following:
> 
> ```
> import useMediaQuery from "@material-ui/core/useMediaQuery";
> const isMobile = !useMediaQuery(theme.breakpoints.up("md"));
> ```

I tried useMediaQuery but React threw an error about out of order hooks or something similarly breaking like that so I went with this instead.

PR #778, opened 2020-09-01
Sidebar and AppBar tweaks for mobile

kmshin1397:
- Avatar + Username is now replaced by just the Avatar on mobile for the profile dropdown
- The sidebar automatically closes when the window is resized to a small breakpoint
- Got rid of minimum width for SourceList component that made it extend far beyond mobile viewport widths

sidebar


static/js/components/SearchBox.jsx
```diff
@@ -88,7 +88,7 @@ const SearchBox = ({ sources }) => {
     root: {
       display: "flex",
       flexWrap: "wrap",
-      width: "35rem",
+      // width: "35rem",
```
stefanv, line 91:
```suggestion
```

PR #781, opened 2020-09-01
Add FETCH_SOURCES WS handler

kmshin1397:
FETCH_SOURCES WS messages were not being handled by the sources.js duck

### DIFF
--- a/static/js/ducks/sources.js
+++ b/static/js/ducks/sources.js
@@ -1,3 +1,5 @@
+import messageHandler from "baselayer/MessageHandler";
+
 import * as API from "../API";
 import store from "../store";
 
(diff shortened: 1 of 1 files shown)


acrellin:
@kmshin1397 I think we want to preserve any filter param values if a user happens to be on the source list page and is viewing search results -- in that case we can just re-submit their search form from this WS message handler. I had implemented that in the past, but we took it out for various reasons. @stefanv does that sound reasonable?

acrellin:
In the case of many sources being POSTed in rapid succession via API calls, we'll also want to throttle this somehow, otherwise someone running a batch POST script will generate way too many front-end fetch requests.

stefanv:
Yes, I think this may cause too much churn on the frontend; if you do a search, you get what you get, and next time you do it again you get the newer results.

acrellin:
@stefanv so are you proposing this be closed?

profjsb (approved):

stefanv:
Yes, I think this may introduce too many API requests.

PR #786, opened 2020-09-01
Put candidate annotations within scrollbars

kmshin1397:
annotations


profjsb:
Can we also make comment entries more compact -- there's utility in not needing to scroll a lot to see entries. Currently there's a lot of whitespace between entries and some wasted vertical space in the div.

kmshin1397:
Screenshot 2020-09-01 161111
How does that look?

stefanv:
Looks good!

PR #788, opened 2020-09-01
Add previous/next page buttons below candidate list & show spinner when query is in progress

acrellin:
Closes https://github.com/skyportal/skyportal/issues/784
Closes https://github.com/skyportal/skyportal/issues/785

scanning_page_spinner


stefanv:
> Do you think we still need the previous/next page buttons on the FilterCandidateList form if we now have it on the table? I guess once you've scrolled down to the bottom of the table, the form is hidden so it's probably a good thing. Otherwise, looks good to me!

Yes, I requested it for that reason: the bottom of the page is quite confusing otherwise.

stefanv:
Thanks, Ari!

PR #790, opened 2020-09-02
Make home page widgets only draggable via handle

kmshin1397:
drag


dmitryduev:
The code looks good, thanks @kmshin1397! 
The top right corner looks a bit messy with this bc the icons are not aligned (even though I guess it's intentional). Could you please try implementing something like
image
(as always, if not too much trouble)

static/js/components/HomePage.jsx
```diff
@@ -87,17 +89,31 @@ const HomePage = () => {
       cols={{ xlg: 16, lg: 12, md: 10, sm: 6, xs: 4 }}
       margin={[15, 15]}
       onLayoutChange={LayoutChangeHandler}
+      draggableHandle=".dragHandle"
     >
       <div key="sourceList" className={styles.homePageWidgetDiv}>
+        <DragHandleIcon className={`${styles.dragHandle} dragHandle`} />
```
dannygoldstein, line 95: Can these `DragHandleIcon`s be incorporated directly into the corresponding components?

kmshin1397:
Refactored a bit so that the Home Page defines general styles to pass along to child widgets, who use them when making the Paper, drag handler, and other icons (i.e. preferences).

Screenshot 2020-09-01 190415

dmitryduev:
Looks awesome, thanks! Merging.

PR #812, opened 2020-09-02
Ensure state.groups.all includes single user groups upon ShareDataForm render

acrellin:
The issue was that sometimes the `fetchGroups()` call in `hydrate` (which does not fetch single user groups) was returning after the `fetchGroups(includeSingleUserGroups=true)` call in `ShareDataForm`. Now, when the component initially renders and when the groups list is updated, the groups list is checked, and if no single user groups are present, `fetchGroups(includeSingleUserGroups=true)` is called.

Closes https://github.com/skyportal/skyportal/issues/787

### DIFF
--- a/static/js/actions.js
+++ b/static/js/actions.js
@@ -17,7 +17,7 @@ export default function hydrate() {
     dispatch(sysInfoActions.fetchSystemInfo());
     dispatch(dbInfoActions.fetchDBInfo());
     dispatch(profileActions.fetchUserProfile());
-    dispatch(groupsActions.fetchGroups());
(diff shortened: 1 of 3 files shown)


acrellin:
The issue was that sometimes the `fetchGroups` in `hydrate` (which does not fetch single user groups) was returning _after_ the `fetchGroups(includeSingleUserGroups=true)` call in the `ShareDataForm`. Now, when the page loads or when the groups list is updated, the groups list is checked and if no single user groups are present, `fetchGroups(includeSingleUserGroups=true)` is called.

PR #814, opened 2020-09-03
Update homepage layouts

kmshin1397:
- Update default layouts to better fit in new widgets
- Move layouts into config YAML and make home page based off a template file
- Make home page layouts respond to resizing from closing/opening the sidebar
- Update source counts widget styling to give more room for longer "last N days" values
layouts


config.yaml.defaults
```diff
@@ -70,6 +70,163 @@ app:
       icon: Info
       url: /about
 
+  homepage_layouts:
```
profjsb, line 73: I think it makes sense to establish the defaults for each widget, so that we can add extra parameters. Most important is a `show` toggle (since not all SP instances will want to show all these widgets):
```yaml
homepage_layouts:
  sourceCounts:
      show: true
      days_past_to_include: 7
      resizeable: false
      layout: 
         xlg: [14, 0, 2, 1]
         lg: ...
```

profjsb (changes requested):
This looks FANTASTIC. I'm suggesting a refactor in the YAML which will make it more compact, more readable, and more extensible.

kmshin1397:
@profjsb Did the YAML refactor you suggested. Also some additional styling (overflows) and added a reset to default layout button.

static/js/components/HomePage.jsx.template
```diff
@@ -0,0 +1,197 @@
...
+    position: "relative",
+    float: "right",
+    margin: "0.5rem",
+  },
+}));
+
+{%- set ns = namespace(xlg=[], lg=[], md=[], sm=[], xs=[]) -%}
+{%- for widgetName, widgetProps in app.homepage_widgets.items() %}
+  {%- if widgetProps.show %}
+    {%- set ns.xlg = [{
```
stefanv, line 52: Use a for loop to generate repeating items like this.

static/js/components/HomePage.jsx.template
```diff
@@ -0,0 +1,197 @@
...
+    dispatch(profileActions.updateUserPreferences(prefs));
+  };
+
+  return (
+    <div>
+      <ResponsiveGridLayout
+        className="layout"
+        layouts={currentLayouts}
+        breakpoints={{ '{{ xlg: 1400, lg: 1150, md: 996, sm: 650, xs: 0 }}' }}
+        cols={{ '{{ xlg: 16, lg: 12, md: 10, sm: 6, xs: 4 }}' }}
```
stefanv, line 146: This should probably appear in the config too?

stefanv (commented):
This looks great, thanks @kmshin1397.

Can you look at the templating again and see whether you can polish that up a bit?  Otherwise I think this is ready.

static/js/components/HomePage.jsx.template
kmshin1397: @stefanv I tried for awhile yesterday to get it to work with for loops but couldn't get past Jinja limitations. Specifically, I couldn't figure out a way to access namespace variables via variable keys, i.e.
Screenshot 2020-09-03 183740
^ and variations from that didn't work at all because Jinja didn't like the bracket syntax at all on the namespace for some reason

If you have any suggestions on how to use the for loops, that'd be great.

profjsb (approved):
Looks great. I agree with all of @stefanv's comments. Will leave my approval here so that we can merge after you address his PR comments.

config.yaml.defaults
```diff
@@ -75,6 +75,72 @@ app:
...
+  homepage_widgets:
+    # Layouts arrays should be [x, y, w, h]
+    # See https://github.com/STRML/react-grid-layout
+
+    # component_props are properties passed down to the
+    # underlying React component, indicated by component_name
+
+    sourceCounts:
+      show: true
+      component_name: SourceCounts
```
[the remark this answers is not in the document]
kmshin1397: Yeah, it was mainly a convenience thing because I couldn't find a quick and easy way to capitalize the first letter in the template to get to the component name from the section name (the builtin capitalize filter also makes the rest of the string lowercase). But in retrospect, the section names are only used as CSS id values so they could easily be capitalized in the configs to get around that.

kmshin1397:
@stefanv Suggestions addressed. I think it's just newsfeed prefs failing

static/js/components/HomePage.jsx.template
```diff
@@ -46,59 +46,15 @@ const useStyles = makeStyles(() => ({
...
-{% endfor %}
-
 const defaultLayouts = {
-  xlg: {{ ns.xlg | replace("True", "true") | replace("False", "false") }},
-  lg: {{ ns.lg | replace("True", "true") | replace("False", "false") }},
-  md: {{ ns.md | replace("True", "true") | replace("False", "false") }},
-  sm: {{ ns.sm | replace("True", "true") | replace("False", "false") }},
-  xs: {{ ns.xs | replace("True", "true") | replace("False", "false") }},
-};
+  {%- for breakpoint in app.homepage_grid.breakpoints.keys() %}
```
profjsb, line 50: nice!

config.yaml.defaults
```diff
@@ -75,17 +75,67 @@ app:
...
+    #
+    # The breakpoints describe screen sizes at which a different set of widget
+    # layouts should be used.
+    #
+    # The cols describe the number of evenly spaced columns that make up the
+    # grid at a given breakpoint. For example, on extra-large screens (greater
+    # than ${breakpoints.xlg} pixels), the grid will use ${cols.xlg} columns of
+    # equal width to describe sizes of widgets based on the layouts provided.
+    #
+    # Optionally, you may provide a "row_height: {number of pixels}" property in
```
stefanv, line 90: Why is the row height set to 150px?  Can we set it in rem?

config.yaml.defaults
```diff
@@ -75,17 +75,67 @@ app:
...
+    # this section to change the height of a row on the grid. By default, this
+    # value is 150px.
+
+    breakpoints:
+      xlg: 1400
+      lg: 1150
+      md: 996
+      sm: 650
+      xs: 0
```
stefanv, line 99: Is it necessary to specify the 0 point (xs) ?

stefanv: I guess it's confusing: anything >0 should fit within the next higher up bracket?

config.yaml.defaults
```diff
@@ -75,17 +75,67 @@ app:
...
+      md: 10
+      sm: 6
+      xs: 4
 
-    sourceCounts:
-      show: true
-      component_name: SourceCounts
-      component_props:
+  homepage_widgets:
+    # This section describes the specific widgets shown on the Home Page and how
```
stefanv, line 109: Very nice description, thanks.

static/js/components/HomePage.jsx.template
```diff
@@ -46,59 +46,15 @@ const useStyles = makeStyles(() => ({
...
+  {{ breakpoint }}: [
```
stefanv, line 51: Super!

static/js/components/HomePage.jsx.template
```diff
@@ -46,59 +46,15 @@ const useStyles = makeStyles(() => ({
...
+    {%- for widgetName, widgetProps in app.homepage_widgets.items() %}
+    { i: "{{ widgetName }}", x: {{ widgetProps.layouts[breakpoint][0] }}, y: {{ widgetProps.layouts[breakpoint][1] }}, w: {{ widgetProps.layouts[breakpoint][2] }}, h: {{ widgetProps.layouts[breakpoint][3] }}, isResizable: {{ widgetProps.resizeable | lower }} },
```
stefanv, line 53: Some newlines here may help with readability, but not strictly necessary.

stefanv (commented):
Looks great, thanks @kmshin1397

config.yaml.defaults
kmshin1397: I used pixels because that's what the react-grid-layout package takes. But yeah, I guess we could take in rem and just convert it?

config.yaml.defaults
kmshin1397: Ah yeah, this is confusing but I noticed the breakpoints passed to react-grid-layout aren't quite the same as breakpoints for, say, MUI. These are min-width values instead of max-width values. So the small breakpoint would apply to any window (or rather, content divs) from 650 up to 996, the xs breakpoint would match with windows 0 up to 650, etc. Not sure if I'm missing something, but that's what I see when I hook up to the onBreakpointChange and print out the breakpoints it's using.

static/js/components/HomePage.jsx.template
kmshin1397: Adding newlines here would transfer to the generated JSX like this:
Screenshot 2020-09-03 215039

I thought it made the generated file unnecessarily long (adds like 125 lines extra). Do we want to prioritize readability on the template or the generated JSX?

config.yaml.defaults
stefanv: We can use the same utility as MUI: https://www.npmjs.com/package/convert-css-length

See https://github.com/mui-org/material-ui/pull/14573/files#diff-c3e0d8891a5bacebd344a4f81ff36539R36 and the next 10 lines.

config.yaml.defaults
stefanv: OK, may be worth pointing that out, since it works differently in MUI where we specify the lower boundaries for transitions.

static/js/components/HomePage.jsx.template
stefanv: Not a big deal; you can leave as is.

stefanv (approved):

PR #815, opened 2020-09-03
Filter page

dmitryduev:
This PR introduces a basic Filter page and adds a corresponding frontend test. [This gets replaced in F]

sp-filters-01


skyportal/tests/frontend/test_filters.py
```diff
@@ -0,0 +1,49 @@
...
+    driver.execute_script("arguments[0].click();", add_stream)
+
+    # add filter
+    filter_name = str(uuid.uuid4())
+    driver.wait_for_xpath_to_be_clickable(
+        f'//button[contains(.,"Add filter")]', timeout=10
+    )
+    flt = driver.switch_to.active_element
+    flt.click()
+    driver.wait_for_xpath(f'//button[contains(.,"Add filter")]').click()
```
kmshin1397, line 32: Consider using the driver.click_xpath() functions from baselayer where possible in this test instead? We had issues with MUI buttons before right? Wouldn't hurt I think.

static/js/components/Group.jsx
```diff
@@ -335,17 +335,20 @@ const Group = () => {
...
                   <List component="nav" disablePadding>
                     {group.filters?.map((filter) =>
                       filter.stream_id === stream.id ? (
-                        <ListItem button component="a" key={filter.id} href="#">
-                          <ListItemText
-                            className={classes.nested}
-                            primary={filter.name}
-                          />
+                        <ListItem button key={filter.id}>
+                          <Link to={`/filter/${filter.id}`}>
```
kmshin1397, line 345:
```suggestion
                          <Link to={`/filter/${filter.id}`} className={classes.filterLink}>
```
And define filterLink above to have `width: "100%"`. Right now, the link only exists on the text of the filter name and not along the whole row in the list, but the whole row is highlighted on hover, leading to non-intuitive behavior.

static/js/components/Filter.jsx
```diff
@@ -0,0 +1,152 @@
...
+    return (
+      <div>
+        <CircularProgress color="secondary" />
+      </div>
+    );
+  }
+
+  return (
+    <div>
+      <Typography variant="h6" style={{ paddingBottom: 10 }}>
```
kmshin1397, line 119: Inline styling should be moved to the makeStyles()

kmshin1397 (changes requested):

skyportal/tests/frontend/test_filters.py
kmshin1397: Yeah, seems like the test failed on Travis so this is probably a good idea.

dmitryduev: Switched to using `driver.click_xpath()` everywhere where it didn't cause a failure locally, will see what the CI has to say...

static/js/components/Group.jsx
dmitryduev: Nice catch, thanks! Fixed

static/js/components/Filter.jsx
dmitryduev: Fixed

skyportal/tests/frontend/test_filters.py
dmitryduev: @dannygoldstein discovered a nice trick using `.pause(2)` to allow for MUI transitions to finish, but I didn't get a chance to put it in before this got merged 😂 will do in a next pr ;-)

PR #816, opened 2020-09-03
use & to separate parameters in URL

profjsb:
A user without a gravatar will not cause an insecure URL call.


dmitryduev (approved):
That was it, nice catch @profjsb!

PR #820, opened 2020-09-03
Disable transitions to help with testing

stefanv:
Let's see how this goes.


stefanv:
This has now been updated to only be enacted on testing.

dannygoldstein (approved):
Failures unrelated

PR #823, opened 2020-09-03
Update News Feed and Preferences widget

kmshin1397:
I noticed that the newsfeed GET API was returning duplicate items for sources added because each source gets a row for every group it is a part of in the Sources table. This PR makes it so the API filters for distinct source additions by time.


kmshin1397:
Screenshot 2020-09-03 103727

skyportal/tests/frontend/test_newsfeed.py
```diff
@@ -1,6 +1,6 @@
 import uuid
 import pytest
-
+from selenium.webdriver.common.action_chains import ActionChains
```
dmitryduev, line 3: Neat, going to start using this!

dmitryduev (approved):
LGTM! `test_news_feed_prefs_widget` is flaking - have you looked into it?

skyportal/tests/frontend/test_newsfeed.py
kmshin1397: Yeah starting to really look into it now... Using ActionChains fixed some flakiness in test_groups for me yesterday but doesn't seem like it's helping here...

kmshin1397:
Okay, this was originally just a quick tweak to the backend news feed query, but in the process of trying to get test_news_feed_prefs_widget green I ended up doing some reskinning of the preferences widget as well.

Basically, I couldn't get it green so I tried just refactoring the preferences to use MUI components and that at least got Travis to green (without using .pause() either).

New preferences widget look:
Screenshot 2020-09-03 155421

@dmitryduev @stefanv

stefanv:
Nice work.  An enemy of pause is a friend of mine.

PR #831, opened 2020-09-03
Distinguish between loading recent sources and no recent sources

kmshin1397:


stefanv (approved):

acrellin:
Are there any other components that expect sources to always be a list and might just jump straight into calling map, etc on it?

acrellin:
Ah, nm, I misread. This is just recent sources.

PR #841, opened 2020-09-04
Observing Run Page

dannygoldstein:
This PR adds a page that shows all of the observing runs in the skyportal DB and links to them from the side bar. It also tests this functionality.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -51,6 +51,8 @@ app:
       component: ShareDataForm
     - path: "/filter/:fid"
       component: Filter
+    - path: "/runs"
+      component: ObservingRunList
 
   sidebar:
     # See https://material-ui.com/components/material-icons/
@@ -70,6 +72,10 @@ app:
     - name: Groups
       icon: GroupWork
       url: /groups
+
+    - name: Observing Runs
+      icon: LocalCafe
+      url: /runs
 
     - name: About
       icon: Info
--- a/static/js/components/ObservingRunList.jsx
+++ b/static/js/components/ObservingRunList.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
-import { Link } from "react-router-dom";
 
 import List from '@material-ui/core/List';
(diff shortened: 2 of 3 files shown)


acrellin (approved):

static/js/components/ObservingRunList.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    width: "100%",
+    maxWidth: "22.5rem",
+    backgroundColor: theme.palette.background.paper,
+  },
+}));
+
+const SimpleList = ({ observingRuns }) => {
```
acrellin, line 20: Why define this separately instead of just including the relevant code in `ObservingRunList`? It'd read a lot easier to me if it were a single component

dannygoldstein: Good point. I think I was writing this when I was first learning React. Crazy to think that was only like 3 or 4 weeks ago. Looking back on it now, it all seems very crude...

dannygoldstein: (This is a port of #616)

acrellin: yeah, amazing how quickly you've become a React pro!

config.yaml.defaults
```diff
@@ -71,6 +73,10 @@ app:
       icon: GroupWork
       url: /groups
 
+    - name: Observing Runs
+      icon: LocalCafe
```
profjsb, line 77: Is this the right icon? Perhaps "Visibility" icon is more obvious?

dmitryduev: I appreciate Danny's sense of humor here, would leave it as is ;)

static/js/components/ObservingRunList.jsx
dmitryduev: Agree with @acrellin, but don't think it's a showstopper. I had the same feeling looking at my first React component on F that Stéfan somehow allowed to get merged...

dmitryduev (approved):

PR #843, opened 2020-09-04
Make PDF generation async

stefanv:
- Execute the PDF generation in a thread
- Download the file in chunks

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -1,10 +1,14 @@
 import datetime
 
+import tornado
+from tornado.ioloop import IOLoop
+import io
 from dateutil.parser import isoparse
 from sqlalchemy.orm import joinedload
 from sqlalchemy import func
 import arrow
 from marshmallow.exceptions import ValidationError
+import functools
 import healpix_alchemy as ha
 from baselayer.app.access import permissions, auth_or_token
 from ..base import BaseHandler
@@ -648,7 +652,7 @@ class SourceOffsetsHandler(BaseHandler):
 
 class SourceFinderHandler(BaseHandler):
     @auth_or_token
-    def get(self, obj_id):
+    async def get(self, obj_id):
         """
         ---
         description: Generate a PDF finding chart to aid in spectroscopy
@@ -733,7 +737,8 @@ class SourceFinderHandler(BaseHandler):
         min_sep_arcsec = facility_parameters[facility]["min_sep_arcsec"]
         mag_min = facility_parameters[facility]["mag_min"]
 
-        rez = get_finding_chart(
+        finder = functools.partial(
+            get_finding_chart,
             source.ra,
             source.dec,
             obj_id,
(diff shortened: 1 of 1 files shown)


acrellin (approved):

skyportal/handlers/api/source.py
```diff
@@ -763,4 +770,26 @@ def get(self, obj_id):
             'Cache-Control', 'no-store, no-cache, must-revalidate, max-age=0'
         )
 
-        return self.write(image)
+        # From
+        # https://bhch.github.io/posts/2017/12/serving-large-files-with-tornado-safely-without-blocking/
+        chunk_size = 1024 * 1024 * 1  # 1 MiB
+        while True:
```
profjsb, line 776: Seems like we should have a timeout check. Like if it takes more than X seconds (X ~ 60) then throw a notification and break. As written it's not clear if this loop cannot go forever ...

profjsb (commented):
Looks good. Comments inline. Perhaps send a front end notification saying that "Finding chart generation has begun. The PDF will download automatically once it's complete."

skyportal/handlers/api/source.py
stefanv: The loop terminates in one of two ways:

1. No more bytes left in buffer
2. Stream is closed

We always try to write inside the loop, so if that fails we're out.  Other exceptions will bubble through, so I can keep an eye on the logs to see when that happens.

stefanv: This is also a stopgap measure, until I can get the caching infrastructure fixed up—then we can serve up directly via nginx.

stefanv:
@profjsb I've added the front-end notification; thanks, that was a good idea.

skyportal/handlers/api/source.py
stefanv: OK, I've now made this a for-loop, predicated on the max file size we want to be able to send.

skyportal/handlers/api/source.py
```diff
@@ -752,8 +758,23 @@ def get(self, obj_id):
...
+
         filename = rez["name"]
-        image = rez["data"]
+        image = io.BytesIO(rez["data"])
+
+        # Adapted from
+        # https://bhch.github.io/posts/2017/12/serving-large-files-with-tornado-safely-without-blocking/
+        mb = 1024 * 1024 * 1
+        chunk_size = 1 * mb
+        max_file_size = 5 * mb
```
profjsb, line 773: These can be pretty large if someone chooses a large image size window in the finder. I'd make this max size = 15 Mb

stefanv: Sure, np

skyportal/handlers/api/source.py
```diff
@@ -752,8 +758,23 @@ def get(self, obj_id):
...
+        if not (image.getbuffer().nbytes < max_file_size):
+            return self.error(
+                f"Refusing to send files larger than {max_file_size / mb} MB"
```
profjsb, line 776: Maybe clip this to 1 sig fig after the decimal place

stefanv: Addressed.

skyportal/handlers/api/source.py
stefanv: Done

profjsb (approved):

PR #844, opened 2020-09-04
Fix upload photometry front-end tests: use Instrument filter list specified in test fixtures

acrellin:
Fixes test failures introduced in https://github.com/skyportal/skyportal/commit/ee2d8894b262cd4b61bcf0e9a63ca3730c931e3f

### DIFF
--- a/skyportal/tests/frontend/test_upload_photometry.py
+++ b/skyportal/tests/frontend/test_upload_photometry.py
@@ -13,8 +13,8 @@ def test_upload_photometry(
     csv_text_input = driver.wait_for_xpath('//textarea[@name="csvData"]')
     csv_text_input.send_keys(
         "mjd,flux,fluxerr,zp,magsys,filter\n"
-        "58001,55,1,25,ab,sdssg\n"
-        "58002,53,1,25,ab,sdssg"
+        "58001,55,1,25,ab,ztfg\n"
+        "58002,53,1,25,ab,ztfg"
     )
 
     inst_select = driver.wait_for_xpath('//*[@id="mui-component-select-instrumentID"]')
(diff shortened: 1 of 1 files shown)


acrellin:
Travis is green, only GA failure is `test_news_feed_prefs_widget`

acrellin:
Merging as this fixes consistent failures on master

dannygoldstein:
@acrellin This does not fix the issue. What is happening is that the mouse moving is selecting not SEDM but the adjacent ZTF in the dropdown even though it is supposed to be selecting SEDM. The wrong instrument is being selected. So the test is not uploading photometry to the requested instrument. The filter checking caught this before - the test tried to upload 'sdssg' to ZTF, failing. If the right instrument (SEDM) is selected then this does not happen.

dannygoldstein:
(Sometimes SEDM is correctly selected, in which case this new vesion (uploading ztfg to SEDM) will fail).

acrellin:
@dannygoldstein then that's another issue, but the instrument that *was* being selected had the wrong filters. this fixed that. those tests now pass on CI

dannygoldstein:
test_upload_photometry_multiple_groups

The correct instrument (SEDM) is sandwiched between two adjacent ZTFs. Sometimes those are selected instead of SEDM. The tests here will now pass when that happens. However, sometimes the correct instrument (SEDM) will be selected. SEDM does not have a ztfg. So when that happens this test will fail. So this is switching out one of these cases for another...

dannygoldstein:
The source of this bug is Material UI. There are animations after both scrolling and clicking that need to be waited for, or else the wrong menu element will be selected. The current tests clicked on `//body` immediately after the select element, which caused the wrong select element to end up being chosen (the selected one ended up always being 18 elements behind the targeted one). This is fixed in #889.

Issue #824, opened 2020-09-03
"No taxonomies loaded" in source page classifications component after running make load_demo_data w/ freshly created tables


acrellin:
![image](https://user-images.githubusercontent.com/7230285/92152762-10468f80-edd8-11ea-864a-0f52a38292e2.png)

The browser console shows that FETCH_TAXONOMIES_OK has a data payload of `[]` (empty list).

PR #846, opened 2020-09-04
Return taxonomies associated with user's accessible groups

acrellin:
Closes https://github.com/skyportal/skyportal/issues/824


acrellin:
@profjsb feel free to merge this one after approval (something this small shouldn't require more than one)

PR #847, opened 2020-09-04
Set up alert stream access to invited users

dmitryduev:
New users invited to groups by admins get alert stream access by association.


skyportal/onboarding.py
```diff
@@ -110,19 +120,45 @@ def setup_invited_user_permissions(strategy, uid, details, user, *args, **kwargs
...
         DBSession().add(Group(name=user.username, users=[user], single_user_group=True))
+        # Add stream access to single user group
+        single_user_group = (
+            DBSession().query(Group).filter(Group.name == user.username).first()
+        )
+        for stream_id in stream_ids:
+            DBSession.add(
+                GroupStream(group_id=single_user_group.id, stream_id=stream_id)
+            )
+            DBSession.add(StreamUser(stream_id=stream_id, user_id=user.id))
```
acrellin, line 148: Should this be outside of this if block? We want to do this even if the single user group might exist, right?

acrellin (commented):
Sounds like we're still discussing some of the higher level stuff going on here, but will leave this one for now.

skyportal/onboarding.py
dannygoldstein: Does this even need a block (i.e., do we need the `if` on 137)? How would the single user group exist already for an invited user?

dannygoldstein (commented):
see rider to ari's comment

skyportal/onboarding.py
acrellin: yeah, that can probably be removed. It was an extra precaution during the development phase, but I think the checks prior to this point are robust enough

dmitryduev: Ok, removed the check

dannygoldstein (approved):

skyportal/onboarding.py
```diff
@@ -110,19 +120,42 @@ def setup_invited_user_permissions(strategy, uid, details, user, *args, **kwargs
...
+            .filter(GroupStream.group_id == group_id)
+            .all()
+        )
+        for stream in streams:
+            stream_ids.add(stream.id)
+
+    # Create single-user group
+    DBSession().add(Group(name=user.username, users=[user], single_user_group=True))
+    # Add stream access to single user group
+    single_user_group = (
```
acrellin, line 140: Minor thing that can wait til later -- we can just assign this to the group instance being added above, and won't need to re-fetch it here

acrellin (approved):
Looks good

PR #850, opened 2020-09-04
Update single user group name when username updated

acrellin:


stefanv (approved):

skyportal/handlers/api/internal/profile.py
```diff
@@ -150,6 +152,16 @@ def patch(self):
...
+            if current_username != username:
+                user_group = (
+                    DBSession()
+                    .query(Group)
+                    .filter(Group.name == current_username)
+                    .first()
+                )
+                if user_group is not None:
+                    user_group.name = username
+                username_updated = True
```
stefanv, line 164: Should this be indented or not?

stefanv (approved):

skyportal/handlers/api/internal/profile.py
acrellin: @stefanv no, because we still use it to trigger a re-fetch of the users list

PR #858, opened 2020-09-05
Robust request gets

stefanv:


acrellin (approved):

skyportal/utils/offset.py
```diff
@@ -54,6 +56,15 @@
 }
 
 
+def get_url(*args, **kwargs):
+    # Connect and read timeouts
+    kwargs['timeout'] = (3.05, 20)
```
profjsb, line 61: what does this mean/do?

stefanv: It means we allow 3.05 seconds for a connection to be established, and up to 20s for the data to come back down.

profjsb: ok

PR #867, opened 2020-09-07
Simplify robotic followup-request menu

dannygoldstein:
This simplifies the followup request frontend to only have one select for instrument/allocation instead of 2 which were degenerate with each other (only an allocation_id is needed to fully specify everything). 

This also fixes a bug that @mcoughlin found where the form was not working for more than 1 instrument. 
Kapture 2020-09-07 at 13 52 35


dannygoldstein:
pinging @profjsb @acrellin

acrellin (approved):

PR #870, opened 2020-09-08
Only post comments to groups that associated Obj is actually associated with

acrellin:


acrellin:
GA failure: `test_news_feed_prefs_widget` (unrelated)

dmitryduev (approved):
Thanks for realizing this! LGTM!

skyportal/handlers/api/comment.py
```diff
@@ -87,15 +87,56 @@ def post(self):
...
         user_accessible_group_ids = [g.id for g in self.current_user.accessible_groups]
-        group_ids = data.pop("group_ids", user_group_ids)
-        group_ids = [gid for gid in group_ids if gid in user_accessible_group_ids]
+        user_accessible_filter_ids = [
+            filtr.id
+            for g in self.current_user.accessible_groups
+            for filtr in g.filters
+            if g.filters is not None
+        ]
+        group_ids = data.pop("group_ids", user_accessible_group_ids)
```
stefanv, line 97: Can this be more succinctly written as an intersection operation?

skyportal/handlers/api/comment.py
```diff
@@ -87,15 +87,56 @@ def post(self):
...
                 f"Invalid group IDs field ({group_ids}): "
                 "You must provide one or more valid group IDs."
             )
+
+        # Only post to groups source/candidate is actually associated with
+        if DBSession().query(Candidate).filter(Candidate.obj_id == obj_id).all():
+            candidate_group_ids = [
+                f.group_id
+                for f in (
+                    DBSession()
```
stefanv, line 112: Perhaps splitting these out.  I.e., assign a variable to the inner query, then operate on that, etc.

That way, there are some human-readable handles to what those queries bring in.

skyportal/handlers/api/comment.py
```diff
@@ -87,15 +87,56 @@ def post(self):
...
+            candidate_group_ids = []
+        matching_sources = (
+            DBSession().query(Source).filter(Source.obj_id == obj_id).all()
+        )
+        if matching_sources:
+            source_group_ids = [source.group_id for source in matching_sources]
+        else:
+            source_group_ids = []
+        group_ids = [
+            gid for gid in group_ids if gid in candidate_group_ids + source_group_ids
```
stefanv, line 135: Another candidate for an intersection

skyportal/handlers/api/comment.py
acrellin: the `data.pop` statement? I'm not sure what you mean... `user_accessible-groups` is the default value passed to `pop`

stefanv: Sorry, this comment should have been one line up.

acrellin:
Merging as review comments have been addressed

PR #876, opened 2020-09-09
fix rendering of observing run titles

dannygoldstein:
Currently, on the observing run page, the titles of observing runs that are not associated with any group render as "Loading...". This is because the render only completes successfully if a group ID is defined. 

This PR fixes this bug by defining render behavior when group ID is not defined for a run.

Before:

Screen Shot 2020-09-08 at 10 11 19 PM

After:

Screen Shot 2020-09-08 at 10 11 13 PM


dmitryduev (approved):
LGTM!

kmshin1397 (approved):

PR #878, opened 2020-09-09
Render login error page with informative error message upon auth failure

acrellin:
This PR is one of two possible implementations (second option in forthcoming PR) for rendering a login error page upon unsuccessful authentication attempts. This PR utilizes `os.environ` variables and therefore will not be appropriate if a user's initial auth request might processed on a different node than the subsequent redirect request. The following PR will have a more general login error page that won't have the specific auth error message, but will not need to use environment variables.

login_error_page_specific_message


stefanv:
Closing in favor of generic approach.

PR #889, opened 2020-09-11
API for Receiving Messages from Remote Facilities

dannygoldstein:
This PR adds the final piece of functionality needed to complete the Robotic Follow-up Request ecosystem: the ability for SkyPortal to receive and process unsolicited messages from remote facilities containing follow-up request status updates.

Here is how the system works:

  * This PR defines a `FacilityMessageHandler` which is served on `/api/facility`. This handler implements `POST`. 
  * Requests sent to this handler must contain `followup_request_id` in the payload. The handler loads up the corresponding followup request and checks whether the user who hit the endpoint has permission to view that followup request. If not, the request is rejected.
  * If the user has permission to view the followup request, the `Listener` interface for the robotic instrument is looked up. Similar to what was done for the related case of sending requests to facilities, in this PR instruments are associated with a `listener_classname` which defines a static API for handling incoming requests. 
  * If the API exists, it automatically defines an associated ACL that determines whether a user is allowed to POST to the API. If the User does not have this ACL, the request is rejected.
  * If the User has the ACL, then the schema of the user's request is validated against the schema that the API expects using JSONSchema. (Each Listener must define its expected schema). 
  * If the schema is valid, the request is passed to the API's `receive_message` static method, which updates things in the database as needed.
  * A record of each of these Transactions is stored in the database by serializing the tornado and requests objects associated with each of these calls. 

API documentation is automatically generated for each Listener API by reading its JSONschema:

Screen Shot 2020-09-10 at 8 24 34 PM

An example implementation is provided for SEDM.


dannygoldstein:
green on travis, GA failure unrelated

skyportal/facility_apis/_base.py
```diff
@@ -1,3 +1,38 @@
+from jsonschema_to_openapi.convert import convert
+from copy import deepcopy
+
+
+class _ListenerBase:
+
+    # subclasses should not modify this
+    @classmethod
+    def complete_schema(cls):
+        base = deepcopy(cls.schema)
```
stefanv, line 10: Can you add a docstring to explain what's happening in the code here?

stefanv (approved):
This looks great, thanks @dannygoldstein 

I don't see any tests related to the serialized capturing?

Overall, some higher level documentation (as a docstring for technical detail, and perhaps as a section in the manual for the rest) would be very welcome.

skyportal/facility_apis/_base.py
dannygoldstein: done

acrellin: Is this requested change pushed? I don't see a docstring

skyportal/handlers/api/facility_listener.py
```diff
@@ -0,0 +1,82 @@
...
+        instrument.listener_class.process_message(self)
+
+        transaction_record = FacilityTransaction(
+            request=http.serialize_tornado_request(self),
+            followup_request=request,
+            initiator=self.associated_user_object,
+        )
+
+        DBSession().add(transaction_record)
+        DBSession().add(request)
```
acrellin, line 48: Is this necessary?

skyportal/handlers/api/followup_request.py
```diff
@@ -458,7 +470,19 @@ def put(self, request_id):
...
+        response = followup_request.instrument.api_class.update(followup_request)
+
+        transaction = FacilityTransaction(
+            request=http.serialize_requests_request(response.request),
+            response=http.serialize_requests_response(response),
+            followup_request=followup_request,
+            initiator=self.associated_user_object,
+        )
+
+        DBSession().add(followup_request)
```
acrellin, line 482: Is this necessary?

skyportal/handlers/api/followup_request.py
```diff
@@ -458,7 +470,19 @@ def put(self, request_id):
...
+        DBSession().add(transaction)
+        DBSession().commit()
```
acrellin, line 484: With two `commit()` calls, this can still fail after the first commit, leading to a partially-updated DB state. Perhaps remove the first commit in this handler or replace it with a flush call if necessary

skyportal/handlers/api/followup_request.py
```diff
@@ -497,14 +521,21 @@ def delete(self, request_id):
...
+        transaction = FacilityTransaction(
+            request=http.serialize_requests_request(response.request),
+            response=http.serialize_requests_response(response),
+            followup_request=followup_request,
+            initiator=self.associated_user_object,
+        )
 
-        DBSession().delete(followup_request)
+        DBSession().add(transaction)
+        DBSession().add(followup_request)
```
acrellin, line 534: Again, I don't think this is necessary. 
e.g.
``` python
In [2]: u = User.query.get(4)                                                                                                                
                                                                                                                                              
In [3]: u                                                                                                                                    
Out[3]: <User(id=4, created_at=2020-09-12 16:45:55.210345, modified=2020-09-12 16:45:55.210345, username=groupadmin@cesium-ml.org, first_name=
None, last_name=None, contact_email=None, contact_phone=None, oauth_uid=None, preferences=None)>                                              
                                                                                                                                              
In [4]: u.preferences = {"ab": 123}                                                                                                          
                                                                                                                                              
In [5]: DBSession().commit()                                                                                                                  
                                                                                                                                              
In [6]: u = User.query.get(4)                                                                                                                
                                                                                                                                              
In [7]: u.preferences                                                                                                                        
Out[7]: {'ab': 123}                                                                                                                           
```

skyportal/models.py
```diff
@@ -1635,17 +1645,23 @@ class FollowupRequest(Base):
     payload = sa.Column(
         psql.JSONB, nullable=True, doc="Content of the followup request."
     )
-    status = sa.Column(sa.String(), nullable=False, default="pending submission")
+    status = sa.Column(
```
acrellin, line 1648: Is there a pre-determined set of valid statuses? If so, perhaps this should be an enum

acrellin (approved):
Looks great @dannygoldstein , just a few minor comments below

skyportal/facility_apis/_base.py
dannygoldstein: not yet

skyportal/models.py
dannygoldstein: there is not.

skyportal/handlers/api/facility_listener.py
dannygoldstein: removed

skyportal/handlers/api/followup_request.py
dannygoldstein: removed

skyportal/handlers/api/followup_request.py
dannygoldstein: This is intended. We want to differentiate between whether the request failed before it was submitted to the observatory or after.

dannygoldstein: fixed

skyportal/handlers/api/followup_request.py
dannygoldstein: fixed

skyportal/facility_apis/_base.py
dannygoldstein: pushed

dannygoldstein: The serialized capturing is not currently used anywhere and there is no API, so nothing to test at the moment. For now, it is purely record keeping in anticipation of possible future usages and for debugging. When we begin using the contents of the table, I will add tests.

dannygoldstein: I added docstrings for all the functions. I request deferring the deployment of official docs for my forthcoming docs pr.

PR #894, opened 2020-09-12
Use custom user_details PSA pipeline func so that first_name, last_name, contact_email are not overwritten upon login

acrellin:


acrellin:
Failures are all known to be failing elsewhere and are unrelated: 
```
FAILED skyportal/tests/frontend/test_followup_requests.py::test_delete_followup_request
FAILED skyportal/tests/frontend/test_followup_requests.py::test_submit_new_followup_request
FAILED skyportal/tests/frontend/test_followup_requests.py::test_edit_existing_followup_request
FAILED skyportal/tests/frontend/test_recent_sources.py::test_recent_sources
FAILED skyportal/tests/frontend/test_quick_search.py::test_quick_search - sel...
```

skyportal/onboarding.py
```diff
@@ -165,3 +165,57 @@ def setup_invited_user_permissions(strategy, uid, details, user, *args, **kwargs
 
     invitation.used = True
     DBSession().commit()
+
+
+def user_details(strategy, details, backend, uid, user=None, *args, **kwargs):
+    """Update user details using data from provider."""
+    if not user:
+        return
+    existing_user = DBSession().query(User).filter(User.oauth_uid == uid).first()
```
stefanv, line 174: What happens for users added programmatically, who do not have an oauth_uid.

In that case, existing user will probably evaluate to None?

stefanv (commented):
Good find, Ari!

skyportal/onboarding.py
acrellin: Yes, thanks, I will check if it's None first!

acrellin: @stefanv actually, I just looked at it and think it's right as is. If we're adding users programmatically that will auth using real OAuth, we'll need to set their oauth_uid when we create the users for it to work anyways. So this shouldn't be an issue. I just checked it with debug_login and it works as expected.

Issue #883, opened 2020-09-10
Cannot add users to groups if username contains space character 


dmitryduev:

profjsb:
I think we should not allow space characters in the username.

PR #895, opened 2020-09-12
Slugify usernames

acrellin:
Closes  https://github.com/skyportal/skyportal/issues/883 

Depends on https://github.com/cesium-ml/baselayer/pull/145

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -1,17 +1,17 @@
 user:
-  - username: testadmin@cesium-ml.org
+  - username: testadmin
     roles:
       - Super admin
     =id: testadmin
-  - username: groupadmin@cesium-ml.org
+  - username: groupadmin
     roles:
       - Group admin
     =id: groupadmin
-  - username: fulluser@cesium-ml.org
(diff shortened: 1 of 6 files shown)


stefanv (approved):

profjsb (changes requested):
If the user enters a username with a space or non-alpha character (other than -) it will slugify after they make the change, which can be disconcerting. Instead you should have a tooltip saying something like "Username is any combination of letters and dashes, 3-16 characters in length". Then you can do form validation onsubmit (using a regex to check like `/^[a-zA-Z\-]{3,16}$/`).

acrellin:
@profjsb how about just a note saying that all usernames will be slugified, with a blurb about what that means? When the user clicks save they'll immediately see the slugified username when the new profile data is re-fetched. It won't be a surprise to them.

acrellin:
This one still needs a bit of updating in other parts of the app to get working smoothly

acrellin:
@stefanv @profjsb screen capture: 
slugify_username

profjsb (approved):

PR #898, opened 2020-09-12
Catch bad emails and phone numbers upon user creation

profjsb:
The PR adds code to the `user` POST handler to catch when a phone number or email is not valid. Before we were not catching the exception upon database insert.


acrellin (approved):

acrellin:
This is great, thanks @profjsb!

PR #900, opened 2020-09-13
Use @id instead of @class to select elements in followup request tests

dannygoldstein:
Fixes followup request failures in #888

PR #901, opened 2020-09-13
fix `test_bulk_delete_photometry`

dannygoldstein:
`api/test_bulk_delete_photometry` is currently failing on master as it has a hardcoded instrument id (id=1) which can sometimes refer to an instrument without `ztfg` filters. This fix has it use the `ztf_camera` fixture, which will always have the right filter set.


dannygoldstein:
failures are a) unrelated or b) fixed in #900

acrellin (approved):

PR #904, opened 2020-09-13
speed up saving of sources on scanning page

### LINKED ISSUE #808: speed up source saving
Currently on the scanning page when you save a source, FETCH_CANDIDATES is dispatched, which re-fetches the entire candidate list. On private.caltech.edu, this results in it taking several seconds to save a single source, which is  a long time. We should only need to re-fetch the saved source, i.e., dispatch FETCH_CANDIDATE instead of FETCH_CANDIDATES.

dannygoldstein:
Fixes #808 .

This PR speeds up the scanning process by reducing the time it takes to save a single candidate. This is achieved by replacing the `FETCH_CANDIDATES` action that is triggered on successful post of a new source with a new `FETCH_CANDIDATE_AND_MERGE` action which fetches the updated information for the single candidate and merges it into the `candidates` branch of the state tree. This results in a factor of `num_candidates_on_scanning_page` less data being transferred and parsed into the state tree each time a candidate is saved.


acrellin (approved):

static/js/ducks/candidates.js
```diff
@@ -1,12 +1,19 @@
...
 
 export const FETCH_CANDIDATES = "skyportal/FETCH_CANDIDATES";
 export const FETCH_CANDIDATES_OK = "skyportal/FETCH_CANDIDATES_OK";
 export const FETCH_CANDIDATES_FAIL = "skyportal/FETCH_CANDIDATES_FAIL";
 
+export const FETCH_CANDIDATE_AND_MERGE = "skyportal/FETCH_CANDIDATE_AND_MERGE";
+export const FETCH_CANDIDATE_AND_MERGE_OK =
+  "skyportal/FETCH_CANDIDATE_AND_MERGE_OK";
+
+export const REFRESH_CANDIDATE = "skyportal/REFRESH_CANDIDATE";
```
profjsb, line 15: does this need a corresponding `REFRESH_CANDIDATE_OK`?

dannygoldstein: nope, as this action originates from the backend, not the frontend, and is handled by the websocket messagehandler.

skyportal/handlers/api/source.py
```diff
@@ -394,7 +394,7 @@ def post(self):
         DBSession().commit()
 
         self.push_all(action="skyportal/FETCH_SOURCES")
-        self.push_all(action="skyportal/FETCH_CANDIDATES")
+        self.push_all(action="skyportal/REFRESH_CANDIDATE", payload={"id": obj.id})
```
acrellin, line 397: This should also use the internal_key so as to not leak sensitive obj IDs

acrellin (changes requested):
Just realized this should not push_all with the obj ID -- should be updated to use the internal_key as we do elsewhere so as to not leak source/cand IDs

dannygoldstein:
[ch1242]

skyportal/handlers/api/source.py
dannygoldstein: fixed

dannygoldstein:
ready for re-review

acrellin (approved):

profjsb (approved):

PR #905, opened 2020-09-13
fix overfetching of users in AssigmentList

### LINKED ISSUE #699: AssignmentList calls `fetchUser` too many times
As `fetchUser` is not currently encapsulated using `useEffect` in AssignmentList, it often issues far too many requests to the backend on loading the source page. 

image

dannygoldstein:
Fixes #699 .

The currently merged solution to this issue (#759 ) is flawed in that it provides a dependency list with a length that can change (spread syntax). This can cause React errors.

The solution here avoids the spread syntax and passes a fixed dependency list to `useEffect` each time the component is rerendered. A useState hook is used to prevent redundant `fetchUser` calls between renders.

### DIFF
--- a/static/js/components/AssignmentList.jsx
+++ b/static/js/components/AssignmentList.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import PropTypes from "prop-types";
 import { useSelector, useDispatch } from "react-redux";
 import * as Actions from "../ducks/source";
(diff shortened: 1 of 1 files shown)


static/js/components/AssignmentList.jsx
```diff
@@ -15,20 +15,21 @@ const AssignmentList = ({ assignments }) => {
...
-  uniqueRequesterIDs.sort((a, b) => a - b);
 
   // use useEffect to only send 1 fetchUser per User
   useEffect(() => {
-    uniqueRequesterIDs.forEach((id) => {
-      if (!users[id]) {
+    requesterIDs.forEach((id) => {
+      if (!users[id] && !loadingUsers.includes(id)) {
         dispatch(UserActions.fetchUser(id));
+        loadingUsers.push(id);
```
acrellin, line 28: I think modifying the state object directly is an anti-pattern; we should make a copy of `loadingUsers`, operate on that copy, and then pass it into `setLoadingUsers`

static/js/components/AssignmentList.jsx
```diff
@@ -15,20 +15,21 @@ const AssignmentList = ({ assignments }) => {
   const { users } = useSelector((state) => state);
   const { observingRunList } = useSelector((state) => state.observingRuns);
   const { instrumentList } = useSelector((state) => state.instruments);
+  const [loadingUsers, setLoadingUsers] = useState([]);
```
acrellin, line 18:
```suggestion
  const [loadingOrLoadedUsers, setLoadingOrLoadedUsers] = useState([]);
```
So that we know it's not just for in-progress queries

static/js/components/AssignmentList.jsx
```diff
@@ -15,20 +15,21 @@ const AssignmentList = ({ assignments }) => {
...
-      if (!users[id]) {
+    requesterIDs.forEach((id) => {
+      if (!users[id] && !loadingUsers.includes(id)) {
         dispatch(UserActions.fetchUser(id));
+        loadingUsers.push(id);
+        setLoadingUsers(loadingUsers);
       }
     });
-  }, [...uniqueRequesterIDs, users, dispatch]);
+  }, [requesterIDs, users, dispatch]);
```
acrellin, line 32: Should `loadingUsers` be in the deps array here?

dannygoldstein:
@acrellin opted for a different approach that does not use `useState`

acrellin (approved):
Nice

PR #926, opened 2020-09-16
Refactor NewGroupForm to use MUI and update API to use user IDs instead of user emails

acrellin:
This PR updates the new group form component to use Material UI and uses a multi-select instead of a comma-separated text input for specifying new group admins. The API is also updated to use group admin user IDs instead of (outdated, error-prone) user email.

new_group_form

### DIFF
--- a/skyportal/tests/frontend/test_groups.py
+++ b/skyportal/tests/frontend/test_groups.py
@@ -33,7 +33,9 @@ def test_add_new_group(driver, super_admin_user, user):
     driver.get('/groups')
     driver.wait_for_xpath('//input[@name="name"]').send_keys(test_proj_name)
     driver.click_xpath('//div[@id="groupAdminsSelect"]')
-    driver.wait_for_xpath_to_be_clickable(f'//li[text()="{user.username}"]').click()
+    driver.wait_for_xpath_to_be_clickable(
+        f'//li[contains(text(),"{user.username}")]'
+    ).click()
     driver.click_xpath('//button[contains(.,"Create Group")]', wait_clickable=False)
     driver.wait_for_xpath(f'//a[contains(.,"{test_proj_name}")]')
 
@@ -48,7 +50,7 @@ def test_add_new_group_explicit_self_admin(driver, super_admin_user, user):
     driver.wait_for_xpath('//input[@name="name"]').send_keys(test_proj_name)
     driver.click_xpath('//div[@id="groupAdminsSelect"]')
     driver.wait_for_xpath_to_be_clickable(
-        f'//li[text()="{super_admin_user.username}"]'
+        f'//li[contains(text(),"{user.username}")]'
     ).click()
     driver.click_xpath('//button[contains(.,"Create Group")]', wait_clickable=False)
     driver.wait_for_xpath(f'//a[contains(.,"{test_proj_name}")]')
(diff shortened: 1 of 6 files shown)


static/js/components/NewGroupForm.jsx
```diff
@@ -1,78 +1,146 @@
...
+    newState[event.target.name] = event.target.value;
     setState({
       ...formState,
       ...newState,
     });
   };
 
+  const useStyles = makeStyles((theme) => ({
+    formControl: {
+      margin: theme.spacing(1),
```
kmshin1397, line 68:
```suggestion
      margin: `${theme.spacing(1)}px 0`,
```
I think we should only have the top/bottom margins here so that the admin select stays aligned with the group name input. Change results in:
Screenshot 2020-09-16 153248

vs right now:
Screenshot 2020-09-16 154117

skyportal/tests/frontend/test_groups.py
```diff
@@ -32,9 +32,9 @@ def test_add_new_group(driver, super_admin_user, user):
     driver.refresh()
     driver.get('/groups')
     driver.wait_for_xpath('//input[@name="name"]').send_keys(test_proj_name)
-    driver.wait_for_xpath('//input[@name="groupAdmins"]').send_keys(user.username)
-    driver.save_screenshot('/tmp/screenshot1.png')
-    driver.click_xpath('//button[contains(.,"Create Group")]')
+    driver.click_xpath('//div[@id="groupAdminsSelect"]')
+    driver.wait_for_xpath_to_be_clickable(f'//li[text()="{user.username}"]').click()
```
kmshin1397, line 36:
```suggestion
    driver.wait_for_xpath_to_be_clickable(
        f'//li[contains(text(),"{user.username}")]'
    ).click()
```
The ripple span in the li element messes up the text()= xpath matching I think. Using the contains seems to work for me locally.

skyportal/tests/frontend/test_groups.py
```diff
@@ -46,10 +46,11 @@ def test_add_new_group_explicit_self_admin(driver, super_admin_user, user):
     driver.refresh()
     driver.get('/groups')
     driver.wait_for_xpath('//input[@name="name"]').send_keys(test_proj_name)
-    driver.wait_for_xpath('//input[@name="groupAdmins"]').send_keys(
-        super_admin_user.username
-    )
-    driver.click_xpath('//button[contains(.,"Create Group")]')
+    driver.click_xpath('//div[@id="groupAdminsSelect"]')
+    driver.wait_for_xpath_to_be_clickable(
+        f'//li[text()="{super_admin_user.username}"]'
```
kmshin1397, line 51:
```suggestion
        f'//li[contains(text(), "{super_admin_user.username}")]'
```

kmshin1397 (commented):
Fixed your tests

skyportal/tests/frontend/test_groups.py
acrellin: Interesting, thanks for looking into that! I couldn't reproduce the errors locally. Much appreciated.

acrellin:
@kmshin1397 thanks for the good suggestions and test fixes! I've pushed up those changes.

skyportal/tests/frontend/test_groups.py
kmshin1397: @acrellin Just kidding I might be wrong about that. I think I looked at the wrong test and thought this fixed made it pass.

kmshin1397:
@acrellin I see this when I run the test locally. It seems like the `driver.click_xpath('//div[@id="groupAdminsSelect"]')` line is somehow scrolling the view past the form (and the user select list hasn't popped up either if I manually scroll back up). 
test

kmshin1397 (approved):
LGTM!

dmitryduev (approved):
LGTM! All tests are passing for me locally, except for test_add_stream_add_delete_filter_group, which xfails -- I'll look into fixing it.

PR #936, opened 2020-09-17
Refactor Transaction Logging in Remote Facility Framework

dannygoldstein:
This PR moves the logging of remote facility transactions to the `staticmethod`s within the Listener and API classes for individual facilities. The previous implementation legislated a universal method for serializing these transactions based on `requests`. This didn't work, for example, for LT, which uses an XML/urllib based communication protocol. This PR solves this problem, accommodating the different HTTP request/response formats that different facilities have. 

This should allow #866 to move forward.

### DIFF
--- a/skyportal/facility_apis/sedm.py
+++ b/skyportal/facility_apis/sedm.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta
 import json
 import requests
 
+from ..utils import http
 
 env, cfg = load_env()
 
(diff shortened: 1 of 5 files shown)


mcoughlin:
It looks OK. Does request=http.serialize_requests_request(r.request) become the XML output or similar for non-request telescopes?

dannygoldstein:
> It looks OK. 

@mcoughlin Add review by clicking on "Add your review" green button at top of page (can be simply approval). 

> Does request=http.serialize_requests_request(r.request) become the XML output or similar for non-request telescopes?

No, that method only processes `requests` requests. For an XML based one, add a new function to the  `utils.http` module.

mcoughlin (approved):

dmitryduev (commented):
LGTM! However, before I merge this: the PR changes the data model, so looks like a perfect opportunity to practice db migrations. Also, I'd prefer not to zoink the db on private. Could you please elaborate on what exactly needs to be done on the ops end?

dannygoldstein:
It should be a two step thing:

1. Run a migration as described in #740 
2. Redeploy the web server

dannygoldstein:
@dmitryduev Can we merge this?

PR #937, opened 2020-09-17
add an explanatory blurb on the share data page

dannygoldstein:
[ch1273] @scizen9

This PR adds an explanatory blurb to the share data page to explain what the page is and how it works.

### DIFF
--- a/static/js/components/ShareDataForm.jsx
+++ b/static/js/components/ShareDataForm.jsx
@@ -178,7 +178,10 @@ const ShareDataForm = ({ route }) => {
           users or groups. Select the photometry or spectra you would like to
           share from the list below, then select the users or groups you would
           like to share the data with. When you click submit, the access
-          permissions on the data will be updated.
+          permissions on the data will be updated. This page does <em>not</em>
+          reproduce the &ldquo;transfer source&ldquo; functionality of the
+          GROWTH marshal. Data shared via this page will not cause the source to
+          be saved to another group.
         </p>
       </div>
       <br />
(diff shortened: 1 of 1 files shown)


dmitryduev (changes requested):
Could you please add that this is not "transferring a source" (as on the GROWTH marshal), so the source won't show up in another group as people might expect.

dannygoldstein:
@dmitryduev Can you modify the text as needed (can even be via the github text editor) and merge when you are happy?

dmitryduev (approved):
👍

static/js/components/ShareDataForm.jsx
```diff
@@ -173,6 +173,16 @@ const ShareDataForm = ({ route }) => {
...
           </Link>
         </Typography>
+        <p>
+          This page allows you to share data for {`${route.id}`} with other
+          users or groups. Select the photometry or spectra you would like to
+          share from the list below, then select the users or groups you would
+          like to share the data with. When you click submit, the access
+          permissions on the data will be updated. This page does <em>not</em>
+          reproduce the &ldquo;transfer source&ldquo; functionality of the
+          GROWTH marshal. Data shared via this page will not cause the source to
```
stefanv, line 183: We would like to keep SkyPortal survey-agnostic, so mentioning the GROWTH marshal should not happen here.

dmitryduev: Ah, good point @stefanv ! @dannygoldstein could you please remove that sentence?

PR #940, opened 2020-09-17
Better Finder Positions

profjsb:
The PR introduces a new default behavior to use photometry-based centroiding of a source position for the purposes of making finding charts and offset/starlists. Rather than use the position at discovery time or the highest SNR detection of the source, combine all the positions available from associated photometry. If the calculated position is far (> 0.5 arcsec by default) from discovery, revert back to the original discovery image.

- always save ra,dec as discovery position if not specified
- introduce two weighting schemes (`snr` and `err`) to determine the offset from fiducial
- default finder imaging to ztfref and change the display scale to look better around galaxies

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -385,10 +385,6 @@ class CandidateHandler(BaseHandler):
         if not filters:
             return self.error("At least one valid filter ID must be provided.")
 
-        # use ra, dec as the discovery position unless it's already specified
-        obj.ra_dis = obj.ra if obj.ra_dis is None else obj.ra_dis
-        obj.dec_dis = obj.dec if obj.dec_dis is None else obj.dec_dis
-
         DBSession().add(obj)
         DBSession().add_all(
             [
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -312,7 +312,7 @@ class CandidateHandler(BaseHandler):
     def post(self):
         """
         ---
-        description: POST a new candidate.
+        description: Create a new candidate.
         requestBody:
           content:
             application/json:
@@ -384,6 +384,11 @@ class CandidateHandler(BaseHandler):
         filters = Filter.query.filter(Filter.id.in_(filter_ids)).all()
         if not filters:
             return self.error("At least one valid filter ID must be provided.")
+
+        # use ra, dec as the discovery position unless it's already specified
+        obj.ra_dis = obj.ra if obj.ra_dis is None else obj.ra_dis
+        obj.dec_dis = obj.dec if obj.dec_dis is None else obj.dec_dis
+
         DBSession().add(obj)
         DBSession().add_all(
             [
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -439,6 +439,11 @@ class SourceHandler(BaseHandler):
                 "Invalid group_ids field. Please specify at least "
                 "one valid group ID that you belong to."
             )
+
+        # use ra, dec as the discovery position unless it's already specified
+        obj.ra_dis = obj.ra if obj.ra_dis is None else obj.ra_dis
+        obj.dec_dis = obj.dec if obj.dec_dis is None else obj.dec_dis
+
         DBSession().add(obj)
         DBSession().add_all([Source(obj=obj, group=group) for group in groups])
         DBSession().commit()
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -822,7 +822,7 @@ class SourceFinderHandler(BaseHandler):
             best_ra, best_dec = initial_pos[0], initial_pos[1]
 
         facility = self.get_query_argument('facility', 'Keck')
-        image_source = self.get_query_argument('image_source', 'desi')
+        image_source = self.get_query_argument('image_source', 'ztfref')
 
         how_many = 3
         obstime = self.get_query_argument(
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -34,7 +34,7 @@ from ...utils import (
     facility_parameters,
     source_image_parameters,
     get_finding_chart,
-    calculate_best_position,
+    _calculate_best_position_for_offset_stars,
 )
 from .candidate import grab_query_results_page
 
@@ -442,10 +442,6 @@ class SourceHandler(BaseHandler):
                 "one valid group ID that you belong to."
             )
 
-        # use ra, dec as the discovery position unless it's already specified
-        obj.ra_dis = obj.ra if obj.ra_dis is None else obj.ra_dis
-        obj.dec_dis = obj.dec if obj.dec_dis is None else obj.dec_dis
-
         DBSession().add(obj)
         DBSession().add_all([Source(obj=obj, group=group) for group in groups])
         DBSession().commit()
@@ -658,14 +654,13 @@ class SourceOffsetsHandler(BaseHandler):
         if source is None:
             return self.error('Invalid source ID.')
 
-        initial_pos = (
-            source.ra_dis if source.ra_dis is not None else source.ra,
-            source.dec_dis if source.dec_dis is not None else source.dec,
-        )
+        initial_pos = (source.ra, source.dec)
 
         try:
-            best_ra, best_dec = calculate_best_position(
-                source.photometry, fallback=(initial_pos[0], initial_pos[1]), how="snr",
+            best_ra, best_dec = _calculate_best_position_for_offset_stars(
+                source.photometry,
+                fallback=(initial_pos[0], initial_pos[1]),
+                how="snr2",
             )
         except JSONDecodeError:
             self.push_notification(
@@ -806,13 +801,12 @@ class SourceFinderHandler(BaseHandler):
         if imsize < 2.0 or imsize > 15.0:
             return self.error('The value for `imsize` is outside the allowed range')
 
-        initial_pos = (
-            source.ra_dis if source.ra_dis is not None else source.ra,
-            source.dec_dis if source.dec_dis is not None else source.dec,
-        )
+        initial_pos = (source.ra, source.dec)
         try:
-            best_ra, best_dec = calculate_best_position(
-                source.photometry, fallback=(initial_pos[0], initial_pos[1]), how="snr",
+            best_ra, best_dec = _calculate_best_position_for_offset_stars(
+                source.photometry,
+                fallback=(initial_pos[0], initial_pos[1]),
+                how="snr2",
             )
         except JSONDecodeError:
             self.push_notification(
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -1,4 +1,5 @@
 import datetime
+from json.decoder import JSONDecodeError
 
 import tornado
 from tornado.ioloop import IOLoop
@@ -33,6 +34,7 @@ from ...utils import (
     facility_parameters,
     source_image_parameters,
     get_finding_chart,
+    calculate_best_position,
 )
 from .candidate import grab_query_results_page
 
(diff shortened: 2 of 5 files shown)


skyportal/handlers/api/candidate.py
```diff
@@ -384,6 +384,11 @@ def post(self):
         filters = Filter.query.filter(Filter.id.in_(filter_ids)).all()
         if not filters:
             return self.error("At least one valid filter ID must be provided.")
+
+        # use ra, dec as the discovery position unless it's already specified
+        obj.ra_dis = obj.ra if obj.ra_dis is None else obj.ra_dis
```
dannygoldstein, line 389: This uses the first detection that passed a filter as the discovery coordinate. Are we sure we want to do that, rather than use the  first detection (discovery detection), as `*_dis`?

skyportal/handlers/api/source.py
```diff
@@ -439,6 +441,11 @@ def post(self):
                 "Invalid group_ids field. Please specify at least "
                 "one valid group ID that you belong to."
             )
+
+        # use ra, dec as the discovery position unless it's already specified
+        obj.ra_dis = obj.ra if obj.ra_dis is None else obj.ra_dis
```
dannygoldstein, line 446: How do you know `obj.ra` always correspond to the discovery ra? And vice versa for dec?

skyportal/handlers/api/source.py
```diff
@@ -643,10 +650,30 @@ def get(self, obj_id):
               application/json:
                 schema: Error
         """
-        source = Source.get_obj_if_owned_by(obj_id, self.current_user)
+        source = Source.get_obj_if_owned_by(
```
dannygoldstein, line 653: This is already an `Obj`, right? So no need to join on `Source.obj`

skyportal/handlers/api/source.py
```diff
@@ -761,7 +788,11 @@ async def get(self, obj_id):
...
+            obj_id,
+            self.current_user,
+            options=[joinedload(Source.obj).joinedload(Obj.photometry)],
```
dannygoldstein, line 794: Already an obj

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+    photometry, fallback=(None, None), how="snr", max_offset=0.5, sigma_clip=4.0
+):
+    """Calculates the best position for a source from its photometric
+       points
+
+    Parameters
+    ----------
+    photometry : list
+        List of Photometry objects
+    fallback : tuple, optional
```
dannygoldstein, line 170: Why not simply raise an exception and let that propagate up to the calling function, rather than handle the exceptions inside this function?

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+    if not isinstance(photometry, list):
+        log(
+            "Warning: No photometry given. Falling back to "
+            " original source position."
+        )
+        return fallback
+
+    # convert the photometry into a dataframe
+    phot_s = "[" + ",".join([str(x) for x in photometry]) + "]"
+    df = pd.DataFrame(json.loads(phot_s))
```
dannygoldstein, line 191:
```suggestion
    phot = [x.to_dict() for x in photometry]
    df = pd.DataFrame(phot)
```

dannygoldstein: No need to convert to string here

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+
+    # remove limit data (non-detections)
+    try:
+        df = df[(df["flux"].notnull()) & (df["fluxerr"].notnull())]
+    except KeyError:
+        log(
```
dannygoldstein, line 197: Again, why not just raise an exception here?

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+            " original source position."
+        )
+        return fallback
+
+    # remove observations with distances more than max_offset away
+    # from the median
+    try:
+        med_ra, med_dec = np.median(df["ra"]), np.median(df["dec"])
+    except TypeError:
+        log(
```
dannygoldstein, line 208: Again, why not just raise an exception here?

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+        if np.abs(sep) > max_offset * u.arcsec:
+            log(
+                "Warning: calculated source position is too far from the"
+                " fiducial. Falling back to the fiducial "
+            )
+            return fallback
+
+    df["ra_offset"] = np.cos(np.radians(df["dec"])) * (df["ra"] - med_ra) * 3600.0
+    df["dec_offset"] = (df["dec"] - med_dec) * 3600.0
+    df["offset_arcsec"] = np.sqrt(df["ra_offset"] ** 2 + df["dec_offset"] ** 2)
```
dannygoldstein, line 229: Why not use `astropy.SkyCoord.sep`? 
https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.separation

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+    df = df[df["offset_arcsec"] <= max_offset]
+
+    # remove outliers
+    if len(df) > 4 and sigma_clip is not None:
+        df = df[df["offset_arcsec"] < sigma_clip * np.std(df["offset_arcsec"])]
```
dannygoldstein, line 234: Suggestion: use MAD std instead of regular std, less sensitive to outliers.

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+
+    # TODO: add the ability to use only use observations from some filters
+    if how == "snr":
```
dannygoldstein, line 237:
```suggestion
    if how == "snr2":
```

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+        df["snr"] = df["flux"] / df["fluxerr"]
+        diff_ra = np.average(df["ra_offset"], weights=df["snr"] ** 2)
+        diff_dec = np.average(df["dec_offset"], weights=df["snr"] ** 2)
+    elif how == "err":
```
dannygoldstein, line 241:
```suggestion
    elif how == "invvar":
```

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+    Parameters
+    ----------
+    photometry : list
+        List of Photometry objects
+    fallback : tuple, optional
+        The position to use if something goes wrong here
+    how : str
+        how to weight positional data:
+          snr = use the signal to noise
+          err = use the photometric uncertainty
```
dannygoldstein, line 175:
```suggestion
          snr2 = use the signal to noise squared
          invvar = use the inverse photometric variance
```

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+        log(
+            "Warning: could not find the median of the positions"
+            " from the photometry data associated with this source "
+        )
+        return fallback
+
+    # check to make sure that the median isn't too far away from the
+    # discovery position
+    if fallback != (None, None):
+        c1 = SkyCoord(med_ra * u.deg, med_dec * u.deg, frame='icrs')
```
dannygoldstein, line 217: Why not use `SkyCoord` elsewhere for consistency?

skyportal/utils/offset.py
```diff
@@ -156,6 +157,101 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
...
+          snr = use the signal to noise
+          err = use the photometric uncertainty
+    max_offset : float, optional
+        How many arcseconds away should we ignore discrepant points?
+    sigma_clip : float, optional
+        Remove positions that are this number of std away from the median
+    """
+
+    if not isinstance(photometry, list):
+        log(
```
dannygoldstein, line 183: Why not raise this?

skyportal/handlers/api/source.py
```diff
@@ -643,10 +650,30 @@ def get(self, obj_id):
...
+        )
         if source is None:
             return self.error('Invalid source ID.')
 
+        initial_pos = (
+            source.ra_dis if source.ra_dis is not None else source.ra,
+            source.dec_dis if source.dec_dis is not None else source.dec,
```
dannygoldstein, line 663: The discovery RA and Dec will often correspond to some of the most astrometrically uncertain photometry. Why not default to the highest S/N detection as the fall back coordinate?

skyportal/handlers/api/candidate.py
profjsb: We weren't doing anything before....
I'm assuming `discovery` is the moment something passes a filter, rather than the first time it's observed. So yes I think this correct.

skyportal/handlers/api/source.py
profjsb: Again, it's just to get the *_dis variables loaded at save time.

skyportal/utils/offset.py
profjsb: These are very small angles (< 1 arcsec), so no need to make expensive SkyCoords

skyportal/handlers/api/source.py
profjsb: I'd prefer to keep this at the position when the source is first saved (ie. passes a filter).

skyportal/utils/offset.py
profjsb: I'd prefer to have `calculate_best_position` always return a position rather than the possibility of raising an exception especially given all the ways this calc can go wrong.

skyportal/utils/offset.py
profjsb: Unlike the other calculations could be a huge angle, if there is a catastrophic error in the photometry relative to the fiducial -- so I need the great circle distance here.

skyportal/utils/offset.py
```diff
@@ -2,6 +2,7 @@
 import os
 import datetime
 import warnings
+import json
```
kmshin1397, line 5:
```suggestion
```
json import isn't used - causing the failure on GA

skyportal/tests/tools/test_offset_util.py
```diff
@@ -1,10 +1,110 @@
...
+def test_calculate_best_position_no_photometry():
+
+    ra, dec = calculate_best_position(
+        [], fallback=(10.0, -20.0), how="snr", max_offset=0.5, sigma_clip=4.0
+    )
+    npt.assert_almost_equal(ra, 10)
+    npt.assert_almost_equal(dec, -20)
+
+
+def test_calculate_best_position_with_photometry(
```
kmshin1397, line 29: Maybe mark this as flaky? Failed on Travis but seems fine locally.

kmshin1397 (commented):
Couple minor inline comments. Can't contribute much to the astronomy-focused points Danny raised, but code itself looks good to me.

skyportal/tests/tools/test_offset_util.py
profjsb: Done

skyportal/utils/offset.py
profjsb: Done

skyportal/utils/offset.py
profjsb: This is not an error - it's certainly possible to have an Obj without any associated photometry (e.g. DESI). It's graceful to just return the fallback position (ie. the Obj ra/dec)

skyportal/handlers/api/candidate.py
profjsb: Ok, removing this in favor of moving the discovery position to Source/Candidate (another PR)

skyportal/handlers/api/source.py
profjsb: Correct but this pattern is used throughout source.py, so keeping it here.

skyportal/handlers/api/source.py
profjsb: See above.

skyportal/utils/offset.py
profjsb: Renamed to make clear what this function is supposed to be used for.

dannygoldstein (approved):

kmshin1397 (approved):

PR #958, opened 2020-09-20
Cosmology calculations and display

profjsb:
Kapture 2020-09-20 at 22 29 31

  - Site-wide configured  cosmology in `config.yaml` (named cosmologies from `astropy` or custom):
   ```
  misc:
     cosmology: Planck18_arXiv_v2
    #cosmology: {H0: "65.0", Om0: 0.3, Ode0: 0.7, name: 'skyportal_user_cosmo'}
  ```
 - Adding the distance calculations to Obj using either information in `altdata` (e.g. `altdata["parallax"]`) or using redshift + cosmology.
 - Show on front-end if DM/luminosity distance is determined
 - If redshift puts the source nearby (<350 km/s) then do no show DM/D_l unless a distance is established in altdata.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -228,6 +228,10 @@ services:
 misc:
   days_to_keep_unsaved_candidates: 7
   public_group_name: "Sitewide Group"
+  # Use a named cosmology from `astropy.cosmology.parameters.available`
+  # or supply the arguments for an FLRW cosmological instance
+  cosmology: Planck18_arXiv_v2
+  #cosmology: {H0: "65.0", Om0: 0.3, Ode0: 0.7, name: 'skyportal_user_cosmo'}
 
 cron:
   - interval: 1440
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -228,8 +228,11 @@ services:
 misc:
   days_to_keep_unsaved_candidates: 7
   public_group_name: "Sitewide Group"
-  # Use a named cosmology from `astropy.cosmology.parameters.available`
-  # or supply the arguments for an FLRW cosmological instance
+  # Use a named cosmology from `astropy.cosmology.parameters.available` cosmologies
+  # or supply the arguments for an `astropy.cosmology.FLRW` cosmological instance.
+  # If {"flat": True} then use a subclass of the FLRW, called `FlatLambdaCMD`
+  # otherwise use `LambdaCDM`. See `utils.cosmology.py` for the way in which
+  # the user-supplied cosmology parameter set constructs the site-wide cosmology.
   cosmology: Planck18_arXiv_v2
   #cosmology: {H0: "65.0", Om0: 0.3, Ode0: 0.7, name: 'skyportal_user_cosmo'}
 
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -467,7 +467,11 @@ class Obj(Base, ha.Point):
     def luminosity_distance(self):
         """
         The luminosity distance in Mpc, using either DM or distance data
-        in the altdata fields or using the cosmology/redshift
+        in the altdata fields or using the cosmology/redshift. Specifically
+        the user can add `dm` (mag), `parallax` (arcsec), `dist_kpc`,
+        `dist_Mpc`, `dist_pc` or `dist_cm` to `altdata` and
+        those will be picked up (in that order) as the distance
+        rather than the redshift.
 
         Return None if the redshift puts the source not within the Hubble flow
         """
@@ -498,8 +502,11 @@ class Obj(Base, ha.Point):
             if self.redshift * 2.99e5 * u.km / u.s < 350 * u.km / u.s:
                 # stubbornly refuse to give a distance if the source
                 # is not in the Hubble flow
+                # cf. https://www.aanda.org/articles/aa/full/2003/05/aa3077/aa3077.html
+                # within ~5 Mpc (cz ~ 350 km/s) a given galaxy velocty
+                # can be between between ~0-500 km/s
                 return None
-            return (cosmo.luminosity_distance(self.redshift)).value
+            return (cosmo.luminosity_distance(self.redshift)).to(u.Mpc).value
         return None
 
     @property
@@ -518,8 +525,7 @@ class Obj(Base, ha.Point):
                 # see eq (20) of https://ned.ipac.caltech.edu/level5/Hogg/Hogg7.html
                 return dl / (1 + self.redshift) ** 2
             return dl
-        else:
-            return None
+        return None
 
     def airmass(self, telescope, time, below_horizon=np.inf):
         """Return the airmass of the object at a given time. Uses the Pickering
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -4,7 +4,6 @@ from datetime import datetime, timezone
 import arrow
 from astropy import units as u
 from astropy import time as ap_time
-from astropy import cosmology
 
 import astroplan
 import numpy as np
@@ -22,6 +21,7 @@ from sendgrid.helpers.mail import Mail
 from astropy import coordinates as ap_coord
 import healpix_alchemy as ha
 import timezonefinder
+from .utils.cosmology import establish_cosmology
 
 from baselayer.app.models import (  # noqa
     init_db,
@@ -35,7 +35,6 @@ from baselayer.app.models import (  # noqa
 )
 from baselayer.app.custom_exceptions import AccessError
 from baselayer.app.env import load_env
-
 
 from . import schema
 from .enum_types import (
@@ -54,8 +53,8 @@ from skyportal import facility_apis
 PHOT_ZP = 23.9
 PHOT_SYS = 'ab'
 
-# TODO: make this configurable
-cosmo = cosmology.WMAP9
+_, cfg = load_env()
+cosmo = establish_cosmology(cfg)
 
 
 def is_owned_by(self, user_or_token):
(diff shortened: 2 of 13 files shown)


profjsb:
[ch1226]

profjsb:
(failed test appears to be unrelated)

skyportal/models.py
```diff
@@ -459,6 +463,64 @@ def gal_lon_deg(self):
...
+                return (float(self.altdata.get("dist_kpc")) * 1e-3 * u.Mpc).value
+            if self.altdata.get("dist_Mpc") is not None:
+                return (float(self.altdata.get("dist_Mpc")) * u.Mpc).value
+            if self.altdata.get("dist_pc") is not None:
+                return (float(self.altdata.get("dist_pc")) * 1e-6 * u.Mpc).value
+            if self.altdata.get("dist_cm") is not None:
+                return (float(self.altdata.get("dist_cm")) * u.Mpc / 3.085e18).value
+
+        if self.redshift:
+            if self.redshift * 2.99e5 * u.km / u.s < 350 * u.km / u.s:
```
dannygoldstein, line 498: Citation for 350kms?

skyportal/models.py
```diff
@@ -459,6 +463,64 @@ def gal_lon_deg(self):
...
+    @property
+    def angular_diameter_distance(self):
+        dl = self.luminosity_distance
+        if dl:
+            if self.redshift and self.redshift * 2.99e5 * u.km / u.s > 350 * u.km / u.s:
+                # see eq (20) of https://ned.ipac.caltech.edu/level5/Hogg/Hogg7.html
+                return dl / (1 + self.redshift) ** 2
+            return dl
+        else:
+            return None
```
dannygoldstein, line 522:
```suggestion
        return None
```

skyportal/models.py
```diff
@@ -459,6 +463,64 @@ def gal_lon_deg(self):
...
+                return (float(self.altdata.get("dist_pc")) * 1e-6 * u.Mpc).value
+            if self.altdata.get("dist_cm") is not None:
+                return (float(self.altdata.get("dist_cm")) * u.Mpc / 3.085e18).value
+
+        if self.redshift:
+            if self.redshift * 2.99e5 * u.km / u.s < 350 * u.km / u.s:
+                # stubbornly refuse to give a distance if the source
+                # is not in the Hubble flow
+                return None
+            return (cosmo.luminosity_distance(self.redshift)).value
```
dannygoldstein, line 502:
```suggestion
            return (cosmo.luminosity_distance(self.redshift)).to(u.Mpc).value
```

skyportal/models.py
```diff
@@ -459,6 +463,64 @@ def gal_lon_deg(self):
...
+        """
+        The luminosity distance in Mpc, using either DM or distance data
+        in the altdata fields or using the cosmology/redshift
+
+        Return None if the redshift puts the source not within the Hubble flow
+        """
+
+        # there may be a non-redshift based measurement of distance
+        # for nearby sources
+        if self.altdata:
```
dannygoldstein, line 477: Is there a way to incorporate photo-z's here? Or if there are somehow multiple z estimates (no firm spec-z)?

skyportal/utils/cosmology.py
```diff
@@ -0,0 +1,64 @@
+from astropy import cosmology
+from astropy import units as u
+
+from baselayer.log import make_log
+
+log = make_log('cosmology')
+
+
+def establish_cosmology(cfg={}, fallback_cosmology=cosmology.Planck18_arXiv_v2):
```
dannygoldstein, line 9: Why have a fallback cosmology defined here and not in `config.yaml.defaults`?

skyportal/utils/cosmology.py
```diff
@@ -0,0 +1,64 @@
...
+    def _get_cosmology():
+        if cfg.get('misc'):
+            if cfg["misc"].get('cosmology'):
+                log("here1")
```
dannygoldstein, line 13:
```suggestion
```

skyportal/utils/cosmology.py
```diff
@@ -0,0 +1,64 @@
...
+                if cfg["misc"]["cosmology"] in cosmology.parameters.available:
+                    cosmo = cosmology.default_cosmology.get_cosmology_from_string(
+                        cfg["misc"]["cosmology"]
+                    )
+                elif isinstance(cfg["misc"]["cosmology"], dict):
+                    log("here")
```
dannygoldstein, line 19:
```suggestion
```

skyportal/utils/cosmology.py
```diff
@@ -0,0 +1,64 @@
...
+                    par = cfg["misc"]["cosmology"]
+                    log(par)
```
dannygoldstein, line 21:
```suggestion
```

skyportal/utils/cosmology.py
```diff
@@ -0,0 +1,64 @@
+from astropy import cosmology
+from astropy import units as u
+
+from baselayer.log import make_log
+
+log = make_log('cosmology')
```
dannygoldstein, line 6: I think this can safely be removed? Or is there a way to set log_level = DEBUG?

dannygoldstein (commented):
Looks good. Minor comments inline and 2 bigger picture comments:

1. It would be nice to have a frontend component that summarizes the cosmology assumed, showing the parameters and reference (if any) 

2. I am a bit worried about the pattern of letting the program continue and defaulting to a fallback if a function fails to execute properly. I think we should be letting exceptions propagate up the call stack as they may indicate a fatal error in the program state that should halt the program and alert the user / developer in a way that will not get buried in the logs.

config.yaml.defaults
```diff
@@ -228,6 +228,10 @@ services:
 misc:
   days_to_keep_unsaved_candidates: 7
   public_group_name: "Sitewide Group"
+  # Use a named cosmology from `astropy.cosmology.parameters.available`
+  # or supply the arguments for an FLRW cosmological instance
```
kmshin1397, line 232: I think this comment could be more descriptive in terms of specifying a custom FLRW cosmological instance, i.e. explaining that the `FlatLambdaCDM` and `LambdaCDM` are the two FLRW classes made available based on having a `"flat"` flag set. It seems like from the astropy documentation that there are other classes inheriting from FLRW available that aren't in the utils/cosmology.py (like `FlatwCDM`) so just saying "arguments for an FLRW cosmological instance" may be misleading. Of course, I don't know how relevant those unused cosmologies actually are from practical astronomy sense so this may be a moot point.

skyportal/utils/cosmology.py
kmshin1397: Not sure whether removing or setting to debug is preferable but taking a quick look through the baselayer code, it doesn't seem like the concept of log levels are implemented yet like with the Python logging module. Maybe it's just subtly hidden somewhere, but if not it would probably be generally good to have. 

@stefanv Does baselayer logging have log levels?

kmshin1397 (commented):
Not much to add from me on top of Danny's comments - mostly just the comment on the config.yaml.

skyportal/models.py
profjsb: the API for `luminosity_distance` says it will return in Mpc

skyportal/models.py
profjsb: Done.

skyportal/models.py
profjsb: Done

skyportal/models.py
profjsb: Done.

skyportal/models.py
profjsb: A user (currently only via the API) can add `dm`, `parallax`, `dist_kpc`, `dist_Mpc`, `dist_pc` or `dist_cm` to `altdata` and those will be picked up (in that order) as the distance rather than the redshift.

profjsb: Added some more info in the docstrings about this

skyportal/utils/cosmology.py
profjsb: Because I dont want to the cosmology instantiation to fail if the user makes a mistake.

skyportal/utils/cosmology.py
profjsb: I think it's good (for posterity) to have a record in the logs about what cosmology is being used. It shows up only a boot time, not throughout a running SP.

config.yaml.defaults
profjsb: Done.

profjsb:
@dannygoldstein wrote:
>     1. It would be nice to have a frontend component that summarizes the cosmology assumed, showing the parameters and reference (if any)
> 

Done. Showing the cosmology in the About page.
 
image

>     2. I am a bit worried about the pattern of letting the program continue and defaulting to a fallback if a function fails to execute properly. I think we should be letting exceptions propagate up the call stack as they may indicate a fatal error in the program state that should halt the program and alert the user / developer in a way that will not get buried in the logs.

I disagree -- I dont want the entire app to fail because someone made a mistake in the `misc.cosmology` section of the config file. IMHO it's EXTREMELY unlikely for most use cases in SP that the specific cosmology will matter to the way in which science conducted. What likely matters most is the relative values of redshift, which is data supplied outside of the context of this app.

kmshin1397 (approved):

skyportal/utils/cosmology.py
dannygoldstein: I thought more about this, and am cool with it. Astropy does a similar thing sometimes, with warnings. As long as it is clear in the logs.

Issue #963, opened 2020-09-23
Recently Added Sources should show newly saved Sources


profjsb:
Right now it looks like the Recently Added Sources widget shows when the candidate was observed/or first added to the DB. Instead it should show the most recently saved sources. In the screenshot below the source saved recently isn't showing up.

<img width="535" alt="image" src="https://user-images.githubusercontent.com/480799/93949344-659a0080-fcf5-11ea-8107-448739b24d78.png">

PR #967, opened 2020-09-23
Query for recent sources based on time saved

kmshin1397:
Closes Issue #963


profjsb:
@kmshin1397 can you confirm that the failing tests are not related to this PR?

kmshin1397:
The tests all passed for me locally, and GA and Travis are failing on different tests, so I don't think they are related to the PR. I'll try re-running the CI and see what happens.

profjsb:
Small ask here: I think it would be good to add or modify a test for this -- with just the demo data I cannot see if the change has the desired effect.

kmshin1397:
Sure, that makes sense. I'll add a test

profjsb (approved):
Great!

PR #977, opened 2020-09-24
Parse Spectra POSTed as ASCII into Skyportal `Spectrum` objects

dannygoldstein:
@scizen9

This PR adds a new endpoint, `/api/spectrum/ascii`, which parses ASCII-formatted text into a `skyportal.Spectrum` object, then returns a JSONified `Spectrum` as a response. 

The lines that make up the ASCII header must all be formatted the same way within a single file, and can be formatted in one of two ways. 

1. `# KEY: VALUE` (YAML)
2. `# KEY = VALUE / COMMENT` (representation of FITS header as ASCII).

`astropy.io.ascii.read` is used to load the table into Python memory. An attempt is made to parse the header first using method 1, then method 2. 

The data must be at least 2 column ascii (wavelength, flux). If three columns are given, they are interpreted as (wavelength, flux, fluxerr). If more than 3 columns are given, by default the first three are interpreted as (wavelength, flux, fluxerr). The column indices of each of these arguments can be controlled by passing the integer column index to the POST JSON. 

The tests use 4 random real spectra (each formatted differently) off the marshal to verify that all of the data and header values are correctly parsed.


skyportal/handlers/api/spectrum.py
```diff
@@ -176,6 +177,100 @@ def delete(self, spectrum_id):
         return self.success()
 
 
+class ASCIIHandler(object):
```
profjsb, line 180: Dont need to subclass object in Python 3.x
```suggestion
class ASCIIHandler:
```

skyportal/handlers/api/spectrum.py
```diff
@@ -176,6 +177,100 @@ def delete(self, spectrum_id):
...
+            raise ValidationError('Invalid instrument id.')
+
+        group_ids = json.pop('group_ids')
+        user_group_ids = [g.id for g in self.current_user.accessible_groups]
+        for group_id in group_ids:
+            if group_id not in user_group_ids:
+                raise PermissionError('Insufficient permissions.')
+
+        filename = json.pop('filename')
+        ascii = json.pop('ascii')
```
profjsb, line 207: It's not clear what will happen if filename and/or ascii is None (is that even allowed by the schema)?

skyportal/models.py
```diff
@@ -1577,39 +1584,161 @@ class Spectrum(Base):
     assignment_id = sa.Column(sa.ForeignKey('classicalassignments.id'), nullable=True)
     assignment = relationship('ClassicalAssignment', back_populates='spectra')
 
+    altdata = sa.Column(
+        psql.JSONB, doc="Miscellaneous alternative metadata.", nullable=True
+    )
+
```
profjsb, line 1590:
```suggestion
```

skyportal/models.py
```diff
@@ -1577,39 +1584,161 @@ class Spectrum(Base):
...
+    original_file_string = sa.Column(
+        sa.String,
+        doc="Content of original file that user passed to upload the spectrum.",
```
profjsb, line 1593:
```suggestion
        doc="Content of original file that was passed to upload the spectrum.",
```

skyportal/models.py
```diff
@@ -1577,39 +1584,161 @@ class Spectrum(Base):
...
+    )
+    original_file_filename = sa.Column(
+        sa.String, doc="Name of original file that user passed to upload the spectrum."
```
profjsb, line 1596:
```suggestion
        sa.String, doc="Name of original file name that was passed to upload the spectrum."
```

skyportal/models.py
```diff
@@ -1577,39 +1584,161 @@ class Spectrum(Base):
...
+                    # this line does not raise a warning
+                    card = fits.Card.fromstring(line)
+                    try:
+                        # this line warns (exception in this context)
+                        card.verify()
+                    except AstropyWarning:
+                        continue
+                    cards.append(card)
+
+            # this ensure lines like COMMENT and HISTORY are properly dealt with
```
profjsb, line 1707:
```suggestion
            # this ensures lines like COMMENT and HISTORY are properly dealt with
```

skyportal/models.py
```diff
@@ -1577,39 +1584,161 @@ class Spectrum(Base):
...
+
+        filename = Path(filename)
+
+        if data is None:
+            table = ascii.read(filename, comment='#', header_start=None)
+            bytes = open(filename).read().encode('ascii')
+        else:
+            file_obj = io.BytesIO(data.encode('ascii'))
+            table = ascii.read(file_obj, comment='#', header_start=None)
+            bytes = data
```
profjsb, line 1665: It's very likely that this is the most volatile component of the ingest. You should catch errors thrown by `ascii` read and propagate it to the API response.

skyportal/handlers/api/spectrum.py
```diff
@@ -176,6 +177,100 @@ def delete(self, spectrum_id):
...
+        DBSession().commit()
+        return self.success(data={'id': spec.id})
+
+
+class SpectrumASCIIFileParser(BaseHandler, ASCIIHandler):
+    @permissions(['Upload data'])
+    def post(self):
+        """
+        ---
+        description: Parse spectrum from ASCII file
```
profjsb, line 254: Could give some more detail here about expected/typical formats. For example astropy `ascii` accepts many formats (https://docs.astropy.org/en/stable/io/ascii/#id3).

skyportal/handlers/api/spectrum.py
```diff
@@ -176,6 +177,100 @@ def delete(self, spectrum_id):
...
+        spec = Spectrum.from_ascii(filename, data=ascii, **json)
+        return spec
+
+
+class SpectrumASCIIFileHandler(BaseHandler, ASCIIHandler):
+    @permissions(['Upload data'])
+    def post(self):
+        """
+        ---
+        description: Upload spectrum from ASCII file
```
profjsb, line 217: Do you want the upload to fail even if it cannot get parsed? Perhaps a non-standard file is getting uploaded. Even if it cannot be displayed a user may want to store it.

skyportal/handlers/api/spectrum.py
```diff
@@ -176,6 +177,100 @@ def delete(self, spectrum_id):
...
+                            id:
+                              type: integer
+                              description: New spectrum ID
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+
+        spec = self.spec_from_ascii_request()
```
profjsb, line 243: Set a maximize size. Someone may mistakenly upload a very large file that is parsed but is not a spectrum.

profjsb (changes requested):
Looking good. You are using relatively new "random" spectra for testing...do you have permissions to show these spectra openly and for this purpose?

skyportal/handlers/api/spectrum.py
dannygoldstein: Won't happen, this would be rejected before this point by the marshmallow schema validator.

skyportal/handlers/api/spectrum.py
dannygoldstein: fixed

skyportal/models.py
dannygoldstein: just following black rules here

dannygoldstein: just following black rules here

skyportal/models.py
dannygoldstein: fixed

skyportal/models.py
dannygoldstein: fixed

skyportal/models.py
dannygoldstein: fixed

skyportal/models.py
dannygoldstein: done

skyportal/handlers/api/spectrum.py
dannygoldstein: I added a detailed write up to the API docs

skyportal/handlers/api/spectrum.py
dannygoldstein: Delegated this to a future PR.

dannygoldstein:
@profjsb Comments addressed and the spectra have been purposely corrupted to avoid sharing sensitive data (all fluxes have been significantly modified.

skyportal/handlers/api/spectrum.py
dannygoldstein: Done

dannygoldstein:
@profjsb Here are the new docs

Screen Shot 2020-09-26 at 12 11 06 PM
Screen Shot 2020-09-26 at 12 11 11 PM

dannygoldstein:
green

dannygoldstein:
@profjsb @stefanv re: the commentary on today's Q&A, it's very important to get this merged ASAP. Can you both please re-review? thanks

cc @scizen9

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
-        instrument_id : int
-           ID of the Instrument with which this Spectrum was acquired.
-        observed_at : string or datetime
-           Median UTC ISO time stamp of the exposure or exposures in which the Spectrum was acquired."
+           The original name of the ASCII file containing the spectrum,
+           or the name of the file on disk to read. If `data` is not provided,
+           `from_ascii` tries to read this file and parse its contents. If `data`
+           is provided, the filename is used only for bookkeeping purposes; no
+           attempt is made to actually read the file.
+        data: str, optional
```
stefanv, line 1689: It would be simpler to avoid the joint handling of filename and data.  How about opting for `file` as a parameter, and allow that parameter to be a filename or file-like object.  If you want to read bytes, you put those into BytesIO and pass it in.

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+           filesystem.
+        obj_id : str, optional
+           The id of the Obj that this Spectrum is of, if not present
+           in the ASCII header.
+        instrument_id : int, optional
+           ID of the Instrument with which this Spectrum was acquired,
+           if not present in the ASCII header.
+        observed_at : string or datetime, optional
+           Median UTC ISO time stamp of the exposure or exposures in which
+           the Spectrum was acquired, if not present in the ASCII header."
```
stefanv, line 1704:
```suggestion
           the Spectrum was acquired, if not present in the ASCII header.
```

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+        wave_colindex: integer, optional
```
stefanv, line 1705: `wave_column` ?

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+           The 0-based index of the ASCII column corresponding to the wavelength
+           values of the spectrum (default 0).
+        flux_colindex: integer, optional
```
stefanv, line 1708: `flux_column`?

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+           The 0-based index of the ASCII column corresponding to the flux
+           values of the spectrum (default 1).
+        fluxerr_colindex: integer or None, optional
```
stefanv, line 1711: `fluxerr_column`?

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
            The Spectrum generated from the ASCII file.
 
         """
-        data = np.loadtxt(filename)
-        if data.shape[1] != 2:  # TODO support other formats
-            raise ValueError(f"Expected 2 columns, got {data.shape[1]}")
+
+        filename = Path(filename)
+
+        if data is None:
```
stefanv, line 1727: If we do the`file` argument which accepts filename or file-like, then these two duplicate clauses will collapse.

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+            try:
+                table = ascii.read(filename, comment='#', header_start=None)
+            except Exception as e:
+                e.args = (f'Error parsing ASCII file: {e.args[0]}',)
+                raise
+            bytes = open(filename).read().encode('ascii')
```
stefanv, line 1733: Where's the checking for format 2?

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+                table = ascii.read(file_obj, comment='#', header_start=None)
+            except Exception as e:
+                e.args = (f'Error parsing ASCII file: {e.args[0]}',)
+                raise
+            bytes = data
+
+        tabledata = np.asarray(table)
+        colnames = table.colnames
+
+        if len(colnames) < 2:
```
stefanv, line 1746: Is this enough validation?  Perhaps factor out validation to function and do a slightly more rigorous check

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+            # XTENSION: IMAGE
+            # BITPIX: -32
+            # NAXIS: 2
+            # NAXIS1: 433
+            # NAXIS2: 1
+
+            header = {}
+            for line in table.meta['comments']:
+                try:
+                    result = yaml.load(line, Loader=yaml.FullLoader)
```
stefanv, line 1764: This can't be very effective?  Why not bundle and load all at once?

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+                    except AstropyWarning:
+                        continue
+                    cards.append(card)
+
+            # this ensures lines like COMMENT and HISTORY are properly dealt with
+            fits_header = fits.Header(cards=cards)
+            serialized = dict(fits_header)
+
+            for key in serialized:
+                if isinstance(serialized[key], fits.header._CardAccessor):
```
stefanv, line 1789: This doesn't seem safe; checking for a private type.

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+                        card.verify()
+                    except AstropyWarning:
+                        continue
+                    cards.append(card)
+
+            # this ensures lines like COMMENT and HISTORY are properly dealt with
+            fits_header = fits.Header(cards=cards)
+            serialized = dict(fits_header)
+
+            for key in serialized:
```
stefanv, line 1788: What is this block necessary for?

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+                    serialized[key] = str(serialized[key])
+                if len(fits_header.comments[key]) > 0:
+                    header[key] = {
+                        'value': serialized[key],
+                        'comment': fits_header.comments[key],
+                    }
+                else:
+                    header[key] = serialized[key]
+
+            # coerce things to serializable JSON
```
stefanv, line 1800: Our JSON writer can already handle this?

skyportal/models.py
```diff
@@ -1645,39 +1652,170 @@ class Spectrum(Base):
...
+                if isinstance(header[k], (datetime, date)):
+                    header[k] = header[k].isoformat()
+        else:
+            header = None
 
         return cls(
-            wavelengths=data[:, 0],
-            fluxes=data[:, 1],
+            wavelengths=tabledata[colnames[wave_colindex]],
+            fluxes=tabledata[colnames[flux_colindex]],
```
stefanv, line 1809: No checking that these column indices are valid?

stefanv (commented):
Thanks, Danny.

The ASCII Spectrum reader is eminently testable outside of the frontend tests, and should be tested separately and thoroughly.

See, e.g., the cache tests.

profjsb (approved):

profjsb:
Approved, based on my review. Please address @stefanv's comments before merging.

skyportal/handlers/api/spectrum.py
```diff
@@ -176,6 +177,105 @@ def delete(self, spectrum_id):
...
+        group_ids = json.pop('group_ids')
+        user_group_ids = [g.id for g in self.current_user.accessible_groups]
+        for group_id in group_ids:
+            if group_id not in user_group_ids:
+                raise PermissionError('Insufficient permissions.')
+
+        filename = json.pop('filename')
+        ascii = json.pop('ascii')
+
+        # maximum size 200MB - above this don't parse. Assuming ~1 byte / char
```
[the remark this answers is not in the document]
dannygoldstein: changed to 10MB.

skyportal/handlers/api/spectrum.py
```diff
@@ -176,6 +177,105 @@ def delete(self, spectrum_id):
...
+                            id:
+                              type: integer
+                              description: New spectrum ID
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+
+        spec = self.spec_from_ascii_request()
```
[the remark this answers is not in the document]
dannygoldstein: Done

skyportal/models.py
dannygoldstein: fixed

skyportal/models.py
dannygoldstein: done

skyportal/models.py
dannygoldstein: done

skyportal/models.py
dannygoldstein: done

skyportal/models.py
dannygoldstein: The reason I do it line by line is because, if we parse the whole header as yaml and a single comment line is not formatted as yaml but is instead of a different form, e.g., the third row here:

``` 
# key1: value1
# key2: value2
# col1 col2 col3 col4 
```

then parsing of the entire header will fail. Having one or two generic comment lines in the header is common: they often contain column labels or some similar data that we don't need to keep track of or use in any way. Similar to how we don't want an improperly configured cosmology to cause the entire app to crash, we don't want a single misformatted header line to cause the entire upload to fail. 

On the frontend, we will use the `SpectrumParser` handler to return the parsed ascii spectrum for users to inspect before they can click an upload button to actually post the spectrum to the database.

skyportal/models.py
dannygoldstein: `fits.Header` exposes a dict-like interface but its serialization of key: value mappings to pure python types is not perfect. Some values are left as `astropy` types or other types that are not directly serializable to JSON with our `JSON` encoder. This block handles the edge cases of JSON-serialization from serialized fits headers. It also ensures that comments on key: value mappings are propagated to the database if they are present.

skyportal/models.py
dannygoldstein: Thanks, this is much nicer.

skyportal/models.py
dannygoldstein: Thanks, this is much nicer.

skyportal/models.py
dannygoldstein: Starts on line 1776 in this diff.

skyportal/models.py
dannygoldstein: I beefed up this section, adding some more checks.

skyportal/models.py
dannygoldstein: Moved this part up, to the table validation section.

dannygoldstein: Added index validity checks in that section too.

skyportal/models.py
dannygoldstein: I remember having a good reason for putting this in when I did, but I commented it out and the tests pass.

skyportal/models.py
dannygoldstein: Fixed this to be safer.

skyportal/models.py
```diff
@@ -1645,39 +1650,190 @@ class Spectrum(Base):
...
         """Generate a `Spectrum` from an ascii file.
 
         Parameters
         ----------
-
-        filename : str
-           The name of the ASCII file containing the spectrum.
+        file : str or file-like object
+           Name or handle of the ASCII file containing the spectrum. If `data`
+           is not provided, `from_ascii` tries to read this file and parse its
```
stefanv, line 1682: Presumably data is gone now?

stefanv (commented):
Looks great, thanks for the tests!

Do the new tests cover both file formats supported?

dannygoldstein:
@stefanv Yes they do.

dannygoldstein:
GA failures unrelated

skyportal/models.py
dannygoldstein: Now I remember why I put this in. It's for serializing *to* the database, not for serializing *from* the database. I.e., to avoid errors like ```JSON response: {‘data’: {}, ‘version’: ‘0.9.dev0’, ‘status’: ‘error’, ‘message’: “(builtins.TypeError) Object of type datetime is not JSON serializable\n[SQL: INSERT INTO spectra (created_at, modified, wavelengths, fluxes, errors, obj_id, observed_at, origin, instrument_id, followup_request_id, assignment_id, altdata, original_file_string, original_file_filename) VALUES (timezone(%(timezone_1)s, CURRENT_TIMESTAMP), timezone(%(timezone_1)s, CURRENT_TIMESTAMP), %(wavelengths)s, %(fluxes)s, %(errors)s, %(obj_id)s, %(observed_at)s, %(origin)s, %(instrument_id)s, %(followup_request_id)s, %(assignment_id)s, %(altdata)s, %(original_file_string)s, %(original_file_filename)s) RETURNING spectra.id]\n[parameters: [{‘original_file_string’: ‘# SIMPLE: True\\n# BITPIX: -64\\n# NAXIS: 2\\n# NAXIS1: 214\\n# NAXIS2: 220\\n# EXTEND: True\\n# CDELT1: 25.570776255704914\\n# CRVAL1: 3776.7123287671147\\n ... (9831 characters truncated) ... 556e-33\\n9121.0 1.591e-16 1.680e-33\\n9146.6 1.619e-16 1.214e-33\\n9172.1 1.528e-16 1.309e-33\\n9197.7 1.428e-16 1.388e-33\\n9223.3 1.357e-16 1.630e-33\\n’, ‘observed_at’: datetime.datetime(2020, 9, 1, 5, 42, 29, tzinfo=datetime.timezone.utc), ‘obj_id’: ‘ZTF20abxgwjr’, ‘altdata’: {‘SIMPLE’: True, ‘BITPIX’: -64, ‘NAXIS’: 2, ‘NAXIS1’: 214, ‘NAXIS2’: 220, ‘EXTEND’: True, ‘CDELT1’: 25.570776255704914, ‘CRVAL1’: 3776.7123287671147,  ... (3960 characters truncated) ...  ‘SNIDAGEMED’: -6.15, ‘SNIDAGEMEDERR’: 5.38, ‘SNIDFRAC_IA’: 0.97, ‘SNIDFRAC_IB’: 0.01, ‘SNIDFRAC_IC’: 0.02, ‘SNIDFRAC_II’: 0.0, ‘SNIDFRAC_NOTSN’: 0.0}, ‘instrument_id’: 2, ‘wavelengths’: array([3776.7, 3802.3, 3827.9, 3853.4, 3879. , 3904.6, 3930.1, 3955.7,\n       3981.3, 4006.8, 4032.4, 4058. , 4083.6, 4109.1, 4134.7, 4160.3,\n       4 ... (1601 characters truncated) ... 8814.2, 8839.7, 8865.3,\n       8890.9, 8916.4, 8942. , 8967.6, 8993.2, 9018.7, 9044.3, 9069.9,\n       9095.4, 9121. , 9146.6, 9172.1, 9197.7, 9223.3]), ‘fluxes’: array([ 5.403e-16,  1.065e-16, -6.951e-17,  1.738e-16,  3.601e-16,\n        3.258e-16,  3.241e-16,  4.001e-16,  4.439e-16,  5.263e-16,\n        7.520e-1 ... (2569 characters truncated) ... 16,  1.504e-16,  1.566e-16,\n        1.523e-16,  1.276e-16,  1.227e-16,  1.423e-16,  1.591e-16,\n        1.619e-16,  1.528e-16,  1.428e-16,  1.357e-16]), ‘original_file_filename’: ‘spec_auto_contsep_lstep1__crr_b_ifu20200901_05_42_29_ZTF20abxgwjr.txt’, ‘errors’: None, ‘assignment_id’: None, ‘followup_request_id’: None, ‘origin’: None}]]“}
```

PR #978, opened 2020-09-25
improve stability of followup request tests

dannygoldstein:
The followup request tests have been failing recently because they wait for the plots to load, and the plots have been taking a very long time to load on CI (>10s) causing the tests to timeout. 

This PR replaces the previous method of dealing with flakiness in the followup request frontend tests (refreshing the page and trying again in case of a plot timeout), and simply executes the page load once, but with an increased  timeout on the plot load from 10 to 20s.

### DIFF
--- a/skyportal/tests/frontend/test_followup_requests.py
+++ b/skyportal/tests/frontend/test_followup_requests.py
@@ -1,7 +1,6 @@
 import uuid
 import pytest
 import requests
-from selenium.common.exceptions import TimeoutException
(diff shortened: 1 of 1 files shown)


dannygoldstein:
I have a feeling that the slow page loading is also causing the other failures we have been seeing recently on CI.

profjsb (approved):
This looks good. We may still want to add the TimeoutException if these continue to fail.

Issue #961, opened 2020-09-22
Querying metadata associated with obj's


dmitryduev:
We need the ability to query obj by metadata stored in either comments or alt_data. 

We need to explore this in detail.

- Allow storing collections in comments and not just strings?
- Use alt_data to store structured (meta)data, e.g. auto-annotations? In this case, it needs to be exposed on a number of pages.
- Ability to query and filter obj's based on the metadata. Use something like `https://github.com/kapouille/mongoquery`? Or query postgres' jsonb fields directly?

profjsb:
The good news is that the query engine on JSONB is JSONPATH (like XPATH) and is fully implemented in PG12:
https://www.postgresql.org/docs/12/functions-json.html#FUNCTIONS-SQLJSON-PATH

dannygoldstein:
Right now autoannotations are implemented as comments. But autoannotations, unlike comments, need to be searchable, filterable, sortable, and storable as not just plain text but also as floating points and integers. 

I'd advocate for developing a separate `Annotation` model that encapsulates these notions, with the main column storing JSONB @profjsb and @dmitryduev suggest.

guynir42:
I am now working on this, and wanted to know if right now we are generating any auto-annotations and where in the code they are generated? Thanks!

dannygoldstein:
Fixed by #979 and #1068

PR #979, opened 2020-09-25
Auto Annotations: Backend

guynir42:
Attempts to fix issue #961 

I have added two fields to the comment model:
-is_auto which is true/false
-altdata which is a parsed JSONB field that can contain all sorts of data. 

The idea is that when you input "z=0.5" that would be registered into altdata and so can be searchable. 

The is_auto needs to be specified by any API call from any robots making comments (I still have not found where such comments are made). 

I've also added a simple parser that handles human comments given as keyword=value to also be added into altdata even when is_auto is false. 

I don't think this is done but it is a starting point so please tell me what else we need to do. 

Also, I am not sure but I think this requires some changes to the database.


skyportal/handlers/api/comment.py
```diff
@@ -172,6 +189,50 @@ def post(self):
...
+                    except ValueError:
+                        continue  # bad input, skip to next pair
+
+                    if value:
+                        if altdata is None:
+                            altdata = {}
+                        if instruction[0] not in altdata:
+                            altdata[instruction[0]] = value
+
+        return altdata
```
dannygoldstein, line 234: I am not sure we want to be getting into the business of writing API middleware to parse user comments like this. The logic here is already somewhat complicated (strip this, lower case that, check the data type, etc. -- many places for subtle issues to crop up when dealing with real-world user data). Once we set the precedent of doing this, we open ourselves up to having to write all kinds of custom parsers, each one adding a new layer of complexity to the backend. I say we should let the data types speak contractually for themselves, and, to the extent possible, not deviate from CRUD.

skyportal/handlers/api/comment.py
```diff
@@ -64,6 +64,15 @@ def post(self):
...
                         type: string
+                  is_auto:
+                    type: boolean
+                    description: |
+                        specify True when a robot makes the comment.
+                        Defaults to False for human authors.
+                  altdata:
+                    type: object
+                    description: machine readable (searchable) data in JSON format.
+
```
dannygoldstein, line 75: `altdata` and `text` creep into each other here somewhat. The content of `text` influences the content of `altdata`, and the content of `altdata` is handled different depending on whether `is_auto` is flagged. I think this is a sign that a separate database model is needed for autoannotations.  To the extent possible, I would like to keep our API for individual database classes as close to basic CRUD operations (https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) as possible. This makes the API contract presented to the user extremely clear, improves the long-term maintainability of the code, and makes it easier to extend the data model later on.

dannygoldstein (changes requested):
I worry that the data model here overloads the `Comment` object beyond what it was designed to accommodate. I think it would be simplest to break out the notion of arbitrary JSON user data to an `Annotation` object that has its own handler and frontend components (i.e., displayed in a materialui-datatable, instead of a CommentList).

guynir42:
I did a lot of copy-paste for this. I'm not sure if this is better than combining the comments and annotations, or at least having annotations inherit from comments. 

Either way, I couldn't find where to get a new ACL for annotations, so I left it the same ACL as comments.

dannygoldstein:
@guynir42 Looks like maybe something is not being imported, causing the docs build to fail?

skyportal/handlers/api/annotation.py
```diff
@@ -0,0 +1,263 @@
...
+            data=annotation_data,
+            obj_id=obj_id,
+            origin=origin,
+            author=author,
+            groups=groups,
+        )
+
+        DBSession().add(annotation)
+        if 'redshift' in annotation_data:
+            obj.redshift = annotation_data['redshift']
```
dannygoldstein, line 153:
```suggestion
```

We dont want to handle redshifts here. There is going to be a separate handler for redshifts based on the classifications API

skyportal/handlers/api/annotation.py
```diff
@@ -0,0 +1,263 @@
...
+
+        DBSession().commit()
+
+        self.push_all(
+            action='skyportal/REFRESH_SOURCE',
+            payload={'obj_key': annotation.obj.internal_key},
+        )
+        return self.success(data={'annotation_id': annotation.id})
+
+    @permissions(['Annotation'])
```
dannygoldstein, line 163:
```suggestion
    @permissions(['Annotate'])
```

skyportal/model_util.py
```diff
@@ -6,6 +6,7 @@
 all_acl_ids = [
     'Become user',
     'Comment',
+    'Annotation',
```
dannygoldstein, line 9:
```suggestion
    'Annotate',
```

skyportal/models.py
```diff
@@ -1369,6 +1408,64 @@ def get_if_owned_by(cls, ident, user, options=[]):
...
+        "User", back_populates="annotations", doc="Annotation's author.", uselist=False,
+    )
+    author_id = sa.Column(
+        sa.ForeignKey('users.id', ondelete='CASCADE'),
+        nullable=False,
+        index=True,
+        doc="ID of the Annotation author's User instance.",
+    )
+
+    origin = sa.Column(sa.String, doc='Annotation origin.')
```
dannygoldstein, line 1425:
```suggestion
    origin = sa.Column(sa.String, doc='Annotation origin.', index=True, nullable=False)
```

dannygoldstein: Can you expand the docs here somewhat?

skyportal/handlers/api/annotation.py
```diff
@@ -0,0 +1,263 @@
...
+        obj_id = data.get("obj_id")
+        origin = data.get("origin")
+
+        if obj_id is None:
+            return self.error("Missing required field `obj_id`")
+
+        if origin is None or origin == '':
+            return self.error('Missing required field `origin`')
+
+        annotation_data = data.get("data")
```
dannygoldstein, line 90: This needs to be validated

skyportal/tests/api/test_annotations.py
```diff
@@ -0,0 +1,296 @@
...
+        data={
+            'obj_id': public_source.id,
+            'data': {'redshift': 0.15},
+            'group_ids': [public_group.id],
+        },
+        token=annotation_token,
+    )
+
+    assert status == 400
+    assert data["message"] == "Missing required field `origin`"
```
dannygoldstein, line 20: This will need to be changed when the validation is handled via a Marshmallow schema

skyportal/tests/api/test_annotations.py
```diff
@@ -0,0 +1,296 @@
...
+
+    # this should not work, since "origin" is empty
+    status, data = api(
+        'POST',
+        'annotation',
+        data={
+            'obj_id': public_source.id,
+            'origin': '',
```
dannygoldstein, line 28: The empty string should be allowed as a valid origin. It should only barf when origin is null

skyportal/tests/api/test_annotations.py
```diff
@@ -0,0 +1,296 @@
...
+def test_post_same_origin_fails(annotation_token, public_source, public_group):
+
+    # first time adding an annotation to this object from Kowalski
+    status, data = api(
+        'POST',
+        'annotation',
+        data={
+            'obj_id': public_source.id,
+            'origin': 'kowalski',
+            'data': {'redshift': 0.15},
```
dannygoldstein, line 48: Defer redshift handling to redshift handler - to be implemented cc @acrellin

skyportal/tests/api/test_annotations.py
```diff
@@ -0,0 +1,296 @@
...
+            'group_ids': [public_group.id],
+        },
+        token=annotation_token,
+    )
+
+    assert status == 400
+    assert 'duplicate key value violates unique constraint' in data["message"]
```
dannygoldstein, line 71: I would refrain from checking the messages as this may increase the brittleness of the test suite but I defer to @stefanv  on this if he thinks otherwise

skyportal/tests/api/test_annotations.py
```diff
@@ -0,0 +1,296 @@
...
+
+    assert status == 200
+    annotation_id = data['data']['annotation_id']
+
+    # This token belongs to public_group2
+    status, data = api(
+        'GET', f'annotation/{annotation_id}', token=annotation_token_two_groups
+    )
+    assert status == 200
+    assert data['data']['data'] == {'redshift': 0.15}
```
dannygoldstein, line 146:
```suggestion
    assert data['data']['data'] == {'sdss_redshift': 0.15}
```
Suggest replacing this throughout the test suite with `sdss_redshift` or similar to avoid any confusion about how and where redshifts are set

dannygoldstein (changes requested):
Few changes requested below, thanks for the thorough test suite.

skyportal/handlers/api/annotation.py
dannygoldstein: I.e., need to validate that `data.get('data')` returns valid JSON

skyportal/handlers/api/annotation.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        data = self.get_json()
+        obj_id = data.get("obj_id")
+        origin = data.get("origin")
+
+        if obj_id is None:
+            return self.error("Missing required field `obj_id`")
+
+        if origin is None or origin == '':
+            return self.error('Missing required field `origin`')
```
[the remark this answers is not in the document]
guynir42: Ok I will try to find some examples.

skyportal/handlers/api/annotation.py
guynir42: I know, but this functionality exists in the current Comment code, so I thought it would be good to pull it out only after we have something else instead.

skyportal/models.py
guynir42: yes, good idea.

skyportal/handlers/api/annotation.py
guynir42: I need to check that I get a python dictionary?

skyportal/tests/api/test_annotations.py
guynir42: I'm not sure about this. If people are allowed to enter empty origin it could cause a lot of trouble... Part of the whole annotations API is that you MUST give an origin when posting these. If '' is legal how is that enforced?

skyportal/tests/api/test_annotations.py
guynir42: I would prefer to use more realistic examples, both for the data and the origin. Can you think of something more concrete than "kowalski" and "redshift"?

skyportal/handlers/api/annotation.py
dannygoldstein: not just any python dictionary, but valid JSON. if you use a marshmallow schema this validation will be handled automatically for you.

skyportal/tests/api/test_annotations.py
dannygoldstein: `''` is not an empty origin. It is the empty string, which is a valid string. If you want to require that an origin string consist of at least 1 alphanumeric character then that is not what is being validated in the current handler. For example, I could post `' '` and the handler as written would accept that as valid user input. That is fine. As long as it is a unique string, that is OK. We should, to the extent possible, let data types specify valid contracts with the users.

skyportal/tests/api/test_annotations.py
dannygoldstein: Kowalski is fine, and I would use something like `offset_from_host_galaxy`

skyportal/tests/api/test_annotations.py
guynir42: I see your point. But if we don't limit anything (maybe 1+ alphanumeric is good) then it leaves the option for abusing the system. If I'm writing a bot and I don't have patience to read the definition of "origin" I could just leave it empty. Or put a " " or a "-" as input until it works and I wouldn't care if that is actually not informative. 

I think we should enforce some reasonable constraint on this. It is easier than "educating" the users to be good and properly fill out their origin...

skyportal/handlers/api/annotation.py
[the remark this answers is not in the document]
guynir42: I added a Annotations.__schema__() and used it to validate inputs but now it complains that it doesn't get an author_id and that it doesn't recognize group_ids.

skyportal/handlers/api/annotation.py
```diff
@@ -0,0 +1,279 @@
...
+        """
+        data = self.get_json()
+
+        group_ids = data.pop('group_ids', None)
+
+        schema = Annotation.__schema__(exclude=["author_id"])
+        try:
+            schema.load(data)
+        except ValidationError as e:
+            return self.error(f'Invalid/missing parameters: {e.normalized_messages()}')
```
dannygoldstein, line 99: Excellent

skyportal/models.py
```diff
@@ -1369,6 +1408,69 @@ def get_if_owned_by(cls, ident, user, options=[]):
...
+
+        if annotation is not None and not annotation.is_owned_by(user):
+            raise AccessError('Insufficient permissions.')
+
+        # Grab basic author info for the annotation
+        annotation.author_info = annotation.construct_author_info_dict()
+
+        return annotation
+
+    __table_args__ = (UniqueConstraint('obj_id', 'origin'),)
```
dannygoldstein, line 1465: Move to top of class

dannygoldstein (approved):
Excellent work @guynir42

skyportal/tests/api/test_annotations.py
```diff
@@ -0,0 +1,276 @@
+from skyportal.tests import api
+
+
+def test_post_without_origin_fails(annotation_token, public_source, public_group):
```
kmshin1397, line 4: I love how thorough the tests are for the annotations. Great work!

kmshin1397 (approved):
@guynir42 I have one very small change (just copy-paste to duplicate a recent change to Comments) but otherwise looks great to me! Ping me on Slack when you get that in and I'll merge this.

skyportal/handlers/api/annotation.py
```diff
@@ -0,0 +1,279 @@
...
+            ]
+        else:
+            candidate_group_ids = []
+        matching_sources = (
+            DBSession().query(Source).filter(Source.obj_id == obj_id).all()
+        )
+        if matching_sources:
+            source_group_ids = [source.group_id for source in matching_sources]
+        else:
+            source_group_ids = []
```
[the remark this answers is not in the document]
guynir42: OK I just uploaded it. You can make sure it is ok and merge.

kmshin1397 (approved):
Great, will merge when green

PR #985, opened 2020-09-25
fix default zoom on photometry plot

### LINKED ISSUE #969: Fix the default zoom on the Bokeh plot to account for non-detections
Discovered by @stefanv:

This is what it looks like currently:
image

This is what it should be like:
image

dannygoldstein:
Fixes #969. 

This PR updates the photometry plot to make the default zoom take non-detections into account, showing both detections and non detections by default in magnitude space. 
Screen Shot 2020-09-25 at 3 18 32 PM


dmitryduev (approved):
Looks great, thanks for the fix, @dannygoldstein !

PR #987, opened 2020-09-25
Add absolute magnitude to the photometry plot right axis if `obj.dm is not None`

### LINKED ISSUE #975: Requested feature: 4 axis plots
It has been requested that we add a days-from-today x-axis and an abs mag y-axis to light curve plots.

dannygoldstein:
[ch1266]

Closes #975 . 
Screen Shot 2020-09-25 at 4 14 13 PM

This PR adds an absolute magnitude axis to the photometry plot if `obj.dm is not None`, i.e., if the object has a distance modulus. 

If the distance modulus is not `None`, the loaded cosmology is used to calculate the distance modulus.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -606,7 +606,7 @@ def photometry_plot(obj_id, user, width=600, height=300):
             "Absolute Mag": Range1d(start=ymax - obj.dm, end=ymin - obj.dm)
         }
         plot.add_layout(
-            LinearAxis(y_range_name="Absolute Mag", axis_label="m - μ"), 'right'
+            LinearAxis(y_range_name="Absolute Mag", axis_label="m - DM"), 'right'
         )
 
     toggle = CheckboxWithLegendGroup(
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -606,7 +606,7 @@ def photometry_plot(obj_id, user, width=600, height=300):
             "Absolute Mag": Range1d(start=ymax - obj.dm, end=ymin - obj.dm)
         }
         plot.add_layout(
-            LinearAxis(y_range_name="Absolute Mag", axis_label="Absolute Mag"), 'right'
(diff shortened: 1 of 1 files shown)


skyportal/plot.py
```diff
@@ -600,6 +600,15 @@ def photometry_plot(obj_id, user, width=600, height=300):
     plot.yaxis.axis_label = 'AB mag'
     plot.toolbar.logo = None
 
+    obj = DBSession().query(Obj).get(obj_id)
+    if obj.dm is not None:
+        plot.extra_y_ranges = {
+            "Absolute Mag": Range1d(start=ymax - obj.dm, end=ymin - obj.dm)
+        }
+        plot.add_layout(
+            LinearAxis(y_range_name="Absolute Mag", axis_label="m - μ"), 'right'
```
profjsb, line 609: This labeling is really confusing since you have to know how to orient your head to understand that the axis is saying "m - μ" (otherwise it looks like garbage). Maybe say `M_<sub>absolute</sub> (no K-correction)`?

profjsb (changes requested):

skyportal/plot.py
dannygoldstein: I changed it to m - DM. I considered `M_abs (no K-correction)` but this leaves out the fact that there is also no EBV correction. Adding the "no ebv correction" makes the label too long. I think this is more readable than m - µ was, and is also sufficiently precise to avoid any confusion about what corrections (or lack thereof) have been made to the photometry

dannygoldstein: Screen Shot 2020-09-26 at 12 09 34 PM

kmshin1397 (approved):
@dannygoldstein Merge conflict from my merging the "days ago" x-axis, but otherwise looks good to me

dannygoldstein:
@kmshin1397 fixed merge conflicts

profjsb (approved):

Issue #949, opened 2020-09-18
Query for new photometry and spectra


cfremling:
We need a way to query for data from specific telescopes. For example, to see all sources that have been observed within a certain time interval and to list the observations (photometry or spectra) that were obtained. e.g., recent photometry from LCO-1m, LCO-2m, recent spectra from P200/DBSP, Keck/LRIS

acrellin:
Addressed by #997

dannygoldstein:
Closed by #1507

PR #997, opened 2020-09-27
Add an endpoint to query for photometry taken on certain instruments over a time range

dannygoldstein:
Addresses #949 (for photometry).

@cfremling.

This PR adds the following new endpoint:

Screen Shot 2020-09-26 at 8 47 21 PM


profjsb (approved):
Looks very good. My only suggestion would be to change the endpoint from `/recent/` to something else: there is nothing in the capability that requires the photometry to be recent. Maybe `/range/'? Change the handler names accordingly.

guynir42 (approved):
LGTM!

dannygoldstein:
> Looks very good. My only suggestion would be to change the endpoint from `/recent/` to something else: there is nothing in the capability that requires the photometry to be recent. Maybe `/range/'? Change the handler names accordingly.

Fixed

dannygoldstein:
This is ready for a relook @profjsb

dannygoldstein:
Ping @profjsb

PR #1003, opened 2020-09-28
Fix scanning page warning

kmshin1397:
Fix warning from recent scanning page candidates-per-page fix
Screenshot 2020-09-28 150525


dannygoldstein (approved):
Failures unrelated

Issue #951, opened 2020-09-19
Sharing data is too restrictive


dmitryduev:
SP complains that the user must be a member of the target group, including the single-user groups:
![image](https://user-images.githubusercontent.com/7557205/93662502-7278d480-fa15-11ea-98a5-866974bf8730.png)

Expected behavior: user can choose to share the data with anyone; (optionally) add a step where the target group can either accept or reject the offered data.

PR #1004, opened 2020-09-28
Allow sharing of data with any group(s)

acrellin:
Closes https://github.com/skyportal/skyportal/issues/951


dmitryduev:
Thanks, @acrellin. Thinking about this a bit more, let's leave the things as they are now, this seems a bit dangerous from the permissioning point of view.

Issue #982, opened 2020-09-25
anyone should be able to create a group


dannygoldstein:
Currently I have no option to create a group on my groups page: 
<img width="540" alt="Screen Shot 2020-09-25 at 11 18 25 AM" src="https://user-images.githubusercontent.com/2769632/94302295-da9d4e00-ff20-11ea-9287-ff43a685c874.png">

PR #1006, opened 2020-09-28
Allow anyone to create groups

acrellin:
This PR changes the permissions requirements around creating and editing groups, allowing anyone to create a group (which they are then an admin of by default) and allowing anyone who is an admin of a group to modify/delete that group. This allows for users to delete groups that they have created.

Closes https://github.com/skyportal/skyportal/issues/982 

anyone_can_create_groups


dannygoldstein:
Looks like there is an issue with make demo data?

acrellin:
> Looks like there is an issue with make demo data?

@dannygoldstein green now. GA failure is `test_news_feed_prefs_widget` (unrelated)

dannygoldstein (approved):

PR #1009, opened 2020-09-29
Use both date and time in candidate filtering form

acrellin:
image

Closes https://github.com/fritz-marshal/fritz/issues/43


acrellin:
@stefanv Updated screenshots with your suggestions implemented: 
Empty:
image
Selected:
image

PR #1015, opened 2020-09-29
Add functionality to add groups to an existing source

### LINKED ISSUE #966: Previously saved candidates cannot be saved to a new group via frontend
Probably should add a "save to new group" button to either the scanning page or the candidate/source page

kmshin1397:
- On scanning page, previously saved sources info includes the groups that the source is currently saved to
- Add buttons to "re-save" a candidate to new groups on both the scanning page and the Source page
- Update recent sources widget to be "Recently Saved Sources", with sources that are re-saved showing up with a tag indicating so
- Update news feed to distinguish between a candidate being saved for the first time and re-saved
addgroup_scan
addgroup_source
Screenshot 2020-09-29 160705
Screenshot 2020-09-29 154352

Closes #966


dmitryduev (changes requested):
Looks great, thanks @kmshin1397! My one comment is that I think instead of disabling the buttons when there are no more user-accessible groups left, it would make more sense to `{dislpay: none}` -- could you please implement that?

kmshin1397:
> Looks great, thanks @kmshin1397! My one comment is that I think instead of disabling the buttons when there are no more user-accessible groups left, it would make more sense to `{dislpay: none}` -- could you please implement that?

Implemented

dmitryduev (approved):
👍

static/js/components/AddSourceGroup.jsx
```diff
@@ -0,0 +1,174 @@
...
+  const classes = useStyles();
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  // Dialog logic:
+
+  const dispatch = useDispatch();
+  const [dialogOpen, setDialogOpen] = useState(false);
+
+  const { handleSubmit, errors, reset, control, getValues } = useForm();
+
+  useEffect(() => {
```
acrellin, line 41: As this file looks to be a slightly modified version of SaveCandidateButton.jsx, I'm wondering if there's a way to generalize that one enough to be used both places so as to not have so much duplicated code. Thoughts?

static/js/components/AddSourceGroup.jsx
```diff
@@ -0,0 +1,174 @@
...
+    } else if (result.status === "error") {
+      setIsSubmitting(false);
+    }
+  };
+
+  return (
+    <>
+      <div style={displayStyle(unsavedGroups.length !== 0)}>
+        {icon ? (
+          <Tooltip title="Add a group to this source">
```
acrellin, line 86:
```suggestion
          <Tooltip title="Save source to new group(s)">
```

acrellin (commented):
Looks excellent, thanks @kmshin1397 . A few minor comments/requests below

skyportal/handlers/api/internal/recent_sources.py
```diff
@@ -42,24 +42,52 @@ def get_recent_source_ids(self, current_user):
...
                 obj_id,
                 self.current_user,
                 options=[joinedload(Source.obj).joinedload(Obj.thumbnails)],
             )
+
+            # Get the entry in the Source table to get the accurate saved_at time
+            source_entry = (
+                Source.query.filter(Source.obj_id == obj_id)
+                .order_by(desc('created_at'))
+                .offset(recency_index)
```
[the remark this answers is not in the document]
kmshin1397: Yeah, this is to get a specific instance of a source being saved (not just the latest time the obj was saved). The recency_index is to keep track of which instances are which

skyportal/handlers/api/news_feed.py
```diff
@@ -70,15 +70,27 @@ def fetch_newest(model):
...
-                'type': 'source',
-                'time': s.created_at,
-                'message': 'New source added',
-                'source_id': s.obj_id,
-            }
-            for s in sources
-        ]
+        news_feed_items = []
+        source_seen = set()
+        # Iterate in reverse so that we arrive at re-saved sources second
```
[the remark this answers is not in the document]
kmshin1397: The sources list coming in is in order of newest->oldest, so going backward means I see the older time an obj was saved first, so we can know that if this is not the first time we see a given obj_id then it's a case of a source being re-saved and modify the newsfeed entry accordingly

static/js/components/AddSourceGroup.jsx
kmshin1397: Good point. It might be a good idea to factor out the dialog form for selecting and submitting the new group. That's the part that's most duplicated, and I believe Dima uses the same logic in the Alerts page on Fritz for saving directly from the alert as well.  Should I do that as part of this PR?

acrellin: Let's defer this to another PR

acrellin (approved):
Looks good. Will merge when green

acrellin:
Travis failures look unrelated

Issue #578, opened 2020-07-17
Add quick admin/not admin toggle button next to existing users in group page


acrellin:
Currently users need to be deleted and then re-added as admin/non-admin in order to change their admin status

PR #1016, opened 2020-09-30
Add toggle admin status button in group users list

acrellin:
This PR introduces a button to change the admin status of group users (only displayed to users with the "Manage users" ACL). `GroupUserHandler.patch` is implemented, along with an associated action creator. An additional change: disable the delete user icon button in the case of a single remaining group admin instead of setting its display property to "none", which caused a jarring misalignment.

Closes https://github.com/skyportal/skyportal/issues/578 

Screen capture demo:
toggle_admin_status

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -422,6 +422,62 @@ class GroupUserHandler(BaseHandler):
         )
 
     @permissions(["Manage users"])
+    def patch(self, group_id, *ignored_args):
+        """
+        ---
+        description: Update a group user's admin status
+        parameters:
+          - in: path
+            name: group_id
+            required: true
+            schema:
+              type: integer
+        requestBody:
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  userID:
+                    type: integer
+                  admin:
+                    type: boolean
+                required:
+                  - username
+                  - admin
+        responses:
+          200:
+            content:
+              application/json:
+                schema: Success
+        """
+        data = self.get_json()
+        try:
+            group_id = int(group_id)
+        except ValueError:
+            return self.error("Invalid group ID")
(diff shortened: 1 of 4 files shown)


acrellin:
Travis is green; GA failures are unrelated

kmshin1397 (changes requested):
Mostly looks good to me. I had a couple suggestions on the front-end that I decided to go ahead and implement. I'll open a PR to your branch. Mainly, wanted to suggest:

- Moving the admin chip to the left to be next to the usernames to clearly distinguish the chip from the secondary action buttons (the right side of the list elements were looking a little crowded to me)
- The new button was overlapping with the admin chips on mobile screen sizes -> should condense the admin toggle and delete buttons into a single expand button that leads to a popover with the two action buttons.

stefanv:
https://material-ui.com/components/chips/

How about a chip with a cross instead?

acrellin:
> https://material-ui.com/components/chips/
> 
> How about a chip with a cross instead?

@stefanv how would that handle granting admin status to users that don't already have it? Sounds like that'd only working for revoking admin status (like we currently have on the manage users page).

kmshin1397 (approved):

acrellin:
Revoke admin access button is now disabled if the user in question is the sole group admin so that groups do not end up admin-less: 

disable_revoke_sole_admin

stefanv:
That's a good point, Ari; current design is fine then.  One small thing: the text---"remove admin" sounds a bit like the user will be removed.  Maybe "Remove admin privilege"?

acrellin:
> That's a good point, Ari; current design is fine then. One small thing: the text---"remove admin" sounds a bit like the user will be removed. Maybe "Remove admin privilege"? (Or "Remove admin role"?)

@stefanv good point. Done: 
toggle_admin_status

Issue #1018, opened 2020-09-30
Stream/group access when using bulk invite


dmitryduev:
When you do `email@institute.io,1 2,1,false` and program 1 has access to only stream 1 or 2, that would be the only stream the user will get access to. Could we please modify that so that users are first granted access to the specified streams and then added to groups (throwing an error if groups have access to streams that the user was not granted access to, and not adding more streams -- that should prevent accidents, I think)?

PR #1019, opened 2020-09-30
Grant user access to streams specified in invitation; check group streams against this list

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1018

### DIFF
--- a/skyportal/onboarding.py
+++ b/skyportal/onboarding.py
@@ -4,10 +4,8 @@ from .models import (
     User,
     Group,
     GroupUser,
-    GroupStream,
     Invitation,
     Role,
-    Stream,
     StreamUser,
 )
 from baselayer.app.env import load_env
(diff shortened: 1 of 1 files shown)


acrellin:
@dmitryduev I've tested this locally and is :heavy_check_mark:

skyportal/onboarding.py
```diff
@@ -124,16 +122,21 @@ def setup_invited_user_permissions(strategy, uid, details, user, *args, **kwargs
...
-        DBSession()
-        .query(Stream)
-        .join(GroupStream)
-        .filter(GroupStream.group_id.in_(group_ids))
-        .all()
-    )
-    stream_ids = [stream.id for stream in streams]
-
+    # Add user to specified streams
     for stream_id in stream_ids:
```
dmitryduev, line 140: Wait, so it should have been happening as intended, no? (i.e. add stream access first, then add to groups) I wonder why it was misbehaving?

dmitryduev (approved):

skyportal/onboarding.py
acrellin: @dmitryduev I needed to update where the streams list was coming from (now it comes from invitation.streams, whereas before it was coming from the set of streams associated with the specified groups)

PR #1025, opened 2020-10-02
Add a recent git changelog to the About page

profjsb:


skyportal/handlers/api/sysinfo.py
```diff
@@ -26,14 +28,50 @@ def get(self):
...
+                              type: array
+                              description: Recent git commit lines
+                          cosmology:
+                              type: string
+                              description: Details of the cosmology used here
+                          cosmoref:
+                              type: string
+                              description: Reference for the cosmology used.
         """
+        default_lines = 25
```
profjsb, line 44: I'm opting not to make this configurable. I think it's ok to be opinionated here.

skyportal/handlers/api/sysinfo.py
```diff
@@ -26,14 +28,50 @@ def get(self):
...
+            pr_number_start = result.find("(#")
+            if pr_number_start != -1:
+                pr_number_end = result.find(")", pr_number_start)
+                pr_str = result[(pr_number_start + 2):(pr_number_end)]
+                pr = pr_url + pr_str
+                result = result.replace(
+                    "(#" + pr_str + ")",
+                    "(<a target='_blank' rel='noopener noreferrer'"
+                    f" href='{pr}'>#{pr_str}</a>)",
+                )
```
profjsb, line 67: This makes the PR number linkable.

stefanv:
Please note that the git repository is not available in the built Docker container.

profjsb:
> Please note that the git repository is not available in the built Docker container.

Ug. I wish someone had told me that when I proposed this solution in the standup! What's the recourse?

stefanv:
> > Please note that the git repository is not available in the built Docker container.
> 
> Ug. I wish someone had told me that when I proposed this solution in the standup! What's the recourse?

Sorry, I had mentioned this to Dima when we spoke about this afterwards, but I didn't realise you were working on it.  I suggested we look at towncrier, which NumPy uses: https://github.com/numpy/numpy/tree/master/doc/release/upcoming_changes (see also the README in that directory).

Another option that may be easier for now would be to modify the `./fritz build` command to generate this list as a component, similar to what you've done already.

profjsb:
> > > Please note that the git repository is not available in the built Docker container.
> > 
> > 
> > Ug. I wish someone had told me that when I proposed this solution in the standup! What's the recourse?
> 
> Sorry, I had mentioned this to Dima when we spoke about this afterwards, but I didn't realise you were working on it. I suggested we look at towncrier, which NumPy uses: https://github.com/numpy/numpy/tree/master/doc/release/upcoming_changes (see also the README in that directory).
> 
> Another option that may be easier for now would be to modify the `./fritz build` command to generate this list as a component, similar to what you've done already.

I looked at towncrier -- it looks like it adds an extra burden on the developers to add more text during commits. Seems like overkill for what we're trying to solve during beta testing. The `build` command could send through a string with the log...

stefanv:
> > Another option that may be easier for now would be to modify the `./fritz build` command to generate this list as a component, similar to what you've done already.
> 
> I looked at towncrier -- it looks like it adds an extra burden on the developers to add more text during commits. Seems like overkill for what we're trying to solve during beta testing. The `build` command could send through a string with the log...

Yes, or populate a JSON / JS file that can be imported by that component.  But given that this will be built into Fritz only, perhaps a new page?

profjsb:
> > > Another option that may be easier for now would be to modify the `./fritz build` command to generate this list as a component, similar to what you've done already.
> > 
> > 
> > I looked at towncrier -- it looks like it adds an extra burden on the developers to add more text during commits. Seems like overkill for what we're trying to solve during beta testing. The `build` command could send through a string with the log...
> 
> Yes, or populate a JSON / JS file that can be imported by that component. But given that this will be built into Fritz only, perhaps a new page?

Ok @stefanv the implementation here works for a standalone SP instance but now also can work when an external file is written into `data/gitlog.txt`. This is now implemented in Fritz (https://github.com/fritz-marshal/fritz/pull/60).

acrellin (approved):
Looks great, thanks @profjsb

dannygoldstein (changes requested):
@profjsb, the log is nicely formatted and works great on my machine.

one big-picture issue: sometimes SP runs from outside a git repository. e.g., in Docker containers. cc @dmitryduev. The docker container on private does not contain the `.git` directory so I think this would break there, in production? 

Confirm that `.git` is not added to docker containers - see the .dockerignore file here:

https://github.com/skyportal/skyportal/blob/master/.dockerignore

Can you add a contingency where this is not shown when there is no .git directory? Something that checks the subprocess return code and logs the error if status code != 0 but does not halt the program?

dannygoldstein:
@profjsb I see the discussion above, but am still wondering - what about the case where there is no external log file and no .git directory?

profjsb:
> @profjsb I see the discussion above, but am still wondering - what about the case where there is no external log file and no .git directory?

In that case, a short message should show up about where to see PRs.

skyportal/handlers/api/sysinfo.py
```diff
@@ -26,14 +32,61 @@ def get(self):
...
+                    "log",
+                    "--pretty=format:'%C(cyan)[%ci]%Creset %s %C(auto)%h'",
+                ],
+                capture_output=True,
+                universal_newlines=True,
+            )
+            loginfo = p.stdout
+
+        raw_gitlog = loginfo.splitlines()[:default_log_lines]
+        gitlog = []
```
dannygoldstein, line 67: Looking at this block, `loginfo` starts as `None` and then if the external logfile exists it reads that and if not it runs `git` (which expects a `.git` directory to be there) By the time it gets to L66 then it expects `loginfo` to have been populated as a string. But if there is no `.git` and no external log then it will still be `None` and L66 will throw an exception.

dannygoldstein: I think it would be something like

```
AttributeError: 'NoneType' object has no attribute 'splitlines'
```

skyportal/handlers/api/sysinfo.py
```diff
@@ -26,14 +32,61 @@ def get(self):
...
+                              description: Recent git commit lines
+                          cosmology:
+                              type: string
+                              description: Details of the cosmology used here
+                          cosmoref:
+                              type: string
+                              description: Reference for the cosmology used.
         """
+        # if another build system has written a gitlog file, use it
+        loginfo = None
```
profjsb, line 49:
```suggestion
        loginfo = []
```

skyportal/handlers/api/sysinfo.py
```diff
@@ -26,14 +32,61 @@ def get(self):
...
+        if os.path.exists(gitlog_file):
+            with open(gitlog_file, "r") as spgl:
+                loginfo = spgl.read()
+        if not loginfo:
```
profjsb, line 53:
```suggestion
        if loginfo == "":
```

skyportal/handlers/api/sysinfo.py
profjsb: Fixed. Using `""` as the default. `"".splitlines` will produce `[]`, which will not error.

PR #1027, opened 2020-10-02
No longer deploy SkyPortal on merge to deploy branch

stefanv:
We have new Helm machinery, and that is being tested with Fritz


stefanv:
Failures unrelated.

PR #1031, opened 2020-10-02
Handle Group Bookkeeping for Posted Followup Request Data

dannygoldstein:
This PR alters the data model for posting ascii spectra to skyportal, using followup request ID to keep track of what groups the data should be shared with.

- Additional group ids to share a spectrum posted through the API with can be specified with the usual `group_ids` parameter.

- This PR also brings the test fixtures in line with the data model by adding each user to the sitewide group and to a single user group. 
 
- The followup request form has been augmented to control group access.


dannygoldstein:
cc @scizen9

static/js/components/GroupShareSelect.jsx
```diff
@@ -0,0 +1,129 @@
...
+import { makeStyles, useTheme } from "@material-ui/core/styles";
+import Select from "@material-ui/core/Select";
+import MenuItem from "@material-ui/core/MenuItem";
+import InputLabel from "@material-ui/core/InputLabel";
+import FormControl from "@material-ui/core/FormControl";
+import Chip from "@material-ui/core/Chip";
+import Input from "@material-ui/core/Input";
+
+const ITEM_HEIGHT = 48;
+const ITEM_PADDING_TOP = 8;
```
kmshin1397, line 12: Could we define these in terms of the theme instead of raw pixels? You could use the useTheme hook and move these props into the component below.

static/js/components/GroupShareSelect.jsx
```diff
@@ -0,0 +1,129 @@
...
+                    label={groupIDToName[value]}
+                    className={classes.chip}
+                  />
+                ))}
+              </div>
+            );
+          }
+          return (
+            <div className={classes.chips}>
+              <Chip
```
kmshin1397, line 89: I like the summary chip pattern!

static/js/components/GroupShareSelect.jsx
```diff
@@ -0,0 +1,129 @@
...
+      name: PropTypes.string,
+    })
+  ).isRequired,
+  groupIDs: PropTypes.arrayOf(PropTypes.number).isRequired,
+  setGroupIDs: PropTypes.func.isRequired,
+  maxGroups: PropTypes.number,
+};
+
+GroupShareSelect.defaultProps = {
+  maxGroups: 2,
```
kmshin1397, line 126: A default of 2 here seems a little low for me. There's a lot more space to display more groups unless you're on mobile. The other inputs in this form tend to extend across the width of the accordion, so limiting this to max 2 group names feels bad to me. Suggested change would be to set this to something like 3-4, but passing in maxGroups: 2 from the parent component if you detect a mobile screen size per media queries.

kmshin1397 (commented):
Mostly LGTM! A couple of minor comments. I also went ahead and resolved a merge conflict in schema.py for you.

static/js/components/GroupShareSelect.jsx
dannygoldstein: thanks

static/js/components/GroupShareSelect.jsx
dannygoldstein: set to 3

static/js/components/GroupShareSelect.jsx
dannygoldstein: these were unnecesssary and I just removed them

kmshin1397 (approved):
Nice! A bunch of tests failing but I don't think any of them are related. Maybe try merging in the latest master and see if that fixes any of them?

dannygoldstein:
@kmshin1397 I think they are related ... investigating now.

dannygoldstein:
Failures are now unrelated (same failures as on master)

dannygoldstein:
@dmitryduev This is ready for you to look at now. Can you please review. Thanks for waiting ...

dannygoldstein:
LMK when you both approve, then will add a db migration script. @kmshin1397 Requesting your re-review as a lot of time has passed since the last review. There haven't been any substantive changes, just keeping stuff in sync with master

kmshin1397 (approved):

skyportal/tests/frontend/test_followup_requests.py
```diff
@@ -185,3 +188,111 @@ def test_delete_followup_request(
...
+    )
+    driver.wait_for_xpath(
+        f'''//table[contains(@data-testid, "followupRequestTable")]//td[contains(., "1")]'''
+    )
+    driver.wait_for_xpath(
+        f'''//table[contains(@data-testid, "followupRequestTable")]//td[contains(., "submitted")]'''
+    )
+
+    filename = glob.glob(
+        f'{os.path.dirname(__file__)}/../data/ZTF20abwdwoa_20200902_P60_v1.ascii'
```
dmitryduev, line 264: Does this have to be hardcoded?

static/js/components/FollowupRequestForm.jsx
```diff
@@ -27,21 +32,32 @@ const FollowupRequestForm = ({
...
+    setSelectedGroupIds([allocationList[0]?.group_id]);
+  }, [setSelectedAllocationId, allocationList, setSelectedGroupIds]);
 
   // need to check both of these conditions as selectedAllocationId is
   // initialized to be null and useEffect is not called on the first
   // render to update it, so it can be null even if allocationList is not
   // empty.
-  if (allocationList.length === 0 || !selectedAllocationId) {
+  if (
+    allocationList.length === 0 ||
```
dmitryduev, line 48: `allocationList?.length`?

static/js/components/FollowupRequestForm.jsx
```diff
@@ -27,21 +32,32 @@ const FollowupRequestForm = ({
...
+  ) {
     return <h3>No robotic instruments available...</h3>;
   }
 
-  if (!allGroups || allGroups.length === 0 || telescopeList.length === 0) {
+  if (
+    !allGroups ||
+    allGroups.length === 0 ||
+    telescopeList.length === 0 ||
+    instrumentList.length === 0
```
dmitryduev, line 59: Optional chaining for all these?

dmitryduev (approved):
LGTM! It touches the data model, so needs a migration script -- could you please add that?

skyportal/tests/frontend/test_followup_requests.py
dannygoldstein: Yeah, the data file is a part of the repo.

static/js/components/FollowupRequestForm.jsx
dannygoldstein: No need, `allocationList` is initialized in redux to be `[]`.

static/js/components/FollowupRequestForm.jsx
dannygoldstein: `telescopeList` and `instrumentList` don't need it, they are both initialized to be `[]`.  `allGroups` I have modified.

dannygoldstein: actually need to keep it the way it is now. We want to check if `allGroups` is null or has length zero if nonnull, and the chaining wouldn't handle that. `allGroups.length` is only accessed if `!!allGroups` (previous line), so no need for chaining.

dannygoldstein:
migration script added. this is ready to go in.

PR #1044, opened 2020-10-02
Comment groups permissions updates & bug fixes

acrellin:
This PR fixes a bug where a user with System admin ACL could view a source saved to group(s) they're not part of, but could not use the front-end form to comment, as the front-end form only selects the users groups they're part of (here changed to all groups accessible to user).

Also, malformed comment attachments are better handled (used to fail silently, now returns informative error message)

### DIFF
--- a/skyportal/handlers/api/comment.py
+++ b/skyportal/handlers/api/comment.py
@@ -106,32 +106,24 @@ class CommentHandler(BaseHandler):
             )
 
         # Only post to groups source/candidate is actually associated with
-        if DBSession().query(Candidate).filter(Candidate.obj_id == obj_id).all():
-            candidate_group_ids = [
-                f.group_id
-                for f in (
-                    DBSession()
-                    .query(Filter)
-                    .filter(Filter.id.in_(user_accessible_filter_ids))
-                    .filter(
-                        Filter.id.in_(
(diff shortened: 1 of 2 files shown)


acrellin:
Failures unrelated (sole failure on Travis is `FAILED skyportal/tests/frontend/test_recent_sources.py::test_recently_saved_candidate`)

skyportal/handlers/api/comment.py
```diff
@@ -106,42 +106,41 @@ def post(self):
...
+                .all()
+            )
+        ]
+        source_group_ids = [
+            source.group_id
+            for source in DBSession()
+            .query(Source)
+            .filter(Source.obj_id == obj_id)
+            .all()
+        ]
```
acrellin, line 126: The logic hasn't changed here, it's just been simplified & cleaned up

kmshin1397 (approved):

acrellin:
@dannygoldstein @dmitryduev whichever of you has a moment to review first, go ahead and review if happy (has 1 approval already)

dmitryduev (approved):
LGTM, thanks for the fix!

dmitryduev:
just in case, CI failures seem unrelated

PR #1046, opened 2020-10-02
Truncate long file attachment names for comments

kmshin1397:
Screenshot 2020-10-02 122542

### DIFF
--- a/static/js/components/CommentList.jsx
+++ b/static/js/components/CommentList.jsx
@@ -19,6 +19,19 @@ import UserAvatar from "./UserAvatar";
 dayjs.extend(relativeTime);
 dayjs.extend(utc);
 
+export const shortenFilename = (filename) => {
+  if (filename.length <= 15) {
+    return filename;
+  }
+  const extensionLength = filename.split(".", 2)[1].length;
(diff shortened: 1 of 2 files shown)


static/js/components/CommentList.jsx
```diff
@@ -19,6 +19,19 @@ import UserAvatar from "./UserAvatar";
 dayjs.extend(relativeTime);
 dayjs.extend(utc);
 
+export const shortenFilename = (filename) => {
+  if (filename.length <= 15) {
+    return filename;
+  }
+  const extensionLength = filename.split(".", 2)[1].length;
```
dmitryduev, line 26: What if there are no dots in the filename for some reason?

dmitryduev (changes requested):
Looks great, one small thing probably needs addressing.

acrellin (approved):
Thanks for the quick fix!

static/js/components/CommentList.jsx
kmshin1397: Good point. Case implemented

dmitryduev (approved):
LGTM!

PR #1053, opened 2020-10-05
Improved Source/Candidate classification and position view

### LINKED ISSUE #1052: Show small/large cap Classifications on the Source page.
Request from Dan Perley.

profjsb:
- Shows the classification capitalization as it is in the taxonomy (before it was all caps)
- puts the RA/DEC sexigesimal position out in front, in big bold font.
image

Closes #1052

### DIFF
--- a/static/js/components/Candidate.jsx
+++ b/static/js/components/Candidate.jsx
@@ -32,6 +32,10 @@ export const useStyles = makeStyles(() => ({
   accordionItem: {
     maxWidth: "900px",
   },
+  position: {
+    fontWeight: "bold",
+    fontSize: "110%",
+  },
 }));
 
 const Candidate = ({ route }) => {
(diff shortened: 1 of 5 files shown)


static/js/components/ShowClassification.jsx
```diff
@@ -1,7 +1,16 @@
 import React from "react";
 import PropTypes from "prop-types";
-import Button from "@material-ui/core/Button";
 import Tooltip from "@material-ui/core/Tooltip";
+import Chip from "@material-ui/core/Chip";
+
+import { makeStyles } from "@material-ui/core/styles";
+
+// Export to allow Candidate.jsx to use styles
+export const useSourceStyles = makeStyles((theme) => ({
```
kmshin1397, line 9:
```suggestion
export const useStyles = makeStyles((theme) => ({
```
This seems to be copy pasted from Source styling, but we don't actually need to export this makeStyles() instance outside of this component.

static/js/components/ShowClassification.jsx
```diff
@@ -12,8 +21,9 @@ const groupBy = (array, key) =>
   }, {});
 
 function ShowClassification({ classifications, taxonomyList }) {
-  // Here we compute the most recent non-zero probability class for each taxonomy
+  const classes = useSourceStyles();
```
kmshin1397, line 24:
```suggestion
  const classes = useStyles();
```

static/js/components/Candidate.jsx
```diff
@@ -32,6 +32,10 @@ export const useStyles = makeStyles(() => ({
   accordionItem: {
     maxWidth: "900px",
   },
+  position: {
+    fontWeight: "bold",
+    fontSize: "110%",
+  },
```
kmshin1397, line 38:
```suggestion
```
This is actually not necessary as the `classes.position` class is actually imported from SourceMobile.jsx. Putting rules for `position` here will assign it to `candidateStyles.position` (candidate page specific styling not imported from the source page), but that's not necessary here.****

kmshin1397 (changes requested):
Functionally looks good - a few minor changes for semantics

kmshin1397 (approved):

PR #1057, opened 2020-10-06
Allow configuring multiple logos in header bar

kmshin1397:
Will put up corresponding PR to Fritz to add the GROWTH logo as in sample image below:
Screenshot 2020-10-06 134117

Also a small spacing improvement on mobile for header bar (removed a margin if there's less space available)

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -1,6 +1,8 @@
 app:
   title: SkyPortal
-  logo: /static/images/skyportal_logo.png
+  logos: # Logos to place in the top-left of the header bar (zero or more)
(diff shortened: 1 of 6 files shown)


stefanv:
@kmshin1397 Quick check: how does this interact with the logo rotation on load?

kmshin1397:
> @kmshin1397 Quick check: how does this interact with the logo rotation on load?

Yes, I slightly refactored the existing Logo.jsx component to take different image src paths instead of just the Skyportal one. So all logos are hooked up to the same store variable with the rotation styling.

acrellin (approved):
Looks good, thanks @kmshin1397 !

PR #1060, opened 2020-10-06
add a new ACL for managing allocations - manage allocations

dannygoldstein:
This PR creates a new ACL - "Manage allocations", and replaces the requirement that only a System admin manage allocations with anyone who has this ACL. This will allow designated users to manage allocations (e.g., @rswalters).


dannygoldstein:
@dmitryduev failure on GA is unrelated and fixed by #1061

dmitryduev (approved):
👍

PR #1061, opened 2020-10-06
Downgrade redoc-cli from 0.9.12 to 0.9.8

acrellin:


dannygoldstein:
Merging as this doesn't affect the docs and fixes the regression on master.

PR #1062, opened 2020-10-06
Decouple thumbnails from photometry; add PS1 thumbnails; show more thumbnails on source/scanning pages

### LINKED ISSUE #946: Show PS1 and/or SDSS cutouts directly on source and scanning pages
It can be of great help to see SDSS or PS1 to visually check for a host galaxy and/or stellar counterparts.
On the scanning page this could be optional through a checkbox in case there is a general slowdown. However, it is critical that other surveys / catalogs can be easily accessed when scanning even before a source has been saved.

acrellin:
Closes #946 

scanning_page_thumbnails_5

source_page_thumbnails

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -11,7 +11,6 @@ from ...models import (
     DBSession,
     Obj,
     Candidate,
-    Thumbnail,
     Photometry,
     Instrument,
     Source,
@@ -171,11 +170,11 @@ class CandidateHandler(BaseHandler):
                 obj_id,
                 self.current_user,
                 options=[
+                    joinedload(Candidate.obj).joinedload(Obj.thumbnails),
                     joinedload(Candidate.obj)
-                    .joinedload(Obj.thumbnails)
-                    .joinedload(Thumbnail.photometry)
+                    .joinedload(Obj.photometry)
                     .joinedload(Photometry.instrument)
-                    .joinedload(Instrument.telescope)
+                    .joinedload(Instrument.telescope),
                 ],
             )
             if c is None:
@@ -261,8 +260,8 @@ class CandidateHandler(BaseHandler):
         q = (
             Obj.query.options(
                 [
-                    joinedload(Obj.thumbnails)
-                    .joinedload(Thumbnail.photometry)
+                    joinedload(Obj.thumbnails),
+                    joinedload(Obj.photometry)
                     .joinedload(Photometry.instrument)
                     .joinedload(Instrument.telescope),
                 ]
@@ -400,6 +399,7 @@ class CandidateHandler(BaseHandler):
                               description: New candidate ID
         """
         data = self.get_json()
+        obj_already_exists = Obj.query.get(data["id"]) is not None
         schema = Obj.__schema__()
 
         passing_alert_id = data.pop("passing_alert_id", None)
@@ -447,6 +447,8 @@ class CandidateHandler(BaseHandler):
             ]
         )
         DBSession().commit()
+        if not obj_already_exists:
+            obj.add_linked_thumbnails()
 
         return self.success(data={"id": obj.id})
 
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -256,10 +256,9 @@ class SourceHandler(BaseHandler):
             if s is None:
                 return self.error("Invalid source ID.")
             if "ps1" not in [thumb.type for thumb in s.thumbnails]:
-                create_thumbnail_link = functools.partial(
-                    add_ps1_thumbnail_and_push_ws_msg, s, self,
+                IOLoop.current().add_callback(
+                    lambda: add_ps1_thumbnail_and_push_ws_msg(s, self)
                 )
-                IOLoop.current().add_callback(create_thumbnail_link)
             comments = s.get_comments_owned_by(self.current_user)
             source_info = s.to_dict()
             source_info["comments"] = sorted(
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -23,7 +23,6 @@ from ...models import (
     Photometry,
     Obj,
     Source,
-    Thumbnail,
     Token,
     Group,
     FollowupRequest,
@@ -236,9 +235,9 @@ class SourceHandler(BaseHandler):
                     .joinedload(ClassicalAssignment.run)
                     .joinedload(ObservingRun.instrument)
                     .joinedload(Instrument.telescope),
+                    joinedload(Source.obj).joinedload(Obj.thumbnails),
                     joinedload(Source.obj)
-                    .joinedload(Obj.thumbnails)
-                    .joinedload(Thumbnail.photometry)
+                    .joinedload(Obj.photometry)
                     .joinedload(Photometry.instrument)
                     .joinedload(Instrument.telescope),
                     joinedload(Source.obj)
@@ -290,10 +289,12 @@ class SourceHandler(BaseHandler):
                 )  # only give sources the user has access to
             )
             .options(
-                joinedload(Obj.thumbnails)
-                .joinedload(Thumbnail.photometry)
-                .joinedload(Photometry.instrument)
-                .joinedload(Instrument.telescope)
+                [
+                    joinedload(Obj.thumbnails),
+                    joinedload(Obj.photometry)
+                    .joinedload(Photometry.instrument)
+                    .joinedload(Instrument.telescope),
+                ]
             )
         )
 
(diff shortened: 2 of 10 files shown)


acrellin:
PS1 thumbnail is now added in the background after the initial source fetch has returned, and a REFRESH_SOURCE WS message is pushed when complete. Demo screen capture:
async_ps1_gen

stefanv:
@acrellin Should we have a placeholder in places to indicate to the user that it is being fetched, so it does not just flash into place?

skyportal/handlers/api/source.py
```diff
@@ -249,6 +255,11 @@ def get(self, obj_id=None):
             )
             if s is None:
                 return self.error("Invalid source ID.")
+            if "ps1" not in [thumb.type for thumb in s.thumbnails]:
+                create_thumbnail_link = functools.partial(
+                    add_ps1_thumbnail_and_push_ws_msg, s, self,
+                )
+                IOLoop.current().add_callback(create_thumbnail_link)
```
stefanv, line 262: Maybe `functools.partial` is overkill here.  How about:

```
IOLoop.current().add_callback(
    lambda: add_ps1_thumbnail_and_push_ws_msg(s, self)
)
```

I was a bit concerned about using `self` (the handler) this way, because they handler is typically destroyed after each request.  But, I *think* you're okay since a reference to self, like you are installing here, might keep the handler alive.  I would test this explicitly by adding a `time.sleep(...)` to `create_thumbnail_link_and_push_ws_msg` (which, btw, I think is fine to rename to `create_thumbnail_link` if you use a lambda); if the sleep lasts a long time, it would give time for the handler to finish, and we would see if it still available when the WS message needs to be pushed.

skyportal/models.py
```diff
@@ -489,6 +489,28 @@ def desi_dr8_url(self):
...
+        """
+        try:
+            ps_query_url = (
+                f"http://ps1images.stsci.edu/cgi-bin/ps1cutouts"
+                f"?pos={self.ra}+{self.dec}&filter=color&filter=g"
+                f"&filter=r&filter=i&filetypes=stack&size=250"
+            )
+            response = requests.get(ps_query_url)
+            match = re.search(
+                'src="//ps1images.stsci.edu.*?"', response.content.decode()
```
stefanv, line 508: I'd probably have opted for BeautifulSoup4, to also avoid the replace below which feels slightly brittle, but I'll leave it up to you.

skyportal/handlers/api/candidate.py
```diff
@@ -447,6 +447,8 @@ def post(self):
             ]
         )
         DBSession().commit()
+        if not obj_already_exists:
+            obj.add_linked_thumbnails()
```
stefanv, line 451: I know this is basically the logic we used to have, but I'm curious: what do we do if this step fails?

stefanv (commented):
Good work, Ari!

I have a few questions, but the implementation is clean—I like it.

acrellin:
@stefanv placeholder added:

async_ps1_gen_w_placeholder2

skyportal/handlers/api/source.py
acrellin: @stefanv confirmed, the handler stays alive even with a long sleep

acrellin: @stefanv regarding functools vs lambda, I'm not sure I see it as a simplification, but I've implemented your suggestion (now using a lambda function)

skyportal/models.py
acrellin: I just resurrected whatever Brett wrote back in the day -- this is his implementation

skyportal/handlers/api/candidate.py
acrellin: Well, currently it returns with an error. What do you propose we should do?

dannygoldstein: @stefanv I dont think this should be too volatile - at this point basically it is just adding a single row to the Thumbs table in the DB with a string link to the external thumbs.

skyportal/handlers/api/source.py
```diff
@@ -249,6 +255,10 @@ def get(self, obj_id=None):
             )
             if s is None:
                 return self.error("Invalid source ID.")
+            if "ps1" not in [thumb.type for thumb in s.thumbnails]:
+                IOLoop.current().add_callback(
+                    lambda: add_ps1_thumbnail_and_push_ws_msg(s, self)
```
dannygoldstein, line 260: What is the scenario in which this will be triggered?

skyportal/models.py
```diff
@@ -463,15 +462,16 @@ def last_detected(cls):
...
+        sdss_thumb = Thumbnail(obj=self, public_url=self.sdss_url, type='sdss')
+        dr8_thumb = Thumbnail(obj=self, public_url=self.desi_dr8_url, type='dr8')
         DBSession().add_all([sdss_thumb, dr8_thumb])
         DBSession().commit()
 
+    def add_ps1_thumbnail(self):
+        ps1_thumb = Thumbnail(obj=self, public_url=self.panstarrs_url, type="ps1")
+        DBSession().add(ps1_thumb)
+        DBSession().commit()
+
```
dannygoldstein, line 474: Why have these separate?

skyportal/models.py
```diff
@@ -489,6 +489,28 @@ def desi_dr8_url(self):
             f"&dec={self.dec}&size=200&layer=dr8&pixscale=0.262&bands=grz"
         )
 
+    @property
+    def panstarrs_url(self):
+        """Construct URL for public PanSTARRS-1 (PS1) cutout.
+
+        The cutout service doesn't allow directly querying for an image; the
+        best we can do is request a page that contains a link to the image we
+        want (in this case a combination of the green/blue/red filters).
```
dannygoldstein, line 498:
```suggestion
        want (in this case a combination of the g/r/i filters).
```

skyportal/models.py
```diff
@@ -489,6 +489,28 @@ def desi_dr8_url(self):
...
+            ps_query_url = (
+                f"http://ps1images.stsci.edu/cgi-bin/ps1cutouts"
+                f"?pos={self.ra}+{self.dec}&filter=color&filter=g"
+                f"&filter=r&filter=i&filetypes=stack&size=250"
+            )
+            response = requests.get(ps_query_url)
+            match = re.search(
+                'src="//ps1images.stsci.edu.*?"', response.content.decode()
+            )
+            return match.group().replace('src="', 'http:').replace('"', '')
```
dannygoldstein, line 510: Nice.

dannygoldstein (commented):
Good stuff @acrellin, I like this data model way more. 

One comment is that the ObservingRun handler should still return the thumbnails as these are used on the observingRun page.

acrellin:
> Good stuff @acrellin, I like this data model way more.
> 
> One comment is that the ObservingRun handler should still return the thumbnails as these are used on the observingRun page.

@dannygoldstein that wasn't removed, just needed to be updated 
https://github.com/skyportal/skyportal/pull/1062/files#diff-7f11563469135cb2cdd534eff9841428R104

skyportal/handlers/api/candidate.py
stefanv: What I meant, and maybe this is already handled, is that we should attempt again in the future if fetching the URL fails.  So, maybe do this call every time an obj is accessed and the URL is not filled out.

stefanv: Looks like this is the logic below already, maybe.

stefanv (approved):

skyportal/models.py
acrellin: Because PS1 is the only one that makes an external request to generate the link - the other two just immediately return a string, so are super quick and can be added as soon as the obj is created without much of a performance hit

skyportal/handlers/api/source.py
acrellin: If the PS1 thumbnail doesn't exist yet when a single source is fetched, we kick off the process that generates it

skyportal/handlers/api/candidate.py
acrellin: @stefanv yes, exactly. The line above adds the two thumbnail links that we already know without making an external request. PS1 requires making the external request - if that fails, it'll try again the next time the source is accessed

dannygoldstein (approved):

dannygoldstein (commented):
@acrellin It looks like there are some consistent new failures on the sources page. Could these be related to the new thumbnails?

acrellin:
> @acrellin It looks like there are some consistent new failures on the sources page. Could these be related to the new thumbnails?

@dannygoldstein hopefully fixed in https://github.com/skyportal/skyportal/pull/1062/commits/da2449cfa0f45ef1aee355fa4110036daf81c066

acrellin:
@dannygoldstein the failure that was related to this PR is passing now: https://github.com/skyportal/skyportal/pull/1062/checks?check_run_id=1228249322#step:13:417

stefanv:
:rocket: Thanks, Ari!

PR #1063, opened 2020-10-06
Fix peer dependencies for redoc and react-dropdown-aria

stefanv:


stefanv:
Not trying to upgrade any of these, given that master is fixed for now.  Will try again once redoc makes a release.

PR #1064, opened 2020-10-07
Fix photometry plots

dmitryduev:
Fix photometry plots if there are either no detections or no non-detections


stefanv (approved):

dannygoldstein (approved):

dannygoldstein:
Thanks for picking this one up @dmitryduev

PR #1065, opened 2020-10-07
Improve scanning page + front-end for annotations refactor

kmshin1397:
#### Changelist ####
- Annotations are now displayed separately on Source and Candidate pages as a MUI DataTable
- Scanning page:
  - The MUI DataTable is now set to server-side loading mode, and the MuiDataTable component is hooked up to the proper candidates list fetching API calls
  - The MUI DataTable's built-in pagination controls are enabled (also using server-side loading) -> **Option to choose number of rows per page enabled**
  - Removed the existing pagination/jump-to-page parts of the input form at the top of the page since these are now available directly on the table 
  - Annotations are grouped by origin and displayed in nested lists
  - Any annotation key with a number-type value can be clicked on to select it -> This will display in the Info column that annotation key/value (if present) for all candidates
  - If an annotation key is selected, the sort tool button can be clicked to sort on that origin[key] value. Re-clicking will sort in the other direction
- Backend:
  - Add annotations and sorting logic/query options to CandidatesHandler

#### Screenshots ####
- Source/Candidate page Annotations table (Desktop)
Screenshot 2020-10-06 162913
- Source/Candidate page Annotations table (Mobile)
Screenshot 2020-10-06 162844
- Pagination
paging
- Sorting
sorting

#### Related Issues ####
- #1010 
- #961 (partially)
- #947
- #864 (Not directly addressed, but should go away since annotations shouldn't go to Comments anymore for future data)

#### Work Left for Future PRs ####
- Filtering on annotations
- Other small improvements, i.e. displaying classifications in Info if present

PR #1067, opened 2020-10-07
Display annotations as MUI data table on source/candidate page

kmshin1397:
Breaks out the Source/Candidate page updates from the rest of #1065 for easier review.

Desktop:
Screenshot 2020-10-06 162913

Mobile:
Screenshot 2020-10-06 162844


dannygoldstein:
merge conflicts

dannygoldstein (commented):
- [ ] Some annotation values are cut off, we should always show the full annotation value. 
Screen Shot 2020-10-07 at 12 42 29 PM'

- [ ] Show who posted the annotation, and when, in the table.

- [ ] I think origin is basically degenerate with `group_id`. I would be in favor of deprecating origin and just having it show the `Group.name` corresponding to `group_id`. (This also solves the problem of acceptable origin names described by @guynir42 ).

kmshin1397:
> * [ ]  I think origin is basically degenerate with `group_id`. I would be in favor of deprecating origin and just having it show the `Group.name` corresponding to `group_id`. (This also solves the problem of acceptable origin names described by @guynir42 ).

@dannygoldstein Annotations allow multiple group_ids to be associated with a single annotation - what would I should in that case?

static/js/components/AnnotationsTable.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+import {
+  makeStyles,
+  createMuiTheme,
+  MuiThemeProvider,
+  useTheme,
+} from "@material-ui/core/styles";
+import MUIDataTable from "mui-datatables";
+
+import { getAnnotationValueString } from "./CandidateAnnotationsList";
```
acrellin, line 16: It'd be helpful to capture in the filenames here and with CandidateAnnotationsList.jsx exactly where they're used, e.g. changing this one to `CandidatePageAnnotations` and the other to `ScanningPageCandidateAnnotations`. It's confusing to have all of these very similarly-named files.

static/js/components/CandidateAnnotationsList.jsx
```diff
@@ -0,0 +1,157 @@
+import React, { useState } from "react";
```
acrellin, line 1: Again, I'm confused by the filename here and how to distinguish this from the other new file in this PR

static/js/components/CandidateAnnotationsList.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+// This component will be used to display annotations in the scanning page.
+// Included in this PR only for the getAnnotationValueString() function above.
+// Can ignore the rest of this component during review of this PR for now.
+const CandidateAnnotationsList = ({ annotations }) => {
+  const classes = useStyles();
+
+  const dispatch = useDispatch();
+
+  const initState = {};
```
acrellin, line 62:
```suggestion
  const initialState = {};
```
just for a bit more readability

static/js/components/CandidateAnnotationsList.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const selectedAnnotationItem = useSelector(
+    (state) => state.candidates.selectedAnnotationItem
+  );
+
+  const handleClick = (origin) => {
+    setopenedOrigins({ ...openedOrigins, [origin]: !openedOrigins[origin] });
+  };
+
+  const handleItemSelect = (origin, key) => {
```
acrellin, line 76: What's an item?  An annotation, or a specific annotation data key? Would be helpful to make the name more explicit

static/js/components/CandidateAnnotationsList.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+  return (
+    <List
+      component="nav"
+      aria-labelledby="nested-list-subheader"
+      className={classes.root}
+      dense
+    >
+      {annotations.map((annotation) => (
+        <div key={`annotation_${annotation.origin}`}>
```
acrellin, line 94: Isn't it possible for many annotations to have the same origin, or are all annotations with the same origin being combined somewhere?

acrellin (commented):
Looks great overall, just a few comments below that I think will help readability/maintainability

static/js/components/AnnotationsTable.jsx
kmshin1397: Good point, I'll do that.

static/js/components/CandidateAnnotationsList.jsx
kmshin1397: Annotations can have the same origin across candidates, but not within the same candidate, Since this list is used to display the annotations of one specific candidate, there is no issue here.

static/js/components/CandidateAnnotationsList.jsx
kmshin1397: Annotation data key. You're right, this could be more specific

dmitryduev:
@dannygoldstein @kmshin1397 using `group_id` would work, but then if a group has multiple filters, and all make annotations, they will be dumped in a single bucket. Also, auto-annotations made by Kowalski and AMPEL would be dumped into a single bucket, too.

kmshin1397:
> @dannygoldstein @kmshin1397 using `group_id` would work, but then if a group has multiple filters, and all make annotations, they will be dumped into a single bucket. Also, auto-annotations made by Kowalski and AMPEL would be dumped into a single bucket, too.

Related to this, if we dump annotations from multiple filters into a single group bucket, I'd imagine we could run into conflicting values for duplicated keys, like if each filter provides a separate "score" field for example. I think as long as users are diligent about descriptive origin names origin makes more sense. I should probably update the documentation for the origin field to be more descriptive though.

dmitryduev (approved):
LGREATTM!

acrellin (changes requested):
Strong request that we try to parameterize a more general component that can be used in both (or more) places. That's (part of) the beauty of React -- reusable, composable components. We should try to minimize code duplication whenever possible, even if it requires refactoring something to make it more generalized.

kmshin1397:
> Strong request that we try to parameterize a more general component that can be used in both (or more) places. That's (part of) the beauty of React -- reusable, composable components. We should try to minimize code duplication whenever possible, even if it requires refactoring something to make it more generalized.

@acrellin Good catch, I think I removed the duplication you were referring to. I think git got confused because I had one file renamed here compared to the other PR branch that got merged, so it made two copies of the same file. 

If that wasn't what you were talking about, and instead you were talking about generalizing the display of annotations between the Source/Candidate page and the scanning page, I'm not sure that that's the best direction to go. I'm all for generalizing where we can, but the two components (ScanningPageCandidateAnnotations vs ObjPageAnnotations) there are fundamentally different. The former is a nested List and must hook up to the larger CandidateList MUIDataTable and the server-sider candidates fetching. The latter is a simpler, isolated MUIDataTable in itself with all the built-in MUIDataTable searching/filtering/etc enabled.

acrellin:
> > Strong request that we try to parameterize a more general component that can be used in both (or more) places. That's (part of) the beauty of React -- reusable, composable components. We should try to minimize code duplication whenever possible, even if it requires refactoring something to make it more generalized.
> 
> @acrellin Good catch, I think I removed the duplication you were referring to. I think git got confused because I had one file renamed here compared to the other PR branch that got merged, so it made two copies of the same file.
> 
> If that wasn't what you were talking about, and instead you were talking about generalizing the display of annotations between the Source/Candidate page and the scanning page, I'm not sure that that's the best direction to go. I'm all for generalizing where we can, but the two components (ScanningPageCandidateAnnotations vs ObjPageAnnotations) there are fundamentally different. The former is a nested List and must hook up to the larger CandidateList MUIDataTable and the server-sider candidates fetching. The latter is a simpler, isolated MUIDataTable in itself with all the built-in MUIDataTable searching/filtering/etc enabled.

The diff looks better now, and yes, I can see that they're different enough to not be combined into a more general component.

acrellin (approved):

Issue #1010, opened 2020-09-29
Allow users to adjust number of objects on the scanning page


dannygoldstein:
From Jesper: 

I want to be able to select how many objects to display. Growth marshall allows some options, easiest could even be to pick your own number.
Skyportal only allows 25 (and actually only display 10) which is very unpractical for a scanner like me (too much reloading time).

dannygoldstein:
Fixed by #1068

Issue #947, opened 2020-09-18
Sorting on scanning page


cfremling:
I would like to request the ability to sort the scanning page, based on computations made by the filters. For example, if we compute an age based on the first historical detection, it would be great if we could sort the entire scanning page to put young/new canididates first.

AndyTza:
I also think it would be particularly useful when the user selects the parameter they want to sort by for it to be displayed right next to the cutout images (i.e see image below)

<img width="1434" alt="Screen Shot 2020-09-25 at 12 22 15 PM" src="https://user-images.githubusercontent.com/22720676/94291811-14268700-ff2a-11ea-8914-e35b2e160eba.png">

stefanv:
Related: https://github.com/skyportal/skyportal/issues/961

dannygoldstein:
@cfremling @AndyTza: Update, we have a good model for this now, @guynir42  is leading up the backend development (PR should be in today) and @kmshin1397 is going to be implementing this on the frontend.

dannygoldstein:
Fixed by #1068

Issue #864, opened 2020-09-06
Don’t show autoannotations in NewsFeed


profjsb:
Annotations swamping the NewsFeed. Don't show them.

dmitryduev:
The simplest way to do that would be to filter out comments posted by provisioned_admin.

dannygoldstein:
Fixed by #1068 and #979

PR #992, opened 2020-09-26
alphabetize and fix links in surveylinklist

dannygoldstein:
[ch1289]

This PR alphabetizes the `SurveyLinkList` as requested by @steveschulze

PR #1068, opened 2020-10-07
Improve scanning page

kmshin1397:
Revised version of #1065 to factor out just the scanning page changes. See #1067 for the other half of what was #1065. 

#### Changelist ####
- Scanning page:
  - The MUI DataTable is now set to server-side loading mode, and the MuiDataTable component is hooked up to the proper candidates list fetching API calls
  - The MUI DataTable's built-in pagination controls are enabled (also using server-side loading) -> **Option to choose number of rows per page enabled**
  - Removed the existing pagination/jump-to-page parts of the input form at the top of the page since these are now available directly on the table 
  - Annotations are grouped by origin and displayed in nested lists
  - Any annotation key with a number-type value can be clicked on to select it -> This will display in the Info column that annotation key/value (if present) for all candidates
  - If an annotation key is selected, the sort tool button can be clicked to sort on that origin[key] value. Re-clicking will sort in the other direction
- Backend:
  - Add annotations and sorting logic/query options to CandidatesHandler

#### Screenshots ####
- Pagination
paging
- Sorting
sorting

#### Related Issues ####
- #1010 
- #961 (partially)
- #947
- #864 (Not directly addressed, but should go away since annotations shouldn't go to Comments anymore for future data)

#### Work Left for Future PRs ####
- Filtering on annotations
- Other small improvements, i.e. displaying classifications in Info if present


dannygoldstein (commented):
- [x] Allow sorting on any field, not just numeric fields.

- [x] Long auto annotations are cut off, make sure to show the full annotation (wrapping text if necessary).
Screen Shot 2020-10-07 at 12 47 03 PM

- [ ] Nested fields show up as [Object object], please coerce these to strings for display purposes.

skyportal/handlers/api/candidate.py
```diff
@@ -521,10 +583,8 @@ def grab_query_results_page(q, total_matches, page, n_items_per_page, items_name
         or (info["totalMatches"] == 0 and page != 1)
     ):
         raise ValueError("Page number out of range.")
-    info[items_name] = (
-        q.limit(n_items_per_page).offset((page - 1) * n_items_per_page).all()
-    )
-
+    s = slice((page - 1) * n_items_per_page, page * n_items_per_page)
```
acrellin, line 586: There could potentially be a very large number of candidates in the DB -- grabbing them all without a `limit` seems like a bad idea -- this could take a very long time

skyportal/handlers/api/candidate.py
```diff
@@ -500,11 +558,15 @@ def patch(self, obj_id):
 
 
 def grab_query_results_page(q, total_matches, page, n_items_per_page, items_name):
+    # The query can return multiple rows for candidates with multiple annotations
+    # Getting the query results and doing the limit/offset on the Python side
+    # simplifies the deduplication
+    items = q.all()
```
acrellin, line 564: See comment below -- this could be a huge number of candidates. I think we want to keep the `limit`

skyportal/models.py
```diff
@@ -1510,6 +1510,8 @@ def get_if_owned_by(cls, ident, user, options=[]):
 
         return annotation
 
+    __table_args__ = (UniqueConstraint('obj_id', 'origin'),)
```
acrellin, line 1513: Why not just specify `unique=True` in `sa.Column` above?

static/js/components/CandidateList.jsx
```diff
@@ -143,12 +231,33 @@ const CandidateList = () => {
   useEffect(() => {
     if (candidates === null) {
       setQueryInProgress(true);
-      dispatch(candidatesActions.fetchCandidates());
+      dispatch(
+        candidatesActions.fetchCandidates({ numPerPage: defaultNumPerPage })
```
acrellin, line 235: Omitting numPerPage results in the backend using a default of 25 already

static/js/components/CandidateList.jsx
```diff
@@ -143,12 +231,33 @@ const CandidateList = () => {
...
+      );
     } else {
       setQueryInProgress(false);
     }
   }, [candidates, dispatch]);
 
+  const candidateHasAnnotationItem = (candidateObj) => {
```
acrellin, line 242:
```suggestion
  const candidateHasAnnotationWithSelectedKey = (candidateObj) => {
```
or is that not what's happening here? The naming could be clearer, perhaps with some in-line comments as well

static/js/components/CandidateList.jsx
```diff
@@ -143,12 +231,33 @@ const CandidateList = () => {
...
+    const annotation = candidateObj.annotations.find(
+      (a) => a.origin === selectedAnnotationItem.origin
+    );
+    if (annotation === undefined) {
+      return false;
+    }
+    return selectedAnnotationItem.key in annotation.data;
+  };
+
+  const getCandidateAnnotationValue = (candidateObj) => {
```
acrellin, line 252:
```suggestion
  const getCandidateSelectedAnnotationValue = (candidateObj) => {
```

static/js/components/CandidateAnnotationsList.jsx
```diff
@@ -0,0 +1,150 @@
...
+  const dispatch = useDispatch();
+
+  const initState = {};
+  annotations.forEach((annotation) => {
+    initState[annotation.origin] = true;
+  });
+  const [openedOrigins, setopenedOrigins] = useState(initState);
+
+  const selectedAnnotationItem = useSelector(
+    (state) => state.candidates.selectedAnnotationItem
```
acrellin, line 62:
```suggestion
  const selectedAnnotationKey = useSelector(
    (state) => state.candidates.selectedAnnotationKey
```
or `DataKey` or something else more explicit

kmshin1397:
> * Allow sorting on any field, not just numeric fields.

@dannygoldstein My reasoning for only allowing numeric fields here was that I felt the other cases were better suited for the filtering work I've started. For example, booleans only have true/false so it would make more sense to just filter on true, for example, than to sort on a boolean field. Similarly, I figured most use cases for string fields would be looking for field == some value and not necessarily sorting in alphabetical order. I can change this if you think otherwise, of course.

static/js/ducks/candidates.js
```diff
@@ -23,6 +26,13 @@ export const fetchCandidates = (filterParams = {}) => {
   return API.GET(`/api/candidates?${queryString}`, FETCH_CANDIDATES);
 };
 
+export const setCandidatesAnnotationItem = (item) => {
```
acrellin, line 29:
```suggestion
export const setSelectedCandidatesAnnotationKey = (item) => {
```

acrellin (changes requested):
Looks great overall. A few suggestions below that I think will improve clarity/maintainability. I think we'll need to add some explanatory text to this page to help point users to some of this new functionality -- it's not obvious when looking at the scanning page.

skyportal/handlers/api/candidate.py
kmshin1397: Agreed. I tried for some time to do this on the database side, but I couldn't see a straightforward solution so I just settled for this for now. Just getting distinct Objs from the passed in query was putting the sorting out of order, at least the way I was trying to do the deduplication. I'll keep pushing on this.

skyportal/models.py
kmshin1397: I believe this is here to define that within a given candidate's annotations, origin must be unique. Different candidates can have the same origin so we don't want to make the origin column itself unique.

data/db_demo.yaml
```diff
@@ -220,6 +230,58 @@ comment:
   - obj_id: 16fil
     text: "Birds are not real"
 
+annotation:
+  - obj_id: 14gqr_unsaved_copy
+    origin: "Program A"
```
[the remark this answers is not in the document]
kmshin1397: Yeah, the docs should be updated to be more clear - origin is what we are using as the equivalent to program in the old marshal. Per discussions with Dima last night on Slack: 

Screenshot 2020-10-07 162405

dannygoldstein:
> > * Allow sorting on any field, not just numeric fields.

I think it might be a bit more intuitive from a UX perspective -- each field can be clicked and sorted, rather than just the numeric types. Feels a bit funky when I go to click on an annotation but can't hover select it when I can do so on the numeric ones. How would you feel about having both the filtering and being able to sort on everything (including null)?

data/db_demo.yaml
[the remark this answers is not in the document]
dannygoldstein: See my thoughts on #1067 regarding `origin` vs `group_id`

kmshin1397:
> > > * Allow sorting on any field, not just numeric fields.
> 
> I think it might be a bit more intuitive from a UX perspective -- each field can be clicked and sorted, rather than just the numeric types. Feels a bit funky when I go to click on an annotation but can't hover select it when I can do so on the numeric ones. How would you feel about having both the filtering and being able to sort on everything (including null)?

Hm yeah I think you might be right on second thought. The selecting is also to get that key to show up in the Info column, so disabling that just to prevent un-interesting sorting does feel bad.

profjsb:
@kmshin1397 you might consider showing the coordinates in the same way we show coordinates on the source/candidate pages: https://github.com/profjsb/skyportal/blob/fe83f9e63794ba45db84059c7ad0ec00db833642/static/js/components/Candidate.jsx#L86

dannygoldstein:
I just updated / improved the coordinate display function in the link list PR #992

kmshin1397:
> @kmshin1397 you might consider showing the coordinates in the same way we show coordinates on the source/candidate pages: 

Good idea. I'll probably make and include that in a separate "minor scanning page updates" PR to try to keep the size of this one down to just the sorting/paging logic.

kmshin1397:
Added a help popover to the Autoannotations column per @acrellin's suggestion for some explanatory text on usage.
help

kmshin1397:
> > @kmshin1397 you might consider showing the coordinates in the same way we show coordinates on the source/candidate pages:
> 
> Good idea. I'll probably make and include that in a separate "minor scanning page updates" PR to try to keep the size of this one down to just the sorting/paging logic.

This is now part of #1087

dannygoldstein (approved):
Beautiful work, only one comment: 

- [x] Dont change color of page rows on hover (no need, the rows arent selectable, and it looks a bit awkward with the other components, e.g., annotationlist).

dannygoldstein:
One more thing - looks like same failures on GA and Travis. Worth investigating ...

acrellin (approved):
Looks good, thanks @kmshin1397

acrellin:
I second Danny's request to check out those identical failures -- LMK when resolved/green and I'll merge.

PR #1074, opened 2020-10-08
group sources upgrades

guynir42:
Includes filterable classifications, TNS name (needs to be finished), date saved, separating decimal/sexagesimal coords and only displaying one of them, fixing the CVS output. 
group_sources_upgrades


static/js/components/GroupSources.jsx
```diff
@@ -238,6 +259,11 @@ const GroupSources = ({ route }) => {
     );
   };
 
+  const renderDateSaved = (dataIndex) => {
+    const source = sources[dataIndex];
+    return <div key={`${source.id}_date_saved`}>{source.created_at}</div>;
```
dannygoldstein, line 264: Use `source.saved_at`

static/js/components/GroupSources.jsx
```diff
@@ -196,28 +188,57 @@ const GroupSources = ({ route }) => {
...
+  const renderClassification = (dataIndex) => {
     const source = sources[dataIndex];
     return (
-      <div key={`${source.id}_dec`}>
-        {source.dec}
-        <br />
-        {dec_to_hours(source.dec)}
-      </div>
+      <Suspense fallback={<div>Loading classifications</div>}>
+        <ShowClassification
```
dannygoldstein, line 233: Only show the most recent classification in this table

dmitryduev:
@guynir42 this looks great! May I request one thing? I would make the table width be 100% of the available screen space regardless of the amount of content, like on the Candidates page. (If there's more stuff, horizontal scrolling if perfectly fine, I think) @dannygoldstein @kmshin1397 you guys agree?

static/js/components/GroupSources.jsx
```diff
@@ -196,28 +188,57 @@ const GroupSources = ({ route }) => {
...
+          classifications={source.classifications}
+          taxonomyList={taxonomyList}
+          shortened
+        />
+      </Suspense>
+      //       <div key={`${source.id}_classification`}>
+      //          {source.latest_class}
+      //       </div>
```
stefanv, line 241:
```suggestion
```

stefanv (commented):
Needs tests, otherwise looks great!

static/js/components/GroupSources.jsx
```diff
@@ -196,28 +188,57 @@ const GroupSources = ({ route }) => {
...
   const renderRA = (dataIndex) => {
     const source = sources[dataIndex];
-    return (
-      <div key={`${source.id}_ra`}>
-        {source.ra}
-        <br />
-        {ra_to_hours(source.ra)}
-      </div>
-    );
+    return <div key={`${source.id}_ra`}>{source.ra}</div>;
```
kmshin1397, line 205: Might be a good idea to specify .toFixed() values for these numerical columns (RA, Dec, redshift) - we don't have to round too aggressively (could be like .toFixed(6) maybe?) but doesn't hurt to just have that in there in case the values are very long I think.

static/js/components/GroupSources.jsx
guynir42: The ShowClassification component shows only the latest classification...

kmshin1397: Agreed - the ShowClassification component will display the most recent classification for each taxonomy, not just overall. 

I think the best way to do this would be to refactor the ShowClassification component slightly to have an prop for toggling showing just the most recent classification vs for each taxonomy. It seems like something we could want to reuse later. Honestly, maybe just use the `shortened` flag you already put in and say, if shortened, only display the most recent classification.

static/js/components/GroupSources.jsx
guynir42: That does not exist. As far as I can tell, for sources "created_at" represents when the object was saved.

skyportal/handlers/api/source.py
```diff
@@ -260,6 +260,7 @@ def get(self, obj_id=None):
             source_info["last_detected"] = s.last_detected
             source_info["gal_lat"] = s.gal_lat_deg
             source_info["gal_lon"] = s.gal_lon_deg
+            source_info["alias"] = ""  # placeholder until we get TNS names
```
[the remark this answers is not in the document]
guynir42: I was under the impression that this is going to come in very soon. 

@dmitryduev do you know when we can expect to have TNS names? or at least where I can find them in them when they appear?

kmshin1397 (commented):
Looks good overall! A couple of minor comments, and probably could use a test or two as Stefan mentioned.

static/js/components/GroupSources.jsx
guynir42: I just wanted to leave the behavior of ShowClassifications as it was for other people using it. I agree that we can make this component more versatile. I will have to make sure it doesn't appear in too many other places where you'd suddenly get all classifications instead of one.

static/js/components/GroupSources.jsx
acrellin: https://github.com/skyportal/skyportal/blob/9c986c12407b9cf418088d7a2b29965b140fc580/skyportal/models.py#L787

guynir42:
I can't get the table to fill the entire screen, some help is needed. 

Also, it looks like the handler for sources actually returns Obj and not Source (at least, when getting multiple sources). I'm not sure how to fix this (see source.py around line 294)

kmshin1397:
> I can't get the table to fill the entire screen, some help is needed.
> 
> Also, it looks like the handler for sources actually returns Obj and not Source (at least, when getting multiple sources). I'm not sure how to fix this (see source.py around line 294)

Hi Guy, I started playing around with this and I think I fixed the width issue - I can just make a PR to your branch give me a few minutes.

kmshin1397:
@guynir42 Opened a PR to fix the width and the saved_at issues [here](https://github.com/guynir42/skyportal/pull/1)

skyportal/handlers/api/source.py
```diff
@@ -377,8 +378,22 @@ def get(self, obj_id=None):
             source_list[-1]["last_detected"] = source.last_detected
             source_list[-1]["gal_lon"] = source.gal_lon_deg
             source_list[-1]["gal_lat"] = source.gal_lat_deg
+            source_list[-1]["alias"] = ""  # placeholder until we get TNS names
             source_list[-1]["luminosity_distance"] = source.luminosity_distance
             source_list[-1]["dm"] = source.dm
+            source_list[-1]["saved_at"] = [
```
acrellin, line 384: I'm not sure what the utility of a list of datetimes is here decoupled from the actual groups they were saved to. I think the saved_at value should be added as a new key to each dict in source_list[-1]["groups"] after that's added on L401 below (will need to call to_dict() on each group beforehand)

acrellin (changes requested):

static/js/components/GroupSources.jsx
```diff
@@ -69,18 +74,42 @@ const GroupSources = ({ route }) => {
...
+        </div>
+        <div>
+          <Typography
+            variant="h4"
+            gutterBottom
+            color="textSecondary"
+            align="center"
+          >
+            Refresh page to show new sources!
+          </Typography>
```
dannygoldstein, line 103:
```suggestion
```

static/js/components/GroupSources.jsx
dannygoldstein: On the group sources page people will only be interested in the classification in the context of their group. Please refactor this to show only the latest classification from the group that is being shown

static/js/components/GroupSources.jsx
```diff
@@ -285,6 +379,13 @@ const GroupSources = ({ route }) => {
         customBodyRenderLite: renderGroups,
```
dannygoldstein, line 379: add a line break beetween each group chip

static/js/components/GroupSources.jsx
```diff
@@ -285,6 +379,13 @@ const GroupSources = ({ route }) => {
...
       },
     },
+    {
+      name: "Latest Date Saved",
```
dannygoldstein, line 383:
```suggestion
      name: "Date Saved",
```

static/js/components/GroupSources.jsx
```diff
@@ -238,6 +287,19 @@ const GroupSources = ({ route }) => {
...
+    const source = sources[dataIndex];
+
+    const group = source.groups.find((g) => {
+      return g.id === group_id;
+    });
+    if (typeof group === "undefined") {
+      return "";
+    }
+    //     group = source.groups[0];
+    return <div key={`${source.id}_date_saved`}>{group.saved_at[0]}</div>;
```
dannygoldstein, line 300: Dont render fractions of a second here - dont need that much precision

static/js/components/GroupSources.jsx
```diff
@@ -196,28 +217,56 @@ const GroupSources = ({ route }) => {
...
+
+  const renderRedshift = (dataIndex) => {
     const source = sources[dataIndex];
     return (
-      <div key={`${source.id}_dec`}>
-        {source.dec}
-        <br />
-        {dec_to_hours(source.dec)}
-      </div>
+      <div key={`${source.id}_redshift`}>{source.redshift.toFixed(6)}</div>
```
dannygoldstein, line 256: Dont use fixed point here - implies a redshfit precision which may not be achieved in reality. I.e., we dont want to render a redshift that is 0.06 as 0.060000

skyportal/handlers/api/source.py
```diff
@@ -291,6 +292,25 @@ def get(self, obj_id=None):
                 )
                 .all()
             )
+            # add the date(s) this source was saved to each of these groups
+            new_group_list = []
+            for g in source_info["groups"]:
+                saved_at = [
```
acrellin, line 298: Shouldn't there be only one saved_at value for each GroupSource row?

acrellin: Probably should be a scalar value for each group rather than a list

acrellin: Also, query.all() already returns a list -- here you're wrapping it in another list

acrellin: I think you can scrap the list comprehension and use `.first()`

guynir42: I think we talked about having the ability to save the same source to the same group more than once. 

In any case, it doesn't hurt to have it as an array if we ever do decide that is an option.

acrellin: That's not even possible with the current data model. There can only be one GroupSource row for a given obj_id, group_id pair, and saved_at is necessarily a single scalar value. Please update this to reflect the current data model. If we ever refactor that in the future, we can update this accordingly. Thanks

dannygoldstein:
merge conflicts

skyportal/handlers/api/source.py
guynir42: OK

skyportal/handlers/api/source.py
```diff
@@ -399,6 +427,34 @@ def get(self, obj_id=None):
...
+                #     s.saved_at
+                #     for s in DBSession()
+                #     .query(Source.saved_at)
+                #     .filter(
+                #         Source.obj_id == source_list[-1]["id"], Source.group_id == g.id,
+                #     )
+                #     .order_by(Source.saved_at.desc())
+                #     .all()
+                # ]
+                saved_at = (
```
acrellin, line 443: This is great! Just need to remove the commented out code above

dannygoldstein (changes requested):
@guynir42 API test is failing on Travis and GA - this failure is real

dannygoldstein (approved):

skyportal/tests/frontend/test_group_sources.py
```diff
@@ -60,3 +84,72 @@ def test_add_new_source_renders_on_group_sources_page(
...
+    # making sure the drawer is still open even after posting a classification!
+    driver.wait_for_xpath("//*[@class='vega-embed']")
+
+    # need to reload the page to see changes!
+    driver.get(f"/group_sources/{public_group.id}")
+
+    # check the classification does show up after a refresh
+    driver.wait_for_xpath(f"//*[text()[contains(., '{'Algol'}')]]")
+
+    print(public_group2.id)
```
acrellin, line 135:
```suggestion
```

skyportal/tests/frontend/test_group_sources.py
```diff
@@ -3,14 +3,34 @@
...
+    super_admin_user_two_groups,
+    public_group,
+    public_group2,
+    upload_data_token_two_groups,
+    taxonomy_token_two_groups,
+    classification_token_two_groups,
 ):
 
-    obj_id = str(uuid.uuid4())
+    print(super_admin_user_two_groups)
```
acrellin, line 21:
```suggestion
```

skyportal/handlers/api/source.py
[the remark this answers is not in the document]
acrellin:
```suggestion
```
Instead of an empty string placeholder, we should simply be able to handle the case where this is null or undefined on the front-end gracefully

skyportal/handlers/api/source.py
```diff
@@ -306,6 +307,22 @@ def get(self, obj_id=None):
...
+                    .query(Source.saved_at)
+                    .filter(Source.obj_id == source_info["id"], Source.group_id == g.id)
+                    .first()
+                    .saved_at
+                )
+                g_new = g.to_dict()
+                g_new['saved_at'] = saved_at
+                new_group_list.append(g_new)
+
+            source_info["groups"] = new_group_list
```
acrellin, line 324: A simplification:
```suggestion
            for i in range(len(source_info["groups"])):
                saved_at = (
                    DBSession()
                    .query(Source.saved_at)
                    .filter(Source.obj_id == source_info["id"], Source.group_id == g.id)
                    .first()
                    .saved_at
                )
                g_new = g.to_dict()
                g_new['saved_at'] = saved_at
                source_info["groups"][i] = g_new
```

skyportal/handlers/api/source.py
```diff
@@ -402,6 +419,7 @@ def get(self, obj_id=None):
             source_list[-1]["last_detected"] = source.last_detected
             source_list[-1]["gal_lon"] = source.gal_lon_deg
             source_list[-1]["gal_lat"] = source.gal_lat_deg
+            source_list[-1]["alias"] = ""  # placeholder until we get TNS names
```
acrellin, line 422:
```suggestion
```

skyportal/handlers/api/source.py
```diff
@@ -419,6 +437,24 @@ def get(self, obj_id=None):
                 .all()
             )
 
+            # add the date(s) this source was saved to each of these groups
+            new_group_list = []
```
acrellin, line 441: Same simplification request here as above

static/js/components/GroupSources.jsx
```diff
@@ -196,28 +207,58 @@ const GroupSources = ({ route }) => {
...
+    return <div key={`${source.id}_ra_sex`}>{ra_to_hours(source.ra)}</div>;
   };
 
   // This is just passed to MUI datatables options -- not meant to be instantiated directly.
   const renderDec = (dataIndex) => {
+    const source = sources[dataIndex];
+    return <div key={`${source.id}_dec`}>{source.dec.toFixed(6)}</div>;
+  };
+
+  const renderDec_sex = (dataIndex) => {
```
acrellin, line 238:
```suggestion
  const renderDecSex = (dataIndex) => {
```

static/js/components/GroupSources.jsx
```diff
@@ -196,28 +207,58 @@ const GroupSources = ({ route }) => {
...
-      <div key={`${source.id}_ra`}>
-        {source.ra}
-        <br />
-        {ra_to_hours(source.ra)}
-      </div>
-    );
+    return <div key={`${source.id}_ra`}>{source.ra.toFixed(6)}</div>;
+  };
+
+  const renderRA_sex = (dataIndex) => {
```
acrellin, line 227:
```suggestion
  const renderRASex = (dataIndex) => {
```

static/js/components/GroupSources.jsx
```diff
@@ -303,17 +399,23 @@ const GroupSources = ({ route }) => {
...
     source.redshift,
-    source.classifications,
-    source.groups,
-    source.recent_comments,
+    source.latest_class,
+    source.groups.map((group) => {
+      return group.name;
+    }),
+    source.saved_at,
+    "", // this last one is for the finder chart
```
acrellin, line 413: Can you please add a note explaining why an empty string is required here for the finder chart?

static/js/ducks/sources.js
```diff
@@ -93,4 +95,13 @@ const reducer = (state = initialState, action) => {
   }
 };
 
+// Websocket message handler
+messageHandler.add((actionType, payload, dispatch, getState) => {
+  const { group } = getState();
+
+  if (actionType === FETCH_GROUP_SOURCES) {
```
acrellin, line 102: I don't see this WS message type ever being pushed on the backend -- why is this needed?

acrellin (changes requested):
Looks great overall @guynir42 , just a few requests/comments/questions below

skyportal/handlers/api/source.py
guynir42: enumerate, pretty cool :)

static/js/ducks/sources.js
guynir42: I used it when I wanted the group sources page to update automatically. 

I guess we have decided to make it non-reactive, so I can remove this handler...

acrellin:
Hey @guynir42, if this is ready for re-review, please indicate that with the re-request review button, otherwise I/we don't know to take another look. I see that some recently merged changes created conflicts - let me know if you'd like a hand resolving those. Thanks!

skyportal/tests/api/test_candidates.py
```diff
@@ -13,10 +13,10 @@ def test_candidate_existence(view_only_token, public_candidate):
     status = api('HEAD', f'candidates/{public_candidate.id}', token=view_only_token)
     assert status == 200
 
-    status = api(
+    status, *_ = api(
```
acrellin, line 16:
```suggestion
    status, _ = api(
```
Just a two-element tuple returned by `api`

acrellin (approved):
Looks great!

skyportal/tests/api/test_candidates.py
guynir42: What happens if it returns a scalar? or is that never supposed to happen?

acrellin: I just mean the naming convention for unused variables. It's fine as-is, but I think `*_` is typically used to indicate multiple unused values, whereas `_` is used for single values (in this case a dict)

guynir42: I think *_ is preferred here because if there are no additional outputs beyond the first then it is ok, and if there are multiple additional outputs it is also ok. 
If you put status, _ = ... it expects exactly two outputs.

acrellin: `api` always returns at most two outputs

guynir42: Yes, but if it returns only one, and you specify status, _ = ... then it doesn't work, I think...

acrellin:
Will merge when green

skyportal/tests/api/test_candidates.py
stefanv: I propose that we always return two outputs, making it consistent for `HEAD` requests.

stefanv:
@acrellin Before (or after) merging, please check that baselayer is not downgraded.  Some PRs pin old versions of baselayer.

acrellin:
> @acrellin Before (or after) merging, please check that baselayer is not downgraded. Some PRs pin old versions of baselayer.

I just pushed up a commit pinning the latest baselayer

skyportal/tests/api/test_candidates.py
acrellin: Agreed

acrellin:
Failures appear unrelated: 
```
FAILED skyportal/tests/frontend/test_sources.py::test_upload_download_comment_attachment
```

PR #1075, opened 2020-10-08
Make position determination more robust in the starlist

profjsb:
When the positional uncertainty is nongiven or zero, make sure that the centroid positioning for the starlist/finder returns a sensible position.

### DIFF
--- a/skyportal/utils/offset.py
+++ b/skyportal/utils/offset.py
@@ -2,6 +2,7 @@ import io
 import os
 import datetime
 import warnings
+from functools import wraps
 
 import pandas as pd
 import requests
@@ -30,8 +31,21 @@ from baselayer.log import make_log
 
 log = make_log('finder-chart')
 
-warnings.simplefilter('ignore', category=AstropyWarning)
-warnings.simplefilter('ignore', category=RuntimeWarning)
+
+def warningfilter(action="ignore", category=RuntimeWarning):
+    """decorator to filter warnings using a context manager"""
+
+    def wrapper(func):
+        @wraps(func)
+        def decorated_function(*args, **kwargs):
+            with warnings.catch_warnings():
+                warnings.simplefilter(action, category=category)
+                return func(*args, **kwargs)
+
+        return decorated_function
+
+    return wrapper
+
 
 facility_parameters = {
     'Keck': {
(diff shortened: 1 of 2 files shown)


acrellin (approved):

skyportal/utils/offset.py
```diff
@@ -285,15 +286,23 @@ def _calculate_best_position_for_offset_stars(
...
+            diff_ra = np.average(df["ra_offset"], weights=df["snr"] ** 2)
+            diff_dec = np.average(df["dec_offset"], weights=df["snr"] ** 2)
+        elif how == "invvar":
+            diff_ra = np.average(df["ra_offset"], weights=1 / df["ra_unc"] ** 2)
+            diff_dec = np.average(df["dec_offset"], weights=1 / df["dec_unc"] ** 2)
+        else:
+            log(f"Warning: do not recognize {how} as a valid way to weight astrometry")
+            return (med_ra, med_dec)
+    except ZeroDivisionError as e:
+        log(f"ZeroDivisionError in calculating position with {how}: {e}")
```
stefanv, line 301: Do we want to do something special in SourceFinderHandler when these errors occur?  If so, this function may want to raise.

profjsb: No, the median position will be fine enough. It's not an error they needs to be dealt with at the user level. We're already in the "likely doesn't matter" regime with the median vs the weighted average pos.

stefanv: OK, great

stefanv:
Looks like this needs a rebase, since it contains commits from the master branch.

skyportal/utils/offset.py
```diff
@@ -31,6 +31,7 @@
 log = make_log('finder-chart')
 
 warnings.simplefilter('ignore', category=AstropyWarning)
+warnings.simplefilter('ignore', category=RuntimeWarning)
```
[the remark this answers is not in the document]
profjsb: Done. Using a decorator to watch warnings now.

stefanv (approved):

stefanv:
The patch looks good to me!  Some weird AstroPy error; related?

profjsb:
No, I think the GA failure is something unrelated. It's happening during the requirements installation.

stefanv:
OK, will test locally.

PR #1079, opened 2020-10-09
Pin newest baselayer so that DB does not need to be stamped in order to run app

acrellin:


acrellin:
@stefanv Travis seems to be quite unhappy with this BL commit for some reason

PR #1087, opened 2020-10-12
Minor scanning page updates

kmshin1397:
Address several scanning page enhancements from user feedback
- Light curves now use `days ago` instead of `mjd` for the x-axis
- Classifications are displayed in the Info column if available for a previously saved candidate
- The "save candidate to a group" button now displays group nicknames over the full name if possible

scanning_update


stefanv:
Thanks, @kmshin1397!

Someone suggested we do mjd as an axis above, and absolute magnitude on the right axis. Thoughts?

kmshin1397:
> Thanks, @kmshin1397!
> 
> Someone suggested we do mjd as an axis above, and absolute magnitude on the right axis. Thoughts?

@stefanv I have nothing against having those, but I think the implementation is not very straightforward. I tried to get the mjd as an axis above on the plot for awhile before settling on just the days ago axis. The documentation essentially just covers adding a second axis for a second dataset, and I couldn't get it to work for adding two axes for the same data (especially since we have other layers already in the plot for the non-detections and error bars making things more complicated).  Might be something we story out and revisit later - I just wanted to get the "days ago" part out first.

stefanv:
> @stefanv I have nothing against having those, but I think the implementation is not very straightforward.

Thanks for taking a look!

profjsb:
@kmshin1397 Can you rebase this PR with the latest master? I'd like to see if the tests are passing.

kmshin1397:
@profjsb master is merged in. Also added a quick update that I remembered you suggested (changing the position display to match the Candidate/Source page)

Screenshot 2020-10-14 101641

profjsb (approved):
LGTM, thanks for making the changes.

dannygoldstein:
merge conflicts

kmshin1397:
One unrelated test failure on each GA and Travis on flaky tests (profile_dropdown and show_starlist respectively)

stefanv (approved):

acrellin (approved):

PR #1090, opened 2020-10-13
We now rely on baselayer to handle all migration logic

stefanv:
See also https://github.com/cesium-ml/baselayer/pull/158


profjsb:
Do you need to wait to pin to the baselayer version once that's merged?

stefanv:
Pinned; let's see how this goes.

acrellin (approved):

PR #1101, opened 2020-10-15
Robustify Flaky Tests - Week of 10/12/20 (#1)

dannygoldstein:


skyportal/tests/frontend/test_observing_planner.py
```diff
@@ -50,7 +50,7 @@ def test_source_is_added_to_observing_run_via_frontend(
 
     driver.scroll_to_element_and_click(submit_button)
     driver.get(f"/run/{red_transients_run.id}")
-    driver.wait_for_xpath(f'//*[text()="{public_source.id}"]')
+    driver.wait_for_xpath(f'//*[text()="{public_source.id}"]', timeout=20)
```
dannygoldstein, line 53: this and the similar invocations later in this file are a poor man's solution (for now) to the root cause of the flakiness of these tests, which is described in #1102

dannygoldstein: basically, the page takes a long time to load right now because there are some non trivial flops that have to be done on the backend handler

dannygoldstein: eventually these will be cached and we can remove these longer timeouts

stefanv: Great, let's add that note of your here so we know about it.

skyportal/tests/frontend/test_sources.py
```diff
@@ -299,32 +299,27 @@ def test_super_user_can_delete_unowned_comment(
...
-    driver.execute_script("arguments[0].scrollIntoView();", comment_div)
-    ActionChains(driver).move_to_element(comment_div).perform()
-    driver.execute_script("arguments[0].click();", delete_button)
-    try:
-        driver.wait_for_xpath_to_disappear(f'//p[text()="{comment_text}"]')
-    except TimeoutException:
-        driver.refresh()
-        driver.wait_for_xpath_to_disappear(f'//p[text()="{comment_text}"]')
+    driver.scroll_to_element_and_click(delete_button)
+    driver.wait_for_xpath_to_disappear(f'//p[text()="{comment_text}"]')
```
stefanv, line 322: For some reason, I've had issues with disappear.  Looks like it works now! :man_shrugging:

stefanv (commented):
Great stuff, Danny!

stefanv (approved):

skyportal/tests/frontend/test_sources.py
```diff
@@ -299,32 +299,27 @@ def test_super_user_can_delete_unowned_comment(
...
     driver.get(f"/source/{public_source.id}")
-    driver.refresh()
+
     comment_text_div = driver.wait_for_xpath(f'//div[./p[text()="{comment_text}"]]')
     comment_div = comment_text_div.find_element_by_xpath("..")
     comment_id = comment_div.get_attribute("name").split("commentDiv")[-1]
-    delete_button = comment_div.find_element_by_xpath(
+
+    # wait for delete button to become interactible - hence pause 0.1
+    ActionChains(driver).move_to_element(comment_text_div).pause(0.1).perform()
```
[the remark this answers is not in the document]
dannygoldstein: for this, we need to do a manual actionchain movetoelement, as the delete button is only interactible if the cursor is hovering directly over the comment div. If we use a straight up `click_xpath` or `scroll_to_element_and_click` here then it will fail with an ElementNotINteractibleException such as 

https://github.com/skyportal/skyportal/runs/1256617595

skyportal/tests/frontend/test_observing_planner.py
dannygoldstein: done

PR #1103, opened 2020-10-15
normalize spectra by their integrated flux in source page spectroscopy plot

### LINKED ISSUE #1051: Spectrum flux scaling should be relative not absolute
When displaying the spectrum, the fluxes should be relative so that spectra from different nights or different telescopes do not show wild variation simply due to different slit losses or transparency variations.  The growth marshal use relative flux scaling for this reason.

dannygoldstein:
Fixes #1051 .

This normalizes the spectra in the `spectroscopy_plot` function so they can be easily compared. The method used is to normalize by the integrated flux. 

Before: 
Screen Shot 2020-10-14 at 10 36 20 PM

After:
Screen Shot 2020-10-14 at 10 35 25 PM

cc @mansikasliwal @scizen9

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -707,7 +707,6 @@ def photometry_plot(obj_id, user, width=600, height=300):
 
 # TODO make async so that thread isn't blocked
 def spectroscopy_plot(obj_id, spec_id=None):
-    """TODO normalization? should this be handled at data ingestion or plot-time?"""
     obj = Obj.query.get(obj_id)
     spectra = Obj.query.get(obj_id).spectra
     if spec_id is not None:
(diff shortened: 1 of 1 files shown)


kmshin1397 (approved):

skyportal/plot.py
```diff
@@ -717,22 +716,25 @@ def spectroscopy_plot(obj_id, spec_id=None):
...
-                }
-            )
-            for i, s in enumerate(spectra)
-        ]
-    )
+    data = []
+    for i, s in enumerate(spectra):
+
+        # normalize spectra to a common total flux (facilitates easy visual comparison)
+        normfac = np.sum(np.gradient(s.wavelengths) * s.fluxes)
```
profjsb, line 723: There is no guarantee that this is non-zero and finite. Please handle these cases.

profjsb (changes requested):

dannygoldstein:
@profjsb  Done

skyportal/plot.py
dannygoldstein: fixed

dannygoldstein:
Failures unrelated

skyportal/plot.py
```diff
@@ -717,22 +716,30 @@ def spectroscopy_plot(obj_id, spec_id=None):
...
+
+        if not (np.isfinite(normfac) and normfac > 0):
+            # otherwise normalize the value at 6000A to 2e-4
+            minimum_wave_index = np.argmin(np.abs(s.wavelengths - 6000.0))
+            normfac = s.fluxes[minimum_wave_index] / 2.0e-4
```
profjsb, line 728: I do think it's good to normalize the spectra so the y-axis isn't scaled to such a small value (since the DC level is arbitrary), but it seems like 2e-4 is a hard-coded factor that is likely related to the wavelength gradient size. A lower or higher-resolution spectrum would likely lead to a different factor. For generality it's worth calculating this factor.

profjsb (changes requested):

profjsb:
Why is the date of the Keck2 spectrum far into the future (2992!)? Perhaps the demo data needs a tweak?

dannygoldstein:
Keck2 spectrum is not part of demo data, I just uploaded it using a random date.

skyportal/plot.py
dannygoldstein: Thanks, I totally agree. I did the following in 3f77297: 

1. Instead of normalizing to a common total flux, I normalized to a common average flux per resolving element of 1 across spectra. So,  the values within a given spectrum should average to 1. 

2. In the case where a common average flux per resolving element is negative or undefined, I normalize the value of the spectrum at the median wavelength to be 1.

PR #1105, opened 2020-10-15
Source/candidate fetching optimization

acrellin:
This PR greatly improves performance when fetching sources/candidates by not including photometry by default. An optional query parameter `includePhotometry` is added to both endpoints that, when set to `true` (JSON) (becomes `True` in Python), includes associated photometry. Locally with the demo data, this results in a 10-20x speed up in fetching source/candidate data.


stefanv (approved):

kmshin1397 (approved):
LGTM!

PR #1106, opened 2020-10-15
Fix showing "new" thumbnail by default on homepage widgets

### LINKED ISSUE #1059: The "new" image should be displayed by default in the homepage widgets
Sometimes the diff image gets shown instead (or sdss/dr8):
image

image

kmshin1397:
Closes #1059 

Screenshot 2020-10-15 133056

### DIFF
--- a/skyportal/handlers/api/internal/recent_sources.py
+++ b/skyportal/handlers/api/internal/recent_sources.py
@@ -96,7 +96,7 @@ class RecentSourcesHandler(BaseHandler):
 
 
 def first_thumbnail_public_url(thumbnails):
-    urls = [t.public_url for t in sorted(thumbnails, key=lambda t: tIndex(t))]
+    urls = [t.public_url for t in sorted(thumbnails, key=lambda t: tIndex(t.type))]
     return urls[0] if urls else ""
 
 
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/internal/recent_sources.py
```diff
@@ -96,7 +96,7 @@ def get(self):
 
 
 def first_thumbnail_public_url(thumbnails):
-    urls = [t.public_url for t in sorted(thumbnails, key=lambda t: tIndex(t))]
+    urls = [t.public_url for t in sorted(thumbnails, key=lambda t: tIndex(t.type))]
```
acrellin, line 99: Is this relying on "new" being before "ref" and "sub" alphabetically? What if a "dr8" thumbnail is present?

acrellin: how about grabbing the "new" thumbnail explicitly, and then falling back to this line here if there's no thumbnail with type "new"?

acrellin: also, since this is duplicated across multiple handlers, let's just define it once and then import it in the other file

dannygoldstein (approved):
ditto ari's comment

skyportal/handlers/api/internal/recent_sources.py
kmshin1397: @dannygoldstein @acrellin No, the `tIndex()` function being called there to determine the sorting is actually defined right below this, and defines the explicit ordering of thumbnail types. So it's not just alphabetical, and all the logic including falling back is already in there.
Screenshot 2020-10-16 092457

I did the deduplication of the function though.

acrellin (approved):

PR #1107, opened 2020-10-16
Implement source transfer/unsave/invite groups to save functionality and allow target group to accept/decline

acrellin:
This PR adds functionality for unsaving a source from selected groups, requesting that selected groups save a source (and thereby covering the transfer source functionality), and adds a new table to the GroupSources page that shows requested/pending sources, with buttons for saving/declining each source. 

A new API endpoint `/api/source_groups` and associated handler are added.

See demo below (computer was being quite laggy during screen capture):
invite_group_to_save_source_then_accept


acrellin:
Note that displaying a source's save history on the front-end will be implemented in an upcoming PR

dmitryduev (changes requested):
Great work, @acrellin! My main concern is the following:
transfer-source-1
Could you try to reproduce this behavior?

dmitryduev (approved):
Awesome, just a q on parameter naming in the request body, but otherwise 👍

PR #1108, opened 2020-10-16
Add optional chaining when fetching photometry in SurveyLinkList

dmitryduev:
Found the reason why the car was driving off the cliff on the Source/Candidate page.


kmshin1397 (approved):
Will merge when green

dannygoldstein:
nice

PR #1111, opened 2020-10-18
Refactor Photometry Handler

dannygoldstein:
This pull request refactors the implementation of the Photometry handler, preserving the results of #1071 but with more stringent guarantees of database integrity.

The key differences between this implementation and that of #1071 are: 

1. A unique 6-column index is created on the photometry table to enforce deduplication *at the database level* based on the potentially duplicated values themselves instead of a derived hash. The uniqueness of floating point columns is enforced by indexing on a decimal (fixed-point, exact) representation of each floating point column, which is robust to the `==` operator. 5 digits of precision after the decimal point are retained for comparisons between single precision columns (e.g., flux, fluxerr), and 10 digits of precision are retained for comparisons between double precision columns (mjd). 

2. Searches for duplicate photometry are performed within postgres instead of within python. This is facilitated by creating temporary tables within select statements by use of the postgres `VALUES` construct. The standing postgres index keeps this lookup fast and prevents indices from having to be recreated in python. 

3. No manual hashes are calculated or used. Everything is done automatically within the database. Since the index and its unique constraint enforce the deduplication, index positions are recalculated any time a dependent photometry column value changes. This fixes the current implementation of PATCH.

4. Origin and flux are forced to be nonnull *within the database* (this PR does not make any changes to the user-facing API). Null origins or fluxes are represented within the database as `''`, and `NaN`, respectively. This enables quick duplicate searches, avoiding some subtleties with null comparisons (in postgres, `NULL != NULL`, so if flux or origin were allowed to be null, then `{mjd:1., flux:None, fluxerr:1., origin: 'abcd', instrument_id:1, obj_id: 'ZTF20abcd'}` and `{mjd:1., flux:None, fluxerr:1., origin: 'abcd', instrument_id:1, obj_id: 'ZTF20abcd'}` would not be considered duplicates).

5. Code that is duplicated in POST and PUT is refactored into methods of `PhotometryHandler` that are called by `post` and `put`. 

EDIT: This now uses floating point instead of fixed point for comparisons. This removes the maximum allowable value for floating point indexed columns, and makes the guarantee that no 2 elements are duplicates _to within floating point precision_.


stefanv (commented):
@dannygoldstein In principle, I really like this approach.  I am afraid my database skills are not quite up to par to review this PR, so I would need a bit of a tour (or maybe @acrellin is able to follow the finer nuances).

I am not sure, e.g., how the race condition is avoided.

Since you have it in your head, some comments around the intent of the various code snippets would be fantastic.

Thanks for this refactor; epic work!

dmitryduev:
Truly, @stefanv, this is absolutely epic! And I was going to say exactly that -- this is well beyond my ("emerging" lol) sql skills, would take me quite some effort to navigate here. @profjsb maybe you also could take a look? 

[Regarding the race condition -- I don't think it exists in the current implementation bc `unique=True` for the hash column, so whichever request gets there first succeeds, and the second one simply fails.]

dannygoldstein:
Thanks guys. Yes @dmitryduev you are right - there is no race. The unique constraint on the hash would have prevented this.

dmitryduev:
A question about the 6-column index -- that b-tree (or whatever postgres is using) will eventually be quite large, no? I feel that even an O(log(n)) look ups would be outperformed by fetching only the necessary bits and sorting things out in python. + the overheads of creating temporary tables

kmshin1397:
> A question about the 6-column index -- that b-tree (or whatever postgres is using) will eventually be quite large, no? I feel that even an O(log(n)) look ups would be outperformed by fetching only the necessary bits and sorting things out in python. + the overheads of creating temporary tables -- I wonder how bad that is given that there will be tens of thousands post/put photometry requests a night (again, a noob here).

I'm nowhere near an expert either, but I don't think the number of columns has a significant impact on the index performance (compared to having to compare against the index on the hash column as we have right now). The b-tree will be the same size (in terms of items in it, at least) for one-column or multi-column indices - the multiple columns are just stored as tuples for the b-tree values I think. So as long as you are comparing all the columns of the multi-column index (or a prefix of them) I don't think it'll be much, if any, slower than our querying on the hash column.

I don't have a good sense for how much overhead the temp table introduces though so still not sure about the overall performance. I can actually try asking the Schmidt Academy instructor about all this - he teaches all the database courses at Caltech so he'll definitely have some expert-level insight.

acrellin:
I'd like to take a look before this goes in, and have tagged myself. I'll try to get to it today

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +566,125 @@ def put(self):
...
-                origin=packet['origin'],
-                hash=packet["hash"],
-            )
+        # now safely drop the duplicates:
+        new = df.loc[new_photometry_ids]
 
-            params.append(phot)
+        if len(new) > 0:
+            try:
+                ids, _ = self.insert_new_photometry_data(new, instcache, group_ids)
```
kmshin1397, line 676: I feel like there is some unnecessary validation here. The "new" photometry has already been filtered for duplicates, but the self.insert_new_photometry_data() function has it's own set of coerce into VALUES and join to check for duplicates logic that I feel can be skipped in this spot.

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +566,125 @@ def put(self):
...
-            f"generate_series(1, {len(df)})"
+        new_photometry_query = (
+            DBSession()
+            .query(values_table.c.pdidx)
+            .outerjoin(Photometry, condition)
+            .filter(Photometry.id.is_(None))
         )
 
-        proxy = DBSession().execute(pkq)
+        new_photometry_ids = [g[0] for g in new_photometry_query]
```
kmshin1397, line 643: I didn't realize you could iterate through the query like this instead of the list returned by `new_photometry_query.all()`. Is there a reason for doing this instead of having the .all() at the end of the query above (and like you have for the duplicated version of this right below)?

kmshin1397 (commented):
After looking at it more and reading the `VALUES` documentation, I feel pretty confident that we should be fine regarding @dmitryduev's concern about the 6-column index and VALUES strategy scaling. At least, I don't think it will scale any worse than the hash-index we had previously. The only potential issue I see is if one photometry somehow has so many points that all the data points being inserted doesn't fit into memory and `VALUES` throws an error but I think 1) that's very very unlikely and 2) I think in that case the pandas dataframe would probably also throw a memory error before we get to the Postgres anyways.

skyportal/handlers/api/photometry.py
dannygoldstein: Just for debugging purposes. Helps to be able to have a reference to the query which I can then print.

skyportal/handlers/api/photometry.py
dannygoldstein: I'm OK with that, I would feel better about it  if we enabled the REPEATBLE READ isolation level for PUT and POST transactions.

https://www.postgresql.org/docs/9.5/transaction-iso.html

dannygoldstein: > The Repeatable Read isolation level only sees data committed before the transaction began; it never sees either uncommitted data or changes committed during transaction execution by concurrent transactions. (However, the query does see the effects of previous updates executed within its own transaction, even though they are not yet committed.) This is a stronger guarantee than is required by the SQL standard for this isolation level, and prevents all of the phenomena described in Table 13-1. As mentioned above, this is specifically allowed by the standard, which only describes the minimum protections each isolation level must provide.

dannygoldstein:
> @dannygoldstein In principle, I really like this approach. I am afraid my database skills are not quite up to par to review this PR, so I would need a bit of a tour (or maybe @acrellin is able to follow the finer nuances).
> 
> I am not sure, e.g., how the race condition is avoided.
> 
> Since you have it in your head, some comments around the intent of the various code snippets would be fantastic.
> 
> Thanks for this refactor; epic work!

I added some comments in 80ab127

skyportal/handlers/api/photometry.py
kmshin1397: Okay, yeah I think REPEATABLE READ makes sense

dannygoldstein: done

dannygoldstein:
I think this can actually probably be done without using fixed point and just keeping floating point.

dannygoldstein:
It works! Now we are really in business...

dannygoldstein:
This whole exercise has given me a lot more appreciation for the beauty of floating point numbers. We take them for granted, but they are an ingeniously efficient way to represent a hugely dynamic range of values. They solve a very difficult problem. As a species we have come a long way in the richness of our representation of numbers...

image

For fun, I love this video of what it took to do floating point operations on some of the first computers:

https://www.youtube.com/watch?v=uFQ3sajIdaM

dannygoldstein:
The index now rejects numbers that have the same floating point representation. I.e., 

```
>>> float(14.2473817487387443873) == float(14.2473817487387443872)
True
```

but 
```
>>> float(14.2474817487387443873) == float(14.2473817487387443872)
False
```

similarly,

```
>>> float(1.2474817487387443873e-40) == float(1.2474817487387443872e-40)
True
```

but 

```
>>> float(1.2474817487387443873e-40) == float(1.2473817487387443872e-40)
False
```

This offers a lot more dynamic range and flexibility than the fixed point index.

dannygoldstein:
This == condition on floats that is now used is equivalent  to the following  condition in pseudocode

`binary(float1) == binary(float2)`

skyportal/handlers/api/photometry.py
acrellin: That doesn't seem to warrant leaving it in a non-standard (at least throughout our codebase) format; it's easily changed if you need to debug and add a print statement. I'd suggest using `.all()` for consistency

acrellin: Also, if these values are indices rather than DB IDs, then the variable name should be changed accordingly (e.g. ids -> df_idxs or (even better) df_indices)

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +586,129 @@ def put(self):
...
-            flux = packet.pop('standardized_flux')
-            fluxerr = packet.pop('standardized_fluxerr')
-
-            # reduce the DB size by ~2x
-            keys = ['limiting_mag', 'magsys', 'limiting_mag_nsigma']
-            original_user_data = {key: packet[key] for key in keys if key in packet}
-            if original_user_data == {}:
-                original_user_data = None
+        # now safely drop the duplicates:
+        new = df.loc[new_photometry_ids]
```
acrellin, line 696:
```suggestion
        new_photometry = df.loc[new_photometry_ids]
```

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +586,129 @@ def put(self):
...
-                    f"Instrument {instrument.name} has no filter "
-                    f"{packet['filter']}."
+        duplicated_photometry = (
+            DBSession()
+            .query(values_table.c.pdidx, Photometry)
+            .join(values_table, condition)
+            .options(joinedload(Photometry.groups))
+        ).all()
+
+        for pdidx, duplicate in duplicated_photometry:
```
acrellin, line 678: Readability
```suggestion
        for df_index, duplicate in duplicated_photometry:
```

skyportal/handlers/api/photometry.py
```diff
@@ -853,7 +752,10 @@ def patch(self, photometry_id):
               application/json:
                 schema: Error
         """
-        _ = Photometry.get_if_owned_by(photometry_id, self.current_user)
+
+        DBSession().rollback()
+        DBSession().execute('SET TRANSACTION ISOLATION LEVEL REPEATABLE READ')
```
acrellin, line 757: Inline, explanatory comments

acrellin (changes requested):
Nice, @dannygoldstein , this is great. Just a few requests below for improved readability. Using descriptive, full word(s) variable names tremendously aids readability and thus maintainability. A piece of code can go from dense and impenetrable to clear and easy to follow just by using more descriptive variable names.

skyportal/handlers/api/photometry.py
```diff
@@ -1,13 +1,21 @@
...
 import uuid
 
+import math
+
+from sqlalchemy.ext.compiler import compiles
+from sqlalchemy.sql.expression import FromClause
+from sqlalchemy.sql import column
+
+import sqlalchemy as sa
+
```
kmshin1397, line 18: Going off of Ari's comments on readability/consistency, I would suggest reordering the imports here so that all the sqlalchemy imports are adjacent; `import math` and `import uuid` is at the top since those are standard library imports.

kmshin1397 (approved):
Looking good! I think @acrellin has some good points regarding the readability & added one of my own.

stefanv:
> For fun, I love this video of what it took to do floating point operations on some of the first computers:
> 
> https://www.youtube.com/watch?v=uFQ3sajIdaM

That was mesmerizing.

dmitryduev (changes requested):
I did some performance checks using the demo alert data from Fritz (~300 alerts with alert photometry only + a few with the full alert history of several hundred points), so running the same `./kowalski.py test` command every time, it's just the version of SP that is different. Here's a typical histogram of the results:

image

On the x axis -- the time it takes to PUT the photometry. Blue is the current implementation, orange -- this pr. The orange distribution is always shifted to the right, but the difference is not huge around the distribution maximum. What concerns me is the tail of the orange distribution. I am consistently seeing ~2x worse performance for longer light curves. We need to test this at scale (the py implementation vs pg) before it goes in.

[BTW as for the deduplication itself, both approaches work great/same lol]

dannygoldstein:
> > For fun, I love this video of what it took to do floating point operations on some of the first computers:
> > https://www.youtube.com/watch?v=uFQ3sajIdaM
> 
> That was mesmerizing.

Also fun - much of the development of   floating point" was done at UC Berkeley  - https://en.wikipedia.org/wiki/William_Kahan

stefanv:
> > > For fun, I love this video of what it took to do floating point operations on some of the first computers:
> > > https://www.youtube.com/watch?v=uFQ3sajIdaM
> > 
> > 
> > That was mesmerizing.
> 
> Also fun - much of the development of floating point was done at UC Berkeley - https://en.wikipedia.org/wiki/William_Kahan

The Kahan from "Kahan summation".  Because of the way floating point numbers are represented, order of operations matter a lot.

Jim Demmel at Berkeley is currently working on new, flexible floating point representations. He also has some interesting papers on "The Complexity of Accurate Floating Point Computation".

skyportal/handlers/api/photometry.py
```diff
@@ -49,6 +57,53 @@ def allscalar(d):
...
+                (", ".join(c.name for c in element.columns)),
+            )
+        else:
+            v = "(%s)" % v
+    return v
+
+
 def serialize(phot, outsys, format):
 
     retval = {
```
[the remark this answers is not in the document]
dannygoldstein: fixed

skyportal/handlers/api/photometry.py
```diff
@@ -365,24 +347,105 @@ def post(self):
...
+        values_table, condition = self.get_values_table_and_condition(df)
+
+        duplicated_photometry = (
+            DBSession()
+            .query(Photometry)
+            .join(values_table, condition)
+            .options(joinedload(Photometry.groups))
+        )
+
+        dict_rep = [d.to_dict() for d in duplicated_photometry]
```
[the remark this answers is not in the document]
dannygoldstein: this is just for pretty printing in the following exception

skyportal/handlers/api/photometry.py
```diff
@@ -393,11 +456,17 @@ def post(self):
...
+
+        df = df.where(pd.notnull(df), None)
+        df.loc[df['standardized_flux'].isna(), 'standardized_flux'] = np.nan
+
+        rows = df.to_dict('records')
+        upload_id = str(uuid.uuid4())
 
         params = []
         for packet in rows:
             if packet["filter"] not in instcache[packet['instrument_id']].filters:
```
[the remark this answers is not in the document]
dannygoldstein: fixed

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +586,129 @@ def put(self):
...
-            return self.error(
-                "Invalid group_ids parameter value. Must be a list of IDs "
-                "(integers) or the string 'all'."
-            )
-
-        if allscalar(data):
-            data = [data]
-
-        upload_id = str(uuid.uuid4())
+            df, instcache = self.standardize_photometry_data()
```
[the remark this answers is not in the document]
dannygoldstein: fixed

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +586,129 @@ def put(self):
...
+        try:
+            group_ids = self.get_group_ids()
+        except ValidationError as e:
+            return self.error(e.args[0])
 
-                w_oid_duplicate = (df.obj_id == oid) & (df.hash.isin(existing_hashes))
-                # now safely drop the duplicates:
-                df = df.drop(index=np.where(w_oid_duplicate)[0])
+        try:
+            df, instcache = self.standardize_photometry_data()
```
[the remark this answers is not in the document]
dannygoldstein: fixed

skyportal/handlers/api/photometry.py
dannygoldstein: fixed

skyportal/handlers/api/photometry.py
dannygoldstein: @acrellin renamed

skyportal/handlers/api/photometry.py
dannygoldstein: done

skyportal/handlers/api/photometry.py
dannygoldstein: done

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +586,129 @@ def put(self):
...
-                        groups = (
-                            DBSession()
-                            .query(Group)
-                            .filter(Group.id.in_(group_ids_update))
-                            .all()
-                        )
-                        # update the corresponding photometry entry in the db
-                        duplicate.groups = groups
+        DBSession().rollback()
+        DBSession().execute('SET TRANSACTION ISOLATION LEVEL REPEATABLE READ')
```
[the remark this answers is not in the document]
dannygoldstein: done

skyportal/handlers/api/photometry.py
```diff
@@ -493,323 +586,129 @@ def put(self):
...
-            return self.error(
-                'Top level JSON must be an instance of `dict`, got ' f'{type(data)}.'
-            )
-
-        if "altdata" in data and not data["altdata"]:
-            del data["altdata"]
-
-        # quick validation - just to make sure things have the right fields
+        DBSession().rollback()
+        DBSession().execute('SET TRANSACTION ISOLATION LEVEL REPEATABLE READ')
```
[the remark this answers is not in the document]
dannygoldstein: done

skyportal/handlers/api/photometry.py
dannygoldstein: OK I added a `validate` kwarg to `insert_new_photometry_data` which turns off the validation on request

skyportal/handlers/api/photometry.py
dannygoldstein: done

dannygoldstein:
> I did some performance checks using the demo alert data from Fritz (~300 alerts with alert photometry only + a few with the full alert history of several hundred points), so running the same `./kowalski.py test` command every time, it's just the version of SP that is different. Here's a typical histogram of the results:
> 
> image
> 
> On the x axis -- the time it takes to PUT the photometry. Blue is the current implementation, orange -- this pr. The orange distribution is always shifted to the right, but the difference is not huge around the distribution maximum. What concerns me is the tail of the orange distribution. I am consistently seeing ~2x worse performance for longer light curves. We need to test this at scale (the py implementation vs pg) before it goes in.
> 
> [BTW as for the deduplication itself, both approaches work great/same lol]

@dmitryduev Deferring this to #1113

skyportal/handlers/api/photometry.py
dannygoldstein: @acrellin No need for `all` here as we are not returning an instance of a mapped class from the query, just a single value (`id`), which would need its own iterator for extraction after `.all()` anyway, as the query returns it as a list of 1-element tuples

acrellin (approved):

kmshin1397 (approved):

dannygoldstein:
Failure on GA is due to GA artifact service being down

dannygoldstein:
Travis failure unrelated and reported on #testing-maintenance

dannygoldstein:
Visualizing floating point precision loss from DM Gregory

The blue dot is end of the desired vector. The black dot snaps to the closest representable point (in a simplified floating point grid with only 4 mantissa bits)

https://twitter.com/i/status/1044008750162604032

What the index is doing is considering duplicates any numbers that map to the same black dot

dmitryduev (approved):
For future reference, this is what the plot looks like after improvements (+ digesting a bit more "longer" time series):
image

stefanv:
Great work, @dannygoldstein; thank you also to all the reviewers for their valuable feedback.

Issue #1083, opened 2020-10-11
In saving object by ID, record who saved it and when the object was saved


profjsb:
Request Steve Schulze

acrellin:
This is already being done -- I suppose the question is then where to display this information. On the source page, we currently show a chip for each group a source is saved to; perhaps we could add who saved the source to that group and when as a tooltip for each group chip.

kmshin1397:
I think the tooltip is a good idea. Maybe we can add a history icon button at the end of the group chips like you did with the redshift as well?

Yashvi-Sharma:
Could you also add this information to API source query output?

PR #1116, opened 2020-10-21
Show source save history on source page

acrellin:
This PR adds save history to the source page, including tooltip for each group chip which shows who saved to each group when, as well as a table that can be displayed containing the entire save history (see screen capture demo below).

Closes https://github.com/skyportal/skyportal/issues/1083

source_save_history_2

### DIFF
--- a/static/js/components/SourceSaveHistory.jsx
+++ b/static/js/components/SourceSaveHistory.jsx
@@ -1,0 +1,92 @@
+import React, { useState } from "react";
+import PropTypes from "prop-types";
+import HistoryIcon from "@material-ui/icons/History";
+import Dialog from "@material-ui/core/Dialog";
+import Tooltip from "@material-ui/core/Tooltip";
+import IconButton from "@material-ui/core/IconButton";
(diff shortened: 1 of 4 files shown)


static/js/components/SourceSaveHistory.jsx
```diff
@@ -0,0 +1,92 @@
...
+            </TableBody>
+          </Table>
+        </DialogContent>
+      </Dialog>
+    </>
+  );
+};
+
+SourceSaveHistory.propTypes = {
+  groups: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
```
stefanv, line 89: How about making this shape be a bit more explicit to show what we expect on the inside?

stefanv (approved):

static/js/components/SourceSaveHistory.jsx
acrellin: Done and pushed!

kmshin1397 (approved):

stefanv:
Nice, thanks Ari!

PR #1117, opened 2020-10-21
Scanning page filtering on annotations

kmshin1397:
Implement filtering on annotation values on scanning page.
- Numeric annotation fields can be filtered for a provided range [min, max]
- Boolean annotations fields are filtered for `true` or `false`
- String annotations (and nested annotations converted to JSON strings) are filtered on exact matches for the string value provided.

filter

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -1,5 +1,7 @@
 import datetime
 from copy import copy
+import re
+import json
 
 import arrow
 
@@ -196,9 +198,9 @@ class CandidateHandler(BaseHandler):
             explode: false
             style: simple
             description: |
-              Comma-separated string of annotation filter strings formatted as
-              "{origin}:{key}:{value}" for non-numeric fields or
-              "{origin}:{key}:{min}:{max}" for numeric fields
+              Comma-separated string of JSON objects representing annotation filters.
+              Filter objects are expected to have keys { origin, key, value } for
+              non-numeric value types, or { origin, key, min, max } for numeric values.
           - in: query
             name: includePhotometry
             nullable: true
@@ -401,17 +403,27 @@ class CandidateHandler(BaseHandler):
         if annotation_filter_list is not None:
             # Parse annotation filter list objects from the query string
             # and apply the filters to the query
-            print(annotation_filter_list)
-            for item in annotation_filter_list.split(","):
-                tokens = item.split(":")
-                new_filter = {"origin": tokens[0], "key": tokens[1]}
-                if len(tokens) < 3:
+
+            for item in re.split(r",(?={)", annotation_filter_list):
+                try:
+                    new_filter = json.loads(item)
+                except json.decoder.JSONDecodeError:
                     return self.error(
-                        f"Invalid annotation filter list item: {item}. Not enough colon-separated tokens - should have at least 3 for origin:key:value."
+                        "Could not parse JSON objects for annotation filtering"
+                    )
+                print(new_filter)
+                if "origin" not in new_filter:
+                    self.error(
+                        f"Invalid annotation filter list item {item}: \"origin\" is required."
                     )
 
-                if len(tokens) == 3:
-                    value = tokens[2]
+                if "key" not in new_filter:
+                    self.error(
+                        f"Invalid annotation filter list item {item}: \"key\" is required."
+                    )
+
+                if "value" in new_filter:
+                    value = new_filter["value"]
                     # Support True/False and true/false convention
                     if value in ["True", "False"]:
                         value = value.lower()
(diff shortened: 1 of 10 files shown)


dannygoldstein:
Rebase off master ?

skyportal/handlers/api/candidate.py
```diff
@@ -383,6 +398,45 @@ def get(self, obj_id=None):
         if end_date is not None and end_date.strip() not in ["", "null", "undefined"]:
             end_date = arrow.get(end_date).datetime
             q = q.filter(Obj.last_detected <= end_date)
+        if annotation_filter_list is not None:
+            # Parse annotation filter list objects from the query string
+            # and apply the filters to the query
+            print(annotation_filter_list)
```
dannygoldstein, line 404:
```suggestion
```

skyportal/handlers/api/candidate.py
```diff
@@ -185,6 +186,19 @@ def get(self, obj_id=None):
...
+            schema:
+              type: array
+              items:
+                type: string
+            explode: false
+            style: simple
+            description: |
+              Comma-separated string of annotation filter strings formatted as
+              "{origin}:{key}:{value}" for non-numeric fields or
+              "{origin}:{key}:{min}:{max}" for numeric fields
```
dannygoldstein, line 201: What about when origin contains a comma ?

skyportal/handlers/api/candidate.py
```diff
@@ -383,6 +398,45 @@ def get(self, obj_id=None):
         if end_date is not None and end_date.strip() not in ["", "null", "undefined"]:
             end_date = arrow.get(end_date).datetime
             q = q.filter(Obj.last_detected <= end_date)
+        if annotation_filter_list is not None:
+            # Parse annotation filter list objects from the query string
+            # and apply the filters to the query
+            print(annotation_filter_list)
+            for item in annotation_filter_list.split(","):
```
dannygoldstein, line 405: Perhaps these should be parsed using json instead ?

skyportal/handlers/api/internal/annotations_info.py
```diff
@@ -0,0 +1,143 @@
...
+            .filter(
+                Obj.id.in_(
+                    DBSession()
+                    .query(Candidate.obj_id)
+                    .filter(Candidate.filter_id.in_(user_accessible_filter_ids))
+                )
+            )
+            .join(GroupAnnotation)
+            .filter(GroupAnnotation.group_id.in_(user_accessible_group_ids))
+        )
```
dannygoldstein, line 132: What exactly is happening here? Can you add an inline comment with example input and output ?

skyportal/handlers/api/candidate.py
kmshin1397: Good point, I think switching to parsing json like you suggested is the right way to go

kmshin1397:
> Rebase off master ?

Standard rebase was kind of annoying with so many commits - just squashed all the commits and then rebased off master

skyportal/handlers/api/internal/annotations_info.py
```diff
@@ -0,0 +1,143 @@
...
+
+# Helper classes/functions below provide a workaround for PostgreSQL functions
+# returning multiple columns, unsupported by SA
+# See https://stackoverflow.com/questions/33865038/joining-with-set-returning-function-srf-and-access-columns-in-sqlalchemy
+# for details
+class FunctionColumn(ColumnClause):
+    def __init__(self, function, name, type_=None):
+        self.function = self.table = function
+        self.name = self.key = name
+        self.key = self.name
```
acrellin, line 26: Looks like this is already set on the line above

skyportal/handlers/api/internal/annotations_info.py
```diff
@@ -0,0 +1,143 @@
...
+        self.is_literal = False
+
+    @property
+    def _from_objects(self):
+        return []
+
+    def _make_proxy(
+        self, selectable, name=None, attach=True, name_is_truncatable=False, **kw
+    ):
+        print('_make_proxy')
```
acrellin, line 37: I know this is just a copy/pasted recipe, but do we want to keep this print statement in here?

skyportal/handlers/api/internal/annotations_info.py
```diff
@@ -0,0 +1,143 @@
...
+        description: Collects valid annotation origin/key pairs to filter on for scanning
+        responses:
+          200:
+            content:
+              application/json:
+                schema:
+                  allOf:
+                    - $ref: '#/components/schemas/Success'
+                    - type: object
+                      description: |
```
acrellin, line 102: I think the object described here needs to be nested under a `data` property, i.e. the response JSON will look something like `{'data': {'originA': 'valueA', ...}, ...}`

skyportal/handlers/api/internal/annotations_info.py
acrellin: I can't quite follow this all, either -- comments would be helpful

static/js/components/CandidateList.jsx
```diff
@@ -286,6 +302,94 @@ const CandidateList = () => {
...
+          max: tokens[3].trim(),
+        };
+        break;
+      default:
+        break;
+    }
+    return returnObject;
+  };
+
+  const filterAnnotationObjToChip = (annotationObj) => {
```
acrellin, line 336: Where are these chip-formatted strings needed? Can we just pass around objects instead, or does MUI-Datatables require chip strings?

acrellin (approved):

static/js/components/CandidateList.jsx
kmshin1397: Uh yeah, I use these to customize the list of chips generated by MUIDataTable to display what filters are currently active - otherwise the component doesn't really know how to get that info from the autoannotation column. You can actually see an example of the chip briefly in the GIF I posted above.

dannygoldstein (approved):

PR #1137, opened 2020-10-22
Return 404 on HEAD request if source not found

stefanv:
Also require a message to be passed when calling a handler's
`error` method.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -87,7 +87,8 @@ class SourceHandler(BaseHandler):
         if num_s > 0:
             return self.success()
         else:
-            return self.error()
+            self.set_status(404)
+            self.finish()
 
     @auth_or_token
     def get(self, obj_id=None):
--- a/skyportal/handlers/base.py
+++ b/skyportal/handlers/base.py
@@ -12,5 +12,5 @@ class BaseHandler(BaselayerHandler):
     def success(self, *args, **kwargs):
         super().success(*args, **kwargs, extra={'version': __version__})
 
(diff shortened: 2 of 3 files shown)


skyportal/handlers/base.py
```diff
@@ -12,5 +12,5 @@ def associated_user_object(self):
     def success(self, *args, **kwargs):
         super().success(*args, **kwargs, extra={'version': __version__})
 
-    def error(self, *args, **kwargs):
-        super().error(*args, **kwargs, extra={'version': __version__})
+    def error(self, message, *args, **kwargs):
```
acrellin, line 15: Why is this needed?

acrellin (approved):

skyportal/handlers/api/source.py
```diff
@@ -87,7 +87,8 @@ def head(self, obj_id=None):
         if num_s > 0:
             return self.success()
         else:
-            return self.error()
+            self.set_status(404)
```
acrellin, line 90: Candidate handler probably also needs this

dmitryduev: Yes, please add the same for candidates.HEAD.

dmitryduev (changes requested):
👍

skyportal/handlers/base.py
stefanv: It's not strictly speaking needed.

skyportal/handlers/api/source.py
stefanv: Done.

dmitryduev (approved):
👍

PR #1140, opened 2020-10-22
Fix observing run MUI datatable bug

dannygoldstein:
This fixes a bug where only the first N elements were rendering in pullout rows on the observing run page, due to the use of rowIndex instead of dataIndex. So if there were 10 elements per page, the same 10 pullout rows would show on every page. This should map the pullout rows to the proper assignments on page number > 1


dannygoldstein:
failures unrelated

kmshin1397 (approved):

Issue #1142, opened 2020-10-22
Soft notification fails on fritz.science


dmitryduev:
![image](https://user-images.githubusercontent.com/7557205/96841556-c9047480-1400-11eb-8f7f-35347fcd94c9.png)

PR #1143, opened 2020-10-22
Fix source notification error associated with None user prefs

kmshin1397:
Didn't catch this during development because our demo users don't have contact_email set initially (but the real users on production do) so the logic never reached the problematic line.


kmshin1397:
Closes #1142

stefanv:
Looks OK to me!

dmitryduev (approved):
👍

PR #1144, opened 2020-10-22
Always return length-two tuple in tests.api when raw_response=False

acrellin:


acrellin:
Test failures unrelated

dannygoldstein (approved):

stefanv:
Not sure if this should be `None` or {}, but good enough for now.  Thanks @acrellin !

Issue #1132, opened 2020-10-22
Obj should not allow null ra and dec


stefanv:
Since this causes problems almost everywhere.

Fix in all API endpoints that add Objs.

Migrate db to make this column required? /cc @dannygoldstein

dannygoldstein:
Fixed in #1126

PR #1146, opened 2020-10-22
Quick fix for null RA and Dec

dannygoldstein:
Until we can converge on the more long-term solution for the ra/dec bug we are seeing on fritz.science via #1132, this PR should fix the problem. It's just a simple change of the API validator to require that RA and Dec be nonnull in Source.POST and Candidate.POST.

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -199,6 +199,8 @@ candidates:
 
 sources:
   - id: 14gqr
+    ra: 353.36647
+    dec: 33.646149
     group_ids:
       # Adding the public_group_id is a hack until the admin token i
       # allowed to create sources outside of its own groups, as per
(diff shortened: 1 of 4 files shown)


dannygoldstein:
The reason #1132 is tricky is that the RA and Dec columns are defined in Healpix alchemy, and cannot be directly overriden, so I think we need a change in healpix_alchemy to get #1132 working in a reasonable way. It also makes a database change, which this PR does not. cc @lpsinger

stefanv (approved):
A good interim measure, I think.

stefanv:
Perhaps we can add a little wrapper around ha.Point that turns it into a required column (not applicable to this PR, just thinking about a database-level solution).

dmitryduev (approved):
Thanks for the quick fix, @dannygoldstein !

dmitryduev:
will merge when green

dannygoldstein:
This is ready to go in, failures unrelated (travis is hanging, but that is unrelated to this PR)

stefanv:
Nice catch; LGTM.

PR #1148, opened 2020-10-22
Testing stability improvements

kmshin1397:
Collecting various testing improvements here as they pop up
- `test_upload_download_comment_attachment` had testing logic assuming old versions of the component where click on the link led directly to the download of attachment files, so I removed that
- Removed line that was not serving any purpose in `test_upload_download_comment_attachment` that was left over from previous tweaking of the test
- Switching to explicit clicking with ActionChains for MUI dropdown/button elements for `test_super_user_can_delete_unowned_comment` and `test_submit_annotations_filtering`.
- Plus a lot more other tweaks to tests in light of [baselayer #175](https://github.com/cesium-ml/baselayer/pull/175)


kmshin1397:
Still failing on `test_upload_spectroscopy` but the other tests seem more stable with these changes so I think it's worth merging what I have here. Still not confident that they're stable enough to disable reruns completely though.

dannygoldstein (approved):

kmshin1397:
DNM: in the long run, we should refactor to fix the root of the issue `driver.click_xpath()` in baselayer instead of patching with action chain pause calls like for much of this PR.

dannygoldstein (approved):

stefanv (approved):
Good stuff, thanks!  Feel free to also add scroll to top (either as own function, or as keyword argument) to existing baselayer utils.

dannygoldstein:
Let there be (green) light !

PR #1157, opened 2020-10-23
Let PS1 thumbnail URL generation function error rather than silently return None

acrellin:
This will keep the `Thumbnail` from being added to the DB w/ null `public_url`, and will instead re-try next time the source page is loaded.


dannygoldstein:
Are we OK not handling this exception anywhere?

acrellin:
> Are we OK not handling this exception anywhere?

@dannygoldstein handled now

stefanv (approved):

PR #1171, opened 2020-10-23
Cache ephemerides

### LINKED ISSUE #1102: cache observing run and assignment ephemeris data for faster page loading
Currently when we hit `ObservingRun.get`, all of the ephemerides for the assigned targets are calculated within the handler. This takes up a fair amount of compute, causing the page to load slowly sometimes, especially on CI. I propose caching the ephemerides on database insert / update to fix this.

dannygoldstein:
Fixes #1102


dannygoldstein:
This is not the right approach, there is a better way using vectorization.

PR #1174, opened 2020-10-23
vectorize calculation of ephemerides

### LINKED ISSUE #1102: cache observing run and assignment ephemeris data for faster page loading

dannygoldstein:
Supersedes #1171, Fixes #1102 

This PR vectorizes the calculation of rise and set times of observing run assignments. Previously, this was being done one-by-one, and consequently took ~200ms per assignment. This PR should reduce this to ~200ms for the whole run.

### DIFF
--- a/skyportal/handlers/api/observingrun.py
+++ b/skyportal/handlers/api/observingrun.py
@@ -153,14 +153,16 @@ class ObservingRunHandler(BaseHandler):
                     del a['obj'].sources
 
                 # vectorized calculation of ephemerides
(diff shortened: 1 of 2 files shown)


stefanv (approved):

dannygoldstein:
Failures unrelated

acrellin (approved):

PR #1180, opened 2020-10-24
Source list MUI table

guynir42:
Turn the table in the sources page to a MUI table. 

The search box will come in a separate PR. 

source_list_table


guynir42:
I think the test failures are unrelated...

dannygoldstein (changes requested):
In the pullout, can we add 

- [ ] A spectrum plot (I added a VegaPlot for this that is being used to preview spectra in the UploadSpectrum form)

- [ ] An autoannotations table

guynir42:
> In the pullout, can we add
> 
> * [ ]  A spectrum plot (I added a VegaPlot for this that is being used to preview spectra in the UploadSpectrum form)
> * [ ]  An autoannotations table

I added annotations, in the same style as the scanning page. 

As for spectra: I will open a separate PR for making a VegaPlot for spectra. 
There is an important question, though: there can be multiple spectra for the same source. Which one should we show?

skyportal/handlers/api/source.py
```diff
@@ -426,6 +426,7 @@ def get(self, obj_id=None):
             q = q.filter(Obj.within(other, radius))
         if start_date:
             start_date = arrow.get(start_date.strip())
+            print(start_date)
```
acrellin, line 429:
```suggestion
```

acrellin (changes requested):
My main request for now is to use the source list table that we're already using in the group sources page. That might need to be slightly modified to be even further generalized to use here. I don't think we should be duplicating this much code.

skyportal/handlers/api/source.py
guynir42: Yeah that's what I was thinking, too. But I had the feeling there would be a lot of different stuff between the two pages, but it is starting to look like the two pages would be very similar.

dannygoldstein:
> > In the pullout, can we add
> > 
> > * [ ]   A spectrum plot (I added a VegaPlot for this that is being used to preview spectra in the UploadSpectrum form)
> > * [ ]   An autoannotations table
> 
> I added annotations, in the same style as the scanning page.
> 
> As for spectra: I will open a separate PR for making a VegaPlot for spectra.
> There is an important question, though: there can be multiple spectra for the same source. Which one should we show?

All.

guynir42:
closes #1224

kmshin1397 (changes requested):
Couple of minor changes but looking good!

skyportal/tests/frontend/test_source_list.py
```diff
@@ -1,119 +1,183 @@
...
+        token=classification_token_two_groups,
+    )
+    assert status == 200
+
+    # check the classification shows up (it should not show up without a page refresh!)
+    try:
+        driver.wait_for_xpath(f"//*[text()[contains(., '{'Algol'}')]]", timeout=1)
+
+    except TimeoutException:
+        pass  # the classification should not appear on its own, so we ignore this error
```
kmshin1397, line 115:
```suggestion
    # check the classification doesn't shows up (it should not show up without a page refresh!)
    driver.wait_for_xpath_to_disappear(f"//*[text()[contains(., '{'Algol'}')]]", timeout=1)
```
Sorry @guynir42 one more thing. When you test for things to not be present, you should use the `wait_for_xpath_to_disappear` instead of the `try-catch` pattern I think - for readability, performance, and consistency.

kmshin1397 (commented):
Thanks for the changes! Sorry, just one last thing I missed last time.

skyportal/tests/frontend/test_source_list.py
guynir42: Thanks! I didn't know about that one...

kmshin1397 (approved):
Okay I think this looks good now. The test is still failing for me locally because we don't have server-side pagination yet and there are a lot of test sources to load before the timeout, but that can be worked on after this PR goes in.

acrellin (approved):

PR #1183, opened 2020-10-25
Filtering fixes for scanning page

### LINKED ISSUE #1169: Candidates (scanning) page filtering is not respected when selecting number of candidates to show
In the MUI datatable

kmshin1397:
In this PR:
- Proper coercion of annotation filter values for fields that are booleans, strings, or nested objects
- Hook up the scanning page filter form (for groups and dates) state with the paging state in the MUI data table

Closes #1169 

Remaining known issue: 
Cannot have multiple filters set on annotation fields from different origins at the same time (i.e. cannot filter for `some_key (origin A): true` and `some_other_key (origin B): true` at the same time even if a candidate has both). Vague impressions is that this will require a bigger change to the sqlalchemy code for the candidates handler and more thinking than a quick change. Something like merge annotation rows into one big jsonb field so that one row knows it has both origins? Right now the queries technically look for entries in the Annotation table that have both fields instead of objects with entries in the Annotation table that fit the fields.

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -433,6 +433,21 @@ class CandidateHandler(BaseHandler):
                             == value,
                         )
                     else:
+                        # Test if the value is a nested object
+                        try:
+                            value = json.loads(value)
+                            # If a nested object, we put the value through the
+                            # JSON loads/dumps pipeline to get a string formatted
+                            # like Postgres will for its JSONB ->> text operation
+                            # For some reason, for example, not doing this will
+                            # have value = { "key": "value" } (with the extra
+                            # spaces around the braces) and cause the filter to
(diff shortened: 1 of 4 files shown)


dannygoldstein (approved):

acrellin (approved):
These changes look good. Could the multiple origins filtering issue be resolved with a combination of `or_` and `and_` filters?

acrellin:
Merge conflicts

kmshin1397:
> These changes look good. Could the multiple origins filtering issue be resolved with a combination of `or_` and `and_` filters?

Yeah, I'm not sure but that would be the first thing I would try. I made an issue for it (#1185) and will revisit it soon.

dannygoldstein:
Failures were unrelated

PR #1184, opened 2020-10-26
Refactor About page to work off a base template and use MUI Paper

kmshin1397:
With the PR, the `About.jsx` component now has Jinja-style "block" declarations throughout the file using comments so that the file remains valid JSX but other applications can now use it as a parent template to easily overwrite specific parts of the component.

Additionally, the contents of the component were refactored from the current version to use a MUI Paper background and MUI Typography components. 

Screenshot 2020-10-25 160821
 
Once this is merged, a corresponding PR will be made to Fritz to actually put in Fritz-specific info.


stefanv:
Thanks, @kmshin1397! One quick question: would it be possible not to make the base template compiled? I.e.,use it as a base template, but have a jsx extension (maybe by using comments in the right places, eg).

I worry that templates are harder to edit and work with, so it would be nice to only use them as needed.

Ideally, skyportal would render a sensibly named default page, and Fritz would inherit and extend as necessary.

kmshin1397:
@stefanv I think I've got it to a nice, elegant state. Essentially just adds the Jinja "block" tags as comments in the original JSX so that nothing really changes on the SkyPortal side, but Fritz can write an `About.jsx.template` to inherit and overwrite the blocks.

stefanv:
This looks good, @kmshin1397.  There's still mention of BaseAbout in the docstring, which I presume is outdated?

I am a bit confused about why this works, though.  Why doesn't the comment also comment out the inserted block of text?

kmshin1397:
@stefanv You can basically write the inserted blocks such that it doesn't comment out any actual code. For example, the block statements commented out with "//" can insert a newline before the actual content you want to insert so that there's just an empty comment line. Similarly you can add "{/*" or "*/}" where relevant in the text you insert into the blocks to close the comments in the html blocks without having it affect the actual code you're inserting. I've tested it and while the produced JSX is somewhat ugly in that you'll end up with a bunch of empty comments, it works.

static/js/components/About.jsx
```diff
@@ -79,40 +114,63 @@ BibLink.propTypes = {
...
+
+  // {% block configVariables -%}
+  // Replace these with Jinja double-brace injection when extending if you just
+  // want to use a config file variable instead of this default appTitle without
+  // overwriting a render block that uses it below.
+  //
+  // This is done like this instead of just directly writing something like
+  // {{ app.title }} below because the actual double-brace injection syntax will
+  // not be valid JSX if not extending the About component.
+  const appTitle = "SkyPortal";
```
profjsb, line 133: appTitle should be important from the config.yaml, not hardcoded here.

static/js/components/About.jsx
```diff
@@ -79,40 +114,63 @@ BibLink.propTypes = {
...
         <a href="https://skyportal.io">https://skyportal.io/docs/</a>
-      </p>
-      <p>
-        You may also interact with SkyPortal through its API. Generate a token
+      </Typography>
+      <Typography variant="body1">
+        You may also interact with {appTitle} through its API. Generate a token
         from your&nbsp;
         <Link to="/profile">profile</Link>&nbsp; page, then refer to the&nbsp;
         <a href="https://skyportal.io/docs/api.html">API documentation</a>.
```
profjsb, line 159: Likewise, should import doc URL from config.yaml

profjsb (changes requested):

static/js/components/About.jsx
kmshin1397: As I explain in the inline comments here, I don't think there's a way to import the values from the config.yaml without restructuring this PR back to the way I had it before @stefanv's suggestions. I'd have to either make this a .template file or make it an invalid .jsx file and get our linters to ignore it. I think converting it to an `About.jsx.template` is bad because the whole point of the PR was so that Fritz wouldn't have to take an entire file and rewrite it; that would lead us back to the `BaseAbout.jsx.template` that was rejected earlier in this PR. I thought this solution was an acceptable compromise. Note that while in the base SkyPortal version of the About component this is hard-coded, any application extending the page (i.e. Fritz) can import relevant app-specific configs by overwriting the `configVariables` Jinja block when extending this file.

static/js/components/About.jsx
kmshin1397: Ah ok, sure I think moving these to the configVariables block above (like with the appTitle) is good. In this case, none of the URL's were actually in the configs before this PR either so I was just carrying that over. But again, I would push back on directly importing from the configs here for the reasons I stated above.

static/js/components/About.jsx
```diff
@@ -1,3 +1,26 @@
...
+  backslash) after {% block %} declarations within the JSX component render()
+  segment, and "{/*" immediately before the {% endblock %} declarations within
+  the render().
+
+  See https://jinja.palletsprojects.com/en/2.11.x/templates/#template-inheritance
+  for documentation on usage of Jinja templates.
+#}
+*/
+
+// {/* {% block imports -%} */}
```
stefanv, line 23: Can we stick to the `//` syntax consistently everywhere?  That would also make it easy to explain.

static/js/components/About.jsx
stefanv: Simply making it possible to override the title should be fine, given that any overrides will happen in a template.  In *that* template, the title *will* be accessible.

static/js/components/About.jsx
stefanv: From the point of fritz, it's sufficient to have these overridable (Fritz can rewrite with its own paragraph).

This paragraph is not really applicable to other projects, so think using SkyPortal by default makes sense here.

static/js/components/About.jsx
kmshin1397: I wish, but within the HTML parts the `//` doesn't actually make a comment so I don't think we can use it for the blocks in the return() statement.

static/js/components/About.jsx
kmshin1397: Actually, with regards to getting doc URL from config.yaml, do we really want that? If we make that change, why not also get the github URL and the project homepage URL? I think it'll start bloating the configs unnecessarily. As Stefan says, these segments will likely be overridden by Fritz and other applications anyways so I don't know if moving the URLs to the configs are worth it.

kmshin1397:
Closing in favor of #1254

PR #1187, opened 2020-10-26
Interactive Finder Page

profjsb:
This PR makes a finding chart page where the user can configure finding chart parameters interactively, then save/print the computed finder.
Kapture 2020-10-29 at 11 27 15


config.yaml.defaults
```diff
@@ -61,6 +61,8 @@ app:
       component: UserManagement
     - path: "/upload_spectrum/:id"
       component: UploadSpectrum
+    - path: "/finder/:id"
```
stefanv, line 64: Should this be `/source/:id/finder` and then specify the correct query parameter below in the URL?

package.json
```diff
@@ -39,14 +39,16 @@
     "react-datepicker": "3.1.3",
     "react-dnd": "7.7.0",
     "react-dom": "16.13.1",
+    "react-image": "4.0.3",
     "react-file-previewer": "0.6.3",
     "react-grid-layout": "1.1.1",
-    "react-hook-form": "5.7.2",
+    "react-hook-form": "6.5.0",
```
stefanv, line 45: @acrellin, will this work overall?  You tried to upgrade last time and ran into some subtleties.

skyportal/handlers/api/source.py
```diff
@@ -993,14 +993,34 @@ async def get(self, obj_id):
...
+        - in: query
+          name: type
+          nullable: true
+          schema:
+            type: string
+            enum: [png, pdf]
+          description: |
+            output type
+        - in: query
+          name: how_many
```
stefanv, line 1005: I realize this is also an issue with the Python implementation, but `how_many` is a very generic argument name.  Makes sense for `get_nearby_offset_stars`, but not so much for `get_finding_chart` (or for this endpoint).

Perhaps use `how_many_offset_stars` or `nr_offset_stars`?

static/js/components/FindingChart.jsx
```diff
@@ -0,0 +1,280 @@
...
+    howmany: 3,
+  });
+
+  const componentRef = useRef();
+
+  const initialFormState = {
+    ...params,
+  };
+
+  const url = [
```
stefanv, line 65: Would be cleaner to use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

stefanv (commented):
I just did a quick read-over; looks good overall!

static/js/components/FindingChart.jsx
```diff
@@ -0,0 +1,280 @@
...
+                          </Controller>
+                        </FormControl>
+                      </Grid>
+                      <Grid item xs={12}>
+                        <FormControl>
+                          <InputLabel
+                            className={classes.items}
+                            id="PositionSelect"
+                          >
+                            Use DR2 positions?
```
dannygoldstein, line 167:
```suggestion
                            Use Gaia DR2 positions for offsets
```

static/js/components/FindingChart.jsx
```diff
@@ -0,0 +1,280 @@
...
+    documentTitle: `finder_${id}.pdf`,
+    pageStyle: "@page {size: landscape}",
+  });
+
+  return (
+    <>
+      <div>
+        <div>
+          <Typography variant="h5" gutterBottom>
+            {`Interactive Finder for ${id}`}
```
dannygoldstein, line 105: Make this a link

dannygoldstein: (to the source page for `${id}`)

dannygoldstein (commented):
Looking good. Few comments:

- Matplotlib figures should be closed after they are done being used, currently the warning 

```

/Users/danny/PycharmProjects/skyportal/skyportal/utils/offset.py:800: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
  fig = plt.figure(figsize=(11, 8.5), constrained_layout=False)
``` 

is flashed

- On both the print button and the update button, the ""Finding chart generation in progress. Download will start soon." notification is often shown twice on a single click

package.json
acrellin: @stefanv @profjsb no, unfortunately it looks like this introduces some failures. We'll need to update our usage in order to upgrade the the new major version. Suggest reverting this change.

profjsb: Ok, I'll revert

skyportal/handlers/api/source.py
profjsb: Sure.

static/js/components/FindingChart.jsx
```diff
@@ -0,0 +1,280 @@
...
+                        <FormControl>
+                          <InputLabel
+                            className={classes.items}
+                            id="PositionSelect"
+                          >
+                            Use DR2 positions?
+                          </InputLabel>
+                          <p />
+                          <Controller
+                            as={
```
stefanv, line 171: https://material-ui.com/components/switches/

kmshin1397 (commented):
`test_starlist` is failing because of your changes to the source offset handler I think (the `how_many` to `num_offset_stars` change

That, and the merge conflict with baselayer, but otherwise looks good!

profjsb:
Unrelated failure on Travis (green on GA)

kmshin1397 (approved):

Issue #914, opened 2020-09-16
Expand user management page


acrellin:
- [x] Add functionality to allow admins to add existing user group membership
- [x] Add functionality to allow admins to add existing user stream membership
- [ ] Show pending invites & allow admins to modify invite group/stream IDs
- [x] Add ability to manage user roles/ACLs

PR #1194, opened 2020-10-27
Allow adding user to streams from user management page

acrellin:
Partially addresses https://github.com/skyportal/skyportal/issues/914 

add_user_to_streams_from_user_mgmt_pg

### DIFF
--- a/skyportal/tests/conftest.py
+++ b/skyportal/tests/conftest.py
@@ -118,6 +118,11 @@ def public_stream():
 
 
 @pytest.fixture()
+def public_stream2():
+    return StreamFactory()
+
+
+@pytest.fixture()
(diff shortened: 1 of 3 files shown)


dmitryduev (approved):
Great, thanks @acrellin!

kmshin1397 (approved):

kmshin1397:
@acrellin I see a failure on GA for `test_add_user_to_group` which I don't remember as being flaky recently. Seeing as that's also on the user management page you could look into it and see if it's real? Otherwise I think this is good to go.

kmshin1397:
@acrellin You're right it seems to be passing for me locally as well when I check out your branch. Must just be some unrelated flakiness that we should keep an eye on.

PR #1195, opened 2020-10-27
Add wheel as a requirement

stefanv:
This is needed by astroplan, which is a dependency of astropy


stefanv:
But, needs to be installed earlier.

PR #1197, opened 2020-10-27
Fix candidate passed_at parsing

dmitryduev:
In this PR:

- In `/api/candidates.POST`, ingest `passed_at` into the db as `arrow.get(passed_at).datetime` (to preserve the current data model)
- Report filter_id,passed_at,passing_alert_id in `/api/candidates.GET` as in
image

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -663,7 +663,7 @@ class CandidateHandler(BaseHandler):
         return self.success(data={"id": obj.id})
 
     @permissions(["Manage sources"])
-    def patch(self, obj_id):
+    def put(self, obj_id):
         """
         ---
         description: Update a candidate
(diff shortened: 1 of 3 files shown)


acrellin (approved):

dannygoldstein:
FAIL: skyportal/tests/api/test_candidates.py:50 in test_token_user_update_candidate

stefanv:
Looks good, but touches the data model, so requires a migration @dannygoldstein ?

dmitryduev:
@stefanv, no, no data model touching! :) Specifically for this reason, I did `passed_at = arrow.get(passed_at).datetime` instead of changing the column type.

stefanv:
You did set nullable to true, though.  Is that the current value?  If so, we can remove.

dmitryduev:
It was nullable and I kept it like that. It's just black's reformatting after I added some clarifying help-text

stefanv:
OK, cool

PR #1200, opened 2020-10-28
make observing run page load much faster

### LINKED ISSUE #1193: Runs page blocks threads and takes long to load
This means that all threads get tied up from time to time, causing the status backend to be targeted (bad).

Launch any long-running processes either as threads, or use websockets to trigger loading of resource on frontend once done computing.

dannygoldstein:
The observing run page was loading slowly due to an inefficient query that was sorting on unindexed columns that were not even needed to render the page. This PR fixes #1193. Fixes #1171 (and also is the true fix to #1102).


dannygoldstein:
failures unrelated

acrellin (approved):

PR #1205, opened 2020-10-28
Remove caching and renewal of IERS table downloads

stefanv:
We were using internal functions that are no longer available.
AstroPy now handles the caching internally, so we don't need these
fixtures any longer.


dannygoldstein (approved):
Awesome!

stefanv:
See also https://github.com/skyportal/skyportal/pull/1206

dannygoldstein:
failures unrelated

PR #1208, opened 2020-10-28
Record Users that reduced and observed a spectrum

dannygoldstein:
cc @mansikasliwal

This PR introduces a dropdown to the upload spectrum page that keeps track of who reduced and observed a spectrum. The uploader can select any number of users on the website from the menu to populate each field. This will make it easier to track who should be cited when a spectrum is used in a paper.


dannygoldstein:
failure unrelated

stefanv:
The PR seems technically sound, but I need someone more familiar with the domain to review. @profjsb? @dmitryduev?

acrellin (approved):

dmitryduev (approved):
LGTM

PR #1211, opened 2020-10-28
Refactor User ACLs

acrellin:
This PR updates user ACL checking to always use User.permissions for accessing the full ACLs list (role-level + those specified by the new `user_acls` table), and pins a new baselayer that adds that new join table and updates `User.permissions` as described above.


stefanv:
Looks good, but requires a migration before merging.

acrellin:
@stefanv migration script is now in place

skyportal/handlers/api/internal/token.py
```diff
@@ -35,8 +35,7 @@ def post(self):
         data = self.get_json()
 
         user = User.query.filter(User.username == self.current_user.username).first()
-        user_acls = {acl.id for acl in user.acls}
-        token_acls = set(data['acls']) & user_acls
+        token_acls = set(data['acls']) & set(user.permissions)
```
dannygoldstein, line 38: I think this should throw an exception when the user posts an ACL they dont have rather than silently not adding it to the token

skyportal/handlers/api/observingrun.py
```diff
@@ -191,7 +191,7 @@ def put(self, run_id):
 
         data = self.get_json()
         run_id = int(run_id)
-        is_superadmin = "System admin" in [a.id for a in self.current_user.acls]
+        is_superadmin = "System admin" in self.current_user.permissions
```
dannygoldstein, line 194: Can we make this a method of `User`?

dannygoldstein (approved):

acrellin:
@dannygoldstein all fair asks, but all related to pre-existing behavior that wasn't changed here. Will file an issue and address soon.

dmitryduev (approved):
LGTM!

acrellin:
Merging with two explicit approvals and @stefanv's verbal approval above

dmitryduev:
I just figured that I didn't check what the CI failures were, so was doing that. Apparently, unrelated; if not -- we'll know soon enough :)

acrellin:
> I just figured that I didn't check what the CI failures were, so was doing that. Apparently, unrelated; if not -- we'll know soon enough :)

@dmitryduev oh, sorry!
Yes, unrelated failures:
```
FAILED skyportal/tests/frontend/test_user_management.py::test_add_user_to_group
```

Issue #1226, opened 2020-10-29
Default to no groups selected on scanning page


cfremling:
Defaults on for example the scanning page need to be more reasonable. e.g.,
default to showing data from the previous night only. And definitely do not start with all programs enabled when initially visiting the page.

All similar issues users may have could alternatively be solved by having the system memorize many of the user selections across the site (when it makes sense). If I select Program. X as the only program to scan, that should still be my selection the next time I visit the page. I think the vast majority of users would agree with this.

acrellin:
@cfremling an alternate approach would be to have no programs selected by default, showing no candidates, and then you can simply click the program(s) you want to filter on and see the results each time. This is much more likely to be implemented than saving your program selection from last time.

acrellin:
I believe this is also how the old marshal works

cfremling:
Right! yes this would be fine for now.

Christoffer
________________________________

Issue #1223, opened 2020-10-29
Display more candidate dates on scanning page


dmitryduev:
Currently, it is only possible to filter by date based on the `last_detected` attribute, which is simply the time stamp of the most recent photometry point. 

The users, however, mostly need to know when a candidate passed (a) particular filter(s) most recently, or for the first time. 

I propose to: 
- ~~Add a dropdown menu on the Scanning page that would allow the user to select `dateType` from `created_at`, `last_detected`, or `passed_at` that will be used to filter and order the candidates on the backend. `created_at` is easy as it's an `Obj` attribute, but `passed_at` is more involved, since it's a `Candidate` attribute and can be null / different for different filters. Would probably make sense to do something like `any([start_date <= passing_alert.get("passed_at", created_at) <= end_date for passing_alert in passing_alerts])`, but other ideas are most welcome.~~ (Addressed in #1228 )
- For each candidate, in addition to `last_detected`, display `created_at` (a single date) and a list of `passed_at` for the filters of the selected groups/programs.

PR #1228, opened 2020-10-29
Scanning page filtering updates

acrellin:
In this PR:
- change the scanning page filtering form to default to zero groups selected;
- apply date range filtering to `Candidate.passed_at` rather than `Obj.last_detected`;
- `Candidate.passed_at` is required both at the API level and DB level (and is now indexed);
- test updates

Closes https://github.com/skyportal/skyportal/issues/1226
Partially addresses https://github.com/skyportal/skyportal/issues/1223

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -135,7 +135,7 @@ class CandidateHandler(BaseHandler):
               type: string
             description: |
               Arrow-parseable date string (e.g. 2020-01-01). If provided, filter by
-              last_detected >= startDate
+              Candidate.passed_at >= startDate
           - in: query
             name: endDate
             nullable: true
@@ -143,7 +143,7 @@ class CandidateHandler(BaseHandler):
               type: string
             description: |
               Arrow-parseable date string (e.g. 2020-01-01). If provided, filter by
-              last_detected <= endDate
+              Candidate.passed_at <= endDate
           - in: query
             name: groupIDs
             nullable: true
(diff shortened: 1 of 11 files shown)


acrellin:
Failures unrelated

 FAILED skyportal/tests/frontend/test_user_management.py::test_add_user_to_group
FAILED skyportal/tests/frontend/test_recent_sources.py::test_recently_saved_candidate

dannygoldstein (commented):
This one is new, and seems related:

FAILED skyportal/tests/frontend/test_recent_sources.py::test_recently_saved_candidate

Does `passed_at` currently have an index? And is it db non-nullable? If not, we will need to add these. 

That would in turn require a migration script.

acrellin:
@dannygoldstein that's beyond the scope of what this PR is meant to
address. Please file an issue for those noted schema changes for a future
PR.

skyportal/handlers/api/candidate.py
```diff
@@ -411,10 +415,10 @@ def get(self, obj_id=None):
             "undefined",
         ]:
             start_date = arrow.get(start_date).datetime
-            q = q.filter(Obj.last_detected >= start_date)
+            q = q.filter(Candidate.passed_at >= start_date)
```
dannygoldstein, line 418: I think it should go in to this PR as this new line is filtering on `passed_at`, which, without an index, could become a bottleneck

acrellin: @dannygoldstein it was previously filtering on an unindexed property, so no change there

acrellin: If you insist, please file a PR against this branch

dannygoldstein: done

dmitryduev (changes requested):
@acrellin looks great, thanks a lot!! I think it would be worth it merging @dannygoldstein's pr with the model changes (once the conflicts are fixed + see my comment on it) -- we have a bunch of candidates in the db with null `passed_at` columns, so the migration script would need to init those with `Obj.created_at`

acrellin:
@dannygoldstein @dmitryduev ready for re-review

dmitryduev (approved):
LGTM!

PR #1271, opened 2020-11-03
Fix missing candidates' saved groups info

### LINKED ISSUE #1270: Already saved candidates on the Scanning page
- The scanning page is not always showing the chips with the group nicknames
- That info is also not always propagated to the Edit Groups modal
bug-4

kmshin1397:
Closes #1270 
savedgroups


dmitryduev (approved):
Lol, thanks @kmshin1397 :)

kmshin1397:
Yeah I was really confused why Python wasn't complaining when I found the error, since I never even initialized that variable. Turns out it was initialized for the single candidate part of the handler, so it was just always None for the scanning page :)

PR #1274, opened 2020-11-04
Manage invitations

### LINKED ISSUE #914: Expand user management page

acrellin:
This PR adds functionality for managing pending user invites including adding/removing groups or streams from the invitation or deleting it entirely.

Closes #914 

manage_pending_invites


skyportal/handlers/api/invitations.py
```diff
@@ -133,3 +133,136 @@ def post(self):
...
+        invitation = Invitation.query.get(invitation_id)
+        group_ids = data.get("groupIDs")
+        stream_ids = data.get("streamIDs")
+        if group_ids is None and stream_ids is None:
+            return self.error(
+                "At least one of either groupIDs or streamIDs are requried."
+            )
+        if group_ids is not None:
+            groups = Group.query.filter(Group.id.in_(group_ids)).all()
+            if len(groups) != len(group_ids):
```
kmshin1397, line 214: Can we do something like a set difference here to return the specific invalid group IDs in the error message? Same thing with the stream IDs right below

static/js/components/UserManagement.jsx
```diff
@@ -271,6 +368,7 @@ const UserManagement = () => {
     const results = await Promise.all(promises);
     if (results.every((result) => result.status === "success")) {
       dispatch(showNotification("User(s) successfully invited."));
+      dispatch(invitationsActions.fetchInvitations());
```
kmshin1397, line 371: This fetch is just to have the new invite show up on the management page, right? I think it would be better to do that through a webhook instead for consistency with the rest of SP

static/js/components/UserManagement.jsx
```diff
@@ -449,6 +547,94 @@ const UserManagement = () => {
       <br />
       {invitationsEnabled && (
         <>
+          {!!invitations?.length && (
+            <>
+              <Typography variant="h5">Pending Invitations</Typography>
+              <Paper elevation={1}>
```
kmshin1397, line 553:
```suggestion
              <Paper elevation={1} className={classes.invitations}>
```
I would add a corresponding styling class at the top of the file to add a marginBottom to this paper (right now the bulk invite header is right up against the pending invitations table).

Something like this:
`  invitations: {
    marginBottom: "1rem",
  },`

kmshin1397 (commented):
A few minor suggestions

static/js/components/UserManagement.jsx
acrellin: I don't think what you're proposing is the right pattern here. Note that the invitations will never be fetched for most users (only fetched when a system admin user loads the user management page, not in hydrate), so for most users that branch of the state will always be null, as opposed to changes to sources, e.g., which should use the websocket push. Most users don't even have access to the endpoint to fetch invitations.

static/js/components/UserManagement.jsx
acrellin: 👍

skyportal/handlers/api/invitations.py
acrellin: Sure

static/js/components/UserManagement.jsx
kmshin1397: Ah okay yeah, you're right let's keep this as is.

kmshin1397 (approved):

acrellin:
@dmitryduev ready for review :)

dmitryduev (approved):
LGTM!

PR #1284, opened 2020-11-05
Fix git-lfs install pubkey error

acrellin:


stefanv:
Why are we installing git-lfs?

acrellin:
> Why are we installing git-lfs?

@stefanv no idea, but there are apt errors related to git-lfs being unsigned

PR #1288, opened 2020-11-05
Fix display of classifications with null probabilities

kmshin1397:
Properly ignore classifications with null probabilities on scanning page and home page widgets


dannygoldstein (approved):

dmitryduev (approved):
👍 thanks for the quick fix!

PR #1302, opened 2020-11-06
fix group display on observing planner

### LINKED ISSUE #1155: Do not show single-user groups on Run page
image

dannygoldstein:
This PR fixes the group chip display on the observing run page. Fixes #1155. 

Whereas before long group names would be shown on the observing planner, causing groups to get cut off, and single user groups would be shown, this PR shows only group nicknames, and does not show single user groups. 

A test is updated to check that this is enforced.


dmitryduev (approved):
👍

acrellin (approved):

Issue #1303, opened 2020-11-06
Source page MUI datatable crashes with  Cannot read property 'substring' of undefined


dannygoldstein:
![Screen Shot 2020-11-06 at 12.32.51 AM.png](https://images.zenhubusercontent.com/59e:2c99cb0222d5de47baddd/d8bcc29e-aa65-4348-b2e5-8f7072e52bda)

To reproduce: 

1. Navigate to fritz.science/sources
2. Advance the table ~7-15 pages until the site crashes

dmitryduev:
The cause is probably the same as for #1294

kmshin1397:
This is due to sources having no `saved_at` field for whatever reason, leading to failure at SourceTable.jsx:350. Example I found was ZTF20acdewdg but there are probably more.

kmshin1397:
actually that might not be it - just realized the table is displaying group.saved_at and not source.saved_at

kmshin1397:
Ah, I think it's due to some sources not having any groups visible to the user (for example, I don't see any groups on ZTF20ackefmn - I'm assuming it was saved to a group I don't have access to). We should probably filter those out in the source handler.

PR #1295, opened 2020-11-06
Sources table server side pagination

acrellin:
source_list_pagination

group_sources_refactored

Closes #1294 
Closes https://github.com/skyportal/skyportal/issues/1285

Issue #1224, opened 2020-10-29
Sorting by save date on the group_sources page does not work


cfremling:
Sorting by "Date Saved" on the group_sources page does not work

kmshin1397:
Duplicate of [fritz #100](https://github.com/fritz-marshal/fritz/issues/100)

PR #1306, opened 2020-11-06
Remove unsaved sources from source GET results

kmshin1397:
Don't include sources that were unsaved from the GET multiple sources handler.

Partially addresses #1303 and #1224 (likely fixes backend cause of the error, frontend patched in #1295)

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -394,15 +394,20 @@ class SourceHandler(BaseHandler):
             query_options.append(
                 joinedload(Obj.photometry).joinedload(Photometry.instrument)
             )
+        source_filter_condition = (
+            or_(Source.requested.is_(True), Source.active.is_(True))
+            if include_requested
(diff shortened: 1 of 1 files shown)


acrellin (commented):
@kmshin1397 do you know why lines 500-504 aren't already taking care of this? when a source is unsaved it becomes `active = False`

kmshin1397:
@acrellin Oh, didn't even notice that was there. It looks like that filter is filtering out the groups that are inactive (so just affecting the `source.groups` field returned) but the object itself is still being returned - hence the empty `source.groups` arrays.

acrellin:
@kmshin1397 oh right that's for groups, the filtering on sources is on 446/447 -- perhaps we should just apply that condition outside of the conditional block it's in, too?

kmshin1397:
@acrellin Oh yeah I see it now. Yeah I think pulling that out is the correct fix here instead of what I added.

acrellin (approved):

PR #1322, opened 2020-11-10
Robustify tests

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1347


skyportal/tests/api/test_sources.py
```diff
@@ -193,7 +193,7 @@ def test_token_user_post_new_source(upload_data_token, view_only_token, public_g
     npt.assert_almost_equal(data['data']['ra'], 234.22)
 
     saved_at = parser.parse(data['data']['groups'][0]['saved_at'] + " UTC")
-    assert abs(saved_at - t0) < timedelta(seconds=2)
+    assert abs(saved_at - t0) < timedelta(seconds=60)
```
acrellin, line 196: This failed locally with a difference of ~25 seconds once, even though the test completed within a few seconds...
This has also failed on CI

acrellin:
@kmshin1397 's upcoming PR to refactor the user management page to use mui-datatables with server-side pagination should address the failures we're seeing on the user management page (during test runs the table of users can grow to many rows, and regular material-ui tables, which are currently being used on the user management page, can be ridiculously slow)

acrellin:
GA failures are down to two on the user management page which will be fixed in an in-progress PR: 
```
FAILED skyportal/tests/frontend/test_user_management.py::test_add_user_to_stream
FAILED skyportal/tests/frontend/test_user_management.py::test_add_user_role
```

PR #1338, opened 2020-11-11
increase GA timeout to 90 minutes

dannygoldstein:
Most of the current GA builds are timing out. This PR increases the GA build timeout to 90 minutes from 45 minutes.


dannygoldstein:
self-merging as this is currently preventing our CI from running.

PR #1340, opened 2020-11-12
attempt to circumvent spam filters from invite text 

dannygoldstein:
Fritz.science invites were being filtered to spam. I think it may be because some of the text of the invite appears in the spam list of triggering words, specifically "click here." 

https://blog.hubspot.com/blog/tabid/6307/bid/30684/The-Ultimate-List-of-Email-SPAM-Trigger-Words.aspx

This PR rewrites the invite text somewhat to clear out these triggering words


dannygoldstein:
Closing as @dmitryduev's investigations revealed that this has to do with a 554 error due to sendgrid being an untrusted source for many mailservers.

PR #1342, opened 2020-11-12
increase timeout on test_token_user_post_new_source

dannygoldstein:
Sometimes this test fails on the CI because because it can take a bit longer than 2s for the above POST to go through. This PR just increases that timeout a bit so that the test can pass


acrellin:
Closing in favor of https://github.com/skyportal/skyportal/pull/1322 
This has been up to 25 seconds locally for me

PR #1343, opened 2020-11-12
track followup request id and assignment id in Spectrum object when posted through handler 

dannygoldstein:
This fixes a small bug where spectra were being shared with correct groups via followup_request_ids, but the followup_request_ids themselves were not saved to the Spectrum objects through Spectrum.POST. Similar for Spectrum.assignment_id


kmshin1397 (approved):

dmitryduev (approved):
👍

PR #1348, opened 2020-11-12
Refactor candidates table to use 3-col index (obj_id, filter_id, passed_at)

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1267

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -589,7 +589,7 @@ class CandidateHandler(BaseHandler):
     def post(self):
         """
         ---
-        description: Create a new candidate.
+        description: Create new candidate(s) (one per filter).
         requestBody:
           content:
             application/json:
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -715,7 +715,7 @@ class CandidateHandler(BaseHandler):
         """
         c = Candidate.query.get(candidate_id)
         if not (
-            self.associated_user_object.is_admin
+            self.associated_user_object.is_system_admin
             or c.uploader_id == self.associated_user_object.id
         ):
             return self.error("Insufficient permissions.")
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -685,6 +685,7 @@ class CandidateHandler(BaseHandler):
                 filter=filter,
                 passing_alert_id=passing_alert_id,
                 passed_at=passed_at,
+                uploader_id=self.associated_user_object.id,
             )
             for filter in filters
         ]
@@ -695,7 +696,7 @@ class CandidateHandler(BaseHandler):
 
         return self.success(data={"ids": [c.id for c in candidates]})
 
-    @permissions(["Manage sources"])
+    @auth_or_token
     def delete(self, candidate_id):
         """
         ---
@@ -713,6 +714,11 @@ class CandidateHandler(BaseHandler):
                 schema: Success
         """
         c = Candidate.query.get(candidate_id)
+        if not (
+            self.associated_user_object.is_admin
+            or c.uploader_id == self.associated_user_object.id
+        ):
+            return self.error("Insufficient permissions.")
         DBSession().delete(c)
         DBSession().commit()
 
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -626,9 +626,11 @@ class CandidateHandler(BaseHandler):
                         data:
                           type: object
                           properties:
-                            id:
-                              type: string
-                              description: New candidate ID
+                            ids:
+                              type: array
+                              items:
+                                type: integer
+                              description: List of new candidate IDs
         """
         data = self.get_json()
         obj_already_exists = Obj.query.get(data["id"]) is not None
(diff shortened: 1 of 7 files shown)


acrellin:
Failures unrelated

dannygoldstein:
Needs a migration

skyportal/models.py
```diff
@@ -723,21 +731,62 @@ class Filter(Base):
...
+        "Filter",
+        foreign_keys=[filter_id],
+        back_populates="candidates",
+        doc="The filter that an Obj passed",
+    )
+    passed_at = sa.Column(
+        sa.DateTime,
+        nullable=False,
+        index=True,
+        doc="ISO UTC time when the Candidate passed the Filter last time.",
```
dannygoldstein, line 773:
```suggestion
        doc="ISO UTC time when the Candidate passed the Filter.",
```

skyportal/models.py
```diff
@@ -723,21 +731,62 @@ class Filter(Base):
...
+        sa.ForeignKey("filters.id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+        doc="ID of the filter the Obj passed",
+    )
+    filter = relationship(
+        "Filter",
+        foreign_keys=[filter_id],
+        back_populates="candidates",
+        doc="The filter that an Obj passed",
```
dannygoldstein, line 767:
```suggestion
        doc="The filter that the Candidate passed",
```

skyportal/models.py
```diff
@@ -723,21 +731,62 @@ class Filter(Base):
...
+        "Obj",
+        foreign_keys=[obj_id],
+        back_populates="candidates",
+        doc="The Obj that passed a filter",
+    )
+    filter_id = sa.Column(
+        sa.ForeignKey("filters.id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+        doc="ID of the filter the Obj passed",
```
dannygoldstein, line 761:
```suggestion
        doc="ID of the filter the Candidate passed",
```

dannygoldstein (commented):
Do we need to update the `head` method of the `CandidateHandler` here? cc @dmitryduev

dmitryduev (changes requested):
LGTM! @dannygoldstein I don't think the HEAD method needs any changes -- it just says whether an Obj has been posted as a candidate to any filter the user has access to. I think what requires a change is the PUT method, as it is now essentially unusable, so I'd simply remove it, and add a DELETE method instead. The workflow when dealing with the candidates would be then: - create new - if unhappy, delete by id (if accessible) and post a new one. @dannygoldstein @acrellin agreed?

acrellin:
@dannygoldstein @dmitryduev all comments/requests addressed. Ready for re-review

acrellin:
Migration script runs successfully locally w/ demo data

skyportal/handlers/api/candidate.py
```diff
@@ -677,117 +679,45 @@ def post(self):
...
-            .query(Candidate)
-            .filter(Candidate.obj_id == obj_id, Candidate.filter_id.in_(filter_ids))
-            .all()
-        )
-
-        for candidate in candidates:
-            candidate.passed_at = passed_at
-            candidate.passing_alert_id = passing_alert_id
-
+        c = Candidate.query.get(candidate_id)
```
dannygoldstein, line 715: Deletion should be handled with care and should follow the ownership model - super admins and candidate creators should be the only ones allowed to delete something. This implementation makes it possible for anyone with the manage sources ACL to delete a candidate - even if they are not in the group that a candidate belongs to !

skyportal/tests/api/test_candidates.py
```diff
@@ -42,62 +42,86 @@ def test_token_user_retrieving_candidate_with_phot(view_only_token, public_candi
     assert all(k in data["data"] for k in ["ra", "dec", "redshift", "dm", "photometry"])
 
 
-def test_token_user_update_candidate(
-    manage_sources_token, public_candidate, public_filter
+def test_token_user_post_delete_new_candidate(
```
dannygoldstein, line 45: This test as written should fail - see comment above. Deletion of data should be handled very carefully - only the person who created a database entry for a candidate or a system admin should be able to delete a row from the database.

skyportal/tests/api/test_candidates.py
```diff
@@ -109,30 +133,72 @@ def test_token_user_post_new_candidate(
...
         data={
-            "id": candidate_id,
+            "id": obj_id,
             "ra": 234.22,
             "dec": -22.33,
             "redshift": 3,
             "transient": False,
             "ra_dis": 2.3,
+            "filter_ids": [public_filter.id],
             "passed_at": str(datetime.datetime.utcnow()),
```
dannygoldstein, line 153: I think filter_ids should be a scalar - filters all run separately and it will be impossible for any two filters to pass the exact same candidate at the same time.

dannygoldstein (changes requested):

skyportal/handlers/api/candidate.py
```diff
@@ -626,9 +626,11 @@ def post(self):
...
                           type: object
                           properties:
-                            id:
-                              type: string
-                              description: New candidate ID
+                            ids:
+                              type: array
+                              items:
+                                type: integer
+                              description: List of new candidate IDs
```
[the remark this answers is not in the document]
acrellin: Done

skyportal/handlers/api/candidate.py
acrellin: Done

skyportal/tests/api/test_candidates.py
acrellin: Not going to change this here -- do you want me to file an issue?

acrellin:
@dannygoldstein all comments addressed and ready for re-review

skyportal/tests/api/test_candidates.py
dannygoldstein: Yes please file an issue

dannygoldstein (approved):

skyportal/tests/api/test_candidates.py
dmitryduev: Makes sense, but I'd request to leave this as is -- the it's implemented now saves (potentially) a lot of calls on the K side of F if an object passes multiple filters.

acrellin: Okay, will _not_ file an issue, then :wink:

acrellin:
@dmitryduev ready for review when you have time. Thanks!

dmitryduev (approved):
👍

PR #1350, opened 2020-11-13
Make SourceOffsetHandler asynchronous

### LINKED ISSUE #1349: Starlist queries trigger 502s
On the observing run page a separate starlist query is issued for each observing run assignment. Each of these queries takes several seconds. With 8 web workers on fritz.science, the web workers can get totally blocked when more than 8 queries are issued simultaneously. Somehow (I am not totally sure how) this causes the website to crash and send back 502s.

dannygoldstein:
This PR makes the compute-intensive part of the source offsets query asynchronous. This should vastly improve site stability and fix those intermittent 502s we have been seeing. 

Fixes #1349 (hopefully).

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -930,13 +930,16 @@ class SourceOffsetsHandler(BaseHandler):
             use_ztfref=use_ztfref,
         )
 
-        (
-            starlist_info,
-            query_string,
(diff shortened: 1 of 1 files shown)


profjsb (approved):
Good. Looks like it works (or at least doesn't break things)

stefanv (approved):

PR #1354, opened 2020-11-13
Refactor User Management page to use MUI DataTable

### LINKED ISSUE #1216: Refactor User Management to use MUI data table
Currently, the user management page loads all users onto a single page instead of paginating. This is causing test failures on the page due to long page load times as all users are loaded into the table on the page. 

The pagination that comes with MUI data table would help a lot with the test robustness and UX when using the page. Also would enable access to features like searching for users, etc.

kmshin1397:
usermanage

In this PR:

- "Manage Users" table now uses MUI DataTable with server-side pagination and filtering
- "Pending Invitations" table also now uses MUI DataTable with server-side pagination and filtering
- "Pending Invitations" table and bulk invite form were factored out into a separate `UserInvitations.jsx` file.
- The `user` and `invitations` GET handlers were modified to also return a `totalMatches` value (for the server-side pagination), as well as supporting various query parameters for filtering, i.e. a `username` parameter for the GET users.
- Fix frontend `test_user_management.py` tests to first filter for the `user` fixture relevant to each test (otherwise, the just-provisioned user will likely not show up on the default loaded user management page).
- Added some tests for testing the GET multiple users handler, including the filtering parameters I added

Closes #1216


kmshin1397:
Test failures are real - looking into them now

dannygoldstein (approved):

acrellin (approved):

PR #1363, opened 2020-11-13
Add savedBefore, savedAfter, and saveSummary query parameters to SourceHandler.get

dannygoldstein:
This PR adds three new options to `SourceHandler.get`:

          - in: query
            name: savedBefore
            nullable: true
            schema:
              type: string
            description: |
              Only return sources that were saved before this UTC datetime.
          - in: query
            name: savedAfter
            nullable: true
            schema:
              type: string
            description: |
              Only return sources that were saved after this UTC datetime.
          - in: query
            name: saveSummary
            nullable: true
            schema:
              type: boolean
            description: |
              Boolean indicating whether to only return the source save
              information in the response (defaults to false). The schema
              of this summary information is:
              ```
                  {
                    "group_id": 2,
                    "created_at": "2020-11-13T22:11:25.910271",
                    "saved_by_id": 1,
                    "saved_at": "2020-11-13T22:11:25.910271",
                    "requested": false,
                    "unsaved_at": null,
                    "modified": "2020-11-13T22:11:25.910271",
                    "obj_id": "16fil",
                    "active": true,
                    "unsaved_by_id": null
                  }
              ```

Fixes https://github.com/fritz-marshal/fritz/issues/99

@cfremling

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -10,6 +10,7 @@ from dateutil.parser import isoparse
 from sqlalchemy.orm import joinedload
 from sqlalchemy import func, or_
 import arrow
+from marshmallow import Schema, fields
 from marshmallow.exceptions import ValidationError
 import functools
 import healpix_alchemy as ha
@@ -235,6 +236,28 @@ class SourceHandler(BaseHandler):
             description: |
               Boolean indicating whether to only include requested/pending saves.
               Defaults to false.
+          - in: query
+            name: savedBefore
+            nullable: true
+            schema:
+              type: string
+            description: |
+              Only return sources that were saved before this UTC datetime.
+          - in: query
+            name: savedAfter
+            nullable: true
+            schema:
+              type: string
+            description: |
+              Only return sources that were saved after this UTC datetime.
+          - in: query
+            name: saveSummary
+            nullable: true
+            schema:
+              type: boolean
+            description: |
+              Boolean indicating whether to only return the source save
+              information in the response (defaults to false).
           responses:
             200:
               content:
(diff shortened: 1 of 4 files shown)


dannygoldstein:
Ah, this needs a migration...

dannygoldstein:
Migration added

skyportal/tests/api/test_sources.py
```diff
@@ -346,3 +346,70 @@ def test_source_notifications_unauthorized(
...
+    assert len(sources) == 1
+    source = sources[0]
+
+    assert source['obj_id'] == public_source.id
+    assert source['group_id'] == public_group.id
+
+    # check the datetime formatting is properly validated
+    status, data = api(
+        "GET",
+        f"sources?saveSummary=true&group_ids={public_group.id}&savedBefore=2020-104-01T00:00:01.2412",
```
dmitryduev, line 412: 104?

dmitryduev (approved):
LGTM!

dannygoldstein:
Yep that’s what’s being tested - this should (and does) cause the test to fail

dannygoldstein:
Er, the query to fail - the test passes

skyportal/handlers/api/source.py
```diff
@@ -235,6 +236,43 @@ def get(self, obj_id=None):
...
+                    "created_at": "2020-11-13T22:11:25.910271",
+                    "saved_by_id": 1,
+                    "saved_at": "2020-11-13T22:11:25.910271",
+                    "requested": false,
+                    "unsaved_at": null,
+                    "modified": "2020-11-13T22:11:25.910271",
+                    "obj_id": "16fil",
+                    "active": true,
+                    "unsaved_by_id": null
+                  }
```
acrellin, line 274: Indicate that the response data will include a list of these dicts under `response_json['data']['sources']`

skyportal/tests/api/test_sources.py
```diff
@@ -346,3 +346,70 @@ def test_source_notifications_unauthorized(
     assert data["message"].startswith(
         "Twilio Communication SMS API authorization error"
     )
+
+
+def test_token_user_source_summary(
```
acrellin, line 351: Check that the data we expect to not be present (obj data) are indeed not present

acrellin (changes requested):

skyportal/handlers/api/source.py
dannygoldstein: done

skyportal/tests/api/test_sources.py
dannygoldstein: done

skyportal/tests/api/test_sources.py
```diff
@@ -346,3 +346,70 @@ def test_source_notifications_unauthorized(
...
+        f"sources?saveSummary=true&group_ids={public_group.id}",
+        token=view_only_token_two_groups,
+    )
+    assert status == 200
+    assert "sources" in data['data']
+    sources = data['data']['sources']
+
+    assert len(sources) == 1
+    source = sources[0]
+    assert 'obj' not in source
```
acrellin, line 368: Obj wouldn't be present anyways, right? Each dict in `sources` would just have all the obj data (ra, dec, mag, magerr, etc)

dannygoldstein: nice catch, thanks. fixed

acrellin (approved):
Looks great. Will merge when green / only unrelated failures

Issue #100, opened 2018-09-14
Investigate moveable homepage components


acrellin:

acrellin:
@kmshin1397 has implemented this

PR #1369, opened 2020-11-17
Implement server-side sorting for source tables

kmshin1397:
Adds sorting for source tables on the group sources page and the sources page (which all use SourceTable.jsx).
Partially closes [Fritz #100](https://github.com/fritz-marshal/fritz/issues/100) (There was a separate request in there about adding additional columns to the table; this PR is just sorting).
sourcesort


kmshin1397:
Hmm looks like I have some test failures to look into

acrellin (approved):
Very nice, thanks @kmshin1397

PR #1370, opened 2020-11-17
Fix source handler bug where invalid ID was not caught soon enough

acrellin:
This PR fixes a bug where an invalid obj ID was not caught soon enough and was passed to `register_source_view` and `add_ps1_thumbnail` before erroring.

Closes https://github.com/skyportal/skyportal/issues/1372

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -392,6 +392,10 @@ class SourceHandler(BaseHandler):
                 query_options.append(
                     joinedload(Obj.photometry).joinedload(Photometry.instrument)
                 )
+            s = Obj.get_if_owned_by(obj_id, self.current_user, options=query_options,)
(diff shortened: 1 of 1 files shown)


kmshin1397 (approved):

kmshin1397:
Looks good! Waiting on CI before merging

kmshin1397:
@acrellin I see a bunch of sqlalchemy invalid request errors in the tests. They seem to be real seeing as I tried rerunning both CIs and got the same behavior. Seems like this change isn't as innocuous as it seems?

acrellin:
@kmshin1397 these are also present on master; fixed in https://github.com/skyportal/skyportal/pull/1376

kmshin1397:
@acrellin Ah okay great, I'll merge this then.

PR #1374, opened 2020-11-17
Update SourceHandler.get access control

### LINKED ISSUE #1373: Use same source access control for fetching multiple sources as for single source fetches
I.e. return all sources w/ at least one photometry data point viewable to user when fetching multiple sources

acrellin:
This PR changes `SourceHandler.get` query filtering to filter on objects with at least one photometry data point viewable to user, rather than filtering on which groups a source is saved to. This accords with the current access control approach for fetching individual sources.

Closes #1373 
Closes https://github.com/fritz-marshal/fritz-beta-feedback/issues/42

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -509,7 +509,12 @@ class SourceHandler(BaseHandler):
                 .query(Obj)
                 .join(Photometry)
                 .join(Source)
-                .filter(Photometry.groups.any(Group.id.in_(user_accessible_group_ids)))
+                .filter(
+                    or_(
+                        Photometry.groups.any(Group.id.in_(user_accessible_group_ids)),
+                        Source.group_id.in_(user_accessible_group_ids),
+                    )
+                )
                 .options(query_options)
             )
         else:
@@ -518,7 +523,12 @@ class SourceHandler(BaseHandler):
                 .query(Source)
                 .join(Obj)
                 .join(Photometry)
-                .filter(Photometry.groups.any(Group.id.in_(user_accessible_group_ids)))
+                .filter(
+                    or_(
+                        Photometry.groups.any(Group.id.in_(user_accessible_group_ids)),
+                        Source.group_id.in_(user_accessible_group_ids),
+                    )
+                )
             )
 
         if sourceID:
@@ -576,7 +586,6 @@ class SourceHandler(BaseHandler):
                     if sort_order == "asc"
                     else [Obj.ra.desc().nullslast()]
                 )
-                print(sort_by, sort_order)
             elif sort_by == "dec":
                 order_by = (
                     [Obj.dec.nullslast()]
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -26,7 +26,6 @@ from ...models import (
     Source,
     Token,
     Group,
-    GroupPhotometry,
     FollowupRequest,
     ClassicalAssignment,
     ObservingRun,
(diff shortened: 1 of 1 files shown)


kmshin1397 (approved):

acrellin:
@stefanv bump

acrellin:
Hmm, this seems to have gained a few failures since pulling in recent changes. Will update and re-ping when ready for review

acrellin:
I think this needs to be an `or_` filter condition so that sources without any associated photometry can still be fetched, which is what's causing some test failures

dannygoldstein:
I think this will break the group sources page, no ?

dannygoldstein:
The group sources page uses this handler and should only display sources that are _saved_ to a user's groups. Photometry accessible to groups will water down the page with sources that haven't been explicitly saved to the sample.

I think the right way to go here is to implement this kind of checking on `Obj.get` handler, which I am working on, and to leave the logic for `Source.get` alone.

acrellin:
@dannygoldstein filtering by group ID(s) is unaffected by this change, which happens elsewhere in the handler. Please do not close other people's PRs without discussion. Make a comment, fine, but do not close someone else's PR. Don't be a disrespectful teammate.

dannygoldstein:
I closed this pull request prematurely, which was a mistake. I take full responsibility, I should have approved this PR.

dannygoldstein (approved):

dannygoldstein:
Are the failures related? Looks like some API failures related to sources.

dannygoldstein (commented):
I missed something on my second look at this, an important subtlety that is not addressed by this implementation.

We want an object to be visible to a user if any of its photometry is associated with any of a user's groups, or if it is a source saved to one of the user's groups. However, in the former case, we do not require that the obj is a source (saved to any group) in order to return it. The object's database record should be accessible to a querying user even if the object is saved to no groups, but it has some photometry that is associated with at least one of the user's groups. Based on past experience, the majority of objects viewed via the source page or API may fall into this category, so it is not an edge case.

Because there is an inner join against the source table here, only objects that are saved to at least one group as sources will ever be returned by this handler. I think it would work if the inner join were converted to an outer join.

But really, this should go in the Obj handler, which is forthcoming. The logic here is more relevant to Objs than Sources. 

This is part of the reason I originally closed this - we had a long discussion about refactoring and overhauling the obj / source/ candidate handlers at stand up the other day, after you disconnected @acrellin. It was decided after discussion with the group that a total refactor and overhaul of the source/candidate/obj handler should proceed, likely resulting in an increment of the source code major version number. 

Hopefully this provides a bit of context for why I closed the PR. On closer inspection, I still think the implementation here is not quite right, and that this should be deferred to the obj / source / candidate refactor and the obj handler that will be included in that.

acrellin:
> I missed something on my second look at this, an important subtlety that is not addressed by this implementation.
> 
> We want an object to be visible to a user if any of its photometry is associated with any of a user's groups, or if it is a source saved to one of the user's groups. However, in the former case, we do not require that the obj is a source (saved to any group) in order to return it. The object's database record should be accessible to a querying user even if the object is saved to no groups, but it has some photometry that is associated with at least one of the user's groups. Based on past experience, the majority of objects viewed via the source page or API may fall into this category, so it is not an edge case.
> 
> Because there is an inner join against the source table here, only objects that are saved to at least one group as sources will ever be returned by this handler. I think it would work if the inner join were converted to an outer join.
> 
> But really, this should go in the Obj handler, which is forthcoming. The logic here is more relevant to Objs than Sources.
> 
> This is part of the reason I originally closed this - we had a long discussion about refactoring and overhauling the obj / source/ candidate handlers at stand up the other day, after you disconnected @acrellin. It was decided after discussion with the group that a total refactor and overhaul of the source/candidate/obj handler should proceed, likely resulting in an increment of the source code major version number.
> 
> Hopefully this provides a bit of context for why I closed the PR. On closer inspection, I still think the implementation here is not quite right, and that this should be deferred to the obj / source / candidate refactor and the obj handler that will be included in that.

@dannygoldstein that makes sense, thanks for explaining your thoughts on this -- will close and defer to the upcoming larger handler/endpoint restructuring PR

PR #1377, opened 2020-11-17
Track whether a telescope is space based or ground based 

dannygoldstein:
This PR keeps track of whether a given telescope is space based or ground based. If it is space based, it does not show the telescope's airmass plot on the observability page, partially addressing https://github.com/fritz-marshal/fritz-beta-feedback/issues/29.

### DIFF
--- a/alembic/versions/2a17ca9d5438_add_fixed_location_col.py
+++ b/alembic/versions/2a17ca9d5438_add_fixed_location_col.py
@@ -1,4 +1,4 @@
-"""add space-based col
+"""add fixed-location column
 
 Revision ID: 2a17ca9d5438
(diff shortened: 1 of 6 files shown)


profjsb (changes requested):
Not to be nickpicky but...I think it's better to have the attribute be something like `fixed_location` (default True) instead of `space_based`: could imagine non-stationary observer which is not groundbased nor spacebased (ie. Sofia).

dannygoldstein:
@profjsb fixed

dannygoldstein:
Looks like there are some CI issues here ...

profjsb (approved):

PR #1378, opened 2020-11-18
Improve reliability of timezone finder 

dannygoldstein:
The current function that we use to look up timezones given positions on the earth is somewhat brittle and can return `None` in some cases (e.g., for coordinates in the Negev Desert). This caused the observability plot to fail to render for Wise Observatory.

This PR replaces the timezone finder with a less brittle function `timezone_at` -> `closest_timezone_at`, which returns the closest timezone in a 5 degree radius. 

This is a somewhat more expensive call, but can be sped up by ~100x using numba JIT compilation. So I have added numba as a dependency.


profjsb (approved):
Should update `timezonefinder` to version 4.5.0...but otherwise fine.

dannygoldstein:
> Should update `timezonefinder` to version 4.5.0...but otherwise fine.

done

dmitryduev (approved):
I love `numba`, so would happily merge this!

PR #1379, opened 2020-11-18
Automatically refresh enums on app start and bump sncosmo to 2.3.0

### LINKED ISSUE #1308: Add swift bandpasses

dannygoldstein:
Fixes #1308.

`sncosmo` 2.3.0 adds the _Swift_ UV Optical Telescope filterset, which is widely used in ZTF. The current `enum` type of allowed bandpasses did not contain these filters. 

I added a hook that, on app deploy, refreshes the database enum types to add any new values to the type, similar to what we do now for refreshing tables. 

This ensures that enum types will always exist, match the docs, and persist across migrations.


dmitryduev (approved):
LGTM

kmshin1397 (approved):

PR #1381, opened 2020-11-18
Fix relationship loading in source handler

acrellin:
Fixes source API test failures
Unsure why these would suddenly be failing, as the relevant code hasn't changed recently (related package version upgrades, perhaps)


acrellin:
Single failing test (skyportal/tests/frontend/test_assignments.py::test_submit_and_delete_new_assignment) is unrelated and will be addressed elsewhere (target out of bounds selenium error)

PR #1382, opened 2020-11-18
Fix assignment front-end test

acrellin:


dannygoldstein:
@acrellin Can you add a bit of info to what the issue is that this fixes

acrellin:
> @acrellin Can you add a bit of info to what the issue is that this fixes

@dannygoldstein https://travis-ci.com/github/skyportal/skyportal/builds/202452106#L3869 
`click_xpath` is much more robust, and using our own `data-testid` value is safer than an auto-generated MUI input ID

acrellin:
One flaky failure each per GA and Travis, though different tests. Will address elsewhere. The test in question is passing on both builds

PR #1384, opened 2020-11-18
Fix source API save_summary option

kmshin1397:


dannygoldstein:
what happened here?

acrellin (approved):
Unrelated to your PR then, it appears. Looks good, thanks. Will merge when green

dannygoldstein:
> Unrelated to your PR then, it appears. Looks good, thanks. Will merge when green

Which PR?

acrellin:
> > Unrelated to your PR then, it appears. Looks good, thanks. Will merge when green
> 
> Which PR?

@dannygoldstein the source table filtering PR that was just merged. Looks like your source handler changes might have been in progress at the same time.

kmshin1397:
@dannygoldstein @acrellin It technically was my source table sorting PR that broke this - I needed to call the `get_query_results` function (that's shared between candidate.py and source.py) to maintain proper order_by for sorting queries and forgot about the `save_summary` option, which is supposed to provide a lighter response and doesn't want all the extra logic in the `get_query_results` function. So yeah, not directly related to the handler refactor Danny is working on.

acrellin:
@kmshin1397 could https://travis-ci.com/github/skyportal/skyportal/builds/202518833#L3965 be related?

kmshin1397:
@acrellin Hmm no that shouldn't be related as the frontend will always hit the if clause before the change in this PR since the table will define a page number. But I should still look into that test failing - my guess is that it's more like the usual frontend test flakiness.

PR #1390, opened 2020-11-19
Limit source assignment to upcoming observing runs

kmshin1397:
Addresses [Beta Feedback Issue #39](https://github.com/fritz-marshal/fritz-beta-feedback/issues/39).

On the source page, the options for "Assign Target to Observing Run" are now limited to observing runs whose end (the next sunrise after the run's calendar date) is after the current time.

### DIFF
--- a/skyportal/handlers/api/observingrun.py
+++ b/skyportal/handlers/api/observingrun.py
@@ -162,7 +162,14 @@ class ObservingRunHandler(BaseHandler):
             return self.success(data=data)
 
         runs = ObservingRun.query.order_by(ObservingRun.calendar_date.asc()).all()
-        return self.success(data=runs)
+        runs_list = []
+        for run in runs:
+            runs_list.append(run.to_dict())
+            runs_list[-1]["run_end_utc"] = run.instrument.telescope.next_sunrise(
+                run.calendar_noon
+            ).isot
+
+        return self.success(data=runs_list)
 
     @permissions(["Upload data"])
     def put(self, run_id):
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/observingrun.py
```diff
@@ -162,7 +162,14 @@ def get(self, run_id=None):
             return self.success(data=data)
 
         runs = ObservingRun.query.order_by(ObservingRun.calendar_date.asc()).all()
-        return self.success(data=runs)
+        runs_list = []
+        for run in runs:
+            runs_list.append(run.to_dict())
+            runs_list[-1]["run_end_utc"] = run.instrument.telescope.next_sunrise(
```
dannygoldstein, line 168: This isn't when the run ends, but the next sunrise on the telescope. For the sunrise of the run, I think you want `run.sunrise`, but check to make sure

dannygoldstein (changes requested):

skyportal/handlers/api/observingrun.py
kmshin1397: @dannygoldstein I don't see a `sunrise` field defined on ObservingRun in models.py. There is only `rise_time()` which only reports rise times for specific targets passed in to the function. I can add a `sunrise` property, but what would be the best way to compute that? So I don't want to go through the telescope instance?

dannygoldstein: I take this back - this is fine. Sorry about this. IDK what I was thinking here ...

dannygoldstein (approved):

dmitryduev (approved):
👍

PR #1394, opened 2020-11-19
Use colon instead of hms, dms as hexadecimal coordinate separators on source page

dannygoldstein:
Fixes https://github.com/fritz-marshal/fritz-beta-feedback/issues/49


profjsb (approved):
Good.

PR #1396, opened 2020-11-20
Fix permissions on spectroscopy plot 

dannygoldstein:
The spectroscopy bokeh plot was not doing any ACL checking and was showing all spectra associated with an object rather than just the ones that were associated with any of the requesting user's groups. This PR fixes this by adding in an ACL check to the `spectroscopy_plot` function.

### DIFF
--- a/skyportal/handlers/api/internal/plot.py
+++ b/skyportal/handlers/api/internal/plot.py
@@ -35,7 +35,7 @@ class PlotSpectroscopyHandler(BaseHandler):
     def get(self, obj_id):
         spec_id = self.get_query_argument("spectrumID", None)
         docs_json, render_items, custom_model_js = plot.spectroscopy_plot(
-            obj_id, spec_id
+            obj_id, self.associated_user_object, spec_id
         )
         if docs_json is None:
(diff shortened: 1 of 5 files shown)


kmshin1397 (approved):
I noticed that the GET spectrum handler seems to just check that the Obj for the spectrum is accessible to the user and doesn't check the GroupSpectrum records. I think the check here is more correct? Would it make sense to change the GET spectrum handler to use the same ACL check as well?

dannygoldstein:
@kmshin1397 you are absolutely right. that's a critical one - will push up a fix

dannygoldstein:
@kmshin1397 fixed

kmshin1397 (approved):

kmshin1397:
@dannygoldstein I think some of the test failures may be related - I see some tests timing out while looking for the bokeh plot.

dannygoldstein:
@kmshin1397  nice catch, again, thanks. - hopefully fixed now.

dannygoldstein:
failures unrelated

dmitryduev (approved):
LGTM!

Issue #555, opened 2020-07-06
Upgrade react-hook-form to v6


acrellin:
Has breaking API changes, especially in `Controller` API

PR #1400, opened 2020-11-20
Upgrade react-hook-form to v6

acrellin:
Notable changes:
- `Controller` API:
  - Simplest usage of `as` prop still supported, but any components that use `onChange`, `value`, etc, now need to use new `render` prop
  - `data-testid` (& other attributes) no longer automatically passed in to child component specified in `render` (can do props spreading or simply move to child component explicitly

Closes https://github.com/skyportal/skyportal/issues/555


acrellin:
Just need to sort out the xpath issues

acrellin:
From the test failures, it looks like I need to update our usage of `validate`, too

acrellin:
It's actually our usage of `Checkbox` components with the `Controller` that needs updating -- I'm working my way through the many instances

stefanv (approved):

PR #1401, opened 2020-11-20
Fix bug where unsaved groups were showing up in saved to groups list on scanning page

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1398


dmitryduev (approved):
Good catch!

Issue #924, opened 2020-09-16
NewGroupUserForm and GroupUserHandler.post should use user ID instead of username


acrellin:
In the request data, that is (not what is displayed to users on front-end)

PR #1404, opened 2020-11-20
Refactor new group user & invite new user to group components; update API to use user ID instead of username

acrellin:
This PR simplifies the logic of the `NewGroupUserForm` component by splitting the disparate cases it covered into two separate components -- one for adding existing users to the group, another for inviting new users to the site (& adding them to the group). This component was previously rather confusing both from a UX & DX perspective. The API is also updated to always use user ID instead of username for adding existing users to groups.

Closes https://github.com/skyportal/skyportal/issues/924

### DIFF
--- a/static/js/components/NewGroupUserForm.jsx
+++ b/static/js/components/NewGroupUserForm.jsx
@@ -38,8 +38,7 @@ const NewGroupUserForm = ({ group_id }) => {
   }, [dispatch, allUsers]);
 
   const handleClickSubmit = async () => {
-    let result = null;
-    result = await dispatch(
+    const result = await dispatch(
       groupsActions.addGroupUser({
         userID: formState.userID,
         admin: formState.admin,
--- a/static/js/components/NewGroupUserForm.jsx
+++ b/static/js/components/NewGroupUserForm.jsx
@@ -71,7 +71,7 @@ const NewGroupUserForm = ({ group_id }) => {
         Add an existing user to this group
       </Typography>
       <Autocomplete
-        id="newUserEmail"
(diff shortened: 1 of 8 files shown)


acrellin:
Travis failure unrelated (`test_source_count_widget`)

static/js/components/NewGroupUserForm.jsx
```diff
@@ -46,54 +37,19 @@ const NewGroupUserForm = ({ group_id }) => {
...
-  };
-
-  const handleClickSubmit = (event) => {
-    event.preventDefault();
-    if (invitationsEnabled && formState.invitingNewUser) {
-      // If user clicks confirm, `submitAndResetForm` will be called
-      setConfirmDialogOpen(true);
-    } else {
-      submitAndResetForm();
+  const handleClickSubmit = async () => {
```
kmshin1397, line 40: I think it would good to add a validation check in the frontend form here that checks for username !== null. Otherwise, the error is caught thrown from the backhand and the end user sees a "UserID is missing" error which may be confusing since the form refers to usernames.

kmshin1397 (changes requested):
One small suggestion, but otherwise looks good! I agree that separating out the two forms is a good idea for both dx and ux.

stefanv (approved):

static/js/components/NewGroupUserForm.jsx
acrellin: :+1: Done

kmshin1397 (approved):

kmshin1397:
Thanks Ari! Will merge when green

PR #1405, opened 2020-11-20
Add {User,Token}.accessible_streams and use where applicable in handlers

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1158


dmitryduev (approved):
👍

PR #1409, opened 2020-11-23
Group page and group API updates

### LINKED ISSUE #1391: Refactor the members list on the group page to use paginated MUI DataTable

kmshin1397:
In this PR:

- Refactor the "Members" list on Group page to a MUI DataTable (not server-side) - closes #1391 
  - Displays username and admin columns by default, and the actions column (i.e. toggle admin button) for admins. A `name` column is displayed if at least one user has a first/last name.
  - Supports sorting/filtering/search etc.
  - Help icons explaining that admin status is group admin, but also sys admin and/or "manage groups" ACL - so even if you take away admin for a group from a super admin, they will continue to have the admin tag.
- Clean up GET response for group handler
  - The current GET response includes a `users` list which just gives `(ID, username)` info, and a separate `group_users` list which is used just for group admin checks. I combined all group user info into a single `users` property (and tacked on other potentially interesting info i.e. first and last name that wasn't being returned at all)
  - This means the front-end doesn't have to keep cross-matching the two lists to check admin status for a user
- Cleaned up the groups page tests
  - Use recent, more robust versions of baselayer test utils i.e. `scroll_parent=True` for clicking on list items and using `click_xpath` where possible
- Noticed that while there was a scrollbar on the "All groups" list on the Groups page, the maximum height was not properly applied, so the list extended forever - fixed this to have reasonable box size for the all groups list.

members


acrellin (approved):

dmitryduev (approved):
LGTM! @kmshin1397 could you please resolve the merge conflict -- will merge after that?

dmitryduev:
Failure seems unrelated.

PR #1421, opened 2020-11-25
Make default date range on scanning page past 24 hours

acrellin:
utc_pickers_scanning_page

Closes https://github.com/fritz-marshal/fritz-beta-feedback/issues/58

### DIFF
--- a/static/js/components/FilterCandidateList.jsx
+++ b/static/js/components/FilterCandidateList.jsx
@@ -137,6 +137,7 @@ const FilterCandidateList = ({
                   format="YYYY/MM/DD HH:mm"
                   ampm={false}
                   showTodayButton={false}
+                  data-testid="startDatePicker"
                 />
               )}
               rules={{ validate: validateDates }}
@@ -153,6 +154,7 @@ const FilterCandidateList = ({
                   format="YYYY/MM/DD HH:mm"
                   ampm={false}
                   showTodayButton={false}
+                  data-testid="endDatePicker"
                 />
               )}
               rules={{ validate: validateDates }}
--- a/static/js/components/FilterCandidateList.jsx
+++ b/static/js/components/FilterCandidateList.jsx
@@ -51,13 +51,20 @@ const FilterCandidateList = ({
 }) => {
   const classes = useStyles();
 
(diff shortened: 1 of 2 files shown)


acrellin:
Will update this shortly such that users work in UTC times (need work around mui pickers always displaying times in local browser time)

acrellin:
After going down a rabbit hole that ended up being a dead end, I landed on a super simple way of using UTC times for the pickers.

kmshin1397 (approved):

static/js/components/FilterCandidateList.jsx
```diff
@@ -117,30 +129,32 @@ const FilterCandidateList = ({
...
+              render={({ onChange, value }) => (
                 <KeyboardDateTimePicker
-                  value={formState.startDate}
-                  label="Start (browser local time)"
+                  value={dayjs.utc(value)}
+                  onChange={(e, date) => onChange(dayjs.utc(date))}
+                  label="Start (UTC)"
                   format="YYYY/MM/DD HH:mm"
                   ampm={false}
                   showTodayButton
```
dannygoldstein, line 139: The today button still goes to today in browser local time, rather than UTC. E.g.,

Right now its 13:26 PT, if I click on "Today" on the picker, it sets the time to 13:26 UTC, but should be 21:26 UT.

dannygoldstein (commented):
Beautiful stuff. Just a quick thing about the today button

static/js/components/FilterCandidateList.jsx
acrellin: Yes, unfortunately the datepicker always uses the browser local time, and there is no callback exposed for changing this. The best we can do is disable to the "Today" button

acrellin: @dannygoldstein that change is pushed

dannygoldstein (approved):

dannygoldstein:
self merge on ci

acrellin:
Looks like one of the CI failures might actually be related to this change. Will push up a fix

acrellin:
All test failures are unrelated

PR #1425, opened 2020-11-25
remove magsystems from enum types

dannygoldstein:
Because they are not an Enum type used by postgres. Only by the webserver.


dmitryduev (approved):
👍

Issue #957, opened 2020-09-20
ra/dec strings in comments are attempted to be rendered as emojis


dmitryduev:
E.g.:
![image](https://user-images.githubusercontent.com/7557205/93705441-322b5c00-fad2-11ea-996f-256677e77073.png)

A simple solution would be to ignore comments posted by provisioned_admin.

mcoughlin:
@kmshin1397 What do you think of something like:
const emojiSupport = (text) => {
    text.value.replace(/:\w+:/gi,
                       emoji.getUnicode(name).includes("undefined") ? name : (name) => emoji.getUnicode(name));
I can't quite make it work, but I think that's simply my lack of java skillz. But some kind of tertiary operator to do the check?

dmitryduev:
up

PR #1435, opened 2020-11-29
Fix emoji parsing for non-emoji colons in comments

kmshin1397:
Address #957 

Comment substrings matching the emoji RegEx but not mapping to a legitimate emoji are now just let through.

@mcoughlin I think the main problem with what you were trying was that the "undefined" you were seeing printed was actually the javascript undefined property instead of a string that said "undefined" so your check wasn't working.


dmitryduev (approved):
Great, thanks for the fix!

mcoughlin (approved):
I knew it was a one liner, thanks for getting it done :)

PR #1439, opened 2020-11-30
fix share with... bug in upload spectrum

dannygoldstein:
Fixes https://github.com/fritz-marshal/fritz-beta-feedback/issues/68


dmitryduev (approved):
👍

PR #1454, opened 2020-12-01
Refactor classifications form to use react-jsonschema-form and add group IDs 

### LINKED ISSUE #1318: Frontend Classifications form doesn't expose group IDs parameter
POST API for classifications has group_ids parameter for who gets to see the classification, but the frontend form does not.

kmshin1397:
classifications

Closes #1318

### DIFF
--- a/static/js/components/ClassificationForm.jsx
+++ b/static/js/components/ClassificationForm.jsx
@@ -1,52 +1,55 @@
-import React, { useReducer } from "react";
-import { useDispatch } from "react-redux";
+import React from "react";
+import { useDispatch, useSelector } from "react-redux";
 import PropTypes from "prop-types";
-import Select from "@material-ui/core/Select";
-import InputLabel from "@material-ui/core/InputLabel";
 import TextField from "@material-ui/core/TextField";
-import Button from "@material-ui/core/Button";
 import MenuItem from "@material-ui/core/MenuItem";
-import Autocomplete from "@material-ui/lab/Autocomplete";
+import Select from "@material-ui/core/Select";
+import Input from "@material-ui/core/Input";
+import InputLabel from "@material-ui/core/InputLabel";
+import Chip from "@material-ui/core/Chip";
 import { makeStyles } from "@material-ui/core/styles";
+import Form from "@rjsf/material-ui";
+
 import { showNotification } from "baselayer/components/Notifications";
 import * as Actions from "../ducks/source";
 
 const useStyles = makeStyles(() => ({
-  taxonomySelect: {
-    maxWidth: "100%",
-    fullWidth: "true",
+  chips: {
     display: "flex",
-    wrap: "nowrap",
+    flexWrap: "wrap",
+  },
+  chip: {
+    margin: 2,
   },
 }));
 
 // For each node in the hierarchy tree, add its full path from root
-// to the node_paths list
-const add_node_paths = (node_paths, hierarchy, prefix_path = []) => {
-  const this_node_path = [...prefix_path];
+// to the nodePaths list
+const addNodePaths = (nodePaths, hierarchy, prefix_path = []) => {
+  const thisNodePath = [...prefix_path];
 
   if (
     hierarchy.class !== undefined &&
     hierarchy.class !== "Time-domain Source"
   ) {
-    this_node_path.push(hierarchy.class);
-    node_paths.push(this_node_path);
+    thisNodePath.push(hierarchy.class);
+    nodePaths.push(thisNodePath);
   }
 
   hierarchy.subclasses?.forEach((item) => {
     if (typeof item === "object") {
-      add_node_paths(node_paths, item, this_node_path);
+      addNodePaths(nodePaths, item, thisNodePath);
     }
   });
 };
 
 // For each class in the hierarchy, return its name
 // as well as the path from the root of hierarchy to that class
-const allowed_classes = (hierarchy) => {
-  const class_paths = [];
-  add_node_paths(class_paths, hierarchy);
+const allowedClasses = (hierarchy) => {
+  const classPaths = [];
+  addNodePaths(classPaths, hierarchy);
 
-  const classes = class_paths.map((path) => ({
+  const classes = classPaths.map((path) => ({
     class: path.pop(),
     context: path.reverse(),
   }));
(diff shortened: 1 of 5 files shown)


static/js/components/ClassificationForm.jsx
```diff
@@ -56,197 +59,229 @@ const allowed_classes = (hierarchy) => {
...
-          ),
-        };
-      default:
-        return {
-          ...state,
-          [action.name]: action.value,
-        };
+  const handleSubmit = async ({ formData }) => {
+    // Get the classification without the context
+    const classification = formData.classification.split(" <> ")[0];
```
stefanv, line 82: Why not get the form to send the value you are interested in instead?

stefanv (commented):
@kmshin1397 What is the motivation from using jsonschema-form here, instead of using react-hook-form?  I don't see a big difference in the code readability.  I think @dannygoldstein introduced jsonschema-form for cases where forms had to be dynamically generated, but where schemas were calculated on the backend?

/cc @acrellin

static/js/components/ClassificationForm.jsx
kmshin1397: I was getting validation errors related to duplicate values in the schema enum because there are classifications with duplicate names (but different contexts) so this was my workaround.

stefanv: OK, so this is slightly bizarre: why is freeform text allowed for classification? 

https://skyportal.io/docs/api.html#/paths/~1api~1classification/post

Should we not validate that classifications are from an existing taxonomy?

kmshin1397:
> @kmshin1397 What is the motivation from using jsonschema-form here, instead of using react-hook-form? I don't see a big difference in the code readability. I think @dannygoldstein introduced jsonschema-form for cases where forms had to be dynamically generated, but where schemas were calculated on the backend?
> 
> /cc @acrellin

I'm not particularly set on using one over the other, to be honest. I do think the jsonschema form is much simpler to set up and read with straightforward forms (when essentially you just have to read a JSON segment instead of an HTML segment) but this form did end up require more custom widgets than I anticipated so the readability was not as big a difference than I expected. And as Danny has said, I think it is definitely easier to do dynamically generated forms like with the followup requests. Doesn't necessarily have to be schemas computed on the backend (for example, I dynamically change form fields on the scanning page filtering form based on the annotation types selected).

But yea, in this particular case, I wouldn't lean one way or the other.

stefanv:
> > @kmshin1397 What is the motivation from using jsonschema-form here, instead of using react-hook-form? I don't see a big difference in the code readability. I think @dannygoldstein introduced jsonschema-form for cases where forms had to be dynamically generated, but where schemas were calculated on the backend?
> > /cc @acrellin
> 
> I'm not particularly set on using one over the other, to be honest. I do think the jsonschema form is much simpler to set up and read with straightforward forms (when essentially you just have to read a JSON segment instead of an HTML segment) but this form did end up require more custom widgets than I anticipated so the readability was not as big a difference than I expected. And as Danny has said, I think it is definitely easier to do dynamically generated forms like with the followup requests. Doesn't necessarily have to be schemas computed on the backend (for example, I dynamically change form fields on the scanning page filtering form based on the annotation types selected).
> 
> But yea, in this particular case, I wouldn't lean one way or the other.

In general, we try to drive towards [There should be one—and preferably only one—obvious way to do it](https://www.python.org/dev/peps/pep-0020/).  So, what I would like to learn from this discussion with @dannygoldstein and @acrellin who have both worked on forms, is whether we want to ultimately steer all our forms in this direction, or whether keeping react-hook-forms in play makes sense.

static/js/components/ClassificationForm.jsx
kmshin1397: That's more of an issue with the docs I think. There is a check for allowed classifications in the handler code (and the frontend form only allows selection from the allowed classes)

kmshin1397: https://github.com/skyportal/skyportal/blob/fe49f4b99fadefb7be08769dab7f63a5a8640ba9/skyportal/handlers/api/classification.py#L123-L136

acrellin:
> > > @kmshin1397 What is the motivation from using jsonschema-form here, instead of using react-hook-form? I don't see a big difference in the code readability. I think @dannygoldstein introduced jsonschema-form for cases where forms had to be dynamically generated, but where schemas were calculated on the backend?
> > > /cc @acrellin
> > 
> > 
> > I'm not particularly set on using one over the other, to be honest. I do think the jsonschema form is much simpler to set up and read with straightforward forms (when essentially you just have to read a JSON segment instead of an HTML segment) but this form did end up require more custom widgets than I anticipated so the readability was not as big a difference than I expected. And as Danny has said, I think it is definitely easier to do dynamically generated forms like with the followup requests. Doesn't necessarily have to be schemas computed on the backend (for example, I dynamically change form fields on the scanning page filtering form based on the annotation types selected).
> > But yea, in this particular case, I wouldn't lean one way or the other.
> 
> In general, we try to drive towards [There should be one—and preferably only one—obvious way to do it](https://www.python.org/dev/peps/pep-0020/). So, what I would like to learn from this discussion with @dannygoldstein and @acrellin who have both worked on forms, is whether we want to ultimately steer all our forms in this direction, or whether keeping react-hook-forms in play makes sense.

I've actually never used jsonschema-form, but would agree that we should be consistent for the sake of simplicity. I guess I had been assuming that we're using react-hook-form wherever possible, and only using rjsf when that was necessitated due to dynamically-determined fields. I find RHF to be simpler, but that may just be because I'm much more familiar with it. I'm open to moving towards rjsf more regularly if there are clear advantages. Thoughts?

dannygoldstein:
I like react json schema form because it makes creating a form as simple as writing down a json schema. You write down the schema of your form data and it will make all the form components for you, using material-ui if desired. With a single schema you not only get the frontend components but you get the validation as well. With a few extra tweaks you can customize the form UI if you want, although I have never had to do that. 

I have used RJSF for some other forms on the site just because I find it easier to use than RHF, specifically I find it intuitive to think of forms in terms of json schemas. 

The one downside in my mind is that the forms automatically fill their containing divs unless you configure otherwise. I think this is fine in most cases but sometimes it can look a bit weird

kmshin1397:
I would agree with Danny that at least for the very simple forms (where you are fine with whatever default Material UI components the jsonschema-form uses for a particular input type) the jsonschema-form approach is much more intuitive since you just have to declare the fields you need and what type in a JSON object. But once you get into customizing these widgets there isn't much of a difference in the complexity between the jsonschema-form and react-hook-forms IMO

acrellin (approved):

stefanv:
@kmshin1397 Please create an issue to remind us to revisit this issue; I think we can go ahead with this PR.

dmitryduev (approved):
Late to the party, but LGTM!

PR #1460, opened 2020-12-03
Remove instructions on how to change cosmology

stefanv:
This can go in the docs, but for most hosted instances it doesn't make sense.


profjsb (approved):
This is fine by me.

PR #1465, opened 2020-12-03
Add includeComments (default = false) query param to sources API and never include attachment data

acrellin:
This PR adds an optional `includeComments` query parameter to `SourceHandler.get` which defaults to `False`, and never ships comment attachment bytes.

Closes https://github.com/skyportal/skyportal/issues/1464

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -288,6 +288,14 @@ class SourceHandler(BaseHandler):
               type: string
             description: |
               The sort order - either "asc" or "desc". Defaults to "asc"
+          - in: query
+            name: includeCommentAttachments
+            nullable: true
+            schema:
+              type: boolean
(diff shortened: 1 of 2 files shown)


dmitryduev (approved):
Looks good, thanks @acrellin !

stefanv:
Can we avoid downloading all attachments by default when loading a source?  I feel like those should only be grabbed when requested.  Otherwise this is :+1:

acrellin:
@stefanv by default when loading a source, the comments are not loaded unless done so manually. Here we're loading the comments manually, and conditionally including the `attachment_name` and `attachment_bytes` columns when doing so.

stefanv (approved):

kmshin1397 (approved):

kmshin1397:
I think this is good to merge on green :)

PR #1466, opened 2020-12-03
GaiaQuery abstraction; push to Gaia EDR3

profjsb:
Push to Gaia EDR3 (`gaiaedr3`). Also, adds an abstraction layer for Gaia queries: first try the ESA TAP+ portal via astroquery, then drop back to `gaia.aip.de` TAP.

### DIFF
--- a/skyportal/utils/offset.py
+++ b/skyportal/utils/offset.py
@@ -6,10 +6,12 @@ from functools import wraps
 
 import pandas as pd
 import requests
+from requests.exceptions import HTTPError
 import matplotlib
 import matplotlib.pyplot as plt
 import seaborn as sns
 import numpy as np
+import numpy.ma as ma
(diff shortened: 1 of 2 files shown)


skyportal/utils/offset.py
```diff
@@ -32,6 +36,90 @@
 log = make_log('finder-chart')
 
 
+class GaiaQuery:
+
+    alt_tap = 'https://gaia.aip.de/tap'
```
dmitryduev, line 41: Maybe make this configurable?

dmitryduev (approved):
LGTM!

skyportal/utils/offset.py
profjsb: Let's refactor this once we also have a pathway to querying Kowalski (ie in another PR).

Issue #1428, opened 2020-11-26
Add more filtering options to the scanning page


dmitryduev:
Namely, convert the current checkbox "Show only unsaved candidates" into radio buttons/dropdown menu (replicating the GM):

![image](https://user-images.githubusercontent.com/7557205/100325427-7842e700-2f7d-11eb-9f7b-ef9c294a4e56.png)

PR #1419, opened 2020-11-25
Standard stars in summary starlist

profjsb:
Kapture 2020-11-25 at 10 08 36

Adds standard stars, formatted in the requested version at the end of a run starlist listing. Functionality in the backend is added for selecting standard star lists and filtering by position in the sky (useful for cutting down the number of stars which will not be viewable in a given run).

- Closes #1419 
- Addresses and closes https://github.com/fritz-marshal/fritz-beta-feedback/issues/45

PR #1474, opened 2020-12-05
Add more saved status filtering options to the scanning page

kmshin1397:
This PR implements more advanced filtering options for the scanning page based on save status.
Instead of a single unsaved only checkbox, we now have:
- Show only candidates which passed a filter from the selected groups...
  - regardless of saved status
  - and is saved to all selected groups
  - and is saved to at least one of the selected groups
  - and is saved to at least one of the user-accessible groups
  - and is not saved to any of the user-accessible groups
  - and is not saved to any of the selected groups
  - and is not saved to all of the selected groups

scanning
Addresses #1428 

Also made small updates to other tests in `test_candidates.py`:
- Consistently using `"` instead of `'`.
- Using the `params` optional argument added to the `api()` function in #1419 to pass query arguments in test API calls


skyportal/handlers/api/candidate.py
```diff
@@ -399,14 +401,84 @@ def get(self, obj_id=None):
...
+                # Retrieve objects that have as many active saved groups that are
+                # in 'group_ids' as there are items in 'group_ids'
+                q = q.filter(
+                    Obj.id.in_(
+                        DBSession()
+                        .query(Source.obj_id)
+                        .filter(Source.active.is_(True))
+                        .filter(Source.group_id.in_(group_ids))
+                        .group_by(Source.obj_id)
+                        .having(func.count(Source.group_id) == len(group_ids))
```
dmitryduev, line 424: Nice!

dmitryduev (approved):
Awesome, thanks @kmshin1397!

stefanv:
From a user perspective, it may not be clear what a "user accessible group is".  Maybe "to at least one group I have access to"?

skyportal/handlers/api/candidate.py
```diff
@@ -399,14 +401,84 @@ def get(self, obj_id=None):
...
+            "notSavedToAnyAccessible",
+            "notSavedToAnySelected",
+            "notSavedToAllSelected",
+        ]:
+            if saved_status == "savedToAllSelected":
+                # Retrieve objects that have as many active saved groups that are
+                # in 'group_ids' as there are items in 'group_ids'
+                q = q.filter(
+                    Obj.id.in_(
+                        DBSession()
```
stefanv, line 419: These three lines are common to all queries, so can lift them out.

stefanv:
I think once the "user accessible" has been clarified a bit, this is good to go.  Thanks!

kmshin1397:
> From a user perspective, it may not be clear what a "user accessible group is". Maybe "to at least one group I have access to"?

Yeah that makes sense. I'll make that change.

skyportal/handlers/api/candidate.py
kmshin1397: Hmm I'm actually not sure the best way to cleanly lift them out - actually half of them use `Obj.id.notin_()` instead of `Obj.id.in_()` so not sure if lifting them out will result in cleaner code. I imagine I'd be assigning the inner queries to variables for each of these if/elif cases, but then will just end up creating another if/elif at the end for the `in_` vs `notin_` at the end anyways. @stefanv

stefanv: I think you can do `~table_a.id.in_(subquery))`?

kmshin1397: @stefanv common lines are now factored out

stefanv (approved):

PR #1477, opened 2020-12-08
Allow users to request admission to groups; add a pending requests management table for admins

acrellin:
request_admission

Closes https://github.com/fritz-marshal/fritz-beta-feedback/issues/21


static/js/components/GroupAdmissionRequestsManagement.jsx
```diff
@@ -0,0 +1,187 @@
...
+      options: {
+        filter: false,
+        customBodyRenderLite: renderUserInfo,
+      },
+    },
+    {
+      name: "status",
+      label: "Status",
+      options: {
+        filter: false,
```
kmshin1397, line 146:
```suggestion
        filter: true,
```
I think filtering by status would be potentially useful - what do you think? Otherwise, should just set `filter: false` in the `options` below to get rid of an essentially useless toolbar button.

static/js/components/GroupAdmissionRequestsManagement.jsx
```diff
@@ -0,0 +1,187 @@
...
+  };
+
+  const renderUserInfo = (dataIndex) => {
+    const { user } = requests[dataIndex];
+    let userInfoString = user.username;
+    if (!!user.first_name && !!user.last_name) {
+      userInfoString += ` (${user.first_name} ${user.last_name})`;
+    }
+    return userInfoString;
+  };
```
kmshin1397, line 131:
```suggestion
  const renderUserInfo = (value) => {
    let userInfoString = value.username;
    if (!!value.first_name && !!value.last_name) {
      userInfoString += ` (${value.first_name} ${value.last_name})`;
    }
    return userInfoString;
  };
```
And set `customBodyRenderLite: renderUserInfo` below to `customBodyRender: renderUserInfo` - that will allow using the search tool to search the usernames and user names

static/js/components/NonMemberGroupList.jsx
```diff
@@ -0,0 +1,139 @@
...
+  ];
+
+  const options = {
+    responsive: "standard",
+    download: false,
+    search: true,
+    selectableRows: "none",
+    rowsPerPage: 10,
+    rowsPerPageOptions: [10, 25, 50, 100, 200],
+    filter: true,
```
kmshin1397, line 113:
```suggestion
    filter: false,
```
Don't think we need to have filtering for any of these columns (which is already the case), so just turn it off here to get rid of the toolbar icon.

static/js/actions.js
```diff
@@ -28,5 +29,6 @@ export default function hydrate() {
     dispatch(observingRunsActions.fetchObservingRuns());
     dispatch(telescopesActions.fetchTelescopes());
     dispatch(taxonomyActions.fetchTaxonomies());
+    dispatch(groupActions.fetchGroup(7));
```
kmshin1397, line 32: Can you explain why we need this here? And this wouldn't work in production either way since groupID 7 doesn't have a special meaning right?

skyportal/handlers/api/group_admission_request.py
```diff
@@ -0,0 +1,213 @@
...
+        responses:
+          200:
+            content:
+              application/json:
+                schema: Success
+        """
+        data = self.get_json()
+        status = data.get("status")
+        if status is None:
+            return self.error("Missing required parameter `status`")
```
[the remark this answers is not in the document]
acrellin: :+1:

static/js/components/GroupAdmissionRequestsManagement.jsx
acrellin: Yes, good call

static/js/components/NonMemberGroupList.jsx
acrellin: :+1:

static/js/actions.js
acrellin: Ah woops, this was a left-behind debugging-only addition. Removed :)

acrellin: I spent a few hours tracking down a typo in one of the reducers and added this to help pin down the issue

static/js/components/NonMemberGroupList.jsx
acrellin: Done

acrellin:
@kmshin1397 thanks for the review -- I've implemented all of your suggestions

kmshin1397 (commented):
Looking good! I think this should be good once tests get added

stefanv:
@acrellin We need some way to notify admins of the groups that there are requests waiting.  Maybe this is the time to add the "notifications waiting" bell that @profjsb had on the initial Figma design?

acrellin:
> @acrellin We need some way to notify admins of the groups that there are requests waiting. Maybe this is the time to add the "notifications waiting" bell that @profjsb had on the initial Figma design?

@stefanv yes, this, along with source save requests, should ideally trigger a notification to the respective group admins. That should be in another PR, though, as this one is already sizeable. If there isn't already an issue for that I will open one

stefanv:
The PR looks good to me otherwise; it still needs a migration.

skyportal/handlers/api/group_admission_request.py
```diff
@@ -0,0 +1,217 @@
...
+        if (
+            self.associated_user_object.id != user_id
+            and not self.current_user.is_system_admin
+        ):
+            return self.error("Insufficient permissions")
+        if Group.query.get(group_id) is None:
+            return self.error("Invalid group ID")
+        if User.query.get(user_id) is None:
+            return self.error("Invalid user ID")
+        admission_request = GroupAdmissionRequest(
```
dmitryduev, line 133: We should exclude single-user groups, because now you can submit a request there.

dmitryduev (changes requested):
A minor request, but otherwise LGTM, so can go in once the migration is in place. Thanks, @acrellin

acrellin:
@dmitryduev thanks for the review -- feedback addressed, and tests & migration script are now in place

acrellin:
@kmshin1397 tests and migration script now in place

dmitryduev (approved):
👍🏻

static/js/components/Group.jsx
```diff
@@ -258,27 +259,24 @@ const Group = () => {
...
 
+  useEffect(() => {
+    const fetchGroup = async (groupID) => {
+      const result = await dispatch(groupActions.fetchGroup(groupID));
+      if (result.status === "error") {
+        setGroupLoadError(result.message);
+      }
+    };
+    if (String(id) !== String(group?.id)) {
+      fetchGroup(id);
```
stefanv, line 276: If id is invalid, won't this cause an infinite loop?

I know we've had to make special arrangements in other similar patterns to avoid that.

static/js/components/NonMemberGroupList.jsx
```diff
@@ -0,0 +1,139 @@
...
+  const handleRequestAdmission = async (groupID) => {
+    const result = await dispatch(
+      groupAdmissionRequestsActions.requestGroupAdmission(
+        currentUserID,
+        groupID
+      )
+    );
+    if (result.status === "success") {
+      dispatch(showNotification("Successfully requested admission to group."));
+      // dispatch(groupsActions.fetchGroups(true));
```
stefanv, line 35: leftover comment

static/js/ducks/groupAdmissionRequests.js
```diff
@@ -0,0 +1,65 @@
...
+  if (actionType === FETCH_GROUP_ADMISSION_REQUESTS) {
+    dispatch(fetchGroupAdmissionRequests(payload.group_id));
+  }
+});
+
+const reducer = (state = null, action) => {
+  switch (action.type) {
+    case FETCH_GROUP_ADMISSION_REQUESTS_OK: {
+      const { data } = action;
+      // action.data is an array of records
```
stefanv, line 56: Why is it an array but only the first element is used?

static/js/components/Group.jsx
acrellin: Note this logic is not new to this PR, just got moved around

static/js/components/NonMemberGroupList.jsx
acrellin: Thanks

static/js/ducks/groupAdmissionRequests.js
acrellin: The entire array is used... See reducer return value. Each array element has the same group ID, because we only fetch these for a single group (see action creator)

static/js/components/Group.jsx
acrellin: I'll add a fix to this PR though (dataFetched component state that gets set to true after initial fetch, as we've done elsewhere)

acrellin: Done

kmshin1397 (commented):
@acrellin seems like `test_group_admission_request_and_acceptance` failed on GA

acrellin:
> @acrellin seems like `test_group_admission_request_and_acceptance` failed on GA

@kmshin1397 it looks like the group/user in question is not on the first table page, so I need to filter the table, but there are multiple mui-datatables on the page so multiple matches for the `@data-testid="Search-iconButton"` xpath. We could just grab the first or last match, respectively, but the addition of more datatables would require those tests to be updated in order to still pass. I guess that's okay for now.

kmshin1397 (approved):

acrellin:
Green. Anyone wanna merge?

PR #1484, opened 2020-12-09
make demo observing runs earlier in time to avoid running into outdated IERS file exceptions

dannygoldstein:
@scizen9 was running into this bug locally when trying to access source pages through the web interface  (the error was being flashed as a notification)

https://github.com/astropy/astropy/issues/8788

we found that the reason for this is that when the main IERS mirror at NASA GSFC is down, it reverts to a backup mirror which serves a file that only has IERS data until 2020-08-22, which is not sufficiently late in time to calculate ephemerides for targets on runs in the demo data set (june 2021). The main IERS file from GSFC goes out till 2022, so it worked in that case.

This PR fixes that issue by moving the dates of the demo runs back in time so that they are within the period of IERS coverage for both the main and backup files.

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -295,13 +295,13 @@ taxonomy:
 observing_run:
   - pi: Danny Goldstein
     instrument_id: =ESI
-    calendar_date: "2021-02-16"
+    calendar_date: "2020-02-16"
     group_id: =program_A
     observers: "Danny Goldstein, Peter Nugent"
     =id: esi_run
   - pi: Danny Goldstein
     instrument_id: =LRIS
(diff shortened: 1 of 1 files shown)


scizen9 (approved):
Did a checkout of this PR and verified these changes.  Then reset the database (make db_clear && make db_init) and ran make load_demo_data with no issues.  I also verified that the cached version of the IERS database was not updated and therefore the main source appears to still be down: thus I approve this as a fix for this issue.

dmitryduev (approved):

dmitryduev:
Failure probably unrelated, so merging

Issue #1332, opened 2020-11-11
Add option to retrieve sources by classification to SourceHandler.get


dannygoldstein:
Request from @mansikasliwal:

Users want to be able to retrieve all sources that have been classified a certain way across all their groups. We should be able to specify a list of (taxonomy, class_within_taxonomy) pairs to the sourcehandler to only retrieve sources that are of that class.

kmshin1397:
This is implemented as a part of #1495, but that implementation just filters on class names instead of (taxonomy, class) pairs. Should I change that logic? @dmitryduev

dmitryduev:
Yea, I think using the (t, c) pair is preferable: there could be classes in different taxonomies that use the same name, but mean somewhat different things. For example, what some people might call bogus (meaning "not my science") would be quite different from we call bogus in acai.

Issue #1475, opened 2020-12-07
Sources page does not allow sorting on "classification" column


stefanv:
https://fritz.science/sources

Issue #1392, opened 2020-11-19
Add Classification to Source Page search


profjsb:
The user should be able to search by site classification (not just Simbad class). Suggest an autocomplete for chips (which serve as a OR at the SQL level).

PR #1495, opened 2020-12-11
Classification sorting/filtering on sources page

kmshin1397:
- Add sorting and filtering by classification to the sources page.
  - Addresses #1392 and #1475
- Clean up `api/test_sources.py` to use double-quotes consistently and use the `params` arg to API GET calls

sources


static/js/components/SourceList.jsx
```diff
@@ -43,17 +49,40 @@ const useStyles = makeStyles((theme) => ({
...
+  // Filter out empty fields from an object (form data)
+  Object.keys(data).forEach((key) => {
+    // Empty array ([]) counts as true, so specifically test for it
+    if (!(Array.isArray(data[key]) && data[key].length == 0) && key) {
+      filteredData[key] = data[key];
+    }
+  });
+  return filteredData;
+};
+
```
kmshin1397, line 82: There were issues associated with passing along empty/default form values introduced by #1487 (the ducks used to just ignore form values that weren't filled out but now they're added as query params I think), and this is used to remove those before calling the API fetch action

acrellin: @kmshin1397  Would it suffice to just add a line to `API.GET` that removes any properties whose value are in `[null, undefined, ""]` or something, before passing it to URLSearchParams?

acrellin: i.e. just always pass `filterParams` through the above function in API.GET?

acrellin (approved):
Looks good. I just wonder if we should move `filterOutEmptyValues` to `API` and always pass `filterParams` through that for every request

static/js/components/SourceList.jsx
kmshin1397: @acrellin I think that's a good idea - I can imagine this popping up again as an issue elsewhere in the future if we're not careful. And I don't think there will ever be a case where we need to let those through. I'll go ahead and make that change.

acrellin (approved):
Looks like there are some linter errors, otherwise :+1:

dmitryduev (changes requested):
Mostly LGTM, a few UI requests:
- When you sort by classifications (or actually anything) and then click to the next page, the sorting resets, so there's no way to see sorted results beyond the first page. I think the same may apply to other places where we're using mui datatables with server-side logic. Could you please fix that?
- A comment/request on sorting: how hard would it be to make it work just by clicking the column name (once or twice, depending on the desired order; I think you can specify the default direction per column in mui-dt)? At the moment it takes 5 clicks to do that and I always find myself clicking on the column name first. I think this enhancement would be much appreciated.
- It's a bit counterintuitive to my taste that when you sort the table on a column, the entries where that is non-existent just disappear - I'd expect them to show up after the sorted entries, in whatever order. Would it be hard to implement? Not critical

dmitryduev:
A side note for a future PR: I think it would be really cool to visualize the available taxonomies and interactively select from the available classes using @profjsb's `tdtax`-like layout. This way, users would also be able to filter on, e.g., higher-level classifications, for example on SNe in general, without having to select all the subtypes.

kmshin1397:
@dmitryduev:

1. That's actually the next thing on my list, and as you say, is present in multiple places so I was going to make that a separate PR. It's encapsulated as an issue [here](https://github.com/fritz-marshal/fritz/issues/135)

2. Agreed that it would be much better to have the sorting work by clicking the column. If I remember correctly, I had trouble getting the custom server-side logic to hook into the sorting events exposed through clicking the column, so I ended up disabling the built-in sorting and adding my own toolbar form for it. But I can definitely look into it more - I gave up on it pretty early on it in favor of the current approach pretty early so there may be something I missed.

3. Hmm I thought that functionality was already in there (I think all the API handlers have `nullslast()` tacked on to the sorting filters) but I don't think I ever tested it. I'll look into it.

kmshin1397:
@dmitryduev I just tested sorting by redshift and my test source with null redshift was properly placed as the last entry. Can you tell me in which scenario you saw this not working properly?

dmitryduev:
I saw it when sorting by classification. What I did was saving 12 random ZTF sources and adding classifications to 10 of them (+ classification of the demo sources), then when sorting by classification, the two sources without a classification would disappear from the table.

kmshin1397:
DNM: will refactor to use (taxonomy, class) instead of just class for filtering

kmshin1397:
> I saw it when sorting by classification. What I did was saving 12 random ZTF sources and adding classifications to 10 of them (+ classification of the demo sources), then when sorting by classification, the two sources without a classification would disappear from the table.

Figured out the issue - will push a fix soon :)

kmshin1397:
Refactored to use (taxonomy name, classification) pairs for filtering:

Screenshot 2020-12-15 183950

kmshin1397:
@acrellin @dmitryduev tagging for re-review because I:

- Added fix to null classification sources disappearing from sorted results
- Refactored to use (taxonomy.name, classification) pairs for filtering instead of just the classification (which closes #1332)

Dima, your requests 1 and 2 will be addressed in future PRs (#1505 and then a follow-up once that goes in for the sorting via column header change - I figured out how to do it)

dmitryduev (approved):
Excellent, thanks a lot @kmshin1397 for addressing these! LGTM!

acrellin (approved):

PR #1497, opened 2020-12-11
Ensure photometry is returned in the correct order for photometry api tests 

dannygoldstein:
Object photometry used to be returned in MJD order, an assumption that went into this test

https://github.com/skyportal/skyportal/runs/1535057564#step:13:579

However, that was removed at some point for performance reasons, so the comparisons in the failing test above were comparing photometry points of different IDs.

This PR sorts the photometry by MJD order in that test so that the correct photometry points are compared.

### DIFF
--- a/skyportal/tests/api/test_photometry.py
+++ b/skyportal/tests/api/test_photometry.py
@@ -1240,6 +1240,7 @@ def test_token_user_retrieving_source_photometry_and_convert(
     assert 'mjd' in data['data'][0]
     assert 'ra_unc' in data['data'][0]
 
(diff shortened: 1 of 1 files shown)


acrellin (commented):
test changes look good, but the table creation changes are in here, too

dannygoldstein:
thanks.fixed

acrellin (approved):
and this works because 58000 < 58000 + np.random.random()?
looks good. will merge when green

acrellin:
Sole failure (`test_source_is_added_to_observing_run_via_frontend`) is unrelated

PR #1500, opened 2020-12-12
Skinny endpoint to retrieve source groups info

dmitryduev:
This PR implements an endpoint `/api/sources/<source_id>/groups.GET` to retrieve basic source groups info. This will significantly speed up alert posting from K on F as currently K has to `/api/sources/<source_id>.GET` just to get group IDs for a saved source.


stefanv (approved):

acrellin:
Does https://github.com/skyportal/skyportal/blob/master/skyportal/handlers/api/source.py#L254 not suffice?

stefanv:
This returns all groups, though; do you get that from the save summary?

acrellin:
Every group the source is saved to, yes

dmitryduev:
@acrellin I need it per obj_id, whereas saveSummary is meant for bulk queries. There's also the sourceId parameter, but it's used as in `q = q.filter(Obj.id.contains(sourceID.strip()))` whereas I need an exact match (e.g. if I'm looking for ZTF20bsbsbs and someone created ZTF20bsbsbs_1, I'll get that too). The use case is: when processing an alert, K needs to figure out just which groups a corresponding source is saved to (for posting photometry, autosaving and other such purposes). The point is that is has to be snappy as there are potentially hundreds of thousands such queries every night (we're in tens of thousands regime already).

dmitryduev:
Failure in `test_submit_new_followup_request_two_groups ` seems unrelated.

acrellin (approved):

PR #1505, opened 2020-12-14
Properly save state across page changes in sources table

kmshin1397:
For the SourceTable.jsx component and the pages that use it (Sources and GroupSources):
- Save sorted column info across page changes (pass sort state to pagination handlers)
- Save displayed column info across component re-renders
- Properly detect and display sources that have no photometry points

Address https://github.com/fritz-marshal/fritz/issues/135

Also:
- Clean up `test_top_source_prefs` to use a `data-testid` XPATH instead of current one that was ambiguous and leading to flakiness

sourcetablestate

### DIFF
--- a/skyportal/tests/frontend/test_group_sources.py
+++ b/skyportal/tests/frontend/test_group_sources.py
@@ -3,7 +3,7 @@ import pytest
 from .. import api
 
 from tdtax import taxonomy, __version__
-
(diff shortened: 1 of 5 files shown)


acrellin (approved):

dmitryduev (approved):
👍🏻

Issue #1491, opened 2020-12-11
Add option to reset sorting column for source table 


kmshin1397:
Currently, the only way to remove sorting on a column on SourceTable.jsx is to sort on something else or refresh the page.

kmshin1397:
Also explore moving sorting functionality to built-in "click on column header" behavior instead of custom toolbar form

PR #1510, opened 2020-12-16
Refactor source table to use sorting via clicking on column header

kmshin1397:
Addresses #1491 

sorting


acrellin (approved):
Nice

dmitryduev:
Big like!

PR #1512, opened 2020-12-17
Fix encoding of plaintext spectrum data

dannygoldstein:
This PR fixes a bug where some spectrum files would not be properly encoded for download due to inclusion of characters that were not handled by `encodeURI`. 

The solution here casts these to blobs and then encodes those to plaintext urls using `URL.createObjectURL` as recommended by https://javascript.info/blob#blob-as-url. 

Tested this locally and it fixes the issue.


acrellin (approved):

acrellin:
Failure is unrelated

PR #1515, opened 2020-12-18
Add md5hash query to bundle to ensure reload when necessary

stefanv:
This needs https://github.com/cesium-ml/baselayer/pull/183

Closes https://github.com/skyportal/skyportal/pull/1492

### DIFF
--- a/static/index.html.template
+++ b/static/index.html.template
@@ -12,6 +12,9 @@
   </head>
   <body>
     <div id="content"></div>
-    <script src="/static/build/main.bundle.js"></script>
+    <script
+      src="/static/build/main.bundle.js?{{'static/build/main.bundle.js' | md5sum}}"
+    >
+    </script>
   </body>
 </html>
--- a/static/index.html.template
+++ b/static/index.html.template
@@ -13,7 +13,7 @@
   <body>
     <div id="content"></div>
     <script
-      src="/static/build/main.bundle.js?{{'static/build/main.bundle.js' | md5sum}}"
+      src="/static/build/main.bundle.js?{{ 'skyportal' | version | hash('md5') }}"
     >
     </script>
   </body>


static/index.html.template
```diff
@@ -12,6 +12,9 @@
   </head>
   <body>
     <div id="content"></div>
-    <script src="/static/build/main.bundle.js"></script>
+    <script
+      src="/static/build/main.bundle.js?{{'static/build/main.bundle.js' | md5sum}}"
```
acrellin, line 16: @stefanv is the vertical bar jinja-specific? I'm not sure what it achieves here

stefanv: Yes, that is the extension I wrote into our template parser.  It takes the filename, calculates its md5sum, and replaces the `{{ ... }}` with that value.

acrellin:
@stefanv wanna pin to latest BL master? I merged the corresponding PR there

acrellin:
@stefanv  How does this address the issue we discussed before where fill_conf_values runs before webpack bundling (necessarily, as we have templated JS files)?

stefanv:
Yes, you're right: there will need to be an extra run of the filler before deployment. Perhaps `make run` and `make production` could do that?

acrellin:
That seems far from ideal -- that's why I just calculated the hash on the static/js directory contents, so it can be done before webpack runs. How about just switch to doing that?

stefanv:
My concerns with that were:

- It templates webpack.config.js, making it harder to edit
- Adds a somewhat arbitrary variable into the template filling namespace, and one that may be quite expensive to compute (it hsa to hash every single file in that directory!)

If you prefer that kind of approach to avoid running the filler one additional time for the index file, we can just as well modify the  filler to 1) check the timestamp of the JS bundle 2) hash it whenever it changes and 3) provide that as a variable that can be used to template out the index file.  WebPack can [already calculate md5 hashes](https://www.npmjs.com/package/webpack-md5-hash), so that should address the above concerns of mine, while sticking more closely to your preferred approach.

acrellin:
@stefanv  I'm not sure if I get all of the latter part of your comment, but what I was suggesting was just keeping everything you have here except hashing the JS sources directory contents instead of the bundled file.

stefanv:
We could, it's just a very expensive operation; maybe we can calculate a cheaper hash.  Something that checks, say, file timestamps and sizes or... not sure what would be sufficient to guarantee a unique bundle?

acrellin:
@stefanv it's not expensive at all: 
``` 
In [2]: %time dirhash("./static/js", "md5")       
CPU times: user 5.73 ms, sys: 0 ns, total: 5.73 ms
Wall time: 5.4 ms                                 
```

stefanv:
Haha, I guess I'm the only person in the world who feels sorry for the computers.

OK, fine, let's go with that then.

stefanv:
I just realized that the same problematic condition exists here: if we checksum the directory, it is the checksum of the directory at the point where the index is created.  But, files *inside* the directory can be modified by the template filler later on, rendering the checksum invalid (or, more importantly, insensitive to changes).

So, I think we have no choice but to calculate the checksum after the filler completes to get a checksum that reflects the bundled state of the code.

acrellin:
Or just add the contents of config.yaml to the hash?

stefanv:
Why don't we just substitute the version number instead of the hash?  Every time we redeploy, we'd want clients to reload anyway.

acrellin:
Because that can lead to a lot of unnecessary re-fetching of an identical
bundle (e.g. when we commit back-end only changes). But maybe we don't
care? What's the minified bundle size?

stefanv:
I think the main bundle is 4MB. I think we seldom don't touch the JS with a new deploy.

acrellin:
@stefanv yes, good point. Probably fine to just use the version instead, then. Great!

stefanv:
This will fail the test suite for now---it requires the BL patch from https://github.com/cesium-ml/baselayer/pull/193

But, it shows the general idea.

acrellin:
So you hash the SP version number here? That looks good. Wanna pin latest BL?

stefanv:
OK, done

stefanv:
Err, not sure how to resolve that conflict.  How can this be in conflict with master?

acrellin:
Nice! Tested locally, looking good

PR #1517, opened 2020-12-18
Fix broken profile test

kmshin1397:
Test failure with `test_insufficient_name_entry_in_profile` was due to a fixture change - was testing empty first names in the profile form, but the user fixture was changed in #1253  to initialize users with a random first name


dmitryduev (approved):
👍🏻

Issue #1383, opened 2020-11-18
Event-Driven News Feed with Notifications


dannygoldstein:
Summarizing a discussion from stand-up this morning:

- [x] Augment the news feed to more types of events, see #1353
- [x] Make the types of events tracked in the news feed configurable via the profile page
- [ ] Add a "+" button on the home page that allows users to set up different types of news feeds on the home page that each track different types of events 
- [x] Set up user-configurable push, email, text, and in-app notifications on desired news feed events

mcoughlin:
@Theodlz Do you feel like we have fulfilled the spirit of this one? Or is there a separate issue to make that should still be implemented?

Theodlz:
Hi @mcoughlin. `Add a "+" button on the home page that allows users to set up different types of news feeds on the home page that each track different types of events` has not been addresses yet, but I'm not sure we still want that (anything in the newsfeed itself, we added notifications that do the same I believe)?

mcoughlin:
@Theodlz exactly. Want to close it then?

Theodlz:
I think we can yes!

PR #1521, opened 2020-12-18
Make news feed categories configurable via associated WidgetPrefsDialog

### LINKED ISSUE #1518: Make configurable the types of records that appear in news feed
Sub-issue of #1383

### LINKED ISSUE #1519: Allow both text and checkbox group inputs in WidgetPrefsDialog; refactor to use RHF

acrellin:
This PR makes the news feed categories configurable via the widget preferences dialog form, and refactors that form to accommodate both text fields and checkbox groups (represented by an object whose keys are value names, and values are `true`/`false`, corresponding to checked/unchecked). A test is added.

configure_newsfeed_categories

Closes #1518 
Closes #1519 
Partially addresses #1383

### DIFF
--- a/static/js/components/WidgetPrefsDialog.jsx
+++ b/static/js/components/WidgetPrefsDialog.jsx
@@ -12,6 +12,7 @@ import { makeStyles } from "@material-ui/core/styles";
 import SaveIcon from "@material-ui/icons/Save";
 import TextField from "@material-ui/core/TextField";
 import FormControlLabel from "@material-ui/core/FormControlLabel";
+import Typography from "@material-ui/core/Typography";
 
 const useStyles = makeStyles(() => ({
   saveButton: {
@@ -77,8 +78,7 @@ const WidgetPrefsDialog = ({
               ) {
                 return (
                   <div key={key} className={classes.inputSectionDiv}>
-                    Select {key}:
-                    <br />
+                    <Typography variant="subtitle2">Select {key}:</Typography>
                     {Object.keys(initialValues[key]).map((subKey) => (
                       <FormControlLabel
                         key={subKey}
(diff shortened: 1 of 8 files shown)


static/js/components/WidgetPrefsDialog.jsx
```diff
@@ -63,38 +65,82 @@ const WidgetPrefsDialog = ({
...
+            onSubmit={handleSubmit(formSubmit)}
+          >
+            {Object.keys(initialValues).map((key) => {
+              if (
+                typeof initialValues[key] === "object" &&
+                initialValues[key].constructor === Object
+              ) {
+                return (
+                  <div key={key} className={classes.inputSectionDiv}>
+                    Select {key}:
```
kmshin1397, line 80: Maybe wrap this in a Typography tag? Maybe variant="subtitle2" or something like that. I think it looks a little off right now because the header here is a smaller font than the checkbox options.

kmshin1397 (commented):
The code here looks good (other than my one comment). One concern though - existing users will already have their newsFeedPrefs populated with only the numItems and that will always get loaded in instead of the new default prefs, so there won't be a way for them to set these new checkbox prefs through the front-end... Not sure what the best way to modify that is. I guess either a direct database query to update everyone's user newsfeed prefs to the new default, or we add some check to the JS to add the default `categories` prefs if none exists.

acrellin:
Yep, good catch - I'll add the default categories dict to the user
preferences if it doesn't already exist when it's passed in.

Will update the headers as suggested.

Thanks for the review.

static/js/components/WidgetPrefsDialog.jsx
acrellin: @kmshin1397 done: 
image

acrellin:
@kmshin1397 both suggestions implemented, thanks. Ready for re-review

kmshin1397 (approved):

dmitryduev (approved):
LGTM!

PR #1526, opened 2020-12-21
fix coordinate formatting

dannygoldstein:
Fixes this issue reported by  dan perley:

Hi all, I think there is a major bug with the reporting/formatting of sexagesimal coordinates: as of right now, the coordinates of ZTF20acytfkf on its sources page (https://fritz.science/source/ZTF20acytfkf) are reported as:
05:02:33.6 +15:36:24.15 (α,δ= 75.6377796, 15.606709)
But I think the sexagesimal string conversion of the decimal coordinates should actually be
05:02:33.06
... which is an offset of order 10 arcseconds!

### DIFF
--- a/static/js/units.js
+++ b/static/js/units.js
@@ -15,7 +15,9 @@ const ra_to_hours = (ra, sep = null) => {
     .padStart(2, "0");
   const ra_s = (((ra % 15) * 4) % 1) * 60;
   const ra_s_integer = Math.floor(ra_s).toString().padStart(2, "0");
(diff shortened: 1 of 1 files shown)


dmitryduev (approved):
Oh boi... thanks for the fix, @dannygoldstein !

stefanv (approved):

stefanv:
I'd suggest merging and rolling out ASAP, then writing a test in a new PR (can be part of the existing source page test).

PR #1529, opened 2020-12-21
Fix coordinate formatting, round 2

dannygoldstein:
The previous solution had a single (but important) failure mode. 

For sources with ra= hh:mm:ss.99 or dec = dd:mm:ss.99, due to roundoff error the coordinates would be rendered as ra = hh:mm:ss.100 or dec = dd:mm:ss.100, which is incorrect.

Rather than trying to mess around with formatting these in raw javascript (which is error prone), I just used a library, numeral, to handle the formatting directly. This works. 

And I added a test that includes the two pathological cases.


acrellin (approved):

acrellin:
thanks @dannygoldstein , please merge when green

dannygoldstein:
Related to #1526

dannygoldstein:
failures unrelated

PR #1540, opened 2020-12-27
WIP: ZTF API

mcoughlin:
This PR provides a submission interface for the ZTF interface.


dmitryduev:
@mcoughlin this is really nice, but I would close it here and move to Fritz as an extension. @dannygoldstein do you agree? https://github.com/fritz-marshal/fritz/blob/master/extensions/skyportal/skyportal/handlers/api/kowalski_filter.py would be a good reference for talking to K from SP/F - we have `penquins` as a requirement there, so processing requests would be more robust.

mcoughlin:
I definitely understand having the talking to K from SP/F bit in fritz, but I am a bit surprised everything else would be there too, given all the instruments are otherwise in SP.

dmitryduev:
yea, that's right, but can you separate the concerns easily? @dannygoldstein @stefanv @kmshin1397 @acrellin what do you guys think?

requirements.txt
```diff
@@ -39,3 +39,5 @@ twilio>=6.45.3
 jsonschema-to-openapi>=0.2.1
 responses>=0.12.0
 numba>=0.51.2
+sshtunnel
+pyjwt
```
acrellin, line 43: `pyjwt` is already a requirement in `baselayer`

acrellin:
The general followup request test improvements and the new instrument in the demo data should go in SkyPortal, but it looks like pretty much everything else is ZTF-specific, right?

kmshin1397:
Hmm is there a reason not to have any ZTF references in particular at all in the SkyPortal code? It does feel like since LT and SEDM are already part of the codebase with code/URLs specific to those APIs that at least some form of the ZTF API should also be allowed in here. Perhaps we could stub out just the `authenticate` and `query_kowalski` parts here and fill those in on the Fritz side? The `query_kowalski` method could be renamed to something generic like `send_ztf_request` for example, and both that and `authenticate` could just return a `NotImplementedError` in SkyPortal?

PR #1554, opened 2021-01-04
Correct hybrid properties dealing with 'NaN' flux values

kmshin1397:
- The `flux` column of the `Photometry` model defines a default value of `'NaN'` and is not nullable. Thus, we should be checking for `nan` float values when looking for valid flux points instead of something like `self.flux is not None` or `if self.flux:`. For example:
Screenshot 2021-01-04 113355
- Additionally, Postgres considers NaN values to be greater than all non-NaN values, while python will properly return false for all comparisons against NaN:

Screenshot 2021-01-04 113748
(https://www.postgresql.org/docs/9.1/datatype-numeric.html)

This PR addresses issues stemming from these two points by using `np.isnan()` to check for NaN flux points instead of checking for `None` values in the python side of relevant hybrid properties, and using ` != 'NaN'` instead of `!= None` in the SQL expressions of the hybrid properties.

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -591,7 +591,7 @@ class Obj(Base, ha.Point):
         detections = [
             phot.iso
             for phot in self.photometry
-            if not np.isnan(phot.snr) and phot.snr > 5
+            if phot.snr is not None and phot.snr > 5
         ]
         return max(detections) if detections else None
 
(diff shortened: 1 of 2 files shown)


dannygoldstein (approved):

stefanv:
Looks good, thanks!  Could you add a test to ensure we don't break this again?

kmshin1397:
> Looks good, thanks! Could you add a test to ensure we don't break this again?

Test added

acrellin (approved):

PR #1557, opened 2021-01-05
Exclude candidates on having an annotation

guynir42:
This PR adds minor improvements to the candidates handler. The GET now has some new optional query arguments:
1) includeSpectra: the same as includePhotometry, just loads the spectra along with the object.
2) annotationExcludeOrigin and annotationExcludeDate

The annotationExcludeOrigin, if given, gets all candidates that already have an annotation from the specified origin. 
The annotationExcludeDate, if given, will modify the above, and only get candidates that have an annotation from that origin if the annotation was modified at or after the given date. 

The reason I need this is that I want to download candidates, process them, and post an annotation with the results. So I want to only get the candidates I haven't processed yet. 
If I update the analysis code, I may want to re-process older annotations, so I want to include the ones with annotations made before some date as they are outdated.


skyportal/handlers/api/candidate.py
```diff
@@ -473,6 +512,37 @@ def get(self, obj_id=None):
...
+                        Annotation.modified >= expire_date,
+                    )
+                    .subquery()
+                )
+                # q.filter(
+                #     Annotation.origin
+                #     != annotation_exclude_origin | Annotation.origin
+                #     == annotation_exclude_origin & Annotation.modified
+                #     < expire_date
+                # )
```
kmshin1397, line 542: Remove old commented out code

skyportal/handlers/api/candidate.py
```diff
@@ -355,6 +390,10 @@ def get(self, obj_id=None):
         end_date = self.get_query_argument("endDate", None)
         group_ids = self.get_query_argument("groupIDs", None)
         filter_ids = self.get_query_argument("filterIDs", None)
+        annotation_exclude_origin = self.get_query_argument(
+            'annotationExcludeOrigin', None
+        )
+        annotation_exclude_date = self.get_query_argument('annotationExcludeDate', None)
```
kmshin1397, line 396:
```suggestion
        annotation_exclude_date = self.get_query_argument('annotationExcludeOutdatedDate ', None)
```

skyportal/tests/api/test_candidates.py
```diff
@@ -632,3 +643,62 @@ def test_candidate_list_redshift_range(
...
     assert data["data"]["candidates"][0]["id"] == obj_id1
+
+
+def test_exclude_by_outdated_annotations(
+    annotation_token, view_only_token, public_group, public_candidate, public_candidate2
+):
+    status, data = api(
+        "GET",
+        "candidates",
+        params={"redshiftRange": "(0,0.5)", "groupIDs": f"{public_group.id}"},
```
kmshin1397, line 654:
```suggestion
        params={"groupIDs": f"{public_group.id}"},
```

Might want to remove the redshiftRange filter from these queries I think. All the test objects by default have redshift 0.0 so I don't think it actually filters anything, and just in case sometime in the future someone decides to make the ObjFactory randomize the redshifts or something like that and the `public_candidate` is filtered out by the redshift instead of the annotation origin check, you know? I think it's good to touch only the most relevant parts of the code when writing tests to minimize random failures like that.

kmshin1397 (commented):
A few minor comments from me

kmshin1397 (approved):

kmshin1397 (commented):
Ok, looks good but "test_exclude_by_outdated_annotations" is failing so probably need to look into that

guynir42:
@danny all checks are now green...

kmshin1397 (approved):

PR #1558, opened 2021-01-05
Spectra normalization

guynir42:
spectra normalization uses median(abs(flux)) instead of median(flux)

### DIFF
--- a/skyportal/handlers/api/spectrum.py
+++ b/skyportal/handlers/api/spectrum.py
@@ -573,7 +573,7 @@ class ObjSpectraHandler(BaseHandler):
         if normalization is not None:
             if normalization == "median":
                 for s in return_values:
-                    norm = np.median(s["fluxes"])
+                    norm = np.median(np.abs(s["fluxes"]))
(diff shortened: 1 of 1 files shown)


dmitryduev (approved):
👍🏻

skyportal/handlers/api/spectrum.py
```diff
@@ -573,7 +573,7 @@ def get(self, obj_id):
         if normalization is not None:
             if normalization == "median":
                 for s in return_values:
-                    norm = np.median(s["fluxes"])
+                    norm = np.median(np.abs(s["fluxes"]))
```
profjsb, line 576: Need to avoid the case where the median is exactly zero (which will lead to a divide by zero error)
```suggestion
                    norm = np.median(np.abs(s["fluxes"]))
                    norm = norm if norm != 0.0 else 1e-20
```

profjsb (changes requested):
Small change requested

profjsb (approved):

PR #1561, opened 2021-01-06
Empty spectra in source table

guynir42:
This PR fixes a few issues with plotting on the source tables (group sources and source list):
- empty photometry and spectra are now replaced with some text saying they're empty. 
- spectrum did not load in the VegaSpectrum because the API was changed.


acrellin (approved):

dmitryduev (approved):
Very nice, thanks @guynir42

PR #1533, opened 2020-12-22
Refactor photometry detection threshold nsigma to be configurable

kmshin1397:
There are multiple places throughout the codebase where some n-sigma value/ signal-to-noise ratio is hard-coded and used to determine whether a photometry point is a detection or not. This PR ties all that to a single value in the configs for consistency/configurability. 

Would particularly appreciate careful review of the inline comments I modified to make sure that they make sense in the scientific sense :)

PR #1568, opened 2021-01-07
Add more sources table filtering options and port filtering to GroupSources tables

### LINKED ISSUE #1506: Add filtering for SourceTable component

kmshin1397:
This PR mainly enables filtering for the SourceTable.jsx MUIDataTable (used on the Sources and GroupSources pages):
- Move filtering form from the Sources page into a new SourceTableFilterForm.jsx component, which is used as the filter form opened by the filtering tool button on the MUIDataTable
- Add filtering fields for filtering by redshift, peak magnitude, latest magnitude, and whether a sources has a spectrum
- Add new hybrid properties to the Obj class to facilitate some of these:
  - `last_detected` is now `last_detected_at`
  - `last_detected_mag` added
  - `peak_detected_at` and `peak_detected_mag` added
-  Also updated API docs to show `includePhotometry`, `includeComments`, `includePhotometryExists`, and `includeSpectrumExists` (the logic was already there for both single and multiple source GETs, but the API docs only showed the options for the multiple sources GET)
- Added a bunch of API tests for sources filtering, not just for the new filtering fields but also for the existing ones (like spatial filtering) that were not tested at all

sourcesfiltering

Closes #1506 

Note that once #1533 goes in, this PR will be updated to change the new hybrid properties implemented here to use the configurable thresholds


acrellin:
@kmshin1397  Can you pull in the latest master? Should resolve the test failures.

acrellin (approved):
Beautiful, thanks @kmshin1397!

Issue #1567, opened 2021-01-06
Korean initials improperly formatted


acrellin:
Breaking this out from another issue. 

@AlexGKim  can you please provide more details here, e.g. the exact text you input as first, last name and what you expect to see in the profile avatar? Thanks

AlexGKim:
I suggest using the first name only, not the first characters of the first and last names only.  In most all cases the first name will be <=2 characters.

kmshin1397:
Agreed - just the full first name would make sense for Korean names.

acrellin:
I'm looking into language detection libraries

kmshin1397:
@acrellin I also tried looking around but I found that names are not generally handled well because the string lengths were too short (the detection libraries out there are generally based on models I think so it's hard for them to really determine a language off of a few characters without distinctive words, etc.

For this specific issue, it's pretty easy to just regex match for Korean alphabet characters as no other language uses them. I can make a PR for it soon.

If you're curious, this is the best library I found: https://github.com/wooorm/franc

kmshin1397:
@AlexGKim I just filed the PR for this ^

acrellin:
Addressed by https://github.com/skyportal/skyportal/pull/1570

PR #1570, opened 2021-01-07
Add support for Korean user names

kmshin1397:
Update the user avatar and profile page name display to properly format Korean names. Korean names are generally written in last name, first name order with no space in between; a more natural (to a Korean speaker) set of initials for a Korean name would just be the full first name as first names are almost always <=2 characters in length and seeing first name -> last name order of the characters is strange (given the short length of names I don't think I've actually seen Koreans use "initials" in the manner of English names at all). 

Addresses #1567 

Screenshot 2021-01-07 182925

### DIFF
--- a/static/js/components/UserAvatar.jsx
+++ b/static/js/components/UserAvatar.jsx
@@ -1,7 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { makeStyles } from "@material-ui/core/styles";
-
 import Avatar from "@material-ui/core/Avatar";
 
 const useStyles = makeStyles((theme) => ({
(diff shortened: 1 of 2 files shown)


kmshin1397:
@acrellin Sorry if I stepped on your toes a little here - I didn't realize you were already looking into this. I saw it was unassigned and thought it would be good for me to take it as a native speaker.

acrellin (approved):
Awesome, thanks @kmshin1397 !

acrellin:
> @acrellin Sorry if I stepped on your toes a little here - I didn't realize you were already looking into this. I saw it was unassigned and thought it would be good for me to take it as a native speaker.

Not at all -- thanks a lot for taking care of this!

acrellin:
Will merge when green

PR #1577, opened 2021-01-09
Favorites frontend

guynir42:
The front end part of the "favorites" feature. 
A little star appears next to the source ID, and if you click it, the source gets added to the favorites list. 

favorites-frontend

Some stuff the still need doing: 
[ ] need to add the star to mobile source page, and maybe to other places like source table. 
[ ] if the source is in favorites, the star is not filled in until the entire page finishes loading. I don't know how to fix this... 
[ ] need to add some tests.


acrellin:
Looks great at first glance, thanks @guynir42 !  I'd love to see a star button added to the source table before this goes in. Will review more thoroughly soon

skyportal/handlers/api/source.py
```diff
@@ -707,6 +716,14 @@ class Validator(Schema):
...
         if saved_after:
             q = q.filter(Source.saved_at >= saved_after)
+        if list_name:
+            obj_list = Listing.query.filter(
+                Listing.list_name == list_name,
+                Listing.user_id == self.associated_user_object.id,
+            ).all()
+            obj_id_list = [obj.obj_id for obj in obj_list]
+            if obj_id_list is not None:
+                q = q.filter(Source.obj_id.in_(obj_id_list))
```
kmshin1397, line 726: I believe this will not use the db indices as well as if you used joins. There's a concept called "sargable" which describes which sorts of queries best use indices - basically, I think using "IN" is less optimal than using joins and WHERE statements (filter with SA). I would suggest doing something like the classifications where you join in `Listing` earlier in the handler and then filtering without the `in_`.

kmshin1397 (commented):
Okay I took a quick glance and don't see anything wrong with your approach. One suggestion for now, and I'll do a more thorough review once you add the button to more spots like Ari mentioned and add tests.

static/js/components/FavoritesButton.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+import PropTypes from "prop-types";
+
+import IconButton from "@material-ui/core/IconButton";
+import StarIcon from "@material-ui/icons/Star";
+import StarBorderIcon from "@material-ui/icons/StarBorder";
+
+import * as Actions from "../ducks/favorites";
+
+const FavoritesButton = ({ source_id }) => {
```
acrellin, line 12:
```suggestion
const FavoritesButton = ({ sourceID }) => {
```
throughout (camelCase in JS)

skyportal/handlers/api/source.py
guynir42: OK it looks a lot better now. I still don't see a big change in the responsiveness of the button. 
On the other hand, I only have one list and two sources... 
So I think something is going on in the JS side that takes too long until I get the favorites list in the state tree.

acrellin:
Needs tests

guynir42:
Sorry @acrellin I was sure I added these changes yesterday... probably I didn't commit again after the lint hook or something... It should be ok now!

static/js/ducks/sources.js
```diff
@@ -12,6 +12,9 @@ const FETCH_PENDING_GROUP_SOURCES = "skyportal/FETCH_PENDING_GROUP_SOURCES";
 const FETCH_PENDING_GROUP_SOURCES_OK =
   "skyportal/FETCH_PENDING_GROUP_SOURCES_OK";
 
+const FETCH_FAVORITE_SOURCES = "skyportal/FETCH_FAVORITE_SOURCES";
+const FETCH_FAVORITE_SOURCES_OK = "skyportal/FETCH_FAVORITE_SOURCES_OK";
+
```
kmshin1397, line 17: I think you want a REFRESH_FAVORITE_SOURCES websocket message here as well, so that the favorites page is also updated upon changes (right now only the star icons are updated as those rely on the favorites ducks but not the source table on the favorites page)

static/js/components/ClassificationForm.jsx
```diff
@@ -201,7 +201,7 @@ const ClassificationForm = ({ obj_id, taxonomyList }) => {
     );
   };
   CustomGroupsWidget.propTypes = {
-    value: PropTypes.string.isRequired,
+    value: PropTypes.arrayOf(PropTypes.string).isRequired,
```
kmshin1397, line 204: Nice!  Good catch - actually just made a story about this yesterday #1593. Thanks for the fix!

kmshin1397 (changes requested):
A few minor change that are needed, I think, but this is looking good.

static/js/ducks/sources.js
guynir42: Yes but updating the table automatically could be sort of annoying, as the MUI table will have to re-render whenever something is added / removed. So I think maybe it is better to have the favorites table only refresh on a reload, like the group sources page...

static/js/components/FavoritesPage.jsx
```diff
@@ -0,0 +1,102 @@
...
+  const dispatch = useDispatch();
+  const [sourcesRowsPerPage, setSourcesRowsPerPage] = useState(100);
+
+  useEffect(() => {
+    dispatch(sourcesActions.fetchFavoriteSources());
+  }, [dispatch]);
+
+  const sourcesState = useSelector((state) => state.sources.favorites);
+
+  const handleSourcesTableSorting = (sortData) => {
```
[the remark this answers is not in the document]
guynir42: I am not sure exactly how that works but I think I copied the logic correctly from group sources...

guynir42:
I've decided against having a dialog "are you sure you want to remove this favorite source". 
Instead I moved the favorites button into the pull-out row (only on the favorites page, in other tables it is still next to the object name). 

favorites_page_updates

guynir42:
Test failures appear to be unrelated...

acrellin:
@guynir42 master is green as of a few minutes ago, so I am concerned about these test failures as being related to the code changes here

guynir42:
@acrellin lets see what happens when I pull in master...

acrellin:
Sole failure appears unrelated

skyportal/handlers/api/source.py
```diff
@@ -707,6 +716,12 @@ class Validator(Schema):
             q = q.filter(Source.saved_at <= saved_before)
         if saved_after:
             q = q.filter(Source.saved_at >= saved_after)
+        if list_name:
+            q = q.join(Listing)
```
acrellin, line 720: I don't think you can add a join after applying filters -- I assume this is working in testing because none of the previous filter statements are in effect. This will need to be moved, as I don't think it will work at all as is when any of the previous filter blocks are being used

acrellin (changes requested):

acrellin:
Failure appears related: https://github.com/skyportal/skyportal/pull/1577/checks?check_run_id=1730969056#step:13:663

skyportal/handlers/api/source.py
guynir42: I moved the join up to where the classifications/taxonomies are joined. But there is still a call to filter on this query above these statements, so I am not sure what the difference is.

static/js/components/ClassificationForm.jsx
```diff
@@ -135,73 +133,69 @@ const ClassificationForm = ({ obj_id, taxonomyList }) => {
...
-          {options.enumOptions.length > 0 &&
-            options.enumOptions.map((group) => (
-              <MenuItem
-                value={group.value}
-                key={group.value.toString()}
-                data-testid={`notificationGroupSelect_${group.value}`}
-              >
-                {group.label}
-              </MenuItem>
+    <>
```
acrellin, line 157: @guynir42 here's the critical bug -- there's no return statement here

guynir42: Good catch! 

I removed the return statement because of the new lint rules about arrow functions with a single statement. But I needed to also remove the braces and semicolon. 
Now I'm back to where a few unrelated tests are still flakey...

static/js/components/ClassificationForm.jsx
```diff
@@ -97,32 +97,30 @@ const ClassificationForm = ({ obj_id, taxonomyList }) => {
 
   // Custom form widget for the classifications to format and display the contexts as well
   const CustomClassificationWidget = ({ value, onChange, options }) => {
```
kmshin1397, line 99: @guynir42 I think same thing about the return statement for here and the CustomProbabilityWidget as well. I think that's perhaps what's causing the `test_classifications` test to fail

acrellin:
@guynir42 also note that `eslint` has a `--fix` option that'll autofix some issues, including these arrow function body syntax issues

static/js/components/ClassificationForm.jsx
acrellin: Yep, this still needs to be fixed.

guynir42:
I am fairly sure the latest failures are unrelated.

kmshin1397 (commented):
Good to see the tests fixed! 

Can I ask the reason you moved to a text button to remove a source from the favorites page from the star? I assume you wanted to make it more obvious what would happen and help prevent accidentally clicking the star and losing the favorite? I think that in itself is fine, but maybe we could put the "Remove Favorite" into a new column which displays if `favoritesRemoveButton = true` instead of having it in the pullout row? Less clicks to get to it and I think having it hidden in the pullout feels less intuitive and could be cause for confusion. Having a delete button as the last column in a data table is pretty standard throughout the app (like the FollowupRequestsList) so I think that would be a better approach.

guynir42:
> Good to see the tests fixed!
> 
> Can I ask the reason you moved to a text button to remove a source from the favorites page from the star? I assume you wanted to make it more obvious what would happen and help prevent accidentally clicking the star and losing the favorite? I think that in itself is fine, but maybe we could put the "Remove Favorite" into a new column which displays if `favoritesRemoveButton = true` instead of having it in the pullout row? Less clicks to get to it and I think having it hidden in the pullout feels less intuitive and could be cause for confusion. Having a delete button as the last column in a data table is pretty standard throughout the app (like the FollowupRequestsList) so I think that would be a better approach.

I wanted to make the delete button a little harder to accidentally click. It seems to me that deleting the wrong source would be a big hassle as you'd need to go and look for it to bring it back. If you click the wrong row you might actually not even notice which source was removed. So this is my best idea for preventing accidental deletions. 

When you have to open the pullout row, that makes it a lot harder to click the wrong source, because you would open it and see all the details, and if you accidentally clicked the wrong row you'd notice. 

Putting it in the end of the row doesn't help prevent accidental deletions, and it would actually make it easier to click the wrong source.

kmshin1397 (approved):

kmshin1397:
> > Good to see the tests fixed!
> > Can I ask the reason you moved to a text button to remove a source from the favorites page from the star? I assume you wanted to make it more obvious what would happen and help prevent accidentally clicking the star and losing the favorite? I think that in itself is fine, but maybe we could put the "Remove Favorite" into a new column which displays if `favoritesRemoveButton = true` instead of having it in the pullout row? Less clicks to get to it and I think having it hidden in the pullout feels less intuitive and could be cause for confusion. Having a delete button as the last column in a data table is pretty standard throughout the app (like the FollowupRequestsList) so I think that would be a better approach.
> 
> I wanted to make the delete button a little harder to accidentally click. It seems to me that deleting the wrong source would be a big hassle as you'd need to go and look for it to bring it back. If you click the wrong row you might actually not even notice which source was removed. So this is my best idea for preventing accidental deletions.
> 
> When you have to open the pullout row, that makes it a lot harder to click the wrong source, because you would open it and see all the details, and if you accidentally clicked the wrong row you'd notice.
> 
> Putting it in the end of the row doesn't help prevent accidental deletions, and it would actually make it easier to click the wrong source.

Ah yeah okay sure I can see where you're coming from. My main concern was for when users were looking at just the basic data columns without expanding to see the plots and the remove functionality not being clear in that case, but I suppose most everyone will be looking at the plots and thumbnails when deciding which sources to remove anyways.

dmitryduev:
@guynir42 merge conflict, could you please resolve?

guynir42:
@dmitryduev conflicts resolved.

dmitryduev (changes requested):
Super cool @guynir42! I would love to have this feature in asap, I'll use it a lot!

I think @acrellin and @kmshin1397 did a great job in reviewing the code, so my requests are from the UI/UX point of view.

My main concern is the Remove Favorite button in the Favorites table -- it takes up way too much space and throws things off in the row (I removed the button for the first source below):
favs-1
I agree with your reasoning that it should not be easy to accidentally rm a fav, so I propose the following: add a new column to the table called "Actions" and place a red trash bin icon button there that upon clicking brings up a confirmation dialog. Once you click "confirm", it becomes disabled until the action returns. [Also, now when you click the remove button, it briefly turns into "add fav" button before disappearing with the row] @kmshin1397 what do you think?

Next, I can't click the star button in the Group Sources table without jumping to the Source page, I think that should be fixed:
favs-2

Finally - this is optional - what do you think about converting the star next to the source name into a separate column so that the users can sort by it (I think it should be possible to write a custom sort func for a column, should that be necessary)? If you think it's a good idea, then I'd implement that in a future PR.

guynir42:
@dmitryduev re: putting a trash can icon:
I know the way it is now is fairly ugly. I think we can improve. But the thing I most want to avoid is the confirmation dialog. It is evil and we all know that after you've seen that dialog a few times you no longer notice it, and click without thinking. 
I deliberately put the remove button inside the pull-out. This is the confirmation click that you really need, as it exposes the entire row and makes it much harder to remove the wrong source. 

What we can do is make it prettier. Maybe put the remove button at the end of the pull out row after all the other stuff is rendered. Maybe make it a trash can icon (but then I'm not sure if it's purpose would be obvious). 

Re the other stuff: I think putting the star in a different column would solve both problems. I will try it and we'll see how it goes. 

Thanks!

guynir42:
@dmitryduev I've put the star in a different column. It is hidden by default in the favorites page, but the user can bring it back. 

The remove favorites is only displayed on the favorites page. I put it at the end of the pullout row. 

There was some weird error when running the favorites page frontend test, I got errors when tearing down the public source fixture. This seems unrelated to my PR but worth a look.

kmshin1397:
@guynir42 @dmitryduev Regarding the remove favorites, I would say my preference is for what Dima describes with the confirm dialog, as that is most in line with the rest of the app and what I believe is standard practice for such tables. That said, I understand the desire to force users to be looking at the thumbnails/plots when removing a source and I think moving the remove button to the end did resolve the issue of it throwing off the other parts of the pullout row, so I'm fine with proceeding with this as is.

I think moving the star to a new column was a good idea, but you'll need to tweak the API handler to truly enable sorting on it @guynir42. You'll see that for the other columns which enable sorting, the source handler has query filtering behavior set based on the `sort_by` parameter. You'll need to add logic for the favorites in that part of the handler logic as well. Right now, the frontend table is displaying the arrows for sorting but no sorting actually happens. But as Dima said, I think that could be saved until a future PR.

Regarding the test failures on teardown for the favorites tests - do you still see this? I don't seem to see it locally and CI seems fine as well.

guynir42:
@kmshin1397 the errors in the teardown are still happening locally, but I see it is gone from the CI. So maybe it is just some issue with my database or something. 

I think we can move on with the PR then. 

Lets see if there's anything else we need to solve and merge this in... @dmitryduev

dmitryduev (approved):
Looks great, I think everyone's comments have been addressed, so let's merge it! 🚀

PR #1578, opened 2021-01-09
LCO Telescopes - Sinistro, Spectral, Muscat, Floyds APIs

mcoughlin:
LCO triggering interface. No tests currently committed as they require existing program to test (no response from LCO on powerless account).

### DIFF
--- a/skyportal/facility_apis/lco.py
+++ b/skyportal/facility_apis/lco.py
@@ -17,6 +17,19 @@ class SINISTRORequest:
 
     """A JSON structure for LCO 1m SINISTRO requests."""
 
+    def __init__(self, request):
+        """Initialize SINISTRO request.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        """
+
+        self.requestgroup = self._build_payload(request)
+
     def _build_payload(self, request):
         """Payload json for LCO 1m SINISTRO queue requests.
 
@@ -108,6 +121,20 @@ class SPECTRALRequest:
 
     """A JSON structure for LCO 2m SPECTRAL requests."""
 
+    def __init__(self, request):
+        """Initialize SPECTRAL request.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        """
+
+        self.requestgroup = self._build_payload(request)
+
+
     def _build_payload(self, request):
         """Payload json for LCO 2m SPECTRAL queue requests.
 
@@ -196,6 +223,19 @@ class SPECTRALRequest:
 class MUSCATRequest:
 
     """An XML structure for LCO 2m MUSCAT requests."""
+
+    def __init__(self, request):
+        """Initialize MUSCAT request.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        """
+
+        self.requestgroup = self._build_payload(request)
 
     def _build_payload(self, request):
         """Payload json for LCO 2m MUSCAT queue requests.
@@ -295,6 +335,19 @@ class MUSCATRequest:
 class FLOYDSRequest:
 
     """An XML structure for LCO 2m FLOYDS requests."""
+
+    def __init__(self, request):
+        """Initialize FLOYDS request.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        """
+
+        self.requestgroup = self._build_payload(request)
 
     def _build_payload(self, request):
         """Payload header for LCO 2m FLOYDS queue requests.
@@ -600,8 +653,8 @@ class SINISTROAPI(LCOAPI):
         if not altdata:
             return
 
-        ltreq = SINISTRORequest()
-        requestgroup = ltreq._build_payload(request)
+        lcoreq = SINISTRORequest(request)
+        requestgroup = lcoreq.requestgroup
 
         r = requests.post(
             'https://observe.lco.global/api/requestgroups/',
@@ -714,8 +767,8 @@ class SPECTRALAPI(LCOAPI):
         if not altdata:
             return
 
-        ltreq = SPECTRALRequest()
-        requestgroup = ltreq._build_payload(request)
+        lcoreq = SPECTRALRequest(request)
+        requestgroup = lcoreq.requestgroup
 
         r = requests.post(
             'https://observe.lco.global/api/requestgroups/',
@@ -829,8 +882,8 @@ class MUSCATAPI(LCOAPI):
         if not altdata:
             return
 
-        ltreq = MUSCATRequest()
-        requestgroup = ltreq._build_payload(request)
+        lcoreq = MUSCATRequest(request)
+        requestgroup = lcoreq.requestgroup
 
         r = requests.post(
             'https://observe.lco.global/api/requestgroups/',
@@ -935,8 +988,8 @@ class FLOYDSAPI(LCOAPI):
         if not altdata:
             return
 
-        ltreq = FLOYDSRequest()
-        requestgroup = ltreq._build_payload(request)
+        lcoreq = FLOYDSRequest(request)
+        requestgroup = lcoreq.requestgroup
 
         r = requests.post(
             'https://observe.lco.global/api/requestgroups/',
--- a/skyportal/facility_apis/lco.py
+++ b/skyportal/facility_apis/lco.py
@@ -514,8 +514,8 @@ class LCOAPI(FollowUpAPI):
         uid = content["id"]
 
         r = requests.post(
-            'https://observe.lco.global/api/requestgroups/{}/cancel/'.format(uid),
-            headers={'Authorization': 'Token {}'.format(altdata["API_TOKEN"])},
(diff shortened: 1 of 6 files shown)


mcoughlin:
Maybe @dannygoldstein @kmshin1397 can review this like LT?

dannygoldstein:
@mcoughlin Sure, ping us when ready.

mcoughlin:
@dannygoldstein this version is fine, but we need to discuss what you would like done about tests, if any.

dannygoldstein:
@mcoughlin - can you add these to demo data so that I can play around with the rendered forms?

mcoughlin:
@dannygoldstein You mean without the keys required to actually submit? Or do you want I send you those too separately?

dannygoldstein:
Hm, didn't think about that. Is there a test account at LCO?

mcoughlin:
@dannygoldstein no

dannygoldstein:
OK. Can we get one going? Will need one for tests and debugging.

mcoughlin:
@dannygoldstein they showed no interest in making such an account.

dannygoldstein:
I think we'll need one to move forward, or at least some way of validating that dummy requests get the expected output.

mcoughlin:
@dannygoldstein it would seem a bit disappointing if we are going to delay getting a key piece of our follow-up ecosystem into skyportal for this. can we discuss alternatives?

dannygoldstein:
@mcoughlin: @kmshin1397  will also need the creds

dannygoldstein:
@mcoughlin 

When I try to submit a request to SPECTRAL locally, I get:

```
Error: Bad Request for url: https://observe.lco.global/api/requestgroups/
```

mcoughlin:
@dannygoldstein Is it in the right hemisphere?

dannygoldstein:
Not sure - is there a way to introspect the error message provided by LCO? (I.e., "this request failed because ... ")

dannygoldstein:
object was 16fil -  21:30:52.53  +27:34:26.81

mcoughlin:
@dannygoldstein ok I pushed something like that now.. let's see.

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1032 @@
...
+            'type': 'ICRS',
+            'ra': request.obj.ra,
+            'dec': request.obj.dec,
+            'epoch': 2000,
+        }
+
+        exp_time = request.payload["exposure_time"]
+        exp_count = int(request.payload["exposure_counts"])
+
+        # The configurations for this request. In this example we are taking 2 exposures with different filters.
```
kmshin1397, line 63: The comment about a 2-exposure example is no longer relevant, no?

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1032 @@
...
+        if not altdata:
+            return
+
+        content = req.transactions[0].response["content"]
+        content = json.loads(content)
+        uid = content["id"]
+
+        r = requests.post(
+            'https://observe.lco.global/api/requestgroups/{}/cancel/'.format(uid),
+            headers={'Authorization': 'Token {}'.format(altdata["API_TOKEN"])},
```
kmshin1397, line 518:
```suggestion
            f"https://observe.lco.global/api/requestgroups/{uid}/cancel/",
            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
```
I think using the f formatting is preferable to format here for consistency with the rest of the codebase

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1032 @@
...
+        if not altdata:
+            return
+
+        content = req.transactions[0].response["content"]
+        content = json.loads(content)
+        uid = content["id"]
+
+        r = requests.get(
+            'https://observe.lco.global/api/requestgroups/{}/'.format(uid),
+            headers={'Authorization': 'Token {}'.format(altdata["API_TOKEN"])},
```
kmshin1397, line 572:
```suggestion
            f"https://observe.lco.global/api/requestgroups/{uid}/",
            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
```

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1032 @@
...
+        altdata = request.allocation.altdata
+        if not altdata:
+            return
+
+        lcoreq = SINISTRORequest(request)
+        requestgroup = lcoreq.requestgroup
+
+        r = requests.post(
+            'https://observe.lco.global/api/requestgroups/',
+            headers={'Authorization': 'Token {}'.format(altdata["API_TOKEN"])},
```
kmshin1397, line 620:
```suggestion
            "https://observe.lco.global/api/requestgroups/",
            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
```

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1032 @@
...
+
+        if not altdata:
+            return
+
+        lcoreq = SPECTRALRequest(request)
+        requestgroup = lcoreq.requestgroup
+
+        r = requests.post(
+            'https://observe.lco.global/api/requestgroups/',
+            headers={'Authorization': 'Token {}'.format(altdata["API_TOKEN"])},
```
kmshin1397, line 734:
```suggestion
            "https://observe.lco.global/api/requestgroups/",
            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
```

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1032 @@
...
+        altdata = request.allocation.altdata
+        if not altdata:
+            return
+
+        lcoreq = MUSCATRequest(request)
+        requestgroup = lcoreq.requestgroup
+
+        r = requests.post(
+            'https://observe.lco.global/api/requestgroups/',
+            headers={'Authorization': 'Token {}'.format(altdata["API_TOKEN"])},
```
kmshin1397, line 850:
```suggestion
            "https://observe.lco.global/api/requestgroups/",
            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
```

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1032 @@
...
+        altdata = request.allocation.altdata
+        if not altdata:
+            return
+
+        lcoreq = FLOYDSRequest(request)
+        requestgroup = lcoreq.requestgroup
+
+        r = requests.post(
+            'https://observe.lco.global/api/requestgroups/',
+            headers={'Authorization': 'Token {}'.format(altdata["API_TOKEN"])},
```
kmshin1397, line 956:
```suggestion
            "https://observe.lco.global/api/requestgroups/",
            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
```

kmshin1397 (commented):
I'm also seeing the SPECTRAL error that Danny mentioned:
image
Also, trying to delete the failed request from above also threw an error:
Screenshot 2021-01-14 093354

mcoughlin:
@kmshin1397 So basically what is going on is I need a good way to ensure that for Spectral, requestgroup['requests']['target']["dec"] is less than 17 degrees. Its not quite the same as the min and max in the form, can you think of a good way to ensure this? I think throwing an error afterwards doesnt seem to be a good solution.

kmshin1397:
> @kmshin1397 So basically what is going on is I need a good way to ensure that for Spectral, requestgroup['requests']['target']["dec"] is less than 17 degrees. Its not quite the same as the min and max in the form, can you think of a good way to ensure this? I think throwing an error afterwards doesnt seem to be a good solution.

Ah I see, yeah can't really use the form validation as it's taken from the obj itself. I would say the best approach would be to check for it and throw an error in the `_build_payload` functions so that the error occurs before the LCO API is actually hit, and we can send back a more informative message in that case.

mcoughlin:
@kmshin1397 great idea. Implemented.

kmshin1397:
> @kmshin1397 great idea. Implemented.

@mcoughlin Awesome. Looks like I lied earlier - FLOYDS seems fine to me now but still seeing MUSCAT fail. Is there a similar constraint for it we should apply?

kmshin1397:
Also still need to add handling for failed submissions in the delete method I think. The list index error would indicate it's probably failing on the `content = req.transactions[0].response["content"]` line (lco.py:514). I guess since we failed the submission there is no transaction there. It would still be good to be able to remove the failed record from the list of requests.

mcoughlin:
@kmshin1397 I have run out of time on this instrument in my program. Had to check my lco page to be sure.

mcoughlin:
@dannygoldstein How do you feel about the failed deletes?

dannygoldstein:
Same as the rest - we need to be able to test.

mcoughlin:
@kmshin1397 @dannygoldstein If I remove r.raise_for_status, I can get the tests to throw issues like the attached. Do we need the raise for status? 

Screen Shot 2021-01-16 at 3 59 54 PM

mcoughlin:
@dannygoldstein So checking the validates step, it returns 
status_code == 200
instead of
status_code == 201
but otherwise seems like it can be made to imitate submission OK. Is there a good way to determine from
env, cfg = load_env() 
if we are testing or not? If so, we could do just some kind of if then to check with url to query?

acrellin:
@dannygoldstein Any updates? Looks like @mcoughlin is waiting for feedback from you.

mcoughlin:
@acrellin @dannygoldstein @kmshin1397 I think we will redo this test with Kyung to use his interface (submitting to an active program and then deleting). But we need to sit down together or I send my info or something.

PR #1583, opened 2021-01-12
add migrations missing from #1432 and #1261

dannygoldstein:
Forgot to check for database migrations on #1432 and #1261. This PR adds them in.


dannygoldstein:
failure unrelated

stefanv:
Thanks, Danny.

We should probably simulate this in our test environment by taking the latest master, initializing the DB, and then running the test suite.

kmshin1397 (approved):

Issue #1528, opened 2020-12-21
Pin sphinx to 3.4.1 when it's released


acrellin:
3.4.0 has a critical bug whose patch will be included in 3.4.1. For now we've pinned to 3.3.1
https://github.com/sphinx-doc/sphinx/pull/8569

PR #1584, opened 2021-01-12
Pin sphinx to 3.4.3, skipping over broken release

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1528


stefanv:
This is the type of situation where we could probably have pinned to `>= 3.4` instead to allow more versions to co-exist.

It used the be pinned, though, so go ahead and merge when ready.

Issue #80, opened 2018-03-08
Finding chart generator


profjsb:
Add a button to generate (and cache) a finding chart for a given source.

The finder should show (by default):
  - an optical image of the field (5 arcmin on a side, PS1 or SDSS or DSS)
  - the location of three-five stars near the position
  - a calculation of the angular offsets from those stars to the source position

Should be viewable on the page and downloadable as a PDF.

A second version should allow the user to configure field size, orientation, number of stars, etc.

stefanv:
@profjsb I searched for Python packages that would generate these and found (ordered from what I consider to be most to least usable for our purpose):

- https://fpavogt.github.io/fcmaker/
- https://astroplan.readthedocs.io/en/latest/tutorials/plots.html#finder-image
- https://github.com/Fingel/fchart
- https://github.com/zkbt/the-friendly-stars

Would the output of any of these suffice?

stefanv:
Closed by #398.

PR #1591, opened 2021-01-13
Implemented rest wavelength axis on spectrum plots

scizen9:
Implemented initial version of rest wavelength axis on spectrum plots.  Remaining question: should it react to the redshift slider, or is that just for the atomic lines?

Also found that the hover on the spectrum plot does not work, so I commented it out.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -756,31 +756,35 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, height=300):
     smoothed_data = pd.concat(dfs)
 
     split = data.groupby('id')
-    hover = HoverTool(
-        tooltips=[
-            ('wavelength', '$x'),
-            ('flux', '$y'),
-            ('telesecope', '@telescope'),
-            ('instrument', '@instrument'),
-            ('UTC date observed', '@date_observed'),
-            ('PI', '@pi'),
-        ]
-    )
+    # hover = HoverTool(
+    #    tooltips=[
+    #        ('wavelength', '$x'),
+    #        ('flux', '$y'),
+    #        ('telesecope', '@telescope'),
+    #        ('instrument', '@instrument'),
+    #        ('UTC date observed', '@date_observed'),
+    #        ('PI', '@pi'),
+    #    ]
+    # )
     smoothed_max = np.max(smoothed_data['flux'])
     smoothed_min = np.min(smoothed_data['flux'])
     ymax = smoothed_max * 1.05
     ymin = smoothed_min - 0.05 * (smoothed_max - smoothed_min)
     xmin = np.min(data['wavelength']) - 100
     xmax = np.max(data['wavelength']) + 100
+    if obj.redshift is not None and obj.redshift > 0:
+        xmin_rest = xmin / (1.0 + obj.redshift)
+        xmax_rest = xmax / (1.0 + obj.redshift)
     plot = figure(
-        aspect_ratio=2,
+        aspect_ratio=1.5,
         sizing_mode='scale_width',
         y_range=(ymin, ymax),
         x_range=(xmin, xmax),
         tools='box_zoom,wheel_zoom,pan,reset',
         active_drag='box_zoom',
     )
-    plot.add_tools(hover)
+    # Does not work: scizen9 2020-01-13
+    # plot.add_tools(hover)
     model_dict = {}
     for i, (key, df) in enumerate(split):
         model_dict['s' + str(i)] = plot.step(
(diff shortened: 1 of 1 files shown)


skyportal/plot.py
```diff
@@ -756,31 +756,35 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, height=300):
...
+    # hover = HoverTool(
+    #    tooltips=[
+    #        ('wavelength', '$x'),
+    #        ('flux', '$y'),
+    #        ('telesecope', '@telescope'),
+    #        ('instrument', '@instrument'),
+    #        ('UTC date observed', '@date_observed'),
+    #        ('PI', '@pi'),
+    #    ]
+    # )
```
dannygoldstein, line 768: This works fine on my machine.

```suggestion
     hover = HoverTool(
        tooltips=[
            ('wavelength', '$x'),
            ('flux', '$y'),
            ('telesecope', '@telescope'),
            ('instrument', '@instrument'),
            ('UTC date observed', '@date_observed'),
            ('PI', '@pi'),
        ]
     )
```

skyportal/plot.py
```diff
@@ -756,31 +756,35 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, height=300):
...
+        aspect_ratio=1.5,
         sizing_mode='scale_width',
         y_range=(ymin, ymax),
         x_range=(xmin, xmax),
         tools='box_zoom,wheel_zoom,pan,reset',
         active_drag='box_zoom',
     )
-    plot.add_tools(hover)
+    # Does not work: scizen9 2020-01-13
+    # plot.add_tools(hover)
```
dannygoldstein, line 787: see above

dannygoldstein:
I take it back, the hover doesn't work for me either. But let's fix that in another PR. Best practice is to keep the changes introduced in a given PR to only what is relevant to the PR title / description

skyportal/plot.py
scizen9: OK.  Will submit another issue on the hover code.

skyportal/plot.py
scizen9: Uncommented, will submit separate issue.

scizen9:
Submitted issue #80 for hover code on fritz-beta-feedback.

skyportal/plot.py
```diff
@@ -793,6 +796,12 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, height=300):
     plot.xaxis.axis_label = 'Wavelength (Å)'
     plot.yaxis.axis_label = 'Flux'
     plot.toolbar.logo = None
+    if obj.redshift is not None and obj.redshift > 0:
+        plot.extra_x_ranges = {"rest_wave": Range1d(start=xmin_rest, end=xmax_rest)}
+        plot.add_layout(
+            LinearAxis(x_range_name="rest_wave", axis_label="Rest Wavelength (Å)"),
```
mcoughlin, line 802: Will "Å" always render for folks coded like this?

mcoughlin (commented):
The changes after Danny's comments look OK to me. Just the one question on the Angstrom symbol.

mcoughlin (commented):
The changes after Danny's comments look OK to me.

skyportal/plot.py
scizen9: Good question.  This was a cut and paste from the other axis label.  Since that other axis has been used for quite a while without issue, I would assume it will.

skyportal/plot.py
```diff
@@ -772,8 +772,11 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, height=300):
     ymin = smoothed_min - 0.05 * (smoothed_max - smoothed_min)
     xmin = np.min(data['wavelength']) - 100
     xmax = np.max(data['wavelength']) + 100
+    if obj.redshift is not None and obj.redshift > 0:
+        xmin_rest = xmin / (1.0 + obj.redshift)
+        xmax_rest = xmax / (1.0 + obj.redshift)
     plot = figure(
-        aspect_ratio=2,
+        aspect_ratio=1.5,
```
dannygoldstein, line 779: why change this ? separate PR?

dannygoldstein (approved):

mcoughlin (approved):

dannygoldstein:
@mcoughlin Looks like this is causing some LT-related frontend tests to fail?

mcoughlin:
@dannygoldstein looks to me like LT's system is just being slow.

PR #1594, opened 2021-01-13
Fixed spectrum hover in skyportal/plot.py

scizen9:
Changed line plot style for spectra from step to line to enable hover tool.  Modified tooltips struct to access wavelength and flux and formatted same.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -783,8 +783,15 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, height=300):
     plot.add_tools(hover)
     model_dict = {}
     for i, (key, df) in enumerate(split):
-        model_dict['s' + str(i)] = plot.line(
+        model_dict['s' + str(i)] = plot.step(
             x='wavelength', y='flux', color=color_map[key], source=ColumnDataSource(df),
+        )
+        model_dict['l' + str(i)] = plot.line(
+            x='wavelength',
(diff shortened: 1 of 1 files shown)


skyportal/plot.py
```diff
@@ -783,12 +783,8 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, height=300):
...
     model_dict = {}
     for i, (key, df) in enumerate(split):
-        model_dict['s' + str(i)] = plot.step(
-            x='wavelength',
-            y='flux',
-            color=color_map[key],
-            source=ColumnDataSource(df),
-            mode="center",
+        model_dict['s' + str(i)] = plot.line(
+            x='wavelength', y='flux', color=color_map[key], source=ColumnDataSource(df),
```
dannygoldstein, line 787: we need `step` to properly show the spectra, otherwise they end up looking weird.

dannygoldstein: this was requested in https://github.com/fritz-marshal/fritz/issues/111

scizen9:
Found a way to preserve step plot and enable hover tool.  I simply overplot with line, but with line_alpha=0.  Now all you see is the step plot, but the hover tool actually works.

skyportal/plot.py
scizen9: See latest commits that keep step plots.

PR #1595, opened 2021-01-14
Make source page more responsive

kmshin1397:
In this PR:

- Shorten placeholder text in Classifications form so that it doesn't overrun the width of the screen on mobile
- Refactor the AssignmentList and FollowupRequestLists components to use MUI DataTables instead of native tables, as those have built-in responsiveness. 
- Also added accordions for each instrument's follow-up request list and the assignments list so that users on mobile don't have to scroll so much to skip past the follow-up request/assignment lists to get to things below them. 
- Restructure the source info at the top of the page (classification, position, redshift, etc) so that things wrap more naturally on smaller screens (organized into a flexbox container and designated some items as belonging on the same line, etc) 
- Put styling on the Starlist component so that it doesn't explode beyond the bounds of the screen when loaded

**Follow-up**
Before:
followupBefore
After:
followupAfter

**Source info**
Before: 
sourceInfoBefore
After:
sourceInfo

**Starlist**
Before:
starlistBefore
After:
starlistAfter


dannygoldstein (approved):
Looks great. One comment - photometry plot looks a bit crowded on mobile - any way of giving a bit more breathing room here? 

Screen Shot 2021-01-15 at 11 13 33 AM

kmshin1397:
> Looks great. One comment - photometry plot looks a bit crowded on mobile - any way of giving a bit more breathing room here?
> 
> Screen Shot 2021-01-15 at 11 13 33 AM

Ah right, I'd forgotten I was going to look into that. Why don't I include that in a follow-up PR along with excluding the upload data button like we mentioned during standup today?

stefanv (approved):

profjsb:
Looks good. Minor thing: we should protect the position information from being broken across lines. Ie. the numerical values for ra, dec should be on the same line (same with l and b values).

profjsb (approved):
Small change to make if you can.

kmshin1397:
@profjsb Good idea. Implemented:
position

dmitryduev:
@kmshin1397 a minor comment: looks like there's unnecessary padding above the main Paper on smaller screens - could you please remove it?

kmshin1397:
@dmitryduev good catch. Surprisingly, the cause was a random ProfileDropdown component we were adding to mobile screens hidden there in that blank space.
Screenshot 2021-01-19 161122

dmitryduev:
@kmshin1397 👍🏻 Must have been a remnant of the old dropdown or something.

PR #1597, opened 2021-01-15
Fix permissions-checking bug affecting fetching of Candidates/Sources/Objs

### LINKED ISSUE #1535: "Insufficient permissions" to view sources
Users get "Insufficient permissions" when opening the source page of candidates from a filter for which they have stream and group access.

acrellin:
This fixes a critical bug affecting the fetching of Objs/Candidates.

The docstring for (`Source.get_obj_if_readable_by`) `get_source_if_readable_by` (which `Obj.get_if_readable_by` calls) states that it should  
> Return an Obj from the database if the Obj is a Source in at least
    one of the requesting User or Token owner's accessible Groups. If the Obj is not a
    Source in one of the User or Token owner's accessible Groups, raise an AccessError.
    If the Obj does not exist, return `None`.

This fix brings the actual functionality in line with what is expected and what is stated in the docstring.

Closes #1535

### DIFF
--- a/skyportal/handlers/api/comment.py
+++ b/skyportal/handlers/api/comment.py
@@ -3,7 +3,7 @@ from distutils.util import strtobool
 from marshmallow.exceptions import ValidationError
 from baselayer.app.access import permissions, auth_or_token
 from ..base import BaseHandler
-from ...models import DBSession, Source, Comment, Group, Candidate, Filter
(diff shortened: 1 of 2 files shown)


dannygoldstein:
Will merge on green.

dannygoldstein:
Hm..

E: The repository 'https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04  Release' no longer has a Release file.

PR #1601, opened 2021-01-19
Phase folding photometry

scizen9:
Initial draft of phase folding light curve plotting.  Requires 'period' to exist for object.  This draft includes data in the test instance of sky portal that includes a period and RRLyr light curve data for one of the test objects.

'Period' is implemented as a separate tab on photometry and is currently not show if 'period' is null.
The slider can be used to adjust period or one can enter period in text box.

1. Need to decide how periods are handled
2. Need to implement toggling of filter groups
3. Need to implement hover for plot
4. Need to decide how to handle period slider increment units (days, hours, minutes)


dmitryduev:
@scizen9 as discussed at the F iteration meeting, let's handle periods as annotations.
We may want to ditch the slider altogether and just have an input box for now.

skyportal/plot.py
```diff
@@ -687,7 +687,212 @@ def photometry_plot(obj_id, user, width=600, height=300):
...
+            tools='box_zoom,wheel_zoom,pan,reset,save',
+            y_range=(ymax, ymin),
+            x_range=(-0.1, 2.1),  # cover two phases: a and b
+            toolbar_location='above',
+            toolbar_sticky=False,
+            x_axis_location='below',
+        )
+
+        # axis labels
+        period_plot.xaxis.axis_label = 'phase'
```
profjsb, line 715:
```suggestion
        period_plot.xaxis.axis_label = r'phase ($\phi$) [radian]'
```

skyportal/plot.py
```diff
@@ -687,7 +687,212 @@ def photometry_plot(obj_id, user, width=600, height=300):
...
+        period_model_dict = {}
+
+        # iterate over each filter
+        for i, (label, df) in enumerate(split):
+
+            # fold x-axis on period in days
+            df['mjd_folda'] = (df['mjd'] % period) / period
+            df['mjd_foldb'] = df['mjd_folda'] + 1.0
+
+            # phase a plotting
```
profjsb, line 744: Seems like you could make this section more compact by iterating over `a` and `b` to plot the points/errorbars.

skyportal/plot.py
scizen9: Not familiar with this labeling.  I couldn't find any examples of this in the literature.  I don't believe radians is the correct unit since a full cycle should be 2 pi radians, no?

profjsb: the phase is in units of radians. A full cycle is indeed from 0 to 2pi. Phase going from 0-1 is sometimes shown but I do not think that it is as intuitive as radian units.

profjsb (changes requested):
The for loop starting on line 738, should be nested as suggested.

skyportal/plot.py
scizen9: Can you find me an example of this notation in the literature?  I cannot find any such example.  Uniformly, researchers in periodic variables use simple phase.  I worry that adopting a non-standard notation would confuse people, regardless of its correctness.

skyportal/plot.py
scizen9: I admit I am not very familiar with javascript and I don't quite know how to implement this suggestion over both python and javascript.  I will experiment with this.

scizen9: Got it.  The most recent commit should address this.

scizen9:
> The for loop starting on line 738, should be nested as suggested.

Done.  More compact as suggested.

dmitryduev (approved):
Awesome work, thanks @scizen9!

profjsb (approved):

PR #1599, opened 2021-01-16
Get object annotations

dmitryduev:
In this PR:

- `/api/sources/<obj_id>/annotations` endpoint to get an object's annotations.

PR #1611, opened 2021-01-21
Remove mistakenly added another get_obj_annotations_readable_by in models.py

dmitryduev:
It was added to `models.py` by mistake in #1599, but was already present there, with a different footprint, which is causing some issues on the frontend.


stefanv (approved):

stefanv:
Tests should probably have caught this.

dmitryduev:
@stefanv and I they did...
all green now, self-merging

PR #1621, opened 2021-01-25
Adjust source plot layouts based on device type

kmshin1397:
This PR updates the internal plot API to take an optional `device` parameter and adjust the layouts of the plots according to the type of device the plot is being requested on. It is set up to distinguish between `["browser", "mobile_landscape", "mobile_portrait", "tablet_landscape", "tablet_portrait"]`, with different plotting specifications based on the type. 

## Some example differences:
- Plot sizes are no longer hard-coded for mobile and tablet devices but calculated based on device screen width. 
  - Kept browser the same (fixed size with scrolling) because there will be a lot more window resizing and we don't want to have to reload the plot every time or have a weird size based on the initial window size stick around. 
- Plots are given a narrower aspect ratio on `"mobile_landscape"` to better match the aspect ratio of phones. 
- Mobile and tablet devices no longer expose irrelevant buttons like the "wheel_zoom" (no scroll-wheel mouse) or the "save" tools, nor the "Upload new data" buttons. 
- The number of columns to divide the spectral line checkboxes into is adjusted based on device
- Box zoom is turned off by default on mobile devices (it's still there just not active by default) because often on the smaller screens I found that I will touch the plot as I try to scroll on the page and instead accidentally zoom in on random parts of the plots. 

Note that right now there is no real difference between the plot layouts for "tablet_landscape" and "tablet_portrait" modes but I still distinguished the two in the enumeration of device types in case future additions to the plots require more granular tuning.

## Screenshots
Mobile (iPhone 8):
iphone
Tablet (iPad):
ipad


acrellin:
Only one of these followup request tests is failing on master (SEDM, as usual). Could be related?

acrellin (approved):
Looks great, thanks @kmshin1397

acrellin:
After reviewing, I don't see how the failures could be related to these changes, but thanks for the test update here anyways

acrellin:
Failures now appear to be limited to the typical flaky culprits.

dmitryduev (approved):
Looks great!

stefanv:
Eventually, we should aim to get away from fixed height and width specifications, and allow Bokeh to render according to the container size + aspect ratio.

PR #1623, opened 2021-01-25
Use New Permissions Framework on Annotations Handler 

dannygoldstein:
Updates the annotations handler to use the permissions checking infrastructure introduced in #1542

### DIFF
--- a/skyportal/handlers/api/annotation.py
+++ b/skyportal/handlers/api/annotation.py
@@ -2,7 +2,7 @@ import re
 from marshmallow.exceptions import ValidationError
 from baselayer.app.access import permissions, auth_or_token
 from ..base import BaseHandler
-from ...models import DBSession, Source, Annotation, Group, Candidate, Filter, Obj
+from ...models import DBSession, Annotation, Group, Obj
 
 
 class AnnotationHandler(BaseHandler):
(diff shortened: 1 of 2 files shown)


dmitryduev (approved):
So much cleaner!

skyportal/handlers/api/annotation.py
```diff
@@ -110,51 +115,6 @@ def post(self):
             return self.error("Input `origin` must begin with alphanumeric/underscore")
 
         annotation_data = data.get("data")
-
-        # Ensure user/token has access to parent source
```
guynir42, line 114: So what happens if the user doesn't have access to the source? it fails upon trying to post to the database?

guynir42 (approved):
I agree, this is a lot more concise... I want to go and see where this magic happens, is it somewhere inside baselayer?

skyportal/handlers/api/annotation.py
dannygoldstein: Yes

dannygoldstein:
@guynir42 it happens here - 

https://github.com/skyportal/skyportal/blob/master/skyportal/models.py#L2059
https://github.com/cesium-ml/baselayer/blob/master/app/models.py#L288
https://github.com/cesium-ml/baselayer/blob/master/app/handlers/base.py#L84

dannygoldstein:
Failure unrelated - was flaky on #1542

PR #1624, opened 2021-01-25
Use New Permissions Framework on Classifications, Comments, and Filter Handlers

dannygoldstein:
Updates the classification, comment, and filter handlers to use the permissions checking infrastructure introduced in #1542


skyportal/handlers/api/comment.py
```diff
@@ -118,47 +113,14 @@ def post(self):
...
-        group_ids = [int(id) for id in data.pop("group_ids", user_accessible_group_ids)]
-        group_ids = set(group_ids).intersection(user_accessible_group_ids)
+        group_ids = data.pop('group_ids', None)
         if not group_ids:
-            return self.error(
-                f"Invalid group IDs field ({group_ids}): "
-                "You must provide one or more valid group IDs."
-            )
-
-        # Only post to groups source/candidate is actually associated with
```
acrellin, line 138: Is there any assurance that specified groups are actually associated with the source?

acrellin (approved):
So much cleaner. Nice!

acrellin:
Failures are related

skyportal/handlers/api/comment.py
dannygoldstein: Since `Obj`s are `read=public`, no need to check this anymore. But if one wanted to impose a condition like this, it could be done in `GroupComment.create`.

dannygoldstein: That condition would be automatically enforced in `self.verify_permissions()`.

dannygoldstein:
Failures were due to tests grepping for specific text in API error messages, updated those tests.

dmitryduev (approved):
👍🏻

skyportal/handlers/api/comment.py
acrellin: I believe the reason we had added this check was that under certain conditions we were actually getting error responses for certain fetch queries. Hopefully we don't run into that again with the new access control infrastructure.

skyportal/handlers/api/classification.py
```diff
@@ -28,12 +28,9 @@ def get(self, classification_id):
               application/json:
                 schema: Error
         """
-        classification = Classification.get_if_readable_by(
-            classification_id, self.current_user
+        classification = Classification.get_if_accessible_by(
+            classification_id, self.current_user, raise_if_none=True
```
stefanv, line 32: How does this `raise_if_none` tie in with `self.error`?

skyportal/handlers/api/comment.py
dannygoldstein: Shouldn't be an issue because if a source exists, it will be fetchable by anyone. The only way this would error is if someone tries to fetch a nonexistent source. The only source info that is subject to non-public access controls are the attendant data (photometry, spectra, etc.).

skyportal/handlers/api/classification.py
dannygoldstein: I think it more or less produces the same effect. It causes an AccessError to raise which looks like: 

```
class AccessError(tornado.web.HTTPError):
    def __init__(self, reason, status_code=400):
        tornado.web.HTTPError.__init__(self, reason=reason, status_code=400)

    def __str__(self):
        return self.reason
```

I confirmed by looking at the logs that this produces the same log state (no python Exception in the logs) as calling self.error explicitly, e.g., 

[20:41:34 basehandler] Error response returned by [/api/sources/14gqr]: [Insufficient permissions.]

is the only log entry produced for an API call that triggers `raise_if_none`.

dannygoldstein:
@acrellin @stefanv Reviving this. I think this  is  ready to go in once the CI passes, if you guys feel that your comments have been addressed.

acrellin (approved):

acrellin:
@stefanv were your concerns addressed here? good to merge?

skyportal/handlers/api/classification.py
stefanv: Ah, OK, it's a bit confusing, because we try very hard not to raise exceptions from API calls.

Perhaps we can rename this to give a better sense of what's happening behind the scenes.  `error_if_none`?

stefanv:
@acrellin Just the one suggestion above; I find the `raise` terminology confusing.

PR #1626, opened 2021-01-26
Render markdown in browser notifications text

acrellin:
For now, just adds italicization to source/group/user names in message text.

image

Closes https://github.com/skyportal/skyportal/issues/1618

### DIFF
--- a/skyportal/handlers/api/comment.py
+++ b/skyportal/handlers/api/comment.py
@@ -189,7 +189,7 @@ class CommentHandler(BaseHandler):
                 DBSession().add(
                     UserNotification(
                         user=user_mentioned,
-                        text=f"{self.current_user.username} mentioned you in a comment on {obj_id}",
+                        text=f"*@{self.current_user.username}* mentioned you in a comment on *{obj_id}*",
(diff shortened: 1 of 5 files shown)


acrellin:
Updating associated tests

guynir42 (approved):
LGTM

profjsb:
Is there a place where the notification text gets limited in length? For example, if the message is really long, then we might want to take the X characters around the at-mention and add ellipses as appropriate:

`notification.text=`
"@ari you'll never guess what I found in this remarkable galaxy. It's a SN inside a comet..."

or 
`notification.text=`
"...this remarkable galaxy is a comet the shape of an H-R diagram. @ari it's just like you predicted..."

Also, I dont see where you handle multiple at-mentions in the same comment.

acrellin:
@profjsb Currently the notification text just says "_@josh_ mentioned you in a comment on _14gqr_"  without any of the actual comment text.

Regarding mentioning multiple users in a single comment, that's handled here:
https://github.com/skyportal/skyportal/blob/master/skyportal/handlers/api/comment.py#L188

PR #1627, opened 2021-01-27
Robustify comments tests

acrellin:


acrellin:
I don't think it's at all fixed, but hopefully this will help decrease flakiness a bit

skyportal/tests/frontend/test_sources.py
```diff
@@ -331,24 +322,19 @@ def test_super_user_can_delete_unowned_comment(
...
-    comment_text_div = driver.wait_for_xpath(f'//div[./p[text()="{comment_text}"]]')
-    comment_div = comment_text_div.find_element_by_xpath("..")
+    comment_text_p = driver.wait_for_xpath(f'//p[text()="{comment_text}"]')
+    comment_div = comment_text_p.find_element_by_xpath("../..")
     comment_id = comment_div.get_attribute("name").split("commentDiv")[-1]
 
     # wait for delete button to become interactible - hence pause 0.1
-    driver.scroll_to_element(comment_text_div)
-    ActionChains(driver).move_to_element(comment_text_div).pause(0.1).perform()
+    driver.scroll_to_element(comment_text_p)
```
stefanv, line 336: Maybe this should become a hover method in the baselayer test suite?

stefanv (approved):
Nice, thanks Ari.  I think in general we could do with more of this type of refactoring, where we put logic in one place that can then be fine-tuned.

PR #1632, opened 2021-02-01
Feature/1305 dark mode

### LINKED ISSUE #1305: Improve app color palette/dark mode
There are likely pages throughout the app where dark mode does not look as it should or just doesn't look good. Exploring enhancements to the overall color theme (i.e. a darker blue for the header?) would be good to do in conjunction. Potentially could expose the color theme to the configs as well.

joelachance:
Resolves #1305.

I updated and added to the `theme.palette` colors, and updated some of the static hex colors throughout to use more of the `theme.palette`. I also updated the color scheme [with this](https://coolors.co/e63946-f1faee-a8dadc-457b9d-1d3557). Of course, we can easily change this in this PR or in the future since these are mostly in the theme now. If anyone has a different preference, please post! I was also finding this color scheme has some quirks when it comes to accessibility (according to some MUI tools I stumbled across), so there's likely some tweaks we may want to make down the road if we stick with this color theme.

I'll note I wasn't able to find any issues with the current dark mode. I did remove the background colors we had in the Theme.jsx and let MUI do it's thing. I'm hoping this might solve some issues with the tables.

One warning: I'm not sure `make load_demo_data` is currently working for me as much of the dashboard is still pretty empty. Due to that, there's some things I wasn't able to test 100%, and I left any of the colors along that were obviously non-testable in my environment. 

Some of the colors I didn't get to either due to not being able to test or because they haven't been migrated into a Component yet:

```
CommentList.css
ClassificationList.css
Group.css
Main.css
NewsFeed.css/jsx
RunSummary.css
StarList.css
login.css
```

Here's a screenshot of the work:
Screen Shot 2021-01-31 at 10 57 58 PM


stefanv:
Thanks, Joe!  @acrellin is out this week, so apologies if reviews come in more slowly than usual.

I was hoping to modify the theme, depending on whether light or dark is set, and to not manually choose colors (but to stick with "x.secondary" etc.).  Do you think that is an option?

joelachance:
Not a problem at all, thanks for letting me know, @stefanv!

Reading through the Material UI docs, it seems MUI really wants to take care of switching colors itself. I might be misunderstanding your comment, but maybe you're referring to the `dark` and `light` key/value pairs in the primary and secondary theme objects? We can remove those and this should still work. My understanding is that those are to override what MUI is doing when it switches the color, so if you give it a primary color of `#333` and you don't specify the dark and light variations, it'll take it's best shot at the contrasting color. If you specify you can override this, which might be the right thing to do sometimes given accessibility concerns.
The ones I included in the theme were from a MUI tool I found regarding accessibility, but we can remove them.

I've tried to remove some of the static colors in styling throughout components and move towards using theme, but we're not quite there yet.

We can also simply have an empty palette and toggle the `type` key to either `light` or `dark`, and we'll get the default MUI theme.
Does this help?

Attached is a compressed video of what the theme looks like:
[skyportal-theme.mov.zip](https://github.com/skyportal/skyportal/files/5914947/skyportal-theme.mov.zip)

stefanv:
What I tried to say is that, ideally, we would define the color palette in one place, give names to colors that are not related to the colors themselves, and then use those all over the place.  That way, if you use "secondary highlight" as a color, we can easily swap out the palette and have the colors change accordingly.  A new color palette design can easily be installed simply by adjusting the colors in one place.

joelachance:
Ah, understood @stefanv. Based [on this](https://github.com/mui-org/material-ui/issues/14185), it looks like you can add custom colors without too much trouble, you just can't use them as props (e.g. `<Component color="custom-name" />`. So, I can update those to be non-color related, and put them inside the palette.

stefanv (approved):
Thanks, Joe, this looks great.

Who else wants to review; @kmshin1397 ?

kmshin1397:
@joelachance Do you mind merging in the latest commit on master so we can try running the CI properly on this? Thanks!

joelachance:
@kmshin1397 just merged master into my branch, hopefully this works

static/js/components/RecentSources.jsx
```diff
@@ -63,7 +63,7 @@ export const useSourceListStyles = makeStyles(() => ({
     fontSize: "1rem",
   },
   sourceTime: {
-    color: "gray",
+    color: theme.palette.info.main,
```
kmshin1397, line 66:
```suggestion
    color: theme.palette.text.secondary,
```
I found these quite hard to see in light mode (looks good on dark mode though!) because the info.main is a light color against the white background:
Screenshot 2021-02-09 130034
I'd suggest using the text color as those adjust based on "light" or "dark" type for the palette.

static/js/components/SourceDesktop.jsx
```diff
@@ -83,7 +83,7 @@ export const useSourceStyles = makeStyles((theme) => ({
   name: {
     fontSize: "200%",
     fontWeight: "900",
-    color: "darkgray",
+    color: theme.palette.gray[700],
```
kmshin1397, line 86:
```suggestion
    color: theme.palette.grey[700],
```
Not sure if this was working for you locally, but I had to change to this to get it to render correctly. Otherwise, got this error:
Screenshot 2021-02-09 125854

static/js/components/SurveyLinkList.jsx
```diff
@@ -26,7 +26,7 @@ const useStyles = makeStyles(() => ({
     padding: 0,
     margin: 0,
     fontWeight: "bold",
-    color: "gray",
+    color: theme.palette.gray[500],
```
kmshin1397, line 29:
```suggestion
    color: theme.palette.grey[500],
```

static/js/components/QuickSearchBar.jsx
```diff
@@ -11,20 +11,10 @@ import CircularProgress from "@material-ui/core/CircularProgress";
...
-    "& label.Mui-focused": {
-      color: "white",
-    },
-  },
-  label: {
-    "& label": {
-      color: "white",
-    },
+    background: theme.palette.primary.light,
+    borderRadius: '4px',
```
kmshin1397, line 17: Getting rid of the other styling seems to have made it good for dark mode, but now I feel that the light mode version could use some love:
Screenshot 2021-02-09 130454
Perhaps we can bring back the other styling rules about `label` and use `theme.palette.info.main` or something like that to get the search box to match the other sidebar links?

kmshin1397 (changes requested):
@joelachance This looks great! I have a few suggested changes - mainly to do with things that now look off in light mode now that they've been fixed for dark mode. 

As for the CI failures, I believe you've committed an old version of baselayer into this branch. I'd suggesting making sure your baselayer is pinned to whatever commit master is pinned to. Additionally, you should set up and run the pre-commit hooks to address the linter failures. You'll find instructions for that in the SkyPortal docs, I believe. I ran it on you branch and seems like it was all just cases of using single-quoted strings instead of double-quotes.

static/js/components/SourceDesktop.jsx
joelachance: I wasn't seeing errors before, but I'm surprised I wasn't, given `gray` doesn't exist! I'll take a look once this branch builds.

static/js/components/QuickSearchBar.jsx
joelachance: Yep! I wasn't happy with any versions of this, but agreed, I'd like to get this closer to the color of the other menu items.

static/js/components/RecentSources.jsx
joelachance: I didn't catch this one, but I agree, this isn't great. I'll update!

joelachance:
@kmshin1397 

> As for the CI failures, I believe you've committed an old version of baselayer into this branch. I'd suggesting making sure your baselayer is pinned to whatever commit master is pinned to. Additionally, you should set up and run the pre-commit hooks to address the linter failures. You'll find instructions for that in the SkyPortal docs, I believe. I ran it on you branch and seems like it was all just cases of using single-quoted strings instead of double-quotes.

I've run the following:

`git fetch upstream`
`git merge upstream/master`

and when that didn't do the trick, I tried:
`git rebase upstream/master`

The latter command in tandem with the git hook allowed the linter to run, but I'm still running into baselayer issues. Is there a way to pin the baselayer that I'm missing here?

Strangely, the linter didn't call out a single instance of single quotes. All linting errors were for Components I hadn't touched, and it was mostly complaining about using curlies with arrow functions with immediate returns. I fixed those and committed them, but I don't expect any of this to fix the build.

Let me know what you're thoughts are, and thanks for the in depth review!

kmshin1397:
@joelachance 

You can check the current baselayer version your branch is on by `cd`ing into the baselayer subdirectory and doing `git status`. To pin a particular commit for baselayer you would just do `git checkout <Hash>` within the baselayer folder. Generally, I think using `git submodule update` from the root skyportal directory will generally update baselayer to whatever commit the branch is pinned to, if you fall behind (though I don't think that would work in this case since you've already committed a another version into the branch). Normal use of `git pull` or `git merge` etc won't do the submodule update for you (you have to specify a certain option I think) so you should run the `git submodule update` whenever you find that baselayer shows up as modified in your `git status` - unless you've actually modified baselayer files yourself, of course.

As for the linter, did you run just ESLint  or the entire set of pre-commit hooks? I believe your particular issues were with Prettier and not ESLint. So I guess that was a bit of mislabeling on my part - should've said formatter and not linter. The ESLint issues you saw must have sneaked by us for whatever reason - thanks for fixing those! The pre-commit hooks only check the files specifically modified in the commit/PR so it can be easy for linting errors like that to slip by for awhile until they're modified again. If you want to run the same pre-commit hooks the CI will run you can go into the .github YAML files to look for the command used there (should be in the workflows that are not doc building). I believe it's something like `pre-commit run --from-ref <Latest master HASH> --to-ref <Your current commit>`.

Hope that helps!

stefanv:
Yes, so something like

```
pre-commit run --from-ref origin/master --to-ref HEAD
```

should work, presuming your upstream is called `origin`.

dmitryduev:
@joelachance I'm loving the new color palette, can't wait to have it merged and deployed!

Issue #1434, opened 2020-11-28
Reticles for ZTF image cutouts


dmitryduev:

guynir42:
closed by #1634

PR #1634, opened 2021-02-01
cross hair to all thumbnails

guynir42:
I added the little green cross-hairs (reticle) to all thumbnails, except SDSS which already has that. 

Screenshot from 2021-02-01 15-35-11

The old files in the demo data have a built-in cross-hairs so there's two sets in my instance of SkyPortal. I think this will go away for newer data where the thumbnails don't have anything on them.

### DIFF
--- a/static/js/components/ThumbnailList.jsx
+++ b/static/js/components/ThumbnailList.jsx
@@ -100,7 +100,7 @@ const Thumbnail = ({ ra, dec, name, url, size }) => {
             loading="lazy"
           />
         </a>
-        {(name === "dr8" || name === "ps1") && (
+        {name !== "sdss" && (
           <img
             className={classes.crosshair}
             src="/static/images/crosshairs.png"


dmitryduev (approved):
👍🏻

guynir42:
Closes #1434

stefanv:
I see the green crosshairs in the screenshot are misaligned; is this because they were not correctly placed on the static images?

guynir42:
> I see the green crosshairs in the screenshot are misaligned; is this because they were not correctly placed on the static images?

Yeah I think that was the reason, I don't know how to check if the new crosshairs are correct or the old ones...

stefanv:
How about taking a look at a known point source and making sure the cross-hairs intersect where they should?

guynir42:
> How about taking a look at a known point source and making sure the cross-hairs intersect where they should?

I double checked the position of the lines. They are on the middle pixel of each axis. How well these are placed on top of an image is a bit trickier. 

I am having some trouble posting a PNG from the command line: something about encoding is not sitting well when I send it as the JSON data.

PR #1650, opened 2021-02-01
Clean up test fixture database rows

### LINKED ISSUE #1602: Clean up test fixture DB rows upon going out of scope

kmshin1397:
Use yield fixtures (https://docs.pytest.org/en/stable/fixture.html#yield-fixtures-recommended) to implement teardown logic for each fixture which deletes fixture-generated database rows upon fixtures going out of scope.

- Each model factory defined in `fixtures.py` has a teardown function defined which deletes any data it generates, including any created by SubFactory fields.
- Fixtures that use the factories make sure to call the teardown functions for the factory they use upon getting control back from a `yield` statement.
- Update the Testing section of the developer docs to reflect this new fixture approach
- Fix some inconsistencies in foreign key constraint cascade behaviors (i.e. a column set to `CASCADE SET NULL` but also set to `nullable=False`) and also update some SQLAlchemy relationships to make sure relevant child rows are deleted properly (mostly adding `passive_deletes=True`). Migration script added for these.
- Fix `test_add_stream_add_delete_filter_group`. It was trying to add `public_stream` to `public_group` but the `public_group` fixture already adds the `public_stream` fixture upon instantiation. Not sure why the test was passing originally but it started erroring once the teardown stuff got added. Now the test tries to add `public_stream2` to the group instead.

Note: I think the logic in the fixture factory teardown functions could be streamlined more (a lot of places where I could leverage cascade delete behavior to delete rows without deleting them explicitly, for example) but I was running into some weird errors when I tried to do this so I'm just going to leave it like this for now where it definitely works and everything is definitely cleaned up. I guess it's just hard to make everything general enough to clean up correctly no matter what fixtures are combined + whether the tests call DELETE API calls themselves, etc. 

Closes #1602


kmshin1397:
LT test errors seem to be due to failures on the LT server side

skyportal/model_util.py
```diff
@@ -139,3 +138,12 @@ def create_token(ACLs, user_id, name):
     DBSession().add(t)
     DBSession().commit()
     return t.id
+
+
+def delete_token(token_id):
+    t = Token.query.get(token_id)
+    if t is not None:
+        DBSession().expire(t)
```
stefanv, line 146: What is this needed for?

skyportal/models.py
```diff
@@ -652,7 +654,7 @@ class Obj(Base, ha.Point):
     classifications = relationship(
         'Classification',
         back_populates='obj',
-        cascade='save-update, merge, refresh-expire, expunge',
+        cascade='save-update, merge, refresh-expire, expunge, delete-orphan',
```
stefanv, line 657: it would be really good to have these (`save-update`, `merge`, etc.) outlined in the docs, and why they are necessary in various forms (not necessary for this PR).

skyportal/tests/conftest.py
```diff
@@ -145,54 +166,79 @@ def take_screenshot_and_page_source(webdriver, nodeid):
...
     super_admin_user, group_admin_user, user, view_only_user, public_stream
 ):
-    return GroupFactory(
+    group = GroupFactory(
         users=[super_admin_user, group_admin_user, user, view_only_user],
         streams=[public_stream],
     )
+    group_id = group.id
+    yield group
+    GroupFactory.teardown(group_id)
```
stefanv, line 228: The pattern everywhere seems to be identical, so it might be possible to make this a generic fixture, something like:

```
def DBObjectFixture(obj_factory):
    obj = yield from obj_factory
    yield obj
    obj.__class__.teardown(obj.id)
```

And then:

```
@pytest.fixture
@ObjFixture
def group_with_stream(...):
    group = ...
    return group
```

stefanv (approved):
Thanks @kmshin1397, this looks good to me!  I left some thoughts, but no blockers.

skyportal/model_util.py
kmshin1397: Ah good catch I think it's something left over from previous iterations of this PR and not needed any more

dannygoldstein (approved):

dannygoldstein:
Excellent work

PR #1654, opened 2021-02-03
Redshift output with api/obj_id/classifications

Yashvi-Sharma:
Appended redshift key to all classifications for output of api/obj_id/classifications and added a test for it in test_classifications.py. This functionality was requested by the RCF group for easy and fast querying of RCF sources for BTS sample explorer page.


package.json
```diff
@@ -16,7 +16,7 @@
     "@rjsf/core": "2.4.0",
     "@rjsf/material-ui": "2.4.0",
     "autosuggest-highlight": "3.1.1",
-    "check-dependencies": "1.1.0",
+    "check-dependencies": "^1.1.0",
```
dannygoldstein, line 19:
```suggestion
    "check-dependencies": "1.1.0",
```

dmitryduev, line 19: Please keep `package.json` unmodified unless necessary (e.g., if you are adding a new dependency) -- please keep this pinned to this version.

skyportal/handlers/api/classification.py
```diff
@@ -324,6 +324,7 @@ def get(self, obj_id):
         classifications = obj.get_classifications_readable_by(self.current_user)
         for classification in classifications:
             del classification.groups
+            classification.redshift = obj.redshift
```
dannygoldstein, line 327: Our data model associates redshifts with Objs, not classifications. Can you return the `obj` as a nested property of the classification, rather than promoting redshift to a property of the classification?

dannygoldstein (changes requested):
Revert the baselayer update.

skyportal/handlers/api/classification.py
dmitryduev, line 327: I think this should be optional, controlled by a query parameter, such as:
```
- in: query
  name: includeRedshift
  nullable: true
  schema:
    type: boolean
    description: |
      Number of candidates to return per paginated request. Defaults to 25
```
Then `include_redshift = self.get_query_argument("includeRedshift", False)` and check on it before adding to the classification object.

dmitryduev (changes requested):
Good work, thanks @Yashvi-Sharma! See comment + please pin `baselayer` to the version that master is pinned to. Also, it's ok for now, but in future PR please create a branch on your fork with your feature, i.e. don't do the work on your master branch (see https://skyportal.io/docs/contributing.html and https://docs.fritz.science/developer.html#how-to-contribute)

dmitryduev:
@Yashvi-Sharma I agree with @dannygoldstein's comment, but still think it should be optional.

skyportal/handlers/api/classification.py
Yashvi-Sharma: Something like this?
`        for classification in classifications:
            del classification.groups
            classification.obj_data = {'redshift':obj.redshift}`

skyportal/handlers/api/classification.py
```diff
@@ -306,6 +310,13 @@ def get(self, obj_id):
             required: true
             schema:
               type: string
+          - in: query
+            name: includeRedshift
```
dannygoldstein, line 314:
```suggestion
            name: includeObj
```

skyportal/handlers/api/classification.py
```diff
@@ -306,6 +310,13 @@ def get(self, obj_id):
...
+            nullable: true
+            schema:
+              type: boolean
+              description: |
+                Option to include redshift output with classifications
```
dannygoldstein, line 319:
```suggestion
                Option to include nested Obj data with classifications
```

skyportal/handlers/api/classification.py
```diff
@@ -318,12 +329,15 @@ def get(self, obj_id):
         """
 
         obj = Obj.get_if_readable_by(obj_id, self.current_user)
+        include_redshift = self.get_query_argument("includeRedshift", False)
```
dannygoldstein, line 332:
```suggestion
        include_obj = self.get_query_argument("includeObj", False)
```

skyportal/handlers/api/classification.py
```diff
@@ -318,12 +329,15 @@ def get(self, obj_id):
...
         if obj is None:
             return self.error('Invalid obj_id.')
 
         classifications = obj.get_classifications_readable_by(self.current_user)
         for classification in classifications:
             del classification.groups
+            if include_redshift:
+                classification.obj_data = {'redshift': obj.redshift}
```
dannygoldstein, line 340: `obj_data` is not documented, does not contain all fields of Obj, and does not follow the naming conventions we use to refer to database models in nested output. The right way to go here would be to use `joinedload` along with the new permissions interface. But since that API refactor is still in progress, it's fine as a placeholder for now to  do:

```suggestion
            if include_obj:
                classification.obj = obj
```

skyportal/tests/api/test_classifications.py
```diff
@@ -262,3 +262,52 @@ def test_obj_classifications(
...
+        token=classification_token,
+    )
+    assert status == 200
+    status, objdata = api(
+        'GET', f'sources/{public_source.id}', token=classification_token
+    )
+    assert status == 200
+    assert data['data'][0]['classification'] == 'Algol'
+    assert data['data'][0]['id'] == classification_id
+    assert data['data'][0]['obj_data']['redshift'] == objdata['data']['redshift']
```
dannygoldstein, line 312:
```suggestion
    assert data['data'][0]['obj']['redshift'] == objdata['data']['redshift']
```

dannygoldstein (changes requested):

skyportal/handlers/api/classification.py
Yashvi-Sharma: I had tried this earlier but it gave this error in api output
`{'status': 'error',
 'message': 'Circular reference detected',
 'data': {},
 'version': '0.9.dev0+git20210208.b7194a7'}`

dannygoldstein:
Closing after discussing an alternative solution with @Yashvi-Sharma  based on reducing the verbosity of the sources.get handler when certain parameters are flagged.

PR #1655, opened 2021-02-04
handle phot ra,dec being sometimes being none in offfsets handler

dannygoldstein:
When a detection with null photometry coordinates  is used to calculate the best position for offset stars in https://github.com/skyportal/skyportal/blob/master/skyportal/handlers/api/source.py#L1309, the resulting `best_ra` and `best_dec`, will be `nan`. This can cause the offsets handler to fail and prevent StarList components from rendering properly.


dmitryduev (approved):
Thanks @dannygoldstein, I think it should fix this. It's a straightforward one-liner, so I don't think another review is necessary. Merging.

PR #1661, opened 2021-02-10
Loosen pip requirements

dmitryduev:
Some pinned reqs are causing trouble: https://github.com/fritz-marshal/fritz/runs/1867100185#step:6:2691 (conflict with https://github.com/cesium-ml/baselayer/blob/master/requirements.txt#L7)
Putting out this fire in this PR, but I suggest to loosen more requirements, otherwise we'll likely encounter similar problems in the future.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy==1.19.5
 scipy==1.6.0
 pandas==1.2.0
-dask==2020.12.0
+dask>=2020.12,<2021.3
 joblib==1.0.0
 seaborn==0.11.1
 bokeh==2.2.3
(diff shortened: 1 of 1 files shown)


stefanv:
I think we agreed to pin to minor versions, so say something like `>=1.18,<1.19` for NumPy.  How does that sound?

stefanv (approved):

dmitryduev:
@stefanv sure, np, as long as it does not case any issues. How about https://github.com/skyportal/skyportal/pull/1661/commits/2173555c295fd7737e97754bb3ccedaa244f5456 ?

stefanv:
That's perfect, thanks.  I tend to be on the adventurous side when it comes to pinning (I prefer less of it), but I understand others' concerns about things breaking.

dmitryduev:
Fixed the broken diff. Failure on the CI was unrelated (weather widget). @dannygoldstein should be mergeable.

PR #1666, opened 2021-02-12
Phase folding photometry

scizen9:
Minor addition to make phase 'b' invisible on startup.  Currently, it starts with 'One phase' selected and the correct x-axis limits, but the second phase 'b' is visible, which looks a bit wrong.


scizen9:
Still learning.  Ignore this one.

PR #1667, opened 2021-02-12
fix inital display of phase 'b'

scizen9:
Upon startup or hard refresh, phase plot shows both phase 'a' and phase 'b' even though 'One phase' is selected.  Plot x-axis limits are correct, but phase 'b' data visible.  This PR fixes this and initially sets phase 'b' data to visible=False.


dmitryduev (approved):
👍🏻

PR #1668, opened 2021-02-12
Mobile improvements to sources and candidates filtering and source quick view

kmshin1397:
# In this PR:
- Filtering forms for the sources tables and scanning page use the entire screen on smaller screens to prevent things being squished/going off the page and becoming unclickable
  - On the sources filtering form, submitting now replaces the form with a "Form submitted to server" message because MUI DataTable doesn't expose ability to manually close the form, and using their built-in submit button has issues with server-side mode. Replacing the full form with the message reduces the size of the form (esp on mobile) so that it doesn't cover up the table and the user can see that the table was filtered.
- Make source quick view for recent and top source widgets scroll naturally on mobile
- Minor adjustments to pagination controls on other MUI DataTables on mobile 
- Fix console warning with dashboard widget prefs

# Sources Table 
Before:
source_before
After:
sources_new

# Scanning page
Before:
candidates_before
After:
candidates_new

# Source Quick View
Before:
quickview_old
After:
quickview_new


scizen9 (approved):
Tested each of the improvements using developer console at 375 x 667px and each behaved as expected.  Much better navigation.

scizen9:
Just tried it on my ancient iPhone SE (very small 6) and everything worked even on that tiny screen (320x568).  The only thing I noticed was the SkyPortal label overlaps with the notifications bell.  This is probably not worth worrying about as most newer phones have enough real estate to avoid that particular problem.

PR #1673, opened 2021-02-16
Make including group users optional in GET /api/groups/<group_id>

dmitryduev:
Currently a GET call to `/api/groups/<group_id>` return the full list of group members with all their info, which could be hundreds of potentially unwanted entries.

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -143,23 +143,31 @@ class GroupHandler(BaseHandler):
             ) and group.id not in [g.id for g in self.current_user.accessible_groups]:
                 return self.error('Insufficient permissions.')
 
-            include_group_users = self.get_query_argument(
-                "includeGroupUsers", True
-            )
+            if self.get_query_argument("includeGroupUsers", "true").lower() in (
+                "f",
+                "false",
+            ):
+                include_group_users = False
+            else:
+                include_group_users = True
             # Do not include User.groups to avoid circular reference
-            users = [
-                {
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/group.py
```diff
@@ -149,9 +159,12 @@ def get(self, group_id=None):
...
                 }
                 for user in group.users
-            ]
+            ] if include_group_users else None
             group = group.to_dict()
-            group['users'] = users
+            if users is not None:
+                group['users'] = users
+            else:
+                group.pop("users", None)
```
acrellin, line 167: Is this necessary? I don't think relationships are loaded by default when calling to_dict

dmitryduev: True, removed.

acrellin (approved):

acrellin:
Looks great. Will merge when green

skyportal/handlers/api/group.py
```diff
@@ -136,22 +143,34 @@ def get(self, group_id=None):
             ) and group.id not in [g.id for g in self.current_user.accessible_groups]:
                 return self.error('Insufficient permissions.')
 
+            if self.get_query_argument("includeGroupUsers", "true").lower() in (
```
acrellin, line 146: when you pass in `includeGroupUsers=false`, does that get converted to the Python boolean `False`, or just the string `"false"`?

acrellin: nm, just checked, this looks good

Issue #81, opened 2018-03-08
Track multiple source positions


profjsb:
At the DB level, we should keep track of a few positions (and their uncertainties) of the source:

1) The source position at the time of discovery 
   - if this isn't in the initial packet as a separate field, then use the first detection value

2) A user updated position
   - default is to the use the initial source
   - a user may choose to update this over time (not allowed to move it from the original position by XX arcsec or more)

3) A *calculated* position
  - a weighted average position derived from the detections table -- this position will change as more candidates are added to the source.

dannygoldstein:
I believe this is more or less done - we track ra, dec, ra_dis, dec_dis, and the calculated positions are used in the finder chart generation

PR #1674, opened 2021-02-17
Sort spectra on date

scizen9:
Very simple change to address issue [#81](https://github.com/fritz-marshal/fritz-beta-feedback/issues/81) requesting that spectra be sorted on date

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -1032,6 +1032,8 @@ def spectroscopy_plot(
         data.append(df)
     data = pd.concat(data)
 
+    data.sort_values(by=['date_observed', 'wavelength'], inplace=True)
+
     dfs = []
     for i, s in enumerate(spectra):
         # Smooth the spectrum by using a rolling average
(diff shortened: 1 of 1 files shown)


dmitryduev (approved):
👍🏻

skyportal/plot.py
```diff
@@ -1032,6 +1032,8 @@ def spectroscopy_plot(
         data.append(df)
     data = pd.concat(data)
 
+    data.sort_values(by=['date_observed', 'wavelength'], inplace=True)
+
```
guynir42, line 1036: I'm not sure how sorting on wavelength works? I assume each spectrum has a whole array of wavelengths?

scizen9: Yes.  Without including wavelength in the sort, they end up scrambled.

scizen9: I tested having two spectra on the same date, but with different times and it sorted them correctly.  The hover in the spectrum plot only shows date, however, and this should be fixed in a different issue/PR (which I will log).

PR #1680, opened 2021-02-18
Migrations testing

### LINKED ISSUE #1662: Incorporate database migrations into CI
Likely will take the form of a new GA workflow where we checkout the last migration point, try running any new migrations on it, etc.

kmshin1397:
The workflow is only run when a change is detected in the `alembic/versions` folder and/or a change in `models.py` (thus should detect when someone forgets to add a migration at all when it's needed). 

Steps for testing:
- Check out latest master 
  - Should now be the on the last database revision
- Initialized the database and stamp it as head 
  - We now have a database presumably correct and at the latest version
- Check out the PR branch
- Try running the migration 
  - If no migration exists (i.e. schema changed without migration being added), this will do nothing
  - Will catch actual errors in the migration like duplicate columns, not respecting null constraints, etc.
- Try autogenerating a new migration script after the migration is run. 
  - This checks that all schema changes in the PR are actually covered in the new migration script. 
  - If there are no missing migration steps detected, the generated migration will just have `pass` blocks in the `upgrade`/`downgrade` functions, so we check for this as the success condition.
  - This generated migration script is uploaded as an artifact so that it can be reviewed for debugging.

Closes #1662


dmitryduev (approved):
Looks very nice, great work @kmshin1397!

acrellin:
@stefanv this is blocking another PR -- can you please review when you have a moment?

stefanv:
> @stefanv this is blocking another PR -- can you please review when you have a moment?

Sure, sorry about that.

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,157 @@
...
+          # Stamp Alembic HEAD
+          PYTHONPATH=. alembic -x config=config.yaml stamp head
+
+      - name: Checkout PR branch
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Formatting and linting checks
```
stefanv, line 126: No need for this here.

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,157 @@
...
+
+      - name: Check migration completeness
+        run: |
+          # Try generating a new migration
+          PYTHONPATH=. alembic -x config=config.yaml revision --autogenerate -m "Test migration"
+
+          # The generated migration should be empty if the migration script provided in the PR
+          # completely encompasses the changes in the database.
+          pass_lines=$( cat alembic/versions/*test_migration.py | grep -c pass )
+          if [[ $pass_lines -eq 0 ]]
```
stefanv, line 144: Should this be `-neq 2`?

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,157 @@
...
+          echo node $(node --version)
+          nginx -v
+          firefox --version
+
+      - name: Install SkyPortal dependencies
+        run: |
+          export PYTHONPATH=$PYTHONPATH:$(pwd)
+          make dependencies
+
+          pip install black
```
stefanv, line 89: Remove?

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,157 @@
...
+
+      - name: Test loading demo data
```
stefanv, line 91: Is demo data required for migration testing?  Seems like it may be, so just checking.

stefanv (commented):
Thanks, @kmshin1397, this looks good.

My only concern is that this does not exercise the migration manager, which is supposed to handle migrations automatically.

Perhaps on the *main* testing branch, we can do something similar to what you do here (check out master, initialize the database, then switch back) so we also get to see migration manager in action.

.github/workflows/test_migrations.yaml
kmshin1397: Black is needed because it's included as a post-generation step (in `alembic.ini`) for the migration autogenerating with alembic.

stefanv: Of course, yes!

.github/workflows/test_migrations.yaml
kmshin1397: Yeah I believe it's required. Didn't look too much into it, but if I never loaded in the demo data, the autogenerate step would generate a huge migration script with what seemed like the entire schema. The SP documentation on migrations also had something along the lines of "assuming you've run `make load_demo_data`" so I'm assuming we've seen this before

.github/workflows/test_migrations.yaml
kmshin1397: Good point

.github/workflows/test_migrations.yaml
kmshin1397: Removed

kmshin1397:
@stefanv Testing the migration manager is a good idea. I'll incorporate that

kmshin1397:
@stefanv Ok I think this is ready for re-review

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,151 @@
...
+        run: |
+          make run &
+          sleep 10 && make load_demo_data
```
stefanv, line 94: The sleep 10 shouldn't be necessary here; perhaps we should make the default timeout on load_demo_data longer.

stefanv (approved):
This is fantastic; thank you!

kmshin1397:
Summary of new updates:

- Instead of just trying the PR migration against the latest revision on master, we checkout the commit right before the first migration script when in, and replay all migrations in order including the new PR revision.
- The migrated database is compared against a fresh database using [migra](https://github.com/djrobstep/migra), as well as the check with the Alembic diff we were doing before
  - The diff migration script generated by migra is also uploaded as an artifact
- There is also a check for migration conflicts, both locally in the PR branch, and against master.
  - Check the `alembic history` of the PR branch and make sure that the latest revision on master is somewhere in that history to make sure the PR is up to date with the latest master database
  - Also checks to make sure there is only one `head` in the `alembic history`.

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,246 @@
+name: Test Skyportal Migrations
+
+on:
+  # Run only if changes in Alembic versions
+  pull_request:
+    paths:
+      - "skyportal/models.py"
```
acrellin, line 7: Might be worth tracking changes to the pinned baselayer/app/models.py, too

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,246 @@
...
+
+      - name: Check for migration branching
+        run: |
+          # Make sure latest revision on master is included in Alembic history
+          n_current_head=$( PYTHONPATH=. alembic history | grep -c $CURRENT_HEAD )
+          if [[ $n_current_head -eq 0 ]]
+          then
+            cat << EOF
+              The current head Alembic revision on master is not present in this branch's
+              Alembic history. Please merge in the latest commit on master.
```
stefanv, line 153: Just to be doubly clear, perhaps say "Please merge in the latest commit from master into this PR."

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,246 @@
...
+              Please resolve the branching revisions.
+          EOF
+            exit 1
+          else
+            echo "Alembic history has no conflicts in branch."
+          fi
+
+      - name: Run migration
+        run: |
+          # Update dependencies that may break migrations since the first commit, i.e. deepdiff
```
stefanv, line 175: What does this comment mean?

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,246 @@
...
+          make dependencies
+          PYTHONPATH=. alembic -x config=config.yaml upgrade head
+          PYTHONPATH=. alembic current
+
+      - name: Check migration completeness with Alembic
+        run: |
+          psql -U skyportal -h localhost -c "ALTER TABLE candidates DROP CONSTRAINT candidates_pkey;" skyportal
```
stefanv, line 182: Why do you drop the contraint here?

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,246 @@
...
+          kill %1
+          PYTHONPATH=. alembic -x config=config.yaml stamp head
+
+      - name: Compare databases with migra
+        run: |
+          pip install migra
+
+          # SQL needed to go from migrated database to fresh database
+          migra --unsafe postgresql://skyportal:anything@localhost:5432/skyportal postgresql://skyportal:anything@localhost:5432/skyportal2 > migra.sql || true
+          if [[ $(wc -l migra.sql | awk '{ print $1 }') ]]
```
stefanv, line 226:
```
if [[ $(cat migra.sql | wc -l) -ne 0 ]]
```

stefanv (commented):
Thanks, Kyung!  This looks good overall.  I have some small clarification requests.

GitHub Actions is working on reusable YAML blocks, which will really help us here; but they're not available yet.

.github/workflows/test_migrations.yaml
kmshin1397: I ran into an error when attempting to run the migration here without updating the dependencies. It was an error associated with the deepdiff package and seemed to have to do with some breaking change between whatever version of deepdiff is installed per old requirements.txt compared to more recent commits.  I didn't dig into it the exact causes though.

stefanv: I see, sort of, I think :)

Maybe just comment this as `# Update to latest dependencies`?

stefanv (approved):

.github/workflows/test_migrations.yaml
kmshin1397: Sure, that makes sense

.github/workflows/test_migrations.yaml
kmshin1397: I imagine it will be the same case here as I indicated in https://github.com/skyportal/skyportal/pull/1726#discussion_r585081969

kmshin1397: @acrellin Update here as well to just run on all changes to "baselayer"

acrellin: Great, I was worried about unnoticed baselayer model changes going in with a missing or incomplete migration script.

.github/workflows/test_migrations.yaml
```diff
@@ -0,0 +1,245 @@
...
+          echo "localhost:5432:*:skyportal:anything" > ~/.pgpass
+          chmod 600 ~/.pgpass
+
+          createdb -h localhost -U skyportal skyportal_test
+          psql -U skyportal -h localhost -c "GRANT ALL PRIVILEGES ON DATABASE skyportal_test TO skyportal;" skyportal_test
+
+          make db_init
+
+          pip list --format=columns
+          npm ls --depth 0
```
dmitryduev, line 128: Why is this needed?

dmitryduev (approved):
Looks great, awesome work @kmshin1397!

.github/workflows/test_migrations.yaml
kmshin1397: It's not really necessary - I mostly copied this section from existing workflows. I guess this was added for potential debugging purposes - say if some dependency upgrade ends up breaking the build for whatever reason.

kmshin1397:
@acrellin @dannygoldstein I think I've found an easy solution to the ENUM types problem. I can just run `make run` on the migrated database before checking the diffs, and this will run `refresh_enums()` and ensure the ENUM types are synced without doing any manual checking of the `migra` diff results for ENUMs. Also, there were a couple ENUM types missing from the list of `sqla_enum_types` defined in `enum_types.py` so I went ahead and updated that so that `refresh_enums` would properly refresh all ENUM types being used in the database.

Thoughts?

stefanv:
I'm lukewarm on the refres_enums idea.  I think it's fine to have as a safeguard, but ideally we should have these included in our migrations.  At least for forward migrations, I see no problem with creating new enums explicitly.

stefanv:
Also, running SkyPortal is a heavy hammer for only executing the enums refresh; it would be quite easy to run those programmatically.

kmshin1397:
Okay @stefanv yeah I see what you're saying. Then how about just doing only forward migrations and leaving the back migrations blank for ENUM types sound? And I'll remove the `refresh_enums` step in this workflow since it'll hide any missing forward migrations for ENUM type additions. And we can just keep it around as a safeguard like you said when actually running the app.

stefanv:
That sounds good, thanks!

acrellin:
That sounds good to me!

kmshin1397:
@stefanv @acrellin @dmitryduev Someone want to merge this now that #1704 is in?

acrellin:
🚀

stefanv:
NICE WORK!!!

This is a big reliability milestone.  :pray: :popcorn:

PR #1684, opened 2021-02-18
Improve sidebar menu item highlighting logic

### LINKED ISSUE #1683: Bolded sidebar link logic should be based on the current URL rather than the last-clicked sidebar link
Currently the last-clicked sidebar link stays highlighted when the user navigates to new pages:
sidebar_link_bold

acrellin:
Simplifies & improves sidebar link highlighting logic.
Closes #1683 
sidebar_link_bold_better

### DIFF
--- a/static/js/components/SidebarAndHeader.jsx.template
+++ b/static/js/components/SidebarAndHeader.jsx.template
@@ -97,11 +97,25 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 
+const SidebarLinkText = ({ route, title }) => {
+  const classes = useStyles();
+  const currentRoute = useLocation().pathname;
+
+  return (
+    <ListItemText
+      primary={
+        <Typography className={currentRoute === route ? classes.bold : null}>
+          {title}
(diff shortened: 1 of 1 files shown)


stefanv:
What do you think, Ari, is it worth building a lightweight wrapper component that unifies setting bold on click?  It'd be nice not to have to specify that check for every link individually.

acrellin:
> What do you think, Ari, is it worth building a lightweight wrapper component that unifies setting bold on click? It'd be nice not to have to specify that check for every link individually.

We could do that, e.g. add a `SidebarLinkText` component, but we'd still need to pass it the link's route and the current route so it can determine the style, so I don't know that it'd really simplify it at all. Thoughts?

stefanv:
I guess the component can get the current route by itself, so it would just need to receive the component's route.

acrellin:
@stefanv okay, done

stefanv (approved):

acrellin:
Okay to merge, @stefanv ?

Issue #1511, opened 2020-12-17
Add a GH Action to test building and running the Docker images


dmitryduev:
(and also loading the demo data)

dmitryduev:
We do that on F so closing.

kmshin1397:
@dmitryduev Aha, I remember that we closed this in the last iteration meeting. Do we want to revive it and add that SP workflow that you mentioned yesterday?

dmitryduev:
Yes, I think it's worth it. I would only run such action on commits/merges to master (not to waste the computational resources on all the PRs).

kmshin1397:
Yeah I think that makes sense

PR #1687, opened 2021-02-19
Add GA workflow for Docker build and update docs on Docker containers

kmshin1397:
Adds a new workflow to build and spin up SkyPortal using Docker, and tests it by adding demo data.

Also updates documentation to reflect issues I discovered while trying to spin up using Docker:
- Since the image runs `make run_production`, tables need to be manually created the very first time the images are spun up (since the docker volume will be empty). 
  - Added documentation describing this, and how to create the tables 
  - Touched up the `initial_setup.py` script (which I guess was used to add users before invitations was implemented) so that it can be used for creating tables in the Docker containers easily. 
- The existing documentation suggests using `docker-compose run` to run `make load_demo_data` in the SkyPortal container, but this doesn't seem to work anymore as it creates a new image from the SkyPortal container to run the command instead of running the command in the image already running in the SP docker-compose network. Fixed these to use `docker exec` instead.

Only runs on pushes to master to save computation resources.

Addresses #1511


dmitryduev (approved):
Looks great!

acrellin:
@stefanv thoughts here?

.github/workflows/test_docker_build.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Install Docker and Docker Compose
+        run: |
+          sudo apt-get update
+          sudo apt-get install \
+            apt-transport-https \
+            ca-certificates \
+            curl \
+            gnupg-agent \
+            software-properties-common
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
stefanv, line 42: Isn't docker installed on these images by default?

.github/workflows/test_docker_build.yaml
```diff
@@ -0,0 +1,71 @@
...
+          make docker-local
+
+      - name: Start containers
+        run: |
+          docker-compose up -d
+
+      - name: Create database tables
+        run: |
+          # Give plenty of time for containers to fully spin up
+          sleep 90
```
stefanv, line 66: Is there a way to query this rather, e.g. by using curl to query the sysinfo endpoint?

.github/workflows/test_docker_build.yaml
```diff
@@ -0,0 +1,71 @@
...
+          docker exec skyportal_web_1 bash -c 'source /skyportal_env/bin/activate && PYTHONPATH=$PYTHONPATH:"." python skyportal/initial_setup.py --config=config.yaml --username testuser@cesium-ml.org'
```
stefanv, line 67: @acrellin Would know more about the history of initial_setup.py.  I thought we didn't use it anymore.  Would `make db_init` followed by running the data loader not do the trick?

stefanv (commented):
Looks good, thanks @kmshin1397.  Just a few quick questions.

kmshin1397:
Oops accidentally deleted the branch for this

.github/workflows/test_docker_build.yaml
kmshin1397: Ah yes I think you're right and docker is already in there by default.

.github/workflows/test_docker_build.yaml
kmshin1397: Yeah I didn't think we were using `initial_setup.py` so I decided to repurpose it for initial setup of container volumes (hence the updates to the docs as well). The `make db_init` approach will not work here because `db_init.py` (in baselayer) actually only creates the DB but does not involve table creation. That's usually handled by `make run`, but since the Docker images run in production mode, the `make run` will not create the tables either. The data loader as well does not do any table creation as it just makes API calls to upload data.

.github/workflows/test_docker_build.yaml
kmshin1397: I think the issue with querying with sysinfo was that the app actually fails to spin up as there are no initial data tables/enumerate types in the image database, but they don't get created either by `make run` since it's in production mode, so then the app_server.py fails on trying to refresh enums, create the provisioned_admin token, etc. 

Might be possible to query the docker image directly with `docker logs` or something like that but just decided to take the straightforward approach here and just sleep. I can look into that alternative if you want though.

.github/workflows/test_docker_build.yaml
stefanv: Could initial_setup.py be replaced by the data loader?  If so, it already has some waiting logic built-in (can be improved).

If not, yes, it would be good to find a way to ping the system for being online.  I think you can connect to the migration manager and ask it for status, e.g.

kmshin1397: The solution I arrived at is to remove the sleep in the workflow and instead add waiting/retry logic into initial_setup.py. I don't think replacing it with the data loader is the best idea as it would involve logic pretty different from what is currently in the data loader. What we want to do is wait for the PostgreSQL database to be up and running and run `create_tables` in it, and not wait for the API to be up as the data loader currently does (as the API will fail to stand up properly until the tables are created, as I described in my other comment). Thoughts?

acrellin:
@stefanv thoughts here?

.github/workflows/test_docker_build.yaml
stefanv: We can run supervisord inside the container, or we can connect to the migration manager, e.g.?

.github/workflows/test_docker_build.yaml
stefanv: This seems related to https://github.com/skyportal/skyportal/issues/1802#event-4575401691 then

stefanv: @acrellin Any thoughts?

doc/deploy.md
```diff
@@ -38,6 +38,16 @@ docker-compose up -d
 
 Once both services are up and running, you may browse to SkyPortal at `http://localhost:9000`.
 
+Note that by default, the SkyPortal image will run the application in production (`make run_production`).
+
+The key behavior to note when running production mode is that the application will not create any database tables automatically (to avoid messing with production data). This means that the very first time you spin up SkyPortal using containers, the `skyportal_dbdata` volume will only have an empty database. You will need to set up the database tables manually. The easiest way to do this would be to run:
+
+```
+docker exec skyportal_web_1 bash -c 'source /skyportal_env/bin/activate && PYTHONPATH=$PYTHONPATH:"." python skyportal/initial_setup.py --config=config.yaml --username testuser@cesium-ml.org'
```
stefanv, line 46: @acrellin If you can look at this line too

stefanv (approved):

.github/workflows/test_docker_build.yaml
acrellin: I think it'd be better to move away from the otherwise deprecated initial_setup and use the data loader if possible, but I'm not opposed if the latter would be much more complicated

doc/deploy.md
acrellin: Same comment here

.github/workflows/test_docker_build.yaml
kmshin1397: Ok sure, let me see if I can rework the data loader to work for this purpose

kmshin1397:
@acrellin @stefanv I've refactored this PR to just add the functionality I need into the data loader and use that. The changes to the docs have been updated to reflect this as well. The initial_setup.py file was essentially reverted back to the way it is currently on master.

stefanv (approved):

acrellin (approved):

PR #1690, opened 2021-02-22
Explicitly create primary key for id column of all subclasses of JoinModel

acrellin:


kmshin1397:
@acrellin need to add explicitly add `candidates` to this as well right? Since it's not a subclass of JoinModel anymore

acrellin:
@kmshin1397 thanks, just added

acrellin:
The pre-existing primary keys were dropped (https://github.com/skyportal/skyportal/commit/4c835d6989d8c767e4a3ba5ef6c12d26d09d9bab#diff-a8b8ce53246045674f5fe1eb615544b731469eaa6b6455cac427ec112f697fe6R32) so it should be as simple as adding a new one, as is done in this PR

dmitryduev (approved):
Looks good to me! What happens if a PK already exists, would this throw an error, have you tried that @acrellin?

acrellin:
@dmitryduev yes, it errors in that case

dannygoldstein (commented):
Logic and code look correct. But I suggest merging this into 

https://github.com/skyportal/skyportal/blob/master/alembic/versions/eb5ad7587e4e_refactor_permissions.py

so that someone running through the migrations from the first commit would never run into an incorrect database state. 

this is how we handled a similar situation on https://github.com/skyportal/skyportal/commit/5f616766439b1f491ef67bdca6603fe3aec31a0d.

acrellin:
@dannygoldstein  would we need to somehow re-wind the migration history and re-apply all migrations from that point on if we simply updated that past script?

PR #1633, opened 2021-02-01
Interactive Periodogram

profjsb:
Allow the user to determine the period from a light curve in a selectable filter. The Lomb-Scargle periodogram is calculated on the clientside (in browser).

https://user-images.githubusercontent.com/480799/108526481-0d583200-7286-11eb-8f96-537b043377c0.mp4

TODO:
 - DONE ~~selectable time window to calculate period~~
 - FUTURE PR: use the graphics card to speed up the L-S calc
 - DONE ~~error checking on the photometry (not checking upperlimits now, e.g.)~~

PR #1696, opened 2021-02-23
Periodogram

profjsb:
See this [release video](https://berkeley.box.com/s/t2tif50884bac3qus4yhui08rfggczt7) for a preview of the features.

Reinstates the PR from #1633. Fixes for staging errors with multiple instruments.


stefanv:
@profjsb Could you explain the concept of an established filter?

profjsb:
> @profjsb Could you explain the concept of an established filter?

It's a boolean that captures whether the instrument and filter list has been loaded from the available photometry.

profjsb:
Addresses @acrellin's PR feedback in #1633.

acrellin (approved):
Looks good, thanks @profjsb

scizen9:
I'm seeing some conflicting behaviour.  When I select a time range, it calculates a new period automatically.
Hitting the 'Recalculate' button actually resets the time range to the entire data set and recalculates a new
period.  I would relabel the button 'Reset Range' and change the instructions to indicate that a new period is
calculated as soon as you select a time range.

scizen9 (changes requested):
I think I understand now.  The 'Recalculate' button is for the parameters, not for the period range.  As written, I thought I would select the range and then hit the button.  Just need to clarify the different behaviour for the parameters and the range in the text.

kmshin1397 (commented):
Looks good, @profjsb! The one issue I have is that the value displays on hover for each point are hard to see on the phased plots as it overlaps with the plot title. I took a quick glance through the Dygraphs documented options and it didn't seem like there was an easy way to move that hover display to the upper right corner instead of the upper left. Or perhaps we can add some `padding-bottom` to the plot title so that there isn't an overlap with the values?
image

profjsb:
> Looks good, @profjsb! The one issue I have is that the value displays on hover for each point are hard to see on the phased plots as it overlaps with the plot title. I took a quick glance through the Dygraphs documented options and it didn't seem like there was an easy way to move that hover display to the upper right corner instead of the upper left. Or perhaps we can add some `padding-bottom` to the plot title so that there isn't an overlap with the values?
> image

Fixed this by suppressing the legend (not really needed here for the workflow).

profjsb:
> I think I understand now. The 'Recalculate' button is for the parameters, not for the period range. As written, I thought I would select the range and then hit the button. Just need to clarify the different behaviour for the parameters and the range in the text.
@scizen9 made some formatting changes to address your concern.
image

scizen9 (approved):
This does the trick for me.

PR #1698, opened 2021-02-24
Use approximate photometry table count in DB Stats page for improved performance

### LINKED ISSUE #1692: Serve estimate for photometry table row count on DB stats page

acrellin:
Closes #1692 
Autovacuum is enabled by default, so we don't need to run `ANALYZE` as a cron job, as autovacuum already does that. Confirmation that we have autovacuum enabled on production:
```
fritz-1=> SHOW autovacuum;
 autovacuum               
------------              
 on                       
(1 row)                   
```

### DIFF
--- a/skyportal/handlers/api/db_stats.py
+++ b/skyportal/handlers/api/db_stats.py
@@ -8,7 +8,6 @@ from ...models import (
     User,
     Token,
     Group,
-    Photometry,
     Spectrum,
     CronJobRun,
 )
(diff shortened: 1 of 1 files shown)


acrellin:
This takes just milliseconds even on the current production table w/ 18 million rows

acrellin:
On production, 
```
In [10]: %timeit Candidate.query.count()                             
94.3 ms ± 654 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
so I think photometry is the only table that will need to use this approximation

dmitryduev (approved):
LGTM!

dannygoldstein:
What is the typical difference between the approximate and exact counts?

dannygoldstein (commented):
What is the typical difference between the approximate and exact counts?

stefanv (approved):

stefanv:
I do note that the postgres manual claims that counts are no longer slow starting with 9.2, and that you can use indices to get counts very fast: https://wiki.postgresql.org/wiki/Index-only_scans

kmshin1397:
@dannygoldstein I think that would just depend on the difference between the current state and the last time the autovaccum/analyze was run. Looking at the documentation and default settings for autovacuum analyzing I think Postgres will check every min and run the analyze if it sees the table size has changed by ~10% (the default `autovacuum_analyze_scale_factor`). It looks like we can change that config for individual tables though, so it may be worth tweaking that for the photometry since it does feel like 10% of 18 million photometry points is a big threshold. But we also wouldn't want to run it too often either I suppose. 


https://www.postgresql.org/docs/9.6/routine-vacuuming.html#AUTOVACUUM
https://www.postgresql.org/docs/9.6/runtime-config-autovacuum.html

kmshin1397 (approved):

PR #1700, opened 2021-02-25
Include creation time of oldest unsaved candidate on DB Stats page

acrellin:


dannygoldstein:
What's the motivation for this feature ?

dannygoldstein (approved):

acrellin:
To ensure that the deletion of old unsaved candidates is being done
correctly

stefanv:
I don't quite follow—how does this relate to deletion?

dmitryduev:
@stefanv old unsaved candidates should be vacuumed; displaying this will tell us when something is wrong. The page is current displaying the oldest candidate in the db, but that may or may not have been saved, so not very informative.

stefanv:
Oh, I thought there was some deeper hidden meaning :)  Yes, that makes good sense, thanks for the explanation Dima.

PR #1703, opened 2021-02-25
Upgrade redoc, mobx and styled-components (interdependencies)

acrellin:
Supersedes https://github.com/skyportal/skyportal/pull/1702


acrellin:
Self-merging, as this is just pinning package versions (dependabot's PR failed because of a few interdependencies)

PR #1704, opened 2021-02-26
Add migration for LT instruments

kmshin1397:
I discovered that there was never a migration script written for adding the Liverpool Telescope APIs to the ENUM type for `instrument.api_classname`. This PR introduced a migration for that purpose.

I guess we maybe added the values manually? Although I don't see LT instruments in the followup requests form on production either, so maybe they were never added at all? Regardless, the migration uses `IF NOT EXISTS` when adding the values so that if they were manually added already no error will be thrown when upgrading.

### DIFF
--- a/alembic/versions/3e1852612f34_add_lt_apis.py
+++ b/alembic/versions/3e1852612f34_add_lt_apis.py
@@ -1,0 +1,51 @@
+"""Add LT APIs
+
+Revision ID: 3e1852612f34
+Revises: c1354c411fab
+Create Date: 2021-02-25 19:10:55.114202
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '3e1852612f34'
+down_revision = 'c1354c411fab'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Add new Liverpool Telescope API values to the followup_apis ENUM type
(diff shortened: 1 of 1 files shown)


acrellin (approved):

dannygoldstein:
Sorry for the late response here @kmshin1397 . 

There is a bit of a subtlety here that extends to other ENUMs. 

Do we want to delete data when enum types are deprecated? 

For example, say I have enum type A with values (B, C). 

Then I add in another type so type A has (B, C, D).

Then I insert a bunch of rows with A = D. 

Then I run the migration backwards. Do I want to delete all the rows with A=D? That is what would happen in this case. 

To prevent those rows from being deleted, I implemented `refresh_enums` so that we only _add_ to enum types and never take values away.

Thoughts?

acrellin:
@kmshin1397 I think what you suggest around setting `api_classname` to NULL instead of deleting the instrument record is a good solution

kmshin1397:
Okay @acrellin, this is now modified to change rows for removed instrument APIs to NULL.

@dannygoldstein thoughts on this approach?

kmshin1397:
Now changed to just do nothing on the back migration so that we never remove values from the ENUM type

acrellin (approved):

stefanv:
Thank you, @kmshin1397

PR #1710, opened 2021-03-01
Update SQLAlchemy relationship cascade params in Obj.classifications

acrellin:
I had been seeing SQLA warnings saying something like "`delete` is required when specifying `delete-orphan` in `cascade`", and this is the only relationship in which we specify `delete-orphan`


kmshin1397 (approved):

acrellin:
@kmshin1397 wanna merge? I think something this small only needs one reviewer

stefanv:
We really should document these terms and how they apply to our code base ;)

Issue #1707, opened 2021-02-26
Create GA workflow to automatically label config changes, migrations, etc


kmshin1397:
Probably can use https://github.com/marketplace/actions/labeler

PR #1726, opened 2021-03-01
Add a GA workflow to automatically add labels to PRs

kmshin1397:
Check modified files in PRs and automatically add the following labels where relevant (some labels were already in the repository, some I just created):
- config-change
- demo-data
- dependencies
- docker
- documentation
- formatting-and-linting
- javascript
- migration
- needs-migration?
- python
- workflows

See inline comments for detailed specification of what triggers each label. Feel free to suggest more labels if you see come up with any.

Addresses #1707

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -54,3 +54,7 @@ needs-migration?:
 # Add 'python' label to any change to *.py files in the skyportal folder
 python:
   - skyportal/**/*.py
+
+# Add 'workflows' label to any changes to GA workflows (.github folder)
+workflows:
+  - .github/**/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -36,11 +36,6 @@ formatting-and-linting:
   - ".pep8speaks.yml"
   - ".pre-commit-config.yaml"
 
-# Add 'javascript' label to any change to *.js files in the static folder
-javascript:
-  - "static/js/**/*.js"
-  - "static/js/**/*.jsx"
-
 # Add 'migration' label to any change within the 'alembic' directory
 migration:
   - alembic/versions/*
@@ -50,10 +45,6 @@ migration:
 needs-migration?:
   - any: ["skyportal/models.py"]
     all: ["!alembic/versions/*"]
-
-# Add 'python' label to any change to *.py files in the skyportal folder
-python:
-  - skyportal/**/*.py
 
 # Add 'workflows' label to any changes to GA workflows (.github folder)
 workflows:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,52 @@
+# Labels to be automatically added to PRs
+
+# Add 'config-change' label to any change to default config files
+config-change:
+  - any: ["config.yaml.defaults", "test_config.yaml"]
+
+# Add 'demo-data' label to any change to provided demo data
+demo-data:
+  - data/**/*
+
+# Add 'dependencies' label to any change to dependencies
+dependencies:
+  - any: ["requirements.txt", "requirements.docs.txt", "package.json"]
+
+# Add 'docker' label to any change to files related to Docker deployment
+docker:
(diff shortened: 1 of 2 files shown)


kmshin1397:
PR on my fork where I tested this, if you're interested: https://github.com/kmshin1397/skyportal/pull/3

.github/labeler.yml
```diff
@@ -0,0 +1,60 @@
...
+  - "static/js/**/*.jsx"
+
+# Add 'migration' label to any change within the 'alembic' directory
+migration:
+  - alembic/versions/*
+
+# Add 'needs-migration?' label to any change to models.py without changes to
+# the Alembic versions
+needs-migration?:
+  - any: ["skyportal/models.py"]
```
acrellin, line 51: any way to see if the pinned baselayer/app/models.py was touched, too?

kmshin1397: Ah yeah good point - I'll look into it

stefanv:
This is cute, thanks!

We should think about which labels are helpful in development.  I think, e.g., that `python` and `javascript` are not.  Config changes and migrations clearly are important.  How about the others?  What is the motivation for each?

kmshin1397:
@stefanv Here are my thoughts on the other labels. I'm definitely open to discussion here:

- demo-data
  - Developers may see that a `demo-data` labelled PR has recently been merged, and to rerun `make load_demo_data` so that their development environment has test data in line with what is on master. 
- dependencies
  - The dependabot PRs already use this label automatically, I believe. So I added this to match that behavior. I imagine it's good for reviewers to immediately see when new dependencies are being added/updated.
- docker
  - I imagine if there are developers using Dockerized deployments locally for development, they may want to quickly see if any recent PRs include changes to those configurations and decide if they want to rebuild local images? I was also thinking it would be good to emphasize if there are Docker/Docker Compose changes to nudge reviewers to actually try building/running them but I guess that's less of a reason once #1687 is merged. 
- documentation
  - No profound reason - this label already existed (probably to tag issues) so I decided to support it since it's already there
- formatting-and-linting
  - If we ever decide to change certain style rules/guidelines and modify linting/formatting rules accordingly it would be good to highlight that PR so developers can weigh in and be aware of those changes, as all developers are expected to adhere to these style rules.
- javascript
  - This was also something that was already there so I just included it. It seems like it might just be a label used by dependabot? 
- needs-migration?
  - Although #1680 will error if migrations are actually missing, this tag will provide a very clear human-readable message to the developer that they may need to add a migration script without having to go into the GA run and see what actually threw the red X. To be fair, this will also include false positives in that any changes (even innocuous ones like comments) to `models.py` without a subsequent migration will trigger the label. It's why I have the question mark, but we could also just forgo this label.
- python
  - Same thing with the `javascript` tag 
- workflows
  - Honestly just added as an afterthought for completeness as it was the only top-level directory in the repository that didn't really have a label associated with it.

stefanv:
Thanks for that thorough justification!  I'd remove the javascript and Python tags, then we see how it goes with the rest.

kmshin1397:
> Thanks for that thorough justification! I'd remove the javascript and Python tags, then we see how it goes with the rest.

Sounds good!

.github/labeler.yml
kmshin1397: @acrellin Hmm I don't think this is straightforward to do. I enabled the debug logs for this action, and the only thing it gets back from the github API for files changed is `baselayer` and there is no info of what changed in the submodule.

acrellin:
@kmshin1397 in that case, can we add `needs-migration?` whenever baselayer changes so that we can make sure there are no changes there to DB models? Then it can be manually removed after someone has checked

kmshin1397:
@acrellin Sure, that can be done

kmshin1397:
@stefanv I've removed the javascript and python tags here. Are you good with this going in?

stefanv:
Thanks!

PR #1730, opened 2021-03-02
Limit size of Joblib memory cache

stefanv:
See also https://github.com/joblib/joblib/issues/574


acrellin (approved):

profjsb (approved):
Since `reduce_size` is called first, technically the size of the cache could grow to larger than the max size as a new file is added. But I think it's fine as written. The test for this functionality is done in joblib (https://github.com/joblib/joblib/blob/49a2c10bfb6b47e4ca60d9c1297675e79f2c2e6e/joblib/test/test_memory.py#L913)

acrellin:
@profjsb @stefanv is this ready for merge?

PR #1741, opened 2021-03-04
Add sample query parameter request to API docs header

### LINKED ISSUE #1734: Add sample request using query parameters to API docs

acrellin:
Closes #1734

### DIFF
--- a/skyportal/api_description.md
+++ b/skyportal/api_description.md
@@ -34,6 +34,23 @@ if response.status_code in (200, 400):
 curl -s -H 'Authorization: token ea70a5f0-b321-43c6-96a1-b2de225e0339' http://localhost:5000/api/sysinfo
 ```
 
+#### Using query parameters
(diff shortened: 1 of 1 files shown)


skyportal/api_description.md
```diff
@@ -34,6 +34,23 @@ if response.status_code in (200, 400):
...
 
+#### Using query parameters
+
+```python
+import requests
+
+token = 'ea70a5f0-b321-43c6-96a1-b2de225e0339'
+
+response = requests.get(
+    'http://localhost:5000/api/candidates?numPerPage=100&pageNumber=1',
```
stefanv, line 45: Better to use `params` keyword?

acrellin: That's `requests`-specific, but I just want to illustrate how to include them in the URL generally, as this extends to requests generated via other means

PR #1750, opened 2021-03-05
Allow group admins to manage other group members

acrellin:
This PR updates the permissions regarding modifying group members' admin status such that any group admin (regardless of ACLs) can grant/revoke the admin status of other group members, or remove users entirely.

Closes https://github.com/fritz-marshal/fritz/issues/191

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -28,7 +28,9 @@ def has_admin_access_for_group(user, group_id):
         .first()
     )
     return len(
-        {"System admin", "Manage groups"}.intersection(set(user.permissions))
+        {"System admin", "Manage groups", "Manage_users"}.intersection(
+            set(user.permissions)
+        )
     ) > 0 or (groupuser is not None and groupuser.admin)
 
 
@@ -500,10 +502,7 @@ class GroupUserHandler(BaseHandler):
     def patch(self, group_id, *ignored_args):
         """
         ---
-        description: |
-          Update a group user's admin status.
-          Group admins can grant other group members admin status, but cannot
-          revoke it or remove members unless they have sufficient ACLs.
+        description: Update a group user's admin status
         tags:
           - groups
           - users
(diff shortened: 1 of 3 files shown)


stefanv:
That's an odd asymmetry; who requested this?

acrellin:
@stefanv Any group admin can already add users & make them an admin in the process via NewGroupUserForm, but they can't update their admin status after the fact or remove them from the group to be able to re-add them as an admin. That seems a bit backwards to me (and there's a linked issue that was complaining about this), thus this PR.

stefanv:
I just think, in general, if users can do something they should be able to undo it.  Otherwise they accidentally click something, and then need to get help to fix it.

acrellin:
@stefanv but that currently doesn't apply to adding new group users, because only those with particular ACLs can remove users from groups

acrellin:
@stefanv but yes, I agree in general, but apparently at least one PI doesn't 😉   thus the weird requirements here.

stefanv:
Maybe we should mention to them what awkwardness this poses and ask them to re-assess.

acrellin:
@stefanv  I can see the case for it here -- you may want to grant group admin status to a member so they can take care of certain tasks (e.g. reviewing save requests) but not let them go rogue and revoke *your* admin status.

stefanv:
Then maybe our permission granularity should be increased?  Asymmetric permissions just seem like they're going to cause headaches.

acrellin:
But there are already plenty of areas where users are able to post data but
not necessarily delete them, right?

stefanv:
Really?

acrellin:
@stefanv it sounds like your objection is actually to what's already implemented, not this PR then. Let's discuss in the next meeting.

stefanv:
Yes, I was quite surprised!  But you're absolutely right: there's no problem with the PR, we just need to figure out if this, in general, is the way to go.

acrellin:
@stefanv  I've done away with the asymmetrical permissions here

stefanv (approved):

kmshin1397 (approved):

dmitryduev (approved):

PR #1751, opened 2021-03-08
Update scanning page annotations tooltip to note known limitation

### LINKED ISSUE #1185: Annotation filtering on scanning breaks if applying multiple filters with different origins
Cannot have multiple filters set on annotation fields from different origins at the same time.

For example, you cannot filter for some_key (origin A): true and some_other_key (origin B): true at the same time even if a candidate has both. This will just result in no results returned. 

Vague impression is that this will require a bigger change to the sqlalchemy code for the candidates handler and more thinking than a quick change. Something like merge annotation rows into one big jsonb field so that one row knows it has both origins? Right now the queries technically look for entries in the Annotation table that have both fields instead of objects with entries in the Annotation table that fit the fields.

kmshin1397:
Annotation filtering on the scanning page does not support filtering across multiple annotation origins. As this doesn't seem to be a common use case, it seems fine to just make note of the limitation within the column tooltip.
Closes #1185
image

### DIFF
--- a/static/js/components/CandidateList.jsx
+++ b/static/js/components/CandidateList.jsx
@@ -759,7 +759,7 @@ const CandidateList = () => {
               Warning: applying multiple filters on annotations from different
               origins is not supported currently and will return zero results.
               For example, you cannot filter for a specific annotation value in
-              annotations from both &quot;orgin_a&quot; and &quot;origin_b&quot;
+              annotations from both &quot;origin_a&quot; and &quot;origin_b&quot;
               at the same time.
             </i>
           </Typography>
--- a/static/js/components/CandidateList.jsx
+++ b/static/js/components/CandidateList.jsx
@@ -752,6 +752,16 @@ const CandidateList = () => {
             <b>Sorting: </b> Clicking on an annotation field will display it, if
             available, in the Info column. You can then click on the sort tool
             button at the top of the table to sort on that annotation field.
+            <br />
+            <b>Filtering: </b> Filtering on annotations is available through the
+            filtering tool at the top right of the table. <br />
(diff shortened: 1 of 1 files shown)


static/js/components/CandidateList.jsx
```diff
@@ -752,6 +752,16 @@ const CandidateList = () => {
             <b>Sorting: </b> Clicking on an annotation field will display it, if
             available, in the Info column. You can then click on the sort tool
             button at the top of the table to sort on that annotation field.
+            <br />
```
dmitryduev, line 755: Could you please fix the wording on lines 749-750 above? '"they" will be considered' or "That is, two annotation values belonging to..."

dmitryduev (approved):

static/js/components/CandidateList.jsx
```diff
@@ -752,6 +752,16 @@ const CandidateList = () => {
...
+            <b>Filtering: </b> Filtering on annotations is available through the
+            filtering tool at the top right of the table. <br />
+            <i>
+              Warning: applying multiple filters on annotations from different
+              origins is not supported currently and will return zero results.
+              For example, you cannot filter for a specific annotation value in
+              annotations from both &quot;orgin_a&quot; and &quot;origin_b&quot;
```
stefanv, line 762:
```suggestion
              annotations from both &quot;origin_a&quot; and &quot;origin_b&quot;
```

static/js/components/CandidateList.jsx
kmshin1397: Good catch!

acrellin (approved):

PR #1755, opened 2021-03-10
Allow redshift to be nullified in source page component

acrellin:
nullify_redshift

Closes https://github.com/fritz-marshal/fritz-beta-feedback/issues/89

### DIFF
--- a/static/js/components/UpdateSourceRedshift.jsx
+++ b/static/js/components/UpdateSourceRedshift.jsx
@@ -116,7 +116,7 @@ const UpdateSourceRedshift = ({ source }) => {
               startIcon={<ClearIcon />}
               size="large"
               data-testid="nullifyRedshiftButton"
-              disabled={isSubmitting || invalid}
+              disabled={isSubmitting || source.redshift === null}
             >
               Set null
             </Button>
--- a/static/js/components/UpdateSourceRedshift.jsx
+++ b/static/js/components/UpdateSourceRedshift.jsx
@@ -9,6 +9,7 @@ import Button from "@material-ui/core/Button";
 import { makeStyles } from "@material-ui/core/styles";
 import SaveIcon from "@material-ui/icons/Save";
 import ClearIcon from "@material-ui/icons/Clear";
+import Tooltip from "@material-ui/core/Tooltip";
 import TextField from "@material-ui/core/TextField";
 
 import { showNotification } from "baselayer/components/Notifications";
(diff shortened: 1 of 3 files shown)


stefanv:
Ari, would it be hard to have a "clear" button next to SAVE instead of having the users specify "null"?  It feels like we're exposing the underbelly of the beast a bit too much :)

dmitryduev:
I agree with @stefanv, that'll make it more user-friendly.

acrellin:
@stefanv @dmitryduev agreed, much better (and simpler solution). Done: 
nullify_redshift

Ready for review.

stefanv:
@dmitryduev Wording? Maybe "Clear" or "Reset"?

acrellin:
@stefanv those sound like they apply to the form field. Nullify OTOH means to set equal to null

acrellin:
@stefanv  Leaving the wording as is. Ready for review.

acrellin:
How about "Set null" or "Set to null"?

acrellin:
Updated to "Set null": 
nullify_redshift

static/js/components/UpdateSourceRedshift.jsx
```diff
@@ -77,7 +78,7 @@ const UpdateSourceRedshift = ({ source }) => {
         }}
         style={{ position: "fixed" }}
       >
-        <DialogTitle>Update Redshift</DialogTitle>
+        <DialogTitle>Update Redshift (or Nullify)</DialogTitle>
```
guynir42, line 81: Have we decided on "null" as opposed to "clear" the redshift? 
If not, then the title of the dialog should change as well. 
If "null" then it is ok.

guynir42 (approved):
If we are happy with the nomenclature it all looks good. 
I think "clear" or "nullify" are both good options.

stefanv:
I'd like to hear @dmitryduev's take on naming; I suspect users don't know what `null` means and it will confuse them.  Clear is very straightforward.

acrellin:
It's hard to imagine any astronomer not knowing what null means. @dmitryduev can you weigh in here?

stefanv:
:man_shrugging: I only know astronomers that program :)

acrellin:
> 🤷‍♂️ I only know astronomers that program :)

😂 Same here, actually

dmitryduev:
Folks, how about leaving it as is + adding a tooltip (?) on the right with a brief explanation of what it actually does?

acrellin:
@dmitryduev @stefanv done: 
nullify_redshift_tooltip

dmitryduev (approved):
👍🏻

stefanv:
Last thing I'd ask is to remove the "(or Nullify)" from the widget title.

Still leaning to "Clear" instead of "Set null", which feels unnecessarily technical to me, but I will merge either way once that is done.

Thanks, Ari, and sorry for dragging this out.

acrellin:
@stefanv   
image
😉

stefanv:
Now I feel a bit bad—didn't mean to strongarm that one.  Sorry :(

stefanv:
Redemption via merging.

acrellin:
@stefanv no worries, I'm very happy with your proposed wording 🙂

acrellin:
> Redemption via merging.

😂

PR #1756, opened 2021-03-10
Show save candidate button alongside manage groups button on scanning page when relevant

acrellin:
This PR updates the scanning page for candidates that are already saved to at least one group in the following way: If a candidate is already saved to at least one group but is not yet saved to **all** of the groups being scanned (those groups selected in the filter form), then the "Save to ..." button is now displayed in addition to the "Manage Groups" button, with those groups the source is already saved to being filtered out of the "Save to" groups (this applies both to the default button action and the "Edit groups and Save" groups list). 

In this screenshot I was scanning all three groups (i.e. "Program A", "Program B" and "Sitewide Group" were all selected in the candidate search form). Note that those groups that the source is already saved to are filtered out of the "Save to" button groups:

scanning_page_save_cand_buttons_2

### DIFF
--- a/static/js/components/CandidateList.jsx
+++ b/static/js/components/CandidateList.jsx
@@ -623,14 +623,41 @@ const CandidateList = () => {
               label="NOT SAVED"
               className={classes.itemPaddingBottom}
             />
-            <br />
-            <div className={classes.saveCandidateButton}>
-              <SaveCandidateButton
-                candidate={candidateObj}
-                userGroups={userAccessibleGroups}
-                filterGroups={filterGroups}
-              />
-            </div>
+          </div>
+        )}
+        {Boolean(
+          !candidateObj.is_source ||
+            (candidateObj.is_source &&
+              filterGroups.filter(
+                (g) =>
+                  !candidateObj.saved_groups.map((x) => x.id).includes(g.id)
+              ).length)
+        ) && (
+          // eslint-disable-next-line react/jsx-indent
+          <div className={classes.saveCandidateButton}>
+            <SaveCandidateButton
+              candidate={candidateObj}
+              userGroups={
+                candidateObj.is_source
+                  ? userAccessibleGroups.filter(
+                      (g) =>
+                        !candidateObj.saved_groups
+                          .map((x) => x.id)
+                          .includes(g.id)
+                    )
+                  : userAccessibleGroups
+              }
+              filterGroups={
+                candidateObj.is_source
+                  ? filterGroups.filter(
+                      (g) =>
+                        !candidateObj.saved_groups
+                          .map((x) => x.id)
+                          .includes(g.id)
+                    )
+                  : filterGroups
+              }
+            />
           </div>
         )}
         {candidateObj.last_detected_at && (
--- a/static/js/components/CandidateList.jsx
+++ b/static/js/components/CandidateList.jsx
@@ -625,6 +625,7 @@ const CandidateList = () => {
             />
           </div>
         )}
+        {/* If candidate is either unsaved or is not yet saved to all groups being filtered on, show the "Save to..." button */}
         {Boolean(
           !candidateObj.is_source ||
             (candidateObj.is_source &&
@@ -638,6 +639,7 @@ const CandidateList = () => {
             <SaveCandidateButton
               candidate={candidateObj}
               userGroups={
+                // Filter out groups the candidate is already saved to
                 candidateObj.is_source
                   ? userAccessibleGroups.filter(
                       (g) =>
@@ -648,6 +650,7 @@ const CandidateList = () => {
                   : userAccessibleGroups
               }
               filterGroups={
+                // Filter out groups the candidate is already saved to
                 candidateObj.is_source
                   ? filterGroups.filter(
                       (g) =>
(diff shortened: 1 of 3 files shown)


acrellin:
Locally:
```
py.test -v skyportal/tests/frontend/test_scanning_page.py
========================================== test session starts===========================================
platform linux -- Python 3.8.5, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /home/arien/anaconda3/envs/skyportal38/bin/python
cachedir: .pytest_cache
Using --randomly-seed=3703730122
rootdir: /home/arien/projects/skyportal
plugins: randomly-3.5.0, rerunfailures-9.1.1, progress-1.2.1, factoryboy-2.0.3
collected 10 items                                                                                       

skyportal/tests/frontend/test_scanning_page.py::test_candidate_redshift_filtering PASSED           [ 10%]
skyportal/tests/frontend/test_scanning_page.py::test_save_candidate_select_groups PASSED           [ 20%]
skyportal/tests/frontend/test_scanning_page.py::test_submit_annotations_filtering PASSED           [ 30%]
skyportal/tests/frontend/test_scanning_page.py::test_submit_annotations_sorting PASSED             [ 40%]
skyportal/tests/frontend/test_scanning_page.py::test_candidate_date_filtering PASSED               [ 50%]
skyportal/tests/frontend/test_scanning_page.py::test_save_candidate_quick_save PASSED              [ 60%]
skyportal/tests/frontend/test_scanning_page.py::test_candidate_classifications_filtering PASSED    [ 70%]
skyportal/tests/frontend/test_scanning_page.py::test_save_candidate_no_groups_error_message PASSED [ 80%]
skyportal/tests/frontend/test_scanning_page.py::test_candidate_group_filtering PASSED              [ 90%]
skyportal/tests/frontend/test_scanning_page.py::test_candidate_saved_status_filtering PASSED       [100%]

===================================== 10 passed in 106.40s (0:01:46)=====================================
```

static/js/components/CandidateList.jsx
```diff
@@ -623,14 +623,41 @@ const CandidateList = () => {
...
-            <div className={classes.saveCandidateButton}>
-              <SaveCandidateButton
-                candidate={candidateObj}
-                userGroups={userAccessibleGroups}
-                filterGroups={filterGroups}
-              />
-            </div>
+          </div>
+        )}
+        {Boolean(
```
stefanv, line 628: Is this Boolean in the right place?  Why is it here?  Shouldn't there be a check on the size of the filtered response?

This whole check is a bit opaque to me; perhaps worth factoring out and giving it a name?

static/js/components/CandidateList.jsx
```diff
@@ -623,14 +623,41 @@ const CandidateList = () => {
...
+          !candidateObj.is_source ||
+            (candidateObj.is_source &&
+              filterGroups.filter(
+                (g) =>
+                  !candidateObj.saved_groups.map((x) => x.id).includes(g.id)
+              ).length)
+        ) && (
+          // eslint-disable-next-line react/jsx-indent
```
stefanv, line 636: Weird, what does jsx expect?

static/js/components/CandidateList.jsx
```diff
@@ -623,14 +623,41 @@ const CandidateList = () => {
...
+          <div className={classes.saveCandidateButton}>
+            <SaveCandidateButton
+              candidate={candidateObj}
+              userGroups={
+                candidateObj.is_source
+                  ? userAccessibleGroups.filter(
+                      (g) =>
+                        !candidateObj.saved_groups
```
stefanv, line 644: Same check as before; refactor?

static/js/components/CandidateList.jsx
acrellin: Yes, this is in the right place. When I omitted `Boolean` and just left the expression, a `0` was sometimes rendered when the length was zero.

I can add in-line comments to clarify what is being checked here.

static/js/components/CandidateList.jsx
acrellin: prettier and eslint disagree about what should happen here, so I was unable to commit without this

static/js/components/CandidateList.jsx
acrellin: I'll add a clarifying comment

dmitryduev (changes requested):
Thanks for taking this on @acrellin. I fiddled a bit with the UI (without actually looking at the code, shame on me), and noticed something that does not look quite right:
candidates-1

This is immediately after a fresh/clean start. Could you please look into this?

acrellin:
@dima that's because when you unsave a source, it doesn't get deleted, it just gets marked inactive, then the save source button attempts to add a duplicate source. I want to update that handler to just pass for sources that already exist, but have a hunch that you will want it to still error. Am I right? If so I'll do some other finagling to sort it out

acrellin:
@dmitryduev done: 
Peek 2021-03-11 11-36

ready for re-review

static/js/components/CandidateList.jsx
acrellin: Done

static/js/components/CandidateList.jsx
acrellin: Done

acrellin:
Failures are unrelated (followup requests)

guynir42 (approved):
Looks good to me.

stefanv:
This doesn't have to be addressed here, but I think the "unsave" thing is a bit confusing: perhaps we should add radiobutton crosses to the list of groups, then when a user clicks on the cross it pops a confirmation dialog?

Merging this for now.

PR #1757, opened 2021-03-10
Reject candidates

guynir42:
Let the user click a single button to reject a candidate. 
The candidate will no longer show up on the scanning page. 
An option to show all candidates, including those rejected, is set to false by default. 
This uses the Listing table, in a very similar way to "favorites". 

reject_candidates


dmitryduev (commented):
Thanks for the quick implementation @guynir42! As we have discussed, I think this would be more widely applicable if implemented at at the Candidate (Obj x Filter) level (I'd add a "hidden" column to Candidate; this may have to be "broadcast" to all Candidates for a given (Obj, Filter) pair). Consequently, there should be a toggle switch (or better a super complicated dropdown menu to cover all possible cases ;-) ) to display/hide candidates marked as hidden in selected groups.

guynir42:
> Thanks for the quick implementation @guynir42! As we have discussed, I think this would be more widely applicable if implemented at at the Candidate (Obj x Filter) level (I'd add a "hidden" column to Candidate; this may have to be "broadcast" to all Candidates for a given (Obj, Filter) pair). Consequently, there should be a toggle switch (or better a super complicated dropdown menu to cover all possible cases ;-) ) to display/hide candidates marked as hidden in selected groups.

This is fairly straightforward to implement, I think. But there is an issue with having user A trust that user B to hide away candidates for them. Consider that B maybe shares only some of the groups/filters with A. 
It is always possible to show the hidden candidates, but at some point you have to trust the other users that are hiding candidates, otherwise you will spend just the same amount of time scanning the hidden ones. 

I think this is not so simple. I also think that the current implementation is useful on its own. 
Perhaps we should merge it and open for discussion more advanced ways to reject candidates in a different issue/PR?

kmshin1397 (commented):
@guynir42 Do you want to merge in the latest master here to let the CI run fully? Just want to make sure, but this looks good. Thanks!

guynir42:
@kmshin1397 @dmitryduev all tests are now green...

dmitryduev (approved):

dmitryduev (changes requested):
Oh, actually one small ui request: could you please align the icon with the previously saved/not saved tag and enforce it to be on the same line? (prob making it smaller too)

guynir42:
> Oh, actually one small ui request: could you please align the icon with the previously saved/not saved tag and enforce it to be on the same line? (prob making it smaller too)

When the object is saved to multiple groups there are a few chips of "previously saved to" so where should the reject button go in that case?  @dmitryduev

guynir42:
@dmitryduev ok I think I got it... but is there any way to force it to not break the line?

kmshin1397 (approved):

guynir42:
The test failure seems unrelated...

dmitryduev (approved):
👍🏻

PR #1758, opened 2021-03-10
Better phot plot

scizen9:
Use the bokeh.transform factor_mark to map markers onto data['instrument'] in photometry plot.

However, I cannot find an easy way to perform the same mapping for the CheckboxWithLegendGroup(CheckboxGroup) object.  I tried using the same methodology as the colors property in that object, but then the photometry plot does not appear even though the code throws no errors.

I suggest we try using interactive legends and the factor_mark and the factor_cmap bokeh transform functions without splitting the data on the label.  Here is an example from the bokeh docs:

[Mapping marker types](https://docs.bokeh.org/en/latest/docs/user_guide/data.html#mapping-marker-types)

It would change the character of the labels, but would simplify this function greatly.  I'm open to suggestions.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -13,6 +13,8 @@ from bokeh.models import (
     Button,
     LinearAxis,
     RadioGroup,
+    CategoricalColorMapper,
+    Legend,
 )
 from bokeh.models.widgets import CheckboxGroup, TextInput, Panel, Tabs, Div
 from bokeh.plotting import figure, ColumnDataSource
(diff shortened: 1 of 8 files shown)


skyportal/plot.py
```diff
@@ -261,6 +277,22 @@ def photometry_plot(obj_id, user, width=600, height=300, device="browser"):
...
 
+    # get marker for each unique instrument
+    instruments = list(data.instrument.unique())
+    markers = []
+    for i, inst in enumerate(instruments):
+        if i >= len(phot_markers):
+            pmi = i - len(phot_markers)
+        else:
+            pmi = i
+        markers.append(phot_markers[pmi])
```
kmshin1397, line 288:
```suggestion
        markers.append(phot_markers[i % len(phot_markers)])
```
If there are more than `2 * len(phot_markers)` than the current logic will fail. Using the mod will guard against that (although it's probably very unlikely that we ever get that many instruments on a source)

kmshin1397 (commented):
Thanks, this looks really good @scizen9! Definitely much simpler than the old approach. As mentioned during standup, I will probably want to touch up some of the layout stuff in here to make it more responsive but I think it's fine for this PR. 

Just one suggestion otherwise.

acrellin (approved):
LGTM, thanks @scizen9 !

kmshin1397 (approved):

PR #1759, opened 2021-03-10
Adjust photometry plot layouts

### LINKED ISSUE #1753: Photometry plot legend on mobile doesn't handle large number of entries well

kmshin1397:
- Both photometry and spectroscopy plots now dynamically adjust height on smaller screens
  - The spectroscopy plot was actually already doing this - I just removed the height parameter being passed in and going unused.
  - The photometry plot now computes a height proportional to the number of instruments/filters as the legend is moved below the plot on smaller screens and thus will alter the needed height.
- The period tab of the photometry plot also now adjusts plot layouts for smaller screens
Closes #1753 
plots

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -1,4 +1,5 @@
 import itertools
+import math
 
 import numpy as np
 import pandas as pd
@@ -218,7 +219,7 @@ def annotate_spec(plot, spectra, lower, upper):
         )
 
 
-def photometry_plot(obj_id, user, width=600, height=300, device="browser"):
+def photometry_plot(obj_id, user, width=600, device="browser"):
     """Create object photometry scatter plot.
 
     Parameters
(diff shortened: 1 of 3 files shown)


skyportal/plot.py
```diff
@@ -746,6 +749,11 @@ def photometry_plot(obj_id, user, width=600, height=300, device="browser"):
     plot.add_layout(LinearAxis(x_range_name="Days Ago", axis_label="Days Ago"), 'below')
 
     colors_labels = data[['color', 'label']].drop_duplicates()
+    height = (
+        500
+        if device == "browser"
+        else math.floor(width / aspect_ratio) + 18 * len(colors_labels) + 100
```
stefanv, line 755: Maybe add a comment to explain where the 18 comes from

stefanv (approved):

skyportal/plot.py
kmshin1397: Will do!

stefanv: But where does the 18 come from?  Is that 1rem (16px) + 2?

kmshin1397: That would be my guess, but can't be certain as it's controlled by bokeh I think. This will all be a mute point and changed  once #1758 goes in and we stop using the `CheckboxWithLegendGroup`.

kmshin1397:
Closing in favor of a new PR, as the code has changed significantly once adjusted to incorporate the plot restructuring in #1774

PR #1760, opened 2021-03-11
Allow listing git logs from multiple projects

stefanv:
To facilitate this, we now store git logs in JSON files, instead of as
text files.  Each JSON file contains some meta-data about the git log
(where to link PRs and commits to, what the project is called) as well
as the entries to the git log itself.

### DIFF
--- a/skyportal/handlers/api/sysinfo.py
+++ b/skyportal/handlers/api/sysinfo.py
@@ -1,27 +1,24 @@
-import os
-import subprocess
+import glob
+import json
+import itertools
 
 from baselayer.app.env import load_env
 from baselayer.app.access import auth_or_token
 from ..base import BaseHandler
 
 from skyportal.models import cosmo
-from skyportal.utils import gitlog
+from skyportal.utils.gitlog import get_gitlog, parse_gitlog
 
 
 _, cfg = load_env()
 
-# This file is generated with
-#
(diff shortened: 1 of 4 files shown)


acrellin (approved):

dmitryduev (approved):
👍🏻

dmitryduev (changes requested):
Failure on CI is relevant, could you please fix the test (`FAIL: skyportal/tests/utils/test_gitlog.py:13 in test_gitlog_parse`)?

skyportal/handlers/api/sysinfo.py
```diff
@@ -61,33 +58,24 @@ def get(self):
...
+        parsed_log = list(sorted(parsed_log, key=lambda x: x['time'], reverse=True))
+        parsed_log = parsed_log[:max_log_lines]
         parsed_log = [
             entry
             for entry in parsed_log
             if not (entry['description'].lower().startswith(('bump', 'pin')))
         ]
-        parsed_log = parsed_log[:max_log_lines]
+
+        print(parsed_log)
```
profjsb, line 78: Remove
```suggestion
```

profjsb (changes requested):
Looks good. Small change requested. Also, looks like a test needs fixing.

stefanv:
Thanks, should be good now.

Issue #1744, opened 2021-03-05
Suppress authentication error messages in logs


acrellin:

kmshin1397:
Baselayer changes are in - waiting for SP update

PR #1763, opened 2021-03-11
Suppress logging of expected authentication pipeline-related exceptions

acrellin:
Depends on https://github.com/cesium-ml/baselayer/pull/202 

Closes https://github.com/skyportal/skyportal/pull/1744

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -90,6 +90,9 @@ log = make_log('app_server')
 
 class CustomApplication(tornado.web.Application):
     def log_request(self, handler):
+        # We don't want to log expected exceptions intentionally raised
+        # during auth pipeline; such exceptions will have "google-oauth2" in
+        # their request route
         if "google-oauth2" in str(handler._request_summary()):
             return
         return super().log_request(handler)
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -1,4 +1,5 @@
 import tornado.web
+from tornado.log import access_log
 
 from baselayer.app.app_server import MainPageHandler
 from baselayer.app import model_util as baselayer_model_util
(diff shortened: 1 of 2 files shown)


skyportal/app_server.py
```diff
@@ -86,6 +87,35 @@
 
 log = make_log('app_server')
 
+
+class CustomApplication(tornado.web.Application):
+    def log_request(self, handler):
+        """Writes a completed HTTP request to the logs.
+        By default writes to the python root logger.  To change
```
stefanv, line 94: What is "the python root logger"?

skyportal/app_server.py
```diff
@@ -86,6 +87,35 @@
 
 log = make_log('app_server')
 
+
+class CustomApplication(tornado.web.Application):
+    def log_request(self, handler):
+        """Writes a completed HTTP request to the logs.
```
stefanv, line 93:
```suggestion
        """Writes a completed HTTP request to the logs.
        
```

skyportal/app_server.py
```diff
@@ -86,6 +87,35 @@
...
+        By default writes to the python root logger.  To change
+        this behavior either subclass Application and override this method,
+        or pass a function in the application settings dictionary as
+        ``log_function``.
+        """
+        if "google-oauth2" in str(handler._request_summary()):
```
stefanv, line 99: I'm not confident about this method; what code is this based on?  It feels a bit magical.  What does this do that we need, that is not currently being done?

skyportal/app_server.py
acrellin: this is just copy-pasted from the tornado source code, with minor changes

skyportal/app_server.py
acrellin: @stefanv We've discussed this. You said you don't want to see the "access error 500" messages in the logs when they're associated with the PSA pipeline -- this filters those out (the URL always contains "google-oauth2" in those cases). What's magical about it? I'm happy for you to propose an alternate implementation.

stefanv: Looks good now!  Please also add a comment to explain why we do this.

stefanv (approved):
Thanks, Ari!

acrellin:
@stefanv okay to merge this one?

acrellin:
@stefanv pinging you again here

skyportal/app_server.py
stefanv: The comment says "message text" and the comment above says "in the request route".  Request route makes the most sense to me; is `_request_summary()` (private method?) the best way to get to that?

acrellin: Comment updated. Yes, why is `_request_summary` not a good way to get that?

stefanv:
It's a private method; was wondering if that's the best way to get the route?

acrellin:
@stefanv okay, now uses a non-private attribute

stefanv:
Beautiful!

PR #1778, opened 2021-03-12
Re-introduce new photometry plot legend and make it responsive

### LINKED ISSUE #1753: Photometry plot legend on mobile doesn't handle large number of entries well

kmshin1397:
This PR re-introduces the changes in the reverted #1758 (refactoring the legend to use the legend component instead of checkboxes to support better colors and markers), and makes adjustments to the photometry plot layouts to make it properly responsive. This includes changes previously included in the closed #1759 PR, as well as new changes necessitated by the changes in #1758.

- Both photometry and spectroscopy plots now dynamically adjust height on smaller screens
  - The spectroscopy plot was actually already doing this - I just removed the height parameter being passed in and going unused.
  - The photometry plot now computes a height proportional to the number of instruments/filters as the legend is moved below the plot on smaller screens and thus will alter the needed height.
- The period tab of the photometry plot also now adjusts plot layouts for smaller screens

Closes #1753 

## Browser
phot_desktop

## Tablet (iPad)
phot_ipad

## Mobile (iPhone 8)
phot_iphone

### DIFF
--- a/skyportal/handlers/api/internal/plot.py
+++ b/skyportal/handlers/api/internal/plot.py
@@ -20,7 +20,6 @@ device_types = [
 class PlotPhotometryHandler(BaseHandler):
     @auth_or_token
     def get(self, obj_id):
-        height = self.get_query_argument("height", 300)
         width = self.get_query_argument("width", 600)
         device = self.get_query_argument("device", None)
         # Just return browser by default if not one of accepted types
@@ -29,7 +28,6 @@ class PlotPhotometryHandler(BaseHandler):
         json = plot.photometry_plot(
             obj_id,
             self.current_user,
-            height=int(height),
             width=int(width),
             device=device,
         )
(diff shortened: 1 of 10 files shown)


stefanv:
Are we leaving alone BokehModels.js as a proof-of-concept?  Admittedly, it was a massive headache to sort it out, but if we don't foresee using it again we can remove it.

scizen9 (approved):
Much better!  It even works on my tiny-screen iPhone SE.

scizen9:
Wondering if we shouldn't use the same interactive legend for the spectroscopy?

kmshin1397:
@stefanv it should still be in there for now since the spectroscopy plot still makes use of it I think. But as @scizen9 says, it's probably a good idea to transition the spectroscopy plot to the new component as well and then we can revisit removing BokehModels.js. I think that's a separate PR though as I imagine that will also require similar levels of refactoring as this.

stefanv:
Thanks!

scizen9:
Once this PR is merged, I'll create a branch to re-do the spectroscopy plot the same way.

PR #1787, opened 2021-03-15
Disable LT tests

kmshin1397:
LT folks complained about our tests hitting their servers too much - this PR temporarily disables the LT tests. This is probably a good time to revisit trying to do some sort of mocking with these. I'll work on it.


mcoughlin (approved):
Huge thank you for this Kyung

Issue #1786, opened 2021-03-15
Use same interactive legend for spec plots as now used in phot plots


scizen9:
Considering this consistency to not be foolish (and therefore, not a hobgoblin of my little mind), I propose we make the spec plots use the same interactive legend as the phot plots now use.  I will create a PR for this shortly.

scizen9:
fixed with PR #1788

PR #1788, opened 2021-03-15
use interactive legend for spec plots

scizen9:
Useed @kmshin1397 's add_plot_legend to add interactive plot to spec plot.  Now phot and spec plots are consistent, thus satisfying the hobgoblin of my little mind and addressing [Issue 1786](https://github.com/skyportal/skyportal/issues/1786).

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -1253,13 +1253,19 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
     )
     plot.add_tools(hover)
     model_dict = {}
+    legend_items = []
     for i, (key, df) in enumerate(split):
+        renderers = []
+        s = Spectrum.query.get(key)
+        label = f'{s.instrument.name} ({s.observed_at.date().strftime("%m/%d/%y")})'
         model_dict['s' + str(i)] = plot.step(
             x='wavelength',
             y='flux',
             color=color_map[key],
             source=ColumnDataSource(df),
         )
+        renderers.append(model_dict['s' + str(i)])
+        legend_items.append(LegendItem(label=label, renderers=renderers))
         model_dict['l' + str(i)] = plot.line(
             x='wavelength',
             y='flux',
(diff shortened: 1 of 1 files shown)


kmshin1397 (commented):
Looks good @scizen9! I opened a PR to this PR branch with a some small changes to the plot aspect ratios [here](https://github.com/scizen9/skyportal/pull/1) and once that goes in, I think this is good to go.

scizen9:
Merged your PR.  Thanks for helping with this!

kmshin1397 (approved):

skyportal/plot.py
```diff
@@ -1241,8 +1241,14 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
     )
     plot_width = None if device == "browser" else width
+    if device == "mobile_portrait":
+        aspect_ratio = 1
+    elif device == "tablet_portrait":
+        aspect_ratio = 1.5
+    else:
+        aspect_ratio = 1.8
     plot = figure(
-        aspect_ratio=2.0 if device == "mobile_landscape" else 1.5,
```
joelachance, line 1245: It looks like `mobile_landscape` has become either `mobile_portrait` or `tablet_portrait`. Do we know if it's used anywhere else, and if not, maybe we should remove it?

skyportal/plot.py
```diff
@@ -1461,12 +1455,7 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
 
-    row1 = (
-        column(plot, toggle)
-        if "mobile" in device or "tablet" in device
-        else row(plot, toggle)
-    )
     row2 = row(elements_groups)
     row3 = column(z, v_exp) if "mobile" in device else row(z, v_exp)
-    layout = column(row1, row2, row3, sizing_mode='scale_height', width=width)
+    layout = column(plot, row2, row3, sizing_mode='scale_height', width=width)
```
joelachance, line 1460: Plot is definitely a better name than row1 :)

skyportal/plot.py
kmshin1397: I believe it's still used in the photometry plot so I think we might as well keep it around here. I'd like to keep the option available in case future additions to these plots require more fine-tuned layout changes that require specifying something for `mobile_landscape` (same reason why I implemented both `tablet_portrait` and `tablet_landscape` originally even though we didn't actually have differences in the layouts for those two cases until this recent change in legend components)

scizen9:
@joelachance, if possible, can you approve and merge?  I'd like to have this in the next deployment if possible.

joelachance:
@scizen9, unfortunately, I don't have permissions for it. Maybe @acrellin can?

stefanv:
Thanks, Don!

Issue #1368, opened 2020-11-17
Implement dedicated stream groups


dmitryduev:
For bookkeeping:

- Add users to these groups based on their stream access
- Always post stream photometry to these groups

acrellin:
@dmitryduev can you expand on how these will be used? Thanks

acrellin:
@dmitryduev pinging you again here -- can you please expand?

acrellin:
@dmitryduev another ping -- please expand

PR #1793, opened 2021-03-17
Implement stream / group / user access logic and update group and stream handlers

dannygoldstein:
This PR makes use of the framework developed in https://github.com/cesium-ml/baselayer/pull/207/ to enforce the following rules for Group / User / Stream membership:

1. A user can only join a group if they have access to all of the group's streams. 
2. A stream can only be removed from a group if none of the Group's filters currently filter the stream.
3. A stream can only be added to a group if all of the group's members have access to the stream.
4. A group user's admin status can only be modified by group admins.
5. Only a group admin or the target user can remove a user from a group.
6. Only system admins can grant and revoke stream access to users. 
7. Only a group admin can add a user to a group.
8. Users cannot add streams to single user groups.
9. No one can remove a user from their single user group (this can only be done by an event listener).
10. No one can add a user to a single user group through the groups api. 

These rules are enforced within the context of the new permissions framework that is currently being refactored throughout the codebase. The relevant conditions above are automatically checked at the end of the transaction in which  the corresponding CRUD op is performed via the ORM.

The Stream and Group handlers are refactored using these new rules. 

This PR also fixes a bug in the demo data where users without stream=2 access were being succcessfully added to a group that required it. 

Closes https://github.com/skyportal/skyportal/issues/1368

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -49,6 +49,12 @@ groups/=program_A/users:
   - userID: =fulluser
     admin: false
 
+groups/=program_B/users:
+  - userID: =viewonlyuser
+    admin: false
+  - userID: =fulluser
+    admin: false
+
 streams/=ztf_public/users:
   - user_id: =viewonlyuser
   - user_id: =fulluser
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -49,10 +49,6 @@ groups/=program_A/users:
   - userID: =fulluser
     admin: false
 
-groups/=program_B/users:
-  - userID: =viewonlyuser
-    admin: false
-
 streams/=ztf_public/users:
   - user_id: =viewonlyuser
   - user_id: =fulluser
--- a/skyportal/handlers/api/stream.py
+++ b/skyportal/handlers/api/stream.py
@@ -48,11 +48,11 @@ class StreamHandler(BaseHandler):
                   schema: Error
         """
         if stream_id is not None:
-            s = DBSession().query(Stream).filter(Stream.id == stream_id).first()
+            s = Stream.get_if_accessible_by(stream_id, self.current_user)
             if s is None:
                 return self.error("Invalid stream ID.")
             return self.success(data=s)
-        streams = DBSession().query(Stream).all()
+        streams = Stream.get_records_accessible_by(self.current_user)
         self.verify_and_commit()
         return self.success(data=streams)
 
@@ -168,7 +168,10 @@ class StreamHandler(BaseHandler):
               application/json:
                 schema: Success
         """
-        DBSession().delete(Stream.query.get(stream_id))
+        stream = Stream.get_if_accessible_by(
+            stream_id, self.current_user, mode="delete"
+        )
+        DBSession().delete(stream)
         self.verify_and_commit()
 
         return self.success()
@@ -226,7 +229,8 @@ class StreamUserHandler(BaseHandler):
 
         stream_id = int(stream_id)
         su = (
-            StreamUser.query.filter(StreamUser.stream_id == stream_id)
+            StreamUser.query_records_accessible_by(self.current_user)
+            .filter(StreamUser.stream_id == stream_id)
             .filter(StreamUser.user_id == user_id)
             .first()
         )
@@ -264,7 +268,8 @@ class StreamUserHandler(BaseHandler):
                 schema: Success
         """
         su = (
-            StreamUser.query.filter(StreamUser.stream_id == stream_id)
+            StreamUser.query_records_accessible_by(self.current_user, mode="delete")
+            .filter(StreamUser.stream_id == stream_id)
             .filter(StreamUser.user_id == user_id)
             .first()
         )
--- a/skyportal/handlers/api/stream.py
+++ b/skyportal/handlers/api/stream.py
@@ -175,7 +175,7 @@ class StreamHandler(BaseHandler):
 
 
 class StreamUserHandler(BaseHandler):
-    @permissions(["Manage users"])
+    @auth_or_token
     def post(self, stream_id, *ignored_args):
         """
         ---
@@ -238,7 +238,7 @@ class StreamUserHandler(BaseHandler):
 
         return self.success(data={'stream_id': stream_id, 'user_id': user_id})
 
-    @permissions(["Manage users"])
+    @auth_or_token
     def delete(self, stream_id, user_id):
         """
         ---
--- a/skyportal/handlers/api/stream.py
+++ b/skyportal/handlers/api/stream.py
@@ -48,9 +48,9 @@ class StreamHandler(BaseHandler):
                   schema: Error
         """
         if stream_id is not None:
-            s = Stream.get_if_accessible_by(stream_id, self.current_user)
-            if s is None:
-                return self.error("Invalid stream ID.")
+            s = Stream.get_if_accessible_by(
+                stream_id, self.current_user, raise_if_none=True
+            )
             return self.success(data=s)
         streams = Stream.get_records_accessible_by(self.current_user)
         self.verify_and_commit()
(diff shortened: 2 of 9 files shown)


dannygoldstein:
No migration needed.

data/db_demo.yaml
```diff
@@ -57,6 +57,12 @@ streams/=ztf_public/users:
   - user_id: =groupadmin
   - user_id: =testadmin
 
+streams/=ztf_partnership/users:
+  - user_id: =viewonlyuser
+  - user_id: =fulluser
+  - user_id: =groupadmin
+  - user_id: =testadmin
+
```
dannygoldstein, line 65: Needed to grant users this stream so that they could be added to the group that required the ZTF partnership stream.

skyportal/tests/api/test_groups.py
```diff
@@ -245,37 +244,10 @@ def test_add_stream_to_single_user_group_delete_stream(
...
-    )
-    assert len(data["data"]["streams"]) == 0
 
-    # check user still exists
-    status, data = api("GET", f"user/{new_user_id}", token=super_admin_token)
-    assert status == 200
-    assert data["data"]["id"] == new_user_id
+    # check that you can't add a stream to a single user group
+    assert status == 400
+    assert data['status'] == 'error'
```
dannygoldstein, line 250: Refactored the above as users should not be allowed to add stream requirements to single user groups

skyportal/models.py
```diff
@@ -3633,4 +3635,137 @@ def receive_after_flush(session, context):
...
+            GroupStream.stream_id == cls.stream_id,
+        ),
+    )
+    .group_by(cls.id)
+    # no OR here because Users will always be a member of at least one
+    # group -- their single user group.
+    .having(sa.func.bool_and(GroupStream.stream_id.is_(None)))
+)
+
+
```
dannygoldstein, line 3770: This is the meat of this PR.

kmshin1397 (commented):
Thanks, @dannygoldstein. I think the changes here all look good, and it'll definitely be good to get this in and close the loophole. I did notice a lack of tests around the error cases for the permission logic solidified here, unfortunately. By my count, I see two tests in `tests/api/test_groups.py` checking for insufficient permissions and neither of them really involve group members' stream access concerns (I see the can't delete the public group and can't add streams to single user groups). I think ideally each of the 10 bullet points you have in the PR description would have tests representing those insufficient permission cases, but at least a couple so we know it generally works? I mean, the queries for these `CustomUserAccessControl`s are pretty similar so even testing a few of these would give me a lot more confidence about the whole thing. They look perfectly fine to me on paper but doesn't hurt to have tests.

kmshin1397:
Also, do you know what [this test](https://github.com/skyportal/skyportal/blob/772f751fd376bbef743a7f2f1cb2cd3d2261bb47/skyportal/tests/api/test_groups.py#L181-L190) is supposed to do? It's the `test_non_su_add_stream_to_group` test. I'm guessing the `su` stands for `super user` but the `manage_groups_token` fixture used in the test actually does use the `super_admin_user` so it's not actually a non-super user calling the API there, but it still does return 400. Not sure why that's the case either, as I don't the fixture group should have any users with mismatching stream permissions either. 

Maybe there is something real that is off here that I'm not seeing - just noticed that the CI failures seem to be on potentially related tests as well.

skyportal/models.py
```diff
@@ -3633,4 +3635,137 @@ def receive_after_flush(session, context):
...
+# group admins can set the admin status of other group members
+GroupUser.update = accessible_by_group_admins
+
+
+GroupUser.delete = (
+    # users can remove themselves from a group
+    # admins can remove users from a group
+    # no one can remove a user from their single user group
+    (accessible_by_group_admins | AccessibleIfUserMatches('user'))
+    & GroupUser.read
```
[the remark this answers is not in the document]
dannygoldstein: Thanks @kmshin1397 , you are totally right. I will add some more tests. 

The line above is just for future proofing as you intuited.

acrellin:
@dannygoldstein What's the status of this PR?

acrellin:
@dannygoldstein If you could give a status update here, I'm happy to pick it up and carry it over the finish line

dannygoldstein:
Just needs tests for the cases outlined in the OP and I think one or two failing tests need to be fixed

acrellin:
@kmshin1397 Any idea what's going on with the model logic tests here? Seems to be fixture teardown-related

kmshin1397:
> @kmshin1397 Any idea what's going on with the model logic tests here? Seems to be fixture teardown-related

I have a feeling that the teardown errors may be fixed once the one failing model test is fixed. I tried running some of the other tests that error on teardown and they seem fine independently. I have an idea for how to fix that test - will work on it after dinner.

acrellin:
> > @kmshin1397 Any idea what's going on with the model logic tests here? Seems to be fixture teardown-related
> 
> I have a feeling that the teardown errors may be fixed once the one failing model test is fixed. I tried running some of the other tests that error on teardown and they seem fine independently. I have an idea for how to fix that test - will work on it after dinner.

Thanks, no rush

acrellin:
okay @kmshin1397 the API/front-end tests all pass now, and I could use a hand with the tear-down errors in the model tests if you have a minute to take a look (`psycopg2.errors.InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block`)

kmshin1397:
> okay @kmshin1397 the API/front-end tests all pass now, and I could use a hand with the tear-down errors in the model tests if you have a minute to take a look (`psycopg2.errors.InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block`)

@acrellin I would suggest you do something similar to what you did for `GroupUsers` for the `Group` delete policy as well. The one model test that fails (instead of errors) is due to some aliasing issues with `Group` and `accessible_by_admins`. Let's see if fixing the failing test will fix the fixtures.

kmshin1397:
@acrellin nice looks like the fixture errors went away - just some regular model test failures now

acrellin:
Yep, was able to get that fixed. Just finishing updating the last custom access logic methods that still have some issues.

acrellin:
@kmshin1397 Nice and green now -- ready for review.

acrellin (approved):

kmshin1397 (commented):
This all looks good @acrellin. One question: I noticed that we don't use any of the `query_records_accessible_by` or `get_if_accessible_by` in the Stream handlers. I guess the errors will be caught by the `verify_and_commit` regardless, but maybe better to add those in for consistency?

acrellin:
Yes, agreed, thanks for catching that. I had thought that had already been
done.

acrellin:
@kmshin1397 Done

skyportal/handlers/api/stream.py
```diff
@@ -48,11 +48,11 @@ def get(self, stream_id=None):
                   schema: Error
         """
         if stream_id is not None:
-            s = DBSession().query(Stream).filter(Stream.id == stream_id).first()
+            s = Stream.get_if_accessible_by(stream_id, self.current_user)
             if s is None:
                 return self.error("Invalid stream ID.")
```
kmshin1397, line 53: Use `raise_if_none` instead of the manual check?

skyportal/handlers/api/stream.py
```diff
@@ -168,14 +168,17 @@ def delete(self, stream_id):
               application/json:
                 schema: Success
         """
-        DBSession().delete(Stream.query.get(stream_id))
+        stream = Stream.get_if_accessible_by(
```
kmshin1397, line 171: `raise_if_none` here as well

acrellin: Done

kmshin1397 (approved):

kmshin1397:
Ah actually @acrellin `test_group_admin_cannot_grant_delete_user_stream_access` seems like a relevant failure

acrellin:
@kmshin1397 Whack-a-mole ;)  Okay, good to go now

kmshin1397 (approved):

PR #1799, opened 2021-03-18
Add padding for non-period photometry plots

kmshin1397:
image

Unfortunately, the Bokeh plot is plotted with some `position: absolute` styling, so it wasn't straightforward to just get the div with the buttons to get pushed down reactively. 

The overlapping button issue is only with photometry plots without the period tab. I'm not sure exactly which plot widgets the extra needed height is coming from but adding some extra height to the plot layout appears to fix the issue. I was adding similar padding for the plots with periods but my reasoning for those were the extra text fields in the period tab so not exactly sure why the similar padding is needed here but I'll investigate that more later.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -1142,7 +1142,7 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
         tabs = Tabs(tabs=[p2, p1, p3], width=width, height=height, sizing_mode='fixed')
     else:
         # tabs for mag, flux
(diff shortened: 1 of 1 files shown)


acrellin (approved):

dannygoldstein:
Locally, when running the test suite, I am still seeing

Screen Shot 2021-03-19 at 10 20 19 PM

acrellin:
@dannygoldstein are you sure you have these changes pulled in?

PR #1801, opened 2021-03-20
Patch security vulnerabilities and data model violations revealed by test suite after enabling UACL

dannygoldstein:
Merging https://github.com/cesium-ml/baselayer/pull/207 turned on automated permission checking for create, read, and update operations after they were unintentionally disabled since https://github.com/cesium-ml/baselayer/pull/184. Enabling these automated checks revealed a series of security vulnerabilities in skyportal that allowed API access to protected user data (photometry, spectra, etc.) and protected operations on user data (delete, update, etc.). 

The fix also revealed instances in which assumptions in the test suite violated underlying assumptions about the data model. 

This PR patches many of these up. The specific issues fixed in this PR include: 

* The candidate API GET endpoint had a major spectroscopy and photometry leak, returning all spectra and photometry regardless of permissions 
* The facility listener endpoint brought protected Allocation data into the session
* The facility listener handler did not correctly enforce update permissions on followup requests
* The taxonomy handler exposed a vulnerability that allowed any user with the "Delete taxonomy" ACL to delete any taxonomy (even if they couldn't read it) and all of the classifications associated with that taxonomy. There is now logic in place that allows deletion of taxonomies by users with the Delete taxonomy ACL only if a taxonomy is shared with one of their groups and there are no associated classifications (super admins can delete any taxonomy with no classifications)
* Access control policies on telescopes and instruments allowed arbitrary record creation, rather than only by system admins
* Users could not delete themselves, and their update UACL policy did not allow them to update themselves (both of these are now fixed)
* Many users in the test suite did not have any stream associations, but could post candidates to filters that filtered streams (this will be fixed in a subsequent PR). 

Performance impact:

Master: API tests took  235.76s (0:03:55) on my local machine 
This PR: API tests took 240.27s (0:04:00) on my local machine

The difference is 2%.

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -302,6 +302,7 @@ class CandidateHandler(BaseHandler):
                 application/json:
                   schema: Error
         """
+
         user_accessible_group_ids = [g.id for g in self.current_user.accessible_groups]
         include_photometry = self.get_query_argument("includePhotometry", False)
         include_spectra = self.get_query_argument("includeSpectra", False)
@@ -580,6 +581,8 @@ class CandidateHandler(BaseHandler):
                     .subquery()
                 )
             else:
+                # import pdb
+                # pdb.set_trace()
                 expire_date = arrow.get(annotation_exclude_date).datetime
                 right = (
                     DBSession()
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -309,18 +309,6 @@ class CandidateHandler(BaseHandler):
 
         if obj_id is not None:
             query_options = [joinedload(Candidate.obj).joinedload(Obj.thumbnails)]
-            if include_photometry:
-                query_options.append(
-                    joinedload(Candidate.obj)
-                    .joinedload(Obj.photometry)
-                    .joinedload(Photometry.instrument)
-                )
-            if include_spectra:
-                query_options.append(
-                    joinedload(Candidate.obj)
-                    .joinedload(Obj.spectra)
-                    .joinedload(Spectrum.instrument)
-                )
             c = Candidate.get_obj_if_readable_by(
                 obj_id,
                 self.current_user,
@@ -362,6 +350,21 @@ class CandidateHandler(BaseHandler):
                 key=lambda x: x["created_at"],
                 reverse=True,
             )
+
+            if include_photometry:
+                candidate_info['photometry'] = Photometry.get_records_accessible_by(
+                    self.current_user,
+                    mode='read',
+                    options=[joinedload(Photometry.instrument)],
+                )
+
+            if include_spectra:
+                candidate_info['spectra'] = Spectrum.get_records_accessible_by(
+                    self.current_user,
+                    mode='read',
+                    options=[joinedload(Spectrum.instrument)],
+                )
+
             candidate_info["annotations"] = sorted(
                 c.get_annotations_readable_by(self.current_user),
                 key=lambda x: x.origin,
@@ -696,8 +699,6 @@ class CandidateHandler(BaseHandler):
                 n_per_page,
                 "candidates",
                 order_by=order_by,
-                include_photometry=include_photometry,
-                include_spectra=include_spectra,
             )
         except ValueError as e:
             if "Page number out of range" in str(e):
@@ -744,6 +745,23 @@ class CandidateHandler(BaseHandler):
                     )
                 ]
                 candidate_list.append(obj.to_dict())
+
+                if include_photometry:
+                    candidate_list[-1][
+                        "photometry"
+                    ] = Photometry.get_records_accessible_by(
+                        self.current_user,
+                        mode='read',
+                        options=[joinedload(Photometry.instrument)],
+                    )
+
+                if include_spectra:
+                    candidate_list[-1]["spectra"] = Spectrum.get_records_accessible_by(
+                        self.current_user,
+                        mode='read',
+                        options=[joinedload(Spectrum.instrument)],
+                    )
+
                 candidate_list[-1]["comments"] = sorted(
                     [
                         cmt.to_dict()
@@ -940,8 +958,6 @@ def grab_query_results(
     n_items_per_page,
     items_name,
     order_by=None,
-    include_photometry=False,
-    include_spectra=False,
 ):
     # The query will return multiple rows per candidate object if it has multiple
     # annotations associated with it, with rows appearing at the end of the query
@@ -1022,12 +1038,6 @@ def grab_query_results(
 
     items = []
     query_options = [joinedload(Obj.thumbnails)]
-    if include_photometry:
-        query_options.append(
-            joinedload(Obj.photometry).joinedload(Photometry.instrument)
-        )
-    if include_spectra:
-        query_options.append(joinedload(Obj.spectra).joinedload(Spectrum.instrument))
     for item_id in page_ids:
         items.append(Obj.query.options(query_options).get(item_id))
     info[items_name] = items
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -352,17 +352,25 @@ class CandidateHandler(BaseHandler):
             )
 
             if include_photometry:
-                candidate_info['photometry'] = Photometry.get_records_accessible_by(
-                    self.current_user,
-                    mode='read',
-                    options=[joinedload(Photometry.instrument)],
+                candidate_info['photometry'] = (
+                    Photometry.query_records_accessible_by(
+                        self.current_user,
+                        mode='read',
+                        options=[joinedload(Photometry.instrument)],
+                    )
+                    .filter(Photometry.obj_id == obj_id)
+                    .all()
                 )
 
             if include_spectra:
-                candidate_info['spectra'] = Spectrum.get_records_accessible_by(
-                    self.current_user,
-                    mode='read',
-                    options=[joinedload(Spectrum.instrument)],
+                candidate_info['spectra'] = (
+                    Spectrum.query_records_accessible_by(
+                        self.current_user,
+                        mode='read',
+                        options=[joinedload(Spectrum.instrument)],
+                    )
+                    .filter(Spectrum.obj_id == obj_id)
+                    .all()
                 )
 
             candidate_info["annotations"] = sorted(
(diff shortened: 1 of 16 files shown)


skyportal/tests/api/test_sharing.py
```diff
@@ -127,69 +127,6 @@ def test_sharing_photometry_with_foreign_group(
     assert data["data"]["obj_id"] == public_source.id
 
 
-def test_sharing_photometry_shares_source(
```
dannygoldstein, line 130: The behavior in this test is based on outdated logic for Object ownership and sharing and should no longer be included in the suite (automated checking of permissions caught this).

skyportal/models.py
```diff
@@ -2786,16 +2831,49 @@ def from_ascii(
...
     update = delete = (
-        AccessibleIfUserMatches('allocation.group.users')
-        | AccessibleIfUserMatches('requester')
-    ) & read
+        (
+            AccessibleIfUserMatches('allocation.group.users')
+            | AccessibleIfUserMatches('requester')
+        )
+        & read
+    ) | CustomUserAccessControl(updatable_by_token_with_listener_acl)
```
dannygoldstein, line 2876: This allows remote facilities to update, but not read, any followup request associated with an instrument that the remote facility implements an API for. Requesters or others in the allocation group can also update the request.

skyportal/tests/api/test_telescope.py
```diff
@@ -3,7 +3,7 @@
 from skyportal.tests import api
 
 
-def test_token_user_post_get_telescope(upload_data_token):
+def test_token_user_post_get_telescope(super_admin_token):
```
dannygoldstein, line 6: Only super admins can post new telescopes (similar to instruments).

skyportal/tests/conftest.py
```diff
@@ -624,9 +624,11 @@ def public_groupuser(public_group, user):
 
 
 @pytest.fixture()
-def user2(public_group):
+def user2(public_group, public_stream):
```
dannygoldstein, line 627: Users all should have stream access to the streams that their group's filters filter. In this case, `public_group` had a `Filter` that filtered Stream 1, but `user2` did not have that stream. Nevertheless, they were permitted to be in the group. Tests that instantiated the Filter and the user in the same session raised an accesserror because the access policy for filters requires that users accessing them have access to their streams. This fixes that.

skyportal/handlers/api/candidate.py
```diff
@@ -302,24 +302,13 @@ def get(self, obj_id=None):
...
         user_accessible_group_ids = [g.id for g in self.current_user.accessible_groups]
         include_photometry = self.get_query_argument("includePhotometry", False)
         include_spectra = self.get_query_argument("includeSpectra", False)
 
         if obj_id is not None:
             query_options = [joinedload(Candidate.obj).joinedload(Obj.thumbnails)]
-            if include_photometry:
-                query_options.append(
-                    joinedload(Candidate.obj)
-                    .joinedload(Obj.photometry)
```
dannygoldstein, line 314: This line leaked all the candidate's photometry to any API caller.

skyportal/handlers/api/candidate.py
```diff
@@ -302,24 +302,13 @@ def get(self, obj_id=None):
...
-                    .joinedload(Photometry.instrument)
-                )
-            if include_spectra:
-                query_options.append(
-                    joinedload(Candidate.obj)
-                    .joinedload(Obj.spectra)
```
dannygoldstein, line 320: This line leaked all the candidate's photometry to any API caller.

dannygoldstein:
Failures unrelated I think (@kmshin1397, have you seen test_comment_groups_validation fail recently before? I know test_upload_spectroscopy has been flaky for a while. In https://github.com/skyportal/skyportal/runs/2157828399, the CI run for 1685f61, two other comment related tests failed, and then in the CI for the subsequent commit 53d6551, they no longer fail, which makes me think they are flaky).

kmshin1397 (approved):
Looks good, thanks @dannygoldstein! I checked out your branch and the comment tests seem to pass just fine for me locally so I guess it is just flaky not sure why though.

skyportal/handlers/api/photometry.py
```diff
@@ -947,19 +947,26 @@ def delete(self, upload_id):
...
+
+        n = len(photometry_to_delete)
+        if n == 0:
+            return self.error('Invalid bulk upload id.')
 
-        return self.success(f"Deleted {n_deleted} photometry points.")
+        for phot in photometry_to_delete:
+            DBSession().delete(phot)
+
+        # this will raise if the user does not have access to delete any of
```
stefanv, line 966: We should catch raises and use self.error

stefanv (commented):
Thanks very much for the security review, Danny.

skyportal/handlers/api/photometry.py
dannygoldstein: Thanks, yeah, that's what's done here - this raises an AccessError which triggers `self.error` in the exception handling hook and doesn't actually raise up the stack. I'll update the comment here to make that clearer

skyportal/tests/models/test_permissions.py
```diff
@@ -1,383 +1,446 @@
...
+    user,
+    public_group,
 ):
     accessible = public_group.is_accessible_by(user, mode="create")
     assert accessible
 
 
 def test_user_create_public_groupuser(
-    user, public_groupuser,
+    user,
```
acrellin, line 10: is this black's formatting? I would expect such a short parameter list to all be on the above line

acrellin (changes requested):

skyportal/handlers/api/candidate.py
```diff
@@ -361,6 +350,21 @@ def get(self, obj_id=None):
                 key=lambda x: x["created_at"],
                 reverse=True,
             )
+
+            if include_photometry:
+                candidate_info['photometry'] = Photometry.get_records_accessible_by(
```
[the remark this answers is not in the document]
dannygoldstein: Nice catch, yeah, my mistake.

skyportal/handlers/api/candidate.py
```diff
@@ -361,6 +350,21 @@ def get(self, obj_id=None):
...
+                    self.current_user,
+                    mode='read',
+                    options=[joinedload(Photometry.instrument)],
+                )
+
+            if include_spectra:
+                candidate_info['spectra'] = Spectrum.get_records_accessible_by(
```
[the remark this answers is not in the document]
dannygoldstein: Nice catch, yeah, my mistake.

skyportal/handlers/api/candidate.py
```diff
@@ -741,6 +743,23 @@ def get(self, obj_id=None):
                     )
                 ]
                 candidate_list.append(obj.to_dict())
+
+                if include_photometry:
+                    candidate_list[-1][
+                        "photometry"
+                    ] = Photometry.get_records_accessible_by(
```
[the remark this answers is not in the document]
dannygoldstein: Nice catch, yeah, my mistake.

skyportal/models.py
```diff
@@ -1951,6 +1956,30 @@ class Taxonomy(Base):
     )
 
 
+def taxonomy_update_delete_logic(cls, user_or_token):
+
+    if len({'Delete taxonomy', 'System admin'} & set(user_or_token.permissions)) == 0:
+        # nothing accessible
+        return restricted.query_accessible_rows(cls, user_or_token)
```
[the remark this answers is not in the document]
dannygoldstein: This function generates the query for taxonomies  that the current user can delete. L1963 says, if the querying user doesn't have System admin or Delete taxonomy acl, then no taxonomies are delete accessible. Otherwise, it says that the only taxonomies that can be deleted are those that have no associated classifications, preventing situations where classifications get deleted in a cascade when their parent taxonomy is deleted.

skyportal/tests/models/test_permissions.py
dannygoldstein: Yeah, black makes no changes when I run 

```
(skyportal2) Dannys-MBP:skyportal danny$ black skyportal/tests/models/test_permissions.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

skyportal/handlers/api/candidate.py
```diff
@@ -741,6 +743,23 @@ def get(self, obj_id=None):
...
+                        self.current_user,
+                        mode='read',
+                        options=[joinedload(Photometry.instrument)],
+                    )
+
+                if include_spectra:
+                    candidate_list[-1]["spectra"] = Spectrum.get_records_accessible_by(
```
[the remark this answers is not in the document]
dannygoldstein: Good call, thanks. Done in 81ef7c6

skyportal/handlers/api/photometry.py
```diff
@@ -660,7 +660,7 @@ def post(self):
         except ValidationError as e:
             return self.error(e.args[0])
 
-        self.finalize_transaction()
+        DBSession().commit()
```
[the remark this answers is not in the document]
dannygoldstein: Oops, this was just a relic of some local debugging. Thanks

skyportal/handlers/api/photometry.py
```diff
@@ -774,7 +774,7 @@ def put(self):
                 id_map[df_index] = id
 
         # release the lock
-        self.finalize_transaction()
+        DBSession().commit()
```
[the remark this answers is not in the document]
dannygoldstein: Oops, this was just a relic of some local debugging. Thanks

dannygoldstein:
Green now that https://github.com/cesium-ml/baselayer/pull/214 is enabled

skyportal/handlers/api/candidate.py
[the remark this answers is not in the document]
kmshin1397: Nice, yeah my bad for missing this during my review as well

skyportal/handlers/api/photometry.py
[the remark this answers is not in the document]
kmshin1397: So just to clarify, you need to manually call `finalize_transaction` or `verify_permissions`, yes? There's no automatic checking in the base handler before finishing the request or anything like that?

kmshin1397 (approved):

acrellin (approved):
Looks good. I wonder if we should automatically call either verify_permissions or finalize_transaction (depending on the access modes of records in session) in the base handler

dannygoldstein:
> Looks good. I wonder if we should automatically call either verify_permissions or finalize_transaction (depending on the access modes of records in session) in the base handler

Yeah, I thought about this - but I had a hard time thinking of where the right place to put the automatica call would be because generally most handlers do stuff after calling `verify_permissions`, such as send back `self.success` or send stuff over the websockets

skyportal/models.py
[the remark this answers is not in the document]
stefanv: Is this written up adequately in the docs and comments?  If not, we need to do that now, since Danny is right here.

skyportal/handlers/api/photometry.py
```diff
@@ -657,6 +657,7 @@ def post(self):
         except ValidationError as e:
             return self.error(e.args[0])
 
+        # DBSession().commit()
```
stefanv, line 660: Leftover?

dannygoldstein: fixed

skyportal/tests/api/test_candidates.py
```diff
@@ -1315,3 +1315,30 @@ def test_candidate_list_not_saved_to_all_selected_groups(
         )
         == 0
     )
+
+
+def test_correct_spectra_and_photometry_returned_by_candidate(
```
acrellin, line 1320: thank for adding this. a bit more peace of mind

PR #1805, opened 2021-03-22
Consistent photometry plot colors and also now sorts legend by effective wavelength of filters

scizen9:
This is an attempt to fix [Issue 91](https://github.com/fritz-marshal/fritz-beta-feedback/issues/91).

As I experimented, I realized that the legend was not sorted on the filter wavelength.  That can add to the confusion,
so I now gather the effective wavelength for each filter and sort based on that.  This way it will consistently display
filters in the legend in wavelength order with bluest at the top and reddest at the bottom.


scizen9:
Tested with photometry data from ZTF21aakilyd, which contains upper limits and spans a large wavelength range.

scizen9:
I can't include these data in the demo, but here is an example that spans a lot of the wavelengths and includes many instruments: 
image

Notice the consistent colors for the same filter across instruments and the sorting of the filters by effective wavelength.

kmshin1397 (approved):
Look good!

skyportal/plot.py
```diff
@@ -146,38 +146,49 @@ class CheckboxWithLegendGroup(CheckboxGroup):
...
-            cmap = cmap_opt
-            cmap_limits = (3000, 10000)
-        elif 10000 < wave < 1e5:
-            wave = np.log10(wave)
-            cmap = cmap_ir
-            cmap_limits = (4, 5)
-        else:
-            raise ValueError('wavelength out of range for color maps')
-
+def get_effwave(bandpass_name):
```
acrellin, line 155: Does `effwave` mean `effective_wavelength`? If so I'd opt for the latter naming just to be improve readability

skyportal/plot.py
```diff
@@ -146,38 +146,49 @@ class CheckboxWithLegendGroup(CheckboxGroup):
...
+    bandpass = sncosmo.get_bandpass(bandpass_name)
+    return float(bandpass.wave_eff)
+
+
+def get_color(wave):
```
acrellin, line 160:
```suggestion
def get_color(wavelength):
```
and below accordingly

acrellin (approved):
Some minor naming suggestions, otherwise :+1:

skyportal/plot.py
scizen9: Good idea.  Recent commit implements this.

skyportal/plot.py
scizen9: Fixed with recent commit.

PR #1808, opened 2021-03-23
Use frame width for plot scaling

kmshin1397:
Closes fritz-marshal/fritz-beta-feedback#91
Should also close #1803 
plot

Use the `frame_width` parameter instead of `width` in the photometry/spectroscopy plots to remove the legend from width/aspect ratio calculations. Also some other minor tweaks to the plot layouts (spectroscopy sliders are slightly narrower to account for padding; labels wrap to next line a little earlier)

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -342,7 +342,7 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
 
     labels = []
     for i, datarow in data.iterrows():
-        label = f'{datarow["instrument"]}/{datarow["filter"]}'
+        label = f'{datarow["instrument"]}/{datarow["filter"]}asfasdfasfsafdsa'
         if datarow['origin'] is not None:
             label += f'/{datarow["origin"]}'
         labels.append(label)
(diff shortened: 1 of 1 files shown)


skyportal/plot.py
```diff
@@ -342,7 +342,7 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
 
     labels = []
     for i, datarow in data.iterrows():
-        label = f'{datarow["instrument"]}/{datarow["filter"]}'
+        label = f'{datarow["instrument"]}/{datarow["filter"]}asfasdfasfsafdsa'
```
acrellin, line 345: Is this supposed to be here?

acrellin (approved):
Aside from the possible debug code noted above, :+1:

skyportal/plot.py
kmshin1397: Oops, no that's leftover test code good catch

scizen9 (approved):
Thanks @kmshin1397!  Lot of work to solve this problem.

PR #1810, opened 2021-03-24
Commentable Base class

guynir42:
I added a base class that is the target for all comments and annotations (and in the future maybe other meta-data classes). 

At this point I haven't changed anything outside `models.py` because I want feedback on the implementation. 

There are two main issues to solve:
[ ] how do we migrate all the sub-classes into a new table with new IDs. 
[ ] I changed the `Obj` primary key to be a numeric id and added a `name` column to capture the ZTF name. 

The last point will require a big refactor. I think it is the right thing to do, and it has been the right thing to do from the start. 
It will be painful but it makes more sense to have `Obj` using the same integer identifier as all other classes, and refer to its name by the `name` property. 

Let me know what you think.

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -547,13 +547,30 @@ def token_groups(self):
 Token.groups = token_groups
 
 
-class Obj(Base, ha.Point):
+class CommentableBase(Base):
+    """
+    A base class for all other classes that can be assigned comments and annotations.
+    E.g., Obj, Spectrum, Photometry, Thumbnails
+    """
+
+    # id should be defined by the baselayer BaseMixin
+    commentable_type = sa.Column(sa.String)
+
+    __mapper_args__ = {
+        'polymorphic_identity': 'commentable_base',
+        'polymorphic_on': 'commentable_type',
+    }
+
+
+class Obj(CommentableBase, ha.Point):
     """A record of an astronomical Object and its metadata, such as position,
     positional uncertainties, name, and redshift."""
 
     update = public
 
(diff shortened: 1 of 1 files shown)


skyportal/models.py
```diff
@@ -547,13 +547,30 @@ def token_groups(self):
...
+class Obj(CommentableBase, ha.Point):
     """A record of an astronomical Object and its metadata, such as position,
     positional uncertainties, name, and redshift."""
 
     update = public
 
-    id = sa.Column(sa.String, primary_key=True, doc="Name of the object.")
+    # id = sa.Column(sa.String, primary_key=True, doc="Name of the object.")
+    name = sa.Column(sa.String, primary_key=True, doc="Name of the object.")
+    id = sa.Column(sa.Integer, sa.ForeignKey('commentablebases.id'), primary_key=True)
```
acrellin, line 573: This will just be inherited from `Base`, like all other `Base`-based models, so no need for this

acrellin: And I don't think we should be changing `Obj.id` to an int and moving the current `Obj.id` to `Obj.name`

acrellin (commented):
I'll take a closer look soon, but this feels overly complicated to me. I think we can use a much simpler class inheritance schema

dmitryduev (commented):
@guynir42 so what is your plan on this one?

stefanv:
We may need to live with a compromise for now, since this is a somewhat invasive change.

guynir42:
@stefanv I've moved to a different design using a mixin class. see #1841 
I guess I can close this PR for now.

PR #1813, opened 2021-03-24
Use a proxy test server service for caching and mocking LT test requests/responses

### LINKED ISSUE #1792: Implement a general approach for mocking external services for testing

kmshin1397:
This PR sets up a new microservice on testing runs for a `test_server` which acts as a proxy between the test SP instance and external APIs.

The process, explained using LT as an example: 
- The LT host/port is set to the local test server `localhost:64502` instead of the real LT host
- The `test_server` configs are set up so that the proxy server knows that if it sees requests to the LT route, to forward such requests to the real LT address
- When the test server gets a response from LT, it sends that back to the SP API. 
  - For WSDL responses specifically (like the LT GET request), the response body will have the real LT address within the service definitions in the response. This is overwritten to also point to the local test server so that when the SOAP service is invoked, we can intercept that request as well.
- We wrap the logic for forwarding requests/responses within a VCR context so that they are cached for future test runs.
- The VCR cache files are deleted every 7 days by default (number of days are configurable)
- We also define a custom VCR request matcher to identify when an incoming request is in the cache. This checks for not only matching URLs and method (POST vs GET), but also for matching `mode` and `device` within the request body XML.

Also in this PR:
- Shorten the default length of randomly generated UUID's in test fixtures (by using the `uuid4().hex`). This should not only help with flakiness in frontend tests due to unrealistically long names, but was relevant to this PR specifically because the LT API expects target names (our obj IDs) to be at most 32 characters.

Closes #1792

### DIFF
--- a/skyportal/services/test_server/test_server.py
+++ b/skyportal/services/test_server/test_server.py
@@ -44,6 +44,8 @@ def lt_request_matcher(r1, r2):
     """
     Helper function to help determine if two requests to the LT API are equivalent
     """
+
+    # Check that the request modes are matching (should be either "request" or "abort")
     r1_request_mode = re.findall(
         r"mode=&quot;[a-zA-Z]+&quot;", r1.body.decode("utf-8")
     )[0]
@@ -53,6 +55,7 @@ def lt_request_matcher(r1, r2):
         else None
     )
 
+    # For "request" calls, check that the "Device" parameters match up
     r1_device_matches = (
         re.findall(r"&lt;Device name=&quot;.+?&quot;", r1.body.decode("utf-8"))
         if r1.body is not None
@@ -66,6 +69,8 @@ def lt_request_matcher(r1, r2):
     )
     r2_device = r2_device_matches[0] if (len(r2_device_matches) != 0) else None
 
+    # A request matches an LT request if the URL matches, the POST/GET matches,
+    # the mode ("request" or "abort") matches, and the instrument ("Device") matches.
     assert (
         r1.uri == r2.uri
         and r1.method == r2.method
--- a/skyportal/services/test_server/test_server.py
+++ b/skyportal/services/test_server/test_server.py
@@ -1,12 +1,13 @@
 import vcr
-
+import tornado.ioloop
+import tornado.httpserver
 import tornado.web
-
 from suds import Client
 import requests
 
-from baselayer.log import make_log
 from baselayer.app.env import load_env
+from baselayer.log import make_log
+
 
 env, cfg = load_env()
 log = make_log("testserver")
@@ -138,3 +139,17 @@ def make_app():
             (".*", TestRouteHandler),
         ]
     )
+
+
+if __name__ == "__main__":
+    env, cfg = load_env()
+    log = make_log("testserver")
+
+    if "test_server" in cfg:
+        app = make_app()
+        server = tornado.httpserver.HTTPServer(app)
+        port = cfg["test_server.port"]
+        server.listen(port)
+
+        log(f"Listening for test HTTP requests on port {port}")
+        tornado.ioloop.IOLoop.current().start()
--- a/skyportal/services/test_server/test_server.py
+++ b/skyportal/services/test_server/test_server.py
@@ -1,3 +1,8 @@
+import glob
+import datetime
+import re
+import os
+
 import vcr
 import tornado.ioloop
 import tornado.httpserver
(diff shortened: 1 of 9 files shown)


kmshin1397:
LT folks fixed things on their end so the test calls are successful again. When running the followup tests locally, they all pass and generated this VCR YAML file (sent as .txt since GitHub doesn't support attaching YAML files to comments):

[test_server_recordings_2021-03-24.txt](https://github.com/skyportal/skyportal/files/6199997/test_server_recordings_2021-03-24.txt)

mcoughlin:
@kmshin1397 Oh this is real exciting! Can you help me understand the reason for all the hex changes?

kmshin1397:
> @kmshin1397 Oh this is real exciting! Can you help me understand the reason for all the hex changes?

@mcoughlin  The hex changes aren't really necessary - they just serve to get a shorter random string than was provided by just `uuid.uuid4()` because the hyphens are removed. For example:
image

Looking into CI failure now

skyportal/services/test_server/test_server.py
```diff
@@ -0,0 +1,218 @@
...
+    if (today - current_file_date).days > refresh_cache_days:
+        # Delete old cache and return new file path
+        os.remove(current_file)
+        return f"cache/test_server_recordings_{today.isoformat()}.yaml"
+
+    # Cache is still valid
+    return current_file
+
+
+def lt_request_matcher(r1, r2):
```
mcoughlin, line 43: Since we know we will also need to do something like this for LCO and ZTF and so on, is there anything to abstract out here now? Or are the specifics of the bodies enough that we will need to copy and paste things?

test_config.yaml
```diff
@@ -40,3 +40,23 @@ notifications:
...
+
+services:
+  paths:
+    - ./baselayer/services
+    - ./skyportal/services
+
+test_server:
+  port: 64502
+  # Reset the cache every * days
+  refresh_cache_days: 7
```
mcoughlin, line 52: This refresh rate will be changeable for each system? We won't want to have to test LCO too often given the program restrictions, for example.

mcoughlin (commented):
Just a few ideas / comments. Overall looks great to me.

test_config.yaml
```diff
@@ -40,3 +40,23 @@ notifications:
...
+  redirects:
+    # List of routes listened for by the test server, and the real URLs that
+    # requests should be redirected to. Requests will be forwarded by the
+    # test server, cached, and responses played back to the API as if it came
+    # from the real server.
+    /node_agent2/node_agent: http://161.72.57.3:8080 # LT
```
dannygoldstein, line 58: What is `/node_agent2/node_agent`?

dannygoldstein (approved):

skyportal/services/test_server/test_server.py
kmshin1397: Hmm yeah I think this may be the best we can abstract this. There's already decent built-in general options provided by `vcr` like matching on just URL, URL + request type, the entire body, etc. This was written specifically for LT since it's different from the more standard REST APIs like LCO in that the LT requests all go to the same route `/node_agent2/node_agent` and are mostly just distinguished by what's in the body XML. I imagine things like LCO and ZTF will be more amenable to using built-in options like "look for POST requests to this particular URL/route"

test_config.yaml
kmshin1397: That's a good point that I hadn't thought of. As I have it now, requests for everything (LT, LCO, etc) will all go into the same cache file, and the refresh rate is defined for the entire file. Unfortunately VCR doesn't really have good support for request-specific expiration dates as far as I know, so I think we probably have to do something like a separate cache file for each system in order to support different refresh rates for each one.

test_config.yaml
kmshin1397: This is the API route for LT requests. The full real LT URL is `http://161.72.57.3:8080/node_agent2/node_agent`, but they're not being sent to `http://localhost:cfg["test_server.port"]/node_agent2/node_agent` and this configuration tells the test server that all requests coming in on the `/node_agent2/node_agent` route should be forwarded to the LT server.

To be honest, this is the one part of this setup that I'm a little iffy about. I need a way distinguish which requests want to go to which external system, and the real addresses for those external systems. The method I've settled on is to map API routes to real host address in the configs like this, but I can imagine this raising issues in the future if two external services both have some common route like `"GET /users"` that we want to test and it'll be hard to tell which real APIs to forward the requests to when they're both just being sent to `http://localhost:cfg["test_server.port"]/users` by SP

test_config.yaml
mcoughlin: Ok makes sense!

skyportal/services/test_server/test_server.py
mcoughlin: Ok makes sense!

mcoughlin (approved):

PR #1817, opened 2021-03-25
Stringify altdata in PhotometryTable to avoid React errors

acrellin:


static/js/components/PhotometryTable.jsx
```diff
@@ -38,25 +39,30 @@ const PhotometryTable = ({ obj_id, open, onClose }) => {
             if (isFloat(value)) {
               return value.toFixed(key.includes("jd") ? 8 : 6);
             }
+            if (
+              key === "altdata" &&
+              typeof value === "object" &&
+              value !== null
+            ) {
+              return JSON.stringify(value);
+            }
```
acrellin, line 48: These are the only relevant changes here, everything else is due to our current linter not liking other, pre-existing code

Issue #1814, opened 2021-03-24
altdata not being broadcast correctly in photometry post handler


acrellin:
This request fails with `'Unable to coerce passed JSON to a series of packets. Error was: "Mixing dicts with non-Series may lead to ambiguous ordering."'`:
```
r = requests.post(
    "http://localhost:5000/api/photometry", 
    json={
        'mjd': '59203.14312098', 
        'obj_id': '14gqr', 
        'instrument_id': 1, 
        'origin': 'DESI', 
        'magsys': 'ab', 
        'filter': ['ztfg', 'ztfi', 'ztfr'], 
        'group_ids': [7, 8], 
        'flux': [6.69783593559922e-07, 4.372349179111794e-07, 2.843060842255576e-07], 
        'fluxerr': [6.69783593559922e-09, 4.3723491791117935e-09, 2.843060842255576e-09],
        'zp': 0, 
        'altdata': {'subdir': 'daily', 'trunk': 'coadd', 'tile': '80606', 'date': '20201216'},
    }, 
    headers={"Authorization": f"token {token_id}"}
)      
```
yet this request returns successfully (only `altdata` is different here):
```
r = requests.post(
    "http://localhost:5000/api/photometry", 
    json={
        'mjd': '59203.14312098', 
        'obj_id': '14gqr', 
        'instrument_id': 1, 
        'origin': 'DESI', 
        'magsys': 'ab', 
        'filter': ['ztfg', 'ztfi', 'ztfr'], 
        'group_ids': [7, 8], 
        'flux': [6.69783593559922e-07, 4.372349179111794e-07, 2.843060842255576e-07], 
        'fluxerr': [6.69783593559922e-09, 4.3723491791117935e-09, 2.843060842255576e-09],
        'zp': 0, 
        'altdata': {
            'subdir': ['daily', 'daily', 'daily'], 
            'trunk': ['coadd', 'coadd', 'coadd'], 
            'tile': ['80606', '80606', '80606'], 
            'date': ['20201216', '20201216', '20201216'], 
            'panel': ['1', '1', '1'],
        },
    }, 
    headers={"Authorization": f"token {token_id}"}
)      
```

PR #1818, opened 2021-03-25
Fix handling of photometry posted with altdata

acrellin:
This PR fixes & greatly simplifies how altdata is handled in `PhotometryHandler.post`. Now, a list of objects can be provided or a single object can be broadcasted to all photometry points. Tests are added.

Closes https://github.com/skyportal/skyportal/issues/1814

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -145,6 +145,16 @@ class PhotometryHandler(BaseHandler):
 
         if "altdata" in data and not data["altdata"]:
             del data["altdata"]
+        if "altdata" in data:
+            if isinstance(data["altdata"], dict):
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/photometry.py
```diff
@@ -145,6 +145,17 @@ def standardize_photometry_data(self):
...
             del data["altdata"]
+        if "altdata" in data:
+            if isinstance(data["altdata"], dict):
+                max_num_elements = max(
+                    [
+                        len(data[key])
+                        for key in data
+                        if isinstance(data[key], (list, tuple))
+                    ]
+                    + [1]
```
acrellin, line 156: In case there are no list/tuple values

PR #1820, opened 2021-03-26
Fix db_demo.yaml to use instruments instead of telescopes in photometry and add P48 telescope and ZTF instrument

scizen9:
…hotometry point to ZTF instrument


scizen9:
I could not find precise values for latitude, longitude and elevation for P48.  Those values are from the [Wikepedia article](https://en.wikipedia.org/wiki/Samuel_Oschin_telescope) on the P48 and may not be precise.

data/db_demo.yaml
```diff
@@ -273,6 +294,20 @@ sources:
     group_ids:
       - =public_group_id
       - =program_A
+  - id: ZTFe028h94k
```
dannygoldstein, line 297: are these real names ?

acrellin (approved):

data/db_demo.yaml
scizen9: Yes.  They are the names on the ZTF variable marshal.

dmitryduev: (true, but they are auto-generated and random)

PR #1823, opened 2021-03-29
Invert only grayscale thumbnails when "Invert thumbnails" is set

### LINKED ISSUE #1800: Do not invert RGB images
The new image inversion (in the Profile preferences) is most helpful for B&W/greyscale images, but should not be done for RGB images regardless of the inversion toggle. Usually RGB images are helpful for seeing the colors of the host/reference field and doing a color inversion does not make sense.

kmshin1397:
This PR modifies the "Invert thumbnails" UI preference to a "Invert grayscale thumbnails" option so that RGB thumbnails are not inverted. To facilitate this, an `is_grayscale` Boolean column is added to the `thumbnails` table, populated via a Python function `image_is_grayscale` which looks roughly at the color distribution in the image to classify it as colored or grayscale. This is a more forgiving way of choosing which thumbnails should be inverted than, say, checking the color mode in the PNG header, since a PNG file can be marked as colored (encodes all three of RGB color bands) but still only consist of black/white pixels. 

I considered doing this computation on the fly on the front-end instead of adding a new data column, but it was 1) much easier to do on the Python side with `PIL` than with JavaScript and 2) the overhead for computing it on thumbnail inserts was not that bad, so might as well just do it once and cache it in the database.

I did some quick benchmarking of the overhead introduced by running the grayscale classifier logic on inserts to `thumbnails`, and there was roughly ~2 ms added per POST call to `/api/thumbnail`. I think this is acceptable, as this would still place the average POST thumbnail call at around 30ms based on the latest Kowalski performance report (basically the same as our Source/Candidate HEAD calls).

invert

Closes #1800


kmshin1397:
Test failures are real - investigating

skyportal/models.py
```diff
@@ -3129,6 +3130,25 @@ class Thumbnail(Base):
...
+def classify_thumbnail_grayscale(mapper, connection, target):
+    if target.file_uri is not None:
+        target.is_grayscale = image_is_grayscale(target.file_uri)
+    else:
+        try:
+            target.is_grayscale = image_is_grayscale(
+                requests.get(target.public_url, stream=True).raw
+            )
+        except requests.exceptions.RequestException:
+            pass
```
acrellin, line 3151: Have you tested that the front-end doesn't error when `is_grayscale` is null?

kmshin1397: Ah no, not specifically but I don't think that should ever be the case because I've set the data model column to have `nullable=False` with `default=False`. Basically, all thumbnails will be assumed to be colored by default to start out until the event listener to classify it runs. The `try...except...` here will catch errors if external URL thumbnails (like PS1) lead to bad requests for whatever reason and those thumbnails will just keep the default `is_grayscale  = False`.

acrellin: Sounds good

skyportal/utils/thumbnail.py
```diff
@@ -0,0 +1,38 @@
+from PIL import Image, ImageStat, UnidentifiedImageError
+
+
+def image_is_grayscale(file, thumb_size=40, MSE_cutoff=22, adjust_color_bias=True):
```
dmitryduev, line 4: Please make these configurable in `config.yaml` and grab them in the calls from there so that we could tweak things easily if necessary.

dmitryduev (commented):
Thanks @kmshin1397! LGTM, just one request and should be good to go in.

acrellin:
@kmshin1397 What's the status of this PR?

kmshin1397:
@acrellin Just waiting for re-review from Dima, I think. I made the configs change he requested and re-tagged him just now

kmshin1397:
Ah wait I need to update the migrations so we don't have branching I think

dmitryduev (approved):
Looks great, thanks @kmshin1397!

Issue #1824, opened 2021-03-30
Store weather data in separate table rather than telescopes table


acrellin:
We are seeing permissions errors when attempting to update weather data, as a user who has read access to a telescope doesn't necessarily have update access.

PR #1825, opened 2021-03-30
Move weather data into separate table

acrellin:
This PR moves weather data from the `Telescope` model into a new `Weather` model, with public update access (this addresses the permissions issues we were running into when attempting to update weather info). Handler is updated accordingly, with minor front-end changes.

Closes https://github.com/skyportal/skyportal/issues/1824

weather


acrellin:
Failure unrelated:
```
FAILED skyportal/tests/api/test_sources.py::test_starlist - Failed: DID NOT R...
```

alembic/versions/8f086f8c0c21_move_weather_data_to_separate_table.py
```diff
@@ -0,0 +1,70 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        'telescopes',
+        sa.Column(
+            'weather_retrieved_at',
+            postgresql.TIMESTAMP(),
```
kmshin1397, line 53: Should this be of type `sa.DateTime()`?

dmitryduev (changes requested):
Looks good @acrellin, just one comment: when I start from fresh, this is what I see:
image
So I guess the demo data needs to be updated. Otherwise should be good to go.

skyportal/models.py
```diff
@@ -1588,6 +1588,7 @@ class Telescope(Base):
     skycam_link = sa.Column(
         URLType, nullable=True, doc="Link to the telescope's sky camera."
     )
+    weather_link = sa.Column(URLType, doc="Link to the preferred weather site")
```
profjsb, line 1591:
```suggestion
    weather_link = sa.Column(URLType, doc="Link to the preferred weather site.")
```

skyportal/models.py
```diff
@@ -1722,6 +1715,28 @@ def ephemeris(self, time):
         }
 
 
+class Weather(Base):
+    update = public
+
+    weather_info = sa.Column(JSONB, doc="Latest weather information")
```
profjsb, line 1721:
```suggestion
    weather_info = sa.Column(JSONB, doc="Latest weather information.")
```

alembic/versions/8f086f8c0c21_move_weather_data_to_separate_table.py
```diff
@@ -0,0 +1,70 @@
...
+        sa.Column('retrieved_at', sa.DateTime(), nullable=True),
+        sa.Column('telescope_id', sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ['telescope_id'], ['telescopes.id'], ondelete='CASCADE'
+        ),
+        sa.PrimaryKeyConstraint('id'),
+    )
+    op.create_index(
+        op.f('ix_weathers_created_at'), 'weathers', ['created_at'], unique=False
+    )
```
profjsb, line 38: Does this need to be an index?

profjsb (approved):

alembic/versions/8f086f8c0c21_move_weather_data_to_separate_table.py
acrellin: This is what alembic specifies for a `sa.DateTime` column. And the migration checks seem to like this, so seems :+1:

acrellin:
> Looks good @acrellin, just one comment: when I start from fresh, this is what I see:
> image
> So I guess the demo data needs to be updated. Otherwise should be good to go.

@dmitryduev you need a valid weather API key in your local config in order for this to work (look in the fritz-deploy secrets file)

alembic/versions/8f086f8c0c21_move_weather_data_to_separate_table.py
acrellin: All models that inherit from Base have this indexed column.

dmitryduev (approved):
Right, thanks :)

PR #1826, opened 2021-03-30
Refactor GroupAdmissionRequest permissions and add corresponding tests

kmshin1397:
This PR refactors the GroupAdmissionRequest handlers to use the new permissions system. I also catch the `AccessErrors` thrown by the accessible functions so that we can return more informative error messages. Finally, I added a new set of API tests to test these handlers since we only had a couple of frontend tests for them which didn't cover most of the access error cases. I just wanted to get out a smaller PR for just this one class first to make sure I'm on the right track with the new infrastructure. 

Also discovered/fixed a bug in the single GET handler for admission requests (`to_json()` should've been `to_dict()`)


skyportal/handlers/api/group_admission_request.py
```diff
@@ -190,7 +189,14 @@ def post(self):
...
                 )
             )
-        self.verify_and_commit()
+
+        try:
+            self.verify_and_commit()
+        except AccessError:
+            return self.error(
+                "Insufficient permissions: group admission requests cannot be made on behalf of others."
+            )
```
dannygoldstein, line 198:
```suggestion
        except AccessError as e:
            return self.error(
                f"Insufficient permissions: group admission requests cannot be made on behalf of others. (Original exception: {e.args[0]})"
            )
```

skyportal/handlers/api/group_admission_request.py
```diff
@@ -236,13 +242,19 @@ def patch(self, admission_request_id):
...
+            admission_request = GroupAdmissionRequest.get_if_accessible_by(
+                admission_request_id,
+                self.current_user,
+                raise_if_none=True,
+                mode="update",
+            )
+        except AccessError:
+            return self.error(
+                "Insufficient permissions: group admission request status can only be changed by group admins."
+            )
```
dannygoldstein, line 256: see above - it will be useful for debugging to see the original access error

skyportal/handlers/api/group_admission_request.py
```diff
@@ -277,15 +289,18 @@ def delete(self, admission_request_id):
...
+            admission_request = GroupAdmissionRequest.get_if_accessible_by(
+                admission_request_id,
+                self.current_user,
+                raise_if_none=True,
+                mode="delete",
+            )
+        except AccessError:
+            return self.error(
+                "Insufficient permissions: only the requester can delete a group admission request."
+            )
```
dannygoldstein, line 302: see above

skyportal/models.py
```diff
@@ -396,6 +397,12 @@ class Group(Base):
 class GroupAdmissionRequest(Base):
     """Table tracking requests from users to join groups."""
 
+    read = ComposedAccessControl(
+        AccessibleIfUserMatches('user'), accessible_by_group_admins, logic='or'
+    )
```
dannygoldstein, line 402:
```suggestion
    read = AccessibleIfUserMatches('user') | accessible_by_group_admins
```

dannygoldstein (approved):

dannygoldstein:
failures unrelated

dannygoldstein:
looks great @kmshin1397. and thanks for patching up the test suite to catch some of the API vulnerabilities.

acrellin (approved):
Once Danny's comments are addressed, :+1: from me. Good that we're catching the AccessError here -- I've noticed the TB still shows up in the logs when it's raised and uncaught, which is not ideal.

dannygoldstein:
Failures unrelated

PR #1827, opened 2021-03-30
Pin latest BL

dmitryduev:
...where we updated the `pyjwt` requirement to `>2.0.1`.


acrellin (approved):
Will merge when green

Issue #1828, opened 2021-03-30
Query parameter radius units


AlexGKim:
In the documentation it would be useful to give the units for the query parameter radius in the source retrieval.  What is it BTW?

http://skyportal.io/docs/api.html#tag/sources/paths/~1api~1sources~1obj_id/patch

acrellin:
Do you mean in the [GET](http://skyportal.io/docs/api.html#tag/sources/paths/~1api~1sources/get) docs? You linked to the PATCH handler. `ra`, `dec` and `radius` are all in decimal degrees. 

Addressed in https://github.com/skyportal/skyportal/pull/1829

AlexGKim:
Great, this is addressed in that issue.  For some reason I didn't catch that one when I searched.

PR #1829, opened 2021-03-30
Include units for ra, dec, radius in SourceHandler.get docs

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1828


acrellin:
self-merging this docs-only patch

PR #1833, opened 2021-04-01
Update SourceHandler (and related handlers) and NewsfeedHandler to use new permissions logic

acrellin:

PR #1834, opened 2021-04-01
Control access to Photometry via Obj hybrid properties

kmshin1397:
This PR refactors the detection hybrid properties in the `Obj` class to hybrid methods which take in a user instance so that those are computed based on accessible Photometry rows instead of all rows. The Candidates and Sources GET handlers were updated to reflect this; this should take care of Photometry leaks in those handlers. Also added/updated some tests to check that hidden photometry data isn't reflected in returned candidates/sources.

I don't think I touched anything that you did in your #1833 PR @acrellin so we shouldn't have any merge conflicts. I also didn't uncomment the `self.verify_and_commit()` lines in the SourcesHandler, as that'll be done by that PR.

Finally, set `read = accessible_by_groups_members | accessible_by_owner` on Photometry rows to make it clear that whoever posts a photometry point can read it. This was already there logically since the owner's single user group is added to the Photometry groups automatically, but wasn't very clear from the `read = accessible_by_groups_members` definition.

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -22,7 +22,7 @@ from sqlalchemy import cast, event
 from sqlalchemy.dialects import postgresql as psql
 from sqlalchemy.dialects.postgresql import ARRAY
 from sqlalchemy.dialects.postgresql import JSONB
-from sqlalchemy.ext.hybrid import hybrid_property
+from sqlalchemy.ext.hybrid import hybrid_property, hybrid_method
 from sqlalchemy.orm import relationship, joinedload
 from sqlalchemy.schema import UniqueConstraint
 from sqlalchemy_utils import URLType, EmailType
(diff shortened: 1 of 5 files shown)


skyportal/models.py
```diff
@@ -718,99 +718,107 @@ class Obj(Base, ha.Point):
...
     @peak_detected_at.expression
-    def peak_detected_at(cls):
+    def peak_detected_at(cls, user):
         """UTC ISO date at which the object was detected at peak magnitude above a given S/N (3.0 by default)."""
         return (
-            sa.select([Photometry.iso])
-            .where(Photometry.obj_id == cls.id)
-            .where(Photometry.snr.isnot(None))
-            .where(Photometry.snr > PHOT_DETECTION_THRESHOLD)
+            Photometry.query_records_accessible_by(
```
acrellin, line 789: does it matter that this used to return a `select` instance and now returns a `query` instance? not sure if they're interchangeable, so just making sure

acrellin (approved):
Looks good

skyportal/models.py
kmshin1397: No I don't think it matters. I didn't look explicitly for it in the SQLAlchemy docs, but all the tests that use these passed locally for me. We have several tests in test_sources.py for things like filtering on peak mag and latest detection which should use these expressions.

acrellin: nice

skyportal/models.py
```diff
@@ -718,99 +718,107 @@ class Obj(Base, ha.Point):
...
-            sa.select([sa.func.max(Photometry.iso)])
-            .where(Photometry.obj_id == cls.id)
-            .where(Photometry.snr.isnot(None))
-            .where(Photometry.snr > PHOT_DETECTION_THRESHOLD)
+            Photometry.query_records_accessible_by(
+                user, columns=[sa.func.max(Photometry.iso)], mode="read"
+            )
+            .filter(Photometry.obj_id == cls.id)
+            .filter(Photometry.snr.isnot(None))
+            .filter(Photometry.snr > PHOT_DETECTION_THRESHOLD)
```
dannygoldstein, line 742: nicely done

dannygoldstein: bet this translates to a pretty complex sql query under the hood! this is a good illustration of the expressive power of the ORM

dannygoldstein (approved):
excellent

dannygoldstein:
failure unrelated (test get_ztf_ref_url)

PR #1837, opened 2021-04-02
Update permissions for instrument and observing run handlers

kmshin1397:
Refactor the `instrument.py` and `observingrun.py` handlers to use new permissions system. Also added a test for observing runs to make sure it doesn't leak groups nested in the run's assignments.  

@dannygoldstein There was a TODO comment about not turning on verify_and_commit here until the refactor due to "changes to persistent records that shouldn't be flushed". I wasn't sure if I followed exactly what you meant by that - let me know if there's anything more I need to do here.

### DIFF
--- a/skyportal/handlers/api/observingrun.py
+++ b/skyportal/handlers/api/observingrun.py
@@ -102,43 +102,24 @@ class ObservingRunHandler(BaseHandler):
                   schema: Error
         """
         if run_id is not None:
-            run = (
-                DBSession()
-                .query(ObservingRun)
-                .options(
-                    joinedload(ObservingRun.assignments)
-                    .joinedload(ClassicalAssignment.obj)
-                    .joinedload(Obj.thumbnails),
-                    joinedload(ObservingRun.assignments).joinedload(
-                        ClassicalAssignment.requester
-                    ),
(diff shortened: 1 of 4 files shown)


dannygoldstein:
Looks like some related CI API failures ?

skyportal/handlers/api/observingrun.py
```diff
@@ -102,60 +102,39 @@ def get(self, run_id=None):
...
-                .first()
-            )
-
-            if run is None:
-                return self.error(
-                    f"Could not load observing run {run_id}", data={"run_id": run_id}
-                )
+            # These are all read=public, including Objs
+            options = [
+                joinedload(ObservingRun.assignments)
```
acrellin, line 107: You'll probably need to something like `return_data = baselayer.app.model_util.recursive_to_dict(return_data)` in order for these loaded relationships to actually be returned, e.g. https://github.com/skyportal/skyportal/pull/1833/files#diff-0200612f282483e1172207de90a4db4788b3ca5ce55ab5105938d876e11eb2e0R731

dannygoldstein (approved):

acrellin (approved):

PR #1841, opened 2021-04-05
Comment on spectra

guynir42:
Expanding the notion of comments and annotations on the objects themselves, we will be able to add comments/annotations also unto a particular spectrum. 

To do this I've added a `CommentMixin` that has all the columns from the regular `Comment` model, but can be inherited into new types of comments. This does not fundamentally change the database tables, only adds new tables, and two new columns to `Spectrum`.


guynir42:
@dmitryduev @dannygoldstein @stefanv @acrellin Please go over this very *preliminary* PR and tell me if the logic looks ok in general. I will then proceed to add the annotations and the handlers and the front end integration (possibly in a separate PR). 
Thanks!

guynir42:
Hi @stefanv @acrellin I am seeing some weird test failures. 
First is the deploy documentation. Please have a look at the errors it looks like it has some problem ingesting a Mixin class. 
Secondly, there are failures on the regular tests, and I'm not sure if they are related. It looks like maybe some of them are flaky in general (followup requests, etc). 
Lastly I need some help with migration. Mostly I need to add a new table for `CommentOnSpectrum`. Any pointers on where to start? 
Thanks!

skyportal/models.py
```diff
@@ -2081,37 +2084,59 @@ class Comment(Base):
...
+    def author(cls):
+        return relationship(
+            "User",
+            back_populates=cls.backref_name(),
+            doc="Comment's author.",
+            uselist=False,
+            foreign_keys=f"{cls.__name__}.author_id",
+        )
+
+    @declared_attr
```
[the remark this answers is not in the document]
guynir42: > Are you sure these non-relationship columns should be decorated w/ `@declared_attr`?

I think so, @acrellin, I am working off this reference: https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/mixins.html#mixing-in-columns

I think the foreign keys cannot be instantiated in the mixin, only in the sub-class.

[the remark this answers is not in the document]
acrellin: Yep, looks right. Thanks

skyportal/models.py
```diff
@@ -647,6 +649,15 @@ class Obj(Base, ha.Point):
         doc="Comments posted about the object.",
     )
 
+    comments_on_spectra = relationship(
+        'CommentOnSpectrum',
+        back_populates='obj',
+        cascade='save-update, merge, refresh-expire, expunge, delete',
```
acrellin, line 655: You've already got the ondelete cascade set on the obj_id foreign key on the child class, so this should probably just be 
```suggestion
        cascade='save-update, merge, refresh-expire, expunge',
```

acrellin (commented):
Looking good, thanks @guynir42! I'll look into the docs building / mixin errors

skyportal/models.py
```diff
@@ -2053,16 +2064,8 @@ def get_taxonomy_usable_by_user(taxonomy_id, user_or_token):
...
-
-    create = AccessibleIfRelatedRowsAreAccessible(obj='read')
-
-    read = accessible_by_groups_members & AccessibleIfRelatedRowsAreAccessible(
-        obj='read'
-    )
-
-    update = delete = AccessibleIfUserMatches('author')
+class CommentMixin:
+    """All the basic columns needed for comments on objects, spectra, etc."""
```
acrellin, line 2068:
```suggestion
```

acrellin:
@guynir42 The latest commit above fixes docs building locally -- hopefully it does on CI as well.

acrellin:
@guynir42 Here are the commands for generating a migration script: https://skyportal.io/docs/migrations.html

acrellin:
Docs building is fixed here ✅

guynir42:
@acrellin I tried making the table names a little more sensible (`comments_on_spectra` rather than `comments_on_spectrums`) but now I have no idea how to make the migration script work...

acrellin:
@guynir42 do you mean you changed the table name after creating the migration script? If so, I'd just nix the old migration script and start the alembic process over from the beginning.

alembic/versions/0595e877f471_add_streamphotometry_table.py
```diff
@@ -0,0 +1,63 @@
+"""Add StreamPhotometry table
```
acrellin, line 1: Why is this committed here?

guynir42: I think what happened is I tried to re-do the migration after pulling in master into my master, but I didn't pull master into this new branch. 

Next time I will stamp my database before I start working...

kmshin1397:
@guynir42 It looks like the migration is failing because it is attempting to recreate the `comment_type` enumeration, which was used to distinguish between text comments and redshift comments. Since we no longer use the comments to record redshift, I think the simplest way to fix things would be to just drop the `ctype` column entirely from the comments table (and adjust the migration script accordingly)

guynir42:
@acrellin @kmshin1397 I am looking at the errors and they seem unrelated.
a) a lot of "followup requests" test are failing, but I remember these used to be flaky. 
b) I have one API failure on `test_candidates` but I got a status 504 which seems to be unrelated to the test, I think. 
In any case, I think we can do with another review at last...

skyportal/handlers/api/comment.py
```diff
@@ -26,7 +27,7 @@ def users_mentioned(text):
 
 class CommentHandler(BaseHandler):
     @auth_or_token
-    def get(self, comment_id):
+    def get(self, comment_id, comment_on_what=None):
```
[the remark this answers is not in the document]
kmshin1397: Agree with Ari that it would be simpler to just set the default to `"object"` here and not have the `if comment_on_what is None:` line at the beginning of the handler.

skyportal/handlers/api/comment.py
```diff
@@ -48,9 +57,29 @@ def get(self, comment_id):
               application/json:
                 schema: Error
         """
-        comment = Comment.get_if_accessible_by(
-            comment_id, self.current_user, raise_if_none=True
-        )
+
+        if comment_on_what is None:
+            comment_on_what = 'object'
```
kmshin1397, line 62:
```suggestion
```
Don't need this if you set the default in the function signature

skyportal/handlers/api/comment.py
```diff
@@ -48,9 +57,29 @@ def get(self, comment_id):
...
+
+        if type(comment_on_what) != str:
+            return self.error(
+                f'Wrong input type {type(comment_on_what)} given as "comment_on_what" argument'
+            )
```
kmshin1397, line 67:
```suggestion
```
I don't think this actually does anything. I believe Tornado will always just pass the argument parsed from the URL as a string. For example, if you did `GET /api/comments/{comment_id}/123`, then `comment_on_what = "123"` and not the int.  I think just have the error case with the unknown type in the `else` block below is sufficient.

skyportal/handlers/api/comment.py
```diff
@@ -48,9 +57,29 @@ def get(self, comment_id):
...
+                comment_id, self.current_user, raise_if_none=True
+            )
+        elif comment_on_what.lower() == "spectrum":
+            comment = CommentOnSpectrum.get_if_accessible_by(
+                comment_id, self.current_user, raise_if_none=True
+            )
+        # add more options using elif
+        else:
+            return self.error(
+                f'Wrong input "{comment_on_what}" given as "comment_on_what" argument.'
```
kmshin1397, line 80:
```suggestion
                f'Unsupported input "{comment_on_what}" given as "comment_on_what" argument.'
```
I think this would just be a bit clearer

skyportal/handlers/api/comment.py
```diff
@@ -169,7 +209,7 @@ def post(self):
         return self.success(data={'comment_id': comment.id})
 
     @permissions(['Comment'])
-    def put(self, comment_id):
+    def put(self, comment_id, comment_on_what=None):
```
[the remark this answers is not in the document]
kmshin1397: ^

skyportal/handlers/api/comment.py
```diff
@@ -206,16 +262,36 @@ def put(self, comment_id):
...
-            comment_id, self.current_user, mode="update", raise_if_none=True
-        )
+
+        if comment_on_what is None:
+            comment_on_what = 'object'
+
+        if type(comment_on_what) != str:
+            return self.error(
+                f'Wrong input type {type(comment_on_what)} given as "comment_on_what" argument'
+            )
```
kmshin1397, line 272: Same comments here as in the GET

skyportal/handlers/api/comment.py
```diff
@@ -206,16 +262,36 @@ def put(self, comment_id):
...
+        elif comment_on_what.lower() == "spectrum":
+            schema = CommentOnSpectrum.__schema__()
+            c = CommentOnSpectrum.get_if_accessible_by(
+                comment_id, self.current_user, mode="update", raise_if_none=True
+            )
+        # add more options using elif
+        else:
+            return self.error(
+                f'Wrong input "{comment_on_what}" given as "comment_on_what" argument.'
```
kmshin1397, line 287:
```suggestion
                f'Unsupported input "{comment_on_what}" given as "comment_on_what" argument.'
```

skyportal/handlers/api/comment.py
```diff
@@ -260,15 +336,43 @@ def delete(self, comment_id):
...
-            comment_id, self.current_user, mode="delete", raise_if_none=True
-        )
+
+        if comment_on_what is None:
+            comment_on_what = 'object'
+
+        if type(comment_on_what) != str:
+            return self.error(
+                f'Wrong input type {type(comment_on_what)} given as "comment_on_what" argument'
+            )
```
kmshin1397, line 360: Same thing here

skyportal/handlers/api/comment.py
```diff
@@ -260,15 +336,43 @@ def delete(self, comment_id):
...
+                comment_id, self.current_user, mode="delete", raise_if_none=True
+            )
+        elif comment_on_what.lower() == "spectrum":
+            c = CommentOnSpectrum.get_if_accessible_by(
+                comment_id, self.current_user, mode="delete", raise_if_none=True
+            )
+        # add more options using elif
+        else:
+            return self.error(
+                f'Wrong input "{comment_on_what}" given as "comment_on_what" argument.'
```
kmshin1397, line 373:
```suggestion
                f'Unsupported input "{comment_on_what}" given as "comment_on_what" argument.'
```

skyportal/handlers/api/comment.py
```diff
@@ -278,7 +382,7 @@ def delete(self, comment_id):
 
 class CommentAttachmentHandler(BaseHandler):
     @auth_or_token
-    def get(self, comment_id):
+    def get(self, comment_id, comment_on_what=None):
```
kmshin1397, line 385: Same comments for this handler

skyportal/tests/api/test_comments_on_spectrum.py
```diff
@@ -0,0 +1,249 @@
...
+    assert status == 200
+    comment_id = data['data']['comment_id']
+
+    status, data = api('GET', f'comment/{comment_id}/spectrum', token=comment_token)
+
+    assert status == 200
+    assert data['data']['text'] == 'Comment text'
+
+
+def test_add_and_retrieve_comment_group_access(
```
kmshin1397, line 44: Awesome, thanks for making these tests thoroughly cover the access rules as well!

kmshin1397 (commented):
Looks great, @guynir42! A few minor comments, but the crux of the work seems good 👍

skyportal/handlers/api/comment.py
guynir42: Yeah I tried this but it doesn't work. There's something with the handler's default parameters and the regexp that parses the path parameters that just didn't work (it put a None in there, despite having a default parameter...)

guynir42:
@kmshin1397  @acrellin I think this is ready for a (final?) review...

skyportal/handlers/api/comment.py
```diff
@@ -113,6 +143,8 @@ def post(self):
             return self.error("Missing required field `obj_id`")
```
acrellin, line 143: We may want to allow obj_id to be None now, so long as spectrum_id is not None

acrellin (commented):
Looks good. After the below suggestion, and the above change (`comment_on_what` -> `associated_resource_type`), :+1: from me

skyportal/handlers/api/comment.py
kmshin1397: Ah okay, sure no problem then

kmshin1397 (approved):
I like Ari's comment about renaming the variable but otherwise looks good to me

guynir42:
@acrellin is this ready to merge?

acrellin (approved):

PR #1843, opened 2021-04-05
Optimize annotations_info query

kmshin1397:
Since Objs are read public, we can actually skip the candidate filter check entirely, and just check for group membership in `GroupAnnotations`. I tried using `Annotation.query_rows_accessible_by()` but it was like 2x slower than doing it manually like this, at least on the demo data (maybe because read access on annotations has  `AccessibleIfRelatedRowsAreAccessible(obj='read')` defined and public access queries are still slower than not querying it at all).  On the demo data, I'm seeing ~2ms per call with this new version compared to ~7ms on the original query. 

Also, I held off on adding `self.verify_and_commit()` to this as that will only add to the time and it is basically just double-checking the access logic which should already be correct in the main query. I figured we could see how this does on production before deciding it's fast enough to add in the verification guard. 

I think this is the fastest I can get the query for now without doing something drastically different like caching the annotation metadata for each candidate or something like that

### DIFF
--- a/skyportal/handlers/api/internal/annotations_info.py
+++ b/skyportal/handlers/api/internal/annotations_info.py
@@ -8,8 +8,6 @@ from baselayer.app.access import auth_or_token
 from ...base import BaseHandler
 from ....models import (
     DBSession,
-    Obj,
-    Candidate,
     Annotation,
     GroupAnnotation,
 )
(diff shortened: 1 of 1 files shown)


dannygoldstein (approved):

dmitryduev (approved):
Looks good! I suggest we merge this in now and then decide whether we need to do anything drastically different based on the query profiling data.

dmitryduev:
@kmshin1397 please feel free to self-merge once green.

dmitryduev:
btw @dannygoldstein the big photometry post timed out here.

PR #1845, opened 2021-04-05
Refactor of handlers/api/spectrum.py for new permissions method

scizen9:
First attempt.  Did I miss anything?


skyportal/handlers/api/spectrum.py
```diff
@@ -72,39 +70,22 @@ def post(self):
...
-                DBSession()
-                .query(Group)
-                .filter(Group.name == cfg["misc"]["public_group_name"])
-                .all()
-            )
+        # always append the single user group
+        single_user_group = self.associated_user_object.single_user_group
+
+        group_ids = data.pop("group_ids", None)
+        if not group_ids:
```
acrellin, line 77:
```suggestion
        if group_ids is None:
```
(more idiomatic in Python)

skyportal/handlers/api/spectrum.py
```diff
@@ -174,34 +155,18 @@ def get(self, spectrum_id):
...
-            self.verify_and_commit()
-            return self.success(data=spec_dict)
-        else:
-            return self.error(f"Could not load spectrum with ID {spectrum_id}")
+        spec_dict = spectrum.to_dict()
+        spec_dict["instrument_name"] = spectrum.instrument.name
+        spec_dict["groups"] = spectrum.groups
+        spec_dict["reducers"] = spectrum.reducers
+        spec_dict["observers"] = spectrum.observers
+        spec_dict["owner"] = spectrum.owner
```
acrellin, line 167: I'd suggest consolidating this all into a list of joinedloads that you pass into get_if_accessible_by above

skyportal/handlers/api/spectrum.py
```diff
@@ -408,25 +360,19 @@ def post(self):
...
-            .join(GroupUser)
-            .filter(
-                Group.single_user_group == True,  # noqa
-                GroupUser.user_id == self.associated_user_object.id,
+        single_user_group = self.associated_user_object.single_user_group
+
+        group_ids = json.pop('group_ids', [])
+        if not group_ids:
+            groups = [single_user_group]
+        else:
```
acrellin, line 369: should we also allow "all" and handle accordingly here?

acrellin (changes requested):

skyportal/handlers/api/spectrum.py
scizen9: Thanks, good catch.

skyportal/handlers/api/spectrum.py
scizen9: Applied also to same list in ObjSpectraHandler.

skyportal/handlers/api/spectrum.py
```diff
@@ -174,34 +155,20 @@ def get(self, spectrum_id):
...
-            spec_dict["instrument_name"] = spectrum.instrument.name
-            spec_dict["groups"] = spectrum.groups
-            spec_dict["reducers"] = spectrum.reducers
-            spec_dict["observers"] = spectrum.observers
-            spec_dict["owner"] = spectrum.owner
-            self.verify_and_commit()
-            return self.success(data=spec_dict)
-        else:
-            return self.error(f"Could not load spectrum with ID {spectrum_id}")
+        spec_dict = spectrum.to_dict()
```
acrellin, line 169: You'll need to use `baselayer`'s `recursive_to_dict` here in order to preserve the loaded relationships (see current source or candidate handler for examples)

skyportal/handlers/api/spectrum.py
```diff
@@ -566,17 +515,19 @@ def get(self, obj_id):
...
         if obj is None:
             return self.error('Invalid object ID.')
+        query_options = [
+            joinedload(Spectrum.instrument.name)
+            .joinedload(Spectrum.groups)
+            .joinedload(Spectrum.reducers)
+            .joinedload(Spectrum.observers)
+            .joinedload(Spectrum.owner)
+        ]
         spectra = Obj.get_spectra_readable_by(
```
acrellin, line 525: Needs to be updated to use the new access methods

skyportal/handlers/api/spectrum.py
```diff
@@ -566,17 +515,19 @@ def get(self, obj_id):
...
         )
         return_values = []
         for spec in spectra:
             spec_dict = spec.to_dict()
-            spec_dict["instrument_name"] = spec.instrument.name
-            spec_dict["groups"] = spec.groups
-            spec_dict["reducers"] = spec.reducers
-            spec_dict["observers"] = spec.observers
-            spec_dict["owner"] = spec.owner
             return_values.append(spec_dict)
```
acrellin, line 531: I don't see where you're calling verify_and_commit here (and you'll need to use recursive_to_dict before doing so as noted above)

acrellin (changes requested):

acrellin:
@scizen9 For fetching all spectra associated with an obj, I'd suggest something like 
``` python
spectra = (
    Spectrum.query_records_accessible_by(self.current_user)
    .filter(Spectrum.obj_id == obj_id)
    .all()
)
```

skyportal/handlers/api/spectrum.py
scizen9: Updated with latest commit

skyportal/handlers/api/spectrum.py
scizen9: Now uses new access methods

skyportal/handlers/api/spectrum.py
scizen9: added call to verify_and_commit and now uses recursive_to_dict

acrellin (approved):

skyportal/handlers/api/spectrum.py
```diff
@@ -289,27 +245,21 @@ def delete(self, spectrum_id):
...
         self.push_all(
             action='skyportal/REFRESH_SOURCE',
-            payload={'obj_key': spectrum.obj.internal_key},
+            payload={'obj_key': obj_key},
         )
 
         self.push_all(
             action='skyportal/REFRESH_SOURCE_SPECTRA',
-            payload={'obj_id': spectrum.obj_id},
+            payload={'obj_id': obj_key},
```
kmshin1397, line 262: I think this should still be spectrum.obj_id - the internal_key and obj_id are different things, and the spectra duck still expects the obj_id here

kmshin1397 (changes requested):
One quick thing to change but otherwise looks good!

skyportal/handlers/api/spectrum.py
acrellin: Agreed, good catch

scizen9: made this change, thanks @kmshin1397 .

kmshin1397 (approved):

PR #1851, opened 2021-04-06
Combine count and limit/offset queries for scanning

kmshin1397:
This PR adds the `count()` query for `totalMatches` in the candidates query to the main query so that they are just called once together instead of essentially calling the same query twice. The block to check for out-of-range page numbers is moved to below the main query since now we don't know the total count beforehand. Also added a test for the pagination on the candidates query.


acrellin (approved):
Looks good, thanks!

dannygoldstein (approved):

acrellin:
@kmshin1397 Looks like this broke the sources API

kmshin1397:
@acrellin Tests are now fixed (only big photometry post is failing)

PR #1854, opened 2021-04-07
Fix user/group/streams in demo data

kmshin1397:
There are some discrepancies in users and stream access permissions within the current demo data. 
The following were addressed:
- Group A is marked as needing access to the `ztf_partnership` stream, but its members only have access to the `ztf_public` and `ztf_caltech` streams. Now these users have partnership stream access as well.
- `viewonlyuser` was not really special, as they were put in Group A (which requires partnership access) and given access to the Caltech stream. The user was moved to Group B (which only requires public stream access) and their access to the partnership and Caltech streams removed.

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -46,9 +46,11 @@ groups:
     =id: program_B
 
 groups/=program_A/users:
-  - userID: =viewonlyuser
-    admin: false
   - userID: =fulluser
+    admin: false
+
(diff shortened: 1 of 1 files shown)


acrellin (approved):

stefanv (approved):

dmitryduev (approved):
👍🏻

dmitryduev:
Failure seems unrelated, merging.

PR #1855, opened 2021-04-07
Refactor Invitations API permissions and add tests for invitations

kmshin1397:
In this PR:
- The `InvitationsHandler` methods are refactored to use the new permissions system
- A test SMTP server is set up in the `test_server` microservice to be used for receiving test invitations
  - The `.pem` files are for the TLS for this test SMTP server. 
- API tests for `/api/invitations` including tests covering insufficient permissions scenarios
- Front-end tests for invitations to test bulk invitations, single invitations, editing/deleting invitations from the user management page


skyportal/handlers/api/invitations.py
```diff
@@ -99,7 +113,12 @@ def post(self):
                     "Invalid value provided for `streamIDs`; unable to parse "
                     "all list items to integers."
                 )
-            streams = DBSession().query(Stream).filter(Stream.id.in_(stream_ids)).all()
+
+            streams = (
```
acrellin, line 117: Same suggestion here

skyportal/handlers/api/invitations.py
```diff
@@ -304,33 +321,60 @@ def patch(self, invitation_id):
...
         stream_ids = data.get("streamIDs")
         if group_ids is None and stream_ids is None:
             return self.error(
-                "At least one of either groupIDs or streamIDs are requried."
+                "At least one of either groupIDs or streamIDs are required."
             )
         if group_ids is not None:
             group_ids = [int(gid) for gid in group_ids]
-            groups = Group.query.filter(Group.id.in_(group_ids)).all()
+            groups = (
```
acrellin, line 342: same suggestion here as above

skyportal/handlers/api/invitations.py
```diff
@@ -304,33 +321,60 @@ def patch(self, invitation_id):
...
+                Group.query_records_accessible_by(self.current_user, mode="read")
+                .join(GroupInvitation)
+                .filter(GroupInvitation.invitation_id == invitation.id)
+                .all()
+            )
         if stream_ids is not None:
             stream_ids = [int(sid) for sid in stream_ids]
-            streams = Stream.query.filter(Stream.id.in_(stream_ids)).all()
+            streams = (
```
acrellin, line 361: ditto

acrellin (approved):
Nice, looks good! Just a minor compactification suggestion

skyportal/services/test_server/test_server.py
```diff
@@ -210,9 +235,35 @@ def make_app():
...
+        smtp_port = cfg["test_server.smtp_port"]
         server.listen(port)
 
         refresh_cache_days = cfg["test_server.refresh_cache_days"]
 
+        log(f"Listening for test SMTP requests on port {smtp_port}")
+        # SMTP TLS context
+        context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+        context.load_cert_chain(
+            "skyportal/services/test_server/cert.pem",
```
stefanv, line 247: May want to specify this as a path relative to `__file__`

skyportal/services/test_server/test_server.py
```diff
@@ -2,6 +2,12 @@
 import datetime
```
stefanv, line 2: This service name is too generic; test_smtpd or test_email_server or something?

requirements.txt
```diff
@@ -41,3 +41,4 @@ pyvo==1.1
 lxml==4.6.3
 suds-py3==1.4.3
 vcrpy==4.1.1
+aiosmtpd==1.4.2
```
stefanv, line 44: We have enough dependencies now that it may be worth splitting them out into test and production dependencies.  See, e.g.:

https://github.com/networkx/networkx/blob/main/requirements.txt

stefanv, line 44: Since our only need is to complete an SMTP transaction, would https://docs.python.org/3/library/smtpd.html not suffice?

skyportal/tests/conftest.py
```diff
@@ -61,6 +61,14 @@
     DBSession.add(User(username="test factory"))
     DBSession.commit()
 
+# Also add the test driver user (testuser-cesium-ml-org) if needed so that the driver
```
stefanv, line 64: I didn't quite follow this comment.

requirements.txt
kmshin1397: I started with smtpd but then was led to this package instead when attempting to stand up a server supporting the `starttls` method. My impression was that smtpd was slightly outdated and didn't really support TLS servers out-of-the-box.

skyportal/services/test_server/test_server.py
kmshin1397: This service also listens for test external API calls (the LT follow-up API calls, specifically, for now) on a different port because I was trying to keep all the test servers under one service/file, so I'm not sure about that. Maybe it would better to separate into two distinct services?

skyportal/tests/conftest.py
kmshin1397: With invitations enabled on the test configs, the `driver` was failing to login properly because the authenticator was looking for the `testuser-cesium-ml-org` user or an invitation token when neither exists initially. Usually, the `testuser-cesium-ml-org` user was generated by the authenticator with invitiations disabled but we no longer reach that block (in[ this function](https://github.com/kmshin1397/skyportal/blob/32939eb7f496479994f4ad21e19f920965fddf51/skyportal/onboarding.py#L20))

skyportal/services/test_server/test_server.py
kmshin1397: Sure, we can do that

skyportal/tests/conftest.py
stefanv: I guess I meant to say: let's make this easily readable to others who come and also wonder.

requirements.txt
stefanv: Ah, ok, and we need TLS specifically?

We should avoid new dependencies when possible, but if it's useful and simplifies things dramatically it's ok.

skyportal/services/test_server/test_server.py
stefanv: Oh, right, I see.  Yes, I think the idea is to make one microservice per concept.

requirements.txt
kmshin1397: The TLS functionality is used in our STMP client code before sending emails ([here](https://github.com/skyportal/skyportal/blob/41a5e099056b5bc6a9cb3976ed7bb9aeec0c72b9/skyportal/email_utils.py#L26)). I think the alternative is to add something in the configs to say whether to use TLS or not when sending emails over SMTP, which honestly wouldn't be much work either. It just felt weird to add a conditional to the existing API logic to accommodate the test server instead of having the other way around.

kmshin1397:
Part of fritz-marshal/fritz#163

dannygoldstein (approved):

dmitryduev (approved):
LGMT!

PR #1859, opened 2021-04-08
Fix candidates query ordering

kmshin1397:
The root of the candidates list pagination test failing sometimes was that the default ORDER BY on `Candidates.passed_at` was being done on the wrong thing. Instead of the candidates in the candidates subquery, it was trying to order on the base `Candidates` table and was again introducing the `Candidates, Obj` cross-join we saw yesterday.


acrellin (approved):
Good catch

dannygoldstein:
Does the candidate_subquery.c have the passed_at hybrid method? I thought only the mapped class does ...

kmshin1397:
Yea `passed_at` is a regular column (which is why we order on it instead of `last_detected_at`)

PR #1860, opened 2021-04-08
Fix frontend tests broken in #1858

dmitryduev:
This was caused by the addition of the "Annotate" ACL to some user roles.


acrellin (approved):
Thanks Dima!

acrellin:
Will merge when green

Issue #1861, opened 2021-04-08
Do not allow negative `fluxerr` in photometry POST handler


acrellin:

acrellin:
See https://github.com/skyportal/skyportal/issues/1839

Issue #1839, opened 2021-04-04
error when retrieving a source with null photometry point


AlexGKim:
I posted to Skyportal "None" as photometry for a source, it is accepted without error, but an error occurs when I retrieve the source.

In principle the fact that the photometry is None is interesting to me so I would prefer to post it.  Otherwise I can do a None check before posting and clean out the database of these photometry.

```
:32:17 -0700] "GET /api/sources/DESI39627817075347824 HTTP/1.1" 400 145 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36"
  File "/home/akim/miniconda3/envs/skyportal_env/lib/python3.8/site-packages/sqlalchemy/ext/hybrid.py", line 895, in __get__

    return self.fget(instance)
  File "/home/akim/skyportal/skyportal/models.py", line 776, in peak_detected_at
    return max(detections, key=(lambda x: x[1]))[0] if detections else None
TypeError: '<' not supported between instances of 'float' and 'NoneType'
[08:32:17 basehandler] Error response returned by [/api/sources/DESI39627817075347824]: ['<' not supported between instances of 'float' and 'NoneType']
WARNING:tornado.access:400 GET /api/sources/DESI39627817075347824 (127.0.0.1) 66.20ms
```

acrellin:
@kmshin1397 did you recently refactor peak_detected_at? Wondering if the changes there could be related

acrellin:
I think we just need to filter out photometry whose mag value is `None` here https://github.com/skyportal/skyportal/blob/master/skyportal/models.py#L794

acrellin:
@AlexGKim can you confirm that in `config.yaml` there is a valid number specified for `photometry_detection_threshold_nsigma` (under the `misc` section)?

kmshin1397:
I did refactor it recently, but not sure how it could've led to this error as I kept this `is not None` checks in there at
https://github.com/skyportal/skyportal/blob/38f0c7b93f36e8a93e3934a2d9999c305a6b16ac/skyportal/models.py#L781
https://github.com/skyportal/skyportal/blob/38f0c7b93f36e8a93e3934a2d9999c305a6b16ac/skyportal/models.py#L793

but maybe those aren't working as intended now for whatever reason. I'll look into it.

kmshin1397:
ah 

[...]

PR #1862, opened 2021-04-09
Do not allow negative `fluxerr` values in photometry POST handler

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1861
Closes https://github.com/skyportal/skyportal/issues/1839

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -173,7 +173,10 @@ class PhotometryHandler(BaseHandler):
                 )
             else:
                 kind = 'flux'
-                if data['fluxerr'] < 0:
+                if (
+                    isinstance(data['fluxerr'], (list, tuple))
+                    and not all([fluxerr >= 0 for fluxerr in data['fluxerr']])
+                ) or (isinstance(data['fluxerr'], float) and data['fluxerr'] < 0):
                     raise ValidationError("fluxerr must be a non-negative value.")
         else:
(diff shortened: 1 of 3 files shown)


kmshin1397 (approved):

skyportal/handlers/api/photometry.py
```diff
@@ -173,6 +173,14 @@ def standardize_photometry_data(self):
                 )
             else:
                 kind = 'flux'
+                if (
```
stefanv, line 176: The documentation of `PhotFluxFlexible` says

>     """This is one of two classes used for rendering the
    input data to `PhotometryHandler.post` in redoc. These classes are only
    used for generating documentation and not for validation, serialization,
    or deserialization."""

I guess that's not true!

stefanv: Should these checks be specified in `schema.py` using the [validate keyword](https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.Method)?

dmitryduev (approved):
LGTM!

skyportal/handlers/api/photometry.py
acrellin: Agreed -- updated

stefanv (approved):
Nicely done!

Should the validation exception now be caught in the handler?

acrellin:
@stefanv yes, it is already manually caught

PR #1864, opened 2021-04-09
Fix candidates saved groups

kmshin1397:
I think this is the cleanest and most efficient fix: limits the subquery to just the Source records we actually need
Also removes an unnecessary `.filter(Group.id.in_(user_accessible_group_ids))` since we were already doing `Group.query_accessible_records_by`


acrellin (approved):
Great, thanks!

dmitryduev (approved):
LGTM!

dmitryduev:
Let's wait for CI and merge.

PR #1866, opened 2021-04-09
Fix way too slow thumbnail color migration

kmshin1397:
The current migration attempts to pull every existing thumbnail and determine if they are colored or grayscale. On production-scale databases, this would take way too long to do the migration leading to too much downtime. This PR updates the migration script to set all existing thumbnails to just be marked as colored by default. 

I will do a follow-up PR to add a separate utility script to use the API to do this thumbnail processing in a more efficient/parallelized external process without spending all that time in the migration. Until then, the color inversion preference will just have to only work on thumbnails posted after the migration.


dmitryduev (approved):
:+1:

acrellin (approved):

PR #1872, opened 2021-04-12
Fix saved groups for fetched candidate

kmshin1397:
Same thing with subqueries as in the source handler.

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -416,8 +416,8 @@ class CandidateHandler(BaseHandler):
                 candidate_info["saved_groups"] = (
                     Group.query_records_accessible_by(self.current_user)
                     .join(source_subquery, Group.id == source_subquery.c.group_id)
-                    .filter(Source.obj_id == obj_id)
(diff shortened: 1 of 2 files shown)


acrellin (approved):

skyportal/handlers/api/candidate.py
```diff
@@ -416,8 +416,8 @@ def get(self, obj_id=None):
                 candidate_info["saved_groups"] = (
                     Group.query_records_accessible_by(self.current_user)
                     .join(source_subquery, Group.id == source_subquery.c.group_id)
-                    .filter(Source.obj_id == obj_id)
-                    .filter(Source.active.is_(True))
+                    .filter(source_subquery.c.obj_id == obj_id)
```
acrellin, line 419: Is there any performance difference b/w this and applying the filter to the original Source query before calling subquery, as in the Source handler?

kmshin1397: Ah yeah good point, I do think the way I did it in the Source handler is probably a bit faster since you're joining less rows. I'll change this.

PR #1873, opened 2021-04-12
Speed up bulk photometry posting

kmshin1397:
This should speed up large photometry posts. Based on my benchmarking, the step in the photometry post that takes the most time is in checking for duplicate points. This PR tweaks that part of the logic to use `pandas.DataFrame.itertuples()` instead of `pandas.DataFrame.iterrows()` because `itertuples` is generally faster (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.iterrows.html#pandas.DataFrame.iterrows).

This should keep the big photometry post test from timing out so frequently.


dmitryduev (approved):

kmshin1397:
Hmm guess that didn't help as much as I thought. The test still failed here though it was passing just fine on my test branch. Must also have to do with the volume of tests run alongside it

kmshin1397:
Ok this should now be ready for re-review. On top of the change described above, the insert into the `group_photometry` table (which was taking ~240s for the failing test when run with the entire test suite, compared to ~7s for the insertion into the `photometry` table) is now done with the `COPY` command instead of regular `INSERT`. 

The downside here is that I need to do that through running `psql` in a subprocess so that we can pipe in the rows via STDIN (it's either that or writing out the `GroupPhotometry` rows out to a temporary file every time we post), meaning that part of the photometry POST is handled in a different transaction from the rest. I tried to account for that by manually adding logic to rollback changes from the original handler's transaction if the subprocess command fails. The other direction shouldn't be an issue since the post to `group_photometry` is the very last step of the handler logic, so it shouldn't even reach the spawning of the subprocess before the other transaction is rolled back if errors arise there.

dmitryduev (commented):
Thanks @kmshin1397! I have a few concerns regarding this:
- Connecting to PG from a subprocess. Will it eat up some of the connections that would otherwise be available to tornado processes/threads? My worry is that if something goes wrong, will those be properly recycled or not? @stefan could you think of a bad scenario in this case?
- What about the de-duplication logic - is it affected at all by this PR?

dmitryduev (commented):
Hmm. Github insists that I approve this PR even though I select "Comment" 😆

kmshin1397:
> Thanks @kmshin1397! I have a few concerns regarding this:
> 
> * Connecting to PG from a subprocess. Will it eat up some of the connections that would otherwise be available to tornado processes/threads? My worry is that if something goes wrong, will those be properly recycled or not? @Stefan could you think of a bad scenario in this case?
> * What about the de-duplication logic - is it affected at all by this PR?

- It will count towards the max connections allowed by Postgres (I think maybe Stefan said it was 100?) but if I remember correctly, that number is higher than the maximum allowed by our app connection pool so we have some buffer there. My thought is that this would be like our async handlers causing threads to spawn multiple connections occasionally where it will likely not be an issue in normal operation. As for if something goes wrong, I'm assuming `psql` has some garbage collection in place to release connections if a command goes wrong; it won't affect the Python connection pool at all since it's in the subprocess I think. 
- As far as I can tell, I don't think the de-duplication is affected. That check is all done earlier in the handler before inserting the rows into the `photometry` table. This only affects the final step after that, which is to add the `group_photometry` rows for the just-inserted `photometry` rows. But it might be good to have @dannygoldstein look at this as well if he has cycles since I think he implemented a lot of that.

kmshin1397:
Another approach that I did consider: if we write out the rows we want to `COPY` to a temporary file, we can probably use the same session/connection as the rest of the handler so that we're not doing this in two separate transactions, i.e. https://stackoverflow.com/a/44181653

The downside would be the overhead of writing out the temp file and then deleting it, but maybe it's worth it for the cleaner integration

stefanv (approved):

skyportal/handlers/api/photometry.py
```diff
@@ -40,10 +43,37 @@
 )
 from ...enum_types import ALLOWED_MAGSYSTEMS
 
-
 _, cfg = load_env()
 
 
+def save_data_using_copy(rows, table, columns):
+    # Prepare data
+    output = StringIO()
```
stefanv, line 51: I'm curious why this isn't BytesIO?

kmshin1397: Hmm no conscious decision from me - this is just what the Stack Overflow answer had that I was referencing (https://stackoverflow.com/a/44181653). Digging into it more, my guess is that the original reply I was referencing was written before the pandas `DataFrame.to_csv()` function supported binary file objects so that was why they used `StringIO`. I see that binary file support was added relatively recently - [pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html) say it was added version 1.2.0, and we are on pandas 1.2.4.

stefanv: Interesting! Well, as long as it handles extended characters properly, it doesn't matter.  Not sure what encoding StringIO uses by default.

acrellin:
@stefanv Happy with this one? Want to merge if so?

PR #1877, opened 2021-04-13
Make sure annotation data is a valid object

kmshin1397:


acrellin (approved):

skyportal/handlers/api/annotation.py
```diff
@@ -115,6 +114,12 @@ def post(self):
             return self.error("Input `origin` must begin with alphanumeric/underscore")
 
         annotation_data = data.get("data")
+
+        if type(annotation_data) is not dict:
```
dmitryduev, line 118: Perhaps better do:
```
from typing import Mapping
if not isinstance(annotation_data, Mapping):
...
```
?

acrellin, line 118:
```suggestion
        if isinstance(annotation_data, dict):
```
is more idiomatic, but this is fine too

acrellin: What other data types besides `dict` instances might we expect here?

dmitryduev: Theoretically something like OrderedDict, but likely we will only see dicts. I just really like this generalization - it's especially nice for Sequences.

acrellin:
@dmitryduev good to merge?

PR #1881, opened 2021-04-14
Update BL to include new query optimization

acrellin:


acrellin:
6 new failures (all followup request tests) and it's taking almost an hour for the test suite to run now. Same outcome after restarting the build

kmshin1397:
My guess is something like: it's time for the cache to be refreshed, but LT servers are down so the tests are unable to create new cached responses and just failing.

acrellin:
Now it's just phot_big_post

PR #1883, opened 2021-04-16
LCO Triggering

mcoughlin:


skyportal/services/test_server/test_server.py
```diff
@@ -79,6 +80,18 @@ def lt_request_matcher(r1, r2):
     )
 
 
+def lco_request_matcher(r1, r2):
```
kmshin1397, line 83: You'll probably have to update this at some point to look in the body and distinguish between SPECTRAL, MUSCAT, etc, but for now just the URI/method match should be fine since we're just looking at Sinistro.

kmshin1397 (commented):
Yup, this definitely looks like it's going in the right direction! 

How often do you think we should refresh the LCO cache? I was thinking that it may be easiest to just push the YAML file you recorded locally to the repository somewhere so we don't have to worry about encrypting/adding your credentials to the repo for the CI tests. Downside there would be that we'd have to manually "refresh" the VCR cache file if we wanted to.

skyportal/services/test_server/test_server.py
mcoughlin: For now we just have Sinistro access, so this should be OK. I am happy to push the YAML... but cache/ is not part of the repo currently.

kmshin1397: Ah yeah, it can probably go into like `tests/data` and then you can just modify the `get_cache_file()` in the test server code to return that for LCO requests.

mcoughlin: @kmshin1397 nice. pushed that change. lemme know what you think.

kmshin1397: @mcoughlin Ah you may want to remove your auth token from the YAML file (I'm assuming that's your real one?). Just replace it with some test string like "testtoken" or whatever, and have that be what the test allocation uses as well. You probably want to clean up the git history for it as well with `bfg` or something (docs on doing that: https://docs.github.com/en/github/authenticating-to-github/removing-sensitive-data-from-a-repository)

stefanv: In this case, rebasing, pausing at the offending commit, editing, and continuing should be enough.  See `git rebase -i`.

mcoughlin: I replaced with testtoken and also revoked the original one.

mcoughlin:
@kmshin1397 Looks like the failure is unrelated. Nice job!

mcoughlin:
@kmshin1397 Looks like you did it. Failure is unrelated.

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1045 @@
...
+from baselayer.app.env import load_env
+
+from ..utils import http
+
+env, cfg = load_env()
+
+
+# if "PYTEST_CURRENT_TEST" in os.environ:
+#    requestpath = "https://observe.lco.global/api/requestgroups/validate/"
+# else:
```
kmshin1397, line 15: Can delete the commented out lines

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1045 @@
...
+requestpath = f"{cfg['app.lco_protocol']}://{cfg['app.lco_host']}:{cfg['app.lco_port']}/api/requestgroups/"
+print(requestpath)
```
kmshin1397, line 17: Remove the print() statement

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1045 @@
...
+            DBSession()
+            .query(FollowupRequest)
+            .filter(FollowupRequest.id == request.id)
+            .one()
+        )
+
+        altdata = request.allocation.altdata
+
+        if not altdata:
+            return
```
kmshin1397, line 519: Would it make more sense to raise an error here, with a message along the lines of "No credentials found"? Same thing for the other functions

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1045 @@
...
+        requestgroup = lcoreq.requestgroup
+
+        r = requests.post(
+            requestpath,
+            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
+            json=requestgroup,  # Make sure you use json!
+        )
+        print(r.text)
+        print(requestgroup)
+        print(r.status_code)
```
kmshin1397, line 634: Remove print statements

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1045 @@
...
+        lcoreq = SPECTRALRequest(request)
+        requestgroup = lcoreq.requestgroup
+
+        r = requests.post(
+            requestpath,
+            headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
+            json=requestgroup,  # Make sure you use json!
+        )
+
+        # r.raise_for_status()
```
kmshin1397, line 749: Uncomment?

skyportal/facility_apis/lco.py
```diff
@@ -0,0 +1,1045 @@
...
+            return
+
+        lcoreq = MUSCATRequest(request)
+        requestgroup = lcoreq.requestgroup
+
+        # r = requests.post(
+        #    "https://observe.lco.global/api/requestgroups/",
+        #    headers={"Authorization": f'Token {altdata["API_TOKEN"]}'},
+        #    json=requestgroup,  # Make sure you use json!
+        # )
```
kmshin1397, line 860: Remove old stuff

kmshin1397 (commented):
Nice, some comments - mainly about removing debug stuff.

So we don't have any way of making sure the other instrument calls are correct, right? I see that there are some differences in the instrument configs sent over the API so it would really be good if we could record responses for those as well :( Maybe we should keep all the code in but just only enable the SINISTRO class until we are able to get test responses for the others.

mcoughlin:
@kmshin1397 I did pretty extensive testing of all this pre-VCR (when I still had my full allocation). It should be easy enough to add in the other VCRs once the semester turns, and I hopefully have my allocation back. I would advocate leaving it for now.

kmshin1397 (approved):
Ah okay right, got ya you did have access at some point. Okay, in that case yeah I'm fine with this as long as we go back and update once you have access again 👍

stefanv:
There's new infrastructure being used here (VCR), so we'll need documentation on how to re-record sessions, how to use them, etc. so that others elsewhere in the system can use the same.

mcoughlin:
@kmshin1397 Correct me if I am wrong, but running the tests in full are as simple as deleting the cache file no? For them to pass, the proper API key needs to be put in, and then removed from the cache file that is written.

kmshin1397:
Yes, re-recording is a matter of just deleting the existing cache file and running the tests. For the LCO cache specifically, @stefanv, we decided to do this manually as we don't have a test account with their system that we can just hammer away at so we're just going with a static cache checked into the repo in as test data that we can update if needed. Generally, other mock interfaces like the LT one will use the automatically refreshed cache I implemented. That said, I agree that some documentation of the test server microservice and how it works would be good for me to add - prob in a separate PR.

stefanv:
> Yes, re-recording is a matter of just deleting the existing cache file and running the tests. For the LCO cache specifically, @stefanv, we decided to do this manually as we don't have a test account with their system that we can just hammer away at so we're just going with a static cache checked into the repo in as test data that we can update if needed. Generally, other mock interfaces like the LT one will use the automatically refreshed cache I implemented. That said, I agree that some documentation of the test server microservice and how it works would be good for me to add - prob in a separate PR.

This is useful information.  Let's get it in a place where other developers can see it.  Without that information, no-one would have known how to renew the cache (knowing to delete the API key, e.g., is not evident).

OK to document the test server later (please make an issue and assign), but let's not merge any code that does not have functional requirements documented.

mcoughlin:
@kmshin1397 Would you add more info than I did in my last commit?

kmshin1397:
@mcoughlin Hmm it may be more accurate to say that the static cache should be used in cases where regular refreshing is infeasible, i.e. limited access to the real server. The API credentials are technically not enough of a reason in themselves, I'd say, since you can store them securely in your environment/Github settings to allow CI or local deployments to have access to them without having it be committed in the repo.

mcoughlin:
Good point @kmshin1397 ! Clarified.

mcoughlin:
@stefanv @kmshin1397 could you see if anything else?

skyportal/services/test_server/test_api_server.py
```diff
@@ -14,6 +15,22 @@
...
+def get_cache_file_static():
+    """
+    Helper function to get the path to the VCR cache file for requests
+    that must be updated by hand in cases where regular refreshing is
+    infeasible, i.e. limited access to the real server.
+
+    To update this server recording:
+    1) delete the existing recording
+    2) re-run all tests (with API keys for telescopes in place)
+    3) replace any secret information (such as API keys) with dummy values
```
stefanv, line 27: I guess this step could be automated from the configuration?

skyportal/services/test_server/test_api_server.py
```diff
@@ -79,6 +96,41 @@ def lt_request_matcher(r1, r2):
...
+
+    # A request matches an LCO request if the URI and method matches
+
+    r1_uri = r1.uri.replace(":443", "")
+    r2_uri = r2.uri.replace(":443", "")
+    delete_pattern = r"/api/requestgroups/[0-9]+/cancel/"
+    update_pattern = r"/api/requestgroups/[0-9]+/"
+    submit_pattern = r"/api/requestgroups/"
+
+    # Classify r1
```
stefanv, line 112:
```
submit_patterns = {
    'submit': r"/api/requestgroups/",
    'delete': ...
}

r1_type = None
for (submit_type, pattern) in submit_patterns.items():
    if re.search(pattern, r1_uri):
        r1_type = submit_type
```

And then I'd probably turn that into a function, so you can do:

```
r1_type = submit_type(r1_uri)
r2_type = submit_type(r2_uri)
```

skyportal/services/test_server/test_api_server.py
```diff
@@ -90,21 +142,30 @@ class TestRouteHandler(tornado.web.RequestHandler):
 
     def get(self):
         is_wsdl = self.get_query_argument('wsdl', None)
-        cache = get_cache_file()
+        if self.request.uri == "/api/requestgroups/":
+            cache = get_cache_file_static()
+        else:
+            cache = get_cache_file()
+        print(cache)
```
stefanv, line 149: Leftover print

skyportal/services/test_server/test_api_server.py
```diff
@@ -149,30 +210,54 @@ def get(self):
 
     def post(self):
         is_soap_action = "Soapaction" in self.request.headers
-        cache = get_cache_file()
-
+        if "/api/requestgroups/" in self.request.uri:
+            cache = get_cache_file_static()
+        else:
+            cache = get_cache_file()
+        print(cache)
```
stefanv, line 217: Leftover debug print

stefanv (commented):
I suggested some minor tweaks, and I would recommend removing the extra newlines from the docstrings (before and inside); otherwise looks good to me!

skyportal/services/test_server/test_api_server.py
mcoughlin: Fixed

skyportal/services/test_server/test_api_server.py
mcoughlin: Fixed

skyportal/services/test_server/test_api_server.py
mcoughlin: @kmshin1397 This is your code. Do you agree / want to push it?

skyportal/services/test_server/test_api_server.py
mcoughlin: Maybe @kmshin1397 also has thoughts on this one. I think the main issue is that the programs will change with time, so care must be taken no matter what.

skyportal/services/test_server/test_api_server.py
kmshin1397: Agreed, I just made a PR to your branch for this here: https://github.com/mcoughlin/skyportal/pull/6

skyportal/services/test_server/test_api_server.py
kmshin1397: Yes, this is already automated through the configurations by default. It's just hard to do it in this case since as Michael says the programs will change and we are relying on real time assigned on real accounts instead of test allocations to record these responses.

skyportal/services/test_server/test_api_server.py
mcoughlin: @kmshin1397 Tests looked good too! Pulled it in.

stefanv (approved):

stefanv:
Thanks for the quick response to the reviews.  Looks good to me!

mcoughlin:
@kmshin1397 could you merge if comfortable?

kmshin1397 (approved):

PR #1885, opened 2021-04-16
Fix flaky tests associated with UnitializedDBMessage

kmshin1397:
A couple of tests have failed occasionally if they are run in a certain order such that no non-fixture sources from the other tests have been inserted into the test database, due to the page showing the `UninitializedDBMessage.jsx` component instead of the expected page. 

You can see `test_weather_widget` and `test_add_sources_two_groups` failing in such a manner [here](https://github.com/skyportal/skyportal/pull/1793/checks?check_run_id=2364225192).

This PR fixes that by 
- Remove the `UninitializedDBMessage.jsx` on the Home Page. Now that we have widgets like the `SourceCounts.jsx` on the dashboard, that will appear to display correct information even on an empty database, I don't think we need to overwrite all that to show a message that says "the Sources table is empty". The message was more relevant when the dashboard was just the sources list component but that has now been moved to it's own page and all the other widgets added.
- I kept the `UninitializedDBMessage.jsx` in there on the Sources page, as that message is more directly relevant there. To fix `test_add_sources_two_groups`, I just waited until the test's sources were posted before navigating to the Sources page.
- Also made a small change to the `WeatherWidget.jsx` to not make the API call to fetch weather if it sees that there are no telescopes in the database (to avoid a cryptic "Telescope ID 1 could not be found" error on the dashboard for empty databases).

Here is what the home page looks like now on an uninitialized database, instead of the message:
image

### DIFF
--- a/skyportal/tests/frontend/test_remote.py
+++ b/skyportal/tests/frontend/test_remote.py
@@ -13,4 +13,4 @@ def test_remote(driver):
     DBSession.commit()
     driver.get("/")
     assert 'localhost' in driver.current_url
-    driver.wait_for_xpath('//div[contains(.,"Welcome to SkyPortal.")]')
+    driver.wait_for_xpath('//p[contains(.,"New Sources")]')
--- a/skyportal/tests/frontend/test_source_list.py
+++ b/skyportal/tests/frontend/test_source_list.py
@@ -17,14 +17,6 @@ def test_add_sources_two_groups(
     taxonomy_token_two_groups,
     classification_token_two_groups,
 ):
-
-    driver.get(
-        f"/become_user/{super_admin_user_two_groups.id}"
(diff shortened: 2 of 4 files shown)


stefanv (approved):

stefanv:
Looks good, thanks @kmshin1397.  I don't see the relevant component file deleted?

kmshin1397:
> Looks good, thanks @kmshin1397. I don't see the relevant component file deleted, though?

I didn't actually delete `UninitializedDBMessage.jsx` since I decided to keep it in use in `SourceList.jsx`, just removed references to it from `HomePage.jsx.template`. I can just remove it completely if you think that's better.

stefanv:
Fine with me either way. Thanks!

acrellin (approved):

PR #1897, opened 2021-04-20
Address various issues when adding/editing SEDM requests

kmshin1397:
Closes fritz-marshal/fritz#141

This PR addresses this list of issues with the SEDM follow-up request forms per user feedback:

- [x] "End date" appears before "Start date" in the table, which is confusing. 
- [ ] When adding a new request, the default start date is often in the past.
  - Not sure about this one - the default is already set to `utcnow()`, and was unable to recreate issue locally. My only guess is that the user is in a timezone ahead of UTC, so the default date may appear to be in the past. But since we specify that the field is in UTC I think this is fine as is.
- [x] When editing a new request, the defaults for the fields are not the current values but seem to be the same defaults as when entering new values. This makes it very easily to accidentally change more than one intends to.
  - When editing an existing request, the form is populated with the current values instead of the default ones  
- [x] It is possible to edit someone else's request but not to delete it.
  - This has already been addressed by the permissions refactor
- [x] It is possible to create a group with a start date before the end date.
  - Custom validation has been added to the front-end to check for this

sedm

### DIFF
--- a/static/js/components/EditFollowupRequestDialog.jsx
+++ b/static/js/components/EditFollowupRequestDialog.jsx
@@ -40,6 +40,15 @@ const EditFollowupRequestDialog = ({
     handleClose();
   };
 
+  // Set default form values to current request data
+  const { formSchema } = instrumentFormParams[
+    followupRequest.allocation.instrument.id
+  ];
+
+  Object.keys(formSchema.properties).forEach((key) => {
+    formSchema.properties[key].default = followupRequest.payload[key];
+  });
+
   return (
     <span key={followupRequest.id}>
       <Button
@@ -55,10 +64,7 @@ const EditFollowupRequestDialog = ({
       <Dialog open={open} onClose={handleClose} className={classes.dialog}>
         <DialogContent>
           <Form
-            schema={
-              instrumentFormParams[followupRequest.allocation.instrument.id]
-                .formSchema
-            }
+            schema={formSchema}
             uiSchema={
               instrumentFormParams[followupRequest.allocation.instrument.id]
                 .uiSchema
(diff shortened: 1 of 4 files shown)


static/js/components/EditFollowupRequestDialog.jsx
```diff
@@ -40,6 +40,27 @@ const EditFollowupRequestDialog = ({
...
+
+  Object.keys(formSchema.properties).forEach((key) => {
+    formSchema.properties[key].default = followupRequest.payload[key];
+  });
+
+  const validate = (formData, errors) => {
+    if (
+      formData.start_date &&
+      formData.end_date &&
+      formData.start_date > formData.end_date
```
stefanv, line 56: Just double checking whether this is a valid way to compare these objects.

static/js/components/EditFollowupRequestDialog.jsx
```diff
@@ -40,6 +40,27 @@ const EditFollowupRequestDialog = ({
     handleClose();
   };
 
+  // Set default form values to current request data
+  const { formSchema } = instrumentFormParams[
+    followupRequest.allocation.instrument.id
+  ];
+
+  Object.keys(formSchema.properties).forEach((key) => {
+    formSchema.properties[key].default = followupRequest.payload[key];
```
stefanv, line 49: Why is the default being set to the payload value?  A code comment would be useful here.

stefanv (approved):
Looks good to me!  A few questions, but nothing blocking.

static/js/components/EditFollowupRequestDialog.jsx
kmshin1397: You're right, probably better to make these into Date objects and compare. It should work as is with the string comparison since the dates submitted via the frontend are stored in `YYYY-MM-DD` format, but I guess people could always submit weird stuff via the API.

static/js/components/EditFollowupRequestDialog.jsx
kmshin1397: Done

stefanv:
Thanks, @kmshin1397!

PR #1898, opened 2021-04-20
WIP: Skymap ingestion example

mcoughlin:
Showing what skymap ingestion could look like.


mcoughlin:
@kmshin1397 @dmitryduev @stefanv This is to show off what we talked about yesterday with the skymap ingestion. Still confused about the docs TBH.

skyportal/models.py
```diff
@@ -3515,6 +3521,312 @@ class SourceNotification(Base):
...
+        try:
+            notice = self.gcn_notices[0]
+        except IndexError:
+            return None
+        root = lxml.etree.fromstring(notice.content)
+        elem = root.find(".//Param[@name='LightCurve_URL']")
+        if elem is None:
+            return None
+        else:
+            return elem.attrib.get('value', '').replace('http://', 'https://')
```
profjsb, line 3571: this is probably pretty brittle, especially without schema validation (or is that done upstream)? If there's any chance that this will break, you should do another try/except here.

skyportal/models.py
```diff
@@ -3515,6 +3521,312 @@ class SourceNotification(Base):
...
+        try:
+            notice = self.gcn_notices[0]
+        except IndexError:
+            return None
+        root = lxml.etree.fromstring(notice.content)
+        elem = root.find(".//Param[@name='EventPage']")
+        if elem is None:
+            return None
+        else:
+            return elem.attrib.get('value', '')
```
profjsb, line 3585: same comment as above

skyportal/models.py
```diff
@@ -3515,6 +3521,312 @@ class SourceNotification(Base):
...
+    @property
+    def HasNS(self):
+        """Checking if GW event contains NS."""
+        notice = self.gcn_notices[0]
+        root = lxml.etree.fromstring(notice.content)
+        elem = root.find(".//Param[@name='HasNS']")
+        if elem is None:
+            return None
+        else:
+            return 'HasNS: ' + elem.attrib.get('value', '')
```
profjsb, line 3601: same as above

profjsb:
Side note: very nice to see the use of `pygcn` here. It's been 16 years since I release `pyGCNsock`! https://sourceforge.net/projects/pygcnsock/

skyportal/models.py
mcoughlin: We have used this with success the last while, but no problem being extra careful. Are you thinking something like

```
return elem.attrib.get('value', '').replace('http://', 'https://')
->
try:
    return elem.attrib.get('value', '').replace('http://', 'https://')
else:
    return None
```

mcoughlin:
@kmshin1397 Could you take a look at this one?

skyportal/models.py
profjsb: Yes, something like this would be more robust.

profjsb:
I'd like to see the XML document validated as the first step before processing begins. This can be done with, e.g., `xmlschema` (https://pypi.org/project/xmlschema/).  Eg.

```python
import xmlschema
voevent_schema = xmlschema.XMLSchema('VOEvent-v2.0.xsd') # this can be saved locally
...

#in GCNHandler
if voevent_schema.is_valid(payload):
   root = lxml.etree.fromstring(payload)
else:
   raise 
```

kmshin1397 (commented):
Wow, this is a lot and looks like it could be very useful for some people. Two high-level comments to get us moving forward: 

- I think I would prefer to name the database tables something more specific, like `GcnEvent` and `GcnTag` etc. I think otherwise, these names are very generic and it could 1) get confusing for people new to the codebase to see all these specific properties to something generically-named like `Event`, and 2) cause conflicts in the future, like "tag" seems like a data object that could very easily come up again for some other functionality.
- I'd like to see the functionality in the GCN utils exposed via actual API handlers. I like how the offset queries against Gaia are structured, for example. All the Gaia logic is in the utils folder, and all the interaction with the SkyPortal database is handled in the `SourceOffsetsHandler`, which calls the util functions when needed.

mcoughlin:
@kmshin1397 So the idea is that there would be like a class GCNHandler(BaseHandler) with a push function to replace download / from_cone (which push localizations to the database) and get for  get_dateobs/get_tags/get_skymap?

kmshin1397:
> @kmshin1397 So the idea is that there would be like a class GCNHandler(BaseHandler) with a push function to replace download / from_cone (which push localizations to the database) and get for get_dateobs/get_tags/get_skymap?

In general, yes, although we'd map to REST API nomenclature, like POST/PUT instead of push. I can imagine API calls like `GET /api/gcn/dateobs`, `GET /api/gcn/tags`, `GET /api/gcn/skymap`, `POST /api/gcn/cone?ra=*...` etc. That way, people can use it easily from the outside without doing something like spinning up their on SP and calling the utility functions in new scripts or whatever.

mcoughlin:
@kmshin1397 I pushed something in the spirit of what you envisioned. I think the GET's can be added down the line as we figure out what we need, see what you think.

lpsinger:
Big chunks of this are copied from code that I wrote in growth-too/growth-too-marshal. Would you please add comments to those files to attribute them?

lpsinger:
> Big chunks of this are copied from code that I wrote in growth-too/growth-too-marshal. Would you please add comments to those files to attribute them?

Edit: ...big chunks of this are copied from code that _the GROWTH ToO Marshal team_ wrote...

lpsinger:
> Side note: very nice to see the use of `pygcn` here. It's been 16 years since I release `pyGCNsock`! https://sourceforge.net/projects/pygcnsock/

This is a different package: https://github.com/lpsinger/pygcn

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,157 @@
...
+from ...models import (
+    DBSession,
+    GcnEvent,
+    GcnNotice,
+    GcnTag,
+    Localization,
+)
+from ...utils.gcn import get_dateobs, get_tags, get_skymap, get_contour
+
+gcn.include_notice_types(
```
lpsinger, line 25: This does nothing by itself. gcn.include_notice_types() is a decorator. If you want to separate the list of notice types from the handler function, you need to assign this to something and use the variable later as a decorator.

mcoughlin: Yes good point. This can just be removed as we do not setup the cron ingestion in this PR.

lpsinger: Well, I _hope_ that you are not going to use cron. I _hope_ that you are going to receive push notifications.

mcoughlin: Yes total nonsense. By cron, I somehow mean the process by which we will receive notifications.

lpsinger: If I understand correctly, the issue is simply that you have a generic BaseLayer infrastructure for event handling, and you want want to keep the GCN notifications within that framework? So all you need to do is write a trivial handler function here that pumps the event to your BaseLayer handler. Why not do that in this PR?

mcoughlin: I think any persistent service (like those used to received notifications) will need to go in skyportal/services under supervisor... @kmshin1397? I was just thinking we decouple that service from this PR.

lpsinger: :+1: I understand now.

kmshin1397: Yup, exactly. That sounds good @mcoughlin

skyportal/models.py
```diff
@@ -3120,6 +3126,327 @@ class SourceNotification(Base):
...
+    """Records of ingested GCN notices"""
+
+    id = sa.Column(sa.Integer, primary_key=True, autoincrement=True)
+
+    ivorn = sa.Column(
+        sa.String, primary_key=True, comment='Unique identifier of VOEvent'
+    )
+
+    notice_type = sa.Column(
+        sa.Enum(gcn.NoticeType, native_enum=False),
```
kmshin1397, line 3256: What's the reason for `native_enum=False` here?

skyportal/models.py
```diff
@@ -3120,6 +3126,327 @@ class SourceNotification(Base):
...
+    date = sa.Column(sa.DateTime, nullable=False, comment='UTC message timestamp')
+
+    dateobs = sa.Column(
+        sa.DateTime,
+        sa.ForeignKey(GcnEvent.dateobs),
+        nullable=False,
+        comment='UTC event timestamp',
+    )
+
+    content = deferred(
```
kmshin1397, line 3274: Nice, good idea to use `deferred` for these big columns

skyportal/models.py
```diff
@@ -3120,6 +3126,327 @@ class SourceNotification(Base):
...
+    )
+
+    distnorm = deferred(
+        sa.Column(
+            sa.ARRAY(sa.Float),
+            comment='Multiresolution HEALPix distance normalization array',
+        )
+    )
+
+    contour = deferred(sa.Column(sa.JSON, comment='GeoJSON contours'))
```
kmshin1397, line 3374: Did you have a specific reasoning for using regular JSON here instead of JSONB?

skyportal/models.py
```diff
@@ -3120,6 +3126,327 @@ class SourceNotification(Base):
...
 
 
+class GcnEvent(Base):
+    """Event information, including an event ID, mission, and time of the event."""
+
+    id = sa.Column(sa.Integer, primary_key=True, autoincrement=True)
+
+    dateobs = sa.Column(sa.DateTime, doc='Event time', unique=True, nullable=False)
+
+    gcn_notices = relationship(lambda: GcnNotice, order_by=lambda: GcnNotice.date)
```
kmshin1397, line 3136: Mainly for consistency with the rest of SkyPortal, I would prefer to see the lambda functions in this class turned into normal arguments, i.e. `relationship('GcnNotice', order_by=GcnNotice.date)`. To have that work you probably have to define the `GcnEvent` class after the other data tables it depends on but that's perfectly fine.

skyportal/models.py
```diff
@@ -3120,6 +3126,327 @@ class SourceNotification(Base):
     level = sa.Column(sa.String(), nullable=False)
 
 
+class GcnEvent(Base):
+    """Event information, including an event ID, mission, and time of the event."""
+
+    id = sa.Column(sa.Integer, primary_key=True, autoincrement=True)
+
+    dateobs = sa.Column(sa.DateTime, doc='Event time', unique=True, nullable=False)
```
kmshin1397, line 3134: Correct me if I'm wrong, but the `dateobs` is basically the identifier for an event and you don't really use the autoincrementing ID anywhere right? If so, it might make more sense to just make `dateobs` the primary key since that's what you use in all the relationship and api logic anyways.

kmshin1397 (commented):
Hi @mcoughlin sorry it took me so long to look at this again. At first glance this already seems much improved and we're definitely on the right track. This isn't a full review of all the changes yet, but here are some questions to move us along with regards to the database schema changes first.

On that topic, you should also try to generate a database migration script for these schema changes and include that in this PR as well. The section on migrations in the docs can be found here: https://skyportal.io/docs/migrations.html

skyportal/models.py
mcoughlin: I am a little confused how to do this sa.ForeignKey(GcnEvent.dateobs) if we have now reordered these? putting it as a string leads to errors like: Ensure that referencing columns are associated with a ForeignKey or ForeignKeyConstraint, or specify a 'primaryjoin' expression.

skyportal/models.py
mcoughlin: Nope converting to JSONB seems good.

skyportal/models.py
mcoughlin: +1!

skyportal/models.py
mcoughlin: yes good point.

mcoughlin: However, without an id column, I get: {"status": "error", "message": "(psycopg2.errors.NotNullViolation) null value in column \"id\" of relation \"gcnevents\" violates not-null constraint\nDETAIL:  Failing row contains (null, 2021-06-11 02:25:22.42374, 2021-06-11 02:25:22.42374, 2018-01-16 00:36:53).\n\n[SQL: INSERT INTO gcnevents (created_at, modified, dateobs) VALUES (timezone(%(timezone_1)s, CURRENT_TIMESTAMP), timezone(%(timezone_1)s, CURRENT_TIMESTAMP), %(dateobs)s)]\n[parameters: {'timezone_1': 'UTC', 'dateobs': datetime.datetime(2018, 1, 16, 0, 36, 53)}]\n(Background on this error at: http://sqlalche.me/e/13/gkpj)", "data": {}, "version": "0.9.dev0+git20210529.1bf42c0"}

skyportal/models.py
mcoughlin: Looks like strings it is.

PR #1910, opened 2021-04-24
Fix mjdNow computation in VegaPlot.jsx

dmitryduev:
This will fix occasional apparent data points from the future (has been reported a few times + I think I saw that myself).


guynir42 (approved):
I can see how the new code is better. But I am curious why the old code was implemented this way in the first place?

kmshin1397 (approved):

kmshin1397:
> I can see how the new code is better. But I am curious why the old code was implemented this way in the first place?

Ah yeah I think that's on me. I think I just googled a formula for the conversion and copied it over - no particular reason.

PR #1912, opened 2021-04-27
Fix bug in photometry POST handler altdata broadcasting

acrellin:


dmitryduev (approved):
:double-face-palm:... Thanks so much for figuring this out @acrellin!!!

kmshin1397 (approved):

PR #1920, opened 2021-04-29
Added SDSS filters to NOT/ALFOSC

steveschulze:
Hi,

I added the SDSS filters to the NOT/ALFOSC instrument.

Cheers,

Steve


acrellin:
These will still need to be added manually to the production instance, which I will do shortly.

PR #1943, opened 2021-04-30
Fix join in sources filtering by classification

kmshin1397:
There was a bad join when filtering the sources page by classification (a cross-join between all taxonomy and all objs). The current version actually joins in a whole new taxonomy subquery and pairs them with the classifications subquery to look for the filtered on value.

This fixes the subqueries involved in that to get rid of the cross join and actually filter on proper `(taxonomy.name, classification.classification)` pairs from the same subquery.


acrellin (approved):
Nice!

acrellin:
Failures unrelated

Issue #1380, opened 2020-11-18
GET /api/candidates/{obj_id} returns 400 on nonexistent objects


jvansanten:
In 9e4dd70 as deployed on fritz.science, GET /api/candidates/{obj_id} returns 400 Bad Request,

```
curl -i -H 'Authorization: token XXX' 'https://fritz.science/api/candidates/flerpyherp'
HTTP/2 400
server: nginx/1.18.0 (Ubuntu)
date: Wed, 18 Nov 2020 14:46:14 GMT
content-type: application/json; charset=UTF-8
content-length: 99
via: 1.1 google
alt-svc: clear

{"status": "error", "message": "Invalid ID", "data": {}, "version": "0.9.dev0+git20201117.9e4dd70"}
```

while HEAD /api/candidates/{obj_id} returns 404 Not Found as expected:

```
curl --head -i -H 'Authorization: token XXX' 'https://fritz.science/api/candidates/flerpyherp'
HTTP/2 404
server: nginx/1.18.0 (Ubuntu)
date: Wed, 18 Nov 2020 14:46:50 GMT
content-type: text/html; charset=UTF-8
content-length: 0
via: 1.1 google
alt-svc: clear

```

(This example originally used ZTF18aayxiam, which has been posted in the mean time, but any nonexistent ID will do) 

These should probably both return 404 to indicate that the object is missing.

acrellin:
This is now returning 404

Issue #1743, opened 2021-03-05
Return 404 (resource not found) instead of 400 wherever specified resource does not exist


acrellin:
`BaseHandler.error` defaults to `status=400`. We just need to specify `status=404` explicitly.

profjsb:
Fixed by #2551

PR #1965, opened 2021-05-04
Ensure that raised AccessError yields 404 response code

acrellin:
Depends on https://github.com/cesium-ml/baselayer/pull/226
Partially addresses https://github.com/skyportal/skyportal/issues/1743
Closes https://github.com/skyportal/skyportal/issues/1380

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -626,7 +626,9 @@ class SourceHandler(BaseHandler):
                     obj_id, self.current_user, options=[joinedload(Obj.thumbnails)]
                 )
             else:
-                s = Obj.get_if_accessible_by(obj_id, self.current_user, raise_if_none=True)
+                s = Obj.get_if_accessible_by(
+                    obj_id, self.current_user, raise_if_none=True
+                )
             source_info = s.to_dict()
             source_info["followup_requests"] = (
                 FollowupRequest.query_records_accessible_by(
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/source.py
```diff
@@ -1489,9 +1489,7 @@ async def get(self, obj_id):
               application/json:
                 schema: Error
         """
-        source = Obj.get_if_accessible_by(obj_id, self.current_user)
-        if source is None:
-            return self.error('Source not found', status=404)
+        source = Obj.get_if_accessible_by(obj_id, self.current_user, raise_if_none=True)
```
stefanv, line 1492: Why is this better than raising the 404 explicitly?

stefanv (commented):
Looks OK; I'm not sure I 100% understand why this is necessary.  I left a comment on the baselayer PR which uses `error` instead of adding special logic inside of it.  There, I think you can explicitly examine the Exception type to see if it is an AccessError, and then call `error` appropriately.

skyportal/handlers/api/source.py
acrellin: checking manually is fine, but we're largely moving over to the more compact pattern here. this is changed to easily test via the new source API test

stefanv: OK, automating that makes sense.  Thanks for explaining to me.

acrellin:
@stefanv updated, ready for re-review

stefanv (approved):

acrellin:
Depends on https://github.com/cesium-ml/baselayer/pull/230

acrellin:
Now I need to update a bunch of tests ;)

stefanv (approved):

acrellin:
Closing as stalled; will address in a future PR.

PR #1966, opened 2021-05-04
Add plot height to spectroscopy plot if legend overfills plot area

scizen9:
This seems to be a very simple fix, however, we may want to force a vertical scroll bar.
Here is a screenshot for 29 spectra:
image

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -1361,6 +1361,10 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
         + legend_row_height * int(len(split) / legend_items_per_row)
         + 30  # 30 is the height of the toolbar
     )
+    # check if our legend should extend the plot height
+    plot_height_with_legend = len(spectra) * 25 + 30
+    if plot_height_with_legend > plot_height:
+        plot_height = plot_height_with_legend
 
     plot = figure(
         frame_width=frame_width,


acrellin:
Nice! Can I suggest to add a bit more height? The buttons slightly overlap the text input fields at the bottom.

scizen9:
> Nice! Can I suggest to add a bit more height? The buttons slightly overlap the text input fields at the bottom.

I've noticed this with every plot change I've worked on and when it gets deployed they never overlap.  Not exactly sure why.

scizen9:
Updated screenshot from last commit.

image

kmshin1397 (approved):
Excellent, thanks for taking care of this @scizen9

skyportal/plot.py
```diff
@@ -1336,6 +1336,9 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
 
     # These values are equivalent from the photometry plot values
```
dmitryduev, line 1337: Not anymore?

dmitryduev (commented):
@acrellin oh noes, I was in the middle of a review :(

skyportal/plot.py
acrellin: Oh woops! Wanna push your suggested change up to master?

Issue #1913, opened 2021-04-27
Add links to Advanced Usage section in the docs


dmitryduev:
Apparently https://skyportal.io/docs/advanced_usage.html is not referenced in the ToCs, or am I missing something?

guynir42:
closed by #1968

PR #1968, opened 2021-05-06
added advanced_usage to TOC

guynir42:
Is there anything else we need to add except this line? @dmitryduev


guynir42:
addresses #1913

dmitryduev (approved):
That should be it, thanks @guynir42. @stefanv am I right? :)

PR #1969, opened 2021-05-06
Add Obj.alias column and start using Obj.origin

guynir42:


stefanv:
I'm not qualified to review this change. @dmitryduev?

static/js/components/SourceTable.jsx
```diff
@@ -448,12 +448,12 @@ const SourceTable = ({
     return <FavoritesButton sourceID={objid} />;
   };
 
-  const renderAlias = (dataIndex) => {
-    const { id: objid, alias } = sources[dataIndex];
+  const renderOrigin = (dataIndex) => {
```
kmshin1397, line 451: @dmitryduev Is it fine to get rid of the Alias column here? I don't remember what it was for exactly, but it doesn't look like any sources are really using it these days and I think it's just a holdover from old iterations.

dmitryduev: The same physical object usually goes by multiple names in different surveys/services, that's what Alias is meant for - we just didn't get to implementing it fully. I would keep it and add origin like @guynir42 is suggesting, but otherwise it looks good to me.

guynir42:
I agree that we should be using alias. 
I added this column to the Obj table. 
I got sorting and filtering on both alias and origin to work. 
Added some tests for this, too. 

We need to add a way to define which columns are visible in the config file.

skyportal/handlers/api/source.py
```diff
@@ -861,6 +875,10 @@ class Validator(Schema):
                 func.lower(Obj.altdata['simbad']['class'].astext)
                 == simbad_class.lower()
             )
+        if alias:
```
acrellin, line 878:
```suggestion
        if alias is not None:
```

skyportal/handlers/api/source.py
```diff
@@ -861,6 +875,10 @@ class Validator(Schema):
...
+            obj_query = obj_query.filter(Obj.alias.contains(alias.strip()))
+        if origin:
```
acrellin, line 880:
```suggestion
        if origin is not None:
```

skyportal/handlers/api/source.py
```diff
@@ -964,6 +982,18 @@ class Validator(Schema):
         if sort_by is not None:
             if sort_by == "id":
                 order_by = [Obj.id] if sort_order == "asc" else [Obj.id.desc()]
+            elif sort_by == "Alias":
```
acrellin, line 985: why capitalized? also origin below

guynir42:
@acrellin can we merge this as well?

acrellin (approved):
@dmitryduev did you want to review before this goes in?

acrellin:
@guynir42 I took the liberty to update the PR title & description

skyportal/handlers/api/source.py
```diff
@@ -218,6 +218,18 @@ def get(self, obj_id=None):
             schema:
               type: string
             description: Simbad class to filter on
+          - in: query
+            name: alias
+            nullable: true
+            schema:
+              type: string
```
dmitryduev, line 225: I think this should be an array of strings. There could be multiple names for the same object. Could you please update the logic where necessary?

guynir42:
OK so I've made alias an array of strings so you can have multiple aliases. 
I've also added them to the source page. 
Screenshot from 2021-05-13 20-56-16
Screenshot from 2021-05-13 20-57-19
Screenshot from 2021-05-13 20-57-47

dmitryduev (approved):
Nice work!

guynir42:
@acrellin @dmitryduev who will be the one to merge this...?

PR #1972, opened 2021-05-10
Update ZTF triggering

mcoughlin:
ZTF API


test_config.yaml
```diff
@@ -67,10 +67,14 @@ test_server:
...
 
 app:
   lt_host: localhost
   lt_port: 64502
   lco_protocol: http
   lco_host: localhost
   lco_port: 64502
+  ztf_protocol: http
+  ztf_host: localhost
+  ztf_port: 64502
```
kmshin1397, line 80:
```suggestion
  ztf:
    protocol: http
    host: localhost
    port: 64502
```
I think this is why it's not triggering the test server

mcoughlin:
@dmitryduev can you take a look as well?

mcoughlin:
Relevant tests pass.

skyportal/facility_apis/ztf.py
```diff
@@ -0,0 +1,230 @@
...
+
+from ..utils import http
+
+env, cfg = load_env()
+
+
+if not cfg['app.ztf.port'] is None:
+    ZTF_URL = f"{cfg['app.ztf.protocol']}://{cfg['app.ztf.host']}:{cfg['app.ztf.port']}"
+else:
+    ZTF_URL = f"{cfg['app.ztf.protocol']}://{cfg['app.ztf.host']}"
```
dmitryduev, line 17: A bit more idiomatic, otherwise sounds like I not can haz cheeseburger? 
```suggestion
if cfg['app.ztf.port'] is None:
    ZTF_URL = f"{cfg['app.ztf.protocol']}://{cfg['app.ztf.host']}"
else:
    ZTF_URL = f"{cfg['app.ztf.protocol']}://{cfg['app.ztf.host']}:{cfg['app.ztf.port']}"
```

skyportal/facility_apis/ztf.py
```diff
@@ -0,0 +1,230 @@
...
+        payload: json
+            payload for requests.
+        """
+
+        start_mjd = Time(request.payload["start_date"], format='iso').mjd
+        end_mjd = Time(request.payload["end_date"], format='iso').mjd
+
+        bands = {'g': 1, 'r': 2, 'i': 3, 'z': 4, 'J': 5}
+        json_data = {
+            'queue_name': "ToO_" + request.payload["queue_name"],
```
dmitryduev, line 44: Could this ever change? If so, maybe move to config?

dmitryduev (commented):
Nice, @mcoughlin - only a couple of comments.

skyportal/facility_apis/ztf.py
```diff
@@ -0,0 +1,230 @@
...
+            'validity_window_mjd': [start_mjd, end_mjd],
+        }
+
+        if request.payload["program_id"] == "Partnership":
+            program_id = 2
+        elif request.payload["program_id"] == "Caltech":
+            program_id = 3
```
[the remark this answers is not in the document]
mcoughlin: agreed.

skyportal/facility_apis/ztf.py
```diff
@@ -0,0 +1,230 @@
...
+
+                target = {
+                    'request_id': cnt,
+                    'program_id': program_id,
+                    'field_id': field_id,
+                    'ra': request.obj.ra,
+                    'dec': request.obj.dec,
+                    'filter_id': filter_id,
+                    'exposure_time': float(request.payload["exposure_time"]),
+                    'program_pi': 'Kulkarni' + '/' + request.requester.username,
```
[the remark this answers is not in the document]
mcoughlin: Well he don't complain about the papers that come out ;)

mcoughlin:
@kmshin1397 please take another look when you have time. no rush.

skyportal/services/test_server/test_api_server.py
```diff
@@ -124,6 +124,33 @@ def submit_type(uri):
...
+            "put": r"/api/triggers/ztf/",
+        }
+        for (submit_type, pattern) in patterns.items():
+            if re.search(pattern, uri) is not None:
+                return submit_type
+
+        return None
+
+    r1_type = submit_type(r1_uri)
+    r2_type = submit_type(r2_uri)
```
kmshin1397, line 149: Since the routes for the DELETE and PUT are the same, I don't think we need to do all this submit type/pattern stuff. I'd refactor this to just be like a `is_ztf_request` function which just searches for the `/api/triggers/ztf` pattern and returns the boolean without the `submit_type` stuff. That, plus the `r1.method == r2.method` check should be enough.

kmshin1397 (commented):
Just one comment, and otherwise this looks good.

skyportal/services/test_server/test_api_server.py
mcoughlin: @kmshin1397 great point, see update to see if i did what you were imagining.

kmshin1397 (approved):

PR #1973, opened 2021-05-10
Optimize sources query

kmshin1397:
This should speed up the sources page query a bit. The old definition basically had `Source.read = Source.read` so there was an unnecessary self-join in there.


acrellin (approved):

kmshin1397:
Migration workflow failure is addressed in #1974 and is unrelated to this one

stefanv (approved):
Why the brackets around source.read?

Thanks for pinning this down!

PR #1977, opened 2021-05-10
Fix recent sources query

kmshin1397:
The new access control rule somehow made the order by map to `User.created_at` instead of the `Source.created_at`


acrellin (approved):
Thanks!

PR #1978, opened 2021-05-11
Cache scanning page query results & use cache for pagination results

acrellin:
Closes https://github.com/skyportal/skyportal/issues/1766

Screenshot from 2021-05-17 12-59-30


acrellin:
This is working well in local testing.

acrellin:
Hmm, eslint is complaining on CI, but is happy with the code locally

stefanv:
https://github.com/skyportal/skyportal/blob/master/skyportal/utils/cache.py

Example usage: https://github.com/skyportal/skyportal/blob/master/skyportal/utils/offset.py#L324 and below

acrellin:
@stefanv I wonder if expiring cache files based on the number of files vs the time elapsed since they were created is the right approach here. I suppose as long as we keep the limit high enough (100? 200?) we shouldn't run into any problems there. I had been planning on using a time-based cache vacuuming cron job. Thoughts there?

skyportal/handlers/api/candidate.py
```diff
@@ -1057,19 +1063,29 @@ def grab_query_results(
...
+            np.save(f"cache/candidate_queries/{query_id}.npy", ordered_ids.all())
+            results = (
+                ordered_ids.limit(n_items_per_page)
+                .offset((page - 1) * n_items_per_page)
+                .all()
+            )
+        else:
+            ids = np.load(f"cache/candidate_queries/{query_id}.npy")
+            results = ids[
+                (page - 1) * n_items_per_page : (page) * n_items_per_page  # noqa
```
acrellin, line 1079: black and flake8 disagree about whether there should be a space before the `:`

skyportal/handlers/api/candidate.py
```diff
@@ -439,6 +442,7 @@ def get(self, obj_id=None):
 
         page_number = self.get_query_argument("pageNumber", None) or 1
         n_per_page = self.get_query_argument("numPerPage", None) or 25
+        query_id = self.get_query_argument("queryID", None)
```
acrellin, line 445: I've intentionally not documented this, as it is only for front-end usage and could likely confuse users looking through the API docs.

stefanv:
Yes, I suppose you definitely want time-based expiration here.  We can either add time-based expiration to the existing caching mechanism, or you can store a timestamp with the data in the cache and continue to reload it if it is too old.  I'm leaning towards the first option, if it looks straightforward.

A cron job would also do the job, but it would be nice to have some generic caching infrastructure to use across the project.

stefanv:
(Side question: what *should* the timeout be on these objects?  There's a balance between returning old results and returning consistent lists :grimacing: )

acrellin:
@stefanv I've already implemented a simple time-based caching system here and a very simple cron job that cleans up old cache (defaults to 2 hrs, configurable in the config - see diff for details)

stefanv:
Yes, I saw.  Would you prefer not to fold this functionality into the caching system?

acrellin:
Yes, they seem different enough that it'd add undue complexity to extend that to work for this use case.

stefanv:
Here's how that could look, potentially: https://github.com/skyportal/skyportal/pull/1985

As mentioned, the cache invalidates on insert only, but we can change that (there's a similar issue with the implementation given here  that items may live a bit longer than anticipated due to timing of the cron job).

stefanv:
BTW, this is how to fix the conflict between flake8 and black: https://github.com/skyportal/skyportal/pull/1985/files#diff-6951dbb399883798a226c1fb496fdb4183b1ab48865e75eddecf6ceb6cf46442R16

acrellin:
@stefanv Yeah, that part is doable, but there's also the complication that we'll have to convert lists to/from bytes to use that, which is just another complication I'd rather avoid. The proposal here is nice and simple.

stefanv:
Sorry, I'm not following.  Files are always written as bytes, and the key to the cache is the ID.

acrellin:
@stefanv Your caching tool requires a `bytes`-type argument: https://github.com/skyportal/skyportal/blob/master/skyportal/utils/cache.py#L59
I am here caching Python lists. Converting lists to/from bytes is what I'm saying I'd like to avoid if possible (which it is -- see above implementation).

stefanv:
Yes, that's trivial:

```
b = io.BytesIO()
np.save(b, x)
```

Even simpler if you just want to store a list of numbers:

```
x = np.array([1, 10, 255, 1000], dtype=int)
x.tobytes()
```

(and `np.frombuffer(bytes, dtype=int)` to read).

acrellin:
@stefanv `np.frombuffer` doesn't play so nicely with `BytesIO` objects, it seems. Can you post a working example using a mixed-type (list of tuples, each tuple containing strings and numbers) array?

stefanv:
> @stefanv `np.frombuffer` doesn't play so nicely with `BytesIO` objects, it seems. Can you post a working example using a mixed-type (list of tuples, each tuple containing strings and numbers) array? e.g. `[("14gqr", 333), ("16fil", 333)]`

Right, you have to choose one or the other.  You can use the `tobytes` and `frombuffer` combo, or you can use `save` (to a BytesIO object) and `load` (straight from cache file).

Here's a working example:

```
import numpy as np
import io

# Prepare bytes
x = np.array([("14gqr", 333), ("16fil", 333)])
b = io.BytesIO()
np.save(b, x); cached_bytes = b.getvalue()

# This happens inside the caching mechanism
fn  = '/tmp/cachefile.bytes'
with open(fn, 'wb') as f:
    f.write(cached_bytes)

# Load back the array from disk
np.load(fn)
```

acrellin:
Thanks, @stefanv . So, good with sticking with what's proposed here?

stefanv:
Why not use the bytes with the caching infrastructure + the applied patch?

I am concerned about this approach because there is no easy way to test it, whereas the caching infrastructure is well covered.

acrellin:
@stefanv it's not clear to me what you're proposing. Your caching infrastructure doesn't work well with arbitrary lists.

acrellin:
I'm fine with using it in principle, but getting it to work is getting increasingly complex. What's proposed here is simple, and is easily testable with the right test config parameters.

stefanv:
I don't understand what complexity you refer to.  The caching infrastructure just takes (key, bytes) and ensures that they land up in a file, and managed the file removal based on age / size of cache.  You have to control what bytes go in there, but as shown in the code example above obtaining those bytes are trivial:

```
b = io.BytesIO()
np.save(b, x)
bytes_for_cache_use = b.getvalue()
```

So, not sure what the increasing complexity is.

The approach you have here will work, but is not easily testable.

acrellin:
@stefanv to the contrary, it is easily testable, with the right config parameters. How do you load those bytes that the caching infrastructure returns back into a usable Python object?

stefanv:
> @stefanv to the contrary, it is easily testable, with the right config parameters.

How would that work?

> How do you load those bytes that the caching infrastructure returns back into a usable Python object?

You get the filename back, so: `np.load(filename)`

acrellin:
Simple: shorten the expiration and cron interval times, just like what you do in the cache test in your PR

acrellin:
okay @stefanv, I'll port this over to use what you've just implemented in the other PR

skyportal/handlers/api/candidate.py
kmshin1397: Maybe put a quick inline comment here saying that's why it's not in the API docs? In case we forget in the future.

kmshin1397 (approved):
Looks good! I'm happy for it to go in once ported over Stefan's cache util.

kmshin1397:
Ah wait, I guess also add that tooltip like we discussed on the scanning page to just make sure users are aware that results are cached until a new search is made. I agree that this new behavior is intuitive but I imagine some people may have figured out how the old behavior works and will be expecting it now.

acrellin:
Tooltip added: 
Screenshot from 2021-05-17 12-59-30

acrellin:
I'm actually not sure I do want to use the other caching infrastructure. There are too many little complications that make it feel too cumbersome to use in this case. Each one on its own may be "trivial", but on the whole the dev experience is overly cumbersome (needing to creating a BytesIO object and "save" the np array there, the setting/getting asymmetry, etc.)

acrellin:
@dmitryduev can you review when you get a minute?

stefanv:
> I'm actually not sure I do want to use the other caching infrastructure. There are too many little complications that make it feel too cumbersome to use in this case. Each one on its own may be "trivial", but on the whole the dev experience is overly cumbersome (needing to creating a BytesIO object and "save" the np array there, the setting/getting asymmetry, etc.). This implementation is much simpler.

Previously there didn't seem to be many difficulties that weren't easily addressed.  Once the testing has been implemented for this solution, I'm not sure it is so much simpler.  Can we discuss specific difficulties to see if we can sort them out?

dmitryduev (commented):
All looks good to me - nice and simple logic. I would only add a backend test.

skyportal/handlers/api/candidate.py
```diff
@@ -1057,19 +1067,48 @@ def grab_query_results(
...
-        # Now bring in the full Obj info for the candidates
-        results = (
-            ordered_ids.limit(n_items_per_page)
-            .offset((page - 1) * n_items_per_page)
-            .all()
-        )
+        if use_cache:
+            if query_id is None:
+                query_id = str(uuid.uuid4())
+                if not os.path.exists("cache/candidate_queries"):
```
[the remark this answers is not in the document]
acrellin: Done

test_config.yaml
```diff
@@ -13,6 +13,7 @@ server:
 
 misc:
   photometry_detection_threshold_nsigma: 3.0
+  hours_to_keep_candidate_query_cache: 0.000555556 # 2 seconds
```
stefanv, line 16: How about making this granularity minutes?

skyportal/tests/api/test_candidates.py
```diff
@@ -1456,6 +1459,39 @@ def test_candidate_list_pagination(
...
+    assert data["data"]["queryID"] == query_id
+
+    # Wait until cache is expired
+    time.sleep(3)
+
+    # Run cron job that deletes old query cache files
+    root_project_dir = os.path.dirname(
+        os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+    )
+    subprocess.call(
```
stefanv, line 1481: This is not the way the system will be deployed in practice. To test this way, you need to install the cron job as usual, and wait for the timeout (not guaranteed to be 2 seconds, so probably wait 4) to see if the file disappeared.

I don't know what the planned timeout is for the current system, but it's worth noting that the cache items may be retained for up to 2x the number specified in the configuration.

acrellin: This is testing the script itself, just not the cron service, which has a minimum interval of one minute (longer than I think we want to wait in our tests)

dmitryduev (approved):

stefanv (changes requested):
Marking this PR as in discussion.

stefanv (commented):
Nice, Ari, thank you for bearing with me through this!

acrellin:
@stefanv good suggestions, both implemented - ready for re-review

stefanv:
Looks great, Ari!  Last question, is `use_cache` necessary?

acrellin:
> Looks great, Ari! Last question, is `use_cache` necessary?

@stefanv Yes, because the source handler also uses this function, but we don't want to cache those queries, thus the `use_cache` param

stefanv:
@acrellin If you don't provide a cache ID, I guess it caches but you never get the same results back, which is OK?

acrellin:
Can we leave it optional like this? That way it's specifiable, and we don't
need to waste the resources of creating caches we know we'll never want to
utilize.

stefanv:
> Can we leave it optional like this? That way it's specifiable, and we don't need to waste the resources of creating caches we know we'll never want to utilize.

Sure, that's fine!

PR #1980, opened 2021-05-12
Re-skin desktop source page to use MaterialUI grid

dmitryduev:
We are currently using a custom grid on the desktop Source page whose width maxes out at 900 px such that there's a lot of empty space on the right, especially if page view is not at 100% (or on larger screens). I replaced it with MUI's standard grid. Also, there's a very annoying effect when clicking on/selecting dropdown's - MUI adds padding to `<body>`. I was able to disable that effect it in many places, but the trick I found does not work everywhere - we'll have to look into that in the future.

Before:
source-page-1

After:
source-page-2


static/js/components/CommentList.jsx
```diff
@@ -60,7 +60,7 @@ const CommentList = () => {
     );
 
   return (
-    <div className={styles.comments}>
```
dmitryduev, line 63: This class does not exist.

static/js/components/ObjPageAnnotations.jsx
```diff
@@ -18,13 +18,10 @@ import { getAnnotationValueString } from "./ScanningPageCandidateAnnotations";
...
 dayjs.extend(utc);
 
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles(() => ({
   container: {
     width: "100%",
     margin: "auto",
-    [theme.breakpoints.up("md")]: {
-      width: "27.5rem",
-    },
```
dmitryduev, line 27: No need to do this anymore, the grid takes care of auto-scaling.

acrellin (approved):
LGTM!

acrellin:
`prettier` had a new release upstream -- bump your local version

kmshin1397 (approved):
Awesome, thanks for taking care of this. It's definitely long overdue :)

dmitryduev:
@kmshin1397 @acrellin `test_submit_new_followup_request_SEDM` failed - do you think it's relevant?

PR #1981, opened 2021-05-12
Source page spinner

### LINKED ISSUE #1911: Display spinner in source tables while data is loading
...otherwise at first it looks like there's nothing there; the lag seems too long to not bother.

guynir42:
We want to have a spinner running when there is a filtering/sorting request, so it doesn't look like nothing changes when those changes are applied. 

Closes #1911

### DIFF
--- a/skyportal/tests/frontend/test_source_list.py
+++ b/skyportal/tests/frontend/test_source_list.py
@@ -58,12 +58,6 @@ def test_add_sources_two_groups(
     # find the name of the newly added source
     driver.wait_for_xpath(f"//a[contains(@href, '/source/{obj_id}')]")
 
-    # Close filter form
-    driver.click_xpath(
-        "//button[contains(@class, 'MUIDataTableToolbar-filterCloseIcon')]",
-        wait_clickable=False,
-    )
-
     # find the date it was saved
     saved_at_element = driver.wait_for_xpath(
         f"//*[text()[contains(., '{t1.strftime('%Y-%m-%dT%H:%M')}')]]"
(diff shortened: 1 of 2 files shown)


guynir42:
@acrellin I need some help with this. 
I found a good place to put `setQueryInProgress(true)` but I don't know where to put the `false` part. 
Right now the spinner appears when the filter is selected but it doesn't disappear, even when the `FETCH_SOURCES_OK` comes back.

acrellin:
@guynir42 I'd use a `useEffect` hook which sets the value to `false` when `sources` state is updated and is non-null/not-empty

guynir42:
> @guynir42 I'd use a `useEffect` hook which sets the value to `false` when `sources` state is updated and is non-null/not-empty

@acrellin but the `SourceTable` doesn't access the state directly, it can use either the "latest" or "groupSources" depending on where the component is used.

acrellin:
@guynir42 `sources` is passed in as a prop -- the `useEffect` should watch for changes in that variable

acrellin (approved):

acrellin:
@guynir42 can you link to the related issue using the keyword "Closes" so that it'll be closed when this is merged?

guynir42:
Closes #1911

acrellin:
@guynir42 it actually has to be in the PR description in order to work, sorry I should've mentioned that

acrellin:
(I just updated it)

acrellin:
@guynir42 the spinner appears centered in the page -- can you update its containing div/grid's css to make it left-justified? it looks odd centered in the page

guynir42:
I moved the spinner and fixed some tests, I think it is go for second review @acrellin @kmshin1397

acrellin (approved):

guynir42:
@kmshin1397 if you have time to review and merge this one, it's a short one...

kmshin1397 (approved):

PR #1983, opened 2021-05-12
Reduce database queries in photometry PUT by stream id

kmshin1397:
The current version of PUT for photometry using stream IDs makes anywhere from 1 to 3 extra database queries per duplicated photometry point compared to using group IDs.
- For group IDs, each duplicated photometry point just makes one query for the new groups to be added ([these lines](https://github.com/skyportal/skyportal/blob/390d9494e0ef63b6e58f30c7a0786829e99dc7a6/skyportal/handlers/api/photometry.py#L788-L793))
- For stream IDs, we make a call for the stream records associated with the passed-in stream IDs, and then a call to see if the `StreamPhotometry` row exists for each passed-in stream ID as well. 

This PR brings the reduces the number of database queries in the stream ID block to just one per duplicated photometry point (getting the list of streams already associated with a photometry point), which should hopefully bring the posting in line with previous performance.

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -799,7 +799,7 @@ class PhotometryHandler(BaseHandler):
                 # Add new stream_photometry rows if not already present
                 duplicate_stream_ids = set(
                     StreamPhotometry.query_records_accessible_by(
-                        self.current_user, column=[StreamPhotometry.stream_id]
+                        self.current_user, columns=[StreamPhotometry.stream_id]
                     )
                     .filter(
                         StreamPhotometry.photometr_id == duplicate.id,
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -775,11 +775,13 @@ class PhotometryHandler(BaseHandler):
             .query(values_table.c.pdidx, Photometry)
             .join(Photometry, condition)
             .options(joinedload(Photometry.groups))
+            .options(joinedload(Photometry.streams))
         )
 
         for df_index, duplicate in duplicated_photometry:
             id_map[df_index] = duplicate.id
             duplicate_group_ids = set([g.id for g in duplicate.groups])
+            duplicate_stream_ids = set([s.id for s in duplicate.streams])
 
             # posting to new groups?
             if len(set(group_ids) - duplicate_group_ids) > 0:
@@ -797,16 +799,6 @@ class PhotometryHandler(BaseHandler):
             # posting to new streams?
             if stream_ids:
                 # Add new stream_photometry rows if not already present
-                duplicate_stream_ids = set(
-                    StreamPhotometry.query_records_accessible_by(
-                        self.current_user, columns=[StreamPhotometry.stream_id]
-                    )
-                    .filter(
-                        StreamPhotometry.photometr_id == duplicate.id,
-                    )
-                    .all()
-                )
-                # select new streams
                 stream_ids_update = set(stream_ids) - duplicate_stream_ids
                 if len(stream_ids_update) > 0:
                     for id in stream_ids_update:
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -796,24 +796,22 @@ class PhotometryHandler(BaseHandler):
 
             # posting to new streams?
             if stream_ids:
-                streams = Stream.get_if_accessible_by(
-                    stream_ids, self.current_user, raise_if_none=True
-                )
                 # Add new stream_photometry rows if not already present
-                for stream in streams:
-                    if (
-                        StreamPhotometry.query_records_accessible_by(self.current_user)
-                        .filter(
-                            StreamPhotometry.stream_id == stream.id,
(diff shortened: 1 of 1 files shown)


dmitryduev (approved):
Very nice catch, let’s roll this out to staging.

acrellin (approved):

acrellin:
@kmshin1397 How hard would it be to do some local benchmark comparisons?

kmshin1397:
@acrellin Probably not too hard - let me get on that

kmshin1397:
Ah good thing I tested locally there's a typo here too.

acrellin:
@kmshin1397 I get 12.8 seconds on master, and 13.5 seconds with this branch :(

acrellin:
for the big post from the API tests (copied to a Python terminal)

kmshin1397:
That test just makes a POST call and this just affects the PUT call (which is what Kowalski makes) I was going to make that big POST, and then PUT the same data again to test the duplicate photometry update logic. Unfortunately it does look like it's still slower than the group id PUT call, at least locally for me, the PUT with stream IDs times out.

acrellin:
@kmshin1397 first one is against master, second one is against this PR branch: 
```
In [2]: %time r = requests.put("http://localhost:5000/api/photometry", json={'obj_id': "14gqr_unsaved_copy
   ...: ",'mjd': [58000 + i for i in range(50000)],'instrument_id': 1,'mag': np.random.uniform(low=18, hig
   ...: h=22, size=50000).tolist(),'magerr': np.random.uniform(low=0.1, high=0.3, size=50000).tolist(),'li
   ...: miting_mag': 22.3,'magsys': 'ab','filter': 'ztfg','stream_ids': [1,2,3]}, headers={"Authorization"
   ...: : f"token {token_id}"})                      
CPU times: user 78.1 ms, sys: 0 ns, total: 78.1 ms
Wall time: 21.9 s

In [3]: %time r = requests.put("http://localhost:5000/api/photometry", json={'obj_id': "16fil_unsaved_copy
   ...: ",'mjd': [58000 + i for i in range(50000)],'instrument_id': 1,'mag': np.random.uniform(low=18, hig
   ...: h=22, size=50000).tolist(),'magerr': np.random.uniform(low=0.1, high=0.3, size=50000).tolist(),'li
   ...: miting_mag': 22.3,'magsys': 'ab','filter': 'ztfg','stream_ids': [1,2,3]}, headers={"Authorization"
   ...: : f"token {token_id}"})                      
CPU times: user 74.6 ms, sys: 4.56 ms, total: 79.2 ms
Wall time: 23.2 s
```

acrellin:
Ah right, I'm not posting duplicates -- let me try that

acrellin:
they're effectively equivalent (23.2 vs 23.8 seconds for master and this branch PUTting 50k duplicates by stream id)

kmshin1397:
Yeah ok, I think I have another approach that actually works, one sec

kmshin1397:
^ This new commit gets rid of all of those stream ID queries by just loading in the stream IDs with the one big duplicate query. I think we decided not to do this initially because it would load all streams into the session and lead to permission issues but apparently it doesn't because the `test_photometry_stream_put_access` passes and I don't see any bugs in that test's logic. 

Locally, appears even faster than the group ID puts:
image

acrellin:
Hmm, that last commit somehow has the same performance as the previous one for me locally: 
```
In [2]: %time r = requests.put("http://localhost:5000/api/photometry", json={'obj_id': "16fil_unsaved_copy
   ...: ",'mjd': [58000 + i for i in range(50000)],'instrument_id': 1,'mag': np.random.uniform(low=18, hig
   ...: h=22, size=50000).tolist(),'magerr': np.random.uniform(low=0.1, high=0.3, size=50000).tolist(),'li
   ...: miting_mag': 22.3,'magsys': 'ab','filter': 'ztfg','stream_ids': [1,2,3]}, headers={"Authorization"
   ...: : f"token {token_id}"})                      
CPU times: user 81.6 ms, sys: 3.51 ms, total: 85.1 ms
Wall time: 23.2 s

```

acrellin:
ah, nm, these still aren't duplicates -- `np.random`...

kmshin1397:
Ah yeah, I just saved the first object I POST as a dict variable, and then edit just the `stream_ids` to make duplicates for the PUT

acrellin:
@kmshin1397 hmm, with the latest commit here, the large PUT still times out for me locally

kmshin1397:
Hmm @acrellin does it work if you try adding a group id instead of the stream id?

acrellin:
No, that times out, as well. Maybe my laptop is just too slow.

kmshin1397:
Ah ok, maybe we can get @dmitryduev to try it locally as well then

acrellin:
@kmshin1397 with a smaller subset of the test data, I can get it working and confirm this is much faster than master! nice! let's merge when green

acrellin:
~50x faster 👀

dmitryduev:
I checked it with some test data and, too, confirm that this is now much faster than master. Thanks a lot @kmshin1397 @acrellin!!

PR #1984, opened 2021-05-12
Changed img onerror in ThumbnailList.jsx to be more informative

scizen9:
Now onerror script checks if dr8 is being served and uses the "Outside Survey Area" png image.  Default is to serve "Currently Unavailable" png.  Screenshot shows both behaviors.

image


acrellin:
When did new/ref/sub start getting double crosshairs? I thought there was logic that didn't show our crosshairs for those

acrellin:
but couldn't it still be a transient network issue in the dr8 case?

scizen9:
Those demo objects have crosshairs in their thumbnails.  We should probably update those thumbnails.

scizen9:
The origin of this problem was that dr8 errors out when it is outside the survey area, unlike others that serve a cutout with that message in it.  It is possible that dr8 could suffer from a transient problem, should we just always serve currently unavailable?

acrellin:
thoughts @dmitryduev ?

dmitryduev (approved):
I think this is indeed more informative. In the DR8 case, it is more often just outside the survey area vs transient network issues. We could get nerdy and try to discriminate those cases, but I don't think it's worth the effort.

acrellin (approved):

PR #1985, opened 2021-05-12
Add cache invalidation by age

stefanv:
This is a POC to illustrate cache invalidation by age.

Note that, currently, cache invalidation happens only on insert.  We can modify this if it seems more prudent to do it on get, or even in both places.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -13,4 +13,4 @@
 [flake8]
 max-line-length = 80
 select = C,E,F,W,B,B950
-ignore = E501, W503
+ignore = E501, W503, E203
--- a/skyportal/tests/utils/test_cache.py
+++ b/skyportal/tests/utils/test_cache.py
@@ -18,7 +18,7 @@ def cache_parent_dir(tmpdir_factory):
 @pytest.fixture()
 def cache(cache_parent_dir):
     cache_path = pjoin(cache_parent_dir, 'cache')
-    cache = Cache(cache_path, max_items=3)
+    cache = Cache(cache_path, max_items=3, max_age=3)
     yield cache
     shutil.rmtree(cache_path)
 
@@ -53,6 +53,19 @@ def test_cache_cleanup(cache):
 
     for key in range(2, 5):
         assert cache[str(key)] is not None
+
+
+def test_cache_cleanup_by_age(cache):
+    cache['first'] = b'one'
+    f = cache['first']
+    assert f is not None
+
+    # Let object time out of cache
+    time.sleep(3)
+    # Trigger cleanup
+    cache['second'] = b'two'
+    f = cache['first']
+    assert f is None
 
 
 def test_cache_reference_refresh(cache):
--- a/skyportal/tests/utils/test_cache.py
+++ b/skyportal/tests/utils/test_cache.py
@@ -84,3 +84,13 @@ def test_cache_reference_refresh(cache):
 
 def test_cache_nested_root(cache_parent_dir):
     Cache(cache_parent_dir / 'some/deeper/path', 1)
+
+
+def test_cache_no_max_limit(cache_parent_dir):
+    cache = Cache(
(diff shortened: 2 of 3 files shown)


acrellin (changes requested):
Can you also change `fn` to `filename` throughout `cache`? It took me a while to figure out what that variable was supposed to refer to because of the name.

skyportal/utils/cache.py
```diff
@@ -71,20 +76,40 @@ def __setitem__(self, name, data):
...
+        cached_files = [
+            (f.stat().st_mtime, f.absolute()) for f in self._cache_dir.glob('*')
+        ]
+        cached_files = sorted(cached_files, key=lambda x: x[0], reverse=True)
+
+        now = time.time()
+
+        if self._max_age is not None:
+            removed_by_time = [
+                f for (t, f) in cached_files if (now - t) > self._max_age
```
acrellin, line 107:
```suggestion
                filepath for (mod_time, fiilepath) in cached_files if (now - mod_time) > self._max_age
```
for readability

acrellin: and wherever else applicable

stefanv: Sure, I'll push that up.

.flake8
```diff
@@ -13,4 +13,4 @@
 [flake8]
 max-line-length = 80
 select = C,E,F,W,B,B950
-ignore = E501, W503
+ignore = E501, W503, E203
```
acrellin, line 16: I don't think we _always_ want to ignore this rule, though, right?

skyportal/tests/utils/test_cache.py
```diff
@@ -18,7 +18,7 @@ def cache_parent_dir(tmpdir_factory):
 @pytest.fixture()
 def cache(cache_parent_dir):
     cache_path = pjoin(cache_parent_dir, 'cache')
-    cache = Cache(cache_path, max_items=3)
+    cache = Cache(cache_path, max_items=3, max_age=3)
```
acrellin, line 21: Please also make max_items nullable when using max_age

skyportal/utils/cache.py
```diff
@@ -63,28 +68,50 @@ def __setitem__(self, name, data):
...
+
+        if self._max_age is not None:
+            removed_by_time = [
+                filename
+                for (mtime, filename) in cached_files
+                if (now - mtime) > self._max_age
+            ]
+            self._remove(removed_by_time)
+
+        oldest = cached_files[self._max_items :]
```
acrellin, line 113: Please update to make `max_items` optional -- I want to only specify `max_age` w/ no limit on # cache items.

acrellin:
@stefanv also, is there a way around the asymmetrical save/load mechanisms? It's counter-intuitive that `cache` creates the file for you when setting the cache but then just returns the filename rather than the corresponding Python object when loading the cache contents.

.flake8
stefanv: I think we do; that is what is suggested by the Black documentation.

stefanv: https://github.com/psf/black/issues/315#issuecomment-395457972

stefanv:
@acrellin We can certainly modify this in any way that makes sense.

Re: creation: you have to generate the filename and create the file at some point: upon cache insertion seems like the right time.  If you want to change the API, I suggest we return the file object from which the filename & bytes can be accessed (`f.name`, `f.read()`).  I think the existing use cases demanded a filename, hence that is what I used.

The changed API would save having to do `with open(x, 'rb')`.

skyportal/utils/cache.py
stefanv: Done

stefanv:
Failures unrelated.

acrellin (approved):

PR #1974, opened 2021-05-10
Add LCO APIs migration

kmshin1397:
Migration for updating the followup requests API ENUM values for the recently merged LCO requests (#1883)

PR #1990, opened 2021-05-13
Fix migration history

kmshin1397:
Fix branching in Alembic history caused by #1974


dmitryduev (approved):
Thanks for the fix @kmshin1397

PR #1994, opened 2021-05-17
Select new user role when inviting via Group page

### LINKED ISSUE #1769: Allow "View only"-level users to be easily added via front-end
From Daniel Perley:

> it would be useful to give people the ability to scan/see transients from a program but not allow them to save/unsave to it, i.e. have different "tiers" of membership of a program (program owner, member with saving permission, guest without saving permission).  This would help us expand program membership without the concern that people will accidentally save/unsave objects.

To be clear, the back-end infrastructure is already in place, we just need to add an option to make users "View only" when inviting them.

acrellin:
This PR allows one to specify either the "Full user" or "View only" role for a new user when inviting via the group page, as requested.

Closes #1769 

output

Resulting profile page after authing with invite specifying View only user:

Screenshot from 2021-05-17 12-39-01


acrellin:
```
= 1 failed, 430 passed, 3 xfailed, 1 xpassed, 200 warnings in 2165.42s (0:36:05) =
```

dmitryduev (approved):

PR #1991, opened 2021-05-14
WIP: Optimize source handler joins

acrellin:

PR #1999, opened 2021-05-18
Optimize sources filtering by classifications/redshift

kmshin1397:
This takes Ari's changes from #1991 (removing the outer join for filtering on classifications) and adds two more optimizations:
- The accessibility queries for group membership via `groups.users` or `group.users` fields was optimized slightly by matching on the `GroupUser.user_id` join table column instead of the `User.id`, so that we avoid having to join in the `User` table for each of these "is accessible" subqueries
- Explicitly re-order the subqueries joined together for the classification filtering so that the `(Taxonomy.name, Classification.classification)` filtering is applies first before applying a join against the accessible `Classification` rows. This seems to nudge the Postgres query planner in a direction where they are using less layers of nested loops and better using the indices available. Basically, I was seeing logic like nested looping like "for each classification row, check for all groups, check for user membership" etc if I was interpreting the `EXPLAIN` statements correctly, and this seems to free up some of that. 

My general impression of this issue isn't that we're missing some key index somewhere, it's just that when we combine multiple filtering conditions like this along with the permissions stuff, there's just a lot of tables that need to be joined together and there isn't a clear way to fully utilize those indices across all the tables joined together. For example, we have an index on `Classification.classification` but this doesn't really help as much when we need to join in the `Taxonomy` rows and get the taxonomy names associated to each of those before we filter, and so on. This is especially true for classifications because the filtering is based on (classification, taxonomy) pairs which each have their own join tables against groups, etc. 

Closes https://github.com/fritz-marshal/fritz/issues/262


kmshin1397:
@acrellin Do you want to try this commit with however you were testing your own attempts and see if you also see significant improvement? I'm basing my optimization stats on one specific query and the `EXPLAIN` output but just want to make sure that the better estimated cost given by Postgres actually holds true for other queries of the same type.

acrellin:
Current master (pointing at the production DB replica):
```
In [10]: %time r = requests.get("http://localhost:5000/api/sources?maxRedshift=0.9&minRedshift=0.1&pageNum
    ...: ber=1&sortBy=classification&classifications=Sitewide Taxonomy: Ia", headers={"Authorization": f"t
    ...: oken {token}"})  
CPU times: user 3.14 ms, sys: 546 µs, total: 3.68 ms
Wall time: 52.4 s
```
This PR branch:
```
In [12]: %time r = requests.get("http://localhost:5000/api/sources?maxRedshift=0.9&minRedshift=0.1&pageNum
    ...: ber=1&sortBy=classification&classifications=Sitewide Taxonomy: Ia", headers={"Authorization": f"t
    ...: oken {token}"})  
CPU times: user 2.02 ms, sys: 3.71 ms, total: 5.73 ms
Wall time: 50.6 s
```
Pretty similar 😢

acrellin:
I'll try again w/out the redshift filtering to isolate the part of the handler updated here

acrellin:
This PR branch: 
```
In [14]: %time r = requests.get("http://localhost:5000/api/sources?pageNumber=1&sortBy=classification&clas
    ...: sifications=Sitewide Taxonomy: Ia", headers={"Authorization": f"token {token}"})                 
CPU times: user 4.01 ms, sys: 450 µs, total: 4.46 ms
Wall time: 56 s
```

Current master:
```
In [15]: %time r = requests.get("http://localhost:5000/api/sources?pageNumber=1&sortBy=classification&clas
    ...: sifications=Sitewide Taxonomy: Ia", headers={"Authorization": f"token {token}"})                 
CPU times: user 3.67 ms, sys: 0 ns, total: 3.67 ms
Wall time: 59.1 s
```

acrellin:
This is proving to be a tough nut to crack

kmshin1397:
Ahh okay let me take a look at it with this specific query. Which user token are you using for these?

acrellin:
Good point -- it might've been a Sys admin token -- I'll try again w/ a non-sys admin token and report back.

acrellin:
With a non-sys admin token:
This PR branch:
```
In [11]: %time r = requests.get("http://localhost:5000/api/sources?pageNumber=1&sortBy=classification&classifications=Sitewide Taxonomy: Ia",
    ...:  headers={"Authorization": f"token {token.id}"})             
CPU times: user 5.82 ms, sys: 31 µs, total: 5.85 ms
Wall time: 43.6 s
```

Master:
```
Wall time: 43.1 s
```

skyportal/models.py
```diff
@@ -733,14 +836,17 @@ def last_detected_at(cls, user):
...
-            .all()
-            if phot.snr is not None and phot.snr > PHOT_DETECTION_THRESHOLD
-        ]
-        return max(detections, key=(lambda x: x[0]))[1] if detections else None
+            .filter(Photometry.snr.isnot(None))
+            .filter(Photometry.snr > PHOT_DETECTION_THRESHOLD)
+            .order_by(Photometry.mjd.desc())
+            .limit(1)
+            .scalar()
+        )
```
kmshin1397, line 849: This and the `peak_detected_mag` changes should speed up getting these properties slightly by doing all of the computation on the Postgres side. But I made this change earlier in the optimization process, and now this honestly  doesn't matter too much as we only use these in the single GET method and I stopped using it entirely in the multiple GETs (switched to using the SQL expression versions instead).

skyportal/handlers/api/source.py
```diff
@@ -1062,8 +1083,40 @@ class Validator(Schema):
...
+            peak_detected_at = Obj.peak_detected_at(self.current_user)
+            query_results["sources"] = query_results["sources"].add_columns(
+                peak_detected_at
+            )
+
+            # Load in all peak_detected_mag values at once
+            peak_detected_mag = Obj.peak_detected_mag(self.current_user)
+            query_results["sources"] = query_results["sources"].add_columns(
+                peak_detected_mag
+            )
```
kmshin1397, line 1109: Instead of querying the last and peak detection info separately for each object in the result, we now do them in bulk by joining against the SQL expression versions of the properties instead. This speeds up the loop below dramatically.

skyportal/handlers/api/source.py
```diff
@@ -1105,25 +1158,32 @@ class Validator(Schema):
...
-                )
-                obj_list[-1]["peak_detected_at"] = obj.peak_detected_at(
-                    self.current_user
+
+                obj_list[-1]["last_detected_at"] = (
+                    (last_detected_at - last_detected_at.utcoffset()).replace(
+                        tzinfo=UTC
+                    )
+                    if last_detected_at
+                    else None
```
kmshin1397, line 1167: The SQL expression version of the `last_detected_at` and `peak_detected_at` returns the values in local time for some reason, so we just change it back to UTC again here.

skyportal/handlers/api/candidate.py
```diff
@@ -1124,7 +1189,22 @@ def grab_query_results(
...
+            .options(query_options)
+            .join(page_ids_values, page_ids_values.c.id == Obj.id)
+            .order_by(page_ids_values.c.ordering)
+        )
+    else:
+        # If there are no values, the VALUES statement above will cause a syntax error,
+        # so just switch to this simple IN query. Should return nothing either way
+        items = (
+            DBSession().query(Obj).options(query_options).filter(Obj.id.in_(page_ids))
+        )
```
kmshin1397, line 1207: The change here is that instead of looping through the ordered Obj IDs to be returned and manually building up a Python list of the Objs, we return a Query object against `Obj`, with the order maintained by joining against a `VALUES` table. This accomplished two goals:
- All the Obj entities are fetched in one go instead of doing a bunch of `Obj.query.get()` queries.
- Returning a SQLAlchemy Query here instead of a standard Python list means I can actually do the `add_columns()` stuff with the last and peak detections I did in `sources.py` to speed that part up.

kmshin1397 (commented):
New optimizations (on top of what was already part of this PR) are explained in comments:

kmshin1397:
This PR has been updated to optimize the true bottlenecks in the sources query: the fetching of Obj and associated data after getting the initial list of Obj IDs to return. I chose to keep the other optimizations I did for the "getting the list of Obj IDs" step around since it doesn't hurt to have that part faster. With the most recent changes, the sources GET should be much, much faster (especially if `removeNested` is set to True). I didn't really touch the parts that add the data for `if not removeNested`, and those blocks are still pretty slow, but this should be a good first step. 

@acrellin @dmitryduev

skyportal/handlers/api/candidate.py
```diff
@@ -1124,7 +1189,22 @@ def grab_query_results(
 
     items = []
     query_options = [joinedload(Obj.thumbnails)] if include_thumbnails else []
-    for item_id in page_ids:
-        items.append(Obj.query.options(query_options).get(item_id))
+
+    if len(page_ids) > 0:
```
acrellin, line 1193: While we're here, this has always been a bit confusing to me -- can we rename `page_ids` to something more descriptive like `obj_ids_in_page` or something? I frequently find myself wondering what page IDs are :joy:

acrellin (commented):
Looks great so far -- will give a more thorough review a bit later. One quick (unrelated-ish) request for now:

acrellin (approved):
Really nice, thanks @kmshin1397 !

acrellin:
Failures unrelated: 
```
FAILED skyportal/tests/frontend/test_followup_requests.py::test_submit_new_followup_request_SEDM
= 2 failed, 433 passed, 3 xfailed, 1 xpassed, 200 warnings, 1 rerun in 2315.67s (0:38:35) =
```

skyportal/handlers/api/candidate.py
```diff
@@ -1124,7 +1189,25 @@ def grab_query_results(
...
+        )
+    else:
+        # If there are no values, the VALUES statement above will cause a syntax error,
+        # so just switch to this simple IN query. Should return nothing either way
+        items = (
+            DBSession()
+            .query(Obj)
+            .options(query_options)
+            .filter(Obj.id.in_(obj_ids_in_page))
+        )
```
dmitryduev, line 1210: Then why not just set `items = []` or `None` (whatever is expected downstream)?

skyportal/handlers/api/source.py
```diff
@@ -946,28 +920,75 @@ class Validator(Schema):
...
-                    lambda c: (c.split(":")[0].strip(), c.split(":")[1].strip()),
-                    classifications,
+                classification_subquery = classification_query.subquery()
+
+                # We need an outer join here when just sorting by classifications
+                # to support sources with no classifications being sorted to the end
+                obj_query = obj_query.join(
+                    classification_subquery,
+                    Obj.id == classification_subquery.c.obj_id,
+                    isouter=True,
```
dmitryduev, line 989: Wouldn't this be slow compared to what we are currently doing?

skyportal/handlers/api/source.py
dmitryduev: A really nice idea!

skyportal/models.py
dmitryduev: Would still matter if people are iterating over a list of `obj_id`'s with single GET statements. I actually hope it's not what's happening now...

dmitryduev (commented):
Looks great! Some parts are beyond my level of expertise on sql, so reviewed what I could understand :)

skyportal/handlers/api/candidate.py
kmshin1397: Downstream expects a Query object to be able to make `.add_columns()` calls on. This was the way I chose to fix because it was the quickest, but it could easily just be a `None` here and then do a bunch of `if not None` checks or something like that further downstream. That would probably be the more "correct" way of doing it tbh

skyportal/handlers/api/source.py
kmshin1397: You may think so, but not in practice. Under the hood, this generates pretty similar queries from the perspective of the Postgres query planner, with similar amounts of nested looping and such. It really only changes the ordering of those loops, like which records/conditions are the innermost loop etc, or something along those lines. I arrived at this through experimentation and a somewhat vague sense of exactly what the query planner is doing to be honest, so I couldn't explain to you exactly why it's faster but it does indeed seem to be the case.

kmshin1397: At worst I don't think it performs any worse than the previous version (even if it looks a little uglier on the SQLAlchemy side). Definitely, getting rid of the outer join here when not needed is a strictly good thing (and was the idea Ari had originally).

skyportal/handlers/api/candidate.py
kmshin1397: Ok I just pushed a commit to simplify this (just returns `[]` for items if no results) and handle it downstream

skyportal/handlers/api/candidate.py
```diff
@@ -1199,15 +1205,6 @@ def grab_query_results(
...
-        # If there are no values, the VALUES statement above will cause a syntax error,
-        # so just switch to this simple IN query. Should return nothing either way
-        items = (
-            DBSession()
-            .query(Obj)
-            .options(query_options)
-            .filter(Obj.id.in_(obj_ids_in_page))
-        )
 
     info[items_name] = items
```
dmitryduev, line 1209: Wouldn't this error in the `else` case (items undefined)?

kmshin1397: `items` is actually initialized to `[]` already earlier (line 1194 in the newest commit)

dmitryduev: Oops, missed that. In it goes then!

PR #2001, opened 2021-05-19
Update invitations role ID migration script to back-populate

acrellin:


stefanv (approved):
Assuming `Full user` is the default, that works.  Thanks Ari!

kmshin1397 (approved):

Issue #1795, opened 2021-03-18
Add uncertainty to redshift


scizen9:
Should we add and display redshift errors?  For example: 0.003 +- 0.001.

From standup discussion it is important to document that we make no assumptions about the form of the errors.

acrellin:
@scizen9 just a heads up that if you put "Closes #1795" in the PR description, when it's merged, the linked issue will automatically be closed.

PR #2002, opened 2021-05-19
Added redshift_error to Obj model and redshift history update function in candidate.py

scizen9:
Need to add redshift_error to tests and propagate into front end display.


scizen9:
Here are some screenshots: 

1. Display on source page:
image

2. Updating redshift
image

3. Redshift history
image

scizen9:
Will fix [issue #1795 ](https://github.com/skyportal/skyportal/issues/1795)

static/js/components/UpdateSourceRedshift.jsx
```diff
@@ -33,6 +33,10 @@ const UpdateSourceRedshift = ({ source }) => {
   const [state, setState] = useState(
     source.redshift ? String(source.redshift) : ""
   );
+  const [uncstate, setUncState] = useState(
```
acrellin, line 36: Actually, probably best to combine redshift and error into a single state object. Then add relevant "name" attrs to the input fields, and update using event.target.name as the state object key (see https://github.com/skyportal/skyportal/blob/master/static/js/components/NewGroupForm.jsx#L60-L67)

acrellin (commented):
Looks good overall, just one request regarding the React state structure below. Also needs a test (can just expand the existing test)

static/js/components/UpdateSourceRedshift.jsx
scizen9: Great suggestion.  Forced me to learn more about react and update my firefox after 8 years of neglect!

scizen9:
Still need migration script, however.

scizen9:
Apart from errors from 'black' when generating the migration script, it seems to be ready.

acrellin (approved):
Looks great, thanks @scizen9 . Can you also add a test where we update redshift but leave the error blank? Can just copy existing test and removing the error-related lines

static/js/components/UpdateSourceRedshift.jsx
```diff
@@ -113,12 +150,16 @@ const UpdateSourceRedshift = ({ source }) => {
...
+                  handleSubmit({ redshift: null, redshift_error: null });
                 }}
                 startIcon={<ClearIcon />}
                 size="large"
                 data-testid="nullifyRedshiftButton"
-                disabled={isSubmitting || source.redshift === null}
+                disabled={
+                  isSubmitting ||
+                  source.redshift === null ||
+                  source.redshift_error === null
```
guynir42, line 161: Just making sure: the button is disabled if the redshift error is null? event if the redshift itself has a value?

guynir42 (approved):
Very minor suggestions. Other than that it looks good to me.

static/js/components/UpdateSourceRedshift.jsx
acrellin: Good catch -- we should definitely allow the error to be null

skyportal/handlers/api/candidate.py
```diff
@@ -41,6 +41,9 @@ def update_redshift_history_if_relevant(request_data, obj, user):
                 "set_by_user_id": user.id,
                 "set_at_utc": datetime.datetime.utcnow().isoformat(),
                 "value": request_data["redshift"],
+                "uncertainty": request_data["redshift_error"]
```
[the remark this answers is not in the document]
scizen9: Yes.  Updated.

static/js/components/UpdateSourceRedshift.jsx
scizen9: This hinges on a policy decision.  In my opinion, one should not enter a redshift without an error, but others may differ on that.  I'm open to discussion on this.

acrellin: I think for now we should allow it to be nullable, and we can revisit it later if needed

acrellin (changes requested):
Marking this as in discussion, per the nullable redshift conversation

static/js/components/UpdateSourceRedshift.jsx
scizen9: Implemented null entry for redshift_error, and added test for same.

acrellin:
I think we should only render the plus/minus when there's a redshift error. Here it's rendered w/out a redshift value or error: 
image

acrellin:
Also, should the redshift error be factored into the DM and DL values by adding an associated error to those?

image

scizen9:
Current behaviour:

1. No redshift:
image

2. Redshift, no error:
image

3. Redshift and error:
image

acrellin (approved):

Issue #1613, opened 2021-01-21
Comments lists are not responsive to container width


kmshin1397:
The comments list container does not expand to fill the width of its parent Accordion component

dmitryduev:
@kmshin1397 ping?

kmshin1397:
Ah yes still planning to get to this at some point just forgot about it since it was low priority

joelachance:
Looking at this one, here's some changes:

1. I removed the background color for non-hover states. There's only a background when hovered over.
2. Removed some padding to fit some more comments in the space we have
3. Updated the delete button to fit closer with the compact comments ticket #976 

<img width="446" alt="Screen Shot 2021-05-18 at 10 49 36 PM" src="https://user-images.githubusercontent.com/3891660/118753972-a833be80-b82b-11eb-9829-ac41c4b89d38.png">
<img width="437" alt="Screen Shot 2021-05-18 at 10 56 11 PM" src="https://user-images.githubusercontent.com/3891660/118754260-44f65c00-b82c-11eb-8cf4-214010675589.png">

acrellin:
Fixed in https://github.com/skyportal/skyportal/pull/2004

PR #2004, opened 2021-05-20
Feature/976 compact comments

### LINKED ISSUE #976: Requested feature: more compact comments

joelachance:
Resolves #976 and #1613.

Here's some screenshots:
dark_compact_mobile
dark_reg_desktop
dark_reg_mobile
add_comment_button
light_reg_desktop
light_reg_mobile
ui preferences
light_compact_mobile
light_compact_desktop
dark_compact_desktop


skyportal/tests/frontend/test_sources.py
```diff
@@ -412,7 +412,7 @@ def test_centroid_plot(
 
         # Use this commented line to save a new version of the expected plot
         # if changes have been made to the component:
-        # generated_plot.save(expected_plot_path)
+        generated_plot.save(expected_plot_path)
```
joelachance, line 415: @acrellin, I had to un-comment these to get tests to pass, I wanted to call this out in case there's a better way to do this.

acrellin: I think this should just be temporarily un-commented while you run the tests locally, then commit the resulting saved file, and re-comment this line

joelachance: Ok, that's what I thought as well. I did that here:
351fc0e, and the tests failed.
This commit added them back in, tests passed: bb5b674
I commented them back out in the latest commit, and things are failing again on the same two tests: cbb784a

I don't believe these tests aren't connected to this work, so I'm not really sure what to do here. Any thoughts, @acrellin ?

Thanks for the help too, I do appreciate it.

package.json
```diff
@@ -82,15 +82,15 @@
     "@types/node": "14.14.10",
     "babel-eslint": "10.1.0",
     "babel-loader": "8.2.2",
-    "eslint": "7.25.0",
-    "eslint-config-airbnb": "18.2.0",
+    "eslint": "^7.25.0",
```
acrellin, line 85: We prefer to pin specific package versions (npm will annoyingly add these carets automatically oftentimes when running `npm install`)

acrellin (commented):
@joelachance Is the default to render the original full-size comments or the compact comments? Might the tests be expecting one thing and getting another there? Also, did the "Recent comments" -> "Comments" change come before the images we re-created? Let me know if it'd be helpful for me to check out this branch and do some debugging

package.json
joelachance: I noticed this too, I'm going to revert these

joelachance:
> @joelachance Is the default to render the original full-size comments or the compact comments? Might the tests be expecting one thing and getting another there? Also, did the "Recent comments" -> "Comments" change come before the images we re-created? Let me know if it'd be helpful for me to check out this branch and do some debugging

@acrellin 
I'm not sure! I was actually thinking about this same thing this week. I simply added this to Redux, is there a default for user preferences somewhere outside of Redux I can check? At least, we could explicitly set a preference so it's clear to the next dev to look in that file.

I agree with everything you've mentioned so far though, we can't have those plot generation lines un-commented because then those tests are useless, they'll just generate plots every time those tests get hit.

I'll try to join standup next week, I've had a lot of meetings over that time lately.

static/js/components/CommentList.jsx
```diff
@@ -39,6 +170,9 @@ const CommentList = () => {
   const obj = useSelector((state) => state.source);
   const userProfile = useSelector((state) => state.profile);
   const permissions = useSelector((state) => state.profile.permissions);
+  const compactComments = useSelector(
+    (state) => state.profile.preferences.compactedComments
```
acrellin, line 174:
```suggestion
    (state) => state.profile.preferences?.compactComments
```
Just in case `preferences` is uninitialized, and `compacted` -> `compact` for consistency

static/js/components/CommentListMobile.jsx
```diff
@@ -106,6 +223,9 @@ const CommentListMobile = ({ isCandidate }) => {
   const candidate = useSelector((state) => state.candidate);
   const obj = isCandidate ? candidate : source;
   const userProfile = useSelector((state) => state.profile);
+  const compactComments = useSelector(
+    (state) => state.profile.preferences.compactedComments
```
acrellin, line 227:
```suggestion
    (state) => state.profile.preferences.compactComments
```

static/js/components/UIPreferences.jsx
```diff
@@ -14,6 +14,7 @@ const UIPreferences = () => {
   const preferences = useSelector((state) => state.profile.preferences);
   const currentTheme = preferences?.theme;
   const invertThumbnails = preferences?.invertThumbnails || false;
+  const compactedComments = preferences?.compactedComments || false;
```
acrellin, line 17:
```suggestion
  const compactComments = preferences?.compactComments || false;
```

static/js/components/UIPreferences.jsx
```diff
@@ -31,6 +32,13 @@ const UIPreferences = () => {
     dispatch(profileActions.updateUserPreferences(prefs));
   };
 
+  const commentsToggled = (event) => {
+    const prefs = {
+      compactedComments: event.target.checked,
```
acrellin, line 37:
```suggestion
      compactComments: event.target.checked,
```

static/js/components/UIPreferences.jsx
```diff
@@ -47,6 +55,14 @@ const UIPreferences = () => {
     />
   );
 
+  const commpactCommentsSwitch = (
+    <Switch
+      value="Compact Comments"
+      checked={compactedComments}
```
acrellin, line 61:
```suggestion
      checked={compactComments}
```

acrellin (approved):

acrellin:
Looks great, thanks @joelachance !  I hope you don't mind I took the liberty to update a variable name and add a JS object property fail-safe

acrellin:
Will merge when green!

joelachance:
Of course, thanks!

acrellin:
Those two tests are still failing... I'll see what I can do to fix them up.

acrellin:
@joelachance I just re-generated the pngs locally and pushed them up. I'm guessing it might be just a screen resolution-related issue. 🤞🏼

acrellin:
That did it! Just needed to generate them using a 1920 screen resolution, which is what the Firefox driver uses in the tests.

joelachance:
@acrellin just a heads up, this should resolve #1613 as well.

stefanv:
:tada: Thanks, Joe!

Issue #2006, opened 2021-05-20
Security review: ensure relevant ACLs are required for POST, PUT/PATCH, DELETE handlers


acrellin:
Especially relevant now that we will have more "View only" users w/ no ACLs

PR #2009, opened 2021-05-21
Patch potential security vulnerabilities in Tornado handler methods

acrellin:
Closes https://github.com/skyportal/skyportal/issues/2006

### DIFF
--- a/skyportal/handlers/api/group_admission_request.py
+++ b/skyportal/handlers/api/group_admission_request.py
@@ -1,4 +1,4 @@
-from baselayer.app.access import auth_or_token
+from baselayer.app.access import auth_or_token, permissions
 from baselayer.app.custom_exceptions import AccessError
 from ..base import BaseHandler
 from ...models import (
@@ -86,7 +86,7 @@ class GroupAdmissionRequestHandler(BaseHandler):
         self.verify_and_commit()
         return self.success(data=response_data)
 
-    @auth_or_token
+    @permissions(["Upload data"])
     def post(self):
         """
         ---
@@ -203,7 +203,7 @@ class GroupAdmissionRequestHandler(BaseHandler):
             self.flow.push(group_admin.id, "skyportal/FETCH_NOTIFICATIONS", {})
         return self.success(data={"id": admission_request.id})
 
-    @auth_or_token
+    @permissions(["Upload data"])
     def patch(self, admission_request_id):
         """
         ---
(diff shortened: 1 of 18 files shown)


skyportal/handlers/api/group_admission_request.py
```diff
@@ -86,7 +86,7 @@ def get(self, admission_request_id=None):
         self.verify_and_commit()
         return self.success(data=response_data)
 
-    @auth_or_token
+    @permissions(["Upload data"])
```
dmitryduev, line 89: Should still allow even the view-only users to request admission to groups?

skyportal/handlers/api/stream.py
```diff
@@ -242,7 +242,7 @@ def post(self, stream_id, *ignored_args):
 
         return self.success(data={'stream_id': stream_id, 'user_id': user_id})
 
-    @auth_or_token
+    @permissions(["System admin"])
```
dmitryduev, line 245: Oh shoot :)

skyportal/handlers/api/taxonomy.py
```diff
@@ -200,7 +200,7 @@ def post(self):
 
         return self.success(data={'taxonomy_id': taxonomy.id})
 
-    @auth_or_token
+    @permissions(['Delete taxonomy'])
```
dmitryduev, line 203: Oh shoot!!

skyportal/handlers/api/group_admission_request.py
acrellin: Sounds reasonable -- updated.

acrellin:
@dmitryduev okay to merge?

skyportal/handlers/api/stream.py
```diff
@@ -178,7 +178,7 @@ def delete(self, stream_id):
 
 
 class StreamUserHandler(BaseHandler):
-    @auth_or_token
+    @permissions(["System admin"])
```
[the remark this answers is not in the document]
dannygoldstein: These (and below) are not really needed anymore. These ACLs are automatically enforced by the new permissions system when the ORM issues a select query. The only difference is that the ORM would send back a 400 instead of a 403.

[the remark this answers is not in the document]
acrellin: That's true for these stream endpoints, yes, but doesn't hurt to have them

PR #2013, opened 2021-05-25
Implement group-specific source saving access logic

acrellin:
This PR allows group admins to control which group members can save sources to their group via a new `GroupUser.can_save` column, which is specified when either adding existing users to a group or inviting new users to join Fritz via the Group page. A toggle button is added on the group users table so that admins can alter user permissions after they've been added.

add_invite_user_forms

save_access_toggle

add_user_cantsave

cantsave


skyportal/handlers/api/group.py
```diff
@@ -511,10 +527,20 @@ def patch(self, group_id, *ignored_args):
...
+                "Missing required parameter: at least one of `admin` or `canSave`"
+            )
+        admin = data.get("admin", groupuser.admin) in [True, "true", "True", "t", "T"]
+        can_save = data.get("canSave", groupuser.can_save) in [
+            True,
+            "true",
+            "True",
+            "t",
+            "T",
+        ]
```
kmshin1397, line 541: I know it was already like this before, but perhaps it would be good to standardize what we accept as `True` for these parameters. These are more forgiving than the equivalents in the POST handler. I don't really care which one we settle on but it doesn't seem right to have the discrepancy there. Maybe just only accept what's parsed as `True` by the `data.get()` call, since that's probably the most common in the rest of the app?

skyportal/handlers/api/invitations.py
```diff
@@ -148,6 +155,7 @@ def post(self):
                 .all()
             )
         group_admin = data.get("groupAdmin", [False] * len(groups))
+        can_save = data.get("canSave", [True] * len(groups))
         admin_for_groups = [
             el in [True, "True", "true", "t", "T"] for el in group_admin
         ]
```
kmshin1397, line 161: Same comment here

skyportal/models.py
```diff
@@ -1228,9 +1235,21 @@ class Candidate(Base):
...
+    query = DBSession().query(cls)
+    if not user_or_token.is_system_admin:
+        query = query.join(Group).join(GroupUser)
+        query = query.filter(GroupUser.user_id == user_id, GroupUser.can_save.is_(True))
+    return query
+
+
+Source.create = accessible_by_group_members & CustomUserAccessControl(
+    source_create_access_logic
+)
```
kmshin1397, line 1251:
```suggestion
Source.create = CustomUserAccessControl(
    source_create_access_logic
)
```
I don't think we need the `accessible_by_group_members` anymore as that will also be accounted for by the `query = query.filter(GroupUser.user_id == user_id,` part in the `source_create_access_logic`

static/js/components/InviteNewGroupUserForm.jsx
```diff
@@ -70,10 +73,10 @@ const InviteNewGroupUserForm = ({ group_id }) => {
     });
   };
 
-  const toggleAdmin = (event) => {
+  const toggleCheckbox = (event) => {
     setFormState({
       ...formState,
-      admin: event.target.checked,
+      [event.target.name]: event.target.checked,
```
kmshin1397, line 79: Nice, clever way of generalizing this function. I probably would've defined two separate toggle functions haha

kmshin1397 (commented):
Thanks @acrellin - looks mostly good! Just a few minor comments

skyportal/handlers/api/group.py
```diff
@@ -440,7 +445,11 @@ def post(self, group_id, *ignored_args):
                 f"User {user_id} is already a member of group {group_id}."
             )
 
-        DBSession().add(GroupUser(group_id=group_id, user_id=user_id, admin=admin))
+        DBSession().add(
+            GroupUser(
+                group_id=group_id, user_id=user_id, admin=admin, can_save=can_save
```
profjsb, line 450: We're not checking if `can_save` is indeed a Boolean so this will likely throw an error if the API is called with `t` or the like.

profjsb (commented):
LGTM ... I agree with @kmshin1397's comments. Will approve after those are addressed.

acrellin:
@profjsb @kmshin1397 Thanks for the reviews -- all suggestions implemented, ready for re-review.

kmshin1397 (approved):

profjsb (approved):

PR #2014, opened 2021-05-26
Remove duplicates in sources query

kmshin1397:
Not sure if there's another reason for the slowness in the query but this seems to at least resolve the duplicates issue. Will look into it further tomorrow.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -27,6 +27,7 @@ from ...models import (
     Instrument,
     Obj,
     Source,
+    Thumbnail,
     Token,
     Photometry,
     Group,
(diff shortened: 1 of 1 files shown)


acrellin (approved):

skyportal/handlers/api/source.py
```diff
@@ -1138,6 +1143,11 @@ class Validator(Schema):
                         reverse=True,
                     )
 
+                if include_thumbnails and not remove_nested:
+                    obj_list[-1]["thumbnails"] = Thumbnail.query.filter(
```
acrellin, line 1147:
```suggestion
                    obj_list[-1]["thumbnails"] = Thumbnail.query_records_accessible_by(self.current_user).filter(
```

kmshin1397: The thumbnail read policy is just if the obj is readable, I believe, so I just bypassed the accessible part. But this is probably better anyways for future-proofing.

acrellin: Yeah, it's minor, but I heard some discussion about thumbnails not being read-public in the near future, so just wanted to be sure it's future-proofed

PR #2018, opened 2021-05-26
Skip source detection stats by default in GET sources

kmshin1397:
By default, don't return photometry detection statistics like `last_detected_at` and `peak_detected_at` in the GET sources call in order to avoid heavy computations. 

I kept the fields in for the front-end's sources calls since those will all be through the paginated tables and thus limited to 100 sources per page.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -420,6 +420,14 @@ class SourceHandler(BaseHandler):
             description: |
               Boolean indicating whether to include associated thumbnails. Defaults to false.
           - in: query
+            name: includeDetectionStats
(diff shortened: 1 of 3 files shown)


acrellin (approved):

acrellin:
https://github.com/skyportal/skyportal/pull/2019 should address the CI error here

acrellin:
@kmshin1397 wanna try pulling in latest master now?

acrellin:
Model Logic build failure is unrelated and is fixed in https://github.com/skyportal/skyportal/commit/ca18b5bb07b8b4bd080e2c24692c85b13b642b45

PR #2045, opened 2021-06-02
Specify server.ssl = True in config defaults

acrellin:
Closes https://github.com/skyportal/skyportal/issues/2044

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -258,7 +258,7 @@ server:
   # You need to have Google+ API enabled; it takes a few minutes to activate.
   host: localhost
   port: 5000  # This is the public-facing port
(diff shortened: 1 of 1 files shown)


stefanv:
I think the correct thing to do is to simply set ssl to True in our config.

Sorry, I may have misled you in my quick comments this morning.

acrellin:
@stefanv done

acrellin:
See also https://github.com/fritz-marshal/fritz/pull/293

acrellin:
@stefanv can you take another look here? Thanks

stefanv:
So this will make most default Docker deployments break; we don't have SSL certificates enabled by default.  We could probably some kind of self-signed certificate, but there is more work to make that happen.

PR #2046, opened 2021-06-02
Open external thumbnail links in new tab

kmshin1397:
Per Dan Perley:

> Minor functionality suggestion: on the Candidates (scan) page, if you left click the SDSS/PS1 images, the page loads in the same window.  Once this happens there is no way to go back to where you were before - the "back" browser returns you to an empty Candidates query, so you have to re-issue the search and scan again from the beginning.  So, it would be nice if these links (and any other) opened in a new tab instead of in the same window.


acrellin (approved):

acrellin:
Thanks @kmshin1397 , feel free to merge when green

kmshin1397:
Single failure is unrelated: `FAIL: skyportal/tests/frontend/test_scanning_page.py:566 in test_candidate_redshift_filtering` seems like a timeout due to Github Actions variation in vm performance

PR #2047, opened 2021-06-03
Implement an Obj DELETE handler

kmshin1397:
This PR implements an `DELETE Obj` handler to support deletion of `Obj` records through the API. 
- System admins are allowed to delete `Obj` records by default. 
- The `allow_nonadmins_delete_objs` configuration can be set to also allow regular users to delete an `Obj` as long as all associated data with that `Obj` is also able to be deleted by said user. 
  - This is just checked by loading all related rows into the database sessions and letting the `verify_and_commit` check everything that would be deleted if the parent `Obj` was deleted. 
- The change in `test/fixtures.py` is to properly handle teardown of `Obj` fixtures where the test deletes the `Obj` before the teardown code can do it themselves. 

This was the behavior that we discussed during standup as I remember it, but if I forgot or mis-remembered anything, let me know.


skyportal/handlers/api/obj.py
```diff
@@ -0,0 +1,68 @@
...
+
+        DBSession().delete(obj)
+        try:
+            self.verify_and_commit()
+        except AccessError as e:
+            error_msg = (
+                "Insufficient permissions: Objs may only be deleted by system admins"
+            )
+            if cfg["misc.allow_nonadmins_delete_objs"]:
+                error_msg += " or by users who own all associated data with the Obj"
```
acrellin, line 64:
```suggestion
                error_msg += " or by users who own all data associated with the Obj"
```

skyportal/handlers/api/obj.py
```diff
@@ -0,0 +1,68 @@
...
+            obj_id,
+            self.current_user,
+            mode='delete',
+            raise_if_none=True,
+            # Load in all related entities into the session
+            # so that they can be checked by the permission verification
+            options=[
+                joinedload(Obj.photometry),
+                joinedload(Obj.spectra),
+                joinedload(Obj.comments),
```
acrellin, line 44: Others may have commented but the user could still own all other data, i.e. photometry, spectra, etc

skyportal/models.py
```diff
@@ -637,11 +637,17 @@ def token_groups(self):
 Token.groups = token_groups
 
 
+def delete_obj_if_all_data_owned(cls, user_or_token):
```
acrellin, line 640: Shouldn't the check re: associated data happen here rather than via a joinedload passed to the query method?

acrellin (commented):
Looking mostly good, a few questions below

guynir42 (approved):
Looks good to me.

skyportal/handlers/api/obj.py
kmshin1397: Ah yeah, I wasn't sure what to do with that. I'm fine with allowing that if you are.

skyportal/models.py
kmshin1397: That was my first thought as well, but the joinedload method was much simpler to implement so I went with that instead. Doing that logic here would involve doing a bunch of `query_rows_accessible_by` queries for all the associated data, joining them against the Obj, etc. to look for Objs that are allowed to be deleted. The way I have it right now, I just let the loop in the `verify_and_commit` check all those for me instead of having it happen when I first fetch the Obj to be deleted. That said, I agree that doing the check here would be the more correct way of doing it from a good-coding-practices standpoint, so I'm happy to make that change if you feel I should.

acrellin: Yeah, sorry to request a revision here, but I do think the access policy defined here should do all of the relevant checks, just to be consistent with how they're implemented and used everywhere else. That way, if we ever need to call `Obj.get_if_accessible_by(..., mode="delete")` anywhere else, we can rest assured the correct checks are still happening. I think it should be abstracted away here.

skyportal/handlers/api/obj.py
acrellin: Yeah, I think comments should be excluded from this check. And perhaps classifications? Maybe these are questions we should bring to others in the group (the astronomers, specifically)

skyportal/models.py
kmshin1397: No problem, that makes perfect sense. I'll work on that

skyportal/handlers/api/obj.py
kmshin1397: Yeah I think that makes sense. I agree this is probably something we should ask the astronomers for confirmation on. @guynir42 @dmitryduev thoughts? Should users be able to delete an Obj even if not all comments/classifications on it were submitted by themselves?

guynir42: I think this is ok in principle, but it does mean we will have to implement the patch as well as the delete methods. 

I think that if someone is sure that he wants to delete an object, then the conversation should go with it. 

If the poster just wants to make changes to the object it should be through patch, but if that is not available then he will have to delete and repost. Then we could end up losing comments (and annotations!) for no good reason. 

I hope I can get on this when I settle in at Berkeley but I don't know when exactly that would happen...

kmshin1397: Yeah, needing a PATCH does make sense. Looks like we do expose that functionality as the Source PATCH handler already [here](https://skyportal.io/docs/api.html#tag/sources/paths/~1api~1sources~1obj_id/patch). Is there anything missing functionality-wise from that that users would need from a Obj PATCH handler, do you think?

guynir42: Ahhh there's a source PATCH. Great. I think that probably answers most of the use-cases. Except if someone had posted a bunch of un-saved objs that are not considered sources. But that seems like a fairly small edge case that we can handle when we do the full handlers overhaul.

kmshin1397:
@acrellin Ready for re-review :)

skyportal/handlers/api/obj.py
```diff
@@ -0,0 +1,51 @@
...
+            content:
+              application/json:
+                schema: Error
+        """
+        try:
+            obj = Obj.get_if_accessible_by(
+                obj_id,
+                self.current_user,
+                mode='delete',
+                raise_if_none=False,
```
acrellin, line 38:
```suggestion
                raise_if_none=True,
```
otherwise the `DBSession().delete(obj)` below could error if an invalid ID is specified

skyportal/models.py
```diff
@@ -2709,6 +2714,107 @@ class CommentOnSpectrum(Base, CommentMixin):
...
+            isouter=True,
+        )
+        .filter(nondeletable_candidates.c.obj_id.is_(None))
+        .join(
+            nondeletable_sources,
+            nondeletable_sources.c.obj_id == cls.id,
+            isouter=True,
+        )
+        .filter(nondeletable_sources.c.obj_id.is_(None))
+        .filter(allow_nonadmins)
```
acrellin, line 2811: Does this need to be `sa.literal(allow_nonadmins)`?

acrellin (commented):
Thanks for making the updates -- just one potential issue in the handler (`raise_if_none`)

skyportal/models.py
kmshin1397: Doesn't seem like it's necessary as far as I can tell, but I added it in for clarity anyways

skyportal/handlers/api/obj.py
kmshin1397: Good catch, I turned this off for testing purposes and forgot to turn it back on

skyportal/models.py
kmshin1397: Took your feedback, and also moved the permission logic code to right above the Obj class just to group things more logically.

acrellin (approved):
Looks great, thanks @kmshin1397 !

acrellin:
Will merge when green

PR #2062, opened 2021-06-08
Scanning profile improvements

kmshin1397:
This PR implements the following improvements to scanning profiles:

- Previously added scanning profiles can now be edited
  - To support this cleanly, I factored out the profile form into its own component that had slightly different behavior based on whether it was designed for creating or editing, instead of duplicating all that react hook forms logic/JSX twice. This is mainly why the diff is so big: just copy-pasted a bunch of code to its own file.
- Added the empty string `""` option to annotations sorting origin/order, so that the field (and subsequently the other annotation sorting fields) can be cleared.
- Added a `name` field to scanning profiles, to allow users to specify a nickname for saved profiles. This is displayed on the main scanning page so that users are aware which profile is currently loaded.

profile


acrellin:
@stefanv Mind taking a quick look when you have a minute?

stefanv:
Looks good to me.  Only thing I'm curious about is the random IDs, and why they are generated?

static/js/components/CandidatesPreferencesForm.jsx
```diff
@@ -0,0 +1,637 @@
...
+    dispatch(profileActions.updateUserPreferences(prefs));
+
+    if (addOrEdit === "Edit") {
+      // If we just edited the selected profile, let the
+      // parent component know we updated some fields
+      if (selectedScanningProfile?.id === data.id) {
+        setSelectedScanningProfile(data);
+      }
+      closeDialog();
+    } else if (addOrEdit === "Add") {
```
kmshin1397, line 269: @stefanv The randomly generated ids are used basically for internal identification of selected profiles, like here. They're mainly a holdover from the initial version of scanning profiles when I didn't ask the user for a name for the profile. I guess that could now all be replaced by checks against profile names if you'd prefer.

stefanv: If the profile name is required and is unique, that would make sense.  I'll leave this to your best design judgment.

dmitryduev:
@kmshin1397 @stefanv ping. Can we merge this?

kmshin1397:
> @kmshin1397 @stefanv ping. Can we merge this?

Yeah, it's good to go from my POV. Just waiting on Stefan to look at how I addressed his comments,

dmitryduev:
@stefanv does this look good for you now? Can we merge it?

dmitryduev:
Got a verbal approval from @stefanv + failures seem unrelated, so merging.

PR #2064, opened 2021-06-09
Add saved by user to sources table

kmshin1397:
Per Dan Perley via beta-testing channel:

> Would it be possible to make the user ID of the person who saved the transient available as an optional column to display under table of sources in "groups"?

I figured that the username would be more informative on the UI than the user ID, so I chose to display the username of the user who saved the Obj to a specific source instead. On the Sources page, instead of `Saved To Group By`, this column is made instead into a `Last Saved By` username.

savedby

### DIFF
--- a/static/js/components/SourceTable.jsx
+++ b/static/js/components/SourceTable.jsx
@@ -666,6 +666,24 @@ const SourceTable = ({
     );
   };
 
+  const getSavedBy = (source) => {
+    // Get the user who saved the source to the specified group
+    if (groupID !== undefined) {
+      const group = source.groups.find((g) => g.id === groupID);
+      return group?.saved_by.username;
+    }
+    // Otherwise, get whoever saved it last
+    const usernames = source.groups
+      .sort((g1, g2) => (g1.saved_at < g2.saved_at ? -1 : 1))
+      .map((g) => g.saved_by.username);
+    return usernames[usernames.length - 1];
+  };
+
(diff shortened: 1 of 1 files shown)


static/js/components/SourceTable.jsx
```diff
@@ -666,6 +666,24 @@ const SourceTable = ({
     );
   };
 
+  const getSavedBy = (source) => {
+    // Get the user who saved the source to the specified group
+    if (groupID !== undefined) {
+      const group = source.groups.find((g) => g.id === groupID);
+      return group?.saved_by.username;
+    }
+    // Otherwise, get whoever saved it last
```
acrellin, line 675: When would we ever run into this?

guynir42: I think this happens when loading the general sources page, rather than the group sources page. They both use `SourceTable`.

guynir42 (approved):
LGTM

static/js/components/SourceTable.jsx
kmshin1397: Right yeah, the `groupID` here is the param passed into the `SourceTable` based on whether it's on the group sources page and thus has a specific group associated for the data displayed, or the general sources page

acrellin: Ah yeah, I forgot this is used for both GroupSources and general Sources pages

PR #2069, opened 2021-06-11
skymap tables only

mcoughlin:
This PR adds the tables and ingestion API's for skymaps.


mcoughlin:
Test suite failures unrelated. Need help understanding the docs.

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            raise Exception("xml file is not valid VOEvent")
+
+        dateobs = get_dateobs(root)
+
+        try:
+            event = GcnEvent.query.filter_by(dateobs=dateobs).one()
+        except NoResultFound:
+            event = DBSession().merge(GcnEvent(dateobs=dateobs))
+            DBSession().commit()
```
kmshin1397, line 60: Is there a reason you call `DBSession().commit()` throughout the handler logic? We generally try to put off committing until the end of a handler so that we don't flush any changes to the database before the entire logic of the handler is successful, so that we don't have scenarios when for example, the `GcnEvent` is pushed to the database but then the app goes down before the corresponding tags get added so you have end up with a partial update of the database state.

skyportal/tests/api/test_gcn.py
```diff
@@ -0,0 +1,48 @@
...
+        payload = fid.read()
+    data = {'xml': payload}
+
+    status, data = api('PUT', 'gcn/upload', data=data, token=super_admin_token)
+    assert status == 200
+    assert data['status'] == 'success'
+
+    dateobs = Time("2019-04-25 08:18:05", format='iso').datetime
+
+    event = GcnEvent.query.filter_by(dateobs=dateobs).one()
```
kmshin1397, line 22: Instead of querying the database directly to make sure PUTs are successful, the preferred approach is to use the corresponding GET handlers to check these assertions. That way we test those GET handlers as well, and keep the test code separate from underlying database logic by interacting with the test SkyPortal only through the API

skyportal/utils/gcn.py
```diff
@@ -0,0 +1,254 @@
...
+
+    search = root.find("./What/Param[@name='Search']")
+    if search is not None:
+        yield search.attrib['value']
+
+
+def get_skymap(root, gcn_notice):
+    mission = urlparse(root.attrib['ivorn']).path.lstrip('/')
+
+    # Try Fermi GBM convention
```
kmshin1397, line 108: Is there a particular order in which these different conventions should be tested? Obviously I know nothing about the conventions or the format of the skymaps, but just wanted to make sure there weren't potentially cases where you would want to test for LVC before Fermi or whatever, since it's rare programmatically to have a string of `if` cases like this without any `else if`s

skyportal/utils/gcn.py
```diff
@@ -0,0 +1,254 @@
...
+    radius = error * u.deg
+
+    # Determine resolution such that there are at least
+    # 16 pixels across the error radius.
+    hpx = HEALPix(
+        pixel_resolution_to_nside(radius / 16, round='up'), 'nested', frame=ICRS()
+    )
+
+    # Find all pixels in the 4-sigma error circle.
+    ipix = hpx.cone_search_skycoord(center, 4 * radius)
```
kmshin1397, line 173: Are values like the `16 pixels across the error radius` and `4-sigma error circle` standards that are unlikely to change? Just wondering if it would be better to make these configurable via the config.yaml files.

skyportal/utils/gcn.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def get_contour(localization):
+
+    # Calculate credible levels.
+    prob = localization.flat_2d
+    cls = 100 * ligo.skymap.postprocess.find_greedy_credible_levels(prob)
+
+    # Construct contours and return as a GeoJSON feature collection.
+    levels = [50, 90]
```
kmshin1397, line 229: Same question regarding configurability

skyportal/models.py
```diff
@@ -3448,6 +3454,334 @@ class SourceNotification(Base):
     level = sa.Column(sa.String(), nullable=False)
 
 
+class GcnNotice(Base):
```
kmshin1397, line 3457: Another thing that we should probably think about when it comes to these model changes is specifying access control rules. That is, specifying which users have permissions to read these records, who's allowed to delete or update them, etc. You can read more about how to do it here: https://skyportal.io/docs/permissions.html. But yeah, generally I'm looking for things like `delete = restricted` and `read = public`, stuff like that.

kmshin1397 (commented):
@mcoughlin good to see this green and coming along! Some further food for thought for you :)

skyportal/handlers/api/gcn.py
mcoughlin: Ok I think I still need one after Localization is pushed otherwise the query for it fails (because the get_contour is done separately).

skyportal/tests/api/test_gcn.py
mcoughlin: Yes definitely. Fixed.

skyportal/utils/gcn.py
mcoughlin: No they are all separate (LVC, Fermi, neutrino, etc.). So I think its all the same.

skyportal/utils/gcn.py
mcoughlin: Yeah I don't think so. We haven't changed that since Fermi program started at start of ZTF I.

skyportal/utils/gcn.py
mcoughlin: Same as above. Never say never, but I think its ok if that ends up being a small PR down the line.

skyportal/models.py
mcoughlin: Yes I added what I thought made sense. Basically I want these all public, but maybe best not to have just anyone delete them. Can you confirm that I got it right for that?

package.json
```diff
@@ -16,7 +16,7 @@
     "@rjsf/core": "2.4.2",
     "@rjsf/material-ui": "2.4.0",
     "autosuggest-highlight": "3.1.1",
-    "check-dependencies": "1.1.0",
+    "check-dependencies": "^1.1.0",
```
acrellin, line 19:
```suggestion
    "check-dependencies": "1.1.0",
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+
+_, cfg = load_env()
+
+
+class GcnHandler(BaseHandler):
+    @auth_or_token
+    def put(self):
+        """
+        ---
+        description: Ingest GCN xml file
```
acrellin, line 33: We should document here the required top-level JSON body contents (i.e. the `"xml"` parameter)

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+            root = lxml.etree.fromstring(payload.encode('ascii'))
+        else:
+            raise Exception("xml file is not valid VOEvent")
+
+        dateobs = get_dateobs(root)
+
+        try:
+            event = GcnEvent.query.filter_by(dateobs=dateobs).one()
+        except NoResultFound:
+            event = DBSession().merge(GcnEvent(dateobs=dateobs))
```
acrellin, line 59: This must be why you're getting the missing `id` column error -- if you `add` the new record rather than using `merge`, it should create the `id` column as expected.

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+            content=payload.encode('ascii'),
+            ivorn=root.attrib['ivorn'],
+            notice_type=gcn.get_notice_type(root),
+            stream=urlparse(root.attrib['ivorn']).path.lstrip('/'),
+            date=root.find('./Who/Date').text,
+            dateobs=event.dateobs,
+        )
+
+        for tag in tags:
+            DBSession().merge(tag)
```
acrellin, line 73:
```suggestion
            DBSession().add(tag)
```
for same reason as above

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        DBSession().merge(gcn_notice)
```
acrellin, line 74:
```suggestion
        DBSession().add(gcn_notice)
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+
+        skymap = get_skymap(root, gcn_notice)
+        skymap["dateobs"] = event.dateobs
+
+        localization = Localization.query.filter_by(
+            dateobs=dateobs, localization_name=skymap["localization_name"]
+        ).all()
+        if len(localization) == 0:
+            DBSession().merge(Localization(**skymap))
```
acrellin, line 83:
```suggestion
            DBSession().add(Localization(**skymap))
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+            DBSession().commit()
```
acrellin, line 84:
```suggestion
```
commit once at the very end before returning (ideally using `self.verify_and_commit()`)

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        localization = Localization.query.filter_by(
+            dateobs=dateobs, localization_name=skymap["localization_name"]
+        ).one()
+
+        localization = get_contour(localization)
+        DBSession().merge(localization)
+        DBSession().commit()
```
acrellin, line 91:
```suggestion
        self.verify_and_commit()
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+            if len(events) >= max_num_events:
+                continue
+            dateobs = event.dateobs
+            if dateobs < cutoff_day:
+                continue
+            tags = [_ for _ in event.tags]
+            localizations = [_.localization_name for _ in event.localizations]
+            events.append(
+                {'localizations': localizations, 'dateobs': dateobs, 'tags': tags}
+            )
```
acrellin, line 123: prefer to simplify logic here by using `query.filter` and `query.limit`, which should greatly compactify this section

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        q = GcnEvent.query
+
+        events = []
+        for event in q.all():
+            if len(events) >= max_num_events:
+                continue
+            dateobs = event.dateobs
+            if dateobs < cutoff_day:
+                continue
+            tags = [_ for _ in event.tags]
```
acrellin, line 119:
```suggestion
            tags = [tag for tag in event.tags]
```
(`_` usually indicates that the variable goes unused)

acrellin: also, why not just `tags = event.tags`?

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+            localizations = [_.localization_name for _ in event.localizations]
```
acrellin, line 120:
```suggestion
            localizations = [l.localization_name for l in event.localizations]
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+            raise Exception("xml file is not valid VOEvent")
+
+        dateobs = get_dateobs(root)
+
+        try:
+            event = GcnEvent.query.filter_by(dateobs=dateobs).one()
+        except NoResultFound:
+            event = DBSession().merge(GcnEvent(dateobs=dateobs))
+
+        tags = [GcnTag(dateobs=event.dateobs, text=_) for _ in get_tags(root)]
```
acrellin, line 61:
```suggestion
        tags = [GcnTag(dateobs=event.dateobs, text=text) for text in get_tags(root)]
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+            )
+
+        return self.success(data=events)
+
+
+class GcnEventHandler(BaseHandler):
+    @auth_or_token
+    def get(self, dateobs):
+        q = GcnEvent.query.filter_by(dateobs=dateobs)
+        event = q.first()
```
acrellin, line 132:
```suggestion
        event = GcnEvent.query.filter_by(dateobs=dateobs).first()
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        tags = [_ for _ in event.tags]
```
acrellin, line 133:
```suggestion
        tags = [tag for tag in event.tags]
```
or just `tags = event.tags`

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        localizations = [_.localization_name for _ in event.localizations]
```
acrellin, line 134: no underscore here and below

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        notices = [_.content for _ in event.gcn_notices]
+        data = [
+            {
+                'tags': tags,
+                'dateobs': event.dateobs,
+                'localizations': localizations,
+                'lightcurve': event.lightcurve,
+                'notices': notices,
+            }
+        ]
```
acrellin, line 144: Why not just jsonify the record instance?

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        return self.success(data=data)
+
+
+class LocalizationHandler(BaseHandler):
+    @auth_or_token
+    def get(self, dateobs, localization_name):
+        q = Localization.query.filter_by(
+            dateobs=dateobs, localization_name=localization_name
+        )
+        localization = q.first()
```
acrellin, line 154:
```suggestion
        localization = Localization.query.filter_by(
            dateobs=dateobs, localization_name=localization_name
        ).first()
```

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+
+        data = {
+            'flat_2d': localization.flat_2d,
+            'contour': localization.contour,
+            'dateobs': localization.dateobs,
+            'localization_name': localization.localization_name,
+        }
```
acrellin, line 161: Same question here: why not just pass the record instance to `success` and let the `to_json` call there do all this for you?

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+        return self.success(data=data)
+
+
+class ZTFFieldsHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        contours = []
+        import numpy as np
+        import copy
+        import gwemopt.utils
```
acrellin, line 172: Why are these imports in the method definition rather than at the top of the file?

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+    """Records of ingested GCN notices"""
+
+    create = read = public
```
acrellin, line 3460:
```suggestion
```
(these are the default policies already)

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+    update = delete = AccessibleIfUserMatches('sent_by')
+
+    id = sa.Column(sa.Integer, primary_key=True, autoincrement=True)
```
acrellin, line 3463:
```suggestion
```
If you use `DBSession().add()` rather than `merge()` for adding new records, you shouldn't need to redefine this here

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+    def noise(self):
+        return self._get_property(property_name="Terrestrial")
+
+
+class Localization(Base):
+    """Localization information, including the localization ID, event ID, right
+    ascension, declination, error radius (if applicable), and the healpix
+    map."""
+
+    create = read = public
```
acrellin, line 3535:
```suggestion
```

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+    update = delete = AccessibleIfUserMatches('sent_by')
+
+    nside = 512
+    # HEALPix resolution used for flat (non-multiresolution) operations.
+
+    id = sa.Column(sa.Integer, primary_key=True, autoincrement=True)
```
acrellin, line 3541:
```suggestion
```

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+            result = t['PROB'], t['DISTMU'], t['DISTSIGMA'], t['DISTNORM']
+            return hp.reorder(result, 'NESTED', 'RING')
+        else:
+            return (self.flat_2d,)
+
+
+class GcnTag(Base):
+    """Store qualitative tags for events."""
+
+    create = read = public
```
acrellin, line 3652:
```suggestion
```

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+    update = delete = AccessibleIfUserMatches('sent_by')
+
+    id = sa.Column(sa.Integer, primary_key=True, autoincrement=True)
```
acrellin, line 3655:
```suggestion
```

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+        primary_key=True,
+    )
+
+    text = sa.Column(sa.Unicode, nullable=False, primary_key=True)
+
+
+class GcnEvent(Base):
+    """Event information, including an event ID, mission, and time of the event."""
+
+    create = read = public
```
acrellin, line 3670:
```suggestion
```

skyportal/models.py
```diff
@@ -3448,6 +3454,346 @@ class SourceNotification(Base):
...
+    update = delete = AccessibleIfUserMatches('sent_by')
+
+    id = sa.Column(sa.Integer, primary_key=True, autoincrement=True)
```
acrellin, line 3673:
```suggestion
```

skyportal/services/gcn_service/gcn_service.py
```diff
@@ -0,0 +1,50 @@
...
+    gcn.NoticeType.LVC_INITIAL,
+    gcn.NoticeType.LVC_UPDATE,
+    gcn.NoticeType.LVC_RETRACTION,
+    gcn.NoticeType.LVC_TEST,
+    gcn.NoticeType.AMON_ICECUBE_COINC,
+    gcn.NoticeType.AMON_ICECUBE_HESE,
+    gcn.NoticeType.ICECUBE_ASTROTRACK_GOLD,
+    gcn.NoticeType.ICECUBE_ASTROTRACK_BRONZE,
+)
+def handle(payload, root):
```
acrellin, line 41: where is root used?

skyportal/services/gcn_service/gcn_service.py
```diff
@@ -0,0 +1,50 @@
+import yaml
+
+import gcn
+
+from baselayer.app.env import load_env
+from baselayer.log import make_log
+
+from skyportal.tests import api
+
+env, cfg = load_env()
```
acrellin, line 10:
```suggestion
```
unused (associated import above, too)

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,202 @@
...
+from ...models import (
+    DBSession,
+    GcnEvent,
+    GcnNotice,
+    GcnTag,
+    Localization,
+)
+from ...utils.gcn import get_dateobs, get_tags, get_skymap, get_contour
+
+_, cfg = load_env()
```
acrellin, line 25:
```suggestion
```
appears to be unused (should remove associated import above, too)

acrellin (commented):
Looks good overall. A few suggestions below. Thanks!

skyportal/services/gcn_service/gcn_service.py
mcoughlin: good point! holdover from pre-API!

skyportal/handlers/api/gcn.py
mcoughlin: I went through and added Schema's to schema.py... hopefully this was a sensible choice based on the other examples?

acrellin: Yep, looks great!

mcoughlin: Thanks Ari!

skyportal/models.py
kmshin1397: Yup makes sense. In order for these to work properly, you'll need to define the `sent_by` relationships to the `User` table so that the records actually know who sent that data in, and you'll also need to update the `GCNHandler` to store that information. A simple example you could look to would be the `Listing` table, which defines rules against the `user` that the listing entry is for.

skyportal/models.py
kmshin1397: Ah yeah, that could explain the errors you were seeing with getting rid of the `id` column

kmshin1397 (commented):
Sorry I took a few days, but here's some more feedback @mcoughlin. A few more minor suggestions, and I'd second the comments Ari made as well. Thanks!

skyportal/models.py
mcoughlin: @kmshin1397 This piece you mean?

User.listings = relationship(
    'Listing',
    back_populates='user',
    passive_deletes=True,
    doc='The listings saved by this user',
)

mcoughlin:
@acrellin flat_2d is a property and does not automatically come from to_json in Localization table looks like.

mcoughlin:
Similarly, I see that event.tags and event.localizations also do not automatically come along.

skyportal/handlers/api/gcn.py
acrellin: Is this commit really needed?

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,218 @@
...
+            ivorn=root.attrib['ivorn'],
+            notice_type=gcn.get_notice_type(root),
+            stream=urlparse(root.attrib['ivorn']).path.lstrip('/'),
+            date=root.find('./Who/Date').text,
+            dateobs=event.dateobs,
+        )
+
+        for tag in tags:
+            DBSession().add(tag)
+            DBSession().commit()
```
acrellin, line 84: Is this commit needed here?

skyportal/handlers/api/gcn.py
acrellin: If it is, I'd suggest replacing it with `self.verify_and_commit()` so that we're properly enforcing permissions throughout the handler body

skyportal/handlers/api/gcn.py
acrellin: If so, replace with `self.verify_and_commit()`

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,218 @@
...
+        user_prefs = getattr(self.current_user, 'preferences', None) or {}
+        top_events_prefs = user_prefs.get('topGcnEvents', {})
+        top_events_prefs = {**default_prefs, **top_events_prefs}
+
+        max_num_events = int(top_events_prefs['maxNumGcnEvents'])
+        since_days_ago = int(top_events_prefs['sinceDaysAgo'])
+
+        cutoff_day = datetime.datetime.now() - datetime.timedelta(days=since_days_ago)
+        q = (
+            GcnEvent.query.options(joinedload(GcnEvent.localizations))
```
acrellin, line 134: This looks great now!

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,218 @@
...
+
+        event = (
+            GcnEvent.query.options(
+                joinedload(GcnEvent.localizations),
+                joinedload(GcnEvent.gcn_notices),
+            )
+            .filter_by(dateobs=dateobs)
+            .first()
+        )
+        data = {**event.to_dict(), "tags": event.tags}
```
acrellin, line 179: Ah, so concise 😀  -- thanks!

skyportal/models.py
```diff
@@ -3448,6 +3454,355 @@ class SourceNotification(Base):
     level = sa.Column(sa.String(), nullable=False)
 
 
+class GcnNotice(Base):
+    """Records of ingested GCN notices"""
+
+    update = delete = AccessibleIfUserMatches('sent_by')
+
+    ivorn = sa.Column(sa.String, doc='Unique identifier of VOEvent')
```
acrellin, line 3462: Sounds like this should be 
```suggestion
    ivorn = sa.Column(sa.String, unique=True, index=True, doc='Unique identifier of VOEvent')
```

acrellin (approved):
Looks great now, I really like how simple and concise the handlers have become. Only request is the noted column that sounds (from the description) like it should have `unique=True`, and to just look over any columns that originally had `primary_key=True` and add `unique=True` and `index=True` wherever relevant.

skyportal/handlers/api/gcn.py
mcoughlin: Thanks to you!

skyportal/handlers/api/gcn.py
mcoughlin: Again, thanks to you ;)

skyportal/handlers/api/gcn.py
mcoughlin: I found without the commit, calls like ```event = GcnEvent.query.filter_by(dateobs=dateobs).one()''' fail.

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,218 @@
...
+            ivorn=root.attrib['ivorn'],
+            notice_type=gcn.get_notice_type(root),
+            stream=urlparse(root.attrib['ivorn']).path.lstrip('/'),
+            date=root.find('./Who/Date').text,
+            dateobs=event.dateobs,
+        )
+
+        for tag in tags:
+            DBSession().add(tag)
+            self.verify_and_commit()
```
kmshin1397, line 84: Can we remove this `self.verify_and_commit()` call? It doesn't seem like you query for these tags again, which was the main reason for these mid-handler commits, right?

skyportal/handlers/api/gcn.py
```diff
@@ -0,0 +1,218 @@
...
+        dateobs = get_dateobs(root)
+
+        try:
+            event = GcnEvent.query.filter_by(dateobs=dateobs).one()
+        except NoResultFound:
+            DBSession().add(
+                GcnEvent(dateobs=dateobs, sent_by_id=self.associated_user_object.id)
+            )
+            self.verify_and_commit()
+            event = GcnEvent.query.filter_by(dateobs=dateobs).one()
```
kmshin1397, line 69:
```suggestion
            event = GcnEvent(dateobs=dateobs, sent_by_id=self.associated_user_object.id)
            DBSession().add(event)
```
This is the pattern I used to get rid of the early commit with the localization below, and I think can be applied here so we're not committing mid-handler. Sorry to bring it up again haha, but I would really prefer if we could just commit once at the end if we can- it simplifies rolling back transactions if anything goes wrong a lot :)

kmshin1397 (commented):
Sorry, not quite an approval from me yet though we're almost there 😅 . Besides the inline comments, I would also:
- Add the `sent_by` relationship to the other model classes not `GcnEvent` so that those also have working delete/update permissions.
- Write some tests for those permissions.  Simple tests like posting an event and then having another user try to delete it and getting back the `Insufficient permissions` error. If you want examples, there's a bunch of these permission tests in `tests/api/test_groups.py`

skyportal/handlers/api/gcn.py
mcoughlin: Yes!

kmshin1397 (commented):
Almost there! There were some minor tweaks with the permissions stuff that sound scarier to explain than to actually change, so I just went ahead and made it into a PR to your PR branch here: https://github.com/mcoughlin/skyportal/pull/28

I'll be happy to finally approve this once you look over that and merge it in if it makes sense to you :)

mcoughlin:
Thanks @kmshin1397 !

kmshin1397 (approved):
LGTM! @acrellin do you want to look over this one more time and merge if you're happy. We've had a few changes since you approved.

acrellin (approved):
Good call on the new additions, @kmshin1397 -- all looks good! Will merge when green

PR #2071, opened 2021-06-16
Generate front-end notifications for selected activity on favorite sources

acrellin:
This PR introduces an event listener that adds front-end user notifications (when enabled) for activity on users' favorite sources (can be configured to trigger on new comments, spectra and classifications for now -- and can be easily expanded). A generalized SQLAlchemy event listener is used for all three mappers. A new notification settings section is added to the user profile. 

The generated notifications:

notif

On the profile page:

prof

Closes https://github.com/skyportal/skyportal/issues/1619

### DIFF
--- a/skyportal/models.py
+++ b/skyportal/models.py
@@ -3760,6 +3760,8 @@ StreamUser.delete = restricted & CustomUserAccessControl(
 )
 
 
+@event.listens_for(Classification, 'after_insert')
+@event.listens_for(Spectrum, 'after_insert')
 @event.listens_for(Comment, 'after_insert')
 def add_user_notifications(mapper, connection, target):
 
@@ -3775,20 +3777,24 @@ def add_user_notifications(mapper, connection, target):
         users = (
             User.query.join(listing_subquery, User.id == listing_subquery.c.user_id)
             .filter(
-                User.preferences["favorite_sources_activity_notifications"]["comments"]
+                User.preferences["favorite_sources_activity_notifications"][
+                    target.__tablename__
+                ]
                 .astext.cast(sa.Boolean)
                 .is_(True)
             )
             .all()
         )
         for user in users:
-            session.add(
-                UserNotification(
-                    user=user,
-                    text=f"*{target.author.username}* commented on your favorite source *{target.obj_id}*",
-                    url=f"/source/{target.obj_id}",
+            # Only notify users who have read access to the new record in question
+            if target.__class__.get_if_accessible_by(target.id, user) is not None:
+                session.add(
+                    UserNotification(
+                        user=user,
+                        text=f"New *{target.__class__.__name__}* on your favorite source *{target.obj_id}*",
+                        url=f"/source/{target.obj_id}",
+                    )
                 )
-            )
 
 
 schema.setup_schema()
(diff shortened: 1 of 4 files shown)


kmshin1397 (commented):
Looks good @acrellin! I like the idea of using events here! 

I wonder if we may run into performance issues if we end up with a lot of users with big favorites lists that enable the preference. My instinct is that since the classification/spectrum/comment tables are a lot smaller than something like photometry this should be okay. But if it's not too much trouble maybe you could do some performance testing and see if it turns up anything? Create some test users with the notifications enabled and try adding some comments or something and compare the insertion times? Doesn't have to be on the scale of the production database or anything like that, but it would be good to look at some rough numbers I think.

acrellin:
On master: 
``` python
In [12]: DBSession.add(Comment(text="test", author_id=1, obj_id="14gqr"))

In [13]: %time DBSession.commit()                    
CPU times: user 4.81 ms, sys: 0 ns, total: 4.81 ms
Wall time: 40.4 ms
```
On this branch, with 100 test users having added the source in question to their favorites list (probably more than we'll ever encounter in production):
``` python
In [6]: DBSession.add(Comment(text="test", author_id=1, obj_id="14gqr"))

In [7]: %time DBSession.commit()
CPU times: user 371 ms, sys: 19.8 ms, total: 391 ms
Wall time: 468 ms
```
So yeah, definitely slows down the inserts a bit, but as these aren't resources that will be programmatically added in large numbers, the performance seems reasonable to me here, especially considering it's hard to imagine 100 users all having the same source on their favorites list and all having these types of notifications enabled (but even if so, this performance still seems acceptable considering the intermittent nature & low frequency of adding spectra, comments, and classifications). Thoughts @kmshin1397 ?

kmshin1397:
Hmm, I would agree with that when it comes to comments and classifications. But my impressions is that spectra are often added programmatically within a short time frame, at least when it comes to SEDM spectra. I think @scizen9 has scripts which do that sort of thing, if I recall correctly. Is that true @scizen9? You upload spectra through scripts, right? Do those tend to be a lot of spectra at once? How would you feel about a increase in time per spectrum post?

acrellin:
@kmshin1397 @scizen9 As the overhead here is due to inserting new `UserNotification` rows, the overhead should be a linearly-scaling ~400ms per 100 users who are watching the source in question and have enabled the relevant notification types. Does that sound right?

acrellin:
Here's with 100 users watching the obj in question, but without having enabled the relevant notifications: 
``` python
In [7]: %time DBSession.commit()                    
CPU times: user 11.1 ms, sys: 614 µs, total: 11.7 ms
Wall time: 14.2 ms
```

acrellin:
Updated notification text styling: 
notif

guynir42:
This looks good overall. Can you remind me what is the difference between "mark as read" and "delete" in this context?

guynir42 (approved):

acrellin:
> This looks good overall. Can you remind me what is the difference between "mark as read" and "delete" in this context?

That's just the preexisting functionality for the browser notifications - they can be marked read / unread and deleted.

kmshin1397 (approved):

kmshin1397:
Note: approved since while spectra are posted programmatically, they're done so fairly sparsely so the performance hit shouldn't really matter

PR #2072, opened 2021-06-17
Do not unnecessarily fetch favorite sources

acrellin:
This PR addresses a potentially significant resource drain whereby any time a user updated one of their listings, their favorite sources (not just the `Listing`s, but the actual `Obj` data) were re-fetched. The issue was two-fold: the REFRESH_FAVORITE_SOURCES WS message was being pushed regardless of whether the relevant list was updated; and when the front-end received that message, it was always issuing a new (potentially expensive) request, regardless of whether the user was viewing the favorites page, where that information would be relevant.

### DIFF
--- a/static/js/ducks/sources.js
+++ b/static/js/ducks/sources.js
@@ -79,7 +79,9 @@ const initialState = {
 // Websocket message handler
 messageHandler.add((actionType, payload, dispatch) => {
   if (actionType === REFRESH_FAVORITE_SOURCES) {
-    dispatch(fetchFavoriteSources());
+    if (window.location.pathname === "/favorites") {
+      dispatch(fetchFavoriteSources());
+    }
   }
 });
 
--- a/skyportal/handlers/api/user_obj_list.py
+++ b/skyportal/handlers/api/user_obj_list.py
@@ -158,9 +158,11 @@ class UserObjListHandler(BaseHandler):
         DBSession().add(listing)
         self.verify_and_commit()
 
-        self.push(action='skyportal/REFRESH_FAVORITES')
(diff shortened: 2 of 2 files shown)


kmshin1397 (approved):
Good catch, thanks Ari

static/js/ducks/sources.js
```diff
@@ -79,7 +79,9 @@ const initialState = {
 // Websocket message handler
 messageHandler.add((actionType, payload, dispatch) => {
   if (actionType === REFRESH_FAVORITE_SOURCES) {
-    dispatch(fetchFavoriteSources());
+    if (window.location.pathname === "/favorites") {
```
stefanv, line 82: I think this is fine; may be worth briefly looking at React Router (e.g. https://reactrouter.com/web/api/Hooks/uselocation) to see which is best to use.

stefanv (approved):

static/js/ducks/sources.js
acrellin: I think that can only be used inside React components

guynir42:
I didn't have a chance to look at this, but maybe there is a need to refresh the star icon even when not on the favorites page?

acrellin:
@guynir42 The star icon state is determined by the `favorites` state branch, which isn't affected by limiting the requests to `/api/sources` which corresponds to the `sources` state branch and is used by the favorite sources page.

acrellin:
Merging this one so we can redeploy

guynir42:
> @guynir42 The star icon state is determined by the `favorites` state branch, which isn't affected by limiting the requests to `/api/sources` which corresponds to the `sources` state branch and is used by the favorite sources page.

Yeah I forgot about that :)

PR #2073, opened 2021-06-17
Use float priorities for SEDM and LT followup

### LINKED ISSUE #603: Decide on granularity for followup request priorities
Why isn't this numerical? And does it have to be enumerated? Does it make sense to have a priority 1.5, which would slot it between 1 and 2?

_Originally posted by @profjsb in https://github.com/skyportal/skyportal/pull/574#discussion_r456685645_

This issue serves as a forum to determine the right way to enumerate / define follow-up priorities, continuing the conversation started by @profjsb above.

kmshin1397:
This PR converts the `priority` follow-up request form field for the SEDM and LT APIs to be a float value in the `[1, 5]` range instead of an enumerated list of strings representing the integers in that range, to allow for more granularity in the priorities.

Closes #603

### DIFF
--- a/skyportal/facility_apis/lt.py
+++ b/skyportal/facility_apis/lt.py
@@ -528,9 +528,10 @@ class IOOAPI(LTAPI):
             },
             "binning": {"type": "string", "enum": ["1x1", "2x2"], "default": "1x1"},
             "priority": {
-                "type": "string",
-                "enum": ["1", "2", "3", "4", "5"],
-                "default": "1",
+                "type": "number",
+                "default": 1.0,
+                "minimum": 1,
+                "maximum": 5,
                 "title": "Priority",
             },
         },
--- a/skyportal/facility_apis/sedm.py
+++ b/skyportal/facility_apis/sedm.py
@@ -285,9 +285,10 @@ class SEDMAPI(FollowUpAPI):
                 "default": "IFU",
             },
             "priority": {
-                "type": "string",
-                "enum": ["1", "2", "3", "4", "5"],
-                "default": "1",
+                "type": "number",
+                "default": 1.0,
+                "minimum": 1,
+                "maximum": 5,
                 "title": "Priority",
             },
             "start_date": {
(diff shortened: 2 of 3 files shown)


mcoughlin (approved):
Seems straightforward enough. Thanks @kmshin1397.

acrellin (changes requested):
Looks like we'll need to update https://github.com/skyportal/skyportal/blob/master/skyportal/models.py#L3300-L3304 (and convert existing values in a migration script)

kmshin1397:
> Looks like we'll need to update https://github.com/skyportal/skyportal/blob/master/skyportal/models.py#L3300-L3304 (and convert existing values in a migration script)

@acrellin That column is associated with `ClassicalAssignment` records, which I believe are separate from the robotic `FollowupRequest`s that this PR pertains to. I'm not clear on what that means exactly, but I'm guessing they pertain to entire different instruments with different procedures for triggering additional data collection. I don't think we need to change those to be float priorities as well, but I could be wrong of course. Maybe @mcoughlin can shed some light

acrellin:
@kmshin1397 Yeah, I had to trace down what means what, too. It looks like `FollowupRequestForm` posts to `/api/followup_requests`, which actually creates a `ClassicalAssignment` instance. I agree this is all very confusing (Danny did all that)

acrellin:
Ah, nm! I was looking at a different handler inside the `followup_requests` module

mcoughlin:
@kmshin1397 @acrellin Yeah I suppose Robotic and Classical Observations *could* have been made separate tables from the beginning, as maybe they function slightly differently. Anyways, I can't imagine why moving from integer (or string?) to floating point in that table would be an issue.

mcoughlin:
@acrellin ah ok, so not a problem!

acrellin:
@kmshin1397 but where do these priorities even get used, then? I don't see a corresponding column in the `followup_requests` table

mcoughlin:
@acrellin They typically go directly to the scheduler: https://github.com/skyportal/skyportal/blob/master/skyportal/facility_apis/lco.py#L100

kmshin1397:
> @kmshin1397 but where do these priorities even get used, then? I don't see a corresponding column in the `followup_requests` table. Are they just used in the request to the remote facility?

Yeah, I think we just store those within the `payload` JSONB field on our side. They get passed along to the remote facilities, and I assume they keep track of those in their own systems

acrellin (approved):
Sorry for the noise!

PR #2080, opened 2021-06-22
Push FETCH_NOTIFICATIONS WS message when new front-end notification is added in event listener

acrellin:


acrellin:
Thanks for the tip on this one, @stefanv!

skyportal/models.py
```diff
@@ -86,6 +87,8 @@
 _, cfg = load_env()
 cosmo = establish_cosmology(cfg)
 
+ws_flow = Flow()
```
acrellin, line 90: @stefanv Can there be more than one `Flow` instance in the app like this, or do I need to import the instance created in `baselayer`?

acrellin:
Failure unrelated: 
```
FAILED skyportal/tests/frontend/test_followup_requests.py::test_submit_new_followup_request_SEDM
= 1 failed, 447 passed, 3 xfailed, 1 xpassed, 219 warnings in 2979.70s (0:49:39) =
```

stefanv (approved):

stefanv:
The only thing I am not 100% sure about is what happens if the connection gets interrupted.  You instantiate `flow` on the module level, so it is fairly long lasting.  It may be worth instantiating "on-the-fly" if this is a problem.

You can test by bringing down the websocket microservice after the app is running, bringing it back up, and seeing if notifications still go through.

acrellin:
@stefanv I just updated to instantiate `Flow` inside the event listener just to be on the safer side.

kmshin1397 (approved):

PR #2081, opened 2021-06-22
Duplicate sources flag on source page

acrellin:
This PR introduces a check for duplicates in the handler for fetching individual sources by ID. If any duplicates are found, an array of their IDs is included in the response. 

On the front-end source page, if any duplicates exist, a message appears indicating so with links to all possible duplicate sources.

dup_red_no_italics

Partially addresses https://github.com/fritz-marshal/fritz/issues/178

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -660,11 +660,10 @@ class SourceHandler(BaseHandler):
                 .all()
             )
             point = ha.Point(ra=s.ra, dec=s.dec)
-            # Check for duplicates (within 4 arcsecs, magnitude 18 or brighter)
+            # Check for duplicates (within 4 arcsecs)
             duplicates = (
                 Obj.query_records_accessible_by(self.current_user)
                 .filter(Obj.within(point, 4 / 3600))
-                .filter(Obj.last_detected_mag(self.current_user) <= 18)
                 .filter(Obj.id != s.id)
                 .all()
             )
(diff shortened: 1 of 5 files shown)


mcoughlin:
@acrellin What fraction of your sources does this check get?

acrellin:
@mcoughlin I just updated it from mag 15 (catches only two duplicates per 10k sources) to mag 18 (93 per 10k)

mcoughlin (approved):
Looks great post that update @acrellin.

acrellin:
@mcoughlin It's pushed

acrellin:
@mcoughlin What's the rationale for filtering out fainter objects?

mcoughlin:
@acrellin My vision was that these were dominated by bright star PSFs moving the centroid of the subtractions around. But given these numbers, it seems clear that there wont be too many false positives.

acrellin:
@mcoughlin So is there any reason to keep the minimum brightness filtering here?

mcoughlin:
@acrellin no I think probably fine to remove, although I expect it to not change the numbers based on the hypothesis.

acrellin:
It does change the numbers quite a bit -- with `mag <= 18` we catch 93 duplicates for 10k sources, and when removing that criterion it jumps to 301

mcoughlin:
@acrellin In that case, can we look at the faint examples? I think that seems more likely to be truly different source for "normal" stars, where I would think that then we should have a criteria closer to our PSF limit (1 arcsecond).

acrellin:
Removed magnitude filtering per a discussion w/ @mcoughlin

kmshin1397 (approved):
@acrellin Looks like test failure may be related but otherwise looks good

acrellin:
Green now, with two approvals -- merging

PR #2082, opened 2021-06-23
Mock-up of Top GCN Events Widget

mcoughlin:
(very) basic mock-up of Top GCN Events Widget as discussed on our call today. Would be nice to understand how to prettify it ;)

top_events


mcoughlin:
@kmshin1397 @acrellin As I said, would be nice to make it prettier, but looking for suggestions / general feedback.

acrellin:
Looks good! How do you want it to look?

mcoughlin:
@acrellin I think the buttons could be cooler. Could also benefit from the XXX hours ago perhaps. This is how the existing marshal looks.
image

acrellin:
How about making the links MUI chips? https://material-ui.com/components/chips/

mcoughlin:
Looks cool, thanks @acrellin!

acrellin:
If they still need to be links, just wrap the `Chip` tags in `Link` tags (w/ same params as before)

mcoughlin:
Good point, thanks @acrellin. Suggestions on making the Buttons sit next to each other? And on the (XXX hours ago) type of formatting? I see TopSources.jsx has something like it, but I would have thought there would have simply been a line like: https://github.com/growth-astro/growth-too-marshal/blob/main/growth/too/templates/event_header.html#L8
to compute it... Maybe better would simply to compute it as a string in gcn.py?

kmshin1397:
> Good point, thanks @acrellin. Suggestions on making the Buttons sit next to each other? And on the (XXX hours ago) type of formatting? I see TopSources.jsx has something like it, but I would have thought there would have simply been a line like: https://github.com/growth-astro/growth-too-marshal/blob/main/growth/too/templates/event_header.html#L8
> to compute it... Maybe better would simply to compute it as a string in gcn.py?

@mcoughlin I think we use dayjs to get that sort of time formatting. Something like this line: https://github.com/skyportal/skyportal/blob/8135c5b7aeebfdbe32e7c34c3b1bbfe75a77b8ae/static/js/components/RecentSources.jsx#L179

kmshin1397:
As for getting them to sit in a line, I would probably look to putting them in a div with `display: flex`. If you're not familiar with flexbox styling, here's a quick primer: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

mcoughlin:
@kmshin1397 Ok that's good progress, thanks. Is it possible to make buttons different colors than primary and secondary? The internet confuses me. I have like the below (which corresponds to the colors from the old marshal), but I am confused how to use it.

const getColor = (tag) => ({
  BNS: {
    default: "#468847",
  },
  NSBH: {
    default: "#b94a48",
  },
  BBH: {
    default: "#333333",
  },
  GRB: {
    default: "#f89406",
  },
  AMON: {
    default: "#3a87ad",
  },
  Terrestrial: {
    default: "#999999",
  },
});

kmshin1397:
> @kmshin1397 Ok that's good progress, thanks. Is it possible to make buttons different colors than primary and secondary? The internet confuses me. I have like the below (which corresponds to the colors from the old marshal), but I am confused how to use it.
> 
> const getColor = (tag) => ({
> BNS: {
> default: "#468847",
> },
> NSBH: {
> default: "#b94a48",
> },
> BBH: {
> default: "#333333",
> },
> GRB: {
> default: "#f89406",
> },
> AMON: {
> default: "#3a87ad",
> },
> Terrestrial: {
> default: "#999999",
> },
> });

Hi @mcoughlin looks like this is possible but somewhat annoying to do. I found this, which I think should work: https://github.com/mui-org/material-ui/issues/21135#issuecomment-861610819. Basically, you write a wrapper around the Material UI Chip component which takes arbitrary color codes, and then you'll have to pass those in when you render the list of chips.

kmshin1397:
You can probably do something similar if you're looking to customize `Button` components and not `Chip` components. You can also manually style them one-by-one like this: https://codesandbox.io/s/rj34pyo75p?file=/demo.js in the `makeStyles` definition, but that would prob lead to a really long list of style rules which may be more ugly than the wrapper component method

mcoughlin:
@kmshin1397 nice. Can you help me understand how I actually access those different types? I tried: <StyledButton variant={tag} key={tag}> but they are all still red and I think there is probably some kind of `classes.${tag}` call I am supposed to make, but unsure.

kmshin1397:
@mcoughlin See https://github.com/mcoughlin/skyportal/pull/31

mcoughlin:
@acrellin @kmshin1397 To me, this is now beautiful. What more in your mind must be done before we could have it go in? I know @acrellin was looking at making a table for a separate page of these events, but that is separate I think.

kmshin1397:
As I mentioned earlier, I think the one big thing is that these aren't actually showing the top events in terms of views. We aren't keeping track of how many times an event has been viewed at all (which makes sense since we don't have an event page to view just yet). To do that, we would probably need some sort of `GcnEventViews` table much like the `SourceViews` table to keep track of the times someone clicks on an event link, and then sort by that in the `event_views` handler. 

Alternatively, we could just rename everything to be recent events instead of top events and keep all the logic the same. And we could revisit the top events widget once we actually have an event page for people to view.

acrellin:
Nice, I like the updated API endpoint grouping

static/js/components/RecentGcnEvents.jsx
```diff
@@ -0,0 +1,153 @@
...
+  maxNumEvents: "",
+  sinceDaysAgo: "",
+};
+
+const RecentGcnEvents = ({ classes }) => {
+  const styles = useStyles();
+
+  const { gcnEvents } = useSelector((state) => state.recentGcnEvents);
+  const recentEventsPrefs =
+    useSelector((state) => state.profile.preferences.recentGcnEvents) ||
```
acrellin, line 85:
```suggestion
    useSelector((state) => state.profile.preferences?.recentGcnEvents) ||
```
(preferences can be null)

static/js/ducks/recentGcnEvents.js
```diff
@@ -0,0 +1,32 @@
+import messageHandler from "baselayer/MessageHandler";
+
+import * as API from "../API";
+import store from "../store";
+
+export const FETCH_RECENT_GCNEVENTS = "skyportal/FETCH_RECENT_GCNEVENTS";
+export const FETCH_RECENT_GCNEVENTS_OK = "skyportal/FETCH_RECENT_GCNEVENTS_OK";
+
+export const fetchRecentGcnEvents = () =>
```
acrellin, line 9:
```suggestion
export const fetchGcnEvents = () =>
```
b/c there's no time filtering -- you're just fetching all of them.

static/js/ducks/recentGcnEvents.js
```diff
@@ -0,0 +1,32 @@
+import messageHandler from "baselayer/MessageHandler";
+
+import * as API from "../API";
+import store from "../store";
+
+export const FETCH_RECENT_GCNEVENTS = "skyportal/FETCH_RECENT_GCNEVENTS";
```
acrellin, line 6:
```suggestion
export const FETCH_GCNEVENTS = "skyportal/FETCH_GCNEVENTS";
```
etc throughout

static/js/ducks/recentGcnEvents.js
```diff
@@ -0,0 +1,32 @@
...
+      return {
+        gcnEvents,
+      };
+    }
+    default:
+      return state;
+  }
+};
+
+store.injectReducer("recentGcnEvents", reducer);
```
acrellin, line 32:
```suggestion
store.injectReducer("gcnEvents", reducer);
```

static/js/ducks/recentGcnEvents.js
acrellin: then this same action creator can be used for both the widget and the actual events page

acrellin (commented):
Needs a test, and one renaming request below, otherwise looks great!

static/js/ducks/recentGcnEvents.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+});
+
+const reducer = (state = { gcnEvents: [] }, action) => {
+  switch (action.type) {
+    case FETCH_RECENT_GCNEVENTS_OK: {
+      const gcnEvents = action.data;
+      return {
+        gcnEvents,
+      };
```
acrellin, line 25:
```suggestion
      return action.data;
```
no need for the extra layer here

static/js/components/RecentGcnEvents.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+const defaultPrefs = {
+  maxNumEvents: "",
+  sinceDaysAgo: "",
+};
+
+const RecentGcnEvents = ({ classes }) => {
+  const styles = useStyles();
+
+  const { gcnEvents } = useSelector((state) => state.recentGcnEvents);
```
acrellin, line 83:
```suggestion
  const gcnEvents = useSelector((state) => state.gcnEvents);
```
to accord with comments below

static/js/components/RecentGcnEvents.jsx
```diff
@@ -0,0 +1,153 @@
...
+import Button from "@material-ui/core/Button";
+import Chip from "@material-ui/core/Chip";
+import { makeStyles } from "@material-ui/core/styles";
+
+import dayjs from "dayjs";
+import utc from "dayjs/plugin/utc";
+import relativeTime from "dayjs/plugin/relativeTime";
+
+import * as profileActions from "../ducks/profile";
+import * as recentEventsActions from "../ducks/recentGcnEvents";
```
acrellin, line 18:
```suggestion
import * as gcnEventsActions from "../ducks/gcnEvents";
```

mcoughlin:
Looks like failures unrelated.

static/js/components/RecentGcnEvents.jsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(gcnEventsActions.fetchGcnEvents());
+  }, [dispatch]);
+
+  if (!gcnEvents) {
+    return <CircularProgress />;
+  }
```
acrellin, line 97:
```suggestion
```
sorry, I take it back, I forgot this was a homepage widget -- probably a bit distracting in that case

acrellin (approved):
Looks good, thanks @mcoughlin ! 
Just one request (my fault) below

mcoughlin:
@kmshin1397 strikes again.

mcoughlin:
@kmshin1397 looks unrelated.

kmshin1397 (approved):

PR #2083, opened 2021-06-23
Implement expiration of user accounts

kmshin1397:
Using a new `expiration_date` column for Users as included in cesium-ml/baselayer#233, add an optional expiration to user accounts and invitations after which the user is automatically set to a view only role. 
- An expiration date column has been added to both the invitations and user management tables to display and support editing of expiration dates (on the UserManagement page so only available to admins). 
- Expiration dates can also now be specified when creating a new invitation, through the API, Group page, and via bulk invites with CSV.
- A new cron job runs every 24 hours and for every user whose expiration date is expired, sets the user to view only and removes all associated ACLs

Closes fritz-marshal/fritz#180

TODO: Still need to write some tests, but wanted to allow people to start taking a look at the changes

expiration
image


skyportal/handlers/api/user.py
```diff
@@ -359,6 +361,51 @@ def post(self):
...
+        """
+        data = self.get_json()
+        if user_id is not None:
+            user = User.get_if_accessible_by(
+                user_id, self.current_user, mode="update", raise_if_none=True
+            )
+            if data.get("expirationDate") is not None:
+                expiration_date = data.pop("expirationDate").strip()
+                expiration_date = arrow.get(expiration_date).datetime
+                user.expiration_date = expiration_date
```
acrellin, line 402:
```suggestion
            if (expiration_date := data.get("expirationDate")) is not None:
                user.expiration_date = arrow.get(expiration_date.strip()).datetime
```
Just a minor compactification, and an opportunity to add our first walrus operator to SkyPortal! 😁 🎉

acrellin (approved):
All looks good! As I mentioned earlier, I kinda think deactivating accounts altogether upon expiration is a bit more intuitive than making them read-only, but I'm not gonna push for that if no one else agrees/cares. If we did go that route, it'd be simple to add an active/inactive check to the `permissions` and `auth_or_token` handler decorators to manage access. But I'm good with this approach if there are benefits to allowing users to remain read-only after expiration.

kmshin1397:
CI failures should be resolved once the baselayer PR goes in and gets pinned, I think

kmshin1397:
This has been updated to remove the cron job to set expired user to view only, in favor of an update to https://github.com/cesium-ml/baselayer/pull/233 which updates the `auth_or_token` decorator to block requests from expired users. Also added a test for the expiration date form UI as well as the expiration date blocking requests on the homepage.

acrellin (approved):
Nice, thanks!

dmitryduev (commented):
Thanks @kmshin1397. I agree with @acrellin that blocking expired users altogether is a more intuitive behavior. Would you mind re-implementing it like that?

stefanv:
The baselayer changes look good (approved).

kmshin1397:
> Thanks @kmshin1397. I agree with @acrellin that blocking expired users altogether is a more intuitive behavior. Would you mind re-implementing it like that?

@dmitryduev Yep, this is already done, just all the changes are on the baselayer side: cesium-ml/baselayer#233

stefanv:
> > Thanks @kmshin1397. I agree with @acrellin that blocking expired users altogether is a more intuitive behavior. Would you mind re-implementing it like that?
> 
> @dmitryduev Yep, this is already done, just all the changes are on the baselayer side: [cesium-ml/baselayer#233](https://github.com/cesium-ml/baselayer/pull/233)

At least some of the text in this PR suggests that it hasn't been synced with the BL changes yet: https://github.com/skyportal/skyportal/pull/2083/files#diff-48d0ad5d2d0c9d9416a440bedf97d3142f2c58d5589ce93dc1e8316d0bb0bc3eR80

kmshin1397:
> > > Thanks @kmshin1397. I agree with @acrellin that blocking expired users altogether is a more intuitive behavior. Would you mind re-implementing it like that?
> > 
> > 
> > @dmitryduev Yep, this is already done, just all the changes are on the baselayer side: [cesium-ml/baselayer#233](https://github.com/cesium-ml/baselayer/pull/233)
> 
> At least some of the text in this PR suggests that it hasn't been synced with the BL changes yet: https://github.com/skyportal/skyportal/pull/2083/files#diff-48d0ad5d2d0c9d9416a440bedf97d3142f2c58d5589ce93dc1e8316d0bb0bc3eR80

Ah yes good catch @stefanv I forgot about updating the docs - will do that

skyportal/handlers/api/invitations.py
```diff
@@ -359,9 +380,15 @@ def patch(self, invitation_id):
...
+        user_expiration_date = data.get("userExpirationDate")
+        if (
+            group_ids is None
+            and stream_ids is None
+            and role_id is None
+            and user_expiration_date is None
+        ):
             return self.error(
-                "At least one of `groupIDs`, `streamIDs` or `role` is required."
+                "At least one of `groupIDs`, `streamIDs`, `role`, or `userExpirationDate` is required."
```
stefanv, line 391: Just double checking: can a user then be created with just an expiration date?  Is that desired?

skyportal/handlers/api/user.py
```diff
@@ -359,6 +361,50 @@ def post(self):
...
+              application/json:
+                schema: Success
+        """
+        data = self.get_json()
+        if user_id is not None:
+            user = User.get_if_accessible_by(
+                user_id, self.current_user, mode="update", raise_if_none=True
+            )
+            if (expiration_date := data.get("expirationDate")) is not None:
+                user.expiration_date = arrow.get(expiration_date.strip()).datetime
```
stefanv, line 401: No catch for the parse error here?

static/js/components/InviteNewGroupUserForm.jsx
```diff
@@ -73,6 +104,14 @@ const InviteNewGroupUserForm = ({ group_id }) => {
     });
   };
 
+  const handleExpirationDateChange = (date) => {
+    console.log(date);
```
stefanv, line 108: Leftover debug message

skyportal/handlers/api/invitations.py
```diff
@@ -166,6 +173,13 @@ def post(self):
                 "Invalid value provided for `canSave` parameter: "
                 "all elements must be booleans"
             )
+        user_expiration_date = data.get("userExpirationDate")
+        if user_expiration_date is not None:
+            try:
+                user_expiration_date = arrow.get(user_expiration_date).datetime
```
stefanv, line 179: Does arrow know to use UTC by default?

skyportal/handlers/api/invitations.py
kmshin1397: This is in the PATCH handler and not the POST. The assumption is that the required info is already there, and we're allowing users to edit just one of those fields if desired.

stefanv: Got it.

skyportal/handlers/api/invitations.py
kmshin1397: Yup, looks like it does:
image

static/js/components/InviteNewGroupUserForm.jsx
kmshin1397: Fixed

skyportal/handlers/api/user.py
kmshin1397: Added error catching

stefanv (approved):

dmitryduev (approved):
Excellent, thanks @kmshin1397. Can I merge it now?

kmshin1397:
@dmitryduev Yes, I believe it's good to merge

Issue #2087, opened 2021-06-25
Cannot delete owned spectrum comment from main source page


acrellin:
When viewing source comments list, one cannot delete an owned spectrum comment that appears in the list (invalid ID error)

PR #2088, opened 2021-06-25
Fix bug in delete spectrum comments on source page

acrellin:
Closes https://github.com/skyportal/skyportal/issues/2087

### DIFF
--- a/static/js/components/CommentList.jsx
+++ b/static/js/components/CommentList.jsx
@@ -208,7 +208,7 @@ const CommentList = ({
       spectra !== null &&
       objID in spectra
     ) {
-      specComments = spectra[objID].map((spec) => spec.comments).flat();
+      specComments = spectra[objID]?.map((spec) => spec.comments)?.flat();
     }
     if (comments !== null && specComments !== null) {
       comments = specComments.concat(comments);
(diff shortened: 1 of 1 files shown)


kmshin1397 (commented):
@acrellin I'm not sure I understand the bug fix here. Shouldn't `associatedResourceType` already be the same as `spectrum_id ? "spectrum" : "object"`? Is there anything else going on here in terms of changing the logic that addresses the error?

acrellin:
@kmshin1397 `associatedResourceType` is passed in to `CommentList` as a prop (so when it appears on the source page, this is always "object", and when it's on the data management page beside spectra, it's "spectrum"). `spectrum_id`, however, comes from the comments being mapped, and is the relevant var to check here.

kmshin1397:
> @kmshin1397 `associatedResourceType` is passed in to `CommentList` as a prop (so when it appears on the source page, this is always "object", and when it's on the data management page beside spectra, it's "spectrum"). `spectrum_id`, however, comes from the individual comments being mapped, and is the relevant var to check here.

Ah yes that makes sense now. I didn't make the connection that these were for spectrum comments on the obj page. Looks good then, good catch!

kmshin1397 (approved):

PR #2095, opened 2021-06-30
Clean up GcnEvent routes

kmshin1397:
Move the current multiple GET logic to a `RecentGcnEventsHandler` and implement separate GET multiple logic to just fetch all events (can be expanded later to support filtering etc), bringing it more inline with how we handle sources.

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -83,6 +83,7 @@ from skyportal.handlers.api.internal import (
     StandardsHandler,
     NotificationHandler,
     BulkNotificationHandler,
+    RecentGcnEventsHandler,
 )
 
 from . import models, model_util, openapi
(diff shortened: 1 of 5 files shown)


skyportal/app_server.py
```diff
@@ -189,6 +190,7 @@ def log_request(self, handler):
     (r'/api/internal/notifications(/[0-9]+)?', NotificationHandler),
     (r'/api/internal/notifications/all', BulkNotificationHandler),
     (r'/api/internal/ps1_thumbnail', PS1ThumbnailHandler),
+    (r'/api/internal/recent_gcn_events', RecentGcnEventsHandler),
```
mcoughlin, line 193: Can you help me understand why something is internal or is not?

mcoughlin (commented):
Just want to understand logic behind internal or not.. otherwise makes sense.

acrellin (approved):
Looks good, thanks @kmshin1397

skyportal/app_server.py
acrellin: Internal endpoints are those we typically only call from the front-end and don't necessarily intend to be public-facing.

mcoughlin: @acrellin Ok thanks!

mcoughlin (approved):

mcoughlin:
@acrellin I think you have to merge.

PR #2120, opened 2021-07-09
Classification sliders for adding multiple classifications at once

### LINKED ISSUE #2117: Dynamic sliders for applying multiply classifications at once
https://images.zenhubusercontent.com/82240075/2840d76a-19bc-4dbf-9f58-5247541a0090/zvm_labeling_interface.mp4

kmshin1397:
In the pullout rows of the GroupSources page, add sets of sliders based on accessible taxonomies which allow for easy submission of new classifications, organized by hierarchy.

Also add the variables taxonomy to the demo data, using this data: https://github.com/dmitryduev/ztf-variable-marshal/blob/master/ztf-variable-marshal/config.json#L180

Not sure if the behavior of the sliders are the exact same as they were on the current marshal, like how probabilities of ancestor/descendent nodes change when sliders are moved. I just implemented what I thought was reasonable, but would appreciate if you could check out the branch and play around with it a bit :)

sliders

Closes #2117


static/js/components/MultipleClassificationsForm.jsx
```diff
@@ -0,0 +1,363 @@
...
+          probabilityOfSubclasses,
+          newFormState[selectedTaxonomy.id][ancestor]?.probability || 0
+        ),
+      };
+    });
+
+    // Update subclasses' probabilities to be the minimum
+    // between the current probability and the new value
+    const node = getNode(classification, path);
+    updateChildren(node, newValue, newFormState, path.length);
```
kmshin1397, line 183: Let me know if this behavior is what you/users would expect. I took some creative liberty here based on what I thought was logical, but not sure what the exact behavior is in the current marshal

static/js/components/MultipleClassificationsForm.jsx
```diff
@@ -0,0 +1,363 @@
...
+
+  return (
+    <div className={classes.container}>
+      <Typography variant="h6">Edit Classifications</Typography>
+      <FormControl className={classes.taxonomySelect}>
+        <InputLabel id={`taxonomy-select-label-${objId}`}>
+          Select Taxonomy
+        </InputLabel>
+        <Select
+          labelId={`taxonomy-select-label-${objId}`}
```
dmitryduev, line 299: This should prevent the form from adding that annoying padding to <body>:
```suggestion
          labelId={`taxonomy-select-label-${objId}`}
          inputProps={{ MenuProps: { disableScrollLock: true } }}
```

static/js/components/MultipleClassificationsForm.jsx
dmitryduev: The logic makes sense to me. @mcoughlin you agreez?
Poked around with the ui, and seeing some weird behavior at time though:
class-selectors

dmitryduev (commented):
Looks truly awesome, thanks a lot @kmshin1397! A few comments:
- Is there an easy way to "expand" the columns to the full available width if there are not enough of them to fill out all cols?
- Also, could you please add some visual separators like vertical lines indicating indentation, or lines separating different columns?
- Once you select a taxonomy for a source, I think it would make sense to automatically select the same taxonomy for all other sources, that would save quite a few mouse clicks.
- Add this capability to the Sources page, too? It shows up only on the Group Sources page.

static/js/components/MultipleClassificationsForm.jsx
mcoughlin: @kmshin1397 @dmitryduev First of all, playing around with it is awesome. I understand based on this that the parent classification defaults to the sum of the children one. And the child classification cannot exceed that of the parent. I like this behavior as it enforces conservation of classification ;) I am wondering at the children level, if I choose now like 0.75 and then 0.5 for another,  W Uma and detached eclipsing say, eclipsing jumps to 1.0 which makes sense. I am wondering if by default it should bounce back to 0.5+0.5. I think how it *should* behave is a bit ill-defined anyways and seems clear you can get to where it should be without much difficulty.

kmshin1397: Sure, that makes sense and should be doable. I'll work on that change.

kmshin1397:
Still working on improving the ancestor/descendent node probability responses, but I've addressed your other comments @dmitryduev 

image
image

profjsb (commented):
Looks very nice. I'd suggest:
1) Tooltip explaining the expected User interaction, since it's not intuitive. Do all the probabilities sum to 1 at the same-node level? Ie. conditional or marginal probabilities are implied?
2) A dynamic, text-based view of what classification will be assigned based on the sliders. Eg. "supernova=0.1; eclipsing=0.9; SNIa=1.0" (or whatever the probability sum logic is).

kmshin1397:
> Looks very nice. I'd suggest:
> 
> 1. Tooltip explaining the expected User interaction, since it's not intuitive. Do all the probabilities sum to 1 at the same-node level? Ie. conditional or marginal probabilities are implied?
> 2. A dynamic, text-based view of what classification will be assigned based on the sliders. Eg. "supernova=0.1; eclipsing=0.9; SNIa=1.0" (or whatever the probability sum logic is).

@profjsb Thanks for the suggestions, I like #2 a lot and will work on that. 

As for #1, I agree that the current version is not very intuitive, and I've been working on changing it to be more so. In your opinion, what sort of behavior would you intuitively expect with regards to how the probabilities in the hierarchy are related/automatically adjusted?

profjsb:
> @profjsb Thanks for the suggestions, I like #2 a lot and will work on that.
> 
> As for #1, I agree that the current version is not very intuitive, and I've been working on changing it to be more so. In your opinion, what sort of behavior would you intuitively expect with regards to how the probabilities in the hierarchy are related/automatically adjusted?

The classification API does not enforce that probabilities sum to 1 (on purpose). The logic here is that different people int the same group should be allowed to make classification statements at different times (when there's more or less info) and so the classification vector is dynamic. Here, however, the classification is being done at the same moment in time when all the data is available so I think summing to unity is a reasonable expectation. Under the assumption that the taxonomy is acyclic (which we dont enforce in the schema but is what most taxonomies look like), the logic should be that the sum of the child node/leaf classifications should sum to the value of the parent node and that the total probability should sum to 1. Ie. terminal nodes and leafs represent the marginal probabilities, not conditionals. 

So, e.g. 
  P(source) = 1
    - P(pulsational)=0.5    
       - P(RRL)=0.5
          - P(RRab)=0.25
          - P(RRc)=0.25
    - P(eruptive)=0.5
         - P(supernova)=0.5
             - P(SN1a)=0.5

The advantage of this approach is that in searches someone can search for P(RRL) >=0.5 and get this source. The text version of the classifcation for this source should be all those terminal classifications P(SN1a)=0.5, P(RRc)=0.25, P(RRab)=0.25 since the parent classification probabilities are implied by these three classification assignments.

kmshin1397:
newsliders
@profjsb @mcoughlin @dmitryduev I've updated the logic for adjusting the probabilities. I'm fairly certain that it now respects the principle that all direct children's probabilities should sum to the parent's probability (there might be some cases that I missed and/or rounding-induced errors ). If you want to play around with it some more and see if it seems more intuitive to you, that'd be great :)

mcoughlin:
@kmshin1397 I think it's really cool. The interface seems intuitive.

profjsb:
Agreed. Let's merge once green

dmitryduev (approved):
Played around with it some just now and confirm that it is really awesome! Let's merge once green. Need to pin the latest master?

dmitryduev:
@kmshin1397 Failure in `test_candidate_list_classifications` looks relevant - could you please take a look?

dmitryduev:
Green, merging!

PR #2123, opened 2021-07-14
Nginx-based rate limiting

acrellin:
Depends on https://github.com/cesium-ml/baselayer/pull/236


stefanv (commented):
We still need a way to test the rate limiting.  Perhaps connecting to 0.0.0.0 instead of 127.0.0.1 would work?

skyportal/tests/api/test_rate_limiting.py
```diff
@@ -0,0 +1,19 @@
+import time
+from skyportal.tests import api
+
+
+def test_api_rate_limiting(view_only_token):
+    n_successful_requests = 0
+    status = 200
+    while status == 200:
```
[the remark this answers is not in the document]
acrellin: This has been removed altogether.

skyportal/tests/rate_limiting/test_rate_limiting.py
```diff
@@ -1,26 +1,58 @@
...
+
+s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+s.connect(("8.8.8.8", 53))
+localhost_external_ip = s.getsockname()[0]
+s.close()
+
+
 def test_api_rate_limiting(view_only_token):
+    # In case this test gets run after those defined below
+    time.sleep(5)
```
stefanv, line 18: Instead of an arbitrary sleep, perhaps there should be a loop *at the end* of this test that hits the API until it becomes available again?  That way, we are not reliant on test order, since the test becomes self-encapsulated.

stefanv (commented):
Looking good!

skyportal/tests/rate_limiting/test_rate_limiting.py
```diff
@@ -0,0 +1,58 @@
...
+localhost_external_ip = s.getsockname()[0]
+s.close()
+
+
+def test_api_rate_limiting(view_only_token):
+    # In case this test gets run after those defined below
+    time.sleep(1)
+    n_successful_requests = 0
+    status = 200
+    while status == 200 and n_successful_requests < 100:
```
stefanv, line 21: This is fine, but it's a bit easier to write:

```
for n_successful_requests in range(1, 100):
    r = ...
    if r.status_code == 200:
        break
```

acrellin: :+1: done

skyportal/tests/rate_limiting/test_rate_limiting.py
```diff
@@ -0,0 +1,55 @@
...
+        r = requests.get(
+            f'http://{localhost_external_ip}:{cfg["ports.app"]}/api/sysinfo',
+            headers={'Authorization': f'token {view_only_token}'},
+        )
+        assert r.status_code == 200
+        time.sleep(0.2)
+
+
+def test_localhost_unlimited(view_only_token):
+    for i in range(30):
```
stefanv, line 53: Left a comment here, but somehow it disappeared.

Curious what this is testing if the number is the same as the above.  Shouldn't the number here be higher?

acrellin: This test is un-throttled (all requests as quickly as possible) and is coming from localhost (note we're not using the external IP address in this test). In the test above, we are using the external IP address but _throttling_ down to at most 5/s.

acrellin: See test names

stefanv: That makes perfect sense; can you add a comment along those lines, just to help readers along?  I couldn't quite grasp that this was happening from the function names only.

After that, go ahead and merge once @kmshin1397 approves.  Thanks, Ari!

stefanv (approved):

kmshin1397 (approved):
I think it would be good to add some inline comments in the tests to say that these values, i.e. the looping 30 times with sleeps of 0.2s, are based on the default rates specified in the baselayer nginx configs, in case we decide to change those in the future and need to update these tests accordingly. Otherwise, looks good!

acrellin:
@kmshin1397 Good call -- done!

stefanv:
Great, thanks Ari---let's see if it hits green!

kmshin1397:
Test failure unrelated (timeout of big photometry post test due to GA VM performance variance)

PR #2125, opened 2021-07-14
Pin working requests version

kmshin1397:
This PR attempts to fix the issue with Photometry data loading (seen in the CI) caused by the recent update to `requests`.
Specifically, we can no longer have `np.nan` values in the JSON body sent to `requests.post()`: https://github.com/psf/requests/pull/5810.

~~I tried just replacing those NaN's with `None`, and the post at least goes through without error. But I'm not sure if it's a perfect solution, as now I see some weird behavior from the photometry plot with this newly loaded data:
image
Any thoughts?~~

EDIT: 
This PR just pins the `requests` version to be below the breaking-change version of 2.26.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -46,3 +46,4 @@ ligo.skymap==0.5.3
 healpy==1.15.0
 pygcn==1.1.0
 xmlschema==1.6.1
+requests==2.25.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -46,4 +46,4 @@ ligo.skymap==0.5.3
 healpy==1.15.0
 pygcn==1.1.0
 xmlschema==1.6.1
-requests==2.25.1
+requests<2.26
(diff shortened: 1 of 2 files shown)


kmshin1397:
CI still fails because the workflow runs the `make load_demo_data` on current master before checking out the PR branch (to test the migration manager)

kmshin1397:
Changed directions to just pin the last working version of `requests` instead, in order to avoid disrupting existing pipelines.

requirements.txt
```diff
@@ -46,3 +46,4 @@ ligo.skymap==0.5.3
 healpy==1.15.0
 pygcn==1.1.0
 xmlschema==1.6.1
+requests==2.25.1
```
dmitryduev, line 49:
```suggestion
requests<2.26
```
? In case they patch 2.25 in the near future.

dmitryduev (approved):

acrellin (approved):
PR description needs updating, otherwise :+1:

kmshin1397:
> PR description needs updating, otherwise 👍

Done :)

stefanv:
It looks like replacing NaN with None fixes the conversion.

We can also explicitly convert with `simplejson`.

PR #2127, opened 2021-07-16
Add whitelisted IPs section to config

acrellin:


dmitryduev (approved):

stefanv:
Should we also parameterize the burst rate etc.?

acrellin:
> Should we also parameterize the burst rate etc.?

This is the wrong place for that, but sure, I'll add those in baselayer

stefanv:
Yeah, not urgent either. This is good to go.

acrellin:
Merging with two approvals, failure unrelated: `FAILED skyportal/tests/api/test_photometry.py::test_token_user_big_post - ass...`

PR #2129, opened 2021-07-16
WIP: Enabling slicing based on skymaps

mcoughlin:
WIP example for slicing sources based on skymaps. Relies on PR to healpix-alchemy:
https://github.com/skyportal/healpix-alchemy/pull/15

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -6,6 +6,8 @@ import uuid
 
 import arrow
 import numpy as np
+import healpix_alchemy as ha
+import astropy.units as u
 
 from sqlalchemy.orm import joinedload
 from sqlalchemy.sql.expression import case, func, FromClause
@@ -958,6 +960,9 @@ class CandidateHandler(BaseHandler):
 
         if dec is None and not obj_already_exists:
             return self.error("Dec must not be null for a new Obj")
+
+        if not obj_already_exists:
+            data["nested"] = ha.healpix.HPX.lonlat_to_healpix(ra * u.deg, dec * u.deg)
 
         passing_alert_id = data.pop("passing_alert_id", None)
         passed_at = data.pop("passed_at", None)
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -5,6 +5,8 @@ import gcn
 import lxml
 import xmlschema
 from urllib.parse import urlparse
+import healpix_alchemy as ha
+import numpy as np
 
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm import joinedload
@@ -17,6 +19,7 @@ from ...models import (
     GcnNotice,
     GcnTag,
     Localization,
+    LocalizationTile,
 )
 from ...utils.gcn import get_dateobs, get_tags, get_skymap, get_contour
 
@@ -112,6 +115,25 @@ class GcnEventHandler(BaseHandler):
             localization = Localization(**skymap)
             localization = get_contour(localization)
             DBSession().add(localization)
+            self.verify_and_commit()
+
+        tiles, probs = [], []
+        for uniq, probdensity in zip(skymap["uniq"], skymap["probdensity"]):
+            tile = LocalizationTile(
+                localization_id=localization.id, uniq=int(uniq), probdensity=probdensity
+            )
+            area = (tile.nested_hi - tile.nested_lo + 1) * ha.healpix.PIXEL_AREA
+            prob = probdensity * area
+            tiles.append(tile)
+            probs.append(prob)
+
+        idx = np.argsort(probs)[::-1].astype(int)
+        tiles, probs = [tiles[ii] for ii in idx], [probs[ii] for ii in idx]
+        cumprobs = np.cumsum(probs)
+
+        for tile, cumprob in zip(tiles, cumprobs):
+            tile.cumprob = cumprob
+            DBSession().add(tile)
 
         self.verify_and_commit()
 
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -7,6 +7,7 @@ import tornado
 from tornado.ioloop import IOLoop
 import io
 import math
+import astropy.units as u
 from dateutil.parser import isoparse
 from sqlalchemy.orm import joinedload
 from sqlalchemy import func, or_, distinct
@@ -38,6 +39,8 @@ from ...models import (
     Classification,
     Taxonomy,
     Listing,
+    Localization,
+    LocalizationTile,
     Spectrum,
     SourceView,
 )
@@ -555,6 +558,10 @@ class SourceHandler(BaseHandler):
         max_latest_magnitude = self.get_query_argument("maxLatestMagnitude", None)
         has_spectrum = self.get_query_argument("hasSpectrum", False)
 
+        localization_dateobs = self.get_query_argument("localizationDateobs", None)
+        localization_name = self.get_query_argument("localizationName", None)
+        localization_cumprob = self.get_query_argument("localizationCumprob", 1.0)
+
         # These are just throwaway helper classes to help with deserialization
         class UTCTZnaiveDateTime(fields.DateTime):
             """
@@ -1017,6 +1024,28 @@ class SourceHandler(BaseHandler):
                     Obj.id == classification_subquery.c.obj_id,
                     isouter=True,
                 )
+        if localization_dateobs is not None and localization_name is not None:
+            localization = (
+                Localization.query_records_accessible_by(self.current_user)
+                .filter(Localization.dateobs == localization_dateobs)
+                .filter(Localization.localization_name == localization_name)
+                .first()
+            )
+            if localization is None:
+                return self.error("Localization not found", status=404)
+            tiles_subquery = (
+                LocalizationTile.query_records_accessible_by(self.current_user)
+                .filter(LocalizationTile.localization_id == localization.id)
+                .filter(LocalizationTile.cumprob <= localization_cumprob)
+                .subquery()
+            )
+
+            obj_query = obj_query.join(
+                tiles_subquery,
+                Obj.nested.between(
+                    tiles_subquery.c.nested_lo, tiles_subquery.c.nested_hi
+                ),
+            )
 
         source_query = apply_active_or_requested_filtering(
             source_query, include_requested, requested_only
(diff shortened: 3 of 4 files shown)


mcoughlin:
@kmshin1397 Among other things, I could use some help understanding how to get the route.dateobs into the source request (to trigger source -> localization crossmatch).

kmshin1397:
> @kmshin1397 Among other things, I could use some help understanding how to get the route.dateobs into the source request (to trigger source -> localization crossmatch).

Hmm does the `route.dateobs` show up properly within `GcnEventPage`, and not in the `GcnEventSourcesPage` component? Or not at all? Does it not show up when you `console.log(route.dateobs)`? At first glance I don't see any glaring errors...

mcoughlin:
I think I just don't understand how to be sure the route stuff gets segmented properly in the different types of if statements in the duck.

kmshin1397:
> I think I just don't understand how to be sure the route stuff gets segmented properly in the different types of if statements in the duck.

Ah hmm not sure if I understand your question exactly, but I'll take a stab at explaining by just generally explaining the ducks haha.

You're passing the `dateobs` value directly into the `fetchGcnEventSources()` function from the `GcnEventSourcesPage` component, and this function is called the action creator. When you call it through a Redux `dispatch`, like `dispatch(sourcesActions.fetchGcnEventSources(route.dateobs, {}));`, that logic in the function will make a call to the SP backend, specifically the `/api/sources` route with the parameters you specified there. Upon successful completion of that, our frontend redux code is set up so that Redux will see a `FETCH_GCNEVENT_SOURCES_OK` message, which will be passed along to the `reducer` in that `sources` duck (all the `case` statements), which takes that data passed back by the API and puts it in the Redux store (which you then access from components via the `useSelector()`). 

The `messageHandler` stuff is for when you want to do communication in the opposite direction: have the API send a message to the front-end to do something, i.e. refresh a list of sources. This is the websocket messaging stuff, and probably not actually relevant here. We generally define separate action types for those, like the `REFRESH_FAVORITE_SOURCES`, to distinguish websocket messages from the actions triggered by the frontend. 

Let me know if that helps.

mcoughlin:
Yes this is helpful in terms of my understanding, although it remains magic to me. :)

mcoughlin:
@kmshin1397 Very helpful. Is there an example in the JS somewhere where we set:
doing some kind of time math? I need startDate to be dateobs and endDate to be dateobs+3 basically.

stefanv:
You can grep for `dayjs` — it can do all sorts of time calculations.

kmshin1397:
> @kmshin1397 Very helpful. Is there an example in the JS somewhere where we set:
> 
>         start_date = self.get_query_argument('startDate', None)
> 
>         end_date = self.get_query_argument('endDate', None)
> 
> doing some kind of time math? I need startDate to be dateobs and endDate to be dateobs+3 basically.        

Do you mean in the python? That code segment looks like Python haha. If so, you probably want to look up Python datetime.timedelta. If you're looking for JS examples, I think FilterCandidateList.jsx probably has one (setting the default scanning start date to be 24hrs ago), or you can just look up dayjs documentation. We use dayjs for most of our JS time math I think.

mcoughlin:
Thank you @stefanv and @kmshin1397! That did it.

skyportal/handlers/api/source.py
```diff
@@ -1298,6 +1336,45 @@ class Validator(Schema):
             query_results["sources"] = obj_list
 
         query_results = recursive_to_dict(query_results)
+        if includeGeojson:
+            features = []
+
+            # currently necessary for the sources to render
+            features.append(
```
acrellin, line 1343: What's up with these three identical `Feature` instances?

skyportal/handlers/api/source.py
```diff
@@ -1298,6 +1336,45 @@ class Validator(Schema):
...
+                    geometry=Point([float(-180), float(-180)]),
+                    properties={"name": "tmp"},
+                )
+            )
+
+            for source in query_results["sources"]:
+                point = Point((source["ra"], source["dec"]))
+                if source["alias"] is not None:
+                    feat = Feature(
+                        geometry=point, properties={"name": ",".join(source["alias"])}
```
acrellin, line 1366: I suggest just defining the name in these if else blocks, then afterwards use its value in a single Feature instantiation to avoid repeated code

static/js/components/GcnEventPage.jsx
```diff
@@ -148,25 +252,35 @@ const Localization = ({ loc }) => {
...
 };
 
 const GcnEventPage = ({ route }) => {
   const mapRef = useRef();
-  const gcnEvent = useSelector((state) => state.gcnEvent);
   const dispatch = useDispatch();
   const styles = useStyles();
 
+  const gcnEvent = useSelector((state) => state.gcnEvent);
+  const sourcesState = useSelector((state) => state.sources.gcnEventSources);
```
acrellin, line 266:
```suggestion
  const gcnEventSources = useSelector((state) => state.sources.gcnEventSources);
```
and below

acrellin (commented):
Looks like a great start!  A few minor comments below, and this also needs tests. Can you also update the PR description with any relevant screen shots/captures of the front-end additions? Thanks!

skyportal/handlers/api/candidate.py
```diff
@@ -959,6 +961,9 @@ def post(self):
         if dec is None and not obj_already_exists:
             return self.error("Dec must not be null for a new Obj")
 
+        if not obj_already_exists:
+            data["nested"] = ha.healpix.HPX.lonlat_to_healpix(ra * u.deg, dec * u.deg)
```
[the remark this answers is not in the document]
mcoughlin: Yes this is one of the additions in the https://github.com/skyportal/healpix-alchemy/pull/15 PR

skyportal/handlers/api/source.py
mcoughlin: For some reason, I had to put a few down to get them to render on the front-end... Might just me being dumb though, IDK.

acrellin:
Adding @stefanv 's comment that all of the new model types and associated concepts should be well-documented here (and in subsequent PRs where relevant).

static/js/components/GcnEventPage.jsx
mcoughlin: Good call!

mcoughlin:
Screen Shot 2021-07-26 at 6 06 41 PM

Behold the amazingness.

skyportal/handlers/api/source.py
mcoughlin: much better!

skyportal/handlers/api/source.py
acrellin: Can't we just add the relevant data to the demo data?

mcoughlin: I am not quite sure what you mean. The geojson feature table is generated from the sources list.

acrellin: I'm not sure I understand why this is needed, but it definitely doesn't seem like it should be necessary

mcoughlin: I couldn't agree more. I am pretty sure this is just me not being good at front end stuff. But if I include it, points do appear ;)

acrellin: We need to figure out what's going on here, why these dummy points are needed, and either get rid of them or at least document the valid reason for them being here. And instead of all this repeated code, please do something more concise like `features.extend([Feature(...)] * 3)`

skyportal/handlers/api/source.py
```diff
@@ -1298,6 +1345,45 @@ class Validator(Schema):
...
+
+            for source in query_results["sources"]:
+                point = Point((source["ra"], source["dec"]))
+                if source["alias"] is not None:
+                    source_name = ",".join(source["alias"])
+                else:
+                    source_name = f'{source["ra"]},{source["dec"]}'
+
+                feat = Feature(geometry=point, properties={"name": source_name})
+                features.append(feat)
```
acrellin, line 1382:
```suggestion
                features.append(Feature(geometry=point, properties={"name": source_name}))
```

skyportal/handlers/api/source.py
```diff
@@ -1354,6 +1440,9 @@ def post(self):
         if dec is None and not obj_already_exists:
             return self.error("Dec must not be null for a new Obj")
 
+        if not obj_already_exists:
```
acrellin, line 1443: Add in-line comments explaining what's happening here

skyportal/handlers/api/source.py
```diff
@@ -555,6 +559,20 @@ def get(self, obj_id=None):
...
+        includeGeojson = self.get_query_argument("includeGeojson", False)
+
+        print(
+            localization_dateobs,
+            localization_name,
+            localization_cumprob,
+            includeGeojson,
+            start_date,
+            end_date,
+        )
```
acrellin, line 574:
```suggestion
```

skyportal/handlers/api/gcn.py
```diff
@@ -112,6 +115,25 @@ def post(self):
             localization = Localization(**skymap)
             localization = get_contour(localization)
             DBSession().add(localization)
+            self.verify_and_commit()
+
+        tiles, probs = [], []
```
acrellin, line 120: Add in-line comments explaining what's happening here and below

skyportal/models.py
```diff
@@ -3559,7 +3559,9 @@ def noise(self):
 class Localization(Base):
     """Localization information, including the localization ID, event ID, right
     ascension, declination, error radius (if applicable), and the healpix
-    map."""
+    map. The healpix map is a multi-order healpix skymap, and this
```
acrellin, line 3562: Can you explain what healpix means here for the uninitiated? Lay terms

skyportal/models.py
```diff
@@ -3685,6 +3687,37 @@ def flat(self):
         else:
             return (self.flat_2d,)
 
+    tiles = relationship("LocalizationTile")
+
+
+class LocalizationTile(ha.Tile, Base):
+    """This is a single tile within a skymap (as in the Localization table.
```
acrellin, line 3694:
```suggestion
    """This is a single tile within a skymap (as in the Localization table).
```

static/js/components/GcnEventPage.jsx
```diff
@@ -148,25 +252,37 @@ const Localization = ({ loc }) => {
...
+
   if (!gcnEvent) {
     return <CircularProgress />;
   }
 
+  if (!gcnEventSources) {
+    return <CircularProgress />;
+  }
+
+  console.log("data", gcnEventSources);
```
acrellin, line 284:
```suggestion
```

mcoughlin:
@acrellin can you help me confirm the Point behavior on the d3 skymap? It's driving me wild.

PR #2130, opened 2021-07-19
Add hours below airmass plot to observability page

kmshin1397:
Recreate this plot from the GROWTH marshal: 
Screen Shot 2021-06-30 at 10 58 00 AM
By adding this to the SP Observability page:
Screenshot 2021-07-19 145821

Maybe not the most efficient/accurate way to calculate the plot, but I wanted to use the existing code as much as possible and the results seem fine, but open to being corrected :) The curve looks like the same shape, but the values do seem to be slightly off compared to the GROWTH marshal, at least for the one random source I compared (the ZTF21abdmevk from the screenshot) so there may indeed be corrections to be made here. 

The changes to AirmassPlot.jsx are actually all just to make ESLint happier and not relevant.
Closes fritz-marshal/fritz-beta-feedback#61


skyportal/handlers/api/internal/plot.py
```diff
@@ -117,6 +122,52 @@ def get(self, obj_id, telescope_id):
...
+        telescope = Telescope.get_if_accessible_by(
+            telescope_id, self.current_user, raise_if_none=True
+        )
+
+        year = datetime.date.today().year
+        year_start = datetime.datetime(year, 1, 1, 0, 0, 0)
+
+        json = []
+        # Sample every 7 days for the year
+        for day in range(52):
```
mcoughlin, line 154: Maybe best to make 7 a variable (deltat) and then do ceil(365/deltat)?

mcoughlin (approved):
This is really nice @kmshin1397. I wouldn't worry about small implementation differences, this is good enough for round numbers.

dmitryduev (approved):
LGTM!

dmitryduev (commented):
Missed the CI errors, sorry:

```
error  Unable to resolve path to module './HoursBelowAirmassPlot'  import/no-unresolved
```

Could you please check what's wrong please, @kmshin1397?

skyportal/handlers/api/internal/plot.py
kmshin1397: Good suggestion, done

kmshin1397:
> Missed the CI errors, sorry:
> 
> ```
> error  Unable to resolve path to module './HoursBelowAirmassPlot'  import/no-unresolved
> ```
> 
> Could you please check what's wrong please, @kmshin1397?

@dmitryduev It was because ESLint was looking for the JSX file that won't be created until `make run` is run and the template files are used to generate the JSX file. Looks like all previous templated JSX components were imported by other templated components so we never saw this before. I just disabled the linting for that line.

dmitryduev:
@kmshin1397 CI failures seem unrelated, so merging.

Issue #2010, opened 2021-05-24
Limit the number of tokens a user can generate


stefanv:
This quantity should default to 1, but must also be adjustable on a per-site basis by the system administrator.

PR #2132, opened 2021-07-20
Limit regular users to one token

acrellin:
Closes https://github.com/skyportal/skyportal/issues/2010

### DIFF
--- a/skyportal/handlers/api/internal/token.py
+++ b/skyportal/handlers/api/internal/token.py
@@ -11,7 +11,7 @@ class TokenHandler(BaseHandler):
     def post(self):
         """
         ---
-        description: Generate new token
+        description: Generate new token (limit 1 per user)
         requestBody:
           content:
             application/json:
@@ -40,6 +40,11 @@ class TokenHandler(BaseHandler):
             return self.error(
                 "User has attempted to grant token ACLs they do not have "
                 "access to. Please try again."
+            )
+        if len(user.tokens) > 0 and not user.is_admin:
+            return self.error(
+                "You have reached the maximum number of tokens "
+                "allowed for your account type."
             )
         token_name = data['name']
         if Token.query.filter(Token.name == token_name).first():
--- a/skyportal/tests/frontend/test_tokens.py
+++ b/skyportal/tests/frontend/test_tokens.py
@@ -3,7 +3,7 @@ import uuid
 
 
 @pytest.mark.flaky(reruns=2)
-def test_add_token(driver, user, public_group):
+def test_add_token(driver, user):
     token_name = str(uuid.uuid4())
     driver.get(f'/become_user/{user.id}')
     driver.get('/profile')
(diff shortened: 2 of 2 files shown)


stefanv:
The idea behind multiple tokens is that users can generate one for each of their services, each with minimal permissions.  As such, I'm not sure how we benefit by limiting the number of tokens.

One also seems quite restrictive; should we limit to 3?  And, then, should we make that number configurable?

acrellin:
@stefanv We discussed this this morning... why did you not raise objections then before we decided to implement? The group decided on the approach implemented here.

acrellin:
@stefanv per-token rate limiting is pointless if users can have multiple tokens.

stefanv:
I'm sorry for upsetting you.  I will politely excuse myself from this review.

acrellin:
@stefanv no worries, you didn't upset me :) I just didn't know why this was coming up now and not when we were discussing it earlier.

stefanv:
I was there for the conversation earlier, but I didn't think about it too much.  Once I did, I started wondering what exactly we are trying to achieve (given that limiting to 1 is contrary to the typical idea of having lightweight tokens with minimal permissions).

The interaction between rate limiting and tokens makes sense, so a default of 1 is probably a good idea for fritz for now.  It does mean that people who run multiple bots are likely to give their bots too many permissions.  Not ideal, but also not a big deal.

And, we can always parameterize it later.

acrellin:
@stefanv yeah, I agree that if rate-limiting weren't necessary, it'd be ideal not to have to limit the number of tokens per user. Another API token philosophy also seems to be that you create one per account which can do anything you have access to (lots of applications seem to do this), which seems to work well. We can always revisit later if needed.

kmshin1397:
@stefanv @acrellin Another thought I had about this discussion is that the idea of lightweight tokens with minimal permissions relies on users to follow this principle when creating their tokens. I wouldn't be surprised if most users are just simply creating one token for everything they do regardless, seeing as (as far as I know) we don't have any tooltips/guidelines for how we'd like people to use tokens on the Profile page or in the docs. In light of that, I'm in support of the 1-token limit as discussed 👍

kmshin1397 (approved):
LGTM, thanks @acrellin! One thought: how do you think we should handle existing cases of users with more than one token? Maybe we want to do some sort of migration to consolidate existing tokens into one per user? Might be more trouble than it's worth not sure

acrellin:
@dmitryduev Thoughts on this one?

dmitryduev (commented):
@acrellin LGTM. Our current users with multiple tokens won't be affected and I think it is ok to leave them alone. I would, though, check in with other SP users whether it could pose any potential problems for them (+make it configurable in a next PR).

acrellin:
@dmitryduev You commented before, but not sure if you approved; mind giving it a quick review?

dmitryduev (approved):

PR #2133, opened 2021-07-20
Introduce Fields

mcoughlin:
This PR prototypes adding fields attaching to instruments.


mcoughlin:
@acrellin You can see how I did the dask call here, but I think the code needs to know where to look for workers right?

stefanv:
See [BaseHandler](https://github.com/cesium-ml/baselayer/blob/master/app/handlers/base.py#L375) for how to connect to local dask instance.

You are using it in the backend, but for future reference, [here](https://github.com/cesium-ml/baselayer_template_app/blob/master/template_app/handlers/example_computation.py#L21) is how to do it in a handler.

mcoughlin:
@stefanv I assume that turning dask on is the right choice:

services:
  dask: True

and then based on your code I also set the port in the config to something?

acrellin:
:+1: thanks @stefanv

stefanv:
> @stefanv I assume that turning dask on is the right choice:
> 
> services:
> dask: True
> 
> and then based on your code I also set the port in the config to something?

Right, and there should be a port in the default config (63500).

acrellin:
The alembic migration script is definitely wonky -- did you first stamp latest master after initializing the DB @ that commit, then stop the app and checkout your branch, and run the generate migration command? See https://skyportal.io/docs/migrations.html

mcoughlin:
@acrellin Yes I did that exact sequence. But I presumably messed it up. It will need to be updated anyways with rebases and so on. Can you get an idea of how the PR works without it?

acrellin:
@mcoughlin What are fields in this context?

skyportal/handlers/api/instrument.py
```diff
@@ -25,8 +39,32 @@ def post(self):
...
+                )
+                .filter(
+                    Instrument.name == instrument.name,
+                    Instrument.telescope == instrument.telescope,
+                )
+                .one()
+            )
+        except NoResultFound:
+            DBSession().add(instrument)
+            self.verify_and_commit()
```
acrellin, line 56:
```suggestion
        instrument = (
            Instrument.query_records_accessible_by(
                self.current_user,
            )
            .filter(
                Instrument.name == instrument.name,
                Instrument.telescope == instrument.telescope,
            )
            .first()
        )
        if instrument is not None:
            DBSession().add(instrument)
            self.verify_and_commit()
```
but why add this check at all? why wasn't it needed before, but is now?

skyportal/handlers/api/instrument.py
```diff
@@ -17,6 +22,15 @@ def post(self):
...
         )
 
+        if "field_data" in data:
+            field_data = data.pop("field_data")
+            field_of_view_shape = data.pop("field_of_view_shape")
+            field_of_view_size = data.pop("field_of_view_size", None)
+
+            add_fields = True
+        else:
+            add_fields = False
```
acrellin, line 32:
```suggestion
        field_data = data.pop("field_data", None)
        field_of_view_shape = data.pop("field_of_view_shape", None)
        field_of_view_size = data.pop("field_of_view_size", None)
```
paired w/ suggestion below

skyportal/handlers/api/instrument.py
```diff
@@ -25,8 +39,32 @@ def post(self):
...
+                    Instrument.name == instrument.name,
+                    Instrument.telescope == instrument.telescope,
+                )
+                .one()
+            )
+        except NoResultFound:
+            DBSession().add(instrument)
+            self.verify_and_commit()
+
+        if add_fields:
```
acrellin, line 58:
```suggestion
        if field_data is not None:
```

skyportal/handlers/api/instrument.py
```diff
@@ -208,3 +246,23 @@ def delete(self, instrument_id):
...
+):
+    try:
+        tile_args = {'instrument_id': int(instrument_id)}
+        mocs = get_mocs(field_data, field_of_view_shape, field_of_view_size)
+        fields = [
+            InstrumentField.from_moc(moc, field_id=int(field_id), tile_args=tile_args)
+            for field_id, moc in zip(field_data['ID'], mocs)
+        ]
+        for field in fields:
+            field.instrument_id = int(instrument_id)
```
acrellin, line 262: Why is this necessary? You're already passing in instrument ID in tile_args above. And can't you just pass in `instrument_id` directly in the `InstrumentField` instantiation above?

skyportal/models.py
```diff
@@ -1758,6 +1758,65 @@ def api_class(self):
...
 
+    fields = relationship("InstrumentField")
+    tiles = relationship("InstrumentFieldTile")
+
+
+class InstrumentField(ha.Region, Base):
+    """A multi-order healpix representation of a telescope's field shape,
+    as represented by many InstrumentFieldTiles."""
+
+    tile_class = lambda: InstrumentFieldTile  # pylint: disable=E731 # noqa
```
acrellin, line 1769: What's going on here exactly? Not sure why this is needed

skyportal/models.py
```diff
@@ -1758,6 +1758,65 @@ def api_class(self):
...
+        doc='Instrument ID',
+    )
+
+    instrument = relationship(
+        "Instrument",
+        foreign_keys=instrument_id,
+        doc="The Instrument that this field belongs to",
+    )
+
+    field_id = sa.Column(
```
acrellin, line 1783: How is this different from the `id` column this class inherits from `Base`?

skyportal/tests/api/test_instrument.py
```diff
@@ -31,6 +36,9 @@ def test_token_user_post_get_instrument(super_admin_token):
             'band': 'NIR',
             'filters': ['f110w'],
             'telescope_id': telescope_id,
+            'field_data': pd.read_csv(datafile).to_dict(orient='list'),
```
acrellin, line 39: We need to test that the fields/tiles that are expected to be generated actually are

acrellin (commented):
Looking good so far!  A few comments/suggestions/questions below

skyportal/handlers/api/instrument.py
mcoughlin: So for example, ZTF is already in the database, and this will simply add the fields to it.

skyportal/models.py
mcoughlin: This was how Leo had it implemented in his healpix_alchemy implementation (https://github.com/skyportal/healpix-alchemy/pull/15/files#diff-17de8e8b9e279ac99ba4c78cde70804540753339f30642e0faaa614cb6cd843cR72). Is there a better way?

skyportal/models.py
mcoughlin: It shouldn't have to be unique (two telescopes will have a field 1)... is there a unique requirement in Base?

skyportal/tests/api/test_instrument.py
mcoughlin: Do you have a suggestion how to do so aside from the counting?

skyportal/models.py
acrellin: I don't know why you'd need this at all

skyportal/models.py
acrellin: Ah, okay, perhaps make that more explicit in the column desc. Thanks!

skyportal/tests/api/test_instrument.py
acrellin: Wait a reasonable amount of time, then fetch fields/tiles and make sure the ones that should be there are there

skyportal/handlers/api/instrument.py
mcoughlin: InstrumentField.from_moc doesn't have a instrument_id input.

acrellin: Ah, okay.

skyportal/models.py
mcoughlin: How else would I tell it about the tiles in the Region?

skyportal/models.py
mcoughlin: Yes added!

mcoughlin:
@acrellin 

        instrument = (
            Instrument.query_records_accessible_by(
                self.current_user,
            )
            .filter(
                Instrument.name == instrument.name,
                Instrument.telescope == instrument.telescope,
            )
            .first()
        )
        if instrument is not None:
            DBSession().add(instrument)
            self.verify_and_commit()
            
   Doesn't this just end up adding a None to the database?

mcoughlin:
@acrellin It's tricky to ensure the instrument has an ID basically.

skyportal/handlers/api/instrument.py
```diff
@@ -25,8 +34,31 @@ def post(self):
...
+            )
+            .filter(
+                Instrument.name == instrument.name,
+                Instrument.telescope == instrument.telescope,
+            )
+            .first()
+        )
+        if instrument is not None:
+            DBSession().add(instrument)
+            self.verify_and_commit()
```
acrellin, line 50:
```suggestion
        DBSession().add(instrument)
        self.verify_and_commit()
```

acrellin:
@mcoughlin see updated suggested change. A POST request should usually error if the record it attempts to add already exists.

mcoughlin:
@acrellin What if ZTF is in the database but not its fields?

acrellin:
> @acrellin What if ZTF is in the database but not its fields?

Why would that happen if fields are always added when new instruments are added? Back-populating existing instruments with fields we can do in the migration script.

mcoughlin:
Ah i got ya. @acrellin that seems fine then.

skyportal/tests/api/test_instrument.py
mcoughlin: Ok I sliced it to only add a few tiles. Lemme know what you think.

skyportal/handlers/api/instrument.py
```diff
@@ -39,7 +39,17 @@ async def post(self):
         self.verify_and_commit()
 
         if field_data is not None:
-            print(instrument.id)
+            if field_of_view_shape is None:
+                return self.error('field_of_view_shape is required with field_data')
+            elif field_of_view_shape in ['square', 'circle']:
```
stefanv, line 44:
```suggestion
            if field_of_view_shape in ['square', 'circle']:
```

skyportal/handlers/api/instrument.py
```diff
@@ -39,7 +39,17 @@ async def post(self):
...
+                if field_of_view_size is None:
+                    return self.error(
+                        f'field_of_view_size is required for field_of_view_shape={field_of_view_shape}'
+                    )
+            elif field_of_view_shape == "ZTF":
+                pass
+            else:
+                return self.error('field_of_view_shape must be square, circle, or ZTF')
```
stefanv, line 52: I'd replace this with an earlier call:

```
if field_of_view_shape not in ('square', 'circle', 'ZTF'):
    return self.error(...)
```

skyportal/tests/utils/test_instrument.py
```diff
@@ -0,0 +1,32 @@
...
+    mocs = get_mocs(
+        field_data, field_of_view_shape, field_of_view_size=field_of_view_size
+    )
+    assert len(mocs) == 10
+
+    field_of_view_shape = "circle"
+    mocs = get_mocs(
+        field_data, field_of_view_shape, field_of_view_size=field_of_view_size
+    )
+    assert len(mocs) == 10
```
stefanv, line 28: The field_of_view_shape changed, but the only thing that is verified is the length of `mocs`.  Does it not have some other impact?

skyportal/handlers/api/instrument.py
stefanv: So, maybe the whole block could be something like:

```
if field_of_view_shape is None:
    return self.error('`field_of_view_shape` is required with field_data')
if field_of_view_shape not in ('square', 'circle', 'ZTF'):
    return self.error('field_of_view_shape must be square, circle, or ZTF')
if field_of_view_size is None:
    return self.error(
        f'field_of_view_size is required for field_of_view_shape={field_of_view_shape}'
    )
```

skyportal/tests/utils/test_instrument.py
mcoughlin: great point. I improved this.

skyportal/handlers/api/instrument.py
mcoughlin: yes I think I got it fixed like this now.

skyportal/utils/instrument.py
```diff
@@ -0,0 +1,149 @@
...
+    mocpy.MOC
+        MOC representation with length N.
+    """
+
+    IP = '127.0.0.1'
+    PORT_SCHEDULER = cfg['ports.dask']
+
+    client = Client('{}:{}'.format(IP, PORT_SCHEDULER))
+
+    if field_of_view_shape == "ZTF":
```
lpsinger, line 90: When I see these big walls of `if`/`else` blocks, I immediately shout "polymorphism!"

lpsinger:
One of the lessons learned from the GROWTH ToO Marshal is that we should store the instrument fields of view as data files rather than as code. I suggest storing the fields of view as a region file and using [astropy-regions to read it](https://astropy-regions.readthedocs.io/en/stable/region_io.html#reading-writing-region-files). I contributed astropy/regions#307 with this in mind.

PR #2134, opened 2021-07-21
Show first available thumbnail in order of ZTF, SDSS, DR8, PS1 in homepage widgets

acrellin:
Closes https://github.com/skyportal/skyportal/issues/2020

image


skyportal/handlers/api/internal/source_views.py
```diff
@@ -74,3 +78,8 @@ def post(self, obj_id):
         DBSession.add(sv)
         self.verify_and_commit()
         return self.success()
+
+
+def tIndex(t):
```
dmitryduev, line 83:
```suggestion
def t_index(t):
```
? I'm surprised the linter didn't complain...

skyportal/handlers/api/internal/source_views.py
```diff
@@ -49,15 +48,20 @@ def get(self):
...
-            info = first_thumbnail_info(s.thumbnails)
             sources.append(
                 {
                     'obj_id': s.id,
                     'views': view,
                     'ra': s.ra,
                     'dec': s.dec,
-                    'public_url': info[0],
-                    'is_grayscale': info[1],
+                    'thumbnails': [
```
dmitryduev, line 57: 👍🏻

dmitryduev (approved):
LGTM @acrellin! Just one minor comment.

skyportal/handlers/api/internal/recent_sources.py
```diff
@@ -4,6 +4,7 @@
 from baselayer.app.access import auth_or_token
 from ...base import BaseHandler
 from ....models import Obj, Source
+from .source_views import tIndex
```
stefanv, line 7: Camelcase in Python land!  Should become t_index

static/js/components/RecentSources.jsx
```diff
@@ -137,7 +137,31 @@ const RecentSourcesList = ({ sources, styles }) => {
...
+          // Use first thumbnail available
+          const fetchThumbnail = async (index) => {
+            const response = await fetch(source.thumbnails[index]?.public_url, {
+              mode: "no-cors",
+            });
+            return response.status;
+          };
+          let idx = 0;
+          let statusCode = 0;
+          while (idx < source.thumbnails.length - 1) {
```
stefanv, line 149: In general, while loops are dangerous and overly verbose and should be replaced by for-loops.

In JavaScript, this loop can be replaced by [`.each`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every), for example.  Alternatively, you can use a for loop and break out of it.

static/js/components/TopSources.jsx
```diff
@@ -114,7 +114,31 @@ const TopSourcesList = ({ sources, styles }) => {
...
+          let idx = 0;
+          let statusCode = 0;
+          while (idx < source.thumbnails.length - 1) {
+            if (
+              source.thumbnails[idx]?.public_url &&
+              source.thumbnails[idx].public_url.startsWith("/static")
+            ) {
+              break;
+            }
+            statusCode = fetchThumbnail(idx);
```
stefanv, line 133: Shoudn't an await be used here?

stefanv (commented):
Looks good, thanks Ari.  I have some small requests.

skyportal/handlers/api/internal/recent_sources.py
acrellin: Yes, agreed (I didn't author this, just moved it around, as you can see from the diff -- will update anyways).

stefanv: 👍

static/js/components/RecentSources.jsx
acrellin: I don't think it's possible to break out of a [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) loop (I assume that's what you meant). I can replace it with a for loop if you'd like.

stefanv: I meant a `.each` loop---link in comment above.

acrellin: That links to `.every`

acrellin: Which doesn't really apply to this use case

stefanv: You can use `.some` or `.every`.

stefanv:
```
source.thumbnails.some((thumb) =>
  ...
  return statusCode == 200;
);
```

profjsb:
@stefanv @acrellin is this PR still relevant? If so, let's get it merged.

dmitryduev:
@profjsb it is relevant, but as I recall @acrellin wanted to finish something here.

acrellin:
@stefanv ready for review.  I switched to simply incrementing the index used to key into `source.thumbnails` if a given index results in an error when fetching the image.  The server returns the thumbnail list in the desired order.

acrellin:
@stefanv ready for review when you have a moment.

stefanv:
Let's give it a try!

PR #2135, opened 2021-07-22
WIP: Introduce prototype skymap plans API

mcoughlin:
Depends on: https://github.com/skyportal/skyportal/pull/2133


skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
+import datetime
+import os
+import copy
+
+from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.orm import joinedload
+from astropy import time
+import ephem
+import gwemopt.utils
```
acrellin, line 9: These all need to be imported explicitly? You can't just `import gwemopt` once and then access all of these from the top-level import?

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+    'schedule_strategy': 'tiling',
+    'mindiff': 30.0 * 60.0,
+    'doMaxTiles': False,
+    'max_nb_tiles': 1000,
+    'doRASlice': True,
+    'raslice': [0.0, 24.0],
+}
+
+
+class PlanHandler(BaseHandler):
```
acrellin, line 57: Handler methods need docstrings

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+        localization_name = data.get('localization_name')
+
+        telescope = (
+            Telescope.query_records_accessible_by(
+                self.current_user,
+            )
+            .filter(
+                Telescope.name == telescope_name,
+            )
+            .one()
```
acrellin, line 75: probably better to use `first()` so we can gracefully catch when there's no match and manually return an error message, rather than having an uncaught exception add a messy TB to the logs

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+
+        instrument = (
+            Instrument.query_records_accessible_by(
+                self.current_user,
+            )
+            .filter(
+                Instrument.telescope == telescope,
+                Instrument.name == instrument_name,
+            )
+            .one()
```
acrellin, line 86: Same here as above

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+        )
+
+        localization = (
+            Localization.query_records_accessible_by(self.current_user)
+            .filter(
+                Localization.dateobs == dateobs,
+                Localization.localization_name == localization_name,
+            )
+            .first()
```
acrellin, line 95: let's make sure this actually returned a matching record

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+            instrument_id=instrument.id,
+            planned_observation_id=ii,
+            plan_name=plan_name,
+            overhead_per_exposure=overhead_per_exposure,
+        )
+
+        DBSession().add(planned_observation)
+    plan.status = plan.Status.READY
+    DBSession().merge(plan)
+    request_handler.verify_and_commit()
```
acrellin, line 450: let's move this line out to the main handler method definition

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+
+    if params["doPlots"]:
+        gwemopt.plotting.skymap(params, map_struct)
+        gwemopt.plotting.tiles(params, map_struct, tile_structs)
+        gwemopt.plotting.coverage(params, map_struct, coverage_struct)
+
+    return map_struct, tile_structs, coverage_struct
+
+
+def create_plan(
```
acrellin, line 310: docstring

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+    params = gwemopt.utils.params_checker(params)
+
+    if params["doPlots"]:
+        if not os.path.isdir(params["outputDir"]):
+            os.makedirs(params["outputDir"])
+
+    return params
+
+
+def gen_structs(params):
```
acrellin, line 277: docstring

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+            .filter(
+                ObservingPlan.dateobs == dateobs,
+            )
+            .all()
+        )
+
+        return self.success(data=plans)
+
+
+def params_struct(
```
acrellin, line 125: docstring

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+    if params["doPlots"]:
+        if not os.path.isdir(params["outputDir"]):
+            os.makedirs(params["outputDir"])
+
+    return params
+
+
+def gen_structs(params):
+
+    print('Loading skymap')
```
acrellin, line 279:
```suggestion
```

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+    # Function to read maps
+    map_struct = gwemopt.utils.read_skymap(
+        params, is3D=params["do3D"], map_struct=params['map_struct']
+    )
+    print('Generating MOC struct')
```
acrellin, line 284:
```suggestion
```

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+            tile_structs,
+            previous_coverage_struct=params["previous_coverage_struct"],
+        )
+    else:
+        tile_structs, coverage_struct = gwemopt.coverage.timeallocation(
+            params, map_struct, tile_structs
+        )
+
+    if params["doPlots"]:
+        gwemopt.plotting.skymap(params, map_struct)
```
acrellin, line 303: so these calls are modifying the arguments?

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,450 @@
...
+        plan = ObservingPlan(
+            dateobs=localization.dateobs,
+            plan_name=plan_name,
+            instrument_id=instrument.id,
+            validity_window_start=validity_window_start,
+            validity_window_end=validity_window_end,
+            plan_args=plan_args,
+        )
+        DBSession().add(plan)
+        request_handler.verify_and_commit()
```
acrellin, line 367: possible to hold off until the very end before committing?

skyportal/models.py
```diff
@@ -1758,6 +1759,231 @@ def api_class(self):
...
+        sa.Enum(Status), default=Status.WORKING, nullable=False, doc='Plan status'
+    )
+
+    planned_observations = relationship("PlannedObservation")
+
+    @property
+    def start_observation(self):
+        """Time of the first planned observation."""
+        if self.planned_observations:
+            return self.planned_observations[0].obstime
```
acrellin, line 1870: shouldn't we be sorting planned observations by `obstime` to find the earliest one? or will they all have same `obstime`?

skyportal/models.py
```diff
@@ -1758,6 +1759,231 @@ def api_class(self):
...
+
+    def get_probability(self, localization):
+        ipix = np.asarray(list(self.ipix))
+        if len(ipix) > 0:
+            return localization.flat_2d[ipix].sum()
+        else:
+            return 0.0
+
+
+class PlannedObservation(Base):
```
acrellin, line 1933: Should these be named more explicitly to distinguish them from the existing `ObservingRun`, etc, classes?

acrellin (commented):
Looking good here, too! A few comments/questions below. Thanks!

skyportal/handlers/api/plan.py
mcoughlin: Good call!

skyportal/handlers/api/plan.py
mcoughlin: yes!

skyportal/handlers/api/plan.py
mcoughlin: definitely.

skyportal/handlers/api/plan.py
mcoughlin: ah yes

skyportal/handlers/api/plan.py
mcoughlin: yes will fix.

skyportal/handlers/api/plan.py
mcoughlin: but how does that work if it is asynchronous?

skyportal/handlers/api/plan.py
mcoughlin: We need to tell the code we started making the plan so it is PENDING.

skyportal/models.py
mcoughlin: oh yeah!

skyportal/models.py
mcoughlin: Suggestion? They do have Planned in the title.

skyportal/handlers/api/plan.py
acrellin: ah right, nm then

skyportal/handlers/api/plan.py
acrellin: okay, yeah makes sense for polling purposes.

skyportal/models.py
acrellin: Well these are quite distinct from followup requests, right? Something to disambiguate would be nice

skyportal/handlers/api/plan.py
mcoughlin: This was dumb. I am removing.

skyportal/models.py
mcoughlin: ObservingPlan -> EventObservingPlan ?

skyportal/handlers/api/instrument.py
```diff
@@ -1,13 +1,18 @@
...
-from ...models import DBSession, Instrument, Telescope
+from ...models import DBSession, Instrument, InstrumentField, Telescope
 from ...enum_types import ALLOWED_BANDPASSES
+from ...utils.instrument import get_mocs
 
 
 class InstrumentHandler(BaseHandler):
     @permissions(['System admin'])
-    def post(self):
+    async def post(self):
```
stefanv, line 15: Why is this function async?

skyportal/handlers/api/instrument.py
```diff
@@ -17,6 +22,15 @@ def post(self):
             telescope_id, self.current_user, raise_if_none=True, mode="read"
         )
 
+        if "field_data" in data:
+            field_data = data.pop("field_data")
```
stefanv, line 26: Is it verified that these fields will be in data?

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+          - observingplans
+        requestBody:
+          content:
+            application/json:
+              schema: PlanHandlerPost
+        responses:
+          200:
+            content:
+              application/json:
+                schema: Success
```
stefanv, line 61: This is incomplete.

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+
+        data = self.get_json()
+
+        telescope_name = data.get('telescope_name')
```
stefanv, line 70: Error checking for if fields are not available.

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+            .filter(
+                Localization.dateobs == dateobs,
+                Localization.localization_name == localization_name,
+            )
+            .first()
+        )
+        if localization is None:
+            self.error(message=f"Missing localization {localization_name}")
+
+        plan_args = copy.deepcopy(plan_args_default)
```
stefanv, line 111: What is this needed for?

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+
+        create_plan(self, instrument, localization, **plan_args)
+
+    @auth_or_token
+    def get(self):
+        """
+        ---
+          description: Retrieve observing plans
```
stefanv, line 119: Indentation?

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+                  schema: ArrayOfObservingPlan
+            400:
+              content:
+                application/json:
+                  schema: Error
+        """
+
+        data = self.get_json()
+
+        dateobs = data.get('dateobs')
```
stefanv, line 141: Error checking for when `dataobs` isn't specified

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+):
+    """Set gwemopt parameter dictionary for observing plan."""
+
+    observer = ephem.Observer()
+    observer.lat = str(instrument.telescope.lat)
+    observer.lon = str(instrument.telescope.lon)
+    observer.horizon = str(-12.0)
+    observer.elevation = instrument.telescope.elevation
+
+    params = {}
```
stefanv, line 190:
```
params = {
    'config': {
        instrument.name: {
            'tesselation': ...,
            ...
    }
}
```

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+    params["config"][instrument.name]["overhead_per_exposure"] = 0.0
+
+    params["skymap"] = ""
+    params["gpstime"] = -1
+    params["outputDir"] = "output/%s" % dateobs.strftime("%Y%m%dT%H%M%S")
+    params["tilingDir"] = ""
+    params["event"] = ""
+    params["telescopes"] = [instrument.name]
+    if schedule_strategy == "catalog":
+        params["tilesType"] = "galaxy"
```
stefanv, line 212:
```
    if schedule_strategy == "catalog":
        params = {**params, 'tilesType': 'galaxy', ... }
```

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+    elif params["doFootprint"]:
+        params["skymap"] = gwemopt.footprint.get_skymap(params)
+        event_time = time.Time(params["gpstime"], format='gps', scale='utc')
+        params["dateobs"] = event_time.iso
+    elif params["doDatabase"]:
+        event_time = time.Time(params["dateobs"], format='datetime', scale='utc')
+        params["gpstime"] = event_time.gps
+    else:
+        raise ValueError(
+            'Need to enable --doEvent, --doFootprint, ' '--doSkymap, or --doDatabase'
```
stefanv, line 279: Do you have control over these flags?  If so, you may rather want something like `--action=event --action=footprint --action=database` or similar.

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+    if tobs is None:
+        now_time = time.Time.now()
+        timediff = now_time.gps - event_time.gps
+        timediff_days = timediff / 86400.0
+        params["Tobs"] = np.array([timediff_days, timediff_days + 1])
+    else:
+        params["Tobs"] = tobs
+
+    params["doSingleExposure"] = True
+    if filterScheduleType == "block":
```
stefanv, line 291: params["doAlternativeFilters"] = (filtersScheduleType == "block")

skyportal/handlers/api/plan.py
stefanv: If you want to log this explicitly, take a look at how the cache.py does it, e.g..

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+def gen_structs(params):
+    """Use gwemopt to create observing plan for specified instrument
+    and time window."""
+
+    print('Loading skymap')
+    # Function to read maps
+    map_struct = gwemopt.utils.read_skymap(
+        params, is3D=params["do3D"], map_struct=params['map_struct']
```
stefanv, line 315: I don't see `map_struct` anywhere else; how do we know what's supposed to go in there?

skyportal/handlers/api/plan.py
```diff
@@ -0,0 +1,478 @@
...
+            .filter(
+                ObservingPlan.dateobs == dateobs,
+            )
+            .all()
+        )
+
+        return self.success(data=plans)
+
+
+def params_struct(
```
stefanv, line 159: It looks like this function mainly takes keyword arguments and turn them into a dictionary. If so, why not just use a dictionary?

skyportal/models.py
```diff
@@ -1758,6 +1759,236 @@ def api_class(self):
...
+            for i in planned_observation.field.ipix
+        }
+
+    @property
+    def area(self):
+        nside = Localization.nside
+        return hp.nside2pixarea(nside, degrees=True) * len(self.ipix)
+
+    def get_probability(self, localization):
+        ipix = np.asarray(list(self.ipix))
```
stefanv, line 1931: What does `self.ipix` start off as?

skyportal/models.py
```diff
@@ -1758,6 +1759,236 @@ def api_class(self):
...
+        if len(ipix) > 0:
+            return localization.flat_2d[ipix].sum()
+        else:
+            return 0.0
```
stefanv, line 1935: Why? `np.sum([]) == 0`

skyportal/utils/instrument.py
```diff
@@ -0,0 +1,149 @@
+# Inspired by: https://github.com/growth-astro/growth-too-marshal/blob/main/growth/too/models.py and
```
stefanv, line 1: and ... ?

skyportal/utils/instrument.py
```diff
@@ -0,0 +1,149 @@
...
+from mocpy import MOC
+
+import gwemopt.ztf_tiling
+
+from baselayer.app.env import load_env
+
+_, cfg = load_env()
+
+
+def get_ztf_quadrants():
```
stefanv, line 20: All these functions need tests.

skyportal/handlers/api/instrument.py
```diff
@@ -1,13 +1,18 @@
...
 from ..base import BaseHandler
-from ...models import DBSession, Instrument, Telescope
+from ...models import DBSession, Instrument, InstrumentField, Telescope
 from ...enum_types import ALLOWED_BANDPASSES
+from ...utils.instrument import get_mocs
 
 
 class InstrumentHandler(BaseHandler):
     @permissions(['System admin'])
-    def post(self):
```
stefanv, line 10: All handlers need tests.

stefanv (commented):
@mcoughlin I did a quick review with some general pointers; no rush to implement them all immediately, I know these PRs take shape gradually as all the layers come into play.

skyportal/handlers/api/instrument.py
mcoughlin: This is required for having the asynchronous call to the fields or not?

skyportal/handlers/api/plan.py
mcoughlin: So when down the line, we make the parameters configurable through the interface, some of these arguments will change, but maybe a copy isnt necessary? I just get nervous whenever I see something like 
```plan_args = plan_args_default''''
and similar, but maybe it is fine here?

File: skyportal/models.py (hunk 29 above)
Comment (mcoughlin) [reply skyportal/models.py:1931]: It is a set currently.

File: skyportal/models.py (hunk 30 above)
Comment (mcoughlin) [reply skyportal/models.py:1935]: ok can remove!

File: skyportal/handlers/api/plan.py (hunk 1 above)
Comment (mcoughlin) [reply skyportal/handlers/api/plan.py:8]: Ok hmmm anything but specifically calling gwemopt.utils complains it doesn't know how to get gwemopt.utils.blah... I guess I can try to fix that in gwemopt packaging. I guess there is a standard way to do that?

File: skyportal/handlers/api/plan.py (hunk 27 above)
Comment (mcoughlin) [reply skyportal/handlers/api/plan.py:242]: we currently define it in create_plan... it doesn't need to be set in that dictionary I suppose.

File: skyportal/handlers/api/instrument.py (hunk 16 above)
Comment (stefanv) [reply skyportal/handlers/api/instrument.py:15]: I don't think so, unless you use an `await` inside.

File: skyportal/handlers/api/plan.py (hunk 20 above)
Comment (stefanv) [reply skyportal/handlers/api/plan.py:121]: If you are going to modify `plan_args` it makes sense to do a copy.  Deepcopy is usually reserved for instances where you have objects inside the dict that also need to be copied carefully.

Also note this handy new syntax:

```
In [1]: d = {'a': 4, 'b': 6}

In [2]: b = {**d}

In [3]: b['a'] = 5

In [4]: d
Out[4]: {'a': 4, 'b': 6}

In [5]: b
Out[5]: {'a': 5, 'b': 6}
```

Comment (stefanv) [reply skyportal/handlers/api/plan.py:121]: Let me try and be a bit more clear: the deepcopy handles the deeper nested dictionaries, so you don't run into this issue:

```
In [12]: d = {'a': 4, 'b': 6}

In [13]: d = {'a': 4, 'b': 6, 'c': {'foo': 'bar'}}

In [14]: e = {**d}

In [15]: e['c']['foo'] = 'something'

In [16]: d
Out[16]: {'a': 4, 'b': 6, 'c': {'foo': 'something'}}

```

skyportal/handlers/api/instrument.py
mcoughlin: Ok I see. Removed.

skyportal/utils/instrument.py
mcoughlin: Yes added on the other PR.

skyportal/handlers/api/instrument.py
mcoughlin: Yes added on the other PR.

skyportal/handlers/api/plan.py
mcoughlin: Yes added on the other PR.

skyportal/handlers/api/plan.py
mcoughlin: Because error isn't supposed to be a catch all?

skyportal/handlers/api/instrument.py
mcoughlin: Yes added on the other PR.

skyportal/handlers/api/plan.py
stefanv: The Success object needs to be augmented with any additional return fields.

mcoughlin: Got it! ok added.

skyportal/handlers/api/plan.py
mcoughlin: That first example was terrifying ;)

skyportal/handlers/api/plan.py
mcoughlin: Yup!

skyportal/handlers/api/plan.py
mcoughlin: Yes!

skyportal/handlers/api/plan.py
mcoughlin: This was nice.

skyportal/handlers/api/plan.py
mcoughlin: This too.

profjsb:
@mcoughlin please update this issue...have the dependencies been satisfied/clarified?

PR #2136, opened 2021-07-22
Allow free-text observers/reducers for uploading spectra

kmshin1397:
This PR exposes the option to provide free-text for observers and reducers when uploading spectra. If providing free text (via new `external_reducer` and `external_observer` API parameters) valid user IDs must still be provided via the existing `reduced_by` and `observed_by` fields to be used as points of contact for the reduction/observation for citation purposes. This is reflected in the front-end form as well. 

Free-text associate with a spectrum is stored in a new column added to the `SpectrumReducer` and `SpectrumObserver` join tables to minimize the disruptiveness of the migration. That is, I chose to avoid touching the existing relationships to `User` instead of making `reducers` and `observers` string columns, for example, to avoid complicating migrations.

Closes fritz-marshal/fritz-beta-feedback#74

freetext


dmitryduev (approved):
A very elegant solution, LGTM!

skyportal/models.py
```diff
@@ -2777,6 +2777,20 @@ def from_ascii(
 
 # should be accessible only by spectrumowner ^^
 
+SpectrumReducer.external_reducer = sa.Column(
+    sa.String,
+    nullable=True,
+    doc="The actual reducer, provided as free text, for the spectrum, separate "
+    "from the point-of-contact Fritz user designated as reducer",
```
acrellin, line 2784: Maybe add something like "if the reducer is not a user in the database" for clarity?
Below as well

acrellin (approved):
Nice, thanks @kmshin1397 . Impressive RJSF wizardry 😉

dmitryduev:
CI failures seem unrelated, merging.

PR #2137, opened 2021-07-22
Convert CSS files to inline MUI makeStyles()

### LINKED ISSUE #676: Replace CSS with CSS-IN-JSX for MUI
> I see you use both `makeStyles` and `css` in this PR; `makeStyles` is preferred.  However, since we have a mish-mash at the moment, you can leave this as is if you prefer.

_Originally posted by @stefanv in https://github.com/skyportal/skyportal/pull/591_

We should move everything that is currently in CSS to CSS-in-JSX as recommended by the MUI best practices / styles sheet:

https://material-ui.com/styles/basics/

kmshin1397:
This PR deletes the remaining CSS files and moves that styling into inline Material UI `makeStyles()` functions. 

Also made some small tweaks to styles on the Groups page (unifying `Paper` component variants for the various widgets on that page). 

Closes #676


acrellin (approved):
Nice, thanks @kmshin1397

stefanv (approved):

PR #2146, opened 2021-07-26
Split up GA test runs across multiple builds using matrix

### LINKED ISSUE #2143: Split up test suite across multiple GA builds for parallelization
Split up front-end tests across at least two GA builds

acrellin:
This PR reduces the time for GA builds to complete by two thirds (from ~60 minutes to ~20 minutes) by breaking up tests runs into multiple GA workflows run in parallel.

Closes #2143


.github/workflows/test_frontend_part1.yaml
```diff
@@ -0,0 +1,162 @@
...
+          python-version: "3.8"
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache
+            ~/.local
+            ~/.npm
+            cache
+          key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
```
stefanv, line 43: It may be worth figuring out the correct caching key; I'm not sure our `key` is correct, and I imagine it costs us not to utilize the cache properly.

.github/workflows/test_frontend_part2.yaml
```diff
@@ -0,0 +1,162 @@
+name: Test Skyportal Front-end Part 2
```
stefanv, line 1: Instead of duplicating the file, it should be easier to define a [matrix](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix).

.github/workflows/test_models_part2.yaml
```diff
@@ -0,0 +1,156 @@
+name: Test Skyportal Model Logic - pt 2
```
stefanv, line 1: The model logic runs in <20 mins, so probably easiest to keep these together for now.

acrellin: It was actually like 40 minutes on the last build I checked.

.github/workflows/test_frontend_part1.yaml
acrellin: I'd prefer to leave that for another PR, as it's unrelated to this one.

.github/workflows/test_frontend_part2.yaml
acrellin: I'll look into that!

.github/workflows/test_models_part2.yaml
acrellin: e.g. https://github.com/skyportal/skyportal/runs/3167096342

stefanv: I missed the container initialization; 15 minutes?! That's nuts.

stefanv: Playing around with some options in https://github.com/skyportal/skyportal/pull/2148

stefanv: Down to 15s :man_shrugging:

stefanv: The run above might have been a fluke; looks like that step doesn't usually take that long.

.github/workflows/test_api_and_frontend.yaml
```diff
@@ -6,6 +6,10 @@ on:
       - master
   pull_request:
 
+strategy:
+  matrix:
+    test_subset: ["api_and_utils", "frontend_pt1", "frontend_pt2"]
```
stefanv, line 11: Should these have quotes?

.github/workflows/test_api_and_frontend.yaml
```diff
@@ -6,6 +6,10 @@ on:
       - master
   pull_request:
 
+strategy:
```
stefanv, line 9: I think where you had it before was correct.

.github/workflows/test_api_and_frontend.yaml
acrellin: I'll try removing the quotes

.github/workflows/test_api_and_frontend.yaml
acrellin: Moved back to where it was

acrellin:
@stefanv hmm, still no dice

stefanv (approved):

acrellin:
Sole failure in `api_and_utils` build is unrelated: 
```
====== 1 failed, 318 passed, 1 xpassed, 219 warnings in 703.58s (0:11:43) ======
```

@kmshin1397 Ready for review/merge when you have a moment.

stefanv:
Beautiful!  24, 21, and 17 mins for frontend pt 1, pt 2, and API tests!

kmshin1397 (approved):
Looks good @acrellin! Out of curiosity, did you decide to split it like this since the sources and followup tests are the ones with external dependencies?

acrellin:
> Looks good @acrellin! Out of curiosity, did you decide to split it like this since the sources and followup tests are the ones with external dependencies?

I wasn't aware that they had external dependencies :) 
It was pretty arbitrary; I just started with all tests for the source page (followup request form is also there), and just tried to balance the split somewhat evenly while keeping somewhat of a logical grouping

kmshin1397:
> > Looks good @acrellin! Out of curiosity, did you decide to split it like this since the sources and followup tests are the ones with external dependencies?
> 
> I wasn't aware that they had external dependencies :) What do you mean by that?
> It was pretty arbitrary; I just started with all tests for the source page (followup request form is also there), and just tried to balance the split somewhat evenly while keeping somewhat of a logical grouping

Ah okay that makes sense. I meant that I think those files are the ones that include tests depending on external APIs, like the instrument APIs for followup and the source offsets (I think those are in the test_sources not sure). So I think this split also happens to put the tests which can be flaky due to outside stuff together as well :)

acrellin:
Ah, okay, yeah that makes sense. Entirely by accident, but I guess that works!  :)

PR #2148, opened 2021-07-27
Try using tmpfs to speed up postgres

stefanv:


acrellin:
@stefanv I rebased this branch and re-pushed to trigger new GA builds to be able to compare build times with recent PRs/master commits

acrellin:
~23 minutes for the longest run; similar to before, but couldn't hurt to merge this I suppose. Think we should go forward with it @stefanv ?

stefanv:
I'm going to close for now, since it doesn't have the impact I had hoped.

Issue #1266, opened 2020-11-03
Add binning of galaxy spectra 


dannygoldstein:
On the spectroscopy bokeh plot, add a slider bar that allows you to bin spectra in wavelength space

AlexGKim:
a +1 for this issue.

guynir42:
This will be important for DESI spectra.

guynir42:
Closed by #2486

PR #2149, opened 2021-07-27
Binning of spectroscopy plots

kmshin1397:
Implement a slider to apply binning to spectroscopy plots on the Source page. 
binning

Closes fritz-marshal/fritz-beta-feedback#109
Closes skyportal/skyportal#1266


static/js/plotjs/stackSpectra.js
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable */
```
dmitryduev, line 1: Is this to bypass the import error or something else?

dmitryduev (approved):
LAwesomeTM!

static/js/plotjs/stackSpectra.js
kmshin1397: Hmm might be for undeclared variable? Since `bin_slider` is passed in by bokeh and isn't defined in that file. I'm not actually sure - that line is left over from copy-pasting the logic for the photometry binning :)

stefanv: Better to disable individual lines as needed.

stefanv: Can we reuse binning from photometry, or are they too different?

static/js/plotjs/stackSpectra.js
```diff
@@ -0,0 +1,69 @@
...
+const binsize = bin_slider.value;
+const fluxalph = binsize === 0 ? 1.0 : 0.1;
+
+for (let i = 0; i < n_labels; i++) {
+  const spectra = eval(`s${i}`);
```
stefanv, line 6: Is this eval necessary?

static/js/plotjs/stackSpectra.js
```diff
@@ -0,0 +1,69 @@
...
+  const fluxsource = spectra.data_source;
+  const binsource = eval(`bin${i}`).data_source;
```
stefanv, line 8: Same question here.

stefanv (commented):
Some minor comments, otherwise looks good to me.

dmitryduev:
@kmshin1397 would you anticipate finishing this off, or should someone else so that?

kmshin1397:
> @kmshin1397 would you anticipate finishing this off, or should someone else so that?

I can finish this off no problem

static/js/plotjs/stackSpectra.js
kmshin1397: I'll look into alternatives, but this how we've always accessed bokeh plot data sources within custom JS callbacks and I'm not aware of alternatives off the top of my head. Agreed it would be preferable not to use it

static/js/plotjs/stackSpectra.js
kmshin1397: The binning from photometry is fairly different. Namely, the photometry binning must deal with both the photometry points and the associated error bars when replotting based on binning changes, whereas the spectroscopy only cares about the one line. Also, the photometry is a scatter plot where the spectroscopy is a step function so that also slightly alters the data access methodology. I think it would be simpler to just keep them separate.

static/js/plotjs/stackSpectra.js
kmshin1397: Ok, I think I have a clean alternative, will update soon

kmshin1397: Ok @stefanv I believe the linting is much improved, not only here but in all the other Bokeh custom JS stuff. I chose to ignore the `no-undef` rule on a per-file basis because we often call variables dynamically passed in by Bokeh but not visible by ESLint when parsing the custom JS scripts. So that would just be a lot of lines with the "ignore this line because this variable will be there at runtime" behavior.

stefanv: Perfect! Thank you @kmshin1397.

stefanv (approved):
@kmshin1397 As always, your review response exceeds expectation.  We will miss you very much!  Thanks again, and good luck with everything.

dmitryduev:
Thank you, @kmshin1397! Can't add much to what @stefanv said...
CI failures seem unrelated, merging.

dmitryduev:
@stefanv ping!

acrellin:
@stefanv Friendly ping again :)

profjsb:
@guynir42 will check if this conflicts with his new PR

profjsb:
Superceded by #2486

PR #2155, opened 2021-07-29
GA: fix caching of pip and npm

stefanv:
Previously, the npm cache key was generated from **/lockfiles, which we don't use.

This cache key is also not a valid way of invalidating the pip cache.  There are now two caches:
one for NPM (keyed with package.json) and one for pip (keyed with requirements.txt).


stefanv:
Not working correctly yet.

PR #2159, opened 2021-07-30
Reduce big phot post test size from 50k points to 40k points

acrellin:
With 50k points, the test consistently fails on CI.


dmitryduev (approved):
👍🏻

PR #2187, opened 2021-08-04
Photometry SNR bug hotfix: Ensure phot.fluxerr is non-zero number before computing SNR

acrellin:
We were not ensuring that `phot.fluxerr` was a non-zero number before attempting to compute SNR.


acrellin:
test failure unrelated (big_phot_post timeout)

dmitryduev (approved):
Thank you so much for finding and fixing this @acrellin!!

PR #2188, opened 2021-08-05
Allow users with System admin or Become user ACLs to become users when not in debug mode

acrellin:
This fixes a bug in the logic that didn't allow a user with the "Become user" ACL to become another user when not in debug mode, and now also allows users with the "System admin" ACL to do the same.

This PR also updates the GitHub actions workflows to only test `load_demo_data` in one workflow (the model tests), and bumps the data loader timeout to 3s.

### DIFF
--- a/skyportal/handlers/become_user.py
+++ b/skyportal/handlers/become_user.py
@@ -6,12 +6,9 @@ class BecomeUserHandler(BaseHandler):
     def get(self, new_user_id=None):
         if not (
             self.cfg['server.auth.debug_login']
-            or len(
-                {'System admin', 'Become user'}.intersection(
-                    set(self.current_user.permissions)
-                )
+            or {'System admin', 'Become user'}.intersection(
+                set(self.current_user.permissions)
             )
-            > 0
         ):
             return self.error("Insufficient permissions")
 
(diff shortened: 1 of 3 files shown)


skyportal/handlers/become_user.py
```diff
@@ -1,23 +1,29 @@
...
-        if (
+        if not (
             self.cfg['server.auth.debug_login']
-            or ACL.query.get('Become user') in self.current_user.permissions
+            or len(
+                {'System admin', 'Become user'}.intersection(
+                    set(self.current_user.permissions)
+                )
+            )
+            > 0
```
stefanv, line 14: The `len > 0` is probably superfluous.

stefanv (commented):
Looks good, thanks Ari!  There seems to be two unrelated changes here (test_api_and_frontend.yaml and data_loader.py).

skyportal/handlers/become_user.py
acrellin: Updated

acrellin:
@stefanv yes, the data loader timeout and deleting the superfluous test were related to getting this PR green. Can we include them here (they're in the PR description) or would you like me to break them out to separate PRs?

stefanv:
Fine to include them here if they're relevant.  Go ahead and merge on green.

acrellin:
Failure unrelated: ` FAILED skyportal/tests/api/test_photometry.py::test_token_user_big_post`

PR #2191, opened 2021-08-09
Pin latest baselayer

stefanv:
Also update data loader timeouts to be a bit more robust


dmitryduev (approved):
Unrelated failure in `test_get_ztfref_url`, merging.

PR #2192, opened 2021-08-10
Pin baselayer

stefanv:


stefanv:
Some real issues identified.  Not so much with baselayer, but with endpoints in SP itself.  Do not merge.

PR #2197, opened 2021-08-12
Fix spectrum comment previews

### LINKED ISSUE #2200: Spectrum comment attachments not found
A spectrum comment attachment uploaded via the API shows as missing:

image

This is being addressed by [PR#2197](https://github.com/skyportal/skyportal/pull/2197).

stefanv:
- Modify the app routing regexps to ensure that
  `/api/comments/10/attachment/spectrum` is reachable
- Return a sensible error message if no attachment exists
- Modify CommentAttachmentPreview to fetch attachment from
  the appropriate endpoint for either object or spectrum comment

Closes #2200

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -111,13 +111,19 @@ skyportal_handlers = [
     (r'/api/candidates(/[0-9A-Za-z-_]+)/([0-9]+)', CandidateHandler),
     (r'/api/candidates(/.*)?', CandidateHandler),
     (r'/api/classification(/[0-9]+)?', ClassificationHandler),
-    (r'/api/comment', CommentHandler),
-    (r'/api/comment(/[0-9]+)/attachment', CommentAttachmentHandler),
+    (
+        r'/api/comment(/[0-9]+)/attachment(/(?:object|spectrum))?',
+        CommentAttachmentHandler,
+    ),
     # Allow the '.pdf' suffix for the attachment route, as the
     # react-file-previewer package expects URLs ending with '.pdf' to
     # load PDF files.
-    (r'/api/comment(/[0-9]+)/attachment.pdf', CommentAttachmentHandler),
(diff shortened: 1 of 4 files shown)


skyportal/app_server.py
```diff
@@ -111,13 +111,19 @@ def log_request(self, handler):
...
+    ),
     # Allow the '.pdf' suffix for the attachment route, as the
     # react-file-previewer package expects URLs ending with '.pdf' to
     # load PDF files.
-    (r'/api/comment(/[0-9]+)/attachment.pdf', CommentAttachmentHandler),
-    (r'/api/comment(/[0-9]+)?(/.+)?', CommentHandler),
+    (
+        r'/api/comment(/[0-9]+)/attachment.pdf(/(?:object|spectrum))?',
+        CommentAttachmentHandler,
+    ),
```
guynir42, line 124: Does the ?: in the parenthesis do anything in this context or is it just faster to match things?

guynir42 (approved):
Looks good!

skyportal/app_server.py
stefanv: The `?:` means it's a non-matching group.  I needed to specify `(object|spectrum)`, which I'm not sure how to do without brackets.  The `?:` is to say: don't try and *use* those brackets.  I *think* I checked and saw that removing that breaks things, but may be worth double checking.

scizen9 (approved):
Thanks for the quick turnaround.  My only question concerns the pdf specific code as most of the SEDM attachments will be png images.  I'm guessing png is handled by default.

scizen9:
Just added issue #2200 that can be closed when this is merged.

stefanv:
> Thanks for the quick turnaround. My only question concerns the pdf specific code as most of the SEDM attachments will be png images. I'm guessing png is handled by default.

PNG is handled, yes.  We need an endpoint that ends on `.pdf` for browsers to recognize PDF downloads.

PR #2207, opened 2021-08-13
On first run, check out committed version of baselayer

stefanv:
Before, we used to check out the HEAD of the baselayer repository.


dmitryduev (approved):
Huh. I wonder how we never noticed this when we had breaking changes in BL?..

stefanv:
Luckily, the CI checked out the submodules for us, and the Makefile doesn't touch them if they're already there.

PR #2210, opened 2021-08-13
Fix airmass plot

mcoughlin:
Addresses https://github.com/fritz-marshal/fritz/issues/333

The code was missing a check in case sunrise was after sunset to ensure you have positive delta t. Also increased sampling slightly to smooth out behavior.


dmitryduev (approved):
Thanks for the quick fix @mcoughlin! Please feel free to self-merge once green.

stefanv (approved):

Issue #610, opened 2020-07-29
API endpoints should be plural/singular consistently


stefanv:
Currently, we have `/candidates` but `/comment`.  These should all either be singular or plural.

@acrellin and I think they should all be *plural*.

kmshin1397:
Alias with deprecation warning?

stefanv:
Has this been resolved? I thought I still saw this recently.

mcoughlin:
I just don't see us managing to do this at this point.

PR #2212, opened 2021-08-16
Comment API URL reorganization

guynir42:
Changed the order of URLs when appending comments and annotations. E.g., 

`api/comment/<comment_id>/object` is replaced with `api/sources/<source_id>/comment/<comment_id>` 
and
`api/comment/<comment_id>/spectrum` to `api/spectrum/<spectrum_id>/comment/<comment_id>`
(the same is applied to annotations, even though we still do not have spectrum annotations)

When posting a new comment/annotation, the object or spectrum ID in the path must match that in the request body. When using get/delete/put the path object/spectrum ID must match the ID of the object/spectrum that the comment is appended to.


guynir42:
This actually required refactoring a whole bunch of stuff. I also fixed some bugs I found along the way.

skyportal/handlers/api/annotation.py
```diff
@@ -30,18 +49,64 @@ def get(self, annotation_id):
...
-        annotation = Annotation.get_if_accessible_by(
-            annotation_id, self.current_user, raise_if_none=True
-        )
+
+        try:
+            annotation_id = int(annotation_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid annotation ID. ")
+        # the default is to annotate an object
+        if associated_resource_type.lower() in (
```
stefanv, line 58: Why four options here?  And when is this value `object` or `objects`?

skyportal/handlers/api/annotation.py
```diff
@@ -30,18 +49,64 @@ def get(self, annotation_id):
...
+            schema:
+              type: string
+            description: |
+               The ID of the underlying data.
+               This would be a string for an object ID
+               or an integer for other data types like spectrum.
+               The object pointed to by this input must be a valid
+               object or other data type (like spectrum) that
+               can be annotated on by the user/token.
+               It must match the data given in the request body.
```
stefanv, line 109: Why specify it twice?  Should be sufficient to grab it from the URL.

skyportal/handlers/api/comment.py
```diff
@@ -59,32 +71,71 @@ def get(self, comment_id, associated_resource_type=None):
...
 
+        if comment_resource_id_str != resource_id:
+            return self.error(
+                f'Comment resource ID does not match resource ID given in path ({resource_id})'
+            )
         return self.success(data=comment)
 
     @permissions(['Comment'])
-    def post(self):
+    def post(self, associated_resource_type, resource_id, *_):
```
stefanv, line 111: When do more keyword arguments land here?

skyportal/handlers/api/comment.py
```diff
@@ -59,32 +71,71 @@ def get(self, comment_id, associated_resource_type=None):
...
+          - in: path
+            name: resource_id
+            required: true
+            schema:
+              type: string
+            description: |
+               The ID of the underlying data.
+               This would be a string for an object ID
+               or an integer for other data types like spectrum.
+               The object pointed to by this input must be a valid
```
stefanv, line 135: I would make this explicit until we go more general: "This is the ID of the object, source, or spectrum the comment is posted to."

skyportal/handlers/api/annotation.py
guynir42: In principle you could comment on an object that was not saved as a source... we don't have an API for that yet but I can imagine needing it in the future.

skyportal/handlers/api/annotation.py
guynir42: Just for consistency. But if you don't think that matters much then we can remove it.

skyportal/handlers/api/comment.py
guynir42: This fits the signature from the tornado server regexp.

stefanv: I guess I'm asking when the `*_` part is used.

skyportal/handlers/api/annotation.py
stefanv: Let's only specify it in one place.

skyportal/handlers/api/annotation.py
stefanv: It's good to think ahead!  In general, we try to have a design that can be expanded in the future, but to implement only the minimal changes necessary.  Can we take out anything that is not used right now?

guynir42: OK

skyportal/handlers/api/annotation.py
guynir42: OK

skyportal/handlers/api/comment.py
guynir42: It is never used, but if I don't add that last input it complains about not having the right number of arguments.

stefanv: That's not right; for the post there should not be a comment ID there.  A comment ID is only provided for requests that work with an existing comment.

Can you confirm that this is problematic when posting to: `/api/sources/ztf123/comment` ?

guynir42: When I remove `*_` as the fourth argument, I get this error when posting from the front-end:

`post() takes 3 positional arguments but 4 were given`

guynir42: The last argument is "None" but it still exists...

guynir42: The regexp for the server is `r'/api/(sources|spectrum)/([0-9A-Za-z-_\.\+]+)/comment(/[0-9]+)?'` which transmits that last input argument even if the last group doesn't contain anything.

guynir42: I can solve this by adding a `(r'/api/(sources|spectrum)/([0-9A-Za-z-_\.\+]+)/comment', CommentHandler)` route. I don't know why I didn't come up with that earlier...

stefanv: Thanks for the explanation; I understand now.  I also found a SO post that discusses the same issue: https://stackoverflow.com/a/58775110/214686

I think your solution (two routes) is better.  It's a pity we won't easily be able to print out an error message for the case where that argument *is* supplied, but I'm not sure how to solve that simply.  If users follow the API docs, they won't run into the issue, so should be OK.

skyportal/handlers/api/annotation.py
```diff
@@ -8,13 +8,32 @@
...
           - annotations
         parameters:
+          - in: path
+            name: associated_resource_type
+            required: true
+            schema:
+              type: string
+            description: |
+               What underlying data the annotation is on:
+               an "object" (default), or a "spectrum".
```
stefanv, line 25: I don't think the user has the option of using a "default" option?  It has to be `source` or `spectrum`, no?

skyportal/handlers/api/annotation.py
```diff
@@ -8,13 +8,32 @@
...
+          - in: path
+            name: resource_id
+            required: true
+            schema:
+              type: string
+            description: |
+               The ID of the underlying data.
+               This would be a string for an object ID
```
stefanv, line 33: I'm confused that objects are still mentioned here, when we are only handling sources and spectra.  If you think this is the best language we'll leave it as is; but it feels confusing to have `/api/sources/...` and then talk about an object, which is more of an internal SkyPortal concept.

skyportal/handlers/api/annotation.py
```diff
@@ -30,26 +49,65 @@ def get(self, annotation_id):
...
         """
-        annotation = Annotation.get_if_accessible_by(
-            annotation_id, self.current_user, raise_if_none=True
-        )
+
+        try:
+            annotation_id = int(annotation_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid annotation ID. ")
+        # the default is to annotate an object
```
stefanv, line 57: There is no default code path here.

skyportal/handlers/api/annotation.py
```diff
@@ -30,26 +49,65 @@ def get(self, annotation_id):
...
                 schema: Error
         """
-        annotation = Annotation.get_if_accessible_by(
-            annotation_id, self.current_user, raise_if_none=True
-        )
+
+        try:
+            annotation_id = int(annotation_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid annotation ID. ")
```
stefanv, line 56: This should return "Annotation not found" with `status=404`.

skyportal/handlers/api/annotation.py
```diff
@@ -30,26 +49,65 @@ def get(self, annotation_id):
...
+            schema:
+              type: string
+            description: |
+               The ID of the underlying data.
+               This would be a string for an object ID
+               or an integer for other data types like spectrum.
+               The object pointed to by this input must be a valid
+               object or other data type (like spectrum) that
+               can be annotated on by the user/token.
+               It must match the data given in the request body.
```
stefanv, line 104: Will this still be specified in the request body?

skyportal/handlers/api/annotation.py
```diff
@@ -176,15 +266,28 @@ def put(self, annotation_id):
...
 
+        try:
+            annotation_id = int(annotation_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid annotation ID. ")
+
+        if associated_resource_type.lower() == "sources":
+            schema = Annotation.__schema__()
+            a = Annotation.get_if_accessible_by(
+                annotation_id, self.current_user, mode="update", raise_if_none=True
```
stefanv, line 278: I presume this exception (`raise_if_none`) translates into an AccessError; could you verify?

skyportal/handlers/api/comment.py
```diff
@@ -59,46 +72,73 @@ def get(self, comment_id, associated_resource_type=None):
...
+        # the default is to comment on an object
+        if associated_resource_type.lower() == "sources":
             comment = Comment.get_if_accessible_by(
                 comment_id, self.current_user, raise_if_none=True
             )
-        elif associated_resource_type.lower() == "spectrum":
+            comment_resource_id_str = str(comment.obj_id)
+
+            # comment.api_path = f'api/sources/{resource_id}/comment/{comment_id}'
+        elif associated_resource_type.lower() in ("spectrum", "spectra"):
```
stefanv, line 88: Why are we supporting two paths here?

skyportal/handlers/api/annotation.py
guynir42: I'm not sure. I think this catches cases where the ID is None or not an integer. Seems to me "Annotation not found" is more relevant when an integer is given but does not fit any row in the database.

skyportal/handlers/api/annotation.py
guynir42: In Baselayer this calls for a `raise AccessError(f"Invalid {cls.__name__} id: {pk}")` so yeah, it looks like it.

skyportal/handlers/api/annotation.py
stefanv: OK, that makes sense.  In that case, specific guidance would be helpful to the user: "Integer expected for annotation ID" or similar.

profjsb (commented):
Small changes. LGTM.

skyportal/handlers/api/comment.py
profjsb: Minor point: the plural of spectrum is spectra. Do we want to harmonize the plural/singular with `sources` in the API call?

stefanv: Yes, we've been wanting to do that for a while.  See https://github.com/skyportal/skyportal/issues/610

I think we should make things consistent, with deprecation aliases in place (in another PR).

skyportal/handlers/api/annotation.py
stefanv: So you need `except AccessError as e:` to handle it.

skyportal/handlers/api/annotation.py
```diff
@@ -122,30 +187,68 @@ def post(self):
...
+            )
 
         DBSession().add(annotation)
         self.verify_and_commit()
-        self.push_all(
-            action='skyportal/REFRESH_SOURCE',
-            payload={'obj_key': annotation.obj.internal_key},
-        )
+
+        if obj_id is not None:
```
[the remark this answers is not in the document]
stefanv: I think this is the correct check; you want to check whether it was specified (not None), and then do some checks to see if it is the right value.  On integers, you cannot simply do `if obj_id` because it might be zero.

skyportal/handlers/api/annotation.py
```diff
@@ -196,20 +317,48 @@ def put(self, annotation_id):
...
+            return self.error(
+                f'Annotation resource ID does not match resource ID given in path ({resource_id})'
+            )
+
         self.verify_and_commit()
-        self.push_all(
-            action='skyportal/REFRESH_SOURCE', payload={'obj_key': a.obj.internal_key}
-        )
+
+        if a.obj.id is not None:  # comment on object, or object related resources
```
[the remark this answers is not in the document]
guynir42: @profjsb besides catching instances of empty string, are there any other differences between the two tests in this context?

skyportal/handlers/api/comment.py
guynir42: I can easily put both spectra and spectrum in the app_server.py and allow the comments/annotations handlers to comply with both, and we can deprecate `spectrum` at some later time. 

What do you think? @profjsb @stefanv

stefanv: As discussed this morning, please keep PRs to features implemented right now, and the minimal changes necessary for that.

guynir42: OK, in that case the PR is basically ready for a final round of reviews (I hope)

skyportal/handlers/api/annotation.py
```diff
@@ -176,40 +270,87 @@ def put(self, annotation_id):
...
           - annotations
         parameters:
+          - in: path
+            name: associated_resource_type
+            required: true
+            schema:
+              type: string
+            description: |
+               What underlying data the annotation is on:
+               "sources", or a "spectrum" (this option is not yet supported).
```
stefanv, line 342: Related to our last discussion: either implement spectrum annotations here, or move this docs change to another PR.

skyportal/handlers/api/annotation.py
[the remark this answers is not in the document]
profjsb: Ah, didn't know you are also ok with an empty string. Keep the `is not None` in that case. Was just trying to make a more elegant code.

profjsb (approved):

skyportal/handlers/api/annotation.py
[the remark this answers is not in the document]
guynir42: Actually I think empty strings should probably be rejected. I was wondering if there are other edge cases I haven't thought about?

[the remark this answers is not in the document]
stefanv: When is it possible for `a.obj.id` to be None?

skyportal/handlers/api/annotation.py
```diff
@@ -30,18 +49,64 @@ def get(self, annotation_id):
...
+            return self.error("Must provide a valid annotation ID. ")
+        # the default is to annotate an object
+        if associated_resource_type.lower() in (
+            "object",
+            "objects",
+            "source",
+            "sources",
+        ):
+            annotation = Annotation.get_if_accessible_by(
+                annotation_id, self.current_user, raise_if_none=True
```
stefanv: None of the `raise_if_none` exceptions seem to be handled.

skyportal/handlers/api/annotation.py
```diff
@@ -8,13 +8,32 @@
...
+          - in: path
+            name: resource_id
+            required: true
+            schema:
+              type: string
+            description: |
+               The ID of the underlying data.
+               This would be a string for an object ID
+               or an integer for other data types like spectrum.
+               The annotation ID must correspond to an annotation on the
```
stefanv, line 36: Can we clarify this?  I think users may have a hard time parsing what this means.

skyportal/handlers/api/annotation.py
stefanv: Why are we handling multiple cases here?

skyportal/handlers/api/annotation.py
stefanv: Why is this still in here?

skyportal/handlers/api/annotation.py
stefanv: This does not match the implementation below.

skyportal/handlers/api/annotation.py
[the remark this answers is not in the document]
guynir42: Yeah it is actually not nullable...

skyportal/handlers/api/annotation.py
guynir42: I think you are reviewing a previous commit. Let me tag you again after I push a couple more changes.

guynir42:
Hey @stefanv I think some of the stuff you were commenting on last time was from a previous commit. 
In any case, I pushed a few more changes and I hope that catches all the remaining issues. 
Let me know what you think. Thanks!

stefanv:
Sorry about that, not sure how I reviewed an old version! Thanks for the update, I will take another look.

skyportal/handlers/api/annotation.py
```diff
@@ -30,26 +51,65 @@ def get(self, annotation_id):
...
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid (scalar integer) annotation ID. ")
+
+        if associated_resource_type.lower() == "sources":
+            try:
+                annotation = Annotation.get_if_accessible_by(
+                    annotation_id, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible annotations.')
```
stefanv, line 66: Set appropriate status 404

@acrellin Would it make sense to catch AccessError in the errorhandler and do some automated response?

skyportal/handlers/api/annotation.py
```diff
@@ -30,26 +51,65 @@ def get(self, annotation_id):
...
+            annotation_resource_id_str = str(annotation.obj_id)
+        # add more options using elif
+        else:
+            return self.error(
+                f'Unsupported input "{associated_resource_type}" given as "associated_resource_type" argument.'
+            )
+
+        if annotation_resource_id_str != resource_id:
```
stefanv, line 74: Under which circumstances can this occur?

skyportal/handlers/api/annotation.py
```diff
@@ -30,26 +51,65 @@ def get(self, annotation_id):
...
+          - in: path
+            name: resource_id
+            required: true
+            schema:
+              type: string
+            description: |
+               The ID of the underlying data.
+               This would be a string for an object ID.
+               The object pointed to by this input must be a valid
+               object that can be annotated on by the user/token.
```
stefanv, line 106:
```suggestion
```

skyportal/handlers/api/annotation.py
```diff
@@ -98,17 +158,18 @@ def post(self):
...
-            groups = Group.get_if_accessible_by(
-                group_ids, self.current_user, raise_if_none=True
-            )
-
+            try:
+                groups = Group.get_if_accessible_by(
+                    group_ids, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible groups.')
```
stefanv, line 166: Appropriate HTTP status

skyportal/handlers/api/annotation.py
```diff
@@ -122,30 +183,62 @@ def post(self):
...
+            annotation = Annotation(
+                data=annotation_data,
+                obj_id=obj_id,
+                origin=origin,
+                author=author,
+                groups=groups,
+            )
+            annotation_resource_id_str = str(annotation.obj_id)
+        else:
+            return self.error(f'Unknown resource type "{associated_resource_type}".')
```
stefanv, line 197: Appropriate HTTP status

skyportal/handlers/api/annotation.py
```diff
@@ -176,40 +269,88 @@ def put(self, annotation_id):
...
+            required: true
+            schema:
+              type: string
+            description: |
+               The ID of the underlying data.
+               This would be a string for a source ID
+               or an integer for other data types like spectrum.
+               The resource pointed to by the resource_id must
+               be the correct object that is annotated by this
+               annotation.
```
stefanv, line 353:
```suggestion
```

skyportal/handlers/api/comment.py
```diff
@@ -28,26 +30,40 @@ def users_mentioned(text):
...
+            required: true
             schema:
               type: string
             description: |
-               What underlying data the comment is on:
-               an "object" (default), or a "spectrum".
+               The ID of the source or spectrum
+               that the comment is posted to.
+               This would be a string for a source ID
+               or an integer for other data types like spectrum.
```
stefanv, line 57:
```suggestion
               or an integer a spectrum.
```

skyportal/handlers/api/comment.py
```diff
@@ -28,26 +30,40 @@ def users_mentioned(text):
...
+               The resource pointed to by the resource_id must
+               be the correct object that is commented on by this
+               specific comment.
```
stefanv, line 60:
```suggestion
```

skyportal/handlers/api/comment.py
```diff
@@ -59,46 +75,77 @@ def get(self, comment_id, associated_resource_type=None):
...
+            return self.error("Must provide a valid (scalar integer) comment ID. ")
+
+        # the default is to comment on an object
+        if associated_resource_type.lower() == "sources":
+            try:
+                comment = Comment.get_if_accessible_by(
+                    comment_id, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible comments.')
```
stefanv, line 90: Appropriate HTTP status

skyportal/handlers/api/comment.py
```diff
@@ -59,46 +75,77 @@ def get(self, comment_id, associated_resource_type=None):
...
         elif associated_resource_type.lower() == "spectrum":
-            comment = CommentOnSpectrum.get_if_accessible_by(
-                comment_id, self.current_user, raise_if_none=True
-            )
+            try:
+                comment = CommentOnSpectrum.get_if_accessible_by(
+                    comment_id, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible comments.')
```
stefanv, line 99: HTTP status

skyportal/handlers/api/comment.py
```diff
@@ -59,46 +75,77 @@ def get(self, comment_id, associated_resource_type=None):
...
           - comments
+        parameters:
+          - in: path
+            name: associated_resource_type
+            required: true
+            schema:
+              type: string
+            description: |
+               What underlying data the comment is on:
+               "source", or a "spectrum".
```
stefanv, line 129:
```suggestion
               "source" or "spectrum".
```

skyportal/handlers/api/comment.py
```diff
@@ -59,46 +75,77 @@ def get(self, comment_id, associated_resource_type=None):
...
         ---
         description: Post a comment
         tags:
           - comments
+        parameters:
+          - in: path
+            name: associated_resource_type
+            required: true
+            schema:
+              type: string
```
stefanv, line 126: This should be an enum (same elsewhere)

skyportal/handlers/api/comment.py
```diff
@@ -59,46 +75,77 @@ def get(self, comment_id, associated_resource_type=None):
...
+          - in: path
+            name: resource_id
+            required: true
+            schema:
+              type: string
+            description: |
+               The ID of the source or spectrum
+               that the comment is posted to.
+               This would be a string for a source ID
+               or an integer for other data types like spectrum.
```
stefanv, line 139:
```suggestion
               or an integer for a spectrum.
```

skyportal/handlers/api/comment.py
```diff
@@ -59,46 +75,77 @@ def get(self, comment_id, associated_resource_type=None):
...
+               The object pointed to by this input must be a valid
+               source or other data type (like spectrum) that
+               can be commented on by the user/token.
```
stefanv, line 142:
```suggestion
```

skyportal/handlers/api/comment.py
```diff
@@ -139,19 +185,19 @@ def post(self):
...
         else:
-            groups = Group.get_if_accessible_by(
-                group_ids, self.current_user, raise_if_none=True
-            )
+            try:
+                groups = Group.get_if_accessible_by(
+                    group_ids, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible groups.')
```
stefanv, line 200: HTTP status

skyportal/handlers/api/comment.py
```diff
@@ -270,19 +344,31 @@ def put(self, comment_id, associated_resource_type=None):
...
-        elif associated_resource_type.lower() == "spectrum":
+            try:
+                c = Comment.get_if_accessible_by(
+                    comment_id, self.current_user, mode="update", raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible comments.')
+            comment_resource_id_str = str(c.obj_id)
+
+        elif associated_resource_type.lower() in ("spectrum", "spectra"):
```
stefanv, line 362: Which will it be?  Choose one.

skyportal/handlers/api/comment.py
```diff
@@ -204,46 +259,65 @@ def post(self):
...
             description: |
                What underlying data the comment is on:
                an "object" (default), or a "spectrum".
+          - in: path
+            name: resource_id
+            required: true
+            schema:
+              type: string
+            description: |
+               The ID of the object, source or spectrum
```
stefanv, line 309: Is object supported here?

skyportal/handlers/api/comment.py
```diff
@@ -270,19 +344,31 @@ def put(self, comment_id, associated_resource_type=None):
...
             schema = CommentOnSpectrum.__schema__()
-            c = CommentOnSpectrum.get_if_accessible_by(
-                comment_id, self.current_user, mode="update", raise_if_none=True
-            )
+            try:
+                c = CommentOnSpectrum.get_if_accessible_by(
+                    comment_id, self.current_user, mode="update", raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible comments.')
```
stefanv, line 369: HTTP status

skyportal/handlers/api/comment.py
```diff
@@ -335,44 +429,69 @@ def put(self, comment_id, associated_resource_type=None):
...
+            required: true
             schema:
               type: string
             description: |
-               What underlying data the comment is on:
-               an "object" (default), or a "spectrum".
+               The ID of the source or spectrum
+               that the comment is posted to.
+               This would be a string for a source ID
+               or an integer for other data types like spectrum.
```
stefanv, line 456:
```suggestion
               or an integer for a spectrum.
```

skyportal/handlers/api/comment.py
```diff
@@ -335,44 +429,69 @@ def put(self, comment_id, associated_resource_type=None):
...
+               The resource pointed to by the resource_id must
+               be the correct object that is commented on by this
+               specific comment.
```
stefanv, line 459:
```suggestion
```

skyportal/handlers/api/comment.py
```diff
@@ -381,15 +500,23 @@ def delete(self, comment_id, associated_resource_type=None):
 
         obj_key = c.obj.internal_key
         obj_id = c.obj.id
+        # some other logic should come here if comment is not
+        # associated with an object
```
stefanv, line 504:
```suggestion
```

skyportal/handlers/api/comment.py
```diff
@@ -400,26 +527,39 @@ def delete(self, comment_id, associated_resource_type=None):
...
+            required: true
             schema:
               type: string
             description: |
-               What underlying data the comment is on:
-               an "object" (default), or a "spectrum".
+               The ID of the source or spectrum
+               that the comment is posted to.
+               This would be a string for a source ID
+               or an integer for other data types like spectrum.
```
stefanv, line 554:
```suggestion
               or an integer for a spectrum.
```

skyportal/handlers/api/comment.py
```diff
@@ -400,26 +527,39 @@ def delete(self, comment_id, associated_resource_type=None):
...
+               The resource pointed to by the resource_id must
+               be the correct object that is commented on by this
+               specific comment.
```
stefanv, line 557:
```suggestion
```

skyportal/handlers/api/comment.py
```diff
@@ -451,25 +591,42 @@ def get(self, comment_id, associated_resource_type=None):
...
 
-        if associated_resource_type is None:
-            associated_resource_type = 'object'
+        if associated_resource_type.lower() == "sources":
+            try:
+                comment = Comment.get_if_accessible_by(
+                    comment_id, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible comments.')
```
stefanv, line 608: HTTP status

skyportal/handlers/api/comment.py
```diff
@@ -451,25 +591,42 @@ def get(self, comment_id, associated_resource_type=None):
...
         elif associated_resource_type.lower() == "spectrum":
-            comment = CommentOnSpectrum.get_if_accessible_by(
-                comment_id, self.current_user, raise_if_none=True
-            )
+            try:
+                comment = CommentOnSpectrum.get_if_accessible_by(
+                    comment_id, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible comments.')
```
stefanv, line 617: HTTP status

skyportal/handlers/api/annotation.py
guynir42: This would be a 405?

skyportal/handlers/api/comment.py
guynir42: I'm not sure how to do this as an ENUM. Can you think of an example already using that so I can have a look?

skyportal/handlers/api/annotation.py
stefanv: 403, I think.

We may be able to handle this generically, as I asked @acrellin about, but I don't think it currently is.

I made this related PR: https://github.com/cesium-ml/baselayer/pull/253/files which is not quite right yet, but should cover these use cases eventually.

skyportal/handlers/api/comment.py
stefanv: Grep for `enum` in `api/source.py`

guynir42: OK, so it isn't enforced as a data type, just documented as a string that has a closed set of options.

skyportal/handlers/api/annotation.py
guynir42: So looking at this PR I think maybe a 401 is more appropriate?

stefanv: 401 is when the user is unauthenticated

skyportal/handlers/api/annotation.py
guynir42: It can happen if the user tries to get annotation X from object Y but annotation X is actually associated with object Z.

skyportal/handlers/api/annotation.py
guynir42: What's the HTTP status for a wrong resource type? I would think a 400 is good enough?

skyportal/handlers/api/annotation.py
stefanv: In that case, we should simply return a not found.  We cannot leak that the other object exists, or that there is a comment with that ID.

skyportal/handlers/api/annotation.py
stefanv: 400 here is ok

skyportal/handlers/api/annotation.py
guynir42: I'm not sure if this is actually leaking anything. 
If the annotation ID is valid and accessible to the user, then the user should know the annotation exists. 
If it doesn't exist or is inaccessible, the handler returns `Could not find any accessible annotations.` a few lines before this. 

If the annotation is accessible but the object ID (or spectrum ID or whatever) doesn't match, it doesn't tell you that such an object exists, only that it does not match the ID of the object for that annotation (which the user is allowed to view). 
In that case the user should be made aware that he has a problem with the path parameter.

stefanv: OK, I'll resolve

stefanv:
Thanks for the updates, Guy.  @acrellin can you take a look at the `CommentListMobile.jsx` changes?

stefanv:
@guynir42 Thanks, I'm done reviewing.  Waiting on @acrellin to take a look at your frontend changes.

acrellin:
> Thanks for the updates, Guy. @acrellin can you take a look at the `CommentListMobile.jsx` changes?

@stefanv Sure, will do.

static/js/components/CommentListMobile.jsx
```diff
@@ -246,144 +314,175 @@ const CommentListMobile = ({ isCandidate }) => {
...
-                              : { display: "none" }
-                          }
-                          size="small"
-                          color="primary"
-                          name={`deleteCommentButton${id}`}
-                          onClick={() => {
-                            dispatch(sourceActions.deleteComment(id));
-                          }}
-                          className="commentDelete"
+          .map(
```
acrellin, line 317: What changed exactly in this portion of the diff?

static/js/ducks/source.js
```diff
@@ -77,25 +77,47 @@ export function addComment(formData) {
...
+  );
 }
 
-export function deleteComment(comment_id, associatedResourceType = "object") {
-  return API.DELETE(
-    `/api/comment/${comment_id}/${associatedResourceType}`,
-    DELETE_COMMENT
-  );
+export function deleteComment(pathToComment) {
+  return API.DELETE(pathToComment, DELETE_COMMENT);
```
acrellin, line 116: I think we should break this up into two separate action creators for the different resource types; that makes more sense than for the full request path to be passed in as a parameter as is done here. That path should ideally just be hard-coded into each action creator body.

acrellin (changes requested):

static/js/ducks/source.js
stefanv: This is mainly for internal use, and would improve readability, so seems okay to have one action per type.

acrellin: Exactly. Please update.

static/js/components/CommentListMobile.jsx
guynir42: The major addition is the spectrum_id and api_path (the latter will be removed I think). Then I use them when rendering the text, the attachment, and the delete button.

acrellin (approved):
I only looked at the front-end changes as requested, which look good to me! Thanks @guynir42 !

doc/comment_url_migration.md
```diff
@@ -0,0 +1,29 @@
...
+on different underlying data (e.g., spectra),
+and makes it harder to accidentally get the wrong comment,
+as the underlying object ID (source or spectrum)
+must match the object that the comment/annotation
+is associated with.
+
+In the past, a comment was accessed via the endpoint
+`api/comment/<commentID>`
+
+In the new framework the comment must be accessed via
```
acrellin, line 15:
```suggestion
In the new framework, a comment on a source must be accessed via
```

doc/comment_url_migration.md
```diff
@@ -0,0 +1,29 @@
...
+`api/sources/<sourceID>/comment/<commentID>`
+
+This also applies to comments on spectra:
+`api/spectrum/<spectrumID>/comment/<commentID>`
+
+For annotations, use:
+`api/sources/<sourceID>/annotation/<annotationID>`
+
+In all cases, when posting a new comment/annotation,
+do not supply a commentID or annotationID at the end of the path.
```
acrellin, line 26:
```suggestion
do not supply a commentID or annotationID at the end of the path, e.g.
POST to `api/sources/<sourceID>/comment`
```

acrellin (changes requested):

stefanv:
Ari, go ahead and merge once you are satisfied with the changes.

acrellin (approved):

PR #2215, opened 2021-08-24
Upgrade healpix-alchemy

mcoughlin:
Separate PR to update healpix-alchemy and deal with the DB migration script.


alembic/versions/4047c470ab83_healpix_alchemy_migration.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('objs', sa.Column('x', sa.Float(), nullable=True))
+    op.add_column('objs', sa.Column('y', sa.Float(), nullable=True))
+    op.add_column('objs', sa.Column('z', sa.Float(), nullable=True))
+    op.add_column('objs', sa.Column('nested', sa.BigInteger(), nullable=True))
+    op.create_index(op.f('ix_objs_nested'), 'objs', ['nested'], unique=False)
+    op.create_index(op.f('ix_objs_point'), 'objs', ['x', 'y', 'z'], unique=False)
```
acrellin, line 26:
```suggestion
```

alembic/versions/4047c470ab83_healpix_alchemy_migration.py
```diff
@@ -0,0 +1,54 @@
...
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(op.f('ix_photometry_point'), table_name='photometry')
+    op.drop_index(op.f('ix_photometry_nested'), table_name='photometry')
+    op.drop_column('photometry', 'nested')
+    op.drop_column('photometry', 'z')
+    op.drop_column('photometry', 'y')
+    op.drop_column('photometry', 'x')
+    op.drop_index(op.f('ix_objs_point'), table_name='objs')
```
acrellin, line 48:
```suggestion
```

alembic/versions/4047c470ab83_healpix_alchemy_migration.py
```diff
@@ -0,0 +1,52 @@
...
+    op.add_column('photometry', sa.Column('x', sa.Float(), nullable=True))
+    op.add_column('photometry', sa.Column('y', sa.Float(), nullable=True))
+    op.add_column('photometry', sa.Column('z', sa.Float(), nullable=True))
+    op.add_column('photometry', sa.Column('nested', sa.BigInteger(), nullable=True))
+    op.create_index(
+        op.f('ix_photometry_nested'), 'photometry', ['nested'], unique=False
+    )
+    op.create_index(
+        op.f('ix_photometry_point'), 'photometry', ['x', 'y', 'z'], unique=False
+    )
```
acrellin, line 35:
```suggestion
```

alembic/versions/4047c470ab83_healpix_alchemy_migration.py
```diff
@@ -0,0 +1,52 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(op.f('ix_photometry_point'), table_name='photometry')
```
acrellin, line 41:
```suggestion
```

acrellin:
@stefanv Thoughts on merging this even though the migration tester complains about missing indexes (they are indeed created; when we omit the "create only if not exists" part, it complains that they already exist 😮‍💨 )

stefanv:
How about trying the SSH action? It's worth investigating. We don't want red CI for the foreseeable future!

acrellin:
@stefanv I mean pin the resulting commit as the new reference commit in the test workflow

lpsinger:
This is good. Now the migration needs to populate the x, y, and z columns.

acrellin:
> This is good. Now the migration needs to populate the x, y, and z columns.

We discussed that we won't do that in the migration script, didn't we?

lpsinger:
Did we? How do we do it, then?

I don't think that a background job is the right place for this, because ordinarily when you create rows using the ORM layer, the x y z columns _are_ populated. It's only when you bypass the ORM and use the core SQL layer directly that they fail to be populated.

stefanv:
I tend to agree that it is better to swallow the pill during migration, if that does not take prohibitively long.

mcoughlin:
@lpsinger This is the one that failed: https://github.com/skyportal/skyportal/blob/master/skyportal/tests/tools/test_radial_query.py#L11

I tried setting the radius to 0

    # issue the query on the db
    db_ids = (
        sp_models.DBSession()
        .query(sp_models.Obj.id)
        .filter(
            sp_models.Obj.within(
                ha.Point(ra=coord[0].ra.deg, dec=coord[0].dec.deg), 0.0
            )
        )
    )

and do still get many overlaps... Maybe this isn't the best way to test it, but something does seem wonky perhaps.

lpsinger:
This test is strange. Why do you expect there to be _any_ results to this query, unless some of the rows in `Obj` are at precisely ra=0 dec=0?

lpsinger:
Hmm, that's very odd. Can you make a minimal example to reproduce with just healpix-alchemy?

lpsinger:
> Beware that you've populated the Cartesian columns for the `Objs` table but not the `Photometry` table.

Ignore that. My mistake.

lpsinger:
Was this test passing before your changes to the migration script? I don't see how the migration script would have any impact on that test.

PR #2241, opened 2021-09-05
Disable slow Obj filters in Sources tables

dmitryduev:
This is a temporary measure to prevent potential system slowdowns. See https://github.com/fritz-marshal/fritz/issues/338 for the description and background.
I only disabled things on the frontend, didn't touch the backend. In the latter case, there are multiple places where permissions checking potentially cause tons of lookups that are slow (last_detected_at, peak_magnitude, etc.), but let's hope that our theory is right and people were only using the front-end to issue these queries.

In the meantime, we should come up with a permanent solution.


guynir42 (approved):
It looks fine to me. We should think about long term solutions.

stefanv (approved):

Issue #2218, opened 2021-08-26
Update state upon successful spectrum upload


dmitryduev:
We should be updating the app state upon a successful spectrum upload for a given object.

To reproduce the behavior that is confusing to the users:
- Go to any source page
- Upload additional spectroscopy -> upload a spectrum
- Once the app declares success, click on the object name at the top of the page to go back to the Source page
- The newly uploaded spectrum will not be displayed under Spectroscopy *unless the page is refreshed*

guynir42:
will be fixed by #2248

PR #2248, opened 2021-09-08
Upload/delete spectrum updates bokeh plot

guynir42:


guynir42:
Addresses #2218

skyportal/plot.py
```diff
@@ -1351,8 +1351,12 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
         .all()
     )
 
-    if spec_id is not None:
-        spectra = [spec for spec in spectra if spec.id == int(spec_id)]
+    if spec_id is not None and len(spec_id) > 0:
+        spec_id = spec_id.split(',')
+        filtered_spectra = []
+        for id in spec_id:
+            filtered_spectra.extend([spec for spec in spectra if spec.id == int(id)])
```
dmitryduev, line 1358: `id` shadows a built-in python literal
```suggestion
        for sid in spec_id:
            filtered_spectra.extend([spec for spec in spectra if spec.id == int(sid)])
```

dmitryduev (commented):
👍🏻

static/js/components/SourceDesktop.jsx
```diff
@@ -407,7 +411,7 @@ const SourceDesktop = ({ source }) => {
               <div className={classes.photometryContainer}>
                 <Suspense fallback={<div>Loading spectroscopy plot...</div>}>
                   <Plot
-                    url={`/api/internal/plot/spectroscopy/${source.id}?width=800&height=600`}
+                    url={`/api/internal/plot/spectroscopy/${source.id}?width=800&height=600&spectrumID=${specIDs}`}
```
stefanv, line 414: I think this approach can be simplified to the following:

```
<Plot ... cacheIDs=${specIDs}>
```

All we need is for a bunch of string identifiers to be passed to the plot that will change whenever the spectra change.  This way, there is also no need to parse the spectra IDs out---we can use their labels as-is.

dmitryduev:
@stefanv can we merge this now?

skyportal/plot.py
```diff
@@ -1351,8 +1351,12 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
         .all()
     )
 
-    if spec_id is not None:
-        spectra = [spec for spec in spectra if spec.id == int(spec_id)]
+    if spec_id is not None and len(spec_id) > 0:
+        spec_id = spec_id.split(',')
```
stefanv, line 1355: Do you mind explaining what is happening here and why?

guynir42: In case the user wants to plot multiple spectra, there is an option to give a comma separated list of values. I will add a code comment. 

I added this when we added the string with the spectra names, in order to make the plot re-render on a spectrum add/delete. Then we turned the input parameter into a dummy input but I think this functionality could be useful in general so I left it in.

stefanv: I don't see an accompanying documentation change, though, so where is this feature advertised?

guynir42: Should I add a doc string for this function (which, incidentally, doesn't have one yet)?

stefanv: Yes, please!  These functions are ridiculously big, so we should look into refactoring them at some point.

skyportal/plot.py
```diff
@@ -1344,6 +1344,40 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
 
 def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
+    """
+    Create object spectroscopy line plot.
+
+    Parameters
+    ----------
+    obj_id : str
+        ID of Obj to be plotted.
+    user:
```
stefanv, line 1354:
```suggestion
    user :
```

skyportal/plot.py
```diff
@@ -1344,6 +1344,40 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
+        The user object that is requesting the plot.
+    spec_id: str or None
```
stefanv, line 1356:
```suggestion
    spec_id : str or None
```

skyportal/plot.py
```diff
@@ -1344,6 +1344,40 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
+        A string with a single spectrum ID or a
+        comma-separated list of IDs.
+        Only the spectra with matching IDs
+        are plotted.
+        If None (default), will plot all
+        spectra associated with the object
+        and accessible to the user.
+    width: int
```
stefanv, line 1364:
```suggestion
    width : int
```

skyportal/plot.py
```diff
@@ -1344,6 +1344,40 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
+        Size of the plot in pixels. Default=600.
+    device: str
```
stefanv, line 1366:
```suggestion
    device : str
```

skyportal/plot.py
```diff
@@ -1344,6 +1344,40 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
+        the device on which the plot will be displayed:
+        - "browser" (default)
+        - "mobile_portrait"
+        - "mobile_landscape"
+        - "tablet_portrait"
+        - "tablet_landscape"
+    Returns
+    -------
+    dict
+        Returns Bokeh JSON embedding for the desired plot.
```
stefanv, line 1377:
```suggestion
        Bokeh JSON embedding of the plot.
```

stefanv (commented):
Thanks for the docstring!

guynir42:
Can we merge this? @dmitryduev @stefanv

stefanv (approved):

stefanv:
@guynir42 Will one of our demo data sources exercise this functionality?  Otherwise it is completely untested.

guynir42:
> @guynir42 Will one of our demo data sources exercise this functionality? Otherwise it is completely untested.

Do you mean the functionality of plotting only some spectra out of all the object's spectra?

stefanv:
Yes, selecting one or more spectra to plot.

guynir42:
> Yes, selecting one or more spectra to plot.

I would prefer to add a test for that in a separate PR.

stefanv:
> > Yes, selecting one or more spectra to plot.
> 
> I would prefer to add a test for that in a separate PR.

I know it's work to add tests, but we have [clear guidelines](https://skyportal.io/docs/contributing.html#testing) on that in this project.  I'm not doubting your good intensions, but we have learned the hard way that tests that don't come with a change seldom arrive.

guynir42:
> > > Yes, selecting one or more spectra to plot.
> > 
> > 
> > I would prefer to add a test for that in a separate PR.
> 
> I know it's work to add tests, but we have [clear guidelines](https://skyportal.io/docs/contributing.html#testing) on that in this project. I'm not doubting your good intensions, but we have learned the hard way that tests that don't come with a change seldom arrive.

Ok will add a test and re-tag you for review.

stefanv:
Easiest would be to add an API test.  You can do that, even though we typically only use this endpoint in pages.

stefanv:
I.e., call the plot endpoint, then just verify that it has the expected number/types of entries in the resulting Bokeh JSON.

skyportal/plot.py
```diff
@@ -1344,15 +1344,57 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
+        - "mobile_landscape"
+        - "tablet_portrait"
+        - "tablet_landscape"
+    Returns
+    -------
+    dict
+        Bokeh JSON embedding of the plot.
+
+    """
+
```
dmitryduev, line 1380: Thanks for adding this.

dmitryduev (approved):
LGTM! Good to go in once the last point from @stefanv is taken care of.

Could you please propagate the changes to SourceDesktop.jsx and SourceMobile.jsx to Fritz (https://github.com/fritz-marshal/fritz/blob/master/extensions/skyportal/static/js/components/SourceDesktop.jsx and https://github.com/fritz-marshal/fritz/blob/master/extensions/skyportal/static/js/components/SourceMobile.jsx)? Would appreciate a sister PR on F.

Thanks!

skyportal/tests/api/test_spectrum.py
```diff
@@ -742,3 +742,77 @@ def test_spectrum_external_reducer_and_observer(
     assert data['data']['observers'][0]['id'] == user.id
     assert data['data']['external_reducer'] == "Test external reducer"
     assert data['data']['external_observer'] == "Test external observer"
+
+
+def test_spectrum_plot_only_some_spectra(
+    upload_data_token, public_source, public_group, lris, sedm
+):
+    # upload spectra 1
```
stefanv, line 750: Spectrum, since spectra is plural

skyportal/tests/api/test_spectrum.py
```diff
@@ -742,3 +742,77 @@ def test_spectrum_external_reducer_and_observer(
...
+        if 'data' in r['attributes'] and 'instrument' in r['attributes']['data']:
+            instruments.add(r['attributes']['data']['instrument'][0])
+
+    assert len(instruments) > 1
+
+    # now ask for the same plot but with only one spectrum
+    status, data = api(
+        'GET',
+        f'internal/plot/spectroscopy/{public_source.id}',
+        params={'spectrumID': f'{spec_id}'},
```
stefanv, line 805: I'd change this call to have multiple IDs, so as to test the spliting functionality

stefanv (commented):
Two minor tweaks, then I think we can go ahead and merge.

guynir42:
> LGTM! Good to go in once the last point from @stefanv is taken care of.
> 
> Could you please propagate the changes to SourceDesktop.jsx and SourceMobile.jsx to Fritz (https://github.com/fritz-marshal/fritz/blob/master/extensions/skyportal/static/js/components/SourceDesktop.jsx and https://github.com/fritz-marshal/fritz/blob/master/extensions/skyportal/static/js/components/SourceMobile.jsx)? Would appreciate a sister PR on F.
> 
> Thanks!

Hi @dmitryduev I actually don't have a version of Fritz spinned up yet... I think it would be a lot faster if you copied the changes into Fritz...

guynir42:
Test failure is unrelated. @stefanv feel free to merge if it looks good to you...

stefanv:
Happy to merge this; just want to double check why `len(instruments) > 2` when only two spectra are uploaded?  Multiple measurements by each instrument?  I.e. `len(set(instruments)) == 2`?

guynir42:
> Happy to merge this; just want to double check why `len(instruments) > 2` when only two spectra are uploaded? Multiple measurements by each instrument? I.e. `len(set(instruments)) == 2`?

I think the fixture comes with one spectrum.

PR #2251, opened 2021-09-11
Upgrade requests and patch its `requests.compat.json` module to handle nan/inf

stefanv:


stefanv:
@profjsb I thought it would be possible to replace nan's with null, but I forgot that we also occasionally deal with infinity.  So, I followed your suggestion of monkey-patchin and tried to keep it very targeted (`simplejson.dumps`).  This should not cause problems unless we try to use `dumps(x, allow_nan=False)`, which will no longer work anywhere.  But I see no reason we would do that (famous last words, perhaps).

profjsb:
Looks like CI test failures are pointing to a misplaced file? (`from .patch_requests import patch_requests`)

stefanv:
> Looks like CI test failures are pointing to a misplaced file? (`from .patch_requests import patch_requests`)

Thanks for noticing; I updated the commit.

stefanv:
We use the data loader from master to provision the migrations tests.  So, I think this looks good!

dmitryduev (approved):
Looks great!

PR #2253, opened 2021-09-11
Provide explicit feedback when API fetch fails

stefanv:
Before, when an API call failed for an unknown reason, we displayed
the error but did not try and propagate it further.  Now, the error
message is attached to the `*_FAIL` action so that it may be displayed
in a component or logged to the console.

### DIFF
--- a/static/js/ducks/candidate.js
+++ b/static/js/ducks/candidate.js
@@ -29,11 +29,10 @@ const reducer = (state = initialState, action) => {
         ...state,
         loadError: action.message,
       };
-
     case FETCH_CANDIDATE_FAIL:
       return {
         ...state,
-        loadError: "Unknown error while loading candidate",
+        loadError: `Error while loading candidate: ${action.message}`,
       };
     default:
       return state;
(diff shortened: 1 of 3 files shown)


stefanv:
I added this feature because users were seeing `Unknown error while fetching X`.  Which they never should!  But now we can display those errors to them, and hopefully see where they're coming from.

static/js/ducks/candidate.js
```diff
@@ -29,11 +29,10 @@ const reducer = (state = initialState, action) => {
         ...state,
         loadError: action.message,
       };
-
     case FETCH_CANDIDATE_FAIL:
       return {
         ...state,
-        loadError: "Unknown error while loading candidate",
+        loadError: `Unknown error while loading candidate: ${action.message}`,
```
acrellin, line 35:
```suggestion
        loadError: `Error while loading candidate: ${action.message}`,
```

acrellin (approved):
Looks good, thanks @stefanv

acrellin (approved):

Issue #2254, opened 2021-09-13
SDSS and DR8 thumbnails don't always get created


dmitryduev:
This happens when running `make load_demo_data`: since that now happens asynchronously and we kill the process once the serial part is complete, probably we just don't wait long enough for them to be fetched and saved.

Also, as I've noticed, when the app is under heavy load, this could also happen in production.

stefanv:
We should look into implementing something like [this](https://gist.github.com/wonderbeyond/d38cd85243befe863cdde54b84505784) from [this thread](https://github.com/tornadoweb/tornado/issues/1791#issuecomment-596265013) thread.

dmitryduev:
Addressed in #2255, closing.

Issue #1361, opened 2020-11-13
Thumbnail logic


dmitryduev:
It looks like we may have oversimplified the logic regarding the thumbnails when we decoupled them from photometry, For example: https://fritz.science/api/sources/ZTF18abxhsyn I can see, e.g., three "new" thumbnails all pointing to the same static file.

- For inspection purposes, one can look up all thumbnails from the Alert page.
- Do we really need multiple "ZTF" cutouts?
- If yes, then we also need to add the ability to see all posted thumbnails, and not only the most recent ones
- Also, currently there's no way to check if thumbnails exist if your token has no access to the filter it's already saved to

jvansanten:
It may make sense to distinguish between reference images (sdss, ps1, dr8, perhaps the ztf reference itself), which are associated with an object x survey, and subtraction images, which are associated with a particular photometric point. These are conceptually different, and it doesn't really make sense to display a single subtraction image for a transient.

steveschulze:
While you are working on that, could you add an option to Fritz to inspect cut-outs (image, ref, subtraction) of each alert of a given object? This is important and extremely valuable for fast and infant-SN programs where the follow-up strategy is based on 1-2 detections.

dmitryduev:
@steveschulze this is possible. For example:
https://fritz.science/alerts/ztf/ZTF18aaiwsqa
(click "Show Thumbnails" in the table; haven't implemented clicking on the lc yet)
API endpoint: https://fritz.science/api/alerts/ztf/ZTF18aaiwsqa/cutout?candid=1392098985715015002&cutout=template&file_format=png

steveschulze:
Thanks Dima. :)

acrellin:
Closing in favor of https://github.com/skyportal/skyportal/issues/1970

PR #2255, opened 2021-09-13
Grab SDSS and DR8 thumbnails if they don't exists for some reason

dmitryduev:
Addresses https://github.com/skyportal/skyportal/issues/2254.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -693,10 +693,21 @@ class SourceHandler(BaseHandler):
                 self.verify_and_commit()
 
             if include_thumbnails:
-                if "ps1" not in [thumb.type for thumb in s.thumbnails]:
+                existing_thumbnail_types = [thumb.type for thumb in s.thumbnails]
+                if "ps1" not in existing_thumbnail_types:
                     IOLoop.current().run_in_executor(
                         None,
                         lambda: add_ps1_thumbnail_and_push_ws_msg(
+                            obj_id, self.current_user
+                        ),
+                    )
+                if (
+                    "sdss" not in existing_thumbnail_types
+                    or "dr8" not in existing_thumbnail_types
+                ):
+                    IOLoop.current().run_in_executor(
+                        None,
+                        lambda: add_linked_thumbnails_and_push_ws_msg(
                             obj_id, self.current_user
                         ),
                     )


acrellin (commented):
Looks good but we should take care that simultaneous fetches by different users don't trigger duplicate thumbnail generation processes.

dmitryduev:
Hey @acrellin, good to e-see you :)
How should we do that? I don't see any specific logic for PS1 thumbnails, and this basically does the same thing. Also, some of the calls would just fail - the db would complain and won't let the transaction complete, no?

acrellin:
> Hey @acrellin, good to e-see you :)
> How should we do that? I don't see any specific logic for PS1 thumbnails, and this basically does the same thing. Also, some of the calls would just fail - the db would complain and won't let the transaction complete, no?

Yeah, just thinking out loud -- whatever failsafe we might add should be applied to PS1 generation, too. Maybe just a unique constraint that would cause a SQLA error upon attempting to insert a duplicate?

dmitryduev:
@acrellin would that always happen [here](https://github.com/skyportal/skyportal/blob/master/skyportal/models.py#L1042)?

dmitryduev:
Failure in `test_token_user_big_post` - unrelated.

acrellin:
@dmitryduev I don't think so -- we don't have any unique constraints on thumbnails, but adding this should do the trick: 
``` python
Thumbnail.__table_args__ = (
    sa.Index(
        "thumbnail_deduplication_index",
        Thumbnail.obj_id,
        Thumbnail.type,
        unique=True,
    ),
)
```

dmitryduev:
@acrellin ah, you're right. We should be careful with this though: remember https://github.com/skyportal/skyportal/issues/1361. + this will effectively disable multi-epoch thumbnail for now. 
I would advocate for merging this as is, as there have been complains about the missing thumbnails, but we should really get back to 1361...
@stefanv what do you think?

stefanv:
As long as duplicates don't cause issues (which I don't think they will), how about we add this for now and file an issue to discuss duplicate handling?

dmitryduev:
👍🏻 
@acrellin?

acrellin:
Sounds good!

acrellin (approved):

PR #2261, opened 2021-09-13
Set reference commit for migration checks

stefanv:


dmitryduev (approved):

acrellin (approved):

stefanv:
Woohoo, thanks @acrellin!

acrellin:
@stefanv What does this PR need now to be ready?

stefanv:
I think it's good to go.

.github/workflows/test_migrations.yaml
```diff
@@ -3,16 +3,21 @@ name: Test Skyportal Migrations
 on:
   # Run only if potential changes in database schema
   pull_request:
-    paths:
```
stefanv, line 6: @acrellin One question, I guess, is whether this should go back in?

acrellin: Ah, yeah probably good to restrict the runs to changes to these. Shall I add that back in a new commit?

stefanv: Thanks, that'd be great; perhaps also set the reference commit to the point where the new data loader reporting got merged in, then we'll have the advantage of that when running migration tests.

PR #2267, opened 2021-09-14
Make data loader reporting more transparent

stefanv:
Let the user know what HTTP codes were expected/received.

Try 15 times before giving up, so that data loader has a better chance
on slow CI.

### DIFF
--- a/tools/data_loader.py
+++ b/tools/data_loader.py
@@ -94,21 +94,21 @@ if __name__ == "__main__":
 
         try:
             token = yaml.load(open('.tokens.yaml'), Loader=yaml.Loader)['INITIAL_ADMIN']
-            print('Token loaded from `.tokens.yaml`')
             return token
         except (FileNotFoundError, TypeError, KeyError):
-            print(
-                'Error: no token specified, and no suitable token found in .tokens.yaml'
-            )
             return None
 
     print('Testing connection...', end='')
 
-    RETRIES = 10
+    RETRIES = 15
     timeout = 3
+    admin_token = None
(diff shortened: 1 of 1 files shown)


stefanv:
[Example of the output](https://github.com/skyportal/skyportal/pull/2267/checks?check_run_id=3595471599#step:13:110)

Failures are unrelated.

tools/data_loader.py
```diff
@@ -94,21 +94,21 @@ def get_token():
...
     print('Testing connection...', end='')
 
-    RETRIES = 10
+    RETRIES = 15
     timeout = 3
+    admin_token = None
     for i in range(RETRIES):
         try:
+            previous_admin_token = admin_token
             admin_token = get_token()
```
acrellin, line 109:
```suggestion
            if previous_admin_token = None:
                admin_token = get_token()
```
Otherwise we're still fetching the token for each retry, right?

stefanv: We need to, because the token file changes on disk sometimes.  (I had it the way you suggested first, and that's how I found the 400 authorization errors!)

acrellin: Ah, okay, I was confused because it says this PR makes it so that the token is loaded only once. Perhaps I'm just misunderstanding that part, apologies!

stefanv: Durnit; Github grabbed that from my commit message, but of course it doesn't reflect that I updated the commit message!  Sorry, I will edit it.

acrellin (approved):

PR #2268, opened 2021-09-14
Do not include detection stats in sources action creators; remove Peak/Latest mag cols from SourceTable

acrellin:
This is a temporary stopgap to address major performance issues
when fetching sources including detection statistics. We will
revert this when those inefficiencies are addressed.

### DIFF
--- a/static/js/components/SourceTable.jsx
+++ b/static/js/components/SourceTable.jsx
@@ -27,7 +27,7 @@ import CircularProgress from "@material-ui/core/CircularProgress";
 import dayjs from "dayjs";
 import { isMobileOnly } from "react-device-detect";
 
-import { ra_to_hours, dec_to_dms, time_relative_to_local } from "../units";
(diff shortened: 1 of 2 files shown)


dmitryduev (approved):

dmitryduev:
@acrellin thanks! feel free to self-merge once the CI has finished.

stefanv:
Please remember to file appropriate issues so that we can follow up later.

Go ahead and merge after doing so; the failure is unrelated.

dmitryduev:
@acrellin would you mind taking care of filing the issues?

profjsb:
Related to #2298 ?

mcoughlin:
@guynir42 Maybe those columns should be replaced by your stats table relationship?

guynir42:
They don't need to be replaced, but querying will be re-wired to use the PhotStats when it is ready.

PR #2269, opened 2021-09-14
Upgrade bokeh/bokehjs to 2.4

stefanv:
BokehJS now uses more JS features, so added the babel plugins for:

- optional chaining (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
- nullish coalescing operator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator).

Tweaked the models to be in line with the latest BokehJS (mainly usage
of `super.render()`).

<strike>Depends on https://github.com/bokeh/bokeh/issues/11616 to be resolved.</strike>

### DIFF
--- a/static/js/components/BokehModels.js
+++ b/static/js/components/BokehModels.js
@@ -3,19 +3,18 @@
 import { input, label, div, span } from "bokehjs/core/dom";
 import { includes } from "bokehjs/core/util/array";
 import * as p from "bokehjs/core/properties";
-import { bk_inline } from "bokehjs/styles/mixins";
-import { bk_input_group } from "bokehjs/styles/widgets/inputs";
+import * as inputs from "bokehjs/styles/widgets/inputs.css";
+import { InputGroupView } from "bokehjs/models/widgets/input_group";
 import {
   CheckboxGroup,
   CheckboxGroupView,
 } from "bokehjs/models/widgets/checkbox_group";
-import { InputGroupView } from "bokehjs/models/widgets/input_group";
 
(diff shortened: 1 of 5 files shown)


stefanv:
Follow-up requests is failing with:

`E       selenium.common.exceptions.MoveTargetOutOfBoundsException: Message: (248, 1247) is out of bounds of viewport width (1920) and height (1115)`

Unsure how this is related to Bokeh, but seems to be happening consistently after restarting the tests.

static/js/components/BokehModels.js
```diff
@@ -3,19 +3,18 @@
 import { input, label, div, span } from "bokehjs/core/dom";
 import { includes } from "bokehjs/core/util/array";
 import * as p from "bokehjs/core/properties";
-import { bk_inline } from "bokehjs/styles/mixins";
-import { bk_input_group } from "bokehjs/styles/widgets/inputs";
+import * as inputs from "bokehjs/styles/widgets/inputs.css";
```
acrellin, line 6:
```suggestion
import * as bokehInputsCss from "bokehjs/styles/widgets/inputs.css";
```
?

stefanv: This file is a modified version of code in Bokeh (checkbox widget) so I am trying to keep them the same with minimal changes to make it easier to diff.

acrellin: Okay, no problem then.

acrellin (approved):

Issue #2270, opened 2021-09-15
Revert changes to source table columns when detection stats are optimized


acrellin:
https://github.com/skyportal/skyportal/pull/2268 is a temporary stopgap to improve performance until we can sort out the inefficiencies in the detection statistics computation, and should be reverted thereafter.

PR #2273, opened 2021-09-15
DNM: Revert "Do not include detection stats in sources action creators (#2268)"

acrellin:
This reverts commit 87cb4736455872e270bf953cf0c829e3a33657be.

Closes https://github.com/skyportal/skyportal/issues/2270


stefanv (approved):

mcoughlin:
@stefanv do you remember the history of this? https://github.com/skyportal/skyportal/blob/master/skyportal/handlers/api/source.py#L1346 
the detection stats computation is in the query so I would think this should be OK nowadays?

Issue #2214, opened 2021-08-19
PDF file preview and download in comments not working


dmitryduev:
It says "Failed to load PDF file."
The endpoint that it tries to hit looks like `https://fritz.science/api/comment/65728/attachment/object.pdf`, but that throws an error saying that it's an invalid API endpoint. Hitting `https://fritz.science/api/comment/65728/attachment/` grabs the correct file.

dmitryduev:
@scizen9 would you like to take this one on? I recall you/@stefanv dealt with a similar issue recently...

scizen9:
When I try to use the second URL above, it still gives an error.  If I replace 'object.pdf' with 'ztf21abubpng_full_fps_mags.pdf' it also gives an error.  Could this be the special pdf handling problem?  SEDM has been uploading loads of png files to comments and they display/download fine.

PR #2274, opened 2021-09-15
Fix faulty URL for PDF attachment previews

acrellin:
This updates the regex used for matching comment attachment preview
requests to the correct Tornado handler. This was likely broken when
the comment API was updated to include associated resource type.

Closes https://github.com/skyportal/skyportal/issues/2214


dmitryduev (approved):
👍🏻

dmitryduev:
Please feel free to self-merge once green.

PR #2281, opened 2021-09-22
Use optional chaining when filtering JS arrays

acrellin:
This patch adds `?` optional chaining when filtering JS arrays to
ensure that non-existent methods of null objects are not called.

Closes https://github.com/skyportal/skyportal/issues/2259


stefanv (approved):
Nice, how did you catch all of these?

dmitryduev (approved):
That is some formidable grepping effort @acrellin, well done!

PR #2283, opened 2021-09-22
Use optional chaining when attempting to map JS arrays

acrellin:
This patch employs optional chaining (`?.map`) when attempting to
map JavaScript arrays to avoid crashing in the event that the variable
in question is null or undefined.

### DIFF
--- a/static/js/components/AddUsersFromGroupForm.jsx
+++ b/static/js/components/AddUsersFromGroupForm.jsx
@@ -37,7 +37,7 @@ const AddUsersFromGroupForm = ({ groupID }) => {
   };
 
   const onSubmit = async (formData) => {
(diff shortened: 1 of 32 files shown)


stefanv:
It would be nice to only use these when necessary, but that would take forever to find.  So :+1:

dmitryduev (approved):
👍🏻 feel free to self-merge once green(ish)!

acrellin:
> It would be nice to only use these when necessary, but that would take forever to find. So +1

Yes, agreed. I did a quick skim of each use case and ended up skipping those where the variable in question would obviously never be `null`. Those that weren't immediately obvious I just went ahead and added the `?` for extra security, even though some of those are unnecessary.

acrellin:
6 green check marks ✅  -- merging

Issue #2292, opened 2021-09-28
`make load_demo_data` fails on fresh checkout


lpsinger:
Loading demo data fails on master...

```
$ make load_demo_data

Welcome to SkyPortal v0.9.dev0+git20210928.0e0d90d (https://skyportal.io)

Checking system dependencies:
[✓] nginx >= 1.7                  [  1.21.3]
[✓] psql >= 9.6                   [    13.4]
[✓] node (npm) >= 5.8.0           [  7.24.1]
[✓] python >= 3.8                 [  3.8.12]

[✓] Baselayer installed inside of app
--------------------
[✓] Verifying Python package dependencies
[·] Silently executing: baselayer/tools/chec[✓] Silently executing: baselayer/tools/check_js_deps.sh
[14:57:36 baselayer] Missing config files: config.yaml; continuing.
[14:57:36 baselayer] Warning: You are running on the default configuration. To configure your system, please copy `config.yaml.defaults` to `config.yaml` and modify it as you see fit.
  Config files: baselayer/config.yaml.defaults
                config.yaml.defaults
Testing connection...No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
No token specified; reading from SkyPortal generated .tokens.yaml
Traceback (most recent call last):
  File "tools/data_loader.py", line 160, in <module>
    if status not in (200, 400):
NameError: name 'status' is not defined
make: *** [load_demo_data] Error 1
```

acrellin:
Are you also running `make run` in another terminal?

lpsinger:
Yes.

[...]

PR #2296, opened 2021-09-28
Address edge case where status is undefined in data loader

acrellin:
This addresses the edge case where, after the maximum number of
attempts to connect to the running app instance, the `status` variable 
is referenced without being defined.

See https://github.com/skyportal/skyportal/issues/2292#issue-1010100911

### DIFF
--- a/tools/data_loader.py
+++ b/tools/data_loader.py
@@ -103,6 +103,7 @@ if __name__ == "__main__":
     RETRIES = 15
     timeout = 3
     admin_token = None
+    status = None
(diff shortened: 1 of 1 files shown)


acrellin:
Sole test failure unrelated (flaky front-end test)

profjsb (approved):

acrellin:
@stefanv Care to merge?

stefanv:
Ah, yeah, sorry that assignment should probably have been made on line 135, but this will work too.

PR #2299, opened 2021-09-29
sources GET with includeComments now gets spectrum comments as well

guynir42:


skyportal/handlers/api/source.py
```diff
@@ -732,6 +733,18 @@ class Validator(Schema):
...
+                        self.current_user,
+                        options=[
+                            joinedload(CommentOnSpectrum.author),
+                            joinedload(CommentOnSpectrum.groups),
+                        ],
+                    )
+                    .filter(CommentOnSpectrum.obj_id == obj_id)
+                    .all()
+                )
+                comments = comments + comments_on_spectra
```
dmitryduev, line 747: I'd do something like this (esp. in case we add more object-related commentable entities in the future):

```suggestion
                comments = []
                for comment_type in (Comment, CommentOnSpectrum):
                    comments.extend(
                        comment_type.query_records_accessible_by(
                            self.current_user,
                            options=[
                                joinedload(comment_type.author),
                                joinedload(comment_type.groups),
                            ],
                        )
                        .filter(comment_type.obj_id == obj_id)
                        .all()
                    )
```

dmitryduev: @stefanv would you agree?

dmitryduev (changes requested):
Looks good, but several tests need to be fixed (see both https://github.com/skyportal/skyportal/pull/2299/checks?check_run_id=3759170336 and https://github.com/skyportal/skyportal/pull/2299/checks?check_run_id=3759170249)

skyportal/handlers/api/source.py
```diff
@@ -721,17 +722,19 @@ class Validator(Schema):
...
-                            joinedload(Comment.groups),
-                        ],
+                comments = []
+                for comment_type in (Comment, CommentOnSpectrum):
+                    comments.extend(
+                        comment_type.query_records_accessible_by(
+                            self.current_user,
+                            options=[
+                                joinedload(Comment.author),
+                                joinedload(Comment.groups),
```
dmitryduev, line 732: Missed this:
```suggestion
                                joinedload(comment_type.author),
                                joinedload(comment_type.groups),
```

PR #2300, opened 2021-09-29
add spectrum type

guynir42:
Add a column named "type" to the Spectrum table. 
This will have one of a few allowed values (defined in the config). 
By default the types would be "source", "host" and "host center" but it can be anything if the config is changed. 

This can later be used to plot only spectra for the host, or other uses.


alembic/versions/b05b6c4d1a17_spectrum_type.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'b05b6c4d1a17'
+down_revision = '171e9ea18885'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('spectra', sa.Column('type', sa.String(), nullable=True))
```
acrellin, line 21: Do we want to back-populate existing rows with a sensible default?

skyportal/handlers/api/spectrum.py
```diff
@@ -149,7 +149,7 @@ def post(self):
 
         self.push_all(
             action='skyportal/REFRESH_SOURCE_SPECTRA',
-            payload={'obj_key': spec.obj.internal_key},
+            payload={'obj_id': spec.obj.id},
```
acrellin, line 152: ??

acrellin (commented):
What about the proposed enum list in the config?

alembic/versions/b05b6c4d1a17_spectrum_type.py
guynir42: Yes! but I am not sure how to do that...

skyportal/handlers/api/spectrum.py
guynir42: Just a small bug I found along the way.

alembic/versions/b05b6c4d1a17_spectrum_type.py
acrellin: Here's an example: https://github.com/skyportal/skyportal/blob/master/alembic/versions/5efdda259fd0_add_role_id_to_invitations_table.py#L24-L36

skyportal/handlers/api/spectrum.py
acrellin: So the front-end actually checks obj_id and not obj_key?

guynir42: The handler for REFRESH_SOURCE_SPECTRA does. Maybe that is also a bug?

skyportal/models.py
```diff
@@ -2517,6 +2517,9 @@ class Spectrum(Base):
         doc="Median UTC ISO time stamp of the exposure or exposures in which the Spectrum was acquired.",
     )
     origin = sa.Column(sa.String, nullable=True, doc="Origin of the spectrum.")
+    type = sa.Column(
+        sa.String, nullable=True, doc="Type of spectrum: source, host, etc. "
```
acrellin, line 2521: If we're going to enforce a default value at the handler level (that doesn't appear to be committed here yet), we should make this column non-nullable. We can also provide a default value here in the SQLA definition.

guynir42: Agreed.

guynir42:
Test failures seem to be unrelated.

alembic/versions/b05b6c4d1a17_spectrum_type.py
acrellin: Did you apply this migration locally to ensure that it worked to back-populate existing rows?

skyportal/models.py
```diff
@@ -2517,6 +2517,12 @@ class Spectrum(Base):
         doc="Median UTC ISO time stamp of the exposure or exposures in which the Spectrum was acquired.",
     )
     origin = sa.Column(sa.String, nullable=True, doc="Origin of the spectrum.")
+    type = sa.Column(
+        sa.String,
+        nullable=False,
+        default='source',
+        doc="Type of spectrum: source, host, etc. ",
```
acrellin, line 2524:
```suggestion
        doc="Type of spectrum: source, host, etc.",
```

skyportal/schema.py
```diff
@@ -1137,6 +1137,10 @@ class SpectrumPost(_Schema):
 
     origin = fields.String(required=False, description="Origin of the spectrum.")
 
+    type = fields.String(
+        required=False, description="Type of spectrum, e.g., source, host, etc."
```
acrellin, line 1141:
```suggestion
        required=False, description="Type of spectrum, e.g., source, host, etc. Defaults to source."
```

acrellin (commented):
Looks good overall, just a few questions/comments below!

skyportal/models.py
guynir42: I tried applying the migration but it fails on something unrelated to do with photometry owner_id...

guynir42: Nevermind I managed to get it working. 
The spectra from the demo data get type=source after migration.

profjsb:
I think we need to constrain the vocabulary here to make it easier on the front end to render, and also for API-based filtering (eventually). Rather than hardcode enums I think we should put something like this in the config:
```
spectrum_types:
   - source
   - host
   - model
   - host_center
   - default: source
   - view_tabs: ["source": [source],
            "host":   [host],
            "source_model": [source, model]]
```
that last bit can do into another PR when we handle the front-end implications.

guynir42:
> I think we need to constrain the vocabulary here to make it easier on the front end to render, and also for API-based filtering (eventually). Rather than hardcode enums I think we should put something like this in the config:
> 
> ```
> spectrum_types:
>    - source
>    - host
>    - model
>    - host_center
>    - default: source
>    - view_tabs: ["source": [source],
>             "host":   [host],
>             "source_model": [source, model]]
> ```
> 
> that last bit can do into another PR when we handle the front-end implications.

I'm not sure how putting that in the config will constrain (unless you want to also add some code in the handler to check the inputs and compare to the config). 

I think that leaving this up to the users to post whatever strings they like is important, but maybe it is good enough to keep things flexible via the config.

profjsb:
> > I think we need to constrain the vocabulary here to make it easier on the front end to render, and also for API-based filtering (eventually). Rather than hardcode enums I think we should put something like this in the config:
> > ```
> > spectrum_types:
> >    - source
> >    - host
> >    - model
> >    - host_center
> >    - default: source
> >    - view_tabs: ["source": [source],
> >             "host":   [host],
> >             "source_model": [source, model]]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > that last bit can do into another PR when we handle the front-end implications.
> 
> I'm not sure how putting that in the config will constrain (unless you want to also add some code in the handler to check the inputs and compare to the config).
> 
> I think that leaving this up to the users to post whatever strings they like is important, but maybe it is good enough to keep things flexible via the config.

I feel strongly we should not allow the (API) user to populate `spectrum.type` with whatever string they please. They already have `altdata` and `origin` as free-form repositories of unstructured data/strings. More important: we know we want to build generic front-end vizualizations and search/filtering on `spectrum.type` and this will be made much easier if we have a constrained vocabulary.

If you insist on hardcoding types and not using the config to define these, then use the `skyportal/enum_types.py` (like what we do with bandpasses). Here you could add `type` to `Spectrum` like in `models.py`:
```
from .enum_types import (
    allowed_spectral_types,
    default_spectral_type
....
class Spectrum(Base):
    ....
    type = sa.Column(
        allowed_spectral_types,
        nullable=False,
        default=default_spectral_type,
        doc='Type of spectrum (e.g. `source`),
    )
```
and in `skyportal/enum_types.py`:

```
## TODO: this could be grabbed from config one day
ALLOWED_SPECTRAL_TYPES = ("source", "model", "host")
allowed_spectral_types = sa.Enum(
    *ALLOWED_SPECTRAL_TYPES, name="spectral_types", validate_strings=True
)
default_spectral_type = ALLOWED_SPECTRAL_TYPES[0]
```

guynir42:
> > > I think we need to constrain the vocabulary here to make it easier on the front end to render, and also for API-based filtering (eventually). Rather than hardcode enums I think we should put something like this in the config:
> > > ```
> > > spectrum_types:
> > >    - source
> > >    - host
> > >    - model
> > >    - host_center
> > >    - default: source
> > >    - view_tabs: ["source": [source],
> > >             "host":   [host],
> > >             "source_model": [source, model]]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > that last bit can do into another PR when we handle the front-end implications.
> > 
> > 
> > I'm not sure how putting that in the config will constrain (unless you want to also add some code in the handler to check the inputs and compare to the config).
> > I think that leaving this up to the users to post whatever strings they like is important, but maybe it is good enough to keep things flexible via the config.
> 
> I feel strongly we should not allow the (API) user to populate `spectrum.type` with whatever string they please. They already have `altdata` and `origin` as free-form repositories of unstructured data/strings. More important: we know we want to build generic front-end vizualizations and search/filtering on `spectrum.type` and this will be made much easier if we have a constrained vocabulary.
> 
> If you insist on hardcoding types and not using the config to define these, then use the `skyportal/enum_types.py` (like what we do with bandpasses). Here you could add `type` to `Spectrum` like in `models.py`:
> 
> ```
> from .enum_types import (
>     allowed_spectral_types,
>     default_spectral_type
> ....
> class Spectrum(Base):
>     ....
>     type = sa.Column(
>         allowed_spectral_types,
>         nullable=False,
>         default=default_spectral_type,
>         doc='Type of spectrum (e.g. `source`),
>     )
> ```
> 
> and in `skyportal/enum_types.py`:
> 
> ```
> ## TODO: this could be grabbed from config one day
> ALLOWED_SPECTRAL_TYPES = ("source", "model", "host")
> allowed_spectral_types = sa.Enum(
>     *ALLOWED_SPECTRAL_TYPES, name="spectral_types", validate_strings=True
> )
> default_spectral_type = ALLOWED_SPECTRAL_TYPES[0]
> ```

I think constraining it through the config is much better than hard coding it. I'm just not clear how that works. I can add a paragraph like you suggested above, but I don't see how that would be enforced.

guynir42:
> > > I think we need to constrain the vocabulary here to make it easier on the front end to render, and also for API-based filtering (eventually). Rather than hardcode enums I think we should put something like this in the config:
> > > ```
> > > spectrum_types:
> > >    - source
> > >    - host
> > >    - model
> > >    - host_center
> > >    - default: source
> > >    - view_tabs: ["source": [source],
> > >             "host":   [host],
> > >             "source_model": [source, model]]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > that last bit can do into another PR when we handle the front-end implications.
> > 
> > 
> > I'm not sure how putting that in the config will constrain (unless you want to also add some code in the handler to check the inputs and compare to the config).
> > I think that leaving this up to the users to post whatever strings they like is important, but maybe it is good enough to keep things flexible via the config.
> 
> I feel strongly we should not allow the (API) user to populate `spectrum.type` with whatever string they please. They already have `altdata` and `origin` as free-form repositories of unstructured data/strings. More important: we know we want to build generic front-end vizualizations and search/filtering on `spectrum.type` and this will be made much easier if we have a constrained vocabulary.
> 
> If you insist on hardcoding types and not using the config to define these, then use the `skyportal/enum_types.py` (like what we do with bandpasses). Here you could add `type` to `Spectrum` like in `models.py`:
> 
> ```
> from .enum_types import (
>     allowed_spectral_types,
>     default_spectral_type
> ....
> class Spectrum(Base):
>     ....
>     type = sa.Column(
>         allowed_spectral_types,
>         nullable=False,
>         default=default_spectral_type,
>         doc='Type of spectrum (e.g. `source`),
>     )
> ```
> 
> and in `skyportal/enum_types.py`:
> 
> ```
> ## TODO: this could be grabbed from config one day
> ALLOWED_SPECTRAL_TYPES = ("source", "model", "host")
> allowed_spectral_types = sa.Enum(
>     *ALLOWED_SPECTRAL_TYPES, name="spectral_types", validate_strings=True
> )
> default_spectral_type = ALLOWED_SPECTRAL_TYPES[0]
> ```

And if the types in the config were to change, would that require a database migration?

profjsb:
@guynir42 to make this work with configs, do something like
```
spectral_types:
   types:
      - source
      - host
      - model
   - default: source
 ```
in `config.defaults`. Then in `enum_types.py`:
```
...
_, cfg = load_env()
...
ALLOWED_SPECTRAL_TYPES = tuple(cfg.get('spectral_types.types', ["source", "model", "host"]))
allowed_spectral_types = sa.Enum(
    *ALLOWED_SPECTRAL_TYPES, name="spectral_types", validate_strings=True
)
default_spectral_type = cfg.get('spectral_types.default', "source")
```

**Comment by guynir42** [comment]
> default_spectral_type = cfg.get('spectral_types.default', "source")

Yeah I think I got it now. 

Also, I'm not adding a "model" spectrum type. I think model spectra should be handled via annotations, and not as another form of spectrum. For example, a spectrum requires "observed at" and "instrument" and these things don't belong on a model spectrum.

**Comment by guynir42** [comment]
Test failures appear unrelated. 

Please have another look at this, I believe it is ready to merge... @profjsb @acrellin

File: skyportal/models.py (hunk 4 above)
Comment (dmitryduev) [reply skyportal/models.py]: How long did it take to migrate spectra locally? We have 4,360 spectra in the prod db -- how long would it take to update them?

File: skyportal/handlers/api/spectrum.py (hunk 2 above)
Comment (dmitryduev) [reply skyportal/handlers/api/spectrum.py:154]: So if it is a bug, can you please fix it, too? Or make sure it's not a bug. There was an issue when uploading a spectrum, it wouldn't appear on the source page unless you reload the page, but @acrellin fixed it recently.

File: skyportal/models.py [hunk 6]
```diff
@@ -2517,6 +2519,12 @@ class Spectrum(Base):
         doc="Median UTC ISO time stamp of the exposure or exposures in which the Spectrum was acquired.",
     )
     origin = sa.Column(sa.String, nullable=True, doc="Origin of the spectrum.")
+    type = sa.Column(
+        allowed_spectrum_types,
+        nullable=False,
+        default=default_spectrum_type,
+        doc="Type of spectrum: source, host, etc.",
```
dmitryduev, line 2526: Must be one of `ALLOWED_SPECTRUM_TYPES`, not etc, no?

skyportal/schema.py
```diff
@@ -1137,6 +1137,11 @@ class SpectrumPost(_Schema):
 
     origin = fields.String(required=False, description="Origin of the spectrum.")
 
+    type = fields.String(
+        required=False,
+        description="Type of spectrum, e.g., source, host, etc. Default to source.",
```
dmitryduev, line 1142:
```suggestion
        description="Type of spectrum, one of 'source', 'host', 'host_center'. Defaults to source.",
```

dmitryduev (changes requested):

dmitryduev:
@guynir42 also please provide a summary of the changes in the PR description.

skyportal/handlers/api/spectrum.py
```diff
@@ -133,6 +133,8 @@ def post(self):
         spec.instrument = instrument
         spec.groups = groups
         spec.owner_id = owner_id
+        if spec.type is None:
+            spec.type = 'source'
```
profjsb, line 137: Don't hard code this since, it could be difference in different configs:
`spec.type = cfg.get('spectrum_types.default', "source")`

profjsb (commented):
Will await your responses to @dmitryduev. Also, it would be good to add a quick test that a) makes sure that the spec.type is the default when no specified upon upload and b) that you can retrieve a spectrum with type != source and check its type.

acrellin:
Will review when others' comments are addressed; I haven't forgotten about this one :)

skyportal/handlers/api/spectrum.py
guynir42: This is the bug fix. The only thing is that I'm not sure why we are using internal key and not just the object id in all the refresh messages.

guynir42:
> Will await your responses to @dmitryduev. Also, it would be good to add a quick test that a) makes sure that the spec.type is the default when no specified upon upload and b) that you can retrieve a spectrum with type != source and check its type.

Do you mean add a test in the test suite or just try it out locally?

alembic/versions/b05b6c4d1a17_spectrum_type.py
guynir42: Yes it now works locally with old spectra getting automatically populated as "source".

skyportal/models.py
guynir42: yes, correct.

skyportal/handlers/api/spectrum.py
acrellin: We were using the internal key because, for a time, object IDs were considered sensitive, access-restricted information. But since objects became read-public, some message handlers use `obj_id` while others still use `internal_key`. I confirmed that the `spectra` message handler does indeed expect `obj_id`, so this looks good to me. We should bring all of these into consistency soon, while I'll file an issue for (perhaps involving dropping `Obj.internal_key`, as that's no longer relevant).

skyportal/models.py
guynir42: I only have one or two spectra locally, so I really don't know. But I think adding a few thousand strings should not take long under any reasonable conditions...

skyportal/handlers/api/spectrum.py
guynir42: Great. I was hoping I wouldn't need to go replace all the object ids with internal keys.

profjsb:
> > Will await your responses to @dmitryduev. Also, it would be good to add a quick test that a) makes sure that the spec.type is the default when no specified upon upload and b) that you can retrieve a spectrum with type != source and check its type.
> 
> Do you mean add a test in the test suite or just try it out locally?

Add a few tests to the existing spectrum API tests.

dmitryduev (approved):
LGTM!

profjsb (approved):
Looks great, thanks @guynir42

profjsb:
Please ensure that the failing test (`test_submit_new_followup_request_Spectral`) is indeed not related.

skyportal/enum_types.py
```diff
@@ -2,16 +2,23 @@
...
 
+_, cfg = load_env()
+
 
 def force_render_enum_markdown(values):
     return ', '.join(list(map(lambda v: f'`{v}`', values)))
 
 
+ALLOWED_SPECTRUM_TYPES = tuple(
+    cfg.get('spectrum_types.types', ['source', 'host', 'host center'])
```
stefanv, line 20: `host_center`?

guynir42:
Test failures on followup requests seem to be unrelated...

guynir42:
Test failure is unrelated (this test passed locally).

alembic/versions/5b4d9d54142e_add_spectrum_type_enum.py
```diff
@@ -0,0 +1,43 @@
...
+revision = '5b4d9d54142e'
+down_revision = '171e9ea18885'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute(
+        '''CREATE TYPE "spectrumtypes" AS ENUM ('source', 'host', 'host_center')'''
```
acrellin, line 22: Doesn't this need to come from the config?

alembic/versions/5b4d9d54142e_add_spectrum_type_enum.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+    op.add_column(
+        'spectra',
+        sa.Column(
+            'type',
+            sa.Enum('source', 'host', 'host_center', name='spectrumtypes'),
```
acrellin, line 28: Same here

alembic/versions/5b4d9d54142e_add_spectrum_type_enum.py
```diff
@@ -0,0 +1,43 @@
...
+            nullable=True,
+        ),
+    )
+    # ### end Alembic commands ###
+
+    spectra = sa.Table('spectra', sa.MetaData(), sa.Column('type', sa.Enum()))
+    conn = op.get_bind()
+    conn.execute(spectra.update().where(spectra.c.type.is_(None)).values(type='source'))
```
acrellin, line 36: The default should come from the config

skyportal/models.py
```diff
@@ -2517,6 +2520,13 @@ class Spectrum(Base):
         doc="Median UTC ISO time stamp of the exposure or exposures in which the Spectrum was acquired.",
     )
     origin = sa.Column(sa.String, nullable=True, doc="Origin of the spectrum.")
+    type = sa.Column(
+        allowed_spectrum_types,
+        nullable=False,
+        default=default_spectrum_type,
+        doc=f'''Type of spectrum. One of: {', '.join(f"'{t}'" for t in ALLOWED_SPECTRUM_TYPES)}.
+                        Defaults to 'f{default_spectrum_type}'.''',
```
acrellin, line 2528:
```suggestion
        doc=f'''Type of spectrum. One of: {', '.join(f"'{t}'" for t in ALLOWED_SPECTRUM_TYPES)}.
                Defaults to 'f{default_spectrum_type}'.''',
```

skyportal/tests/api/test_spectrum.py
```diff
@@ -819,3 +821,59 @@ def test_spectrum_plot_only_some_spectra(
     assert len(instruments) == 2
     assert 'LRIS' in str(instruments)
     assert 'SEDM' in str(instruments)
+
+
+def test_post_get_spectrum_type(upload_data_token, public_source, public_group, lris):
```
acrellin, line 826: Let's also add a test to ensure that providing an invalid type is handled properly.

acrellin (changes requested):
Looks good overall, thanks @guynir42 ! Main comment is regarding the migration script and adding an additional test (see below).

stefanv:
Whoever does the merge, please edit the combined commit message.

acrellin (approved):

acrellin:
@guynir42 This "Spectral" front-end test is not a familiar flaky one. Please ensure that the failure is not related. https://github.com/skyportal/skyportal/pull/2300/checks?check_run_id=3806395597#step:15:581

guynir42:
> @guynir42 This "Spectral" front-end test is not a familiar flaky one. Please ensure that the failure is not related. https://github.com/skyportal/skyportal/pull/2300/checks?check_run_id=3806395597#step:15:581

@acrellin I ran the tests again and now fail on different followup requests.

PR #2303, opened 2021-09-30
Add handler for GET api/{sources,spectra}/comments endpoint

acrellin:
This patch adds a handler for fetching all comments associated with
the specified resource (obj, spectrum), and adds tests for this
handler as well as the already-existing annotations handler.

### DIFF
--- a/skyportal/tests/api/test_annotations.py
+++ b/skyportal/tests/api/test_annotations.py
@@ -397,3 +397,29 @@ def test_post_invalid_data(annotation_token, public_source, public_group):
 
     assert status == 400
     assert 'Invalid data' in data["message"]
+
+
+def test_fetch_all_annotations_on_obj(annotation_token, public_source, public_group):
+    status, data = api(
+        'POST',
+        f'sources/{public_source.id}/annotations',
+        data={
+            'obj_id': public_source.id,
+            'origin': 'kowalski',
+            'data': {'offset_from_host_galaxy': 1.5},
+            'group_ids': [public_group.id],
+        },
+        token=annotation_token,
+    )
+    assert status == 200
+
+    status, data = api(
+        'GET',
+        f'sources/{public_source.id}/annotations',
+        token=annotation_token,
+    )
+
+    assert status == 200
+    assert len(data['data']) == 1
+    assert data['data'][0]['data'] == {'offset_from_host_galaxy': 1.5}
+    assert data['data'][0]['origin'] == 'kowalski'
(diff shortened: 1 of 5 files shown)


skyportal/tests/api/test_annotations.py
```diff
@@ -397,3 +397,29 @@ def test_post_invalid_data(annotation_token, public_source, public_group):
...
+    assert status == 200
+
+    status, data = api(
+        'GET',
+        f'sources/{public_source.id}/annotations',
+        token=annotation_token,
+    )
+
+    assert status == 200
+    assert len(data['data']) == 1
```
stefanv, line 423: We should probably check with more than one annotation, to ensure that this follows the "fetch all" path.

stefanv (approved):
One minor tweak, then :+1:

skyportal/tests/api/test_annotations.py
acrellin: This already checks that - the data would be an object, not a length one list, otherwise

acrellin:
Failures are the typical flaky tests (unrelated)

skyportal/tests/api/test_annotations.py
stefanv: Two would make that more obvious, but no biggie.

guynir42: I'm not sure I understand. If we posted two annotations with different origins, we would get back two annotations...

skyportal/tests/api/test_comments_on_spectrum.py
```diff
@@ -274,3 +276,47 @@ def test_delete_comment(comment_token, upload_data_token, public_source, lris):
...
+
+    status, data = api(
+        'GET',
+        f'spectra/{spectrum_id}/comments',
+        token=upload_data_token,
+    )
+
+    assert status == 200
+    assert len(data['data']) == 1
+    assert data['data'][0]['text'] == comment_text
```
guynir42, line 322: Maybe two comments here, too?

guynir42 (changes requested):

skyportal/tests/api/test_annotations.py
acrellin: To be extra clear: this is already testing that we're going through the multiple annotations path, as we're ensuring that the response is a length-one list and that the first element is the actual annotation data we expect

acrellin: What's your confusion, Guy?

skyportal/tests/api/test_comments_on_spectrum.py
acrellin: See above -- this is already testing that we go through the "multiple" comments path

dmitryduev (approved):
LGTM 👍🏻

guynir42 (approved):

PR #2338, opened 2021-10-06
Make comment box scale to input size

bparazin:
Closes fritz-marshal/fritz-beta-feedback#103


bparazin:
Aha! Thanks for checking it out!

PR #2342, opened 2021-10-07
Run flaky tests in separate CI workflow

acrellin:
Run known flaky tests in separate GA workflow (which, when failed,
does not report failure for entire job).

Closes https://github.com/skyportal/skyportal/issues/2277


acrellin:
Even the flaky tests are green ;)

PR #2344, opened 2021-10-07
Update BL to include sudo-less db_init

acrellin:


acrellin:
@stefanv Has this been implemented?

stefanv:
We now detect whether sudo is present on Linux, and use it if found.

PR #2374, opened 2021-10-15
Refactor models.py into a submodule

stefanv:
models.py was deleted

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -1,7 +1,6 @@
 import tornado.web
 
 from baselayer.app.app_server import MainPageHandler
-from baselayer.app.models import init_db
 from baselayer.app.model_util import create_tables
 from baselayer.log import make_log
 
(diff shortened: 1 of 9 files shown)


dmitryduev:
Impressive 
image

stefanv:
I separated them for now to make it clear where everything comes from, but if you prefer to have them all in one spot that's easy to add. Let me know your preference!

acrellin:
@stefanv Of course! 

Regarding separate imports for BL models, as long as it's pretty obvious which models come from where, I'm fine with having them separated like this. If devs will need to be looking it up frequently (where do I import ___ from again??), it might be worth importing them all in the SP models init. Thoughts?

acrellin:
@stefanv What about changing the imports in the handlers back to being all from the SP models module?

stefanv:
> @stefanv What about changing the imports in the handlers back to being all from the SP models module?

OK, sure, done.  I left the imports in `skyportal.models` itself alone, since they are *part* of `skyportal.models`.

PR #2384, opened 2021-10-21
Spectrum type&label on front end

guynir42:
We make use of the newly added spectrum type and label to customize the display. 
1) The spectrum plot on the source page now shows tabs for different spectrum types. 

Peek 2021-10-21 15-54

2) The legend labels can be overwritten by the user-supplied label. 

3) This info can be added when posting spectra via the front end form. 

4) The type and label are added to the spectra table on the ManageData page.


guynir42:
Also I couldn't figure out how to make the columns for label/type invisible by default (in the spectrum table on the "ManageData" page). I set them to "display: false" but that did not change the fact that both columns are visible.

skyportal/models/schema.py
```diff
@@ -1035,6 +1035,17 @@ class SpectrumAsciiFilePostJSON(SpectrumAsciiFileParseJSON):
     instrument_id = fields.Integer(
         description='The ID of the instrument that took the spectrum.', required=True
     )
+    type = fields.String(
```
stefanv, line 1038: Should this be an enum populated from the configuration?

skyportal/plot.py
```diff
@@ -1398,397 +1400,423 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
-                        else ""
-                    )
-                ),
-                'origin': s.origin,
-                'altdata': altdata[:20] + "..." if len(altdata) > 20 else altdata,
-            }
+    layouts = []
+    spectrum_types = []
+    for spec_type in ALLOWED_SPECTRUM_TYPES:
+        spectra_of_same_type = [s for s in spectra if s.type == spec_type]
```
stefanv, line 1406: It looks like you want to build a dictionary of spectrum_type -> spectra, and then iterate over that.

skyportal/plot.py
```diff
@@ -1398,397 +1400,423 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
+        if len(spectra_of_same_type) == 0:
+            continue
+
+        rainbow = cm.get_cmap('rainbow', len(spectra_of_same_type))
+        palette = list(map(rgb2hex, rainbow(range(len(spectra_of_same_type)))))
+        color_map = dict(zip([s.id for s in spectra_of_same_type], palette))
+
+        spectrum_types.append(spec_type)
+
+        data = []
```
stefanv, line 1416: The inner part of the loop would become easier to read if it is factored out into a separate function.

skyportal/plot.py
```diff
@@ -1398,397 +1400,423 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
-        + legend_row_height * int(len(split) / legend_items_per_row)
-        + 30  # 30 is the height of the toolbar
-    )
-
-    # check browser plot_height for legend overflow
-    if device == "browser":
-        plot_height_of_legend = (
-            legend_row_height * int(len(split) / legend_items_per_row)
-            + 90  # 90 is height of toolbar plus legend offset
+        if device == "mobile_portrait":
```
stefanv, line 1500: This I'd also factor out into a lookup function

stefanv (commented):
Some restructuring of the plotting code would help to make this more readable, now that there is an additional upper loop.

Please also modify the existing spectrum upload test to use the new functionality, so we can verify that it does the right thing.

skyportal/plot.py
guynir42: Is there a clever way to do that other than a list comprehension expression that sorts each spectrum into the right label?

stefanv:
```
In [1]: spectra = [{'type': 'one', 'data': 1}, {'type': 'two', 'data': 2}]

In [2]: import collections

In [3]: spectra_by_type = collections.defaultdict(list)

In [4]: for s in spectra:
   ...:     spectra_by_type[s['type']].append(s)

```

guynir42:
@stefanv I tried to add the allowed spectrum types enum into the schema but I get errors when posting such as 
`spectrumtypes.host` is not in the allowed spectrum types enum (which does include `host` but without the prefix). I am not sure what I am doing wrong. I left the code I used as commented out rows, maybe you can tell me what I'm doing wrong...

skyportal/plot.py
```diff
@@ -1398,13 +1387,73 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
+    if using different spectrum types.
+
+    Parameters
+    ----------
+    obj:
+        The underlying object that is associated with all these spectra.
+    spectra:
+        The different spectra to be plotted. This can be a subset of
+        e.g., all the spectra of one type.
+    user:
```
stefanv, line 1438: Same as above: spacing + missing type.

skyportal/plot.py
```diff
@@ -1448,8 +1497,43 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
 
     smoothed_data = pd.concat(dfs)
 
+    # These values are equivalent from the photometry plot values
```
stefanv, line 1500: What does this mean?

skyportal/models/schema.py
```diff
@@ -1036,6 +1036,20 @@ class SpectrumAsciiFilePostJSON(SpectrumAsciiFileParseJSON):
         description='The ID of the instrument that took the spectrum.', required=True
     )
 
+    # type = ApispecEnumField(
```
stefanv, line 1039: Should probably be:

```
ApispecEnumField(
    Enum('type', py_allowed_spectrum_types), required=True/False
)
```

skyportal/plot.py
guynir42: This was here before I got to it but I can definitely improve on it :)

skyportal/models/schema.py
guynir42: Tried this. App doesn't start up. Tried replacing `py_allowed_spectrum_types` -> `ALLOWED_SPECTRUM_TYPES` and now I get errors when posting spectra (`'type.host' is not among the defined enum values. Enum name: spectrumtypes. Possible values: source, host, host_center`)

guynir42: This is the error I get when I use `py_allowed_spectrum_types`: 
`Traceback (most recent call last):
  File "baselayer/services/app/app.py", line 36, in <module>
    app_factory = getattr(importlib.import_module(module), app_factory)
  File "/home/guyn/anaconda3/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/guyn/skyportal/skyportal/app_server.py", line 7, in <module>
    from skyportal.handlers import BecomeUserHandler, LogoutHandler
  File "/home/guyn/skyportal/skyportal/handlers/__init__.py", line 11, in <module>
    from .become_user import BecomeUserHandler
  File "/home/guyn/skyportal/skyportal/handlers/become_user.py", line 2, in <module>
    from ..models import User
  File "/home/guyn/skyportal/skyportal/models/__init__.py", line 40, in <module>
    from .schema import setup_schema
  File "/home/guyn/skyportal/skyportal/models/schema.py", line 1031, in <module>
    class SpectrumAsciiFilePostJSON(SpectrumAsciiFileParseJSON):
  File "/home/guyn/skyportal/skyportal/models/schema.py", line 1041, in SpectrumAsciiFilePostJSON
    Enum('type', py_allowed_spectrum_types),
  File "/home/guyn/anaconda3/lib/python3.8/enum.py", line 341, in __call__
    return cls._create_(
  File "/home/guyn/anaconda3/lib/python3.8/enum.py", line 463, in _create_
    member_name, member_value = item
TypeError: cannot unpack non-iterable spectrumtypes object`

skyportal/plot.py
```diff
@@ -1398,13 +1387,73 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
+    for s in spectra:
+        spectra_by_type[s.type].append(s)
+
+    # go over the types in the order they're defined in the config
+    for spec_type in ALLOWED_SPECTRUM_TYPES:
+
+        if len(spectra_by_type[spec_type]) == 0:
+            continue
+
+        spectrum_types.append(spec_type)
```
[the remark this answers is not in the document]
stefanv: Currently, you have to read and parse the entire code snippet to understand what is going on.  If you put the sorting functionality in a function, then the user can simply accept that the function does what it says (sorts a dict by a list of keys), and then read that spectra_by_type now becomes sorted.

skyportal/models/schema.py
stefanv: That error message seems correct.  Why should the enum allow the value of `host.name`?

guynir42: I am posting with `type: 'host"` but that is translated to `type.host` and that isn't allowed...

skyportal/plot.py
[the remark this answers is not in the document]
guynir42: ok I see your point.

skyportal/models/schema.py
stefanv: Do you know where that translation happens and why?

guynir42: No, I just tried to do what is done for other classes and this is what happened...

stefanv: image

acrellin:
@guynir42 I just pushed a commit un-commenting out the enum type specification, which is working just fine for me locally

guynir42:
> @guynir42 I just pushed a commit un-commenting out the enum type specification, which is working just fine for me locally

Did you try posting a spectrum?

acrellin:
> > @guynir42 I just pushed a commit un-commenting out the enum type specification, which is working just fine for me locally
> 
> Did you try posting a spectrum? (you need to explicitly state `'type': 'source'` or any other legal type to see the error)

Yes, posting spectra works.

guynir42:
> > > @guynir42 I just pushed a commit un-commenting out the enum type specification, which is working just fine for me locally
> > 
> > 
> > Did you try posting a spectrum? (you need to explicitly state `'type': 'source'` or any other legal type to see the error)
> 
> Yes, posting spectra works.

Did you make sure the post has `type=source` in the payload?

skyportal/plot.py
```diff
@@ -1398,13 +1387,78 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
...
+        if len(spectra_by_type[spec_type]) == 0:
+            continue
+
+        layouts.append(
+            make_spectrum_layout(obj, spectra_by_type[spec_type], user, device, width)
+        )
+
+    if len(layouts) > 1:
+        panels = []
+        spectrum_types = [s for s in spectra_by_type]
```
stefanv, line 1417: This assumes that only valid spectrum types are given?

I think what you want then is to replace the above with:

```
# Filter out spectra with types that are not allowed
spectra_by_type = {key: spectra_by_type[key] for key in ALLOWED_SPECTRUM_TYPES if key in spectra_by_type}

for spec_type in spectra_by_type:
    layouts.append((spec_type, make_spectrum_layout(...))

if layouts:
    panels = []
    for (stype, layout) in layouts:
        panels.append(Panel(child=layout, title=stype))
```

guynir42: I think that is the same thing as what I'm doing with the function you suggested before. And also, the spectra must be one of the allowed spectrum types (this is enforced by the database).

stefanv: If that's the case, you can simply:

```
for spec_type in spectra_by_type:
    ...
```

And remove the `len(...) == 0` check.

guynir42: I think I just did that on the latest commit.

stefanv (approved):

profjsb:
Talked to the DESI folks today who definitely would like to see this functionality.

guynir42:
> Talked to the DESI folks today who definitely would like to see this functionality.

Yes, I have been talking to them about this for a few weeks. 

At least the backend is ready for this, so they can already start posting data. But it would sure be useful for them to have visual feedback. 

Right now we are just waiting for a second review (Ari) and I think it shouldn't take long after a bunch of improvements went in thanks to Stefan's suggestions.

acrellin (approved):

PR #2390, opened 2021-10-26
Enabling point to skymap crossmatching

mcoughlin:
Uses healpix-alchemy updates to slice sources based on skymaps.


static/js/components/GcnEventPage.jsx
```diff
@@ -56,6 +61,79 @@ const useStyles = makeStyles((theme) => ({
...
+    };
+    if (sortData && Object.keys(sortData).length > 0) {
+      data.sortBy = sortData.name;
+      data.sortOrder = sortData.direction;
+    }
+    dispatch(sourcesActions.fetchGcnEventSources(route.dateobs, data));
+  };
+
+  // eslint-disable-next-line
+  if (sources.sources.length === 0) {
```
acrellin, line 101:
```suggestion
  if (sources.sources.length === 0) {
```
why is eslint complaining? 
if you change this to 
```javascript
  if (!sources.sources.length) {
```
does it still complain? we'd like to minimize the amount of eslint disabling we're doing

static/js/components/GcnEventPage.jsx
```diff
@@ -56,6 +61,79 @@ const useStyles = makeStyles((theme) => ({
...
+    );
+  }
+
+  return (
+    <div className={classes.source}>
+      <Typography variant="h4" gutterBottom align="center">
+        Event sources
+      </Typography>
+
+      {!!sources.sources && (
```
acrellin, line 121:
```suggestion
      {!!sources.sources.length && (
```

static/js/components/GcnEventPage.jsx
```diff
@@ -148,25 +253,35 @@ const Localization = ({ loc }) => {
...
+  }, [route, dispatch]);
+
   if (!gcnEvent) {
     return <CircularProgress />;
   }
 
+  if (!gcnEventSources) {
+    return <CircularProgress />;
+  }
+
```
acrellin, line 284:
```suggestion
  if (!gcnEvent || !gncEventSources) {
    return <CircularProgress />;
  }
```

static/js/components/GcnEventPage.jsx
```diff
@@ -148,25 +253,35 @@ const Localization = ({ loc }) => {
...
 
+  const gcnEvent = useSelector((state) => state.gcnEvent);
+  const gcnEventSources = useSelector((state) => state.sources.gcnEventSources);
+
   useEffect(() => {
     dispatch(gcnEventActions.fetchGcnEvent(route.dateobs));
   }, [route, dispatch]);
 
+  useEffect(() => {
+    dispatch(sourcesActions.fetchGcnEventSources(route.dateobs, {}));
```
acrellin, line 274:
```suggestion
    dispatch(sourcesActions.fetchGcnEventSources(route.dateobs));
```
(this parameter already defaults to that)

static/js/ducks/sources.js
```diff
@@ -73,6 +101,10 @@ messageHandler.add((actionType, payload, dispatch, getState) => {
       dispatch(fetchFavoriteSources());
     }
   }
+  const { gcnEvent } = getState();
+  if (actionType === FETCH_GCNEVENT_SOURCES) {
+    dispatch(fetchGcnEventSources(gcnEvent.dateobs, {}));
```
acrellin, line 106:
```suggestion
    dispatch(fetchGcnEventSources(gcnEvent?.dateobs));
```

static/js/ducks/sources.js
```diff
@@ -73,6 +101,10 @@ messageHandler.add((actionType, payload, dispatch, getState) => {
       dispatch(fetchFavoriteSources());
     }
   }
+  const { gcnEvent } = getState();
```
acrellin, line 104: What if there isn't a gcn event loaded? Won't this still trigger all gcn events being fetched?

acrellin (commented):
Looking good so far, a few comments below, LMK if you want a hand with anything here.

static/js/ducks/sources.js
mcoughlin: What do you suggest here?

skyportal/handlers/api/candidate.py
```diff
@@ -991,6 +993,9 @@ def post(self):
         if dec is None and not obj_already_exists:
             return self.error("Dec must not be null for a new Obj")
 
+        if not obj_already_exists:
```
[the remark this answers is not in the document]
mcoughlin: Agreed.

skyportal/handlers/api/source.py
```diff
@@ -1047,6 +1057,36 @@ class Validator(Schema):
...
+                    .first()
+                )
+            else:
+                localization = (
+                    Localization.query_records_accessible_by(self.current_user)
+                    .filter(Localization.dateobs == localization_dateobs)
+                    .first()
+                )
+            if localization is None:
+                return self.error("Localization not found", status=404)
```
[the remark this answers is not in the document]
mcoughlin: Added.

static/js/ducks/sources.js
acrellin: Only do the dispatch below if `gcnEvent` is not null/empty/etc.

skyportal/handlers/api/gcn.py
```diff
@@ -241,6 +249,44 @@ def add_contour(localization_id, request_handler):
         DBSession.remove()
 
 
+def add_tiles(localization_id, request_handler):
```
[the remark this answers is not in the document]
mcoughlin: Ok I think the new version is closer... what do you think?

skyportal/handlers/api/source.py
```diff
@@ -1327,6 +1367,41 @@ class Validator(Schema):
             query_results["sources"] = obj_list
 
         query_results = recursive_to_dict(query_results)
+        if includeGeojson:
+            # features are JSON representations that the d3 stuff understands.
+            # We use these to render the contours of the sky localization and
+            # locations of the transients.
+
+            # FIXME: currently necessary for the sources to render
+            features = [
```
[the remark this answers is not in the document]
mcoughlin: I don't understand anything. I am sure I am just wrong but I could use someone more experienced in JS than me.

static/js/ducks/sources.js
mcoughlin: Ok I think that hopefully fixed it.

skyportal/handlers/api/source.py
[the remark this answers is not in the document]
acrellin: How did you arrive at this in the first place?

static/js/ducks/sources.js
```diff
@@ -74,6 +102,13 @@ messageHandler.add((actionType, payload, dispatch, getState) => {
     }
   }
 
+  const { gcnEvent } = getState();
+  if (actionType === FETCH_GCNEVENT_SOURCES) {
+    if (gcnEvent?.dateobs) {
+      dispatch(fetchGcnEventSources(gcnEvent?.dateobs));
+    }
+  }
```
acrellin, line 110: Wouldn't this message have a payload like `gcnEventId: ...`? If so, we want to check whether the currently loaded `gcnEvent.id` matches the message payload ID, and only do the fetch then. Something like: 

```suggestion
  if (actionType === FETCH_GCNEVENT_SOURCES) {
    if (gcnEvent && gcnEvent.id === payload.gcnEvent.id) {
      dispatch(fetchGcnEventSources(gcnEvent.dateobs));
    }
  }
```

static/js/ducks/sources.js
acrellin: Resolving in favor of comment below

skyportal/handlers/api/gcn.py
```diff
@@ -221,24 +225,64 @@ def delete(self, dateobs):
         return self.success()
 
 
-def add_contour(localization_id, request_handler):
+def add_contour(localization_id, user_id):
+    session = Session()
     try:
         localization = (
-            Localization.query_records_accessible_by(request_handler.current_user)
+            Localization.query_records_accessible_by(user_id)
```
acrellin, line 232:
```suggestion
            Localization.query_records_accessible_by(session.query(User).get(user_id))
```

skyportal/handlers/api/gcn.py
```diff
@@ -221,24 +225,64 @@ def delete(self, dateobs):
...
     finally:
-        DBSession.remove()
+        session.remove()
+
+
+def add_tiles(localization_id, user_id):
+    session = Session()
+    try:
+        localization = (
+            Localization.query_records_accessible_by(user_id)
```
acrellin, line 253:
```suggestion
            Localization.query_records_accessible_by(session.query(User).get(user_id))
```

skyportal/handlers/api/source.py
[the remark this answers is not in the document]
acrellin: That will hopefully help me understand why this is needed and how to implement a workaround

skyportal/models/localization.py
```diff
@@ -141,3 +146,32 @@ def flat(self):
...
+        doc="The Localization that this tile belongs to",
+    )
+
+    probdensity = sa.Column(
+        sa.Float,
+        nullable=False,
+        doc="Probability density for the tile",
+    )
+
+    cumprob = sa.Column(
```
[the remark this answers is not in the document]
acrellin: Just want to make sure this is indeed supposed to be nullable

static/js/components/GcnEventPage.jsx
acrellin: ⬆️

skyportal/handlers/api/source.py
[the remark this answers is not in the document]
mcoughlin: Just ... trying stuff til it worked :)

static/js/ducks/sources.js
```diff
@@ -58,6 +67,25 @@ export function fetchFavoriteSources(filterParams = {}) {
   return API.GET("/api/sources", FETCH_FAVORITE_SOURCES, filterParams);
 }
 
+export function fetchGcnEventSources(dateobs = null, filterParams = {}) {
+  addFilterParamDefaults(filterParams);
+  filterParams.localizationDateobs = dateobs;
+  filterParams.localizationCumprob = 0.95;
```
acrellin, line 73: Why does this default to 0.95 here but 1.01 on the backend?

static/js/ducks/sources.js
```diff
@@ -58,6 +67,25 @@ export function fetchFavoriteSources(filterParams = {}) {
...
+
+  if (dateobs) {
+    filterParams.startDate = dayjs(dateobs).format("YYYY-MM-DD HH:mm:ss");
+    filterParams.endDate = dayjs(dateobs)
+      .add(7, "day")
+      .format("YYYY-MM-DD HH:mm:ss");
+  }
+
+  filterParams.startDate = null;
```
acrellin, line 82: Isn't this just wiping out the values set above?

acrellin (approved):
Looking better! Some suggestions below -- this is getting really close. Thanks @mcoughlin . I'll be sure to follow up more quickly after you've addressed the below! (please re-tag me for review, otherwise I won't know you're awaiting another review from me).

static/js/ducks/sources.js
mcoughlin: 0.95 makes more sense indeed. Fixing.

static/js/ducks/sources.js
mcoughlin: Commenting this (Debugging line) out!

mcoughlin:
@acrellin I think the cumulative probability stuff can go away with Leo's new healpix_alchemy thing.. so let's wait on that comment. I am less sure what the comment about: 

 // eslint-disable-next-line
  if (sources.sources.length === 0) {
  
  Is referring to...

static/js/components/GcnEventPage.jsx
acrellin: @mcoughlin I'm wondering why you need to disable eslint here. What's the rule that causes it to complain?

mcoughlin: I get with your suggestion:

  100:24  error  'sources.sources.length' is missing in props validation  react/prop-types
  120:26  error  'sources.sources.length' is missing in props validation  react/prop-types
  276:21  error  'gncEventSources' is not defined                         no-undef

acrellin:
Is sources listed as an array in the prop types? If so that's pretty silly
of eslint, and you can leave the disable comment there. Thanks for
looking into it. Fine as is.

mcoughlin:
It does look like it at least!

mcoughlin:
@lpsinger @acrellin I rewrote the ingestor and model to hopefully use healpix_alchemy properly. Before I was just using cumulative probability in the tiles saved in the tile:
https://github.com/mcoughlin/skyportal/blob/crossmatch2/skyportal/handlers/api/source.py#L1089
to determine if an object is inside or outside the 95%. Now what is the correct path here @lpsinger ?

skyportal/models/localization.py
```diff
@@ -141,3 +146,31 @@ def flat(self):
             return healpy.reorder(result, 'NESTED', 'RING')
         else:
             return (self.flat_2d,)
+
+    tiles = relationship("LocalizationTile")
```
lpsinger, line 150:
```suggestion
```

The relationships are not necessary. They are not used in the queries, and I had removed them from my example code.

Removing the relationships will also speed up loading the model instances because it won't load the dependent columns.

skyportal/models/localization.py
```diff
@@ -141,3 +146,31 @@ def flat(self):
...
+        nullable=False,
+        doc='localization ID',
+    )
+
+    localization = relationship(
+        "Localization",
+        foreign_keys=localization_id,
+        doc="The Localization that this tile belongs to",
+        passive_deletes=True,
+    )
```
lpsinger, line 168: Again, you can safely delete the relationships unless you think you will actually use them for something.

skyportal/models/localization.py
```diff
@@ -141,3 +146,31 @@ def flat(self):
...
+    tiles = relationship("LocalizationTile")
+
+
+class LocalizationTile(Base):
+    """This is a single tile within a skymap (as in the Localization table).
+    Each tile has an associated healpix id and probability density."""
+
+    localization_id = sa.Column(
+        sa.ForeignKey('localizations.id', ondelete="CASCADE"),
+        nullable=False,
```
lpsinger, line 159:
```suggestion
        primary_key=True,
```

Otherwise Postgres would complain if you had two different sky maps that contained the same HEALPix tile.```

skyportal/models/obj.py
```diff
@@ -215,6 +216,8 @@ class Obj(Base, conesearch_alchemy.Point):
         sa.ARRAY(sa.String), nullable=True, doc="Alternative names for this object."
     )
 
+    healpix = sa.Column(healpix_alchemy.Point, index=True)
```
lpsinger, line 219:
```suggestion
    healpix = sa.Column(healpix_alchemy.Point, index=True, nullable=False)
```

skyportal/handlers/api/source.py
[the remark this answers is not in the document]
lpsinger: The old D3 code from the GROWTH ToO Marshal might have used the list of sources to automatically pan the view. Maybe your D3 code has a corner case where zero points generates nans?

lpsinger (changes requested):

lpsinger:
I'm not crazy about the name of this PR. I don't see what this has to do with slicing.

skyportal/models/obj.py
mcoughlin: @acrellin has examples of objects without RAs and Decs (they are just None), so I think this needs to be nullable.

skyportal/models/localization.py
mcoughlin: Great point! Committed.

skyportal/models/localization.py
mcoughlin: Great point! Committed.

skyportal/models/obj.py
lpsinger: Why?

mcoughlin: If RA and Dec is not defined, healpix is just None right?

lpsinger (changes requested):

skyportal/handlers/api/gcn.py
```diff
@@ -20,12 +20,16 @@
...
     GcnTag,
     Localization,
+    LocalizationTile,
 )
 from ...utils.gcn import get_dateobs, get_tags, get_skymap, get_contour
 
 log = make_log('api/gcn_event')
 
 
+Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
```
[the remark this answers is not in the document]
mcoughlin: This is the model SP uses for running requests that take long.

requirements.txt
```diff
@@ -22,7 +22,10 @@ marshmallow-enum==1.5.1
 Pillow==8.3.2
 sncosmo==2.5.0
 tdtax==0.1.3
+healpix-alchemy @ git+https://github.com/skyportal/healpix-alchemy.git@95064ae5e06bfe9e50162070212d3d6719fb3bdc#egg=healpix_alchemy
```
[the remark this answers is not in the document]
mcoughlin: yay!

acrellin:
API test failure unrelated; front-end test failure (GCN events) seems related

mcoughlin:
@acrellin I *think* it is 1.3 vs 1.4 but Leo says 1.3 should be fine, so I am stumped.

lpsinger:
All I see in the test log is a Selenium timeout error. Is the actual sqlalchemy error in some other log that is not captured here?

PR #2391, opened 2021-10-27
Galaxy table

mcoughlin:
Adding a table for galaxies, relevant for scheduling galaxy-based follow-up


skyportal/tests/api/test_galaxy.py
```diff
@@ -0,0 +1,22 @@
...
+    data = {
+        'catalog_name': 'CLU_mini',
+        'catalog_data': Table.read(datafile).to_pandas().to_dict(orient='list'),
+    }
+
+    status, data = api('POST', 'galaxy', data=data, token=super_admin_token)
+    assert status == 200
+    assert data['status'] == 'success'
+
+    status, data = api('GET', 'galaxy', token=super_admin_token)
```
profjsb, line 19: use a `view_only_token` here.

requirements.txt
```diff
@@ -22,7 +22,8 @@ marshmallow-enum==1.5.1
 Pillow==8.3.2
 sncosmo==2.5.0
 tdtax==0.1.3
-healpix-alchemy==0.1.2
+mocpy==0.8.5
```
profjsb, line 25: where is this imported or it a "must be pinned" dependency of something else?

skyportal/models/galaxy.py
```diff
@@ -0,0 +1,23 @@
+__all__ = ['Galaxy']
+
+import sqlalchemy as sa
+import healpix_alchemy as ha
+
+from baselayer.app.models import Base
+
+
+class Galaxy(Base, ha.Point):
+    """A record of a galaxy and its metadata, such as position,
```
profjsb, line 10: Should probably have an "alt_name" list field here, since there may be many names for the galaxy.

Would we ever have the case that a `Galaxy` would need user annotations or other `altdata` fields for storing generic data?

profjsb: Some catalogs may have redshift and redshift uncertainty, distance uncertainty, etc.

profjsb (changes requested):
I'd suggesting calling this api something like `galaxy_catalog` and the handler `GalaxyCatalogHandler`. This is because you are doing bulk upload and download here and we may (soon) want to have a `GET` on specific set of galaxies that match a certain criteria (distance > X, within Y deg of this RA, DEC).

requirements.txt
mcoughlin: I am actually thinking here that we should wait for conesearch-alchemy to get dropped in on master and then rebase.

skyportal/tests/api/test_galaxy.py
mcoughlin: Fixed

skyportal/models/galaxy.py
mcoughlin: I went ahead and added alt_name, redshift and redshift uncertainty, and distance uncertainty. Maybe the group can discuss if they want the altdata bit.

mcoughlin:
@acrellin I wonder if this is most close to actually making sense to try to get over the finish line?

acrellin:
@mcoughlin What should we be looking for here instead of `healpix`?
```
Traceback (most recent call last):                                                                                 
  File "/home/arien/anaconda3/envs/skyportal38/lib/python3.8/site-packages/tornado/web.py", line 1699, in _execute 
    result = await result                                                                                          
  File "/home/arien/projects/skyportal/skyportal/handlers/api/galaxy.py", line 36, in post                         
    catalog_data["nested"] = ha.healpix.HPX.lonlat_to_healpix(                                                     
AttributeError: module 'healpix_alchemy' has no attribute 'healpix'                                                

```

mcoughlin:
@acrellin We can get rid of the healpix stuff now, as these Galaxy's are just like point sources now functioning under the cross-match alchemy stuff. So the nested can go.

requirements.txt
```diff
@@ -23,6 +23,8 @@ Pillow==8.3.2
 sncosmo==2.5.0
 tdtax==0.1.3
 conesearch-alchemy==1.0.0
+mocpy==0.8.5
+healpix-alchemy @ git+https://github.com/skyportal/healpix-alchemy.git@cf7acbd3d0aec6fbb71a31e072e067c874de8181#egg=healpix_alchemy
```
acrellin, line 27: @mcoughlin I'm pretty confused about all the healpix-alchemy stuff... is this the version that should be pinned here? Galaxy API tests pass with this.

mcoughlin:
@acrellin I think for this we can remove any healpix_alchemy dependence.

acrellin:
@mcoughlin and replace all references to ha in the galaxy model definition with conesearch-alchemy?

mcoughlin:
@acrellin yes. We don't need the nested line in the model though.

acrellin:
Green!

profjsb:
Lots of test failures. I'll take a close look at this once these go green.

mcoughlin:
@acrellin The migration fails with:
CREATE INDEX ix_galaxys_point ON public.galaxys USING btree (((cosd(ra) * cosd("dec"))), ((sind(ra) * cosd("dec"))), sind("dec"));
Do you know how to fix this given the commands contained? I guess this is because Galaxy inherits the conesearch-alchemy Point class...

skyportal/tests/api/test_galaxy.py
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    status, data = api('POST', 'galaxy_catalog', data=data, token=super_admin_token)
+    assert status == 200
+    assert data['status'] == 'success'
+
+    status, data = api('GET', 'galaxy_catalog', token=view_only_token)
+    assert status == 200
+    data = data["data"]
+    assert len(data) == 10
```
acrellin, line 22: I'd suggest we check that the values of at least one of the returned resources is what we'd expect

acrellin (approved):
Pending the test expansion, this looks good to me 👍🏼

skyportal/handlers/api/galaxy.py
```diff
@@ -0,0 +1,104 @@
...
+import astropy.units as u
+import healpix_alchemy as ha
+
+from ..base import BaseHandler
+from ...models import DBSession, Galaxy
+
+
+class GalaxyCatalogHandler(BaseHandler):
+    @permissions(['System admin'])
+    async def post(self):
```
stefanv, line 12: Why async?

skyportal/handlers/api/galaxy.py
```diff
@@ -0,0 +1,104 @@
...
+              application/json:
+                schema: Success
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+
+        data = self.get_json()
+        catalog_name = data.get('catalog_name')
```
stefanv, line 34: What if these are not specified?

mcoughlin: @stefanv adding an error if this occurs.

skyportal/handlers/api/galaxy.py
mcoughlin: I was thinking in case the database write was slow? In principle, this can run out of time right?

requirements.txt
mcoughlin: This has been resolved.

skyportal/tests/api/test_galaxy.py
mcoughlin: Added from @acrellin's suggestion.

alembic/versions/1aeb080b0eda_galaxy_table.py
```diff
@@ -0,0 +1,62 @@
...
+revision = '1aeb080b0eda'
+down_revision = '80434b53562c'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        'galaxys',
```
profjsb, line 22: This is an awkward plural. Not sure how we handle table naming in models/ but this table should be either `galaxy` or `galaxies`. Likewise the indices should use the proper plural.

skyportal/handlers/api/galaxy.py
```diff
@@ -0,0 +1,109 @@
...
+        catalog_data = data.get('catalog_data')
+
+        if catalog_name is None:
+            return self.error("catalog_name is a required parameter.")
+        if catalog_data is None:
+            return self.error("catalog_data is a required parameter.")
+
+        galaxies = [
```
profjsb, line 42: There's no enforcement of the data structure of the posted json. That's needed. Also bounds checking is needed otherwise we could get nonsensical values added to the DB. This is the place to ensure proper values in the DB. Could defer the checking to the Galaxy post and catch errors here instead.

profjsb (changes requested):
Needs some bounds checking and table name change. Otherwise good.

alembic/versions/1aeb080b0eda_galaxy_table.py
mcoughlin: @acrellin @stefanv The "s" is adding automatically I guess. Any thoughts on this?

skyportal/handlers/api/galaxy.py
mcoughlin: So we could enforce that all distmpc, distmpc_unc, redshift, and redshift_error are all > 0. I guess the minimum viable product allowed should be ra, dec, and name.

mcoughlin: Ok I added some checking along these lines.

skyportal/handlers/api/galaxy.py
stefanv: Yes, unfortunately just making it async won't help; the inside needs to be made async too, and currently nothing inside it exposes async functionality.

mcoughlin: Ok I moved the galaxy write to be async.

mcoughlin:
Failure unrelated.

skyportal/handlers/api/galaxy.py
```diff
@@ -0,0 +1,180 @@
...
+            'distmpc_unc',
+            'redshift',
+            'redshift_error',
+        ]
+        for key in positive_definite_parameters:
+            if any([(x is not None) and (x < 0) for x in catalog_data[key]]):
+                return self.error(f"{key} should be positive definite.")
+
+        # check RA bounds
+        if any([(x < 0) or (x > 360) for x in catalog_data['ra']]):
```
profjsb, line 85:
```suggestion
        if any([(x < 0) or (x >= 360) for x in catalog_data['ra']]):
```

skyportal/handlers/api/galaxy.py
```diff
@@ -0,0 +1,180 @@
...
+            return self.error("ra should span 0<ra<360.")
```
profjsb, line 86:
```suggestion
            return self.error("ra should span 0=<ra<360.")
```

profjsb (changes requested):

mcoughlin:
Great catch. Thanks @profjsb!

profjsb:
I am happy to approve once the table name issue is sorted.

alembic/versions/1aeb080b0eda_galaxy_table.py
acrellin: It's a bit awkward either way -- we can force it to be `"galaxies"`, but then any associated join tables will have `"galaxie_id"` because of how the singular-plural-singular conversion works.  E.g. we forced the photometry table to be `"photometry"`, but now all join tables references that table use `"photometr_id"` which is pretty awkward.  I think `"galaxys"` is fine campared to `"galaxie_id"`. Thoughts?

profjsb (approved):

PR #2393, opened 2021-10-27
Spectrum Annotations front end

guynir42:
Adding the annotations table from the source page / scanning page, but for each individual spectrum in the ManageData page. 
Screenshot from 2021-10-27 14-42-34


acrellin:
> I have no idea how to make the three components in the pull out row behave (i.e., sit in one line). Any help would be appreciated! @acrellin

@guynir42 What's the issue exactly? Happy to try to help, just don't know what you want to do here.

static/js/components/SourceDesktop.jsx
```diff
@@ -484,7 +484,10 @@ const SourceDesktop = ({ source }) => {
               </Typography>
             </AccordionSummary>
             <AccordionDetails>
-              <ObjPageAnnotations annotations={source.annotations} />
+              <ObjPageAnnotations
+                annotations={source.annotations}
+                spectra={spectra}
```
acrellin, line 489: Why not just pass in spectrum annotations after they've been processed rather than the entire spectra state?

skyportal/handlers/api/spectrum.py
```diff
@@ -635,6 +653,10 @@ def get(self, obj_id):
                 key=lambda x: x["created_at"],
                 reverse=True,
             )
+            annotations = [
+                {**a.to_dict(), 'author': a.author.to_dict()} for a in annotations
```
acrellin, line 657: Why is this necessary?

skyportal/handlers/api/spectrum.py
```diff
@@ -195,6 +196,14 @@ def get(self, spectrum_id):
             .filter(CommentOnSpectrum.spectrum_id == spectrum_id)
             .all()
         )
+        annotations = (
+            AnnotationOnSpectrum.query_records_accessible_by(
+                self.current_user,
+                options=[joinedload(AnnotationOnSpectrum.groups)],
```
acrellin, line 202: Does the front-end even utilize the groups list for each annotation? If not, let's omit this

skyportal/handlers/api/spectrum.py
```diff
@@ -616,6 +626,14 @@ def get(self, obj_id):
                 .filter(CommentOnSpectrum.spectrum_id == spec.id)
                 .all()
             )
+            annotations = (
+                AnnotationOnSpectrum.query_records_accessible_by(
+                    self.current_user,
+                    options=[joinedload(AnnotationOnSpectrum.groups)],
```
acrellin, line 632: Same question here

guynir42:
> > I have no idea how to make the three components in the pull out row behave (i.e., sit in one line). Any help would be appreciated! @acrellin
> 
> @guynir42 What's the issue exactly? Happy to try to help, just don't know what you want to do here.

If you look at the example image above (or if you check out this branch) you'll see that the comments component is stretched quite a bit horizontally, and then the annotations component is dropped to the second row. I think it would look better if I could get them more compactly into one row (at least in cases where the browser width is large enough). 
I guess if I could coerce each component to have a maximum size that would do it (maybe?) but I'm not sure where to control that.

skyportal/plot.py
```diff
@@ -1380,7 +1381,9 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
 
     obj = Obj.get_if_accessible_by(obj_id, user)
     spectra = (
-        Spectrum.query_records_accessible_by(user)
+        Spectrum.query_records_accessible_by(
+            user, options=[joinedload(Spectrum.annotations)]
```
acrellin, line 1385: This is not right -- you'll need to use a subquery here, becase annotations on spectra can have different permissioning than their associated spectrum (access to annotations on spectra is determined by group membership, not only whether they can read the underlying spectrum).

skyportal/plot.py
```diff
@@ -1409,6 +1412,12 @@ def spectroscopy_plot(obj_id, user, spec_id=None, width=600, device="browser"):
         normfac = np.nanmedian(np.abs(s.fluxes))
         normfac = normfac if normfac != 0.0 else 1e-20
         altdata = json.dumps(s.altdata) if s.altdata is not None else ""
+        annotations = (
+            json.dumps([{a.origin: a.data} for a in s.annotations])
```
acrellin, line 1416: Why is this necessary?

static/js/components/ManageDataForm.jsx
```diff
@@ -136,13 +137,21 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
-const SpectrumRow = ({ rowData, route }) => {
+const SpectrumRow = ({ rowData, route, annotations }) => {
   const styles = useSourceStyles();
   const colSpan = rowData.length + 1;
+  const spectrumID =
+    typeof rowData[0] === "number" ? rowData[0] : parseInt(rowData[0], 10);
```
acrellin, line 144:
```suggestion
  const spectrumID = parseInt(rowData[0], 10);
```
no harm in passing in a value that's already an int; this is much easier to read/maintain.

static/js/components/ManageDataForm.jsx
```diff
@@ -161,13 +170,13 @@ const SpectrumRow = ({ rowData, route }) => {
...
-                typeof rowData[0] === "number"
-                  ? rowData[0]
-                  : parseInt(rowData[0], 10)
-              }
+              spectrumID={spectrumID}
             />
           </Grid>
+          <Grid item sm={6}>
+            <Typography variant="h6">Annotations</Typography>
+            <ObjPageAnnotations annotations={annotations} />
```
acrellin, line 178: Now that this component is being used in places besides the Obj Page, it makes sense to rename it to `ObjAnnotations`

static/js/components/ObjPageAnnotations.jsx
```diff
@@ -69,11 +69,21 @@ const getMuiTheme = (theme) =>
   });
 
 // Table for displaying Obj annotations on Candidate page and Source page
-const ObjPageAnnotations = ({ annotations }) => {
+const ObjPageAnnotations = ({ annotations, spectra = [] }) => {
```
acrellin, line 72: See comment about passing in spectrum annotations, not the entire spectra list

static/js/components/ObjPageAnnotations.jsx
```diff
@@ -69,11 +69,21 @@ const getMuiTheme = (theme) =>
...
   const classes = useStyles();
   const theme = useTheme();
   const renderValue = (value) => getAnnotationValueString(value);
   const renderTime = (created_at) => dayjs().to(dayjs.utc(`${created_at}Z`));
+  const renderSpectrumDate = (observed_at) => {
+    if (observed_at) {
+      const dayFraction = (parseFloat(observed_at.substring(11, 13)) / 24) * 10;
+      return [observed_at.substring(0, 10), ".", dayFraction.toFixed(0)].join(
+        ""
+      );
```
acrellin, line 82:
```suggestion
      return observed_at.substring(0, 10) + "." + dayFraction.toFixed(0);
```

acrellin: Simpler and easier to maintain

acrellin: or perhaps preferably:
``` javascript
return `${observed_at.substring(0, 10)}.${dayFraction.toFixed(0)}`;
```

static/js/components/ObjPageAnnotations.jsx
```diff
@@ -112,6 +122,38 @@ const ObjPageAnnotations = ({ annotations }) => {
     },
   ];
 
+  // if there are any spectra given, we should add the spectrum annotations after object annotations
+  if (spectra.length > 0) {
```
acrellin, line 126: Is it possible for `spectra` to ever be passed in as `null` or `undefined` while the data are still being fetched? If so we need to be able to handle that here gracefully (this will error as-is if so).

static/js/components/ObjPageAnnotations.jsx
```diff
@@ -69,11 +69,21 @@ const getMuiTheme = (theme) =>
   });
 
 // Table for displaying Obj annotations on Candidate page and Source page
```
acrellin, line 71:
```suggestion
// Table for displaying Obj annotations
```

static/js/components/ObjPageAnnotations.jsx
```diff
@@ -112,6 +122,38 @@ const ObjPageAnnotations = ({ annotations }) => {
     },
   ];
 
+  // if there are any spectra given, we should add the spectrum annotations after object annotations
+  if (spectra.length > 0) {
+    // first add the observed_at property to each annotation in tableData:
+    tableData.forEach((ann) => {
```
acrellin, line 128: Let's do this above where the regular obj annotations are added with a conditional check.

acrellin (changes requested):
Looking good so far! In addition to the below comments, this needs tests.

static/js/components/SourceDesktop.jsx
guynir42: The annotations require additional info like the time each spectrum was detected. That means you have to start passing around data from spectra to annotations before giving it to the component. 
It seemed cleaner to just give the component this one optional argument.

skyportal/handlers/api/spectrum.py
guynir42: I think without this the annotations do not come with an author (which is displayed in the table).

skyportal/handlers/api/spectrum.py
guynir42: Maybe you are right...

skyportal/plot.py
guynir42: OK so query spectrum annotations (accessible by) and then filter on spectrum ID. Got it.

skyportal/plot.py
guynir42: Actually I'm not sure how the data would be rendered without this. I can give it a try.

static/js/components/ManageDataForm.jsx
guynir42: I think it should be named `AnnotationsTable` since it also appears on the ManageData page.

static/js/components/ObjPageAnnotations.jsx
guynir42: I will go with option 2 because option 1 is what I did but the Lint didn't like that.

static/js/components/ObjPageAnnotations.jsx
guynir42: Not sure what you mean?

skyportal/handlers/api/spectrum.py
guynir42: Nevermind it works without this. I don't know why but for comments this was needed (maybe there it required the gravator or something).

skyportal/plot.py
guynir42: I think the json dump is needed to turn the objects into strings that bokeh can render.

skyportal/handlers/api/spectrum.py
guynir42: I don't know why we need this here and not in the object annotations. But it doesn't render an author if I don't add this...

profjsb:
Apart from tests and the remaining enquires from @acrellin, this is looking good.

static/js/components/SourceDesktop.jsx
acrellin: Well, for consistency, I think we should do the pre-processing before we pass the data in here so that we have consistent props like `annotations={...} specAnnotations = {...}`

skyportal/handlers/api/spectrum.py
acrellin: Okay, this is okay to leave as is then. Thanks

skyportal/plot.py
acrellin: Okay

static/js/components/ObjPageAnnotations.jsx
acrellin: Still wondering this

static/js/components/ObjPageAnnotations.jsx
acrellin: Above, there's an iteration over each annotation, where we should add this field if relevant there so we're not iterating again unnecessarily

static/js/components/ObjPageAnnotations.jsx
guynir42: Yeah sorry I think we can just add a couple more conditions to that if, that's not a problem.

acrellin: like `if (spectra && spectra.length)`

static/js/components/AnnotationsTable.jsx
```diff
@@ -84,10 +84,13 @@ const AnnotationsTable = ({ annotations, spectra = [] }) => {
 
   // Curate data
   const tableData = [];
+  annotations.push(...spectrumAnnotations);
```
acrellin, line 87: If `annotations` could ever possibly be `null`, we'd need to add `annotations?.push(...`

static/js/components/AnnotationsTable.jsx
```diff
@@ -120,30 +123,7 @@ const AnnotationsTable = ({ annotations, spectra = [] }) => {
...
-            value,
-            author,
-            created_at,
-            observed_at: spec.observed_at,
-          });
-        });
-      });
-    });
-
+  if (spectrumAnnotations && spectrumAnnotations.length) {
```
acrellin, line 126: `  if (spectrumAnnotations?.length) {`

static/js/components/AnnotationsTable.jsx
```diff
@@ -84,10 +84,13 @@ const AnnotationsTable = ({ annotations, spectra = [] }) => {
 
   // Curate data
   const tableData = [];
+  annotations.push(...spectrumAnnotations);
   annotations?.forEach((annotation) => {
-    const { origin, data, author, created_at } = annotation;
+    const { origin, data, author, created_at, spectrum_observed_at } =
+      annotation;
+    const observed_at = spectrum_observed_at || null;
```
acrellin, line 91:
``` javascript
    const { origin, data, author, created_at, spectrum_observed_at: observed_at = null } =
      annotation;
```

static/js/components/AnnotationsTable.jsx
```diff
@@ -68,18 +68,29 @@ const getMuiTheme = (theme) =>
...
+  };
+
   // Curate data
   const tableData = [];
+  annotations.push(...spectrumAnnotations);
   annotations?.forEach((annotation) => {
-    const { origin, data, author, created_at } = annotation;
+    const { origin, data, author, created_at, spectrum_observed_at } =
+      annotation;
+    const observed_at = spectrum_observed_at || null;
```
acrellin, line 91:
```suggestion
    const { origin, data, author, created_at, spectrum_observed_at: observed_at = null } =
      annotation;
```

acrellin:
@guynir42 Looking much better, thanks! A few more suggestions above. And can you also add some spectrum annotations to the demo data yaml so that I can check out this branch and easily see it in action?

guynir42:
> @guynir42 Looking much better, thanks! A few more suggestions above. And can you also add some spectrum annotations to the demo data yaml so that I can check out this branch and easily see it in action?

Good idea!

acrellin (changes requested):
Looks good except that 14gqr and 16fil are in the process of being replaced; can you please update the demo data yaml to post to other sources? And add regular annotations there if necessary so we can see them alongside each other on the source page.

guynir42:
> Looks good except that 14gqr and 16fil are in the process of being replaced; can you please update the demo data yaml to post to other sources? And add regular annotations there if necessary so we can see them alongside each other on the source page.

Done @acrellin

acrellin (approved):
Looks good, thanks @guynir42!

profjsb:
@guynir42 can you resolve the conflicts with the demo data? After that, I can approve and merge.

profjsb (approved):

PR #2419, opened 2021-11-05
Filter on MJD for last detected

guynir42:
Right now when getting sources using `startDate` or `endDate` the query is built by loading all photometry rows associated with all sources and checking each one for whether it was detected or not. This can be very slow if you have many sources. 

Instead, I propose we do a sub-query on all photometry points with MJD after the `startDate` (or before the `endDate`) and filter out those that were detected, and then join with the sources query. 

benchmarks
----------------
I uploaded 3000 sources with a few hundred photometry points each. 
- If I don't filter at all on `startDate` I get all sources and it takes about 13s (the time it takes to load and send the sources, I guess). 
- If I put a `startDate` very far in the past that gets me all sources, it takes 17s. 
- If I do that with the new method it takes 19s. 
- If I put the `startDate` a little more in the future and grab about 1/3 of the sources, in the old method it takes 8s
- If I do that with the new method it takes 5s. 
- If I put the `startDate` today, and get no sources back at all, the old method still takes 4s. 
- The same thing with the new method takes ~0.1s. 

I think this could be very useful, as long as people set reasonable `startDate` or `endDate` on their queries. On the other hand, if people put other limits like RA/Dec that reduces the number of sources, the old method could be much faster, since it doesn't require polling the entire Photometry table. 

It may be worth while to merge this into staging and run some test, before working this into a full PR.


skyportal/handlers/api/source.py
```diff
@@ -882,9 +885,24 @@ class Validator(Schema):
...
+                    Obj.last_detected_at(self.current_user) >= start_date
+                )
+            elif query_mode == 2:
+                utc_timestamp = start_date.timestamp()
+                mjd = utc_timestamp / 86400 + 40587  # from 1970 Jan 1
+                start_date_subquery = (
+                    Photometry.query_records_accessible_by(self.current_user)
+                    .filter(Photometry.mjd > mjd)
+                    .filter(
+                        Photometry.flux / Photometry.fluxerr > PHOT_DETECTION_THRESHOLD
```
acrellin, line 899: `Photometry` already has `snr` which you can use instead of calculating this here explicitly.
```suggestion
                        Photometry.snr > PHOT_DETECTION_THRESHOLD
```

skyportal/handlers/api/source.py
```diff
@@ -882,9 +885,24 @@ class Validator(Schema):
...
-                Obj.last_detected_at(self.current_user) >= start_date
-            )
+            if query_mode == 1:
+                obj_query = obj_query.filter(
+                    Obj.last_detected_at(self.current_user) >= start_date
+                )
+            elif query_mode == 2:
+                utc_timestamp = start_date.timestamp()
+                mjd = utc_timestamp / 86400 + 40587  # from 1970 Jan 1
+                start_date_subquery = (
```
acrellin, line 895:
```suggestion
                last_detected_after_subquery = (
```

skyportal/handlers/api/source.py
```diff
@@ -882,9 +885,24 @@ class Validator(Schema):
...
+                    Photometry.query_records_accessible_by(self.current_user)
+                    .filter(Photometry.mjd > mjd)
+                    .filter(
+                        Photometry.flux / Photometry.fluxerr > PHOT_DETECTION_THRESHOLD
+                    )
+                    .subquery()
+                )
+                obj_query = obj_query.join(
+                    start_date_subquery, Obj.id == start_date_subquery.c.obj_id
```
acrellin, line 904:
```suggestion
                    last_detected_after_subquery, Obj.id == last_detected_after_subquery.c.obj_id
```

acrellin (commented):
On its own, this provides no improvement in worst-case time complexity. We only start to see benefits when `startDate` and/or `endDate` << median of detection times of accessible photometry. Perhaps we should limit `startDate` and `endDate` to some not-too-distant maximum time from now.

guynir42:
@acrellin I agree it doesn't help the worst case scenario. But I think that if a user insists on making a bad query, the kind that pulls up __all__ the objects in the DB, that would be bad no matter how we set things up. 
We need to protect against very long queries and I'm not sure how to do that. 
On the other hand, for most practical queries, you usually request a start/end time which is close to now... and in that case things will be a lot faster using this method. 
There may be other things we can do, like add a global (all users) cache that lets us quickly sort out objects that are "stale" for all users and get them out from the query before we even start. But again, that doesn't do much if the user request a start date on 1970...

skyportal/handlers/api/source.py
```diff
@@ -882,9 +885,24 @@ class Validator(Schema):
...
-            obj_query = obj_query.filter(
-                Obj.last_detected_at(self.current_user) >= start_date
-            )
+            if query_mode == 1:
+                obj_query = obj_query.filter(
+                    Obj.last_detected_at(self.current_user) >= start_date
+                )
+            elif query_mode == 2:
+                utc_timestamp = start_date.timestamp()
+                mjd = utc_timestamp / 86400 + 40587  # from 1970 Jan 1
```
profjsb, line 894: We use `astropy.time` for mjd calculations (helps avoid off by 0.5 day errors). For example:
```
mjd = Time(request.payload["start_date"], format='iso').mjd
```

skyportal/handlers/api/source.py
```diff
@@ -976,17 +981,69 @@ class Validator(Schema):
...
+                    Photometry.query_records_accessible_by(self.current_user)
+                    .filter(Obj.id == Photometry.obj_id)
+                    .filter(Photometry.mjd >= mjd_end)
+                    .filter(Photometry.snr > PHOT_DETECTION_THRESHOLD)
+                    .scalar_subquery()
+                )
+
+                obj_query = obj_query.outerjoin(
+                    after_range_subquery, Obj.id == after_range_subquery.c.obj_id
+                ).filter(after_range_subquery.c.id.is_(None))
```
guynir42, line 1036: @mcoughlin @stefanv Maybe one of you knows how to do a proper `scalar_subquery` off of an existing query? 
I'm trying to get all objects that don't have even one photometry points after end_date.

skyportal/handlers/api/source.py
```diff
@@ -976,17 +981,69 @@ class Validator(Schema):
...
+                    in_range_subquery, Obj.id == in_range_subquery.c.obj_id
+                )
+                # then must also disqualify objects that have detections after the time range
+                after_range_subquery = (
+                    Photometry.query_records_accessible_by(self.current_user)
+                    .filter(Obj.id == Photometry.obj_id)
+                    .filter(Photometry.mjd >= mjd_end)
+                    .filter(Photometry.snr > PHOT_DETECTION_THRESHOLD)
+                    .scalar_subquery()
+                )
```
guynir42, line 1032:
```suggestion
                after_range_subquery = (
                        DBSession().execute(
                                        Photometry.query_records_accessible_by(self.current_user)
                                        .filter(Obj.id == Photometry.obj_id)
                                        .filter(Photometry.mjd >= mjd_end)
                                        .filter(Photometry.snr > PHOT_DETECTION_THRESHOLD)
                                  )
                    .scalar_subquery()
                )
```

Issue #2422, opened 2021-11-12
Swift TOO API


mcoughlin:
We should probably support this as we do other telescopes.

Copying from the announcement: 

We formally announce the availability of the Swift TOO API. This API is designed to allow automated submission of Target of Opportunity (TOO) Requests for the Neil Gehrels Swift Observatory. Previously TOO requests could only be submitted through a web form. The TOO API is intended to facilitate automated submission of Swift TOOs, or to make simpler the submission of multiple targets.

The API can be accessed using a simple JWT format (http://jwt.io). However, we expect most people to access it using our reference implementation of the Swift TOO API written in Python (supporting Python 3.6 and above), the swift_too module, which is part of swifttools (“pip install swifttools”). If you already have swifttools installed, we recommend you upgrade to the latest v2.1 release (“pip install --upgrade swifttools”).

PR #2423, opened 2021-11-13
Swift TOO Form

mcoughlin:
A first pass at a Swift TOO Form. Not sure how best to test it... they do have an anonymous account:
https://www.swift.psu.edu/too_api/index.php?md=README.md
that let's you validate but not submit.

Closes https://github.com/skyportal/skyportal/issues/2422


profjsb (commented):
I'm confused by the UVOT-centric naming (`UVOTAPI`) but most of the data payload seems to be XRT focused. Worth making different requests for different Swift instruments?

As for testing, since the end point will be brittle, we should probably use a mock return using `vcr`: https://github.com/ktosiek/pytest-vcr

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,198 @@
...
+        too.exp_time_per_visit = request.payload["exposure_time"]
+        too.monitoring_freq = "%d days" % request.payload["monitoring_freq"]
+        too.num_of_visits = int(request.payload["exposure_counts"])
+
+        too.xrt_countrate = request.payload["xrt_countrate"]
+        too.exp_time_just = request.payload["exp_time_just"]
+        too.immediate_objective = request.payload["immediate_objective"]
+
+        too.urgency = int(request.payload["urgency"])
+        too.obs_type = request.payload["obs_type"]
```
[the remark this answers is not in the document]
mcoughlin: It is a dropdown menu that restricts to:
            "obs_type": {
                "type": "string",
                "enum": ["Spectroscopy", "Light Curve", "Position", "Timing"],
                "default": "Light Curve",
                "title": "Observation Type",
            },
currently.

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,198 @@
...
+
+        too.exp_time_per_visit = request.payload["exposure_time"]
+        too.monitoring_freq = "%d days" % request.payload["monitoring_freq"]
+        too.num_of_visits = int(request.payload["exposure_counts"])
+
+        too.xrt_countrate = request.payload["xrt_countrate"]
+        too.exp_time_just = request.payload["exp_time_just"]
+        too.immediate_objective = request.payload["immediate_objective"]
+
+        too.urgency = int(request.payload["urgency"])
```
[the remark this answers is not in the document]
mcoughlin: https://www.swift.psu.edu/too_api/index.php?md=TOO%20parameters.md
I think it is 1-4.

mcoughlin:
My understanding is people just request both instruments when they put in the form, maybe we call it UVOT_XRT or something given it is just this one pathway? I could use help from someone on the VCR stuff if it is decided that is important.

mcoughlin:
@profjsb the VCR wasn't too bad.

skyportal/facility_apis/swift.py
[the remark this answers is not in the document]
acrellin: Looks like we still need to validate the inputs here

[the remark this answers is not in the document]
mcoughlin: What would this look like? Don't we in essence validate at the step where we are giving input options in the dropdown?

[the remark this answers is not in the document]
profjsb: A key idea is that the frontend and the backend need to independently validate and verify inputs. Not doing so on the backend opens up the API to malformed inputs (ie. if someone decides to script again this endpoint). So even if the frontend is constraining the inputs we need to assume the backend has to do the same checks.

skyportal/facility_apis/swift.py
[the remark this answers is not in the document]
profjsb: See comment above--still need the backend to validate the inputs.

skyportal/facility_apis/swift.py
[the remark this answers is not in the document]
profjsb: See lines 88-134 in skyportal/handlers/api/internal/standards.py for range and type checking.

skyportal/facility_apis/swift.py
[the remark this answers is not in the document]
mcoughlin: I added some validation as requested.

mcoughlin:
@acrellin worth a look when you can.

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+
+        self.requestgroup = self._build_payload(request)
+
+    def _build_payload(self, request):
+        """Payload json for Swift UVOT TOO requests.
+
+        Parameters
+        ----------
+
```
acrellin, line 38:
```suggestion
```

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,208 @@
...
+class UVOTRequest:
+
+    """A JSON structure for Swift UVOT requests."""
+
+    def __init__(self, request):
+        """Initialize UVOT request.
+
+        Parameters
+        ----------
+
```
acrellin, line 25:
```suggestion
```

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,208 @@
...
+        """Payload json for Swift UVOT TOO requests.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        Returns
+        ----------
```
acrellin, line 43:
```suggestion
        Returns
        -------
```

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,208 @@
...
+            raise ValueError('Missing allocation information.')
+
+        swiftreq = UVOTRequest(request)
+
+        swiftreq.requestgroup.validate()
+
+        r = requests.post(
+            url=API_URL, verify=True, data={'jwt': swiftreq.requestgroup.jwt}
+        )
+        # r = swiftreq.requestgroup.submit_post()
```
acrellin, line 116:
```suggestion
```

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,208 @@
...
+            request.status = f'rejected: {r.content}'
+
+        transaction = FacilityTransaction(
+            request=http.serialize_requests_request(r.request),
+            response=http.serialize_requests_response(r),
+            followup_request=request,
+            initiator_id=request.last_modified_by_id,
+        )
+
+        DBSession().add(transaction)
```
acrellin, line 130: Need to commit afterwards?

skyportal/tests/frontend/sources_and_followup_etc/test_followup_requests.py
```diff
@@ -52,6 +52,16 @@
     ztf_isonline = True
 
 
+url = "https://www.swift.psu.edu/toop/submit_json.php"
+swift_isonline = False
+try:
+    requests.get(url, timeout=5)
```
acrellin, line 58:
```suggestion
swift_url = "https://www.swift.psu.edu/toop/submit_json.php"
swift_isonline = False
try:
    requests.get(swift_url, timeout=5)
```

acrellin (commented):
Looks good, thanks! A few questions below

data/tests/test_server_recordings_static.yaml
```diff
@@ -240,4 +240,48 @@ interactions:
...
...0IiwiYXBpX3ZlcnNpb24iOiIxLjIiLCJhcGlfZGF0YSI6eyJ1c2VybmFtZSI6ImFub255bW91cyIsInNvdXJjZV9uYW1lIjoiMTc1ZTZkMjNiMWEzNDM4YTljYjA1YTg2YzBmYzM2NTQiLCJzb3VyY2VfdHlwZSI6Ik9wdGljYWwgZmFzdCB0cmFuc2llbnQiLCJyYSI6MC4wLCJkZWMiOjAuMCwicG9zZXJyIjowLjAsImluc3RydW1lbnQiOiJYUlQiLCJ1cmdlbmN5IjozLCJ4cnRfY291bnRyYXRlIjowLjAwMjUsImltbWVkaWF0ZV9vYmplY3RpdmUiOiJXZSB3aXNoIHRvIG1lYXN1cmUgdGhlIFgtcmF5IGVtaXNzaW9uIG9mIGFuIG9wdGljYWxseSBkaXNjb3ZlcmVkIHBvdGVudGlhbCBvcnBoYW4gYWZ0ZXJnbG93L2tpbG9ub3ZhLiIsInNjaWVuY2VfanVzdCI6IkFuIFgtcmF5IGRldGVjdGlvbiBvZiB0aGlzIHRyYW5zaWVudCB3aWxsIGZ1cnRoZXIgYXNzb2NpYXRlIHRoaXMgb2JqZWN0IHRvIGEgcmVsYXRpdmlzdGljIGV4cGxvc2lvbiBhbmQgd2lsbCBoZWxwIHVudmVpbCB0aGUgbmF0dXJlIG9mIHRoZSBwcm9nZW5pdG9yIHR5cGUuIiwiZXhwb3N1cmUiOjQwMDAsImV4cF90aW1lX2p1c3QiOiJBdCB-Mi41ZS0zIGNvdW50cy9zZWMsIDRrcyBzaG91bGQgc3VmZmljZSB0byBhY2hpZXZlIGEgaGlnaCBTTlIsIGFzc3VtaW5nIGEgYmFja2dyb3VuZCBvZiB-MWUtNCBjb3VudHMvc2VjIChQYWdhbmkgZXQgYWwuIDIwMDcpIiwiZXhwX3RpbWVfcGVyX3Zpc2l0Ijo0MDAwLCJudW1fb2ZfdmlzaXRzIjoxLCJtb25pdG9yaW5nX2ZyZXEiOiIxIGRheXMiLCJ4cnRfbW9kZSI6IlBDIiwidXZvdF9tb2RlIjoiZGVmYXVsdCIsIm9ic19uIjoic2luZ2xlIiwib2JzX3R5cGUiOiJMaWdodCBDdXJ2ZSIsInZhbGlkYXRlX29ubHkiOmZhbHNlfX0.PrzYHA42kdg5apdpJD_7xRn7RsW0eGHGj3TDTw0jN5c
```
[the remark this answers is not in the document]
mcoughlin: It is just the anonymous account they suggest for testing.

skyportal/facility_apis/swift.py
mcoughlin: This is the same as in the others... it clearly appears on the page, so I presume it works. But I cannot tell you why. This was a Danny invention...

acrellin: Okay, we must be calling `commit` after this method is invoked.

acrellin (approved):

profjsb (approved):
Looks great. Will merge after tests pass.

profjsb:
@mcoughlin would you mind merging current master into this branch (I dont have permission)? I'll rerun tests then.

mcoughlin:
Rebased

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,205 @@
...
+        """Payload json for Swift UVOT TOO requests.
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        Returns
+        -------
+        payload: json
```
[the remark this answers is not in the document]
stefanv: The return type here doesn't seem to be JSON?

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,205 @@
...
+            payload for requests.
+        """
+
+        altdata = request.allocation.altdata
+
+        too = Swift_TOO()
+        too.username = altdata["username"]
+        too.shared_secret = altdata["secret"]
+        # too.debug = True
```
stefanv, line 51: Remove?

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,205 @@
...
+
+    # subclasses *must* implement the method below
+    @staticmethod
+    def submit(request):
+
+        """Submit a follow-up request to Swift's UVOT
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
```
stefanv, line 97:
```suggestion
        request : skyportal.models.FollowupRequest
```

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,205 @@
...
+
+        altdata = request.allocation.altdata
+        if not altdata:
+            raise ValueError('Missing allocation information.')
+
+        swiftreq = UVOTRequest(request)
+
+        swiftreq.requestgroup.validate()
+
+        r = requests.post(
```
stefanv, line 111: Handle exceptions gracefully. https://docs.python-requests.org/en/latest/user/quickstart/#errors-and-exceptions

skyportal/services/test_server/test_api_server.py
```diff
@@ -390,9 +414,12 @@ def get(self):
                 self.write("Could not find test route redirect")
 
     def post(self):
+
         is_soap_action = "Soapaction" in self.request.headers
         if "/api/requestgroups/" in self.request.uri:
             cache = get_cache_file_static()
+        elif "/toop/submit_json.php" in self.request.uri:
```
stefanv, line 421: Perhaps something like:

```
cached_urls = [".*/api/requestgroups/.*", ".*/toop/submit_json.php$"]

...

if any(re.match(pat, self.request.uri) for pat in cached_urls):
    cache = get_cache_file_static()
```

skyportal/tests/frontend/sources_and_followup_etc/test_followup_requests.py
```diff
@@ -52,6 +52,16 @@
     ztf_isonline = True
 
 
+swift_url = "https://www.swift.psu.edu/toop/submit_json.php"
```
stefanv, line 55: Do we need to test against swift if we have the pre-recorded transactions?

skyportal/utils/http.py
```diff
@@ -1,6 +1,9 @@
 def serialize_requests_request(request):
     if request.body is not None:
-        body = request.body.decode()
+        if isinstance(request.body, bytes):
```
stefanv, line 3: What's happening here?

stefanv (commented):
Nice work, @mcoughlin!

Some minor suggestions, hopefully all very quick.

skyportal/utils/http.py
mcoughlin: just checking if needs to be decoded or not.

skyportal/tests/frontend/sources_and_followup_etc/test_followup_requests.py
mcoughlin: We do the others...

skyportal/services/test_server/test_api_server.py
mcoughlin: added

skyportal/facility_apis/swift.py
mcoughlin: added

skyportal/facility_apis/swift.py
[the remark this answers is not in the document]
mcoughlin: fixed

config.yaml.defaults
```diff
@@ -33,6 +33,10 @@ app:
     host: skipper.caltech.edu
     port: 4000
 
+  swift_protocol: https
```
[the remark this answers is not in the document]
mcoughlin: fixed

data/tests/test_server_recordings_static.yaml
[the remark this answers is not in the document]
mcoughlin: Kyung just wrote a note at the top of the script to delete the file and rerun the tests.

skyportal/utils/http.py
stefanv: Yes, but why would it ever arrive as anything other than bytes?

mcoughlin: Ok will remove it

stefanv: I presume it was added for a reason; just trying to understand why!

mcoughlin: Presumably some bug I introduced in initial testing.

mcoughlin: Well that broke the tests, so something about it is important I guess

stefanv: Hrm, it is used by sedm, lt, ztf, and lco without this switch.  Why would swift be different?

stefanv: Looks like the failures are unrelated?

mcoughlin: I am not sure why. something about how the Swift folks are returning requests?

mcoughlin:
@stefanv can we get this merged?

profjsb:
@mcoughlin please rebase off of current master so we can get CI tests passing.

mcoughlin:
@profjsb I think it is all rebased, according to github. Do you see something else?

profjsb:
Since we're updating `facilities_apis` ENUM with UVOTAPI, you will need a migration script (see https://github.com/profjsb/skyportal/blob/90eccb64de9c68621f2ed25f61c9a3e22abca7d9/alembic/versions/f7c1e785d7a4_unique_token_name_by_user.py). Once you pull in the latest on master and run the tests you'll get migration.py and migra.sql artifact from the (initially failing) migration script. You'll combine those two files as above. Once you include that script you should get tests to pass

mcoughlin:
@profjsb Doing this by hand I get an empty migration script (and github doesn't give me a migration).

requirements.txt
```diff
@@ -51,3 +51,4 @@ pygcn==1.1.2
 xmlschema==1.9.0
 requests==2.27
 typing-extensions>=4.0.1  # astropy v5.0 needs this, but does not correctly specify as much in its install config
+swifttools==2.1
```
profjsb, line 54:
```suggestion
swifttools==2.2.1
```

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,205 @@
...
+        too.exp_time_per_visit = request.payload["exposure_time"]
+        too.monitoring_freq = "%d days" % request.payload["monitoring_freq"]
+        too.num_of_visits = int(request.payload["exposure_counts"])
+
+        too.xrt_countrate = request.payload["xrt_countrate"]
+        too.exp_time_just = request.payload["exp_time_just"]
+        too.immediate_objective = request.payload["immediate_objective"]
+
+        if request.payload["urgency"] not in ["1", "2", "3", "4"]:
+            raise ValueError('urgency outside of allowed bounds.')
```
profjsb, line 66:
```suggestion
            raise ValueError('urgency not in ["1", "2", "3", "4"].')
```

skyportal/facility_apis/swift.py
```diff
@@ -0,0 +1,205 @@
...
+        swiftreq = UVOTRequest(request)
+
+        swiftreq.requestgroup.validate()
+
+        r = requests.post(
+            url=API_URL, verify=True, data={'jwt': swiftreq.requestgroup.jwt}
+        )
+        r.raise_for_status()
+
+        if r.status_code == 201:
```
profjsb, line 115: not 200?

skyportal/facility_apis/__init__.py
```diff
@@ -4,4 +4,5 @@
 from .sedmv2 import SEDMV2API
 from .lt import IOOAPI, IOIAPI, SPRATAPI
 from .lco import SINISTROAPI, SPECTRALAPI, FLOYDSAPI, MUSCATAPI
+from .swift import UVOTAPI
```
profjsb, line 7: Suggest changing this to SWIFTAPI or UVOT_XRTAPI to get the naming consistent with the instruments that will be requested.

profjsb (changes requested):
See https://github.com/skyportal/skyportal/commit/dba29ba7ecf836f8c1ce0aa2157764c450444ab0# to see the upgrade/downgrade path. You should see the commands in the migral.sql script which is an artifact of the migration GHA.

profjsb:
@stefanv can you finalize your review here? I think that the reason that the migration GHA is not getting triggered is because we dont check the change of `skyportal/facility_apis/__init__.py`. That said, I think that the refresh enums code will take care of the changing enums, right? If not, then we need to trigger a migration on enum change.

skyportal/facility_apis/swift.py
mcoughlin: this is  what they send back...

skyportal/facility_apis/__init__.py
mcoughlin: I changed it to UVOTXRTAPI.

profjsb:
@mcoughlin after you fix the migration to the latest revision we can merge.

mcoughlin:
@profjsb it should be latest now. I updated this morning.

profjsb (approved):

PR #2424, opened 2021-11-15
Cap number of sources returned per query

acrellin:


stefanv (approved):
Looks like we already had a `min(..., 100)` in there; why wasn't that doing the trick?

Either way, this PR looks good.  Just double check that pages start at 1, then go ahead and merge on green.

acrellin:
@stefanv because `pageNumber` was defaulting to `None`, in which case we were ignoring `numPerPage`

acrellin:
Yep, pages are 1-indexed

profjsb (approved):

PR #2426, opened 2021-11-17
Allow filtering sources by created_at or modified

acrellin:
This patch adds a new request parameter to the sources
API that allows for filtering on Obj.{created_at, modified}.

Closes https://github.com/skyportal/skyportal/issues/2425


acrellin:
@stefanv Mind taking a quick look here when you have a minute? This is a small one

skyportal/tests/api/test_sources.py
```diff
@@ -1556,3 +1556,174 @@ def test_sources_hidden_photometry_not_leaked(
...
+    )
+    assert status == 200
+    assert len(data["data"]["sources"]) == 0
+
+
+def test_filter_sources_by_modified(upload_data_token, view_only_token, public_group):
+    obj_id1 = str(uuid.uuid4())
+    obj_id2 = str(uuid.uuid4())
+
+    time_before_both = datetime.now(timezone.utc)
```
guynir42, line 1645: Maybe it would be quicker and shorter if we just use the fixture sources and modify them?

acrellin: But I want to be sure we're filtering on created_at, too. Suggest we leave as-is.

guynir42 (approved):

skyportal/handlers/api/source.py
```diff
@@ -943,6 +954,16 @@ class Validator(Schema):
             source_query = source_query.filter(Source.saved_at <= saved_before)
         if saved_after:
             source_query = source_query.filter(Source.saved_at >= saved_after)
+        if created_or_modified_after:
+            created_or_modified_date = arrow.get(
```
profjsb, line 958: This will throw an uncaught exception if `created_or_modified_after` is not in a proper format. Suggest gracefully catching this exception and responding accordingly.

profjsb (changes requested):

skyportal/handlers/api/source.py
```diff
@@ -943,6 +954,16 @@ class Validator(Schema):
...
+                created_or_modified_after.strip()
+            ).datetime
```
acrellin, line 960:
```suggestion
            try:
                created_or_modified_date = arrow.get(
                    created_or_modified_after.strip()
                ).datetime
            except arrow.ParserError:
                return self.error("Invalid value provided for createdOrModifiedAfter")
```

skyportal/handlers/api/source.py
acrellin: Agreed -- done.

profjsb (approved):

Issue #2454, opened 2021-12-02
Decrease verbosity of instrument field generation


acrellin:
Currently the logs are overflowing with 
```
...
[16:39:04 api/instrument] Loaded field 3664 for instrument 13 
[16:39:04 api/instrument] Loaded field 3665 for instrument 13 
[16:39:04 api/instrument] Loaded field 3666 for instrument 13 
[16:39:04 api/instrument] Loaded field 3667 for instrument 13 
[16:39:04 api/instrument] Loaded field 3668 for instrument 13 
[16:39:04 api/instrument] Loaded field 3669 for instrument 13 
[16:39:04 api/instrument] Loaded field 3670 for instrument 13 
[16:39:04 api/instrument] Loaded field 3671 for instrument 13 
[16:39:04 api/instrument] Loaded field 3672 for instrument 13 
[16:39:04 api/instrument] Loaded field 3673 for instrument 13 
[16:39:04 api/instrument] Loaded field 3674 for instrument 13 
[16:39:04 api/instrument] Loaded field 3675 for instrument 13 
[16:39:04 api/instrument] Loaded field 3676 for instrument 13 
[16:39:04 api/instrument] Loaded field 23.0 for instrument 1  
[16:39:04 api/instrument] Loaded field 3677 for instrument 13 
[16:39:05 api/instrument] Loaded field 3678 for instrument 13 
[16:39:05 api/instrument] Loaded field 3679 for instrument 13 
[16:39:05 api/instrument] Loaded field 3680 for instrument 13 
[16:39:05 api/instrument] Loaded field 3681 for instrument 13 
[16:39:05 api/instrument] Loaded field 3682 for instrument 13 
[16:39:05 api/instrument] Loaded field 3683 for instrument 13 
[16:39:05 api/instrument] Loaded field 3684 for instrument 13 
[16:39:05 api/instrument] Loaded field 3685 for instrument 13 
[16:39:05 api/instrument] Loaded field 3686 for instrument 13 
[16:39:05 api/instrument] Loaded field 3687 for instrument 13 
[16:39:05 api/instrument] Loaded field 3688 for instrument 13 
[16:39:05 api/instrument] Loaded field 3689 for instrument 13 
[16:39:05 api/instrument] Loaded field 3690 for instrument 13 
[16:39:05 api/instrument] Loaded field 3691 for instrument 13 
[16:39:05 api/instrument] Loaded field 3692 for instrument 13 
[...]

PR #2456, opened 2021-12-02
Decrease verbosity of instrument field loading

acrellin:
Closes https://github.com/skyportal/skyportal/issues/2454


mcoughlin (approved):
Slightly sad, but probably for the best.

PR #2457, opened 2021-12-02
Upgrade to React v17

acrellin:


acrellin:
Failures unrelated

acrellin:
All failures unrelated.

@stefanv since the last time we tried this, other react-dependent packages have gone through sufficient updates such that this now works seamlessly without modification.

PR #2458, opened 2021-12-02
Upgrade webpack to v5

acrellin:
Upgrade webpack to v5.x and browserify non-polyfilled node modules for
v5 compatibility.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -103,7 +103,9 @@ const config = {
       ),
     },
     extensions: [".js", ".jsx", ".json"],
+    fallback: { path: path.resolve(__dirname, "node_modules/path-browserify") },
   },
+
   watchOptions: {
     ignored: /node_modules/,
     // Set to true if you have trouble with JS change monitoring
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,5 @@
 const path = require("path");
+const webpack = require("webpack");
 
 // const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -85,6 +86,9 @@ const config = {
   plugins: [
     // Uncomment the following line to enable bundle size analysis
     //    new BundleAnalyzerPlugin()
+    new webpack.ProvidePlugin({
```
stefanv, line 89: What is this plugin for?

webpack.config.js
```diff
@@ -103,7 +107,9 @@ const config = {
       ),
     },
     extensions: [".js", ".jsx", ".json"],
+    fallback: { path: path.resolve(__dirname, "node_modules/path-browserify") },
```
stefanv, line 110: This line looks fine.

webpack.config.js
acrellin: This is the only way I could expose the installed `process`; the below fallback approach didn't work here. This appeared to be the most recommended approach for `process` under v5

stefanv: Do you see a problem during runtime?  Because webpack compiles without it.

acrellin: @stefanv Yes, the page crashes with `process is undefined` without this.

stefanv: I cannot reproduce this on the main page.

acrellin: Meaning this branch runs fine locally without this?

stefanv: Yes, as far as I can tell.

acrellin: @stefanv Unless there's a good reason not to include this fix (fails for me locally without this), I propose we move forward here.

stefanv: Sure; how about adding a comment that this should be removed in the future, and then merge (no need to re-run the CI if it works for you locally).

stefanv: Also for the fallback below.

acrellin: Done, thanks!

stefanv:
Thanks, Ari!

Issue #2334, opened 2021-10-04
Consistently use Obj.internal_key for "refresh" WS message payloads


acrellin:
Currently some back-end handlers / front-end message handlers use `obj.internal_key` whilst others use `obj.id` as the WS message payload when sending a `REFRESH_{SOURCE,CANDIDATE,etc}` message. We should be using `internal_key` consistently. Update back-end handlers that send WS messages and associated front-end message handlers.

PR #2459, opened 2021-12-02
Use obj.internal_key in REFRESH_RESOURCE WS message payloads

bparazin:
Right now just making a PR to do test suite since it's much faster on github compared to my machine

Closes https://github.com/skyportal/skyportal/issues/2334 if merged

### DIFF
--- a/static/js/ducks/spectra.js
+++ b/static/js/ducks/spectra.js
@@ -43,7 +43,10 @@ messageHandler.add((actionType, payload, dispatch, getState) => {
     const state = getState().spectra;
 
     Object.entries(state).forEach(([objID, spectra]) => {
-      if (spectra[0]?.obj_internal_key === payload.obj_internal_key) {
+      if (
+        spectra &&
+        spectra[0]?.obj_internal_key === payload.obj_internal_key
+      ) {
         dispatch(fetchSourceSpectra(objID));
       }
     });
--- a/static/js/ducks/spectra.js
+++ b/static/js/ducks/spectra.js
@@ -43,10 +43,7 @@ messageHandler.add((actionType, payload, dispatch, getState) => {
     const state = getState().spectra;
 
     Object.entries(state).forEach(([objID, spectra]) => {
-      if (
-        spectra &&
-        spectra[0]?.obj_internal_key === payload.obj_internal_key
-      ) {
+      if (spectra?.[0]?.obj_internal_key === payload.obj_internal_key) {
         dispatch(fetchSourceSpectra(objID));
       }
     });
--- a/static/js/ducks/spectra.js
+++ b/static/js/ducks/spectra.js
@@ -42,9 +42,11 @@ messageHandler.add((actionType, payload, dispatch, getState) => {
   if (actionType === REFRESH_SOURCE_SPECTRA) {
     const state = getState().spectra;
 
-    if (Object.keys(state).includes(payload.obj_id)) {
-      dispatch(fetchSourceSpectra(payload.obj_id));
-    }
+    Object.entries(state).forEach(([objID, spectra]) => {
+      if (spectra[0]?.obj_internal_key === payload.obj_internal_key) {
+        dispatch(fetchSourceSpectra(objID));
+      }
+    });
   }
 });
 
--- a/skyportal/handlers/api/annotation.py
+++ b/skyportal/handlers/api/annotation.py
@@ -305,7 +305,7 @@ class AnnotationHandler(BaseHandler):
         if isinstance(annotation, AnnotationOnSpectrum):
             self.push_all(
                 action='skyportal/REFRESH_SOURCE_SPECTRA',
-                payload={'obj_id': annotation.obj_id},
+                payload={'obj_id': annotation.obj.internal_key},
             )
         return self.success(data={'annotation_id': annotation.id})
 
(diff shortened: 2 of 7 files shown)


bparazin:
Right now there is a obj.id in line 333 of comment.py that I can't change without breaking test_individual_spectra.py::test_comments and I don't know why.

bparazin:
But beyond that, all test failures are the starlist issues Joshua brought up in slack, so that's good at least

acrellin:
> Right now there is a obj.id in line 333 of comment.py that I can't change without breaking test_individual_spectra.py::test_comments and I don't know why.

@bparazin That's because in `static/js/ducks/spectra.js`, the message handler is expecting `id` -- you'll just need to change that to `internal_key`

bparazin:
How do I use an API.GET with the object_internal key? The issue is feeding back to an API GET call for the spectra that is failing, and I'm really unfamiliar with working with API stuff

acrellin:
> How do I use an API.GET with the object_internal key? The issue is feeding back to an API GET call for the spectra that is failing, and I'm really unfamiliar with working with API stuff

You have to use the actual ID for the GET request, that can't be done with the internal key. We just use the internal key to match against what's in the front-end state to see if anything needs to be updated.

bparazin:
Ok, this there some way to get object ID from internal key then? Because right now the object id feeds into this call, 
image
which calls this GET request that I need an object ID for
image
and if we want to use obj.internal_key, then I am thinking the switch should happen in the messageHandler addition I added in the first screenshot

bparazin:
Looking at things more, every place where obj.id is still used, rather than obj.internal_key is for a call to REFRESH_SOURCE_SPECTRA, which is because it feeds into this call
image
which in turn feeds into the GET call that you need the object ID for
image
Is there a way to get the object Id from the internal key? Because if not, I think that leaving these all as obj.id may have been an intentional design decision

acrellin:
@bparazin  In this case, we'd need to update the spectra handler to also return the associated object's `internal_key` so that we persist that in the `spectra` front-end state branch, then, in the message handler in the `spectra.js` duck, iterate over the objID - objData key-value pairs in that `spectra` state branch to see if one of the loaded spectra has a matching object `internal_key`, and if so, perform the corresponding fetch to update the state, and do nothing otherwise.

bparazin:
image
Is there somewhere else in the state that the internal key is stored? Because looking at the spectra branch, I don't see it there, and that would make check if one of the loaded spectra has that internal key a problem

acrellin:
@bparazin please read my comment again. We'll need to add that information - it doesn't currently exist in the front-end state.

bparazin:
Ah, my bad, I thought the spectra handler was referring to the REFRESH_SOURCE_SPECTRA function, think I get what you mean now

bparazin:
I am still really struggling here, is there some other file in the javascript land I have to edit too? The python stuff doesn't interface directly with the state variable, and seeing that might help figure out what change I have to make for this

acrellin:
@bparazin I meant something like the above commit (https://github.com/skyportal/skyportal/pull/2459/commits/addc093c727e97507e9656f73e926351d8f1c9c9)

bparazin:
All of these failures are passing on my machine, I cannot figure out why I'm just getting random time outs

bparazin:
Ok, the frontend part 1 failures are all marked as flaky in the code, and since I haven't touched anything they work with since the previous commit, I don't think I've got to worry about those. I am still rather confused about the frontend 2 on though

static/js/ducks/spectra.js
```diff
@@ -42,9 +42,14 @@ messageHandler.add((actionType, payload, dispatch, getState) => {
...
     const state = getState().spectra;
 
-    if (Object.keys(state).includes(payload.obj_id)) {
-      dispatch(fetchSourceSpectra(payload.obj_id));
-    }
+    Object.entries(state).forEach(([objID, spectra]) => {
+      if (
+        spectra &&
+        spectra[0]?.obj_internal_key === payload.obj_internal_key
+      ) {
```
acrellin, line 49:
```suggestion
      if (spectra?.[0]?.obj_internal_key === payload.obj_internal_key) {
```
you can just use optional chaining for that

acrellin:
Can you try rebasing on latest master? That might resolve the test failures.

bparazin:
Gonna wait for this test suite to finish, if I'm still seeing errors, will do

bparazin:
Alright, think I rebased it? First time trying that so there is a chance I messed it up, we'll see

bparazin:
Well at least it's a different error in upload spectroscopy now? This one is showing up on my end at least

acrellin:
That test appears to be failing on main as well, so I wouldn't worry about it. Looking good!

bparazin:
I think this might not be us, looking at the console command & error trace, it goes to something being undefined in data-uri-to-buffer, which is something that was imported & something we didn't touch at all?
image

bparazin:
Of course there is also the possibility that I somehow broke that when rebasing.

acrellin:
@bparazin the cause of that test failure may be addressed by https://github.com/skyportal/skyportal/pull/2471

bparazin:
Great, fingers crossed

acrellin:
@bparazin the fix for the failing test is actually in https://github.com/skyportal/skyportal/pull/2473

acrellin:
@bparazin is this PR ready to go? Are there any other instances of using obj_id in the websocket messages?

acrellin:
@bparazin it appears the rebase was botched :/  can you try cherry-picking those commits that are relevant from this PR branch onto a new master-based branch?

acrellin:
@bparazin I fixed the botched rebase

bparazin:
Thank you very much for all the help on this one

PR #2461, opened 2021-12-02
bump default taxonomy version and add info for updating taxonomy in SP

profjsb:
Provide info for how an admin updates a taxonomy in a live version.

### DIFF
--- a/doc/advanced_usage.md
+++ b/doc/advanced_usage.md
@@ -135,7 +135,7 @@ You should then see something like:
  'version': '0.9.dev0+git...'}
 ```
 
-If the `name` of the taxonomy is the exactly the same as a taxonomy already in the system, this new taxonomy will supercede that taxonomy. That is, by default it will be set to be the latest version (`IsLatest = True`) and shown to end users in the dropdown menu on the Source page. Older classifications will be still associated with the `taxonomy_id` they were originally connected to.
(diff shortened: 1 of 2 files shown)


acrellin (approved):
Looks good, thanks @profjsb

doc/advanced_usage.md
```diff
@@ -92,7 +92,52 @@ with the data that was posted.
...
+```
+You should then see something like:
+
+```
+{'status': 'success',
+ 'data': {'taxonomy_id': 9},
+ 'version': '0.9.dev0+git...'}
+```
+
+If the `name` of the taxonomy is the exactly the same as a taxonomy already in the system, this new taxonomy will supercede that taxonomy. That is, by default it will be set to be the latest version (`IsLatest = True`) and shown to end users in the dropdown menu on the Source page. Older classifications will be still associated with the `taxonomy_id` they were originally connected to.
```
acrellin, line 138:
```suggestion
If the `name` of the taxonomy is exactly the same as a taxonomy already in the system, this new taxonomy will supercede that taxonomy. That is, by default it will be set to be the latest version (`IsLatest = True`) and shown to end users in the dropdown menu on the Source page. Older classifications will be still associated with the `taxonomy_id` they were originally connected to.
```

PR #2463, opened 2021-12-03
Fix Gaia slowness leading to broken starlists and finders

profjsb:
It appears that we can no long issue ADQL queries which filter and sort on most columns. As such we need to do cone searches only and then filter on the Python side. This PR fixes the issue. There will be a performance hit here because we need to download all the sources which satisfies the cone search rather than the nearest ~60 as before.


skyportal/utils/offset.py
```diff
@@ -104,6 +104,7 @@ def query(self, q):
         if not self.is_backup:
             job = self.connection.launch_job(q)
             rez = job.get_results()
+
```
profjsb, line 107:
```suggestion
```

acrellin (approved):

acrellin:
Looks good, thanks @profjsb  --  let's see if the relevant tests pass on GA now!

stefanv (approved):

PR #2464, opened 2021-12-05
Autosuggest user mentions in comments

acrellin:
This PR implements auto-completion of usernames (or first name or 
last name) in the comment entry component utilizing a trie for 
performant matching.

username-autosuggest2

Closes https://github.com/fritz-marshal/fritz-beta-feedback/issues/111
Closes https://github.com/fritz-marshal/fritz-beta-feedback/issues/112


static/js/trie.js
```diff
@@ -0,0 +1,66 @@
+const Trie = () => {
```
profjsb, line 1: Should probably add a small comment here to describe what the trie tree is doing and what the data structure looks like after it's constructed (perhaps a reference to https://en.wikipedia.org/wiki/Trie).

profjsb (changes requested):
This looks great @acrellin. How difficult will it be to add the search on the person's name? I think most people do not remember the usernames of people they interact with but instead remember their actual name...it would be very useful for them to be able to autocomplete on real names too.

static/js/trie.js
acrellin: Done

acrellin:
@profjsb To allow simple matching on usernames in addition to first name, last name, I dropped the trie, as that would be a lot more complicated and am not sure we need it for ~few hundred users. We could certainly extend the trie implementation to cover both cases if we decide to go that route, though.

acrellin:
@profjsb actually, with ~500 usernames, performance is painfully laggy without utilizing a trie, so I'll work on extending that implementation to effectively handle both usernames and first/last name matches.

acrellin:
@profjsb @stefanv updated and ready for review.

profjsb (approved):

profjsb:
Looks great. Thanks, @acrellin!

static/js/components/CommentEntry.jsx
```diff
@@ -62,6 +81,28 @@ const CommentEntry = ({ addComment }) => {
...
     setGroupSelectVisible(false);
+    setTextValue("");
+    setAutosuggestVisible(false);
+    setUsernamePrefixMatches({});
+  };
+
+  const handleTextInputChange = (event) => {
+    const text = event.target.value;
+    const cursorIdx = event.target.selectionStart;
+    const currentWord = text.slice(0, cursorIdx).trim().split(" ").pop();
```
stefanv, line 92: Not 100% fool-proof (you can move the cursor, press Del to remove some chars, e.g.), but I think good enough for general editing.

static/js/components/CommentEntry.jsx
```diff
@@ -74,20 +115,71 @@ const CommentEntry = ({ addComment }) => {
     return formState.group_ids?.filter((value) => Boolean(value)).length >= 1;
   };
 
+  const handleClickSuggestedUsername = (username) => {
```
stefanv, line 118: We should also add a keyboard option for selecting the completion.

stefanv (commented):
This looks great, Ari.  Only kbd interaction missing, then I think we're good to go!

static/js/components/CommentEntry.jsx
acrellin: Right, I thought about that, but think it's fine if we only match from the "@" at the beginning of the word to where the cursor currently is -- users will quickly see that's what's being used for the matching in that case.

acrellin:
@stefanv Tab-navigating the list of matches and hitting return already works great!

static/js/components/CommentEntry.jsx
acrellin: See comment below -- this is already possible

stefanv:
Tab is a bit counter-intuitive; can you make it down-arrow like on GitHub and most other sites?

acrellin:
@stefanv That is far from trivial, and is something I see as more of a nice-to-have enhancement rather than strictly necessary for this feature. I've storied out adding that functionality (https://github.com/skyportal/skyportal/issues/2469) and suggest we push that to a later PR.

stefanv:
@acrellin I pushed up a change with key navigation functionality.  Because I removed the `<li>`, this may impact the tests. Hopefully not hard to fix, but if it is I'll look at it.

acrellin:
@stefanv that works, thanks. The styling is worse off now, though.  It's currently centered, which looks bad, and if we left-justify the button text, we end up with weird highlighting when selecting buttons with the arrow keys that doesn't match where the text is.

stefanv:
You can add back the li's, then you just have to use `.parentNode.nextSibling.firstChild` to focus on the next element.

acrellin:
@stefanv also, MDN suggests not using the now-deprecated `keyCode`.

stefanv:
OK, made some changes to address your comments.

stefanv:
One more fix incoming.

stefanv:
The fix I was trying to add is for, once you press enter, the focus returning to the text input field. I tried to add a ref to the input field and to then do `textEntryField.current.focus()` in `handleClickSuggestedUsername`, but that did not work.

acrellin:
@stefanv I got the re-focus on text input functionality working; please go ahead and merge when you're happy with it.

stefanv:
Darn, I was so close 😂

acrellin:
🎉  Thanks for the arrow-key additions @stefanv!

stefanv:
No problem. I thought the trie search addition was pretty neat!

PR #2468, opened 2021-12-07
SEDMv2 triggering interface

mcoughlin:
This PR enables the two pathways that would need to exist here: (i) a local SkyPortal instance that the request is being made on and (ii) an external SkyPortal instance that this instance would take to (like we would for any other system). So we support both pathways.

### DIFF
--- a/skyportal/facility_apis/sedmv2.py
+++ b/skyportal/facility_apis/sedmv2.py
@@ -1,6 +1,7 @@
 from . import FollowUpAPI
 from baselayer.app.env import load_env
 from datetime import datetime, timedelta
+import json
 import requests
 
 from ..utils import http
@@ -58,6 +59,9 @@ class SEDMV2API(FollowUpAPI):
         if cfg['app.sedmv2_endpoint'] is not None:
             altdata = request.allocation.altdata
 
+            if not altdata:
+                raise ValueError('Missing allocation information.')
+
             payload = {
                 'obj_id': request.obj_id,
                 'allocation_id': request.allocation.id,
@@ -103,12 +107,34 @@ class SEDMV2API(FollowUpAPI):
             The request to delete from the queue and the SkyPortal database.
         """
 
-        from ..models import FacilityTransaction, DBSession
+        from ..models import DBSession, FollowupRequest, FacilityTransaction
 
         validate_request_to_sedmv2(request)
 
         if cfg['app.sedmv2_endpoint'] is not None:
-            r = requests.delete(cfg['app.sedmv2_endpoint'], json=request.payload)
+            altdata = request.allocation.altdata
+
+            req = (
+                DBSession()
+                .query(FollowupRequest)
+                .filter(FollowupRequest.id == request.id)
+                .one()
+            )
+
+            altdata = request.allocation.altdata
+
+            if not altdata:
+                raise ValueError('Missing allocation information.')
+
+            content = req.transactions[0].response["content"]
+            content = json.loads(content)
+
+            uid = content["data"]["id"]
+
+            r = requests.delete(
+                f"{cfg['app.sedmv2_endpoint']}/{uid}",
+                headers={"Authorization": f"token {altdata['api_token']}"},
+            )
             r.raise_for_status()
             request.status = "deleted"
 
--- a/skyportal/facility_apis/sedmv2.py
+++ b/skyportal/facility_apis/sedmv2.py
@@ -51,7 +51,7 @@ def validate_request_to_sedmv2(request):
         request.payload["minimum_lunar_distance"] < 0
         or request.payload["minimum_lunar_distance"] > 180
     ):
-        raise ValueError('maximum_airmass must be within 0-180.')
+        raise ValueError('minimum lunar distance must be within 0-180.')
 
     if request.payload["priority"] < 1 or request.payload["priority"] > 5:
         raise ValueError('priority must be within 1-5.')
--- a/skyportal/facility_apis/sedmv2.py
+++ b/skyportal/facility_apis/sedmv2.py
@@ -38,7 +38,7 @@ def validate_request_to_sedmv2(request):
             raise ValueError(f'Filter configuration {filt} unknown.')
 
 
-class SEDMv2API(FollowUpAPI):
+class SEDMV2API(FollowUpAPI):
     """SkyPortal interface to the Spectral Energy Distribution machine (SEDMv2)."""
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


skyportal/facility_apis/sedmv2.py
```diff
@@ -0,0 +1,233 @@
...
+        "exposure_time",
+        "exposure_counts",
+        "maximum_airmass",
+        "minimum_lunar_distance",
+        "priority",
+        "start_date",
+        "end_date",
+    ]:
+        if param not in request.payload:
+            raise ValueError(f'Parameter {param} required.')
```
profjsb, line 35: Probably would benefit from some bounds checking here. I suppose it's the job of the receiver of this to deal bounds but if we can ensure that we are only sending over valid requests then it's worth doing the checking here too.

skyportal/facility_apis/sedmv2.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
+            The request to submit.
+        """
+
+        from ..models import FacilityTransaction, DBSession
+
+        validate_request_to_sedmv2(request)
```
profjsb, line 57: If this raises an error (ie. ill-formed request) then the error will not handled properly. Need to wrap this in a try/except statement and `return self.error(...)` something sensible if an error is raised.

skyportal/tests/flaky/test_followup_requests.py
```diff
@@ -165,6 +165,72 @@ def add_allocation_ztf(instrument_id, group_id, token):
...
+            "instrument_id": instrument_id,
+            "hours_allocated": 100,
+            "pi": "Ed Hubble",
+            '_altdata': '{"access_token": "testtoken"}',
+        },
+        token=token,
+    )
+    assert status == 200
+    assert data["status"] == "success"
+    return data["data"]
```
profjsb, line 183: why is this test returning something?

skyportal/facility_apis/sedmv2.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
+            The request to delete from the queue and the SkyPortal database.
+        """
+
+        from ..models import DBSession, FollowupRequest, FacilityTransaction
+
+        validate_request_to_sedmv2(request)
```
profjsb, line 112: same here with error catching (see above)

skyportal/facility_apis/sedmv2.py
```diff
@@ -0,0 +1,233 @@
...
+            },
+            "maximum_airmass": {
+                "title": "Maximum Airmass (1-3)",
+                "type": "number",
+                "default": 2.0,
+                "minimum": 1,
+                "maximum": 3,
+            },
+            "minimum_lunar_distance": {
+                "title": "Maximum Seeing [arcsec] (0-180)",
```
profjsb, line 187: This key and title seem to be about two different properties.

skyportal/facility_apis/sedmv2.py
```diff
@@ -0,0 +1,233 @@
...
+                "type": "number",
+                "default": 30.0,
+                "minimum": 0,
+                "maximum": 180,
+            },
+            "priority": {
+                "type": "number",
+                "default": 1.0,
+                "minimum": 1,
+                "maximum": 5,
```
profjsb, line 197:
```suggestion
                "minimum": 1.0,
                "maximum": 5.0,
```

profjsb (changes requested):
This looks pretty straightforward to me ... some small changes need IMO.

skyportal/facility_apis/sedmv2.py
mcoughlin: Good catch.

skyportal/facility_apis/sedmv2.py
mcoughlin: All of the other API's just raise Errors. What class is taking care of the error handling here?

skyportal/facility_apis/sedmv2.py
mcoughlin: Added!

skyportal/tests/flaky/test_followup_requests.py
mcoughlin: Great point! Removed!

skyportal/facility_apis/sedmv2.py
profjsb: `validate_request_to_sedmv2` either raises an error (`ValueError`) or returns None. Raising this error within an API request will not necessarily return back the sender a response they can parse. [`baselayer` overwrites the tornado `error` method](https://github.com/cesium-ml/baselayer/blob/ec952bbb3fcde91c19a9c2397c38f16e3d6b29c7/app/handlers/base.py#L286) to handle this gracefully.  So I'd do something like:

```
        try:
            validate_request_to_sedmv2(request)
        except (ValueError, ) as e:
            return self.error('SEMDv2 validation error: {e.normalized_messages()}')
```

mcoughlin: ok but that self isn't available anywhere here is it? this is a static method called in handlers/api/followup_request.py. The exception ends up in the red bar text on the front end.

profjsb: Ah, you're right. I thought that these APIs were subclasses of the API handlers we have in the rest of the app. But instead it's a base of `FollowUpAPI`. So what `self.` is indeed available here `self.error` has not been implemented. I'd argue we should implement it in the `FollowUpAPI` (or its base class) but that's beyond the scope of this PR.

profjsb (approved):

skyportal/facility_apis/sedmv2.py
```diff
@@ -0,0 +1,251 @@
...
+        raise ValueError('exposure_counts must be at least 1.')
+
+    if request.payload["maximum_airmass"] < 1:
+        raise ValueError('maximum_airmass must be at least 1.')
+
+    if (
+        request.payload["minimum_lunar_distance"] < 0
+        or request.payload["minimum_lunar_distance"] > 180
+    ):
+        raise ValueError('maximum_airmass must be within 0-180.')
```
guynir42, line 54:
```suggestion
        raise ValueError('minimum lunar distance must be within 0-180.')
```

skyportal/facility_apis/sedmv2.py
```diff
@@ -0,0 +1,251 @@
...
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
+            The request to delete from the queue and the SkyPortal database.
+        """
+
+        from ..models import DBSession, FollowupRequest, FacilityTransaction
+
+        validate_request_to_sedmv2(request)
```
guynir42, line 130: Do we need to submit all the required parameters even though we are just deleting a followup?

guynir42 (commented):
Some small comments.

skyportal/facility_apis/sedmv2.py
mcoughlin: Great point. Removed.

guynir42:
Looks good. Rerunning tests just to make sure the one failure was unrelated.

guynir42 (approved):

PR #2470, opened 2021-12-08
Attempt to speed up source spectra query

acrellin:
This patch attempts to speed up the source spectra fetching query
in the source spectra handler.


guynir42 (approved):

acrellin:
Failure unrelated (same failure present on main)

PR #2473, opened 2021-12-08
Update webpack config to resolve missing Buffer

acrellin:
This patch includes a module plugin to resolve
the no-longer-polyfilled Buffer library.


acrellin:
@stefanv Self-merging this small hotfix as it's affecting all other GA runs, but happy to discuss and adjust if you'd like.

stefanv:
Looks good. I suspect these will disappear when we go to TypeScript.

Issue #2477, opened 2021-12-09
Limit the number of matches returned in username auto-suggest


acrellin:
This will ensure the implementation will scale well as the user base grows. Currently, e.g., when entering "@", all users are returned and rendered in the list of matches.

PR #2479, opened 2021-12-10
Limit number of matches returned by auto-complete

acrellin:
This patch limits the number of matches returned by the user
mention auto-complete trie to ensure scalable performance with
arbitrarily large user bases.

Closes https://github.com/skyportal/skyportal/issues/2477

### DIFF
--- a/static/js/usernameTrie.js
+++ b/static/js/usernameTrie.js
@@ -4,22 +4,28 @@
 // Initialized empty: `const trie = UsernameTrie()`
 // Returns an object that supports the following methods:
 //  `insertUser({ username: string, firstName: string, lastName: string })` - inserts the user's names into the trie
-//  `findAllStartingWith(prefix)` - returns an object mapping username to `{ firstName, lastName }` for all user
-//    objects in the trie with at least one of username, first name or last name matching the given prefix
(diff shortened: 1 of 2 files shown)


static/js/usernameTrie.js
```diff
@@ -4,23 +4,29 @@
...
-  const getAllMatchesFromHere = (trieNode) => {
+  const getAllMatchesFromHere = (trieNode, limit) => {
     const matches = {};
+    let numMatches = 0;
 
     const traverseAndSaveMatches = (currNode = trieNode) => {
       if (currNode.matchTerminatesHere) {
         Object.keys(currNode.matchingUsersMap).forEach((username) => {
-          if (!matches[username]) {
+          if (!matches[username] && numMatches < limit) {
```
stefanv, line 21: Why is the `numMatches < limit` check needed here?

stefanv (commented):
Looks good! One small question.

static/js/usernameTrie.js
acrellin: In case there is more than one match terminating at the current node.  I suppose in our case that's not possible because each username is unique, but this doesn't hurt to have in here and keeps the implementation more general/extensible

acrellin: Ah, no, there can indeed be multiple matches ending at a given node, because we also store first names and last names in the trie. So this is needed to avoid going over the limit

PR #2481, opened 2021-12-11
Address missing key warning in WidgetPrefsDialog map

acrellin:


stefanv:
Should the FormControlLabel key then be removed?

acrellin:
@stefanv Probably so, yes -- updated.

PR #2483, opened 2021-12-13
Update requirements.txt

chobbaleee:
set python-ligo-lw to 1.7.1 (1.8.0 causes errors)

PR #2485, opened 2021-12-14
bump astropy ligo.skymaps

profjsb:
Supersedes #2483: we need to bump astropy to comport with ligo skymap pinning of astropy==5.0


acrellin:
Closing in favor of https://github.com/skyportal/skyportal/pull/2488

PR #2488, opened 2021-12-14
Upgrade ligo.skymap, astropy, astroquery; Update GA workflows

acrellin:


stefanv:
I re-ran CI, and failures are repeatable.

PR #2491, opened 2021-12-15
Version fixes

profjsb:
attempt to bump versions and geckodriver


.github/workflows/test_api_and_frontend.yaml
```diff
@@ -100,10 +100,12 @@ jobs:
           nginx -v
           firefox --version
 
-      - name: Install SkyPortal dependencies for master
-        run: |
-          export PYTHONPATH=$PYTHONPATH:$(pwd)
-          make dependencies
+      - name: Checkout branch being tested
```
acrellin, line 103: Please also remove the alembic command below

PR #2495, opened 2021-12-16
Require typing-extensions>=4.0.1

acrellin:


stefanv:
Do we use typing extensions? I don't think we're running mypy in our CI.

acrellin:
@stefanv I think it's `astropy` v5.0 that requires it.

stefanv:
So won't it come in automatically via astropy then?

acrellin:
@stefanv apparently astropy doesn't correctly specify the minimum required version of typing-extensions; if you have 3.x installed, you'll just see an error. Manually installing 4.x resolves it.

stefanv:
OK, please add a comment above that line explaining, then go ahead and merge.
(AFAIK `requirements.txt` respects `#` comments, and we should get in the habit of doing so for all strange cases).

stefanv:
Perfect, thanks!

PR #2497, opened 2021-12-17
Use requests.Session in api utility function

### LINKED ISSUE #2496: Specify `requests.Session.trust_env = False` and use session in data loader
Without this, auth will fail if the user has an existing netrc configuration, as requests uses any netrc auth to override provided kwargs by default (unless `trust_env` is set to `False`).

acrellin:
This allows us to bypass pre-existing netrc configurations
on the user's system that would be used to override the specified
authorization headers for every request.

Closes #2496

### DIFF
--- a/skyportal/tests/__init__.py
+++ b/skyportal/tests/__init__.py
@@ -10,6 +10,11 @@ IS_CI_BUILD = "TRAVIS" in os.environ or "GITHUB_ACTIONS" in os.environ
 
 patch_requests()
 
+session = requests.Session()
+session.trust_env = (
+    False  # Otherwise pre-existing netrc config will override auth headers
+)
+
 
 def api(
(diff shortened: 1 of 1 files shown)


stefanv:
This is not the only place requests is being used, so is this enough?

acrellin:
@stefanv My thinking here is that many more people will be spinning up the app for demo/simple testing purposes (the dev team, etc) than will be utilizing all of the functionality (which is more likely to be a production environment that won't have netrc configs laying around) so just starting with making sure it's possible to run the data loader is probably sufficient for now. If we run into cases where we need to extend this to other areas we're using `requests` we can do so then. Thoughts?

stefanv:
Yes, at least everything that uses our `api` function will work then.  We may want to add a warning to the docs about this, in `skyportal/api_description.md`.

stefanv:
Thanks, Ari!

PR #2498, opened 2021-12-28
WIP: ATLAS forced photometry

mcoughlin:
ATLAS forced photometry is a great request similar to our Follow-up requests but instead to an API (and returns photometry). This WIP PR tries that, with a few known issues below.

(i) ATLAS passbands missing from sncosmo: https://github.com/sncosmo/sncosmo/pull/336
(ii) I think what to commit photometry through API upon button press... but how to go from API to API?!


skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,272 @@
...
+                )
+                df['magsys'] = 'ab'
+
+                data_out = {
+                    'obj_id': request.obj_id,
+                    'instrument_id': instrument.id,
+                    'group_ids': 'all',
+                    **df.to_dict(orient='list'),
+                }
+                token_id = "720504e1-f163-478b-b624-a96573a68cbb"
```
profjsb, line 172: what is this? Should it be in a config? in .secrets?

profjsb (commented):
We may want to force photometry from other services (e.g. PTF, ASAS-SN) so either this new endpoint take a `service=atlas` query parameter or rename the endpoint to something like `atlas_photometry_request`

skyportal/facility_apis/atlas.py
mcoughlin: This was just my local token when testing. I assume there is a way to talk to the API directly from within the app without any tokens, but I don't know what that is!

mcoughlin:
@profjsb Agreed on the naming. the Pan-STARRS PR might be a good one to merge in which will force the naming stuff to be sorted out. It's a little funny because this is clearly modeled after the FollowupRequest stuff, so get(self, request_id, service) might be a bit weird because all of that specificity is captured in the api.get(followup_request) call but I am unsure.

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,272 @@
...
+                        'm': 'mag',
+                        'dm': 'magerr',
+                        'mag5sig': 'limiting_mag',
+                        'F': 'filter',
+                    },
+                    inplace=True,
+                )
+                cyan = df['filter'] == 'c'
+                orange = df['filter'] == 'o'
+                snr = df['uJy'] / df['duJy'] < 5
```
profjsb, line 139: how will this behave if there are NaNs? Or if `df['duJy'] == 0`?

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,272 @@
...
+
+                df['filter'].loc[cyan] = 'sdssg'
+                df['filter'].loc[orange] = 'sdssr'
+                df['mag'].loc[snr] = None
+                df['magerr'].loc[snr] = None
+                df = df.replace({np.nan: None})
+
+                df.drop(
```
profjsb, line 147: probably safer to drop all the columns not in the special set you want to keep. Ie.
```python
drop_columns = list(set(df.columns.values) - set(['mjd', 'ra', ...]))
```

skyportal/facility_apis/atlas.py
profjsb: I dont think this is our preferred access pattern (APIs calling APIs). @stefanv @guynir42 ?

skyportal/facility_apis/atlas.py
mcoughlin: great idea!

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,272 @@
...
+
+        mjd_min = Time(request.payload["start_date"], format='iso').mjd
+        mjd_max = Time(request.payload["end_date"], format='iso').mjd
+
+        target = {
+            'ra': request.obj.ra,
+            'dec': request.obj.dec,
+            'mjd_min': mjd_min,
+            'mjd_max': mjd_max,
+            'send_email': False,
```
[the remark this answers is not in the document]
mcoughlin: Added.

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,272 @@
...
+                s = requests.get(
+                    result_url,
+                    headers={
+                        'Authorization': f"Token {altdata['api_token']}",
+                        'Accept': 'application/json',
+                    },
+                )
+                s.raise_for_status()
+                df = pd.read_csv(
+                    StringIO(s.text.replace("###", "")), delim_whitespace=True
```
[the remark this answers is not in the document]
mcoughlin: Yes added.

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,272 @@
...
+                )
+                df.rename(
```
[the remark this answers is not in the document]
mcoughlin: Added.

skyportal/facility_apis/atlas.py
mcoughlin: Tried some handling of this.

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+            The request to add to the queue and the SkyPortal database.
+        """
+
+        from ..models import (
+            DBSession,
+            FollowupRequest,
+            FacilityTransaction,
+            Allocation,
+            Instrument,
+        )
```
profjsb, line 82: why are we not importing at the top of the file? This will slow up the response to the request

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+
+        """Submit a forced photometry request to ATLAS.
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+        """
+
+        from ..models import FacilityTransaction, DBSession
```
profjsb, line 253: same question

skyportal/facility_apis/fake_request.py
```diff
@@ -0,0 +1,33 @@
+from tornado import httputil
+
+
+class FakeApplication(object):
+    def __init__(
+        self,
+    ) -> None:
```
profjsb, line 7: not type annotating!

skyportal/facility_apis/fake_request.py
```diff
@@ -0,0 +1,33 @@
...
+        super().__init__()
+        self.ui_methods = {}
+        self.ui_modules = {}
+        self.settings = {}
+
+
+class FakeRequest(httputil.HTTPServerRequest):
+    def __init__(
+        self,
+    ) -> None:
```
profjsb, line 17: same

skyportal/facility_apis/fake_request.py
```diff
@@ -0,0 +1,33 @@
+from tornado import httputil
+
+
+class FakeApplication(object):
```
profjsb, line 4: in class construction Python 3 does not subclass `object`
```suggestion
class FakeApplication:
```

profjsb, line 4: Missing docstring ... also this code style looks different than elsewhere. Is this a copy of code from elsewhere? Should reference the origin if so.

skyportal/handlers/api/photometry.py
```diff
@@ -649,6 +668,71 @@ def get_stream_ids(self):
...
+        # hold it at a time.
+        DBSession().execute(
+            f'LOCK TABLE {Photometry.__tablename__} IN SHARE ROW EXCLUSIVE MODE'
+        )
+        try:
+            ids, upload_id = self.insert_new_photometry_data(
+                df, instrument_cache, group_ids, stream_ids, user=user
+            )
+        except ValidationError as e:
+            return self.error(e.args[0])
```
profjsb, line 732: is the table gauranteed to be unlocked? What if the exception happens? I dont see an unlock explictly done here.

profjsb (changes requested):
Looking good. Please show a screenshot of the front end in action.

skyportal/facility_apis/atlas.py
mcoughlin: This was following Danny's way of doing things. Also, there is a recursive import issue with the Facility API's because they appear in enum_types with LISTENER_CLASSES.

skyportal/facility_apis/atlas.py
mcoughlin: Same as above.

skyportal/facility_apis/fake_request.py
mcoughlin: I just removed this, which was based loosely on what tornado.web had.

skyportal/facility_apis/fake_request.py
mcoughlin: The docstring now mentions tornado.web directly.

mcoughlin:
image

skyportal/handlers/api/photometry.py
mcoughlin: This is just the same as what is already in POST.

profjsb: I'd like @stefanv to weigh in here since a DB lock seems very risky to me (it might be fine but I don't know now this works so I want me make sure).  It's possible the POST needs to get reworked if this is a concern.

mcoughlin:
@profjsb yes it is because I need the atlas filters in sncosmo sadly.

profjsb:
Ok, @mcoughlin. Now that #2581 is in, can you finalize this PR and retag me for review?

mcoughlin:
@profjsb At least the migrations now pass!

profjsb:
@mcoughlin can you merge in master and we'll see if we can get tests to pass?

data/tests/test_server_recordings_static.yaml
```diff
@@ -285,4 +285,59 @@ interactions:
...
+      Connection:
+      - keep-alive
+      Content-Length:
+      - '83'
+      Content-Type:
+      - application/x-www-form-urlencoded
+      User-Agent:
+      - python-requests/2.26.0
+    method: POST
+    uri: https://fallingstar-data.com/forcedphot/queue/?ra=0.0&dec=0.0&mjd_min=59227.66742310562&mjd_max=59592.66742310584&send_email=False
```
stefanv, line 306: When I hit this endpoint with the data specified here I get a 302 that redirects me to an authentication page.

So, is one call not enough to authorize and get the data?

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+from baselayer.app.env import load_env
+
+from ..utils import http
+
+env, cfg = load_env()
+
+if cfg['app.atlas.port'] is None:
+    ATLAS_URL = f"{cfg['app.atlas.protocol']}://{cfg['app.atlas.host']}"
+else:
+    ATLAS_URL = (
```
stefanv, line 20: I see we've got it this way in a couple of places in the config.  A bit silly really.  This could just be `ATLAS_URL=https://foo.bar:3125`

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+
+    """A dictionary structure for ATLAS forced photometry requests."""
+
+    def _build_payload(self, request):
+        """Payload json for ATLAS forced photometry requests.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
```
stefanv, line 35:
```suggestion
        request : skyportal.models.FollowupRequest
```

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+            The request to add to the queue and the SkyPortal database.
+
+        Returns
+        ----------
+        payload: json
```
stefanv, line 40:
```suggestion
        payload : dict
```

`dict`, unless you return a string specifically formatted according to JSON standards

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+            payload for requests.
+        """
+
+        mjd_min = Time(request.payload["start_date"], format='iso').mjd
```
stefanv, line 44: Any checking for payloads that don't have the required fields, or is this pre-processed to ensure it conforms to some schema?

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+    """An interface to ATLAS forced photometry."""
+
+    @staticmethod
+    def get(request):
+
+        """Get a forced photometry request result from ATLAS.
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
```
stefanv, line 72:
```suggestion
        request : skyportal.models.FollowupRequest
```

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+
+        instrument = (
+            Instrument.query_records_accessible_by(request.requester)
+            .join(Allocation)
+            .join(FollowupRequest)
+            .filter(FollowupRequest.id == request.id)
+            .first()
+        )
+
+        altdata = request.allocation.altdata
```
stefanv, line 99: Same questions as above throughout: are these requests pre-processed to ensure that they have all these fields?  If so, all OK, if not the fields need to be checked.

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+        r = requests.get(
+            content["url"],
+            headers={
+                'Authorization': f"Token {altdata['api_token']}",
+                'Accept': 'application/json',
+            },
+        )
+        r.raise_for_status()
+
+        if r.status_code == 200:
```
stefanv, line 116: Easier to short-circuit here:

```
if r.status_code != 200:
    ... do something ...
```

I don't quite understand the purpose of the if clause as-is; it just seems to set the request status, which never gets used again.

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+            if r.json()['finishtimestamp']:
```
stefanv, line 117: r.json() can error; not handled

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+                result_url = r.json()['result_url']
```
stefanv, line 118: r.json only once

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+                request.status = (
+                    f"Task is complete with results available at {result_url}"
+                )
+
+                s = requests.get(
```
stefanv, line 123: These are not async operations and will block; need to be handled async

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+        if not altdata:
+            raise ValueError('Missing allocation information.')
+
+        content = req.transactions[0].response["content"]
+        content = json.loads(content)
+
+        r = requests.get(
+            content["url"],
+            headers={
+                'Authorization': f"Token {altdata['api_token']}",
```
stefanv, line 110: Is this a token to the Atlas service?

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+                """
+
+                try:
+                    df = pd.read_csv(
+                        StringIO(s.text.replace("###", "")), delim_whitespace=True
+                    )
+                except Exception as e:
+                    raise ValueError(f'Format of response not understood: {e.message}')
+
+                desired_columns = ['MJD', 'RA', 'Dec', 'm', 'dm', 'mag5sig', 'F']
```
stefanv, line 149:
```suggestion
                desired_columns = {'MJD', 'RA', 'Dec', 'm', 'dm', 'mag5sig', 'F'}
```

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+                for col in desired_columns:
+                    if col not in list(set(df.columns.values)):
+                        raise ValueError(f'Missing expected column: {col}')
```
stefanv, line 152:
```suggestion
                if not desired_columns.issubset(df.columns):
                    raise ValueError(f'Missing expected column: {col}')
```

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+
+                df.rename(
```
stefanv, line 154: If you are going to do a rename anyway, then you may just as well do the rename earlier, replace `###MJD` with `mjd`, and not perform the stringio replacement.

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+                        'F': 'filter',
+                    },
+                    inplace=True,
+                )
+                cyan = df['filter'] == 'c'
+                orange = df['filter'] == 'o'
+
+                snr = df['uJy'] / df['duJy'] < 5
+
+                df['filter'].loc[cyan] = 'atlasc'
```
stefanv, line 171: What is this section doing?

skyportal/facility_apis/atlas.py
```diff
@@ -0,0 +1,310 @@
...
+                from skyportal.handlers.api.photometry import PhotometryHandler
+
+                application = FakeApplication()
+                fake_request = FakeRequest()
+
+                photometry_handler = PhotometryHandler(
+                    application=application,
+                    request=fake_request,
+                )
+                photometry_handler.add_external_photometry(
```
stefanv, line 215: This is a risky move; did you check that error handling goes according to plan if PhotometryHandler fails?

test_config.yaml
```diff
@@ -67,12 +67,17 @@ test_server:
...
     /node_agent2/node_agent: http://161.72.57.3:8080 # LT
     /api/requestgroups/: https://observe.lco.global # LCO
     /api/triggers/ztf: http://skipper.caltech.edu:4000 # ZTF
     /cgi-bin/internal/process_kait_ztf_request.py: http://herculesii.astro.berkeley.edu # KAIT
 
 app:
+  atlas:
+    protocol: http
+    host: localhost
+    port: 64502
```
stefanv, line 80: Same port as the LCO host?

stefanv (commented):
I am concerned about the invocation of handlers within facility API; does not seem kosher to me. The lock that needs to be taken out on the database is a good indication that this is a hack, and I'm worried that the lock might remain if something goes wrong.

alembic/versions/59ba401a110a_atlasapi_enum.py
```diff
@@ -0,0 +1,37 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute(
+        """
+alter type "public"."followup_apis" rename to "followup_apis__old_version_to_be_dropped";
+
+create type "public"."followup_apis" as enum ('ATLASAPI', 'KAITAPI', 'SEDMAPI', 'SEDMV2API', 'IOOAPI', 'IOIAPI', 'SPRATAPI', 'SINISTROAPI', 'SPECTRALAPI', 'FLOYDSAPI', 'MUSCATAPI', 'SLACKAPI', 'ZTFAPI');
```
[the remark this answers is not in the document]
mcoughlin: I am just calling what the migration script told me to do here.

skyportal/facility_apis/atlas.py
mcoughlin: I think some earlier telescope API reviews requested splitting them up.

data/tests/test_server_recordings_static.yaml
mcoughlin: I don't think I understand :)

test_config.yaml
mcoughlin: This is for Kyung's VCR stuff.

skyportal/facility_apis/atlas.py
mcoughlin: In https://github.com/skyportal/skyportal/pull/2648, all of these functions are completely pulled out. So It will just be a call to add_external_photometry.

stefanv:
Waiting on #2648 for using refactored photometry methods.

profjsb (commented):
Will wait on the resolution to Stefan's comments but looking pretty good here.

skyportal/facility_apis/atlas.py
mcoughlin: Added.

skyportal/facility_apis/atlas.py
mcoughlin: Adding this.

skyportal/facility_apis/atlas.py
mcoughlin: Yes.

skyportal/facility_apis/atlas.py
mcoughlin: Added.

skyportal/facility_apis/atlas.py
mcoughlin: Added some checking here.

skyportal/facility_apis/atlas.py
mcoughlin: Added here.

mcoughlin:
@profjsb you just need your ATLAS api key input in the data, this is how Danny setup all the external observatories. As to why it is in the dropdown, it's just like a ToO but one that returns quickly. It has API interactions and so on, so the interface is quite nice for it.

mcoughlin:
@profjsb Please consider merging if you are happy.

profjsb:
With #2648 merged (which Stefan was awaiting), I believe we can now merge this.

Issue #1393, opened 2020-11-19
Add "?" to low probability classifications


adamamiller:
When examining sources on fritz/other marshals, it is extremely useful to know if a classification is "good" or not. 

It would be nice if a "?" could be appended to the displayed label for any source with a low classification probability (I'll suggest P < 0.1 to start, but other thresholds could also be acceptable). So, this would then lead to a display of "SN Ia?" for something classified as SN Ia with P = 0.05. 

@dannygoldstein suspects it is a simple change to this line: https://github.com/skyportal/skyportal/blob/master/static/js/components/ShowClassification.jsx#L77

PR #2499, opened 2021-12-29
Add question mark to label of a classification if the probability is …

jadalilleboe:
…< 0.1. Also add classficiation data to the demo data.

This addresses issue #1393. Also, please let me know if I should add a test for this feature or if it's small enough to forgo a test. I also changed db_demo.yaml to make a request to the classification endpoint so I could see some classification chips.

### DIFF
--- a/skyportal/tests/__init__.py
+++ b/skyportal/tests/__init__.py
@@ -10,6 +10,11 @@ IS_CI_BUILD = "TRAVIS" in os.environ or "GITHUB_ACTIONS" in os.environ
 
 patch_requests()
 
+session = requests.Session()
+session.trust_env = (
+    False  # Otherwise pre-existing netrc config will override auth headers
+)
+
 
 def api(
     method, endpoint, data=None, params=None, host=None, token=None, raw_response=False
@@ -48,7 +53,7 @@ def api(
         host = f'http://localhost:{cfg["ports.app"]}'
     url = urllib.parse.urljoin(host, f'/api/{endpoint}')
     headers = {'Authorization': f'token {token}'} if token else None
-    response = requests.request(method, url, json=data, params=params, headers=headers)
+    response = session.request(method, url, json=data, params=params, headers=headers)
 
     if raw_response:
         return response


mcoughlin:
@jadalilleboe Maybe not a test, but could you add a screenshot here?

jadalilleboe:
Screenshot (5)
Screenshot (6)

mcoughlin:
@jadalilleboe I don't see the question mark in the screenshot. Sorry if I am just being silly.

jadalilleboe:
Is it supposed to be on the label chip? "Variable Taxonomy?" in the second screenshot

mcoughlin:
@jadalilleboe Ah yeah I was thinking maybe both initially, but what you did makes more sense since the P=0.02 seems obvious enough.

data/db_demo.yaml
```diff
@@ -471,6 +472,16 @@ sources/ZTFrlh6cyjh/annotations:
...
 
+classification:
+  - taxonomy_id: 1
+    obj_id: ZTF18aabcvnq
+    probability: 0.9
+    classification: "Time-domain Source"
+  - taxonomy_id: 2
+    obj_id: ZTF18aabcvnq
+    probability: 0.02
+    classification: "Variable Taxonomy"
```
guynir42, line 483: I'm a little confused about the name "Variable Taxonomy". Wouldn't "Variable Classification" be more straightforward? Because we have both Taxonomy and Classification objects and this seems to be a Classification...

mcoughlin: Maybe simpler would be "Variable Star"?

guynir42: Yeah "Variable Star" is much better.

jadalilleboe: Oh ok, I chose "Variable Taxonomy" because it was the first allowable class name for the taxonomy with id 2 in the demo data. I think the name of the classification has to match up with a class name from the taxonomy or the http request will fail with "That classification (Variable Star) is not in the allowed classes for the chosen taxonomy (id=2
make: *** [Makefile:59: load_demo_data] Error 255"

jadalilleboe: Here's the possible class names, any preferences?
Screenshot (7)
Screenshot (11)

Screenshot (9)](https://user-images.githubusercontent.com/82007190/147695618-365609d3-b9e5-4e49-b542-98566e92aa94.png)
ser-images.githubusercontent.com/82007190/147695616-60f39300-8948-4e70-9e58-ec8592040346.png)

mcoughlin: @jadalilleboe Ah "variable" then.

mcoughlin: @jadalilleboe Thanks for that. I guess the classification stuff has had no test to this point, so might be nice to take this opportunity to add one.

jadalilleboe:
Sorry the commits are messed up, trying to fix and squash it all into one

profjsb:
I think adding a small test is warranted here since we expect different behavior for different P values. Could add a few lines to an existing test if easier.

stefanv:
Agreed that adding a check to an existing test would be simplest/fastest here.

PR #2500, opened 2021-12-29
PS1 DR2 API access

mcoughlin:
This PR adds the ability to ingest PS1 DR2 data through MAST.


mcoughlin:
@stefanv @profjsb any idea of the magic required to get this migration to work? The crux of it is the sncosmo 2.5->2.7 upgrade..

profjsb (changes requested):
Looks good. Some unexpected behavior when:

 a) I try to delete the request (`Cannot delete requests on this instrument.`)
 b) when I try to retrieve a second time after I retrieved already (error = `objID`)
 c) when I try to reissue the exact request and then retrieve, I dont get an error saying that it's a duplicate. Not clear if the data additions are rejected by our API.

Double heck to make sure that the source position is in the right coordinates for what PS1 expects...I think default for ZTF is in J2015.5 (Gaia DR2).

mcoughlin:
@profjsb all good points. The first two behave correctly now. The third I checked and I see all of the points are rejected by the photometry post, but I am not sure what behavior would be better than that. What do you think?
For the last one, PS1 is J2000, so a mismatch here. I see the source offset handler explicitly calls for J2000. How are we dealing with epochs in the app?

profjsb (approved):

PR #2521, opened 2022-01-03
Add ? label to low probability classifications

jadalilleboe:
Sorry, I messed up the commits in the old pull request so I just made a new one for this feature.

Addresses issue #1393, and I've added a test now.
Screenshot (5)
Screenshot (6)


guynir42 (approved):
Nice work!

jadalilleboe:
Not sure why the ESLint tests are failing, it did pass when I ran the pre-commit hook locally.

guynir42:
> Not sure why the ESLint tests are failing, it did pass when I ran the pre-commit hook locally. 

Sometimes the local hooks ignore files that were not changed since the last commit (e.g., if you commit from the git hub interface). One way around it is to manually change the file and run the hook again and try to re-commit.

jadalilleboe:
Maybe it's a dependency issue because the eslint-config-airbnb was jus bumped? The error message doesn't point out any specific files where changes need to be made.

guynir42:
> Maybe it's a dependency issue because the eslint-config-airbnb was just bumped? The error message doesn't point out any specific files where changes need to be made. 

Yes it sounds like a configuration error. 
Try to pull upstream master, and make sure your config file is updated (from config.yaml.default) and then make sure the newest ESLint runs on your local file (because if it isn't changed then it won't trigger the hooks).

profjsb:
I think the ESLint issue is with a recently updated package. Please consider merging my PR to this fork to fix it.

jadalilleboe:
@profjsb which pr? I tried merging the eslint bump but the tests are still failing

profjsb:
No, I'm suggesting a downgrade of the airbnb eslint style, which introduced a bug. https://github.com/jadalilleboe/skyportal/pull/3

profjsb:
Dont bump eslint itself as in 781d7b8

jadalilleboe:
Ok, thank you. I can investigate why some of the api tests are failing, I'm not sure why because it seems like my changes have nothing to do with it

mcoughlin:
@jadalilleboe looks weird, maybe some of the recent package updates broke the tests and we didn't notice. @profjsb ?

mcoughlin:
@jadalilleboe do those tests still pass locally for you?

jadalilleboe:
@mcoughlin the test_get_ztfref_url test passes for me locally, but the test_starlist doesn't

mcoughlin:
@jadalilleboe can you see if something obvious with fixing the starlist one?

jadalilleboe:
It looks like the API call to the sources endpoint is failing because of authorization. Does this mean there's something wrong with the upload_data_token? Or am I not authorized to send patch requests to this endpoint?
Screenshot (14)

mcoughlin:
I think @profjsb wrote this particular test.

profjsb:
> I think @profjsb wrote this particular test.

It's failing on the PATCH to the sources, which is an auth issue, not related to the starlist stuff itself.

mcoughlin:
Does that mean there is a real issue with something @jadalilleboe added or a transient issue or something else @profjsb ?

profjsb:
> Does that mean there is a real issue with something @jadalilleboe added or a transient issue or something else @profjsb ?

Rerunning tests. It's possible there's a transient issue with the ztf_url test related to IRSA endpoints.

profjsb:
test failures are unrelated.

profjsb (approved):

PR #2531, opened 2022-01-05
downgrade airbnb linter

profjsb:
revert back to previous eslint-config-airbnb in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "babel-eslint": "10.1.0",
     "babel-loader": "8.2.2",
     "eslint": "7.27.0",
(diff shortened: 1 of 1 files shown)


stefanv:
What problem does this address?  It is not feasible for us to keep pinning to an older version in perpetuity, so we need a migration plan.

profjsb:
> What problem does this address? It is not feasible for us to keep pinning to an older version in perpetuity, so we need a migration plan.

Agreed. This was meant to address a linting issue that arose in PR #2521  https://github.com/skyportal/skyportal/pull/2521#issuecomment-1004254883. Not sure if that issue has been cleared up by other means.

PR #2532, opened 2022-01-05
Add KAIT as instrument

### LINKED ISSUE #2383: Add KAIT facility API

mcoughlin:
Closes #2383.

### DIFF
--- a/skyportal/facility_apis/kait.py
+++ b/skyportal/facility_apis/kait.py
@@ -37,9 +37,13 @@ class KAITRequest:
         """
 
         for filt in request.payload["observation_choices"]:
-            assert filt in ["U", "g", "r", "i", "z", "B", "V", "R", "I"]
+            if filt not in ["U", "g", "r", "i", "z", "B", "V", "R", "I"]:
+                raise ValueError(f"Improper observation_choice {filt}")
 
-        assert request.payload["observation_type"] in ["long", "short"]
+        if request.payload["observation_type"] not in ["long", "short"]:
+            raise ValueError(
+                f"Improper observation_type {request.payload['observation_type']}"
+            )
 
         # The target of the observation
         target = {
--- a/skyportal/facility_apis/kait.py
+++ b/skyportal/facility_apis/kait.py
@@ -1,0 +1,125 @@
+import requests
+from requests.auth import HTTPBasicAuth
+
+from . import FollowUpAPI
+from baselayer.app.env import load_env
+
+from ..utils import http
+
+env, cfg = load_env()
+
+
+if cfg['app.kait.port'] is None:
+    KAIT_URL = f"{cfg['app.kait.protocol']}://{cfg['app.kait.host']}"
+else:
+    KAIT_URL = (
+        f"{cfg['app.kait.protocol']}://{cfg['app.kait.host']}:{cfg['app.kait.port']}"
+    )
+
+
+class KAITRequest:
+
+    """A dictionary structure for KAIT ToO requests."""
+
+    def _build_payload(self, request):
+        """Payload json for KAIT queue requests.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        Returns
+        ----------
+        payload: json
+            payload for requests.
+        """
+
+        for filt in request.payload["observation_choices"]:
+            assert filt in ["U", "g", "r", "i", "z", "B", "V", "R", "I"]
+
+        # The target of the observation
+        target = {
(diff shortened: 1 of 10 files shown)


skyportal/facility_apis/kait.py
```diff
@@ -0,0 +1,125 @@
+import requests
+from requests.auth import HTTPBasicAuth
+
+from . import FollowUpAPI
+from baselayer.app.env import load_env
+
+from ..utils import http
+
+env, cfg = load_env()
+
```
profjsb, line 10:
```suggestion
```

skyportal/facility_apis/kait.py
```diff
@@ -0,0 +1,125 @@
...
+        for filt in request.payload["observation_choices"]:
+            assert filt in ["U", "g", "r", "i", "z", "B", "V", "R", "I"]
+
+        # The target of the observation
+        target = {
+            'name': request.obj.id,
+            'ra': request.obj.ra,
+            'dec': request.obj.dec,
+            'filters': "".join(request.payload["observation_choices"]),
+            'exposure': 'long',
```
profjsb, line 48: Will this always be a long exposure or are there other modes?

skyportal/services/test_server/test_api_server.py
```diff
@@ -147,6 +147,29 @@ def is_ztf_request(uri):
     assert r1_is_ztf and r2_is_ztf and r1.method == r2.method
 
 
+def kait_request_matcher(r1, r2):
+    """
+    Helper function to help determine if two requests to the KAIT API are equivalent
```
profjsb, line 152: Might you need to add some fuzzy matching around the ra, dec position? For instance if two requests have the same filter set but are located a small differences from each other (delta << FOV of KAIT), would that be essentially the same request? Thinking about the case of a refined position of a transient.

skyportal/facility_apis/kait.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+
+        for filt in request.payload["observation_choices"]:
+            assert filt in ["U", "g", "r", "i", "z", "B", "V", "R", "I"]
+
+        # The target of the observation
+        target = {
+            'name': request.obj.id,
+            'ra': request.obj.ra,
+            'dec': request.obj.dec,
```
profjsb, line 46: Should probably explicitly cast these floats to a string with a fixed number of decimal places. This will avoid having the URL serializer make the decision for you and might be useful if you want to make sure that a request isn't too precise on the position (e.g. only +- 0.1 arcsec) so that small updates in the astrometry dont lead to effectively duplicate requests.

profjsb (commented):
Looks really good. A few questions/comments.

skyportal/facility_apis/kait.py
mcoughlin: I asked WeiKang and long is all he gave me.

skyportal/services/test_server/test_api_server.py
mcoughlin: But this is really for the tests... this is the same as we have done for the others.

skyportal/facility_apis/kait.py
mcoughlin: Good thinking. Limited to %.5f.

guynir42 (approved):
LGTM

skyportal/facility_apis/kait.py
mcoughlin: Ok went back and confirmed that short is an option, added!

skyportal/facility_apis/kait.py
```diff
@@ -0,0 +1,133 @@
...
+        Returns
+        ----------
+        payload: json
+            payload for requests.
+        """
+
+        for filt in request.payload["observation_choices"]:
+            assert filt in ["U", "g", "r", "i", "z", "B", "V", "R", "I"]
+
+        assert request.payload["observation_type"] in ["long", "short"]
```
profjsb, line 42: Do not use `assert`s in production code (usually only for testing). Instead raise an exception explicitly. E.g.
```python
if request.payload.get("observation_type") not in ["long", "short"]:
     raise ValueError("Improper observation_type")
```
Using the `.get` instead of indexing `[]` should be safer in case `request.payload` does not have the requested key.

profjsb (changes requested):
small change requested.

profjsb:
@mcoughlin small change requested. Also, please resolve conflicts and I'll merge.

mcoughlin:
@profjsb should be ready for re-review.

profjsb (approved):
Thank you for addressing the comments. Looks great and can't wait to see it in action.

PR #2533, opened 2022-01-05
Replace d3 skymap with Aladin

mcoughlin:
This PR replaces the d3 skymap with Aladin by @ReihanMazouz. The npm import with Aladin lite did not yield sensible results, the link/script import works great.

image


mcoughlin:
@profjsb Ease of maintenance and existing APIs for adding tooltips and so on (and a much nicer visualization). You can see the mess of the crossmatch PR for simply adding in transient locations.

PR #2534, opened 2022-01-05
Fix offsets test

profjsb:
Makes the offsets test a bit more robust by checking to make sure that different catalogs are used in comparing the returned offset stars.


guynir42 (approved):
LGTM

profjsb:
self merging based on the PR from @guynir42

Issue #2536, opened 2022-01-06
Dockerfile should run system_setup after copying in configuration


stefanv:
This rebuilds the nginx and other service configs.

See #2535

rknop:
The Dockerfile does run it, but copies the config file after running it:
```
RUN bash -c "\
    source /skyportal_env/bin/activate && \
    make system_setup && \
    \
    ./node_modules/.bin/webpack --mode=production && \
    rm -rf node_modules && \
    \
    chown -R skyportal.skyportal /skyportal_env && \
    chown -R skyportal.skyportal /skyportal && \
    \
    mkdir -p /skyportal/static/thumbnails && \
    chown -R skyportal.skyportal /skyportal/static/thumbnails && \
    \
    cp docker.yaml config.yaml"
```

Probably just moving that cp to the beginning will fix it.

stefanv:
That's right.

PR #2537, opened 2022-01-06
Dockerfile: copy config before setting up services which may rely on it

stefanv:
E.g., the nginx configuration is templated, and depends on the skyportal
config to know which IP addresses to whitelist.

Closes gh-2536


stefanv:
Failure unrelated.

profjsb (approved):
LGTM. Unrelated failure will be addressed in another PR (by me)

PR #2541, opened 2022-01-08
Add GCN notice notifications for slack

mcoughlin:
This PR adds GCN notice notifications for slack.


profjsb (approved):
LGTM. I tested the existing Slack functionality and it seems to still work.

PR #2547, opened 2022-01-11
Add dedicated allocations page

mcoughlin:
This PR adds a counterpart to the "Observing Runs" page, but for the robotic allocations.


guynir42 (commented):
This looks a lot like #2548 so I would apply all the comments on that PR to this one, and then ping me for another look.

config.yaml.defaults
```diff
@@ -74,6 +74,8 @@ app:
       component: Filter
     - path: "/runs"
       component: ObservingRunPage
+    - path: "/allocations"
+      component: AllocationPage
```
guynir42, line 78: Probably want to put this below GCN events so it lines up with instruments and telescopes.

config.yaml.defaults
```diff
@@ -119,6 +121,10 @@ app:
       icon: LocalCafe
       url: /runs
 
+    - name: Allocations
+      icon: Satellite
```
guynir42, line 125: Maybe some clock icon? there must be a bunch of time-related icons that better fit this.

static/js/components/NewAllocation.jsx
```diff
@@ -0,0 +1,78 @@
...
+        ),
+        default: -1,
+        title: "Group",
+      },
+      _altdata: {
+        type: "string",
+        title: "Alternative json data (i.e. {'slack_token': 'testtoken'}",
+      },
+    },
+    required: ["pi", "start_date", "end_date", "instrument_id"],
```
guynir42, line 72: hours allocated should be required? I tried posting without it and it didn't work...

static/js/components/NewAllocation.jsx
```diff
@@ -0,0 +1,78 @@
...
+      start_date: {
+        type: "string",
+        format: "datetime",
+        title: "Start Date",
+      },
+      end_date: {
+        type: "string",
+        format: "datetime",
+        title: "End Date",
+      },
```
guynir42, line 41: Is it really important to have the time as well as the date? If the allocations are nightly (and don't start/end at specific hours) then maybe it is easier not to have to fill out the exact times...

skyportal/handlers/api/allocation.py
```diff
@@ -119,6 +119,7 @@ def post(self):
```
guynir42, line 119: I would add a test to make sure the start date is before the end date.

guynir42 (commented):
Looks good overall, some minor adjustments required.

static/js/components/NewAllocation.jsx
guynir42: Yeah ok it's probably best to leave it like this.

guynir42 (approved):
LGTM. If you merge the telescopes page you may have to pull master and fix conflicts before merging this. But other than that it looks good.

skyportal/handlers/api/allocation.py
mcoughlin: Added.

config.yaml.defaults
mcoughlin: made it an hourglass.

static/js/components/NewAllocation.jsx
mcoughlin: yes good point

config.yaml.defaults
mcoughlin: done

mcoughlin:
Failure unrelated. Can you take a look and merge if happy @guynir42 ?

guynir42:
I hope you don't mind I added a new line before the `(PI:` part. I think it now looks ok, I will merge when green.

mcoughlin:
Your allocations page is my allocations page @guynir42

guynir42:
I believe the problem with the test was this: https://stackoverflow.com/questions/3655549/xpath-containstext-some-string-doesnt-work-when-used-with-node-with-more

mcoughlin:
@guynir42 Ah interesting. I am confused why it works sometimes and not others, but hopefully that does it!

PR #2548, opened 2022-01-11
Add instrument page

mcoughlin:
Similar to the last PR, this page adds the ability to add an instrument.

### DIFF
--- a/skyportal/tests/frontend/test_instruments_page.py
+++ b/skyportal/tests/frontend/test_instruments_page.py
@@ -4,6 +4,12 @@ import uuid
 
 
 def test_instrument_frontend(super_admin_token, super_admin_user, driver):
+
+    driver.get(f"/become_user/{super_admin_user.id}")
+
+    # go to the allocations page
+    driver.get("/instruments")
+
     name = str(uuid.uuid4())
     status, data = api(
         'POST',
@@ -39,12 +45,21 @@ def test_instrument_frontend(super_admin_token, super_admin_user, driver):
     assert status == 200
     assert data['status'] == 'success'
 
-    driver.get(f"/become_user/{super_admin_user.id}")
+    driver.refresh()
 
-    # go to the allocations page
-    driver.get("/instruments")
+    # check for API instrument
+    table = driver.wait_for_xpath('//*[contains(@class, "MuiList-root")]')
+    findinst = False
+    for row in table.find_elements_by_xpath(
+        '//*[contains(@class, "MuiTypography-root")]'
+    ):
+        if row.text == f"{instrument_name}/{name}":
+            findinst = True
+    assert findinst
 
-    driver.wait_for_xpath('//*[@id="root_name"]').send_keys('ZTF')
+    # add dropdown instrument
+    instrument_name2 = str(uuid.uuid4())
+    driver.wait_for_xpath('//*[@id="root_name"]').send_keys(instrument_name2)
     driver.click_xpath('//*[@id="root_type"]')
     driver.click_xpath('//li[contains(text(), "Imager")]')
     driver.wait_for_xpath('//*[@id="root_band"]').send_keys('Optical')
@@ -55,4 +70,12 @@ def test_instrument_frontend(super_admin_token, super_admin_user, driver):
     driver.wait_for_xpath(submit_button_xpath)
     driver.click_xpath(submit_button_xpath)
 
-    driver.wait_for_xpath('//*[text()="f110w"]')
+    # check for API instrument
+    table = driver.wait_for_xpath('//*[contains(@class, "MuiList-root")]')
+    findinst = False
+    for row in table.find_elements_by_xpath(
+        '//*[contains(@class, "MuiTypography-root")]'
+    ):
+        if row.text == f"{instrument_name2}/{name}":
+            findinst = True
+    assert findinst
--- a/skyportal/tests/frontend/test_instruments_page.py
+++ b/skyportal/tests/frontend/test_instruments_page.py
@@ -46,7 +46,7 @@ def test_instrument_frontend(super_admin_token, super_admin_user, driver):
     assert data['status'] == 'success'
 
     # check for API instrument
-    # driver.wait_for_xpath(f'//span[text()="{instrument_name}/{name}"]')
+    driver.wait_for_xpath(f'//span[text()="{instrument_name}/{name}"]')
 
     # add dropdown instrument
     instrument_name2 = str(uuid.uuid4())
@@ -62,7 +62,7 @@ def test_instrument_frontend(super_admin_token, super_admin_user, driver):
     driver.click_xpath(submit_button_xpath)
 
     # check for dropdown instrument
-    driver.wait_for_xpath(f'//span[text()="{instrument_name2}/{name}"]')
+    # driver.wait_for_xpath(f'//span[text()="{instrument_name2}/{name}"]')
 
     # try adding a second time
     driver.wait_for_xpath('//*[@id="root_name"]').send_keys(instrument_name2)
--- a/skyportal/tests/frontend/test_instruments_page.py
+++ b/skyportal/tests/frontend/test_instruments_page.py
@@ -1,0 +1,46 @@
+from skyportal.tests import api
+
+import uuid
+
+
+def test_instrument_frontend(super_admin_token, super_admin_user, driver):
+    name = str(uuid.uuid4())
+    status, data = api(
+        'POST',
+        'telescope',
+        data={
+            'name': name,
+            'nickname': name,
+            'lat': 0.0,
+            'lon': 0.0,
+            'elevation': 0.0,
+            'diameter': 10.0,
+        },
+        token=super_admin_token,
+    )
+    assert status == 200
+    assert data['status'] == 'success'
+    telescope_id = data['data']['id']
+
+    instrument_name = str(uuid.uuid4())
+    status, data = api(
+        'POST',
+        'instrument',
+        data={
+            'name': instrument_name,
+            'type': 'imager',
+            'band': 'NIR',
+            'filters': ['f110w'],
+            'telescope_id': telescope_id,
+        },
+        token=super_admin_token,
+    )
+    assert status == 200
+    assert data['status'] == 'success'
+
+    driver.get(f"/become_user/{super_admin_user.id}")
+
+    # go to the allocations page
+    driver.get("/instruments")
+
+    driver.wait_for_xpath('//*[text()="f110w"]')
--- a/skyportal/tests/frontend/test_instruments_page.py
+++ b/skyportal/tests/frontend/test_instruments_page.py
@@ -46,7 +46,7 @@ def test_instrument_frontend(super_admin_token, super_admin_user, driver):
     assert data['status'] == 'success'
 
     # check for API instrument
-    driver.wait_for_xpath(f'//span[text()="{instrument_name}/{name}"]')
+    driver.wait_for_xpath(f'//span[text()="{instrument_name}/{name}"]', timeout=20)
 
     # add dropdown instrument
     instrument_name2 = str(uuid.uuid4())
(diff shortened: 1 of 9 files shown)


static/js/components/InstrumentPage.jsx
```diff
@@ -0,0 +1,101 @@
...
+        <CircularProgress color="secondary" />
+      </div>
+    );
+  }
+
+  let result = `${instrument?.name}/${telescope?.nickname}`;
+
+  if (instrument?.filters || instrument?.api_classname) {
+    result += "(";
+    if (instrument?.filters) {
```
guynir42, line 41: add a space before parenthesis.

static/js/components/InstrumentPage.jsx
```diff
@@ -0,0 +1,101 @@
...
+      result += `filters: ${instrument.filters}`;
+    }
+    if (instrument?.api_classname) {
+      result += ` / API Classname: ${instrument?.api_classname}`;
+    }
+    result += ")";
+  }
+
```
guynir42, line 49: I would put the name of the instrument/telescope in a separate div from the other info, and maybe make that div in bold or something. It would make it much more readable.

static/js/ducks/instrument.js
```diff
@@ -0,0 +1,44 @@
...
+export const fetchInstrument = (id) =>
+  API.GET(`/api/instrument/${id}`, FETCH_INSTRUMENT);
+
+export const submitInstrument = (run) =>
+  API.POST(`/api/instrument`, SUBMIT_INSTRUMENT, run);
+
+// Websocket message handler
+messageHandler.add((actionType, payload, dispatch, getState) => {
+  const { instrument } = getState();
+  if (actionType === REFRESH_INSTRUMENT) {
```
guynir42, line 22: I think the reason the page doesn't update is that you need to send out a "REFRESH INSTRUMENT" on the back end (python handler for POST instrument) right before transmitting a success return statement.

static/js/components/NewInstrument.jsx
```diff
@@ -0,0 +1,102 @@
...
+  instrumentList?.forEach((instrument) => {
+    instrument.filters?.forEach((filter) => {
+      filters.push(filter);
+    });
+  });
+  const filtersUnique = [...new Set(filters)];
+
+  const uiSchema = {
+    filters: {
+      "ui:widget": "checkboxes",
```
guynir42, line 42: can we make this two-columns? If we get a few more filters this form becomes too tall.

static/js/components/NewInstrument.jsx
```diff
@@ -0,0 +1,102 @@
...
+    if (formData.group_id === -1) {
+      delete formData.group_id;
+    }
+    const result = await dispatch(submitInstrument(formData));
+    if (result.status === "success") {
+      dispatch(showNotification("Instrument saved"));
+    }
+  };
+
+  const api_classnames = [{ enum: [null], title: "No API" }];
```
guynir42, line 22: I'm getting a weird warning in the console: 
`Warning: `value` prop on `input` should not be null.`
I am guessing it has to do with putting null values for the default API class, or something. Worth looking into it.

static/js/ducks/instrument.js
```diff
@@ -0,0 +1,44 @@
...
+    const { instrument_id } = payload;
```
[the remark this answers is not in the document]
mcoughlin: Now I explicitly call instruments to get the new list. That ok?

static/js/components/InstrumentPage.jsx
mcoughlin: done.

static/js/components/InstrumentPage.jsx
mcoughlin: see if you like the update!

static/js/components/NewInstrument.jsx
mcoughlin: https://stackoverflow.com/questions/64529750/multi-column-support-for-react-jsonschema-form
surprisingly hard request... what do you think?

static/js/components/NewInstrument.jsx
mcoughlin: yes made this just "", so I think better

static/js/ducks/instrument.js
mcoughlin: See below.

static/js/components/InstrumentPage.jsx
```diff
@@ -0,0 +1,101 @@
...
+          <div className={classes.paperContent}>
+            <Typography variant="h6">Add a New Instrument</Typography>
+            <NewInstrument />
+          </div>
+        </Paper>
+      </Grid>
+    </Grid>
+  );
+};
+
```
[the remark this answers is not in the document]
mcoughlin: If only I could actually take credit for this ;)

static/js/components/NewInstrument.jsx
guynir42: If it is not easy to do I'd advocate skipping it for now. If the filter list grows too long we can figure out how to change it then.

skyportal/tests/frontend/test_instruments_page.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    assert status == 200
+    assert data['status'] == 'success'
+
+    driver.get(f"/become_user/{super_admin_user.id}")
+
+    # go to the allocations page
+    driver.get("/instruments")
+
+    driver.wait_for_xpath('//*[text()="f110w"]')
```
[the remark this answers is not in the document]
mcoughlin: Ok I improved the test a bit here hitting submit.

static/js/ducks/instrument.js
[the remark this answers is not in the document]
guynir42: That seems to work when adding from the front end. This is probably because you explicitly call `fetchInstruments` from the dispatch. If someone used the API to post a new instrument it would not show up unless you refresh. 
The way around this is to send a REFRESH_INSTRUMENTS websocket message from the handler (instrument.py) and have that picked up by the reducer (instruments.js). Look at what is done in groups (group.py and groups.js).

static/js/components/NewInstrument.jsx
```diff
@@ -0,0 +1,106 @@
+import React from "react";
+import { useDispatch, useSelector } from "react-redux";
+import Form from "@rjsf/material-ui";
+import { showNotification } from "baselayer/components/Notifications";
+import { submitInstrument } from "../ducks/instrument";
+import { fetchInstruments } from "../ducks/instruments";
+
+const NewInstrument = () => {
```
guynir42, line 8: I managed to crash the page when posting an instrument. Something to do with form verification. I will try to recreate but I suggest you play around with it too, you may be able to shake something loose.

mcoughlin: Ok I think I got these. See what you think. There is now a REFRESH_INSTRUMENTS  call (I think), please let me know if it is what you had in mind.

static/js/ducks/instruments.js
```diff
@@ -13,6 +17,13 @@ export const fetchInstruments = () =>
 export const fetchInstrumentForms = () =>
   API.GET(`/api/internal/instrument_forms`, FETCH_INSTRUMENT_FORMS);
 
+// Websocket message handler
+messageHandler.add((actionType, payload, dispatch) => {
+  if (actionType === REFRESH_INSTRUMENTS) {
```
guynir42, line 22: Great. Now you have to make sure this message is actually being sent, by adding a `self.push_all(action='skyportal/REFRESH_INSTRUMENTS)` to the instrument.py handler, for POST, PUT and DELETE. This means if you add/change/remove an instrument using the API, the front end will know about this. 

On the first version of the test you had for this page you posted an instrument via API. I suggest you return that test (besides the front end input you have) and make sure the new instrument appears without a page refresh. That way you'll know you closed the loop.

mcoughlin: Ah yes, added now. And that instrument API call is still there, nominally that is what that filter check is doing.

skyportal/tests/frontend/test_instruments_page.py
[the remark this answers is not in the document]
guynir42: I would reorganize the test:
1) become user and go to the instruments page
2) then post the first instrument via API, and check it appears by name (not by `f110w`). 
3) post another instrument using the front end, with a new name (using uuid again) 
4) check it appears. 
This way you verify the new instruments appear without refresh, both using the API and the front end.

profjsb (commented):
- could you post some screen shots of the look and feel?
- should test the behavior when an instrument is added which already exists

mcoughlin:
Screen Shot 2022-01-13 at 10 02 42 AM
Screen Shot 2022-01-13 at 10 03 10 AM

skyportal/tests/frontend/test_instruments_page.py
```diff
@@ -39,12 +45,31 @@ def test_instrument_frontend(super_admin_token, super_admin_user, driver):
     assert status == 200
     assert data['status'] == 'success'
 
-    driver.get(f"/become_user/{super_admin_user.id}")
+    driver.refresh()
```
guynir42, line 48: This should work without needing refresh!

skyportal/tests/frontend/test_instruments_page.py
```diff
@@ -39,12 +45,31 @@ def test_instrument_frontend(super_admin_token, super_admin_user, driver):
...
+    driver.wait_for_xpath('//*[@id="root_band"]').send_keys('Optical')
+    driver.click_xpath('//*[@id="root_api_classname"]')
+    driver.click_xpath('//li[contains(text(), "ZTFAPI")]')
 
-    driver.wait_for_xpath('//*[@id="root_name"]').send_keys('ZTF')
+    submit_button_xpath = '//button[@type="submit"]'
+    driver.wait_for_xpath(submit_button_xpath)
+    driver.click_xpath(submit_button_xpath)
+
+    driver.refresh()
```
guynir42, line 66: same here.

skyportal/tests/frontend/test_instruments_page.py
```diff
@@ -0,0 +1,81 @@
...
+    driver.wait_for_xpath('//*[@id="root_band"]').send_keys('Optical')
+    driver.click_xpath('//*[@id="root_api_classname"]')
+    driver.click_xpath('//li[contains(text(), "ZTFAPI")]')
+
+    submit_button_xpath = '//button[@type="submit"]'
+    driver.wait_for_xpath(submit_button_xpath)
+    driver.click_xpath(submit_button_xpath)
+
+    # check for dropdown instrument
+    # driver.wait_for_xpath(f'//span[text()="{instrument_name2}/{name}"]')
```
guynir42, line 65: remove commented line

skyportal/tests/frontend/test_instruments_page.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_instrument_frontend(super_admin_token, super_admin_user, driver):
+
+    driver.get(f"/become_user/{super_admin_user.id}")
+
+    # go to the allocations page
+    driver.get("/instruments")
+
+    name = str(uuid.uuid4())
```
guynir42, line 13: call this `telescope_name`.

static/js/components/NewInstrument.jsx
guynir42: I think it looks ok now. Thanks!

guynir42 (approved):
A couple of very minor comments and then approved.

skyportal/tests/frontend/test_instruments_page.py
mcoughlin: ah good idea

skyportal/tests/frontend/test_instruments_page.py
mcoughlin: ah yes!

mcoughlin:
@guynir42 can we get this merged under the new guidelines?

guynir42:
@mcoughlin I think maybe @profjsb would like to have a look. If not then it is good to merge as far as I'm concerned.

profjsb (approved):
LGTM. thanks for working on the comments.

PR #2549, opened 2022-01-11
Draft: Add generic passbands

### LINKED ISSUE #2526: Generic passbands with sncosmo
We are currently restricted to the "named" passbands available in sncosmo.
https://sncosmo.readthedocs.io/en/stable/bandpass-list.html

However, this means we are missing some available filters, especially from smaller telescopes (also no-passband/QE-dominated like Clear). While Danny had suggested ingesting sncosmo ingest all of SVO (see #459), for now, we could simply take in min_wavelength and max_wavelength through their generic interface (https://sncosmo.readthedocs.io/en/stable/bandpasses.html).

jadalilleboe:
Closes #2526 

Adds generic passbands to SkyPortal. Adds a generic passband model in the DB, a generic passband API endpoint, and tests for the endpoint. Also adds demo data for some generic passbands.

In Progress: Registering the generic passbands in the ALLOWABLE_BANDPASSES variable in skyportal/enum_types.py, which will allow instruments to use the custom filters.


mcoughlin:
@jadalilleboe I pushed a little example. I think we want to register the filters so that the rest of the interface can be used as per normal. So no migrations would be required. We could keep track of the extra filters in some kind of CSV file maybe?

jadalilleboe:
@mcoughlin So should the API accept a custom_filters array containing objects with properties name, min_wavelength, and max_wavelength? Then check to see if the filter already exists in the csv file, if not then add it to the allowed bandpasses and the csv file. I see in your example the extra filters are passed in the filters array, so where will the min_wavelength and max_wavelength come from?

mcoughlin:
The user defines the min and max wavelengths, but obviously that is a trivial statement. Maybe we have an API (and frontend element) that simply registers the name min and max wavelength and writes those to a DB table which are then read in by enum_types?

jadalilleboe:
@mcoughlin so a new API endpoint for custom filters, not modifying the api/instrument/ endpoint?

mcoughlin:
Yeah a new table seems better.

jadalilleboe:
@mcoughlin Wait so should the data be kept in a csv file or a table in the database?

mcoughlin:
Sorry I know I am flip flopping. Table in the DB means it will survive taking down and bringing up the app. Maybe the example CSV can be demo data.

jadalilleboe:
@mcoughlin should the wavelengths be floats or ints?

mcoughlin:
@jadalilleboe I think enabling floats would be good.

jadalilleboe:
@mcoughlin I've got the skeleton for the endpoint and model, let me know what you think. the thing I'm having trouble with is passing the generic passbands in to ALLOWED_BANDPASSES in enum_types.py. I was having trouble directly querying the database from enum_types.py because of a circular import, and using a function in enum_types to add new passbands doesn't seem to work either. Also, is ALLOWED_BANDPASSES supposed to be a constant because it's in all caps? So should I even be changing that variable at all?

jadalilleboe:
@mcoughlin Did you also want a frontend element for adding a generic passband? What page should that be on?

mcoughlin:
Maybe it would make sense on the forthcoming Instrument page? (https://github.com/skyportal/skyportal/pull/2548) @guynir42 what do you think?

guynir42:
> Maybe it would make sense on the forthcoming Instrument page? (#2548) @guynir42 what do you think?

Yes I think that makes sense. My feeling is that there are two main use cases:
(1) people have automated systems that post arbitrary filters (e.g., using synthetic photometry) in which case posting from the API is all you need. 
(2) people just have a new instrument and instead of posting a filter for it they just fill out the bandwidth (then they'd want the front end). 
I'm not sure if in general we want to allow (2), maybe it's worth taking up with the team on standup?

mcoughlin:
@guynir42 Makes sense. For 2, it seems like a nice thing given that we are pretty restricted currently using only the sncosmo filter set, and many folks aren't super comfortable with using APIs so enabling front end things is a nice quality of life thing.

guynir42:
@mcoughlin I agree in general. But it may be that we want to have more control over filters being added. For instance, if I want to post some new data with a new filter, I could have my admin add that filter (or set of filters) once, which is not too much to ask of the admin if it doesn't happen very often. This would prevent each user adding similar filters on different occasions.

mcoughlin:
@guynir42 Good point as well. I would generally assume people would not want to go to the trouble of posting filters unless they really needed it, adding some security, but it is true.

profjsb (commented):
As currently implemented I do not think we will get the desired behavior: the `enum_types.py` file makes an SQL alchemy ENUM type for use in the models. Simply updating the sncosmo `_BANDPASSES` variable will not update these model enums.

PR #2550, opened 2022-01-12
Create SLACKAPI to allow instrument triggering via Slack

mcoughlin:
Slack is a common place to track submission requests, and this PR gives creates a SLACKAPI to allow SP to use the microservice to submit requests to a designated Slack channel.

Blocker: Needs migration


profjsb (changes requested):
It is important to document (in `docs/`?) how this functionality will be used. It's not clear from the  code how a user/bot gets access to this -- in particular we only enable Slack in the user profile. It's unclear how these requests will enabled (and checked that they are enabled) using API Tokens.   Remember that the Slack interaction is not guaranteed to work -- so should make clear that a request may never go through (e.g. if Slack is down). 

 It's also not clear why specific filters are allowed/disallowed.

I think a detailed description of the usage patterns will help a lot here.

Missing tests.

mcoughlin:
@profjsb These are good questions. Indeed, I would like the SLACKAPI to be able to take in arbitrary filters and so on, so maybe it should be ingesting from whatever the instrument's options are somehow, but maybe it will be tricky to make a generic enough interface...

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,174 @@
...
+            The request to add to the queue and the SkyPortal database.
+
+        Returns
+        ----------
+        payload: json
+            payload for requests.
+        """
+
+        for filt in request.payload["observation_choices"]:
+            if filt not in ["g", "r", "i", "z"]:
```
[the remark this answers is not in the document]
mcoughlin: Fixed.

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,174 @@
...
+                raise ValueError(f"Improper observation_choice {filt}")
+
+        if request.payload["exposure_time"] <= 0:
```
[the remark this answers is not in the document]
mcoughlin: I made it 2 hours.

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,174 @@
...
+
+if cfg['app.slackapi.port'] is None:
+    SLACK_URL = f"{cfg['app.slackapi.protocol']}://{cfg['app.slackapi.host']}/services"
+else:
+    SLACK_URL = f"{cfg['app.slackapi.protocol']}://{cfg['app.slackapi.host']}:{cfg['app.slackapi.port']}/services"
+
+
+class SLACKRequest:
+
+    """A dictionary structure for SLACK ToO requests."""
```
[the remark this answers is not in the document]
mcoughlin: Added.

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,174 @@
...
+        # The target of the observation
+        target = {
+            'name': request.obj.id,
+            'instrument_request': request.allocation.instrument.name,
+            'ra': "{:.5f}".format(request.obj.ra),
+            'dec': "{:.5f}".format(request.obj.dec),
+            'filters': ",".join(request.payload["observation_choices"]),
+            'exposure_time': request.payload["exposure_time"],
+            'exposure_counts': request.payload["exposure_counts"],
+            'username': request.requester.username,
```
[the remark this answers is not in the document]
mcoughlin: I think so... otherwise it's just the app and no one knows who sent it.

config.yaml.defaults
```diff
@@ -38,6 +38,10 @@ app:
     protocol: http
     host: herculesii.astro.berkeley.edu
 
+  slackapi:
+    protocol: https
+    host: hooks.slack.com
+
```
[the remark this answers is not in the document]
mcoughlin: What exactly needs to be enabled here? What is the slack service checking for exactly?

data/db_demo.yaml
```diff
@@ -290,6 +298,13 @@ instrument:
     filters: ["sdssg", "sdssr", "sdssi", "sdssz"]
     api_classname: SEDMV2API
     =id: SEDMv2
+  - band: optical
+    name: T80CAM
+    telescope_id: =JAST
+    type: imager
+    filters: ['sdssg', 'sdssr', 'sdssi', 'sdssz']
```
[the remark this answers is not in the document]
mcoughlin: Fixed.

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,174 @@
...
+
+    form_json_schema = {
+        "type": "object",
+        "properties": {
+            "observation_choices": {
+                "type": "array",
+                "title": "Desired Observations",
+                "items": {
+                    "type": "string",
+                    "enum": ["g", "r", "i", "z"],
```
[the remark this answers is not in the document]
mcoughlin: Fixed.

doc/slack.md
```diff
@@ -64,3 +64,8 @@ Unlike a fault-tolerant pub-sub mechanism like XMPP Jabber, Slack notifications
 The Slack Admin is trusted to verify that the SkyPortal user they send a Webhook URL to is the same person (or entity) represented in their Slack Workspace.  For most use cases this is a reasonable measure to keep SkyPortal data secure. In addition, while the names and URLs of sources are shown in Slack, no other information (such as comments or classification results) is transmitted. So even if identities are mismatched or channels are compromised, a Slack user with unauthorized access will not be able to access substantive information in SkyPortal.
 
 The Slack admin may delete the Slack App for their workspace at any time, invalidating all webhooks.
+
+
+## Use as an instrument triggering API
```
profjsb, line 69: this section should go above `Notes on Reliability and Security`.

doc/slack.md
```diff
@@ -64,3 +64,8 @@ Unlike a fault-tolerant pub-sub mechanism like XMPP Jabber, Slack notifications
...
+
+We have enabled the use of Slack for instrument "triggering," enabling a push of a transient follow-up request similar to done for other instruments. To enable this, an instrument needs SLACKAPI added as its triggering API, and in the instrument allocation, _altdata should look like {"slack_workspace": "XXX", "slack_channel": "YYY", "slack_token": "ZZZ"} with the appropriate values.
```
profjsb, line 71:
```suggestion
We have enabled the use of Slack for instrument "triggering," enabling a push of a transient follow-up request similar to done for other instruments. To enable this, an instrument needs `SLACKAPI` added as its triggering API, and in the instrument allocation, `_altdata` should look like `{"slack_workspace": "XXX", "slack_channel": "YYY", "slack_token": "ZZZ"}` with the appropriate values. The Slack endpoint will be POSTed to `{SLACK_URL}/{altdata['slack_workspace']}/{altdata['slack_channel']}/{altdata['slack_token']}`
```

profjsb (changes requested):
This will need a migration, as you're adding SLACKAPI to followup_facilities enum. See an example of this in: https://github.com/profjsb/skyportal/blob/90eccb64de9c68621f2ed25f61c9a3e22abca7d9/alembic/versions/f7c1e785d7a4_unique_token_name_by_user.py

mcoughlin:
@profjsb do you understand why the migration checker wasn't triggered on this to give the migra artifact? It worked fine for the observation plan PR which also adds one of these.

stefanv:
No files were touched that would give the workflow a reason to suspect that migrations are involved, so it doesn't run that check.

mcoughlin:
Ah I see. Maybe it could be made to check the init file in facility_apis?

profjsb:
Yes we need to update the GHA action to watch facility APIs

mcoughlin:
@profjsb Yes I am trying it here https://github.com/skyportal/skyportal/pull/2498

profjsb:
I put in a separate PR for watching the files: https://github.com/skyportal/skyportal/pull/2591

profjsb:
Once that gets merged you can rebase and rerun tests and that should trigger the migration manager.

mcoughlin:
yes good. It's not totally trivial to make this migration script work it seems, at least based on the atlas one so far.

mcoughlin:
@profjsb I get migration artifacts that look like the below. Given that the migration script adds SLACKAPI to followup_apis, which (if any) of these lines are relevant to put into the alembic script? All of them?

alter type "public"."followup_apis" rename to "followup_apis__old_version_to_be_dropped";

create type "public"."followup_apis" as enum ('KAITAPI', 'SEDMAPI', 'SEDMV2API', 'IOOAPI', 'IOIAPI', 'SPRATAPI', 'SINISTROAPI', 'SPECTRALAPI', 'FLOYDSAPI', 'MUSCATAPI', 'SLACKAPI', 'ZTFAPI');

alter table "public"."instruments" alter column api_classname type "public"."followup_apis" using api_classname::text::"public"."followup_apis";

drop type "public"."followup_apis__old_version_to_be_dropped";

profjsb:
> @profjsb I get migration artifacts that look like the below. Given that the migration script adds SLACKAPI to followup_apis, which (if any) of these lines are relevant to put into the alembic script? All of them?
> 
> alter type "public"."followup_apis" rename to "followup_apis__old_version_to_be_dropped";
> 
> create type "public"."followup_apis" as enum ('KAITAPI', 'SEDMAPI', 'SEDMV2API', 'IOOAPI', 'IOIAPI', 'SPRATAPI', 'SINISTROAPI', 'SPECTRALAPI', 'FLOYDSAPI', 'MUSCATAPI', 'SLACKAPI', 'ZTFAPI');
> 
> alter table "public"."instruments" alter column api_classname type "public"."followup_apis" using api_classname::text::"public"."followup_apis";
> 
> drop type "public"."followup_apis__old_version_to_be_dropped";

What you're doing in the migration script looks good to me: https://github.com/skyportal/skyportal/blob/9243cc446c878fe6725b0613b68332514f609068/alembic/versions/82f5a89fbe59_slack_trigger_api.py 

migra is detecting a change in the enum and this will (or at least should) satisfy the differences.

mcoughlin:
@profjsb sorry for being dense.. does this mean that file is sufficient or do I need to add those lines into the alembic script?

profjsb:
> @profjsb sorry for being dense.. does this mean that file is sufficient or do I need to add those lines into the alembic script?

No -- those lines above are the alternative route to doing enum updates. Both should work but it's possible migra will only be happy with the verbose way above. Regardless we should not have anything in the `downgrade` path because we dont want to remove enums since that might cause referential integrity issues.

profjsb (approved):

alembic/versions/82f5a89fbe59_slack_trigger_api.py
```diff
@@ -0,0 +1,39 @@
...
+def upgrade():
+    # add SLACKAPI to followup_apis
+    # with op.get_context().autocommit_block():
+    #     op.execute("ALTER TYPE followup_apis ADD VALUE IF NOT EXISTS 'SLACKAPI'")
+
+    op.execute(
+        """
+alter type "public"."followup_apis" rename to "followup_apis__old_version_to_be_dropped";
+
+create type "public"."followup_apis" as enum ('KAITAPI', 'SEDMAPI', 'SEDMV2API', 'IOOAPI', 'IOIAPI', 'SPRATAPI', 'SINISTROAPI', 'SPECTRALAPI', 'FLOYDSAPI', 'MUSCATAPI', 'SLACKAPI', 'ZTFAPI');
```
stefanv, line 27: My guess here:

The reason this works and not the one above is because you need to do a casting operation of sorts, since the ENUM order changes (note that `'SLACKAPI'` does not get appended, but ends up second from last).

What happens then is that a new replacement enum is created.  This new enum is an entirely different enum, so the database needs to be updated to cast between the old and the new one.  That is what the `using api_classname::text::"public"."followup_apis"` part does—it says change the `api_classname` column to use the new enum and use the values currently in that column to look them up *in the new enum* (the underlying numbers stored on disk will change).

This is probably a somewhat costly operation, but it doesn't seem to have caused problems for us so far.

In the future, the cheapest way to update an enum would be to only *append* values, and then use `ADD VALUE` as Michael did before.

Here's an example, borrowing from the postgres docs, illustrating the above:

```
skyportal=> CREATE TYPE mood AS ENUM ('sad', 'ok', 'happy');
CREATE TYPE
skyportal=> CREATE TABLE person (
    name text,
    current_mood mood
);
CREATE TABLE
skyportal=> INSERT INTO person VALUES ('Moe', 'happy');
INSERT 0 1
skyportal=> SELECT * FROM person WHERE current_mood = 'happy';
 name | current_mood 
------+--------------
 Moe  | happy
(1 row)

skyportal=> alter type mood rename to mood_old;
ALTER TYPE
skyportal=> create type mood as enum ('sad', 'ok', 'not happy', 'happy');
CREATE TYPE
skyportal=> alter table person alter column current_mood type mood;
ERROR:  column "current_mood" cannot be cast automatically to type mood
HINT:  You might need to specify "USING current_mood::mood".
                                                             ^
skyportal=> alter table person alter column current_mood type mood using current_mood::text::mood;
ALTER TABLE
skyportal=> SELECT * FROM person WHERE current_mood = 'happy';
 name | current_mood 
------+--------------
 Moe  | happy
(1 row)

skyportal=> alter type mood add value 'grumpy';
ALTER TYPE
skyportal=> SELECT * FROM person WHERE current_mood = 'happy';
 name | current_mood 
------+--------------
 Moe  | happy
(1 row)
```

config.yaml.defaults
```diff
@@ -433,6 +433,8 @@ image_grayscale_params:
 slack:
   microservice_port: 64100
   expected_url_preamble: "https://hooks.slack.com/"
+  slack_followup_requests:
```
stefanv, line 436: Nitpick, can be ignored:

```suggestion
  followup_requests:
```

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+
+    """A dictionary structure for SLACK ToO requests."""
+
+    def _build_payload(self, request):
+        """Payload json for SLACK queue requests.
+
+        Parameters
+        ----------
+
+        request: skyportal.models.FollowupRequest
```
stefanv, line 30:
```suggestion
        request : skyportal.models.FollowupRequest
```

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+class SLACKRequest:
+
+    """A dictionary structure for SLACK ToO requests."""
+
+    def _build_payload(self, request):
+        """Payload json for SLACK queue requests.
+
+        Parameters
+        ----------
+
```
stefanv, line 29:
```suggestion
```

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+        request: skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
+
+        Returns
+        ----------
+        payload: json
```
stefanv, line 35:
```suggestion
        payload : dict
```

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+            payload for requests. payload includes
+              name: object name
+              instrument_request: instrument name
+              ra: right ascension in decimal degrees
+              dec: declination in decimal degrees
+              filters: comma delimited list of filters
+              exposure_time: exposure time requested in seconds
+              exposure_counts: number of exposures requested
+              username: SkyPortal username of requester
```
stefanv, line 44: missing `observation_choices`

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class SLACKAPI(FollowUpAPI):
+
+    """An interface to SLACK operations."""
+
+    # subclasses *must* implement the method below
+    @staticmethod
+    def submit(request):
+
```
stefanv, line 98:
```suggestion
```

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+            f"Exposure time: {target['exposure_time']}\n"
+            f"Exposure counts: {target['exposure_counts']}\n"
+            f"Username: {target['username']}"
+        )
+
+        return target, textwrap.dedent(text)
+
+
+class SLACKAPI(FollowUpAPI):
+
```
stefanv, line 92:
```suggestion
```

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+    # subclasses *must* implement the method below
+    @staticmethod
+    def submit(request):
+
+        """Submit a follow-up request to SLACK.
+
+        Parameters
+        ----------
+        request: skyportal.models.FollowupRequest
```
stefanv, line 103:
```suggestion
        request : skyportal.models.FollowupRequest
```

skyportal/facility_apis/slack.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+        data = json.dumps(
+            {
+                "url": f"{SLACK_URL}/{altdata['slack_workspace']}/{altdata['slack_channel']}/{altdata['slack_token']}",
+                "text": requesttext,
+            }
+        )
+
+        r = requests.post(
```
stefanv, line 128: Since these are used in `followup_requests.py` in `post`, the `submit` call has to be made async and awaited, otherwise this will block our handlers.

Your PR is not the one to introduce the issue, so I will file an issue and we can fix it separately.

stefanv (commented):
Great work, thanks @mcoughlin.

Approved, pending minor fixes to the docstrings (positions of colons, mainly).  I left a few other nitpicky comments, and you are welcome to use or lose as you choose.

The note about async is a bit more dire, but not your fault so we (I?) can address in another PR.

stefanv:
See also: https://github.com/skyportal/skyportal/issues/2594

skyportal/facility_apis/slack.py
mcoughlin: It's returned as filters here, but called observation_choices in the code to be consistent with the other APIs. Down the line this will probably evolve if we have spectrographs using this (which have IFU's, and therefore the more generic term makes sense).

config.yaml.defaults
```diff
@@ -433,6 +433,8 @@ image_grayscale_params:
 slack:
   microservice_port: 64100
   expected_url_preamble: "https://hooks.slack.com/"
+  followup_requests:
```
stefanv, line 436: Could you point out to me where this entry is used?

mcoughlin: You are totally right, it works without this.

PR #2551, opened 2022-01-12
fix migrations for the baselayer=556e74 pin

profjsb:
pinning the latest baselayer requires a change to a number of tests which were looking for a 400 status return upon access error but now require 401. To fix the migrations, we also need a migration for KAIT and SEDMV2 in the `followup_apis` enum.


stefanv:
I think disabling peer dependencies to be installed automatically is dangerous.  Perhaps better to check the relevant peer dependencies with:

```
npm info name-of-module peerDependencies
```

and pin them explicitly?

profjsb:
> I think disabling peer dependencies to be installed automatically is dangerous. Perhaps better to check the relevant peer dependencies with:
> 
> ```
> npm info name-of-module peerDependencies
> ```
> 
> and pin them explicitly?

The problem is (now) that datapicker went from requiring to 16.* to ^17.0.2 which in conflict with 17.0.0 needed to solve the other mui/core issues (which have been fixed with the 17.0.2->17.0.0 downgrade)

stefanv:
For the migration tests, I agree that you can do whatever you like (JS dependencies not important at all).  Maybe leave a comment to that effect.

package.json
```diff
@@ -29,7 +29,7 @@
     "dygraphs": "2.1.0",
     "emoji-dictionary": "1.0.11",
     "emojilib": "3.0.4",
-    "exports-loader": "3.1.0",
+    "exports-loader": "1.1.1",
```
profjsb, line 32: the 3.1.0 caused conflicts in the migration scripts. We recently updated it from 1.1.1 and seems to work for the other tests on the front end. So...reverting it back to 1.1.1

package.json
```diff
@@ -96,11 +96,11 @@
     "eslint-import-resolver-webpack": "0.12.2",
     "eslint-loader": "4.0.2",
     "eslint-plugin-import": "2.25.3",
-    "eslint-plugin-jsx-a11y": "6.3.1",
+    "eslint-plugin-jsx-a11y": "6.4.1",
     "eslint-plugin-react": "7.27.1",
     "eslint-plugin-react-hooks": "4.2.0",
     "json-loader": "0.5.7",
-    "webpack": "5.64.4",
+    "webpack": "5.65.0",
```
profjsb, line 103: minor update

skyportal/tests/__init__.py
```diff
@@ -58,10 +58,8 @@ def api(
...
-                return response.status_code, None
-            else:
-                return response.status_code, response.json()
-        else:
-            return response.status_code, None
+        try:
+            data = response.json()
+        except requests.exceptions.JSONDecodeError:
+            data = None
+        return response.status_code, data
```
profjsb, line 65: Stefan did this to facilitate the error handling in the test APIs

skyportal/tests/api/test_annotations_on_spectrum.py
```diff
@@ -48,7 +48,7 @@ def test_add_and_retrieve_annotation_group_id(
         token=annotation_token,
     )
 
-    assert status == 400
+    assert status in [400, 401]
```
guynir42, line 51: Why 400 and not 401?

skyportal/tests/api/test_annotations_on_spectrum.py
```diff
@@ -79,7 +79,7 @@ def test_add_and_retrieve_annotation_group_id(
         token=annotation_token,
     )
 
-    assert status == 400
+    assert status in [500, 401]
     assert 'duplicate key value violates unique constraint' in data["message"]
```
guynir42, line 83: 500 is when system is revisioning? Would the next assert fail with a different `data['message']` in that case?

skyportal/tests/api/test_annotations_on_spectrum.py
guynir42: Nevermind I see both options.

skyportal/tests/api/test_annotations_on_spectrum.py
stefanv: We should probably get rid of the 500.

skyportal/tests/api/test_annotations_on_spectrum.py
stefanv: This should be one or the other, I suppose.

PR #2552, opened 2022-01-12
Pin mui-datatables 3.8.2; react 17.0.0

profjsb:
Bump mui-datatables to 3.8.2 and downgrade react to 17.0.0 (from 17.0.2) to accommodate the react core failures during migration. Fix the tests that fail because of the update.

replaces #2492


.github/workflows/test_api_and_frontend.yaml
```diff
@@ -189,3 +189,10 @@ jobs:
         with:
           name: test-results
           path: test-results
+
+      - name: Upload generated plots
+        uses: actions/upload-artifact@v2
+        if: ${{ always() }}
+        with:
+          name: generated-plots
+          path: skyportal/tests/data/*_expected.png
```
profjsb, line 198: Save the generated plots for us to debug (these are tiny png's so should not be an issue storage wise)

skyportal/tests/frontend/sources_and_followup_etc/test_group_sources.py
```diff
@@ -245,12 +245,8 @@ def test_sources_sorting(
...
 
     # Now sort by date saved desc by clicking the header twice
-    driver.click_xpath(
-        "//span[contains(@data-testid, 'headcol-')]//div[text()='Date Saved']"
-    )
-    driver.click_xpath(
-        "//span[contains(@data-testid, 'headcol-')]//div[text()='Date Saved']"
-    )
+    driver.click_xpath("//*[text()='Date Saved']")
+    driver.click_xpath("//*[text()='Date Saved']")
```
profjsb, line 249: somehow we lose the data-testid names on the columns, so find them otherwise.

stefanv:
:rocket:

PR #2553, opened 2022-01-12
Return JSON data whenever available from API utility function

stefanv:
Before, we only returned data for HTTP 200 or 400, but there's no
reason to make those special.  In fact, since we recently started
using other error codes like 401, the API function misbehaves in those
cases.


profjsb:
These changes will be merged via #2551

PR #2554, opened 2022-01-13
add check for nan in spectrum smoothing

guynir42:


static/js/plotjs/smooth_spectra.js
```diff
@@ -12,8 +12,12 @@ const smoothing_func = (values, window_size) => {
     const idx_high = i + over < values.length ? i + over : values.length - 1;
     let N = 0;
     for (let j = idx_low; j <= idx_high; j += 1) {
-      N += 1;
-      output[i] += values[j];
+      // check value is not NaN
+      /* eslint no-self-compare: "off" */
+      if (values[j] === values[j]) {
```
profjsb, line 17: use isNaN? https://stackoverflow.com/questions/2652319/how-do-you-check-that-a-number-is-nan-in-javascript

Should also add a test which includes NaN in the input spectrum to see if you get the behavior you're looking for.

profjsb: See also https://github.com/profjsb/skyportal/blob/4b81f1537b526d45d935a67aae49af05d7552b1f/static/js/components/Periodogram.jsx#L106

guynir42: Check the behavior is tricky because the output is a bokeh plot.

profjsb:
Have you tried manually uploading a spectrum with a NaN to see if the plotting works as expected?

guynir42:
> Have you tried manually uploading a spectrum with a NaN to see if the plotting works as expected?

@profjsb I did a manual check and indeed found a bug. Now the behavior seems ok. 

Should I make a separate issue and PR for testing the bokeh output?

guynir42:
The test failure seems to be unrelated. Can we merge this @profjsb ?

profjsb (approved):

PR #2560, opened 2022-01-14
Let database catch duplicate token

stefanv:


profjsb:
But there are likely several million images. Deleting and then pulling from (remote) Kowalski might be very expensive. I think it's worth a 'mv' and 'update DB' loop

mcoughlin:
Given the age of this PR and that things seem to run OK as is, I will close this for now. Feel free to reopen as needed.

PR #2566, opened 2022-01-16
pin sqla_1.4 from baselayer; fix queries

profjsb:


dannygoldstein (commented):
we should use `safe_aliased` here:

https://github.com/skyportal/skyportal/blob/master/skyportal/models/group.py#L29

profjsb:
> we should use `safe_aliased` here:
> 
> https://github.com/skyportal/skyportal/blob/master/skyportal/models/group.py#L29

Done 5071533

skyportal/handlers/api/photometry.py
```diff
@@ -398,27 +397,6 @@ def get_values_table_and_condition(self, df):
...
-                for c in self._column_args:
-                    c._make_proxy(self)  # noqa
-
-            @property
-            def _from_objects(self):
-                return [self]
-
-        # https://github.com/sqlalchemy/sqlalchemy/wiki/PGValues
-        @compiles(_photometry_values)
         def _compile_photometry_values(element, compiler, asfrom=False, **kw):
```
mcoughlin, line 400: Is this actually used anywhere now either?

mcoughlin (approved):
Just one question from the compile stuff, otherwise can merge that branch.

skyportal/handlers/api/photometry.py
profjsb: Good point. Removing.

profjsb:
failure unrelated.

PR #2570, opened 2022-01-18
Fix slugify & metadata= warnings from SA

profjsb:
Fix SAWarning about description needing to be in metadata. 
Suppress warnings on slugify on test fixtures.


mcoughlin:
@profjsb seems great, just a few lint issues looks like.

mcoughlin (approved):
@stefanv you want this merged in the branch or look at it here?

alembic/versions/ea22cd702690_allocations_metadata_type.py
```diff
@@ -0,0 +1,41 @@
...
+# revision identifiers, used by Alembic.
+revision = 'ea22cd702690'
+down_revision = 'f9611048498a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
```
stefanv, line 20: Remove

skyportal/tests/conftest.py
```diff
@@ -54,31 +56,36 @@
...
 # there if running the test server and running tests individually)
-if (
-    not DBSession()
-    .execute(sa.select(User).filter(User.username == "test factory"))
-    .scalar()
-):
-    DBSession.add(User(username="test factory"))
-    DBSession.commit()
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore", category=SAWarning)
```
stefanv, line 64: Why are we suppressing SAWarnings here?

stefanv (commented):
This looks good to me.  The only question I have is why the warnings are suppressed in the test suite.

skyportal/tests/conftest.py
profjsb: Uncaught, this was leading to the error:
```
 skyportal/tests/conftest.py:61: SAWarning: TypeDecorator SlugifiedStr() will not produce a cache key because the ``cache_ok`` attribute is not set to True.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Set this attribute to True if this type object's state is safe to use in a cache key, or False to disable this warning. (Background on this error at: https://sqlalche.me/e/14/cprf)
```
it is not clear how to set `cache_ok` on this query (or even if we want to) so I'm opting to suppress the warning which does not seem to be a real issue (since we're getting the same results as pre sqla 1.4 and the test timing seems similar)

alembic/versions/ea22cd702690_allocations_metadata_type.py
profjsb, line 20:
```suggestion
```

skyportal/tests/conftest.py
stefanv: After scouring the docs and reading [this thread](https://groups.google.com/g/sqlalchemy/c/zCPZKTxM6b0) I am convinced that setting `cache_ok` to True is fine.  Setting it to False is also OK, but less optimal.  Better than these warning workarounds, I think.

profjsb: Can you propose a fix? I don't know where in this call to set the flag.

skyportal/models/spectrum.py
```diff
@@ -32,6 +32,10 @@
 class NumpyArray(sa.types.TypeDecorator):
     """SQLAlchemy representation of a NumPy array."""
 
+    # do not use a cache here, for safety
+    # c.f. https://sqlalche.me/e/14/cprf
+    cache_ok = False
```
stefanv, line 37: Not sure why this is set to False here either.

PR #2571, opened 2022-01-18
SQLAlchemy 1.4.X compat version

profjsb:
This PR upgrades SQLAlchemy to 1.4.* from 1.3. It pins a branch of `baselayer` that has `requirements.txt` explicitly stating the dependency of `sqlalchemy>=1.4.0'. This move has a number of breaking changes. In addition to the original PR, we may opt to merge #2570 into this branch.

* Pin BL that requires SQLA v>=1.4.0

* Use SQLA 1.4-compatible registry mappers in schema

* Refactor Values construct in photometry handler

* Use select API instead of query in PhotometryHandler, PublicGroupHandler, conftest & fixtures

* use a version that eralchemy which works with SQLA1.4

* fix AttributeError on Base in schema.py

* safe alias the group logic

* fix candidate.py

* use VALUES constructor in get_obj_id_values

* use 1.4 func.jsonb_each instead of <1.4 formula

* remove legacy code

Co-authored-by: Ari Crellin-Quick <a.crellinquick@gmail.com>
Co-authored-by: Michael Coughlin <michael.w.coughlin@gmail.com>

### DIFF
--- a/requirements.docs.txt
+++ b/requirements.docs.txt
@@ -4,4 +4,5 @@ numpydoc
 sphinx-rtd-theme
 notedown
 jsx-lexer
-eralchemy==1.2.10
+# need a fix for SQLA1.4
+git+https://github.com/sturzaam/eralchemy@Sturzaam-patch-1#egg=eralchemy
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -9,11 +9,11 @@ import numpy as np
 
 from tornado.ioloop import IOLoop
 
+import sqlalchemy as sa
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm import sessionmaker, scoped_session
-from sqlalchemy.sql.expression import case, func, FromClause
-from sqlalchemy.ext.compiler import compiles
-from sqlalchemy.sql import column
+from sqlalchemy.sql.expression import case, func
+from sqlalchemy.sql import column, Values
 from sqlalchemy.types import Float, Boolean, String, Integer
 from marshmallow.exceptions import ValidationError
 
@@ -822,15 +822,16 @@ class CandidateHandler(BaseHandler):
             if "Page number out of range" in str(e):
                 return self.error("Page number out of range.")
             raise
+
         matching_source_ids = (
             Source.query_records_accessible_by(
                 self.current_user, columns=[Source.obj_id]
             )
-            .filter(Source.obj_id.in_([obj.id for obj in query_results["candidates"]]))
+            .filter(Source.obj_id.in_([obj.id for obj, in query_results["candidates"]]))
             .all()
         )
         candidate_list = []
-        for obj in query_results["candidates"]:
+        for (obj,) in query_results["candidates"]:
             with DBSession().no_autoflush:
                 obj.is_source = (obj.id,) in matching_source_ids
                 if obj.is_source:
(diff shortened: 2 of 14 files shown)


requirements.docs.txt
```diff
@@ -4,4 +4,5 @@ numpydoc
 sphinx-rtd-theme
 notedown
 jsx-lexer
-eralchemy==1.2.10
+# need a fix for SQLA1.4
```
profjsb, line 7: this codebase hasn't been update in many years. This PR fixes the issues introduced in SQLA1.4.

skyportal/handlers/api/internal/annotations_info.py
```diff
@@ -104,9 +32,7 @@ def get(self):
         # filters to apply on the auto-annotations column on the scanning page.
         # For example, if given that an annotation field is numeric we should
         # have min/max fields on the form.
-        annotations = jsonb_each_func(
-            Annotation.data
-        )  # Get each key/value tuple in annotations
+        annotations = func.jsonb_each(Annotation.data).table_valued("key", "value")
```
profjsb, line 35: dozens of lines of old code get turned into this one 🙌

profjsb:
unrelated flaky test failures are fixed in #2570

skyportal/handlers/api/candidate.py
```diff
@@ -1092,52 +1093,22 @@ def get_obj_id_values(obj_ids):
     values_table: `sqlalchemy.sql.expression.FromClause`
         The VALUES representation of the Obj IDs list.
     """
-    # https://github.com/sqlalchemy/sqlalchemy/wiki/PGValues
```
stefanv, line 1095: Nice.

skyportal/tests/api/test_photometry.py
```diff
@@ -478,13 +478,21 @@ def test_retrieve_photometry_error_group_membership_posted_by_other(
         },
         token=upload_data_token,
     )
+    print(
```
stefanv, line 481: Are these debug calls, or useful in the logs?

stefanv (approved):
Fantastic, thanks for all the hard work on this.  I did not scour the PR line by line—it looks to me like the gist is correct, and we should get it in and polish up whatever needs to after.

skyportal/tests/api/test_photometry.py
profjsb: Probably useful in the logs?  But I'll comment them out since we do not need print statements in the test suite as a matter of practice.

stefanv: I don't think it's a big deal to have them there, as long as they provide generally useful information---go with whatever you think is best.

stefanv:
Reminder to remove the warning filters, then we should be good to go.

profjsb:
> Reminder to remove the warning filters, then we should be good to go.

That's done->undone in the other PR (https://github.com/skyportal/skyportal/pull/2570/commits/6ff7724c22fdfbd11562e76fdf39fe9fa2e7c48f)

stefanv:
Only a single failure: `skyportal/tests/frontend/sources_and_followup_etc/test_sources.py:665 in test_obj_page_unsaved_source`

mcoughlin:
@stefanv It passes locally and passed on rerun, might be a bit of a flaky test?

guynir42 (approved):
Looks good!

PR #2574, opened 2022-01-20
GCN comments table

guynir42:
Added a `CommentsOnGCN` model in `models/comments.py`. 

It inherits from `CommentMixin` the same way `CommentOnSpectrum` does. 

The main difference is that this model is not associated at all with any `Obj`. 

The API and frontend elements will be added after this PR is merged.


mcoughlin:
@guynir42 I am super stoked for this.

mcoughlin (approved):
Looks really great. Thanks for factoring out the Obj stuff. Since it seems like there might be a few more tables that have comments to add down the line, might be nice to make a few notes at the top of comment.py to note the couple of places where you need to add things?

alembic/versions/86504d0cab3f_add_gcn_comment_table.py
```diff
@@ -0,0 +1,107 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        'comments_on_gcn',
+        sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column('created_at', sa.DateTime(), nullable=False),
```
profjsb, line 23: should this default to `now`?

skyportal/models/comment.py
```diff
@@ -41,6 +41,8 @@ def backref_name(cls):
             return "comments"
         if cls.__name__ == 'CommentOnSpectrum':
             return 'comments_on_spectra'
+        if cls.__name__ == 'CommentOnGCN':
+            return 'comments_on_gcn'
```
profjsb, line 45: inconsistent naming here. On spectra its <plural>_on_<plural> whereas with GCN its <plural>_on_<singular>. You might change it to `comments_on_gcns`

skyportal/models/comment.py
```diff
@@ -70,13 +72,13 @@ def obj_id(cls):
...
-            back_populates=cls.backref_name(),
-            doc="The Comment's Obj.",
-        )
+    # @declared_attr
+    # def obj(cls):
+    #     return relationship(
+    #         'Obj',
+    #         back_populates=cls.backref_name(),
+    #         doc="The Comment's Obj.",
+    #     )
```
profjsb, line 82:
```suggestion

```

profjsb (changes requested):

alembic/versions/86504d0cab3f_add_gcn_comment_table.py
guynir42: I think this is handled in baselayer. I don't see a default applied to any other migrations so I guess it doesn't need to happen in the migration script.

guynir42:
Test failure is unrelated.

skyportal/models/comment.py
```diff
@@ -13,6 +13,22 @@
...
+- Add the comment as a relationship with back_populates (etc.) on the model you are commenting on.
+  (e.g., for CommentOnSpectrum you need to add "comments" to models/spectrum.py)
+- Add a join to models/group_joins.py so comments will have groups associated with them.
+- Add a join to models/user_token.py so comments will have a user associated with them.
+- Update the API endpoints for comments, and the reducers to listen for changes in the comments.
+- Update the app_server.py paths to accept the new type of comment upon API calls.
+
+"""
+
+
```
profjsb, line 31: This is good instruction for the future. But it should be moved to the development docs (`doc/adding_comment_types.md`?), not as a docstring.

profjsb (changes requested):
Looks good. Should be add testing of the permissions for this table (see `skyportal/tests/models/test_permissions.py`)

skyportal/models/comment.py
guynir42: I think that having it also as a docstring would be more useful to future developers, that sometimes (only very rarely) don't read the docs but do look at the code. Maybe we can have this in both places?

profjsb:
@guynir42 can you fix failing tests or confirm they are unrelated?

guynir42:
@profjsb Now I remember what was blocking this PR: I was waiting to hear from Danny about the `test_permissions` script. I still don't know how to generate that script correctly. 

I suggest we put a pin in this, and add the new models to the test suite in a different PR. This isn't the only model that we are not testing against and it's blocking the entire comments on GCN line.

profjsb:
Ok--can you make an issue and reference it here in the comments so we don't lose track of the need for such test? @guynir42 

After that, you can merge.

profjsb (approved):

PR #2575, opened 2022-01-21
Add LocalizationTile table to determine which objects are inside localization

mcoughlin:
This PR adds a `LocalizationTile` table (with associated healpix id and probability densities) to determine which objects are inside skymaps. It also adds healpix to the Obj table. 

This PR exposes [GeoJSON](https://geojson.org/) for use in visualization tooling. The form of the `["data"]["geojson"]` packet returned when on a GET to `galaxy_catalog` is a list of `dict`:
```python
[
 {'geometry': {'coordinates': [0.60321, -54.76653], 'type': 'Point'},
  'properties': {'name': '2MASX J00022478-5445592'},
  'type': 'Feature'},
 {'geometry': {'coordinates': [0.948375, -54.574083], 'type': 'Point'},
  'properties': {'name': 'IRAS F00012-5451 ID'},
  'type': 'Feature'}
]
```

image


stefanv:
Took a quick look and this looks good to me.

skyportal/handlers/api/galaxy.py
```diff
@@ -121,12 +123,98 @@ def get(self, catalog_name=None):
...
+
         query = Galaxy.query_records_accessible_by(self.current_user, mode="read")
         if catalog_name is not None:
             query = query.filter(Galaxy.catalog_name == catalog_name)
+
+        if localization_dateobs is not None:
+            if localization_name is not None:
+                localization = (
+                    Localization.query_records_accessible_by(self.current_user)
+                    .filter(Localization.dateobs == localization_dateobs)
```
profjsb, line 139: unclear is this is going to work if localization_dateobs is not in the exact string format of dateobs. Would it be safer to coerce this API input with a date parser then recast it to a string of the proper form?

skyportal/handlers/api/galaxy.py
```diff
@@ -121,12 +123,98 @@ def get(self, catalog_name=None):
         """
 
         catalog_name = self.get_query_argument("catalog_name", None)
+        localization_dateobs = self.get_query_argument("localizationDateobs", None)
+        localization_name = self.get_query_argument("localizationName", None)
+        localization_cumprob = self.get_query_argument("localizationCumprob", 0.95)
+        includeGeojson = self.get_query_argument("includeGeojson", False)
```
profjsb, line 129: none of these variable are described in the docstring for this get method.

skyportal/handlers/api/galaxy.py
```diff
@@ -121,12 +123,98 @@ def get(self, catalog_name=None):
...
+
         query = Galaxy.query_records_accessible_by(self.current_user, mode="read")
         if catalog_name is not None:
             query = query.filter(Galaxy.catalog_name == catalog_name)
+
+        if localization_dateobs is not None:
```
profjsb, line 135: what should be returned if `localization_dateobs` is None? I dont see the logic for that.

skyportal/handlers/api/instrument.py
```diff
@@ -271,6 +272,7 @@ def delete(self, instrument_id):
 
 def add_tiles(instrument_id, instrument_name, regions, field_data):
     session = Session()
+    # if True:
```
profjsb, line 275:
```suggestion
```

skyportal/handlers/api/instrument.py
```diff
@@ -301,6 +303,19 @@ def add_tiles(instrument_id, instrument_name, regions, field_data):
...
+            geometry = []
+            for coord in coords:
+                tab = []
+                for i in range(len(coord.ra.deg)):
+                    tab.append([coord.ra.deg[i], coord.dec.deg[i]])
+                tab.append([coord.ra.deg[0], coord.dec.deg[0]])
+                geometry.append(tab)
+
+            # contour['type'] = 'FeatureCollection'
+            # contour['geometry'] = {'coordinates':coordinates,'type': 'MultiLineString'}
```
profjsb, line 315:
```suggestion
```

skyportal/handlers/api/source.py
```diff
@@ -608,6 +614,11 @@ def get(self, obj_id=None):
             "createdOrModifiedAfter", None
         )
 
+        localization_dateobs = self.get_query_argument("localizationDateobs", None)
+        localization_name = self.get_query_argument("localizationName", None)
+        localization_cumprob = self.get_query_argument("localizationCumprob", 0.95)
+        includeGeojson = self.get_query_argument("includeGeojson", False)
```
profjsb, line 620: missing docstring descriptions.

skyportal/handlers/api/source.py
```diff
@@ -1130,6 +1142,65 @@ class Validator(Schema):
                     isouter=True,
                 )
 
+        if localization_dateobs is not None:
```
profjsb, line 1145: same question about logic when this is None...

profjsb (changes requested):
Looking good. A few suggestions:
 - as there's a front-end piece of this, please include a screenshot of what this should look like
 - I dont see any tests for the new query parameter functionality like localization_name and getgeojson.
 - other comments/suggestions in line

alembic/versions/d61a8bea7313_crossmatch.py
```diff
@@ -0,0 +1,68 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        'localizationtiles',
+        sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column('created_at', sa.DateTime(), nullable=False),
```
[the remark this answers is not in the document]
mcoughlin: None of our other tables seems to do that, but I am not opposed to adding it of course.

skyportal/handlers/api/galaxy.py
```diff
@@ -1,14 +1,16 @@
 from baselayer.app.access import permissions, auth_or_token
 from baselayer.log import make_log
 from tornado.ioloop import IOLoop
+from geojson import Point, Feature
 
+import sqlalchemy as sa
 from sqlalchemy.orm import sessionmaker, scoped_session
 
 import astropy.units as u
```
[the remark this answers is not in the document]
mcoughlin: Ah yes of course.

skyportal/handlers/api/galaxy.py
mcoughlin: Currently it just returns the original query, which is the catalog (i.e. the current behavior of the endpoint).

skyportal/handlers/api/galaxy.py
mcoughlin: It's really fundamentally a string here, and really is about checking if the event exists or not, i.e. the logic about checking that the event exists below. So that seems more the crux of it.

skyportal/handlers/api/source.py
mcoughlin: Same deal here, just the original query.

skyportal/handlers/api/source.py
mcoughlin: ah yes. added.

skyportal/handlers/api/galaxy.py
mcoughlin: added.

mcoughlin:
image

Adds a table of sources within the localization.

profjsb (approved):
LGTM. Thanks for working on the initial set of comments.

stefanv:
Wow, that PR title rename had me searching for a long time :)

stefanv:
:clap: Thank you for that PR description, Michael!

profjsb:
> Wow, that PR title rename had me searching for a long time :)

I insisted upon a name change ... sorry

skyportal/handlers/api/galaxy.py
```diff
@@ -104,11 +105,37 @@ def get(self, catalog_name=None):
...
+              name: catalog_name
               schema:
                 type: string
               description: Filter by catalog name (exact match)
+            - in: query
+              name: localizationDateobs
+              schema:
+                type: string
+              description: |
+                Event time in isot format
```
stefanv, line 118:
```suggestion
                Event time in ISO 8601 format (`YYYY-MM-DDTHH:MM:SS.sss`).
```

What timezone is this, or is that not relevant?

skyportal/handlers/api/galaxy.py
```diff
@@ -104,11 +105,37 @@ def get(self, catalog_name=None):
...
+            - in: query
+              name: localizationName
+              schema:
+                type: string
+              description: |
+                Name of localization / skymap to use
```
stefanv, line 124: Guidance for the user on where to find these, or what they can be?

skyportal/handlers/api/galaxy.py
```diff
@@ -104,11 +105,37 @@ def get(self, catalog_name=None):
...
+            - in: query
+              name: localizationCumprob
+              schema:
+                type: number
+              description: |
+                Cumulative probability up to which to include galaxies
+            - in: query
+              name: includeGeojson
```
stefanv, line 132:
```suggestion
              name: includeGeoJSON
```

skyportal/handlers/api/galaxy.py
```diff
@@ -104,11 +105,37 @@ def get(self, catalog_name=None):
...
+              nullable: true
+              schema:
+                type: boolean
+              description: |
+                Boolean indicating whether to include associated geojson. Defaults to
```
stefanv, line 137:
```suggestion
                Boolean indicating whether to include associated GeoJSON. Defaults to
```

mcoughlin:
@stefanv @profjsb made it way better.

skyportal/handlers/api/galaxy.py
```diff
@@ -121,12 +148,99 @@ def get(self, catalog_name=None):
...
+                    )
+
+            cum_prob = (
+                sa.func.sum(
+                    LocalizationTile.probdensity * LocalizationTile.healpix.area
+                )
+                .over(order_by=LocalizationTile.probdensity.desc())
+                .label('cum_prob')
+            )
+            subquery1 = (
```
stefanv, line 193: `subquery_<description>` vs `subquery1` woul dbe helpful for understanding what it does.  Also, below the subqueries have full names.

skyportal/handlers/api/galaxy.py
```diff
@@ -121,12 +148,99 @@ def get(self, catalog_name=None):
...
         galaxies = query.all()
+        query_results = {'sources': galaxies}
+
+        if includeGeojson:
+            # features are JSON representations that the d3 stuff understands.
+            # We use these to render the contours of the sky localization and
+            # locations of the transients.
+
+            features = []
+            for source in query_results["sources"]:
```
stefanv, line 229:
```suggestion
            for source in galaxies:
```

skyportal/handlers/api/gcn.py
```diff
@@ -1,7 +1,6 @@
-# Inspired by https://github.com/growth-astro/growth-too-marshal/blob/main/growth/too/gcn.py
+# inspired by https://github.com/growth-astro/growth-too-marshal/blob/main/growth/too/gcn.py
```
stefanv, line 1:
```suggestion
# Inspired by https://github.com/growth-astro/growth-too-marshal/blob/main/growth/too/gcn.py
```

Presumably this was an unintended edit.

skyportal/handlers/api/gcn.py
```diff
@@ -20,12 +20,16 @@
...
 
 log = make_log('api/gcn_event')
 
 
+Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
+
+
 class GcnEventHandler(BaseHandler):
     @auth_or_token
     def post(self):
```
stefanv, line 35: Note: below it does not look as though it is checked whether `xml` is present in the payload or not, so may result in error.

stefanv: Also, raises a bare exception, instead of using `self.error`

stefanv: Curious how duplicate tags are handled in this handler.

skyportal/handlers/api/gcn.py
```diff
@@ -118,14 +122,10 @@ def post(self):
...
-            contour_func = functools.partial(
-                add_contour,
-                localization.id,
-                self,
-            )
-            IOLoop.current().run_in_executor(None, contour_func)
+            log(f"Generating tiles/contours for localization {localization.id}")
 
-        self.verify_and_commit()
+            IOLoop.current().run_in_executor(None, lambda: add_tiles(localization.id))
```
stefanv, line 127: Question: is there no permission checking surrounding adding of tiles and contours?

skyportal/handlers/api/gcn.py
```diff
@@ -221,24 +221,44 @@ def delete(self, dateobs):
...
 
-def add_contour(localization_id, request_handler):
+def add_tiles(localization_id):
+    session = Session()
     try:
-        localization = (
-            Localization.query_records_accessible_by(request_handler.current_user)
-            .filter(
-                Localization.id == localization_id,
+        localization = session.query(Localization).get(localization_id)
```
stefanv, line 227: Permission checking removed; is this safe?

skyportal/handlers/api/instrument.py
```diff
@@ -301,6 +302,17 @@ def add_tiles(instrument_id, instrument_name, regions, field_data):
                 },
             }
 
+            geometry = []
+            for coord in coords:
+                tab = []
+                for i in range(len(coord.ra.deg)):
+                    tab.append([coord.ra.deg[i], coord.dec.deg[i]])
+                tab.append([coord.ra.deg[0], coord.dec.deg[0]])
```
stefanv, line 310: Maybe something like:

```
tab = zip((*coord.ra.deg, coord.ra.deg[0]), (*coord.dec.deg, coord.dec.deg[0]))
```

skyportal/handlers/api/instrument.py
```diff
@@ -301,6 +302,17 @@ def add_tiles(instrument_id, instrument_name, regions, field_data):
...
+                geometry.append(tab)
+
+            contour['type'] = "Feature"
+            contour['geometry'] = {'coordinates': geometry, 'type': 'MultiLineString'}
+            contour = {'features': [contour], 'type': "FeatureCollection"}
```
stefanv, line 315: This is a bit of a confusing statement.  Perhaps rewrite in its entirety as:

```python
contour = {
    'type': 'FeatureCollection',
    'features': [
        {
            'type': 'Feature',
            'geometry': {
                'type': 'MultiLineString'
                'coordinates': geometry,
            },
        },
    ]
}
```

If I unpacked that correctly!  But you can see that it's quite subtle to unpack it in your head, but once unpacked it's quite easy to follow.

skyportal/handlers/api/source.py
```diff
@@ -534,6 +540,32 @@ def get(self, obj_id=None):
             description: |
               Arrow-parseable date-time string (e.g. 2020-01-01 or 2020-01-01T00:00:00 or 2020-01-01T00:00:00+00:00).
               If provided, filter by created_at or modified > createdOrModifiedAfter
+          - in: query
+            name: localizationDateobs
+            schema:
+              type: string
+            description: |
+              Event time in isot format
```
stefanv, line 548: Same comment as above

skyportal/handlers/api/source.py
```diff
@@ -534,6 +540,32 @@ def get(self, obj_id=None):
...
+            description: |
+              Cumulative probability up to which to include galaxies
+          - in: query
+            name: includeGeojson
+            nullable: true
+            schema:
+              type: boolean
+            description: |
+              Boolean indicating whether to include associated geojson. Defaults to
+              false.
```
stefanv, line 568: Why are we extending the source query to include this, instead of making it its own endpoint?  We try to keep the sources relatively lightweight (we've already somewhat failed at that), but it would be good not to make them even heavier.

stefanv: `/api/sources/14xyz/<new_obj_type>` is what we've used for spectra

stefanv (commented):
This PR implements several features; is it absolutely necessary for those to be combined into one?  It makes for a lot of code to review all at once.

If it's going to be too hairy to pull apart, I'm okay going ahead with this, but let's aim for atomic changes in the future.

Main concerns here:
- Docstrings are likely not descriptive enough to help users
- Source API is being expanded even further, when hopefully we can move out these queries to their own endpoint

skyportal/handlers/api/galaxy.py
mcoughlin: UTC is implied.

skyportal/handlers/api/gcn.py
mcoughlin: This is just adding the Localization tiles, so you would only be doing this on your own localization. Is there an equivalent session query with the accessible bit?

skyportal/handlers/api/instrument.py
mcoughlin: ah fancy.

skyportal/handlers/api/instrument.py
mcoughlin: yes much better.

skyportal/handlers/api/galaxy.py
mcoughlin: Added.

skyportal/handlers/api/galaxy.py
mcoughlin: yes agreed.

skyportal/handlers/api/source.py
mcoughlin: changed same as above.

skyportal/handlers/api/source.py
mcoughlin: but we need all of this other handling for the parameters right?

skyportal/handlers/api/gcn.py
mcoughlin: I am not sure why we would have ownership of those.

skyportal/handlers/api/gcn.py
mcoughlin: What do you suggest for the tag stuff?

skyportal/handlers/api/source.py
stefanv: The parameter handling will be needed, yes.

stefanv: But should it be part of the sources endpoint is what I'm not clear about.

skyportal/handlers/api/galaxy.py
```diff
@@ -115,26 +115,26 @@ def get(self, catalog_name=None):
...
               description: |
-                Event time in isot format
+                Event time in ISO 8601 format (`YYYY-MM-DDTHH:MM:SS.sss`).
             - in: query
               name: localizationName
               schema:
                 type: string
               description: |
-                Name of localization / skymap to use
+                Name of localization / skymap to use. Can be found in Localization.localization_name
```
stefanv, line 124: What does `Localization.localization_name` mean to users?  I guess what I'm asking with the docstrings is to use text that makes it obvious for newcomers to the API on what to do to obtain this value, or how to use it.

stefanv:
Thanks, Michael, this looks good.  Main question for me is whether this should be tacked onto the source handler, or become its own.  I don't have enough perspective to answer that question; @profjsb @guynir42 ?

skyportal/handlers/api/galaxy.py
mcoughlin: Ah yes. I added more information (the endpoint one could use or the fact that it is in the table on the GcnEvents page).

alembic/versions/d61a8bea7313_crossmatch.py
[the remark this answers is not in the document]
guynir42: I think `created_at` is inherited from the Baselayer `Base` class, which defaults to now...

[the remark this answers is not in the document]
guynir42: Same with `modified` if I remember correctly...

skyportal/handlers/api/source.py
guynir42: I think the only reason to include this in the `source` handler is if you are planning to find "all sources inside a boundary". I'm not sure if there's a more straightforward way to do that in a separate handler...

guynir42 (approved):
I have only minor comments on this. Looks really good. Can't wait to see it in action.

mcoughlin:
@guynir42 So those now() defaults seem ok to me. And yeah, I think we want that filtering power in this handler, but I admittedly don't quite know the effect of factoring it out.

mcoughlin:
@guynir42 given the discussion, do you think we can merge this?

stefanv:
Hold off on merging until we've resolved the source handler complexity, please.

stefanv:
Thanks for the discussion this afternoon.

PR #2580, opened 2022-01-21
Allow alembic to handle multiple configuration files

stefanv:
Depends on https://github.com/cesium-ml/baselayer/pull/277


stefanv:
Should probably pin the BL commit here before merging, once that has been merged.

PR #2584, opened 2022-01-25
DRAFT: Add ObservationPlanRequests

mcoughlin:
This PR differentiates FollowupRequests (associated with Objs) and ObservationPlanRequests (associated with GcnEvents). The API stuff so far has baked in the single object follow-up, and this starts to split them apart.


mcoughlin:
Known issue, the API requests currently just loop instead of running once (maybe @stefanv knows why :):

image

stefanv:
My guess would be:

`fetchAllocations` updates allocationList, a dependency of `useEffect` in `ObservationPlanRequestForm`.  That `useEffect` calls `fetchAllocations` again.

PR #2587, opened 2022-01-25
Add @material-ui/styles peer dependency

stefanv:
Webpack would not build on my machine without it, even after removing
`node_modules`.  I am not sure why it works on CI; it could be that I
have a newer version of node/npm.


stefanv:
Failure unrelated.

profjsb (approved):

PR #2590, opened 2022-01-26
fix redshift of lines 

guynir42:
When loading a source page that has a redshift, the spectral lines are not red-shifted when the plot is shown. The redshift is only applied after changing the redshift in the slider or input box. 

This patch fixes this by adding a couple of missing lines in `plot.py`.


profjsb:
LGTM

PR #2592, opened 2022-01-27
Fix the adding groups to spectrum in put request bug and add a test

### LINKED ISSUE #2583: PUT new group_ids to spectrum
There is a Spectrum to which I want to make available to people who are members of a group.  I run the command

```
curl -X PUT -s -H 'Authorization: token XXX' -d '{"group_ids": "[76]"}' https://desi-skyportal.lbl.gov/api/spectra/76957

{
  "status": "success",
  "data": {},
  "version": "0.9.dev0"
}
```
Looking at the contents of https://desi-skyportal.lbl.gov/api/spectra/76957 there appears to be no change.

People in the group with grouped 76 remain unable to see the spectrum.

Am I doing this incorrectly?

jadalilleboe:
Closes #2583 

This pull request fixes the bug mentioned in the linked issue. Adds group checks to the PUT request of the spectrum API endpoint and edits an existing test to test for adding new groups to spectrums within a PUT request.


stefanv:
Is there a way we can use the same functionality in POST and PUT, by factoring it out into a utility function?

jadalilleboe:
@profjsb @stefanv extracted the group checks into a utility function and fixed the `is not None` style, let me know what you think!

stefanv (approved):
Nice work, @jadalilleboe!

skyportal/handlers/api/spectrum.py
```diff
@@ -36,6 +36,19 @@
 
 
 class SpectrumHandler(BaseHandler):
+    def get_groups(self, group_ids):
+        groups = None
+        if group_ids:
```
profjsb, line 41: Is there ever a case now when group_ids can be None? If not you can simplify

skyportal/handlers/api/spectrum.py
```diff
@@ -80,17 +93,10 @@ def post(self):
...
-                groups = Group.query.filter(
-                    Group.name == cfg['misc.public_group_name']
-                ).all()
-            else:
-                groups = Group.get_if_accessible_by(
-                    group_ids, self.current_user, raise_if_none=True
-                )
+            groups = self.get_groups(group_ids)
 
         if single_user_group not in groups:
```
profjsb, line 101: if groups can possibly be None than this will raise an exception since None is not iterable (but I think the logic above is that groups cannot be None here.

jadalilleboe: Yes, I believe the logic is that if group_ids is None then the groups variable will be a list with only the single user group.

skyportal/handlers/api/spectrum.py
jadalilleboe: Fixed

profjsb:
LGTM but I see that tests are failing. Are they possibly related?

jadalilleboe:
@profjsb I think it's unrelated, not sure though. I didn't change anything with the GCN API

jadalilleboe:
@profjsb ok, I pulled from master and reran the tests, I'm pretty positive that the failing test is unrelated

profjsb (approved):

PR #2595, opened 2022-01-27
Add sensitivity_data column to instrument table

### LINKED ISSUE #2527: Instruments with limiting magnitude and associated exposure time
As part of the follow-up API, having some way of reporting a characteristic limiting magnitude for a given exposure time would be nice to then display when making a follow-up request, computing something like:

            nmag = -2.5*np.log10(np.sqrt(config_struct["exposuretime"]/exposureTime))
            limiting_mag = config_struct["magnitude"] + nmag

So adding this information as optional columns to the Instrument table (setting, for example, ZTF has 20.5 mag in 30 seconds).

jadalilleboe:
Closes #2527 

Adds a sensitivity_data column to the instrument table for filters and their associated limiting magnitude and exposure time.


mcoughlin:
To be slightly more generic, I am wondering if perhaps we do something like a json packet that has the possibility of filter, magnitude, characteristic exposure time, as it will depend on the particular filter what those numbers are. Then these could be displayed on the front-end of the planning page when an instrument is chosen. What do you think @profjsb ?

skyportal/models/instrument.py
```diff
@@ -88,6 +88,10 @@ class Instrument(Base):
         doc='List of filters on the instrument (if any).',
     )
 
+    mag = sa.Column(sa.Float, nullable=True, doc="Characteristic limiting magnitude")
```
profjsb, line 91: maybe call this `limiting_mag`?

alembic/versions/f492e8ed9d84_add_magnitude_and_exposure_time_columns_.py
```diff
@@ -0,0 +1,28 @@
...
+Revises: f9611048498a
+Create Date: 2022-01-27 10:06:33.987883
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = 'f492e8ed9d84'
+down_revision = '82f5a89fbe59'
```
profjsb, line 13: will need to update this down revision hash given that there are new migrations.

profjsb:
Once this is out of Draft mode, please retag me for review.

mcoughlin:
@jadalilleboe @profjsb Maybe we discuss at stand up on Thursday, but something more generic like a JSON that knows about filters and so on would probably be more future proof.

jadalilleboe:
@mcoughlin So are you proposing to get rid of the filters, limiting_mag, and exposure_time columns and combine them all into one JSON column?

mcoughlin:
@jadalilleboe yes (or similar). Can we discuss at stand-up?

profjsb:
sensitivity_data ?

jadalilleboe:
@mcoughlin @profjsb ok I think this is ready for review now

skyportal/handlers/api/instrument.py
```diff
@@ -49,6 +49,15 @@ def post(self):
                 'Invalid/missing parameters: ' f'{exc.normalized_messages()}'
             )
 
+        sensitivity_data = data.pop("sensitivity_data", None)
+        if sensitivity_data:
+            filters = data.get("filters")
```
profjsb, line 54: Will want to make sure filters is always a list
```suggestion
            filters = data.get("filters", [])
```

profjsb (changes requested):
Very nice @jadalilleboe -- just a tiny change then good to go.

mcoughlin (approved):
Great @jadalilleboe. Can you confirm the down_revision is up to date? If so, happy to see this merged.

jadalilleboe:
@mcoughlin yes can confirm it's up to date.

jadalilleboe:
not anymore, will update later

alembic/versions/73bc8919f190_create_sensitivity_data_column.py
```diff
@@ -0,0 +1,35 @@
...
+Create Date: 2022-02-16 18:31:00.739464
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '73bc8919f190'
+down_revision = 'd6b24747693a'
```
profjsb, line 14: check this...may have changed.

skyportal/tests/api/test_instrument.py
```diff
@@ -7,6 +7,54 @@
...
+            ],
+            'telescope_id': telescope_id,
+            'field_data': pd.read_csv(fielddatafile)[:5].to_dict(orient='list'),
+            'field_region': Regions.read(regionsdatafile).serialize(format='ds9'),
+        },
+        token=super_admin_token,
+    )
+    assert status == 400
+    assert data['status'] == 'error'
+
```
profjsb, line 56: check that the error message is what you expect.

profjsb (changes requested):
Looks good. A few more suggestions:
1) try testing when you add sensitivity_data with more than one entry.
2) can you add more sensitivity data to an existing instrument (via PATCH?) or does someone need to delete the entire instrument? I think it's a good idea to allow PATCHing of sensitivity_data. In this respect, it might be better to have `sensitivity_data` be a dict with the key as the (allowed) filter name. 

 ```python
'sensitivity_data': 
                { 'f110w': 
                   {
                    'limiting_magnitude': 20.5,
                    'magsys': 'ab',
                    'exposure_time': 30,
                   }
                }
```
3)  add a bit of info about the expected behaviors/data in the docs?

mcoughlin:
@jadalilleboe Closing in favor of:
https://github.com/skyportal/skyportal/pull/2765
where we will apply these in. We can fix any feedback there.

PR #2603, opened 2022-01-28
Fix disappearing photometry/spectroscopy bug

mcoughlin:
When PS1 cutout arrives, the photometry and spectroscopy disappears. It turns out it is because the photometry option only exist in multiple source request (despite what the docstring says). This adds it back in.

See https://github.com/ZwickyTransientFacility/scope/issues/61

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -823,6 +823,16 @@ class SourceHandler(BaseHandler):
                     key=lambda x: x["created_at"],
                     reverse=True,
                 )
+            if include_photometry_exists:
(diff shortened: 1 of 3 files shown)


stefanv:
You mention a docstring being wrong, but this PR doesn't seem to contain a documentation update. Could we address that too?

mcoughlin:
@stefanv The docstring says that the option was available but it actually wasn't (well it didnt do anything).

stefanv:
Cool, add a test and this is good to go.

stefanv (approved):

profjsb (approved):
LGTM

PR #2606, opened 2022-01-28
Adminify the telescope, instrument, and allocation pages

mcoughlin:
@profjsb suggests limiting access to the telescope, instrument, and allocation pages to admins.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -128,18 +128,6 @@ app:
       icon: SettingsInputAntenna
       url: /gcn_events
 
-    - name: Telescopes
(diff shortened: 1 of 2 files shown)


stefanv:
Is it useful for users to see the list of telescopes?  If so, we could also just hide the boxes that add a new telescope.  But fine with me the way it is done here.

stefanv:
(I presume even if users without permissions try right now they cannot add telescopes.)

profjsb:
That's a good idea @stefanv. In that case we'd need to move the permissions logic into the front end components. Do we have exemplars of that?

PR #2609, opened 2022-01-28
Make download CSV file more useful

mcoughlin:
Currently the CSV file downloaded gets very confused by our objects (just says [object Object]).

This makes the CSV file print out relevant things similar to how the table displays in the App.

"id","ra","dec","redshift","classification","groups"
"ZTF21aaqjmps","202.5050541","13.4111859","0.0246","","Sitewide Group;Program A;Program B"
"ZTF18aabcvnq","172.0534465","25.6596908","","Time-domain Source;variable","Sitewide Group;Program A;Program B"
"ZTFe028h94k","229.9620403","34.8442757","","","Sitewide Group;Program A;Program B"
"ZTFrlh6cyjh","342.4596","51.2634","","","Sitewide Group;Program A;Program B"
"ZTFJ201825+380316","304.60791628","38.05469","","","Sitewide Group;Program A;Program B"

### DIFF
--- a/static/js/components/SourceTable.jsx
+++ b/static/js/components/SourceTable.jsx
@@ -1086,9 +1086,70 @@ const SourceTable = ({
     onFilterChange: handleTableFilterChipChange,
     onFilterDialogOpen: () => setFilterFormSubmitted(false),
     search: false,
+    download: true,
     rowsExpanded: openedRows,
     onRowExpansionChange: (_, allRowsExpanded) => {
       setOpenedRows(allRowsExpanded.map((i) => i.dataIndex));
+    },
+    onDownload: (buildHead, buildBody, columnsDownload, data) => {
+      const renderDownloadClassification = (dataIndex) => {
+        const source = sources[dataIndex];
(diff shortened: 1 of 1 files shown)


static/js/components/SourceTable.jsx
```diff
@@ -1086,10 +1086,71 @@ const SourceTable = ({
...
+      };
+
+      return (
+        buildHead([
+          {
+            name: "id",
+            download: true,
+          },
+          {
+            name: "ra",
```
guynir42, line 1119: Maybe add the units here and for Dec?

guynir42 (commented):
I think there are a few columns in the current version of download CSV that may be useful, like Alias, TNS name, Date saved, Origin... maybe worth while to grab those as well.

static/js/components/SourceTable.jsx
mcoughlin: Done

mcoughlin:
@guynir42 Yeah good idea. I added those columns.

PR #2611, opened 2022-01-30
pip 22 breaks our dependencies

mcoughlin:
I think the release of pip 22 broke our dependency installation...

### DIFF
--- a/.github/workflows/build-and-deploy-docs.yaml
+++ b/.github/workflows/build-and-deploy-docs.yaml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install pip==21.3.1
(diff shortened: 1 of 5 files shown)


mcoughlin:
@stefanv Any ideas what to do about this one...?

profjsb:
Are you also getting the dependency failures on your local machine with pip22? I'm totally fine with pinning pip.

mcoughlin:
No I cant get it to fail locally for some reason, but that also seems to be a common issue:
https://github.com/pypa/pip/issues/10851

profjsb:
Looks like the fix is coming out very soon (22.0.2) https://github.com/pypa/pip/commit/347d19063f5deab2fb4b6e83a44be4cf8d15fbf4

stefanv:
We should revert this patch once the new pip is out.

PR #2640, opened 2022-02-04
Bokeh plot tests

guynir42:
I've added an API test to check some of the functionality of the bokeh plots, which have so far not been tested at all. 

So far I've only added tests for the spectrum plotting tool. I will add more tests for the photometry plots later.


skyportal/tests/api/test_bokeh_plots.py
```diff
@@ -0,0 +1,177 @@
+import os
+from skyportal.tests import api
+import numpy as np
+import datetime
+from bokeh.util import serialization
+
+
+def black_body(la, temp):  # black body radiation
```
profjsb, line 8:
```suggestion
def black_body(la, temp):
```

profjsb (approved):

guynir42:
@profjsb should I merge?

profjsb:
Sure

skyportal/handlers/api/internal/plot.py
```diff
@@ -45,6 +45,13 @@ class PlotSpectroscopyHandler(BaseHandler):
     def get(self, obj_id):
         width = self.get_query_argument("width", 600)
         device = self.get_query_argument("device", None)
+        smoothing = self.get_query_argument("smoothing", False)
+        if isinstance(smoothing, str):
+            if smoothing.lower() == 'true':
+                smoothing = True
+            else:
+                smoothing = False
```
stefanv, line 53:
```
smoothing = self.get_query_argument("smoothing", "false")
smoothing = True if smoothing.lower() == "true" else False
```

skyportal/plot.py
```diff
@@ -1343,6 +1344,34 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
 
+def smoothing_function(values, window_size):
+    """
+    Smooth the input "values" using a rolling average
+    where "window_size" is the number of points to use
+    for averaging.
+    This should be the same logic as static/js/plotjs/smooth_spectra.js
+    """
+
+    if values is None or not hasattr(values, '__len__') or len(values) == 0:
```
stefanv, line 1355: What does this code do?

skyportal/plot.py
```diff
@@ -1343,6 +1344,34 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
...
+        return values
+
+    output = np.zeros(values.shape)
+    under = int((window_size + 1) // 2) - 1
+    over = int(window_size // 2)
+
+    for i, v in enumerate(values):
```
stefanv, line 1362: Why not perform all this with NumPy operations?

skyportal/tests/api/test_bokeh_plots.py
```diff
@@ -0,0 +1,177 @@
...
+import numpy as np
+import datetime
+from bokeh.util import serialization
+
+
+def black_body(la, temp):
+    """
+    get the amount of radiation expected from a black body
+    of the given temperature "temp", at the given wavelengths "la".
+    :param la: float array or scalar
```
stefanv, line 12: We don't use this style of docstring in skyportal; please use numpydoc format

skyportal/tests/api/test_bokeh_plots.py
```diff
@@ -0,0 +1,177 @@
...
+    const = (
+        0.014387773538277204  # h*c/k_b = 6.62607004e-34 * 299792458 / 1.38064852e-23
+    )
+    amp = 1.1910429526245744e-25  # 2*h*c**2 * (nm / m) = 2*6.62607004e-34 * 299792458**2 / 1e9 the last term is units
+    la = la * 1e-9  # convert wavelength from nm to m
+
+    return amp / (la ** 5 * (np.exp(const / (la * temp)) - 1))
+
+
+def get_plot_data(data):
```
stefanv, line 28: If this is an internal function, name it `_get_plot_data`.

Please use numpydoc format for the docstring here too.

skyportal/tests/api/test_bokeh_plots.py
```diff
@@ -0,0 +1,177 @@
...
+    the underlying data for all spectra in the plot.
+
+    Returns a list of dictionaries with the flux, wavelength, etc.
+    """
+    objects = []
+    for doc in data['data']['bokehJSON']['doc']['roots']['references']:
+        if 'data' in doc['attributes']:
+            new_obj = {}
+            for field_name in ['wavelength', 'flux', 'flux_original', 'x', 'y']:
+                if field_name in doc['attributes']['data']:
```
stefanv, line 41: Maybe this should be refactored into a small utility function, or an inline function?  It's quite hard to read what's happening here.

stefanv (commented):
@guynir42 Do you mind adding these changes as a separate PR?

skyportal/plot.py
guynir42: Not sure what you mean? 
On line 1355 I just check that I didn't get `None` or an empty array or something that isn't an array or list.

skyportal/plot.py
guynir42: Just trying to make it as close to the JS function as I can.

skyportal/tests/api/test_bokeh_plots.py
guynir42: Sorry. My IDE got the best of me. I'll fix it.

skyportal/tests/api/test_bokeh_plots.py
guynir42: I'm not sure what counts as an internal function. It could conceivably be used by other parts of the code.

guynir42:
@stefanv Please have a look at my responses and then look at the other PR (#2651). Thanks!

skyportal/plot.py
stefanv: OK

skyportal/plot.py
stefanv: Looks like a call to np.convolve

skyportal/tests/api/test_bokeh_plots.py
stefanv: Sure, no worries.

skyportal/plot.py
guynir42: I know I can make this function more efficient, but I want to make sure the logic is the same as the smoothing on the JS side, that's why it is kind of clunky.

skyportal/tests/api/test_bokeh_plots.py
stefanv: Yeah, there's JSONPath, but I don't think that really helps matters along much.

Easiest is to just assign meaningful names, e.g.:

```
attributes = doc['attributes']['data']
if field_name in attributes:
    ...
```

I.e., try and get it to read as close to English.  But, I hear you.

skyportal/plot.py
stefanv: Since we have no test on the JavaScript side to make sure that *it* does the right convolution, perhaps it is good to use numpy here to verify that the JS does, in fact, compute borders etc. correctly?  They should match, surely.

guynir42: I don't understand how making the python side smoothing work with a convolution going to help test the JS side. Also, the edges are not exactly what you'd get from a convolution because I use a smaller window for the edges, where in a convolution you'd likely just pad with zeros or something.

stefanv: Fine, test the outputs with fixed values then.

guynir42: Not sure what you mean...

stefanv: We have a Javascript function that is untested.  We need to test it either against known values or a known implementation of the algorithm.

stefanv: I should go and look at the code again; I thought you needed the Python implementation matching the JS one so you could compare values, but I guess I didn't understand correctly what the Python version was for.

guynir42: The reason we need two versions is that when I call the internal API for the spectrum plot it needs to already send the data for one of the copies of each spectrum as smoothed (if the smooth parameter is given in the API call). 
Once on the JS side, when a user touches the smoothing button it triggers a call to the JS smoothing function, that hopefully does the right thing. 
I couldn't find a way to trigger the JS smoothing without rendering the whole thing, at which point the data is already ingested by BokehJS.

PR #2644, opened 2022-02-07
Testing source detection_stats on Github's CI

mcoughlin:
Using this PR for testing test_sources.py::test_sources_include_detection_stats.

### DIFF
--- a/skyportal/tests/api/test_sources.py
+++ b/skyportal/tests/api/test_sources.py
@@ -954,6 +954,13 @@ def test_sources_include_detection_stats(
     assert status == 200
     assert data["status"] == "success"
 
+    # Note: 40_587 is the MJD of UNIX time 1970-01-01
+    # Because arrow.get views dates as seconds since UNIX time,
+    # s["peak_detected_at"]` is the MJD of 90000 in isodate format.
+
+    # In summary: arrow.get("1970-01-01") - datetime.timedelta(40587) =>
+    # <Arrow [1858-11-17T00:00:00+00:00]>
+
     assert any(
         [
             s["last_detected_at"] == arrow.get((90000.0 - 40_587) * 86400.0).isoformat()
--- a/skyportal/tests/api/test_sources.py
+++ b/skyportal/tests/api/test_sources.py
@@ -906,13 +906,6 @@ def test_sources_include_detection_stats(
 
     assert any(
         [
-            d['name'] == '6dFgs gJ0001313-055904' and d['mstar'] == 336.60756522868667
-            for d in data
-        ]
-    )
-
-    assert any(
-        [
             s["last_detected_at"] == arrow.get((90000.0 - 40_587) * 86400.0).isoformat()
             for s in data["data"]["sources"]
         ]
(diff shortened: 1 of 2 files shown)


mcoughlin:
@guynir42 I don't understand this one at all.

guynir42:
@mcoughlin The order of the tests in the file doesn't matter (they get run in a random order) but it is possible that each time you run the test suite they get run in a different order. I don't know if that should affect this one failing test. 

In the report (https://github.com/skyportal/skyportal/runs/5099149843?check_suite_focus=true#:~:text=AssertionError%3A%20assert%20None%20%3D%3D%20%272105%2D04%2D16T00,04%2D16T00%3A00%3A00%2B00%3A00%27) it says that it returned `None` instead of the date it was last detected at. Under what situation could that happen?

mcoughlin:
@guynir42 I am not really sure. It could be None I suppose if the photometry post failed?

guynir42:
> @guynir42 I am not really sure. It could be None I suppose if the photometry post failed?

@mcoughlin I have a guess: It returns more than one source and sometimes the first source doesn't have a "last detected". This would happen occasionally, depending on what other sources were posted to the DB. 

You could fix it by checking that at least one source has the right date.

skyportal/tests/api/test_sources.py
```diff
@@ -824,6 +824,103 @@ def test_source_photometry_summary_info(
...
+        ]
+    )
+    assert any(
+        [
+            s["peak_detected_at"] == arrow.get((90000.0 - 40_587) * 86400.0).isoformat()
+            for s in data["data"]["sources"]
+        ]
+    )
+
+    assert any([s["last_detected_mag"] is not None for s in data["data"]["sources"]])
```
guynir42, line 920: These could trigger if any of the sources posted during testing have any photometry points above threshold. 
Why not look for the specific magnitudes you put in?

mcoughlin: @guynir42 good call.

skyportal/tests/api/test_sources.py
```diff
@@ -824,6 +824,167 @@ def test_source_photometry_summary_info(
...
+
+    assert any(
+        [
+            s["last_detected_at"] == arrow.get((90000.0 - 40_587) * 86400.0).isoformat()
+            for s in data["data"]["sources"]
+        ]
+    )
+    assert any(
+        [
+            s["peak_detected_at"] == arrow.get((90000.0 - 40_587) * 86400.0).isoformat()
```
profjsb, line 965: Should explain in a comment that `40_587` is the MJD of UNIX time 1970-01-01 and hence, since `arrow.get` views dates as seconds since UNIX time, s["peak_detected_at"]` is the MJD of 90000 in isodate format.

```
arrow.get("1970-01-01") - datetime.timedelta(40587)
# <Arrow [1858-11-17T00:00:00+00:00]>
```

profjsb (approved):

skyportal/tests/api/test_sources.py
mcoughlin: great idea!

guynir42 (approved):

PR #2645, opened 2022-02-08
shift model and API

mcoughlin:
This PR adds the model and API for the scanning shifts.


mcoughlin:
@guynir42 @profjsb I removed any front-end stuff from this one, which hopefully means we can still check interactions enough to decide if it is alright.

alembic/versions/5ab82f23e15c_add_shift.py
```diff
@@ -0,0 +1,53 @@
...
+Create Date: 2022-02-01 05:24:23.927168
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '5ab82f23e15c'
+down_revision = '82f5a89fbe59'
```
profjsb, line 14: will need to update the down revision given that there are new migrations.

skyportal/handlers/api/shift.py
```diff
@@ -0,0 +1,146 @@
...
+                    mode="read",
+                    options=[joinedload(Shift.group).joinedload(Group.users)],
+                )
+                .order_by(Shift.start_date.asc())
+                .all()
+            )
+        self.verify_and_commit()
+        return self.success(data=shifts)
+
+    @permissions(["Upload data"])
```
profjsb, line 108: should be a permissions to "Manage shifts"

skyportal/handlers/api/shift.py
```diff
@@ -0,0 +1,146 @@
...
+from ..base import BaseHandler
+from ...models import (
+    DBSession,
+    Group,
+    Shift,
+)
+
+
+class ShiftHandler(BaseHandler):
+    @permissions(["Upload data"])
```
profjsb, line 13: Should be a permission to "Manage shifts"

skyportal/tests/api/test_shifts.py
```diff
@@ -0,0 +1,22 @@
+from skyportal.tests import api
+
+
+def test_shift(public_group, super_admin_token, super_admin_user):
```
profjsb, line 4: Should be testing the deletion of a shift both with and without proper permissions.

skyportal/model_util.py
```diff
@@ -10,6 +10,7 @@
     'Manage users',
     'Manage sources',
     'Manage groups',
+    'Manage shifts',
```
profjsb, line 13: Also add "Manage shifts" to the "Group admin" role below (line 24). I think that's the default behavior you want...

profjsb (changes requested):

skyportal/model_util.py
mcoughlin: added

skyportal/tests/api/test_shifts.py
```diff
@@ -0,0 +1,22 @@
...
+
+    request_data = {
+        'group_id': public_group.id,
+        'start_date': '3021-02-27T00:00:00',
+        'end_date': '3021-07-20T00:00:00',
```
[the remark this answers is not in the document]
mcoughlin: added

skyportal/tests/api/test_shifts.py
mcoughlin: added

skyportal/handlers/api/shift.py
mcoughlin: added

skyportal/handlers/api/shift.py
mcoughlin: added

alembic/versions/5ab82f23e15c_add_shift.py
mcoughlin: The migration script is now complaining about bandpasses... although I expect this must be not related to this PR

PR #2636, opened 2022-02-03
Executed Observation Table display

mcoughlin:
This PR adds ExecutedObservations to the GcnEventPage in the form of a table with a searchable form.

Co-authored by @jadalilleboe.

PR #2646, opened 2022-02-08
Draft: Add form for modifying source selection on GCN event page

### LINKED ISSUE #2579: Sources in Skymaps
Add a front-end element to modify the start and end time and skymap confidence levels, i.e. the parameters here:
https://github.com/mcoughlin/skyportal/blob/crossmatch3/static/js/ducks/sources.js#L73
to modify the source selection criteria from the GcnEventPage.

jadalilleboe:
Resolves #2579 

Adds a form for modifying the source selection on GCN event page.

I'm having trouble understanding how to update the sources within the GCN event page. I see that the state of the gcnEventSources is in the redux store
`const gcnEventSources = useSelector(
    (state) => state?.sources?.gcnEventSources
  );`
but I'm not sure how to update that state, as of now I'm calling fetchGcnEventSources on submit.
Screenshot (23)
Screenshot (24)


mcoughlin:
@jadalilleboe I made some small modifications. Want to see what you think?

jadalilleboe:
@mcoughlin Is the expected behavior that the Event Sources section should change and be populated with whatever source was just submitted by the form? That's what I'm confused on trying to make happen
Screenshot (25)

jadalilleboe:
@mcoughlin sorry, still confused on how to make it appear and disappear. it worked for you on the test? doesn't happen for me when running skyportal locally

mcoughlin:
I open up the page with that test open (stopping it with like:

import pdb
pdb.set_trace()

and it appeared and disappeared. We can look at it together if it isnt working.

mcoughlin:
I am combining this functionality in #2636 as both sources and observations should be controlled by this form.

PR #2647, opened 2022-02-08
WIP: Issues using Candidate and Photometry Handlers

Theodlz:
This PR is a Work in Progress, it is a set of tests to show that we are having issues posting candidate and photometry data with their respective handlers. We might just be using it wrong, but for now we can't manage to make them work. Also, to give some context,  we are trying to use the handlers to insert data of ztf alerts coming from a kafka server. In order to do so, using the handlers directly and not the API seems necessary, as the number of API calls is limited, it limits the amount of data we an add at once (which can cause problems if we receive a lot of alerts in a short period of time).

This PR adds 2 test files in skyportal/tests/issues where you can find very basic calls to the post functions of telescope, instrument, stream, filter, source, candidate and photometry handlers. In the first file, you'll find the tests that passed for telescope, instrument, stream, filter and source. We considered that they pass if their status codes are all 200, and for some of them we used the get function from the handlers to verify that the data is posted . We plan to assert all results from the get functions to make sure the data has been added to the db, but as we can see it easily on the frontend for those elements, we didn't do it yet.

Then, a second file 'test_handler_issues.py' contains 4 tests, 2 of them test the post functions for candidate and photometry. The last 2 tests use the get functions to verify that data was added to the db (the very last one, that verify that photometry data has been added, is not finished yet).

Also, while re-running make db_clear && make db_init && make run_testing, we noticed that even for the first file, from time to time, the filter is not added to the db even if the status code is 200. We are unsure about the reason that causes this problem, but it seemed that deleting the db manually with postgresql fixed it.


mcoughlin:
@Theodlz Is this error reproducible just using test.api calls?

skyportal/tests/issues/test_handlers_issues.py
```diff
@@ -0,0 +1,179 @@
...
+        'filter_ids': [filter_id],
+        'passed_at': 59000,
+        'ra': 10,
+        'dec': 10,    
+    }
+
+    candidate_handler.current_user = super_admin_user
+    candidate_handler.request.body = json.dumps(body, indent=2).encode('utf-8')
+    try:
+        candidate_handler.post()
```
guynir42, line 85: What kind of error do you get when you try to post?

mcoughlin: I have asked @Theodlz to retry the external photometry post once #2648 is in.

PR #2648, opened 2022-02-08
Refactor photometry handler

mcoughlin:
This PR refactors the photometry handler to enable posting from a different API (useful for the ATLAS and PS1 PRs).


mcoughlin:
@profjsb It looks crazy pants, but this is just removing the 4 space indent for most of the functions at the start of the photometry handler and calling them directly.

skyportal/handlers/api/photometry.py
```diff
@@ -166,489 +172,546 @@ def serialize(phot, outsys, format):
...
+        raise ValidationError(
+            'Top level JSON must be an instance of `dict`, got ' f'{type(data)}.'
+        )
 
-        if not isinstance(data, dict):
-            raise ValidationError(
-                'Top level JSON must be an instance of `dict`, got ' f'{type(data)}.'
+    if "altdata" in data and not data["altdata"]:
+        del data["altdata"]
+    if "altdata" in data:
```
profjsb, line 184: I dont see how line 184 ever evaluates to True given the lines above.

profjsb (commented):
This looks great but I'm still worried about the LOCKs. Looking at https://stackoverflow.com/a/47256217 it seems to make sense for us to put those LOCK lines inside a try statement then in the except make sure we do a `DBSession.rollback()` to make sure we release the lock.

mcoughlin:
@profjsb So basically

    DBSession().execute(
        f'LOCK TABLE {Photometry.__tablename__} IN SHARE ROW EXCLUSIVE MODE'
    )
    try:
        ids, upload_id = insert_new_photometry_data(
            df, instrument_cache, group_ids, stream_ids, user
        )
    except ValidationError as e:
        return log(f"Unable to post photometry: {e}")

=>
    try:
        DBSession().execute(
            f'LOCK TABLE {Photometry.__tablename__} IN SHARE ROW EXCLUSIVE MODE'
        )
        ids, upload_id = insert_new_photometry_data(
            df, instrument_cache, group_ids, stream_ids, user
        )
    except Exception as e:
        DBSession.rollback()
        return log(f"Unable to post photometry: {e}")

stefanv:
At minimum, the lock would need to be wrapped in a `try: ... finally: ...` clause. But if the app gets brought down, the lock could end up hanging around.

I believe we have duplication checks on the DB level, so another option is to just do the check, try the insert, and catch the error if it fails?

mcoughlin:
@stefanv Can you help me understand how your second option is different from above?

mcoughlin:
@stefanv can you see if you are happier with this? I put your try except and with DBSession() as session around the LOCK bits.

skyportal/handlers/api/photometry.py
```diff
@@ -166,489 +172,548 @@ def serialize(phot, outsys, format):
...
+            "Invalid stream_ids parameter value. Must be a list of IDs (integers)."
+        )
+
+    stream_ids = list(set(stream_ids))
+    return stream_ids
+
+
+def add_external_photometry(json, user):
+    """
+    ---
```
stefanv, line 650: This docstring seems to be misplaced

stefanv (commented):
Thanks @mcoughlin! Looks good.  There's just that one docstring left hanging, and a question re: multiple types of session objects.

skyportal/handlers/api/photometry.py
mcoughlin: what does that mean?

mcoughlin: Ah I see, because now we just return the log()

skyportal/handlers/api/photometry.py
```diff
@@ -43,6 +45,10 @@
 _, cfg = load_env()
 
 
+log = make_log('api/photometry')
+Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
```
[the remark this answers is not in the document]
mcoughlin: I have usually seen us using DBSession() in the APIs and Session when called external to the APIs (like when we do the IOSync stuff).

skyportal/handlers/api/photometry.py
stefanv: Well, it looks like a handler (`GET`, `POST`, `PUT`, etc.) docstring in YAML, whereas I would expect a standard Python function docstring that describes the two arguments: `json` and `user`.

mcoughlin: Ah yes, hopefully more clear now?

skyportal/handlers/api/photometry.py
```diff
@@ -166,489 +172,523 @@ def serialize(phot, outsys, format):
...
+        try:
+            session.execute(
+                f'LOCK TABLE {Photometry.__tablename__} IN SHARE ROW EXCLUSIVE MODE'
+            )
+            ids, upload_id = insert_new_photometry_data(
+                df, instrument_cache, group_ids, stream_ids, user
             )
+        except Exception as e:
+            session.rollback()
+            return log(f"Unable to post photometry: {e}")
```
stefanv, line 686: Log does not return anything

skyportal/handlers/api/photometry.py
```diff
@@ -166,489 +172,519 @@ def serialize(phot, outsys, format):
...
 
-        stream_ids = list(set(stream_ids))
-        return stream_ids
+    return log("Successfully posted photometry")
```
stefanv, line 684:
```suggestion
    log("Successfully posted photometry")
```

stefanv (approved):

mcoughlin:
@profjsb please merge if comfortable with this now.

profjsb:
@stefanv should decide on merge now. I'm ok.

profjsb (approved):
Looks great. We should make another issue to look into why we get 400 and 500 sometimes (not suppose to happen).

skyportal/tests/api/test_photometry.py
```diff
@@ -2258,7 +2263,7 @@ def test_problematic_photometry_1276(
         data=payload,
         token=super_admin_token,
     )
-    assert status in [401, 500]
+    assert status in [400, 500]
```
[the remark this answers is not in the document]
stefanv: HTTP 500 is internal server error. BAD.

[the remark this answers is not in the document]
stefanv: This is something that should be investigated by scouring those logs.

PR #2613, opened 2022-02-01
Add concept of scanning shift

mcoughlin:
This PR adds a "Shifts" table to SkyPortal and associated API and front-end element. Similar to a filter, it ties a scanning shift to a group. Longer term will be good to add notifications to remind folks that they are on-call.

Co-authored by: julietteBrt

PR #2654, opened 2022-02-10
WIP: Added concept of shifts - Front-end elements

Theodlz:
This PR adds front-end elements to manage "Shifts" in SkyPortal, as well as associated table, API and tests. Similar to a filter, it ties a scanning shift to a group. Longer term will be good to add notifications to remind folks that they are on-call. This PR supercedes @mcoughlin 's PR: https://github.com/skyportal/skyportal/pull/2613

This PR also fixed some errors regarding timezones both in how it is stored in the DB and how it is displayed in the front-end for the Allocation Page.

Co-authored by: @mcoughlin , julietteBrt, thomasculino


mcoughlin:
@Theodlz please resolve conflicts now that the backend PR was merged.

alembic/versions/5ab82f23e15c_add_shift.py
```diff
@@ -1,7 +1,6 @@
 """add shift
 
 Revision ID: 5ab82f23e15c
-Revises: 66ed6017e9ac
```
mcoughlin, line 4: This should not be removed.

skyportal/models/shift.py
```diff
@@ -31,11 +31,11 @@ class Shift(Base):
     )
 
     start_date = sa.Column(
-        sa.DateTime, nullable=False, index=True, doc="The start time of this shift."
+        sa.DateTime(timezone= True), nullable=False, index=True, doc="The start time of this shift."
```
mcoughlin, line 34: What does this do? I don't think we want to be making different timezone treatment choices than other places in the DB.

static/js/ducks/source.js
```diff
@@ -132,6 +132,7 @@ export function deleteCommentOnSpectrum(spectrumID, commentID) {
   );
 }
 
+
```
mcoughlin, line 135: Probably not needed.

skyportal/models/shift.py
Theodlz: Issue is,  the db received time in local and saving it as is. So when getting it back to the frontend, it wouldn't know it is not in UTC, it didn't have a way to put it back to the right timezone for the user seeing it on the frontend. At least i had that problem while editing the code. I can clone an older version and double check to avoid unnecessary changes

Theodlz: Here's an example. Without that, as the data in UTC is somehow and somewhere, when sent to the DB, turned back to local to be stored in the DB, without any indication of the associated timezone, if a user in GMT+1 saves a shift between 8AM and 10AM, a user in GMT+2, would still see it as 8AM to 10AM, and not 9AM to 11AM. So there is 2 solutions : make sure that it is saved in UTC which means whatever is putting back in local, or saving the 'offset' associated with the local time in the DB, which is bone by the argument timezone=True

Theodlz: Best way to fix this without adding the timezone in the DB would be to figure out what and where the time in UTC is converted back to local time before being added in the DB. We can consider what i have right now a temporary solution, i'll try to find a way to fix it.

PR #2659, opened 2022-02-11
Observation plan migration

mcoughlin:
This PR makes the first database migrations required to add ObservationPlanRequests.


profjsb:
@mcoughlin will probably need to update the down revision since we just merged another migration PR.

skyportal/models/observation_plan.py
```diff
@@ -0,0 +1,347 @@
...
+
+    @num_observations.expression
+    def num_observations(cls):
+        """Number of planned observation."""
+        return cls.planned_observations.count()
+
+    @property
+    def total_time(self):
+        """Total observation time (seconds)."""
+        return sum(_.exposure_time for _ in self.planned_observations)
```
guynir42, line 265: Is there a reason to use underscores instead of a short variable name?

skyportal/models/observation_plan.py
```diff
@@ -0,0 +1,347 @@
...
+        "Instrument",
+        foreign_keys=instrument_id,
+        doc="The Instrument that this observation plan belongs to",
+    )
+
+    dateobs = sa.Column(
+        sa.ForeignKey('gcnevents.dateobs', ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+        doc='UTC event timestamp',
```
guynir42, line 206: Emphasize that this is the time of the GCN trigger, not of the observations.

skyportal/models/observation_plan.py
```diff
@@ -0,0 +1,347 @@
...
+        sa.ForeignKey('instruments.id', ondelete="CASCADE"),
+        nullable=False,
+        doc='Instrument ID',
+    )
+
+    dateobs = sa.Column(
+        sa.ForeignKey('gcnevents.dateobs', ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+        doc='UTC event timestamp',
```
guynir42, line 318: Same as above. In particular when you have both `dateobs` and `obstime` that have very different meanings.

guynir42 (approved):
I have very minor comments. 
Also, I don't remember if you already added some docs describing the MMA concepts, but I think it is also a good place to explain the observations plans, requests, and so on. For me, at least, it is quite confusing.

skyportal/models/observation_plan.py
mcoughlin: fixed.

skyportal/models/observation_plan.py
mcoughlin: extended this

skyportal/models/observation_plan.py
mcoughlin: same here.

mcoughlin:
@guynir42 Yeah I have the doc discussion in the other (mega PR), which we can add to the API PR when I work on it tomorrow.

mcoughlin:
@guynir42 will do. thanks.

PR #2663, opened 2022-02-13
Add shift front-end page

Theodlz:
This PR adds front-end elements to manage "Shifts" in SkyPortal, as well as associated table, API and tests. Similar to a filter, it ties a scanning shift to a group. Longer term will be good to add notifications to remind folks that they are on-call. This PR supercedes @mcoughlin 's PR: https://github.com/skyportal/skyportal/pull/2613

This PR also fixed some errors regarding timezones both in how it is stored in the DB and how it is displayed in the front-end for the Allocation Page. And it works in the same way for the Shifts. User inputs start and end date in local time, it gets stored in the DB in UTC, and it is displayed on the frontend in the local time of the user viewing it.

Co-authored by: @mcoughlin , julietteBrt, thomasculino


guynir42 (commented):
The code looks good, with some minor comments. 

I do have one major issue: when using the shifts page, if I get an error, that error does not clear event after I successfully submit a new shift. That's very confusing because the shift is added but it looks like there were errors in the form.

skyportal/tests/flaky/test_frontend_flaky.py
```diff
@@ -243,3 +243,63 @@ def test_super_user_post_allocation(
...
+    driver.wait_for_xpath('//*[@id="root_start_date"]').send_keys(Keys.TAB)
+    driver.wait_for_xpath('//*[@id="root_start_date"]').send_keys('01:01')
+    driver.wait_for_xpath('//*[@id="root_start_date"]').send_keys('P')
+
+    driver.wait_for_xpath('//*[@id="root_end_date"]').send_keys('03/01/2022')
+    driver.wait_for_xpath('//*[@id="root_end_date"]').send_keys(Keys.TAB)
+    driver.wait_for_xpath('//*[@id="root_end_date"]').send_keys('01:01')
+    driver.wait_for_xpath('//*[@id="root_end_date"]').send_keys('P')
+
+    driver.wait_for_xpath('//*[@id="root_end_date"]').send_keys('02/01/2022 01:01:10')
```
guynir42, line 289: Why do we need this line?

skyportal/tests/flaky/test_frontend_flaky.py
```diff
@@ -243,3 +243,63 @@ def test_super_user_post_allocation(
...
+
+    driver.click_xpath('//*[@id="root_group_id"]')
+    driver.click_xpath('//li[contains(text(), "Sitewide Group")]')
```
guynir42, line 292: I don't know why, but adding a try/except of these two lines seems to make this part of the test run smoothly.

skyportal/tests/flaky/test_frontend_flaky.py
```diff
@@ -243,3 +243,63 @@ def test_super_user_post_allocation(
...
+
+    submit_button_xpath = '//button[@type="submit"]'
+    driver.wait_for_xpath(submit_button_xpath)
+    driver.click_xpath(submit_button_xpath)
+
+    # check for dropdown shift
+    driver.wait_for_xpath(f'//span[text()[contains(.,"{form_name}")]]')
+
+    # check for delete shift button
+    delete_button_xpath = '//button[@id="delete_button"]'
```
guynir42, line 302: This will find a delete button, but not necessarily the one for the right shift (if there are multiple shifts). Maybe this also makes this test flaky. Probably need to make the xpath more specific to the one shift with the right name.

guynir42 (commented):
I've also played around with the test on this PR. 

I think putting the tests directly into flaky is just as good as not testing at all. 
Let's make an effort to get this test back to working order and put it in the regular test folders. 

One of the ways the tests fail is if there are already other shifts, or objects, or groups that are added in previous tests, filling up the pages/tables/form dropdowns, which makes it harder for the driver to find the right button to click, etc. 

For example: 
This test passes for me locally. 
Then I tried putting lines 257-270 in a loop with 10 iterations and now it fails. 
Try this and make sure the test is no longer flaky. 
The real test suite probably doesn't post any shifts outside this test, so maybe try posting a few groups instead, that may help recreate the flakiness of the test suite.

Theodlz:
Hi @guynir42 , Thank you very much for your review. I just tried to reproduce the issue you had with errors and shifts still getting posted. From what i see, it seems that it's not that the shifts that encountered the errors get posted, it seems that its just that the error message persists, even when we input valid start and end date, after having an error with wrong dates. I'll fix that first. Can you confirm that this is the problem you are having too ?

static/js/components/ShiftPage.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+  return result;
+}
+
+const ShiftList = ({ shifts }) => {
+  const dispatch = useDispatch();
+  const deleteShift = (shift) => {
+    dispatch(shiftActions.deleteShift(shift.id)).then((result) => {
+      if (result.status === "success") {
+        dispatch(showNotification("Shift deleted", "warning"));
```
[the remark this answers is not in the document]
Theodlz: Basically what we wanted to do is display the notification that says that we deleted a shift in a yellow/red color. It seemed counter-intuitive to us that a deletion notification could be green. Passing 'warning' as an argument here made it yellow, is there ay other way ? If you think that green is fine then we'll just use a normal notification.

static/js/components/ShiftPage.jsx
```diff
@@ -0,0 +1,167 @@
...
+    ...(shift?.end_date ? [`End Date: ${endDate}`] : []),
+  ];
+
+  // eslint-disable-next-line prefer-template
+  const result = array.join("\n");
+
+  return result;
+}
+
+const ShiftList = ({ shifts }) => {
```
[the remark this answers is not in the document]
Theodlz: Good idea, i'll work on that

skyportal/tests/flaky/test_frontend_flaky.py
Theodlz: Done. Now looks for the list element that contains the same group name and shift name as the shift we just added

skyportal/tests/flaky/test_frontend_flaky.py
Theodlz: Useless you are right, i deleted it

static/js/components/ShiftPage.jsx
```diff
@@ -0,0 +1,167 @@
...
+    <Grid container spacing={3}>
+      <Grid item md={6} sm={12}>
+        <Paper elevation={1}>
+          <div className={classes.paperContent}>
+            <Typography variant="h6">List of Shifts</Typography>
+            <ShiftList shifts={shiftList} />
+          </div>
+        </Paper>
+      </Grid>
+      {(currentUser.permissions?.includes("System admin") ||
```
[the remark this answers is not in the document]
Theodlz: Done.

skyportal/tests/flaky/test_frontend_flaky.py
Theodlz: You mean putting those 2 lines in the same try clause or seperate ? And if there is an exception that is caught, i should get the test to stop and not pass ?

static/js/components/ShiftPage.jsx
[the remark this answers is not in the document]
mcoughlin: Green seems better here.

skyportal/tests/flaky/test_frontend_flaky.py
Theodlz: Noted, i'll do that soon.

guynir42 (approved):
It looks a lot better and the behavior of the errors is much nicer now that the validation is live. 
I have only one minor comment/question below. After that it looks good

static/js/components/NewAllocation.jsx
```diff
@@ -20,6 +35,13 @@ const NewAllocation = () => {
   };
 
   function validate(formData, errors) {
+    if (nowDate > formData.end_date) {
+      errors.end_date.addError(
+        "End date must be after current time, please fix."
+      );
+    } else {
+      errors = [];
+    }
```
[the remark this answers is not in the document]
Theodlz: Hi Guy, this is actually not supposed to be there. I played around with trying to empty the error Array before using the liveValidation. I'll delete it ASAP

Theodlz:
Worth mentionning: Me, thomasculino and @mcoughlin are still thinking about some of the features and how to implement them. So even when the last changes to the tests will be done, we will still have things to add and probably things to change (frontend for sure, backend probably). Maybe that should be the subject of another PR ? Or maybe this one could take care of you, what do you think ?

mcoughlin:
@Theodlz small PRs are good PRs. Let's get this one over the hump.

Theodlz:
Amazing 🔭

guynir42:
@profjsb I think you can review this now. It looks in good shape.

static/js/components/ShiftPage.jsx
```diff
@@ -0,0 +1,184 @@
...
+  const { shiftList } = useSelector((state) => state.shifts);
+  const currentUser = useSelector((state) => state.profile);
+
+  if (!shiftList) {
+    return <CircularProgress />;
+  }
+
+  const permission =
+    currentUser.permissions?.includes("System admin") ||
+    currentUser.permissions?.includes("Manage groups") ||
```
profjsb, line 152: It's not clear that we want to grant this access to someone with `manage groups` -- since we have a more fine grained `manage shifts` if someone needs to be able to manage shifts then they should get that permissions added to their user

profjsb (approved):
This looks really good. Small comment to address then good to go.

skyportal/tests/flaky/test_frontend_flaky.py
Theodlz: Hi @guynir42 , we deleted the tests from the flaky folder and reformatted them as normal tests in the frontend folder

mcoughlin (approved):
Two checkmarks with comments addressed, and the code looks good. Nice job @Theodlz .

PR #2674, opened 2022-02-16
Allow group users to remove themselves from groups

### LINKED ISSUE #2664: Non admin users can't leave groups
We can't leave a group if we are not group admin. The expected behaviour would be that any user of a group, admin or not, can leave it.

mcoughlin:
This PR allows users to remove themselves from groups (even if they are not admins)

Closes #2664.


static/js/components/GroupPageManageUserButtons.jsx
```diff
@@ -63,49 +65,60 @@ const ManageUserButtons = ({ group, loadedId, user, isAdmin }) => {
...
   };
 
+  const handleDelete = () => {
+    dispatch(
+      groupsActions.deleteGroupUser({
+        userID: user.id,
+        group_id: group.id,
+      })
+    );
+    history.goBack();
```
stefanv, line 75: Why is the back navigation necessary here?

mcoughlin: @stefanv Because once you are out of the group, the group page does not exist for you anymore :)

mcoughlin: Maybe it should be some kind of if statement?

profjsb: I wonder if we shouldn't splash a pop up warning saying do you really want to leave etc? This is a pretty big change and we don't want people to click this button without understanding the implication.

mcoughlin: @profjsb @stefanv Is there an example of this in the app already?

mcoughlin: @profjsb @stefanv I just didn't know what to grep for. Indeed that confirmation exists!

mcoughlin:
@profjsb Please take another look when you can.

static/js/components/GroupPageManageUserButtons.jsx
```diff
@@ -63,49 +76,88 @@ const ManageUserButtons = ({ group, loadedId, user, isAdmin }) => {
...
+            disabled={isAdmin(user) && numAdmins === 1}
+          >
+            <DeleteIcon />
+          </IconButton>
+          <Dialog
+            fullWidth
+            open={confirmDeleteOpen}
+            onClose={handleConfirmDeleteDialogClose}
+          >
+            <DialogTitle>Remove user?</DialogTitle>
```
profjsb, line 136: Should have a different title user:
  1) "Remove yourself?" if user.username === currentUser.username
  2) Otherwise "Remove user?"

static/js/components/GroupPageManageUserButtons.jsx
```diff
@@ -63,49 +76,88 @@ const ManageUserButtons = ({ group, loadedId, user, isAdmin }) => {
...
+            <DialogContent dividers>
+              <DialogContentText>
+                Are you sure you want to delete this user from this group?
+                <br />
+                Warning! This will delete the user from the group and all of its
+                filters.
```
profjsb, line 142: Should have a different title user:
  1) "yourself" and "you" if it's the current user
  2) Otherwise "this user" and "the user"

profjsb (changes requested):
Some small changes to suggest here. Note on the demo data, after deleting the current user  I'm getting
`Insufficient permissions for operation "User 3 read Classification {1, 2}".` on the homepage. Not sure if this is out of scope for this PR but maybe an easy fix?

static/js/components/GroupPageManageUserButtons.jsx
mcoughlin: yes

mcoughlin:
@profjsb I tried to catch that AccessError. See what you think!

skyportal/handlers/api/internal/recent_sources.py
```diff
@@ -89,5 +89,12 @@ def get(self):
...
             del source["recency_index"]
 
-        self.verify_and_commit()
+        try:
+            # check for AccessError:
+            # this is in a try-except in case of group deletions
+            # which lead to a change in what classifications are accessible
+            self.verify_and_commit()
+        except AccessError as e:
+            return self.error(f'Classification access failed: {e}')
```
profjsb, line 98: I dont think it makes sense to show classification access errors to the end users since they did not explicitly ask for classifications. Perhaps just log the error using `log()`?

I'm a bit concerned here that we may not have the cascade of classification permissions correct. A user should be able to see classifications that they have made (which means classifications should automatically belong to that user's solo group) and not classifications saved to groups in which they do not belong. I would have thought that the `source_entry = (Source.query_records_accessible_by(self.current_user)...` call would have been enough to enforce correct classifications viewership.

profjsb (changes requested):
Small change but we should look into why the classifications are erroring this way (see my comment)

skyportal/handlers/api/internal/recent_sources.py
mcoughlin: I think Stefan and I resolved this yesterday. We needed a source refresh.

profjsb (approved):
thanks for working on the permissioning/refresh issues.

Issue #2668, opened 2022-02-15
Uncaught exception in lightcurve plot


stefanv:
```
ERROR:tornado.application:Uncaught exception GET /api/internal/plot/photometry/ZTF21acaqdee?width=800&device=browser (78.197.174.32)
Traceback (most recent call last):
  File "/skyportal/skyportal/plot.py", line 331, in get_effective_wavelength
    bandpass = snc.get_bandpass(bandpass_name)
  File "/skyportal_env/lib/python3.8/site-packages/sncosmo/bandpasses.py", line 27, in get_bandpass
    return _BANDPASSES.retrieve(name)
  File "/skyportal_env/lib/python3.8/site-packages/sncosmo/_registry.py", line 160, in retrieve
    self._instances[key] = func(*args, name=name)
  File "/skyportal_env/lib/python3.8/site-packages/sncosmo/builtins.py", line 75, in load_bandpass_remote_aa
    return read_bandpass(abspath, wave_unit=u.AA,
  File "/skyportal_env/lib/python3.8/site-packages/sncosmo/bandpasses.py", line 71, in read_bandpass
    return Bandpass(t['wave'], t['trans'], wave_unit=wave_unit,
  File "/skyportal_env/lib/python3.8/site-packages/sncosmo/bandpasses.py", line 201, in __init__
    s = slice_exclude_below(trans, np.max(trans) * trim_level, grow=1)
  File "<__array_function__ internals>", line 5, in amax
  File "/skyportal_env/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 2754, in amax
    return _wrapreduction(a, np.maximum, 'max', axis, None, out,
  File "/skyportal_env/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
ValueError: zero-size array to reduction operation maximum which has no identity
```

guynir42:
Closed by #2675 (I hope)

guynir42:
Closed unless this bug comes back...

Issue #2667, opened 2022-02-15
Uncaught exception in `/api/photometry`


stefanv:
```
ERROR:tornado.application:Uncaught exception POST /api/photometry (131.215.198.70)
  File "/skyportal/skyportal/handlers/api/photometry.py", line 290, in standardize_photometry_data
    first_offender = np.argwhere(missing)[0, 0]
ValueError: Data must be 1-dimensional
```

guynir42:
@stefanv I'm not sure how this happens. The bug comes from the lines right before: 
```
        for field in PhotMagFlexible.required_keys:
            missing = df[field].isna()
            if any(missing):
                first_offender = np.argwhere(missing)[0, 0]
```
Somehow `df[field].isna()` produces something that is not 1-dimensional. I don't think an empty array would get past the `if any(missing)` so it is likely that indexing by `field` somehow gets a multidimensional dataframe. I just don't have a good guess why that would happen... or what to do if it does (raise? serialize the result into a 1D array?)

stefanv:
Here's the full traceback:

```
HTTPServerRequest(protocol='http', host='fritz.science', method='POST', uri='/api/photometry', version='HTTP/1.0', remote_ip='131.215.198.70')
Traceback (most recent call last):
  File "/skyportal_env/lib/python3.8/site-packages/tornado/web.py", line 1702, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/skyportal/baselayer/app/access.py", line 32, in wrapper
    return method(self, *args, **kwargs)
  File "/skyportal/baselayer/app/access.py", line 61, in wrapper
    return method(self, *args, **kwargs)
  File "/skyportal/skyportal/handlers/api/photometry.py", line 701, in post
    df, instrument_cache = self.standardize_photometry_data()
  File "/skyportal/skyportal/handlers/api/photometry.py", line 290, in standardize_photometry_data
    first_offender = np.argwhere(missing)[0, 0]
  File "<__array_function__ internals>", line 5, in argwhere
  File "/skyportal_env/lib/python3.8/site-packages/numpy/core/numeric.py", line 618, in argwhere
    return transpose(nonzero(a))
  File "<__array_function__ internals>", line 5, in nonzero
  File "/skyportal_env/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 1921, in nonzero
    return _wrapfunc(a, 'nonzero')
[...]

PR #2675, opened 2022-02-16
Fix uncaught errors in photometry handler and plot

guynir42:
There are a couple of uncaught errors found in the logs, see issues #2667 and #2668. 
This PR tries to address those.


skyportal/handlers/api/internal/plot.py
```diff
@@ -52,15 +52,19 @@ def get(self, obj_id):
...
+            json = plot.spectroscopy_plot(
+                obj_id,
+                self.associated_user_object,
+                spec_id,
+                width=int(width),
+                device=device,
+                smoothing=smoothing,
+                smooth_number=smooth_number,
+            )
+        except ValueError as e:
```
stefanv, line 65: Maybe just a blank exception here, i.e. Exception instead of ValueError?

stefanv (approved):

stefanv:
Thanks!

PR #2677, opened 2022-02-17
Add description column to group table

### LINKED ISSUE #2529: Group description
It would be nice if the Group table would allow for some kind of longer description (for example including a reference or url or whatever) in a dedicated column.

jadalilleboe:
Closes #2529 

Adds a column to the group table for a longer description of the group.


skyportal/tests/api/test_groups.py
```diff
@@ -11,7 +11,11 @@ def test_token_user_create_new_group(super_admin_token, super_admin_user):
...
         "groups",
-        data={"name": group_name, "group_admins": [super_admin_user.id]},
+        data={
+            "name": group_name,
+            "description": "Group description",
+            "group_admins": [super_admin_user.id],
+        },
         token=super_admin_token,
     )
     assert status == 200
```
mcoughlin, line 21: I would add to this test you get your correct description back

mcoughlin (approved):
@jadalilleboe just one minor comment, otherwise fine for me. Also, will need to add this to the frontend group creation, but that can be in the next PR.

skyportal/tests/api/test_groups.py
jadalilleboe: added

mcoughlin: ok good for me once the tests are green.

jadalilleboe:
I think I will move this to #2678, it seems small enough to be in 1 PR and the tests are passing there.

PR #2680, opened 2022-02-18
Boolean query validation

guynir42:
- Pin baselayer to get new version of `get_query_argument` that handles boolean defaults / string conversion from query. 
- This immediately fixes all the places where the variables were supposed to be boolean but were actually string (i.e., you could give the argument `"False"` but then it would be evaluated as `True` in the conditionals. 
- Fix all the places where we used various methods to convert the strings "True" or "False" to booleans.
- Now the following values are interpreted as `True`: `"t", "true', "yes", "1" disregarding case. 
- All other values are `False`. 
- The conversion of string to boolean only happens when using a boolean default, e.g., `self.get_query_argument(..., False)`


stefanv:
Bunch of unrelated commits here, looks like?

guynir42:
> Bunch of unrelated commits here, looks like?

I just pulled master after this branch was on hold for a while. The last two commits are where I actually start working.

PR #2683, opened 2022-02-20
Executed observation page

mcoughlin:
This PR adds a page for Executed Observations, similar to for telescopes, instruments, and allocations, for display and upload.


skyportal/tests/frontend/test_upload_observations.py
```diff
@@ -0,0 +1,70 @@
...
+        os.path.join(
+            os.path.dirname(os.path.dirname(__file__)),
+            'data',
+            filename,
+        ),
+    )
+
+    driver.wait_for_xpath(f'//*[contains(., "{filename}")]')
+    submit_button_xpath = '//button[contains(.,"Submit")]'
+    driver.click_xpath(submit_button_xpath, scroll_parent=True)
```
profjsb, line 70: Can you now test that those observations are indeed there/uploaded?

profjsb: Please also test behavior when you upload 
 1) malformed observations
 2) observations without airmass and seeing
 3) identical observations to what was loaded by someone else.

(I dont know what behavior you expect in 3 above, should be documented anyway).

profjsb (changes requested):

skyportal/handlers/api/observation.py
```diff
@@ -514,3 +519,95 @@ def delete(self, observation_id):
...
+                Instrument.id == instrument_id,
+            )
+            .first()
+        )
+        if instrument is None:
+            return self.error(message=f"Missing instrument with ID {instrument_id}")
+
+        observation_data = pd.read_table(StringIO(observation_data), sep=",").to_dict(
+            orient='list'
+        )
```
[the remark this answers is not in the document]
mcoughlin: done.

skyportal/tests/frontend/test_upload_observations.py
mcoughlin: Added. The identical observations one I do not currently test for as that is a log message, as opposed to something returned to the frontend. (just with the baselayer log). I made a few changes to the doc to make it clear what is actually required.

mcoughlin:
@profjsb At @guynir42's suggestion, I went through and ran py.test with --pdb enabled, which led to mostly seeing places where a scroll_parent or timeout was useful. It was interesting indeed to see that the answers would be different each time, running on my own laptop 3 or 4 times.

profjsb (approved):

PR #2684, opened 2022-02-20
Download button for observation plan draft GCN

mcoughlin:
This PR adds a draft GCN (and associated download button) for the observation plans.


mcoughlin:
Thanks @guynir42 .

PR #2689, opened 2022-02-23
WIP: Updated shift model and API

Theodlz:
#### Issues with the current model : 

- Main issue : Shifts belong to a group, but there is no way for this group's users to join, or not join a shift. So we can't assign people to shifts, we can't see who will participate in a shift.
- Secondary issue: We can't add any description/details about a shift. So unless you make it very clear in its name, it is impossible to know what a shift is about / what is its goal (we won't go in details here, we simply added a column to the shift table to add the description, and it is displayed in the frontend).

What we need, is to allow users the group to which the shift belongs, to join or leave that particular shift. And of course, in the DB, there is something that allows you to create a group of users, that can ask to join, join and leave it: the groups.

#### Solution: Instead of belonging to one "main" group, a shift is now linked to 2 distinct groups.

-  The first one is still the **"main group"**. A shift belongs to it, and only members from that main group can see that shift in the shift page and interact with it.
- The second one, is a unique group created at the same time as the shift. This **"shift group"**, is what allows us to have users participate or not to a specific shift. On the frontend, users will be able in the next PR to join or leave that shift group. Now, as users can participate to a shift, or not, we are also able to display the list of people participating to a shift.

#### What has been implemented yet (we will update that list when we make progress):

1. The models for the shifts and the groups has been modified (to add new columns and update relationships as required by sqlalchemy).
2. The handler for the shift has been modified (to create the associated "shift group" automatically when we create a shift).
3. The frontend elements that were in conflict with the backend changes. We made minimal changes here, but we are planning on adding the buttons to join and leave a shift).

#### What to do next ?

We are currently working on another PR that aims to update the frontend to add a calendar element to the Shift Page. You can find it here: [WIP: Updated shift frontend using new backend and API #2701](https://github.com/skyportal/skyportal/pull/2701)


Theodlz:
Hi @mcoughlin , tagging you here to let you know we opened that backend PR.

skyportal/handlers/api/shift.py
```diff
@@ -42,14 +38,47 @@ def post(self):
...
+        schema = Shift.__schema__()
+        data['shift_group_id'] = group.id
+        data.pop('name', None)
+        data.pop('group_admins', None)
         try:
             shift = schema.load(data)
         except ValidationError as exc:
             return self.error(
-                f"Invalid/missing parameters: {exc.normalized_messages()}"
+                f"2Invalid/missing parameters: {exc.normalized_messages()}"
```
profjsb, line 81: 2?

skyportal/models/group.py
```diff
@@ -302,11 +302,19 @@ class Group(Base):
...
+        back_populates="shift_group",
         passive_deletes=True,
         doc='All shifts associated with a group.',
     )
 
+    shift_group = sa.Column(
+        sa.Boolean,
+        default=False,
+        index=True,
+        doc='Flag indicating whether this group ' 'is a shift group or not.',
```
profjsb, line 315:
```suggestion
        doc='Flag indicating whether this group is a shift group or not.',
```

skyportal/tests/api/test_shifts.py
```diff
@@ -1,11 +1,30 @@
 from skyportal.tests import api
+import uuid
 
 
 def test_shift(public_group, super_admin_token, view_only_token, super_admin_user):
 
+    shift_name = str(uuid.uuid4())
+    group_description = "Group description"
```
profjsb, line 8:
```suggestion
    group_description = "Shift Group description"
```

static/js/components/NewShift.jsx
```diff
@@ -62,6 +62,10 @@ const NewShift = () => {
         title: "Group",
         default: groups[0]?.id,
       },
+      name: {
+        type: "string",
+        title: "Shift name (i.e. the Night Shift)",
```
profjsb, line 67:
```suggestion
        title: "Shift name (ie. the Night Shift)",
```

profjsb (changes requested):
Makes sense to make -- thanks for the context in the comments. Either you'll need to modify the group administrative pages as well as any page that shows groups (like classifications, comments, etc.) to only show non-shift groups OR you could create a new model class ShiftGroup. I strongly prefer the latter since you wont need to tough a lot of other code. In this case you'd create a subclass of Group:

```python
class ShiftGroup(Group):
```

which would take as it's cascade delete foreign key one single group ID. Make sense?

Theodlz:
Hi @profjsb , thank you so much for your review and suggestions. You're right, it does make sense to create a subclass of the group model, which is basically creating a new table in the DB. I was also thinking, if we create a new table with this subclass, maybe we can also use that table to hold all the shift infos (start date, end date, ...), so to have one unique table for the shifts instead of having 2 of them ? I'll work on adding that subclass then. Have a nice day !

Theodlz:
> Makes sense to me -- thanks for the context in the comments. Either you'll need to modify the group administrative pages as well as any page that shows groups (like classifications, comments, etc.) to only show non-shift groups OR you could create a new model class ShiftGroup. I strongly prefer the latter since you wont need to tough a lot of other code. In this case you'd create a subclass of Group:
> 
> ```python
> class ShiftGroup(Group):
> ```
> 
> which would take as it's cascade delete foreign key one single group ID. Make sense?

Hi @profjsb , regarding this class ShiftGroup inheriting from Group, what is the expected behaviour you have in mind ? What i had in mind and what i understood from your comment is the following: the ShiftGroup class inherits the same columns as a Group, so it has a name, a description, relationships with users and GroupUsers, ..., and to that we add 2 columns related to shifts: start_date and end_date. But when i use something like : 
 ```python
 class ShiftGroup(Group):
> ```

It does not do that, I doesn't not inherit the columns of the Group table, but sql achemy asks me to have one of the columns of ShiftGroup to be a foreign key to a column in the Group Table. 

Am I doing it wrong or did I not understand what you meant ?

skyportal/handlers/api/shift.py
```diff
@@ -145,3 +174,247 @@ def delete(self, shift_id):
         self.push_all(action="skyportal/REFRESH_SHIFTS")
 
         return self.success()
+
+
+class ShiftUserHandler(BaseHandler):
+    @permissions(["Upload data"])
```
profjsb, line 180: should be a new role "Manage shifts"

profjsb (changes requested):
Looking very good. Some small changes suggested. Please also add some data as appropriate to the demo data. Also, will need some new tests.

skyportal/handlers/api/shift.py
```diff
@@ -145,3 +174,247 @@ def delete(self, shift_id):
...
+              schema:
+                type: object
+                properties:
+                  userID:
+                    type: integer
+                  admin:
+                    type: boolean
+                  canSave:
+                    type: boolean
+                    description: Boolean indicating whether user can save sources to shift. Defaults to true.
```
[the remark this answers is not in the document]
Theodlz: Hi @profjsb , it's a simple copy and paste issue. As mentionned it is a Work in Progress and we copied a lot from the groups table to rapidly get something up and working. We'll delete this soon.

skyportal/handlers/api/shift.py
```diff
@@ -145,3 +174,247 @@ def delete(self, shift_id):
...
+        user_id = data.get("userID", None)
+        if user_id is None:
+            return self.error("userID parameter must be specified")
+        try:
+            user_id = int(user_id)
+        except (ValueError, TypeError):
+            return self.error("Invalid userID parameter: unable to parse to integer")
+
+        admin = data.get("admin", False)
+        if not isinstance(admin, bool):
```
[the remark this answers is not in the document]
Theodlz: You're right. By the way this comes from the group handler, so maybe it needs to be corrected in the group handler too.

skyportal/handlers/api/shift.py
```diff
@@ -145,3 +174,247 @@ def delete(self, shift_id):
...
+            return self.error(
+                "Invalid (non-boolean) value provided for parameter `canSave`"
+            )
+        shiftuser.admin = admin
+        shiftuser.can_save = can_save
+        self.verify_and_commit()
+        return self.success()
+
+    @auth_or_token
+    def delete(self, shift_id, user_id):
```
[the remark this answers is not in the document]
Theodlz: We are actually hesitating. We want to allow a user of a group to join a shift and leave it freely. So really the only thing to check (that we considered), is that the user belongs to the group that has the shift. We could also verify the permission to have a "difference" between users that can participate to shifts or not in a specific group. We'll think about it.

static/js/components/NewShift.jsx
```diff
@@ -74,12 +78,12 @@ const NewShift = () => {
...
       },
-      name: {
+      description: {
         type: "string",
-        title: "Shift name (i.e. the Night Shift)",
+        title: "Shift's description",
       },
     },
-    required: ["group_id", "start_date", "end_date"],
+    required: ["group_id", "name", "start_date", "end_date"],
```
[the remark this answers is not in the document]
Theodlz: II don't think so. We were thinking that giving a name to a shift should be mandatory, but the description optional. What do you think ?

skyportal/handlers/api/shift.py
```diff
@@ -41,7 +45,23 @@ def post(self):
               application/json:
                 schema: Error
         """
+
         data = self.get_json()
+        try:
+            shift_admin_ids = [int(e) for e in data.get('shift_admins', [])]
```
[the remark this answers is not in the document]
Theodlz: Sure. I'll do it

Theodlz:
> Looking very good. Some small changes suggested. Please also add some data as appropriate to the demo data. Also, will need some new tests.

Hi Josh, only thing missing now is the demo data and the tests, but otherwise we addressed your comments the best we can, and fixed other "bugs" that are more behaviors that didn't meet our requirements.

Theodlz:
Hi @guynir42 @profjsb, could you have a look at the new changes please ? Thanks for your help.

skyportal/tests/api/test_shifts.py
```diff
@@ -8,17 +8,18 @@ def test_shift(public_group, super_admin_token, view_only_token, super_admin_use
...
 
-    status, data = api('POST', 'shift', data=request_data, token=super_admin_token)
+    status, data = api('POST', 'shifts', data=request_data, token=super_admin_token)
     assert status == 200
     assert data['status'] == 'success'
 
-    status, data = api('GET', f'shift/{public_group.id}', token=super_admin_token)
+    status, data = api('GET', f'shifts/{public_group.id}', token=super_admin_token)
     assert status == 200
     assert data['status'] == 'success'
```
guynir42, line 24: Maybe check that the shift you posted is in fact recovered. I'd add a uuid name for the shift and make sure you can find that name among one of the returned shifts.

static/js/ducks/shifts.js
```diff
@@ -8,8 +8,34 @@ const FETCH_SHIFTS_OK = "skyportal/FETCH_SHIFTS_OK";
...
+    admin,
+    shift_id,
+  });
+}
+
+export const updateShiftUser = (shiftID, params) =>
+  API.PATCH(`/api/shifts/${shiftID}/users`, UPDATE_SHIFT_USER, params);
+
+export function deleteShiftUser({ userID, shift_id }) {
+  console.log(userID, shift_id);
```
guynir42, line 32: no need for console.log.

guynir42 (commented):
I have minor comments. This looks very close to going in.

profjsb (commented):
Looking good (i agree with @guynir42's comments). Will take a look after @guynir42 approves.

skyportal/handlers/api/shift.py
```diff
@@ -53,6 +103,9 @@ def post(self):
             )
 
         DBSession().add(shift)
+        DBSession().add_all(
+            [ShiftUser(shift=shift, user=user, admin=True) for user in shift_admins]
```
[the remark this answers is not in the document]
Theodlz: It works in the same way as the groups. When creating a group, the current user will be added as its admin, unless you specific an array of sfhit_admins. Later on, using the ShiftUserHandler, you can post non admin users to add them to a shift.

skyportal/tests/api/test_shifts.py
Theodlz: That's a great idea, i'll do that now

static/js/ducks/shifts.js
Theodlz: Yep. Forgot to delete it earlier

skyportal/handlers/api/shift.py
```diff
@@ -36,13 +51,48 @@ def post(self):
...
+        except ValueError:
+            return self.error(
+                "Invalid shift_admins field; unable to parse all items to int"
+            )
+        shift_admins = (
+            User.query_records_accessible_by(self.current_user)
+            .filter(User.id.in_(shift_admin_ids))
+            .all()
+        )
+        if self.current_user not in shift_admins and not isinstance(
```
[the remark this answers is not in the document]
Theodlz: You're right. I mean the shift_admins list is a list of user IDs after all, so thats what should be tested. I think the same changes should be applied in the GroupHandler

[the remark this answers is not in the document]
Theodlz: I think you mentioned this issue last week. From what i saw, it is not just for the shifts but also for the groups and other handlers. Should we just decide a general / common way to handle that in all handlers ?

skyportal/tests/api/test_shifts.py
```diff
@@ -1,13 +1,20 @@
...
+import uuid
 
 
 def test_shift(public_group, super_admin_token, view_only_token, super_admin_user):
 
+    name = str(uuid.uuid4())
+    start_date = date.today().strftime("%Y-%m-%dT%H:%M:%S")
+    end_date = (date.today() + timedelta(days=1)).strftime("%Y-%m-%dT%H:%M:%S")
+    print(start_date)
+    print(end_date)
```
guynir42, line 12:
```suggestion
```

skyportal/tests/api/test_shifts.py
```diff
@@ -23,6 +30,10 @@ def test_shift(public_group, super_admin_token, view_only_token, super_admin_use
     assert status == 200
     assert data['status'] == 'success'
 
+    assert any([request_data['name'] == s['name'] for s in data["data"]])
     assert any([request_data['start_date'] == s['start_date'] for s in data["data"]])
-
     assert any([request_data['end_date'] == s['end_date'] for s in data["data"]])
+    assert (
+        len([s for s in data['data'][0]['users'] if s['id'] == super_admin_user.id])
```
guynir42, line 37: Can't guarantee that your shift will be at index 0. would probably need to do two loops, one with `if s['name']==name` and one inside that to check the users.

guynir42 (commented):
Looks good. Two minor comments.

skyportal/tests/api/test_shifts.py
Theodlz: Oh yes you're very right i tried that first and forgot to add the 2nd loop. Will do it now

Theodlz:
> @Theodlz also please re-run any failing tests to make sure the failures are unrelated.

Hey @guynir42 , i keep getting an error related to the migrations. I'll redo that part entirely soon.

PR #2690, opened 2022-02-23
sncosmo bump to 2.7.0 with migration script

profjsb:
Here's a diff between the two BANDPASSES in sncosmo 2.5.0 and 2.7.0:
image

Create a migration script that adds the enums in the appropriate order to match the 2.7.0 expectation.


profjsb:
Failures appear unrelated. Model logic test passes locally and frontend_pt2 failure appears locally but is related to a recent merged PR.

mcoughlin:
@profjsb This will be a great template for the future.

mcoughlin (approved):
Note to self on this trick of X after Y. Really nice.

PR #2697, opened 2022-02-24
Make observation plan GIF

mcoughlin:
This PR adds the ability to download a GIF of the observation plans.


skyportal/facility_apis/observation_plan.py
```diff
@@ -366,7 +366,7 @@ def delete(request_id):
                 "default": "tiling",
             },
             "exposure_time": {"type": "string", "default": "300"},
-            "filters": {"type": "string", "default": "g,r,i"},
+            "filters": {"type": "string", "default": "ztfg,ztfr,ztfi"},
```
guynir42, line 369: Do we have to have the ZTF in the defaults? Can we easily pull this value from the config?

skyportal/handlers/api/observation_plan.py
```diff
@@ -420,8 +433,211 @@ def get(self, observation_plan_request_id):
...
+        with open(f.name, mode='rb') as g:
+            anim_content = g.read()
+
+    return {
+        "success": True,
+        "name": f"{localization.localization_name}.{output_format}",
+        "data": anim_content,
+        "reason": "",
+    }
+
```
guynir42, line 572: Honestly I didn't understand much of this function but I trust you've kicked the tires on it, so it is probably ok.

guynir42 (approved):
I didn't see anything obviously problematic, and I assume the animations actually work. 
I'm not super happy about having ZTF filters as the default. If it is not to complicated, I'd add a `default_filters` to the config and pull those out. As a default to pulling from the config (ie if there is no `default_filters` key, we should probably use g,r,i or U, B, V, but that would not matter much if you assign the ZTF filters as the default in the config.

skyportal/facility_apis/observation_plan.py
mcoughlin: great idea. fixed.

skyportal/handlers/api/observation_plan.py
mcoughlin: yeah it works ok. Will probably ask a student to beautify it more down the road once it gets some more use.

mcoughlin:
@guynir42 Please merge if happy.

skyportal/facility_apis/observation_plan.py
```diff
@@ -8,11 +8,19 @@
...
 
 log = make_log('api/observation_plan')
 
+env, cfg = load_env()
+
+if cfg['app.observation_plan.default_filters'] is None:
+    default_filters = ['g', 'r', 'i']
+else:
+    default_filters = cfg['app.observation_plan.default_filters']
+
```
guynir42, line 23:
```suggestion
env, cfg = load_env()

default_filters = cfg.get('app.observation_plan.default_filters', ['g', 'r', 'i']):
```

mcoughlin:
@guynir42 Merge if happy?

PR #2700, opened 2022-02-27
Change sqlalchemy queries to use select

mcoughlin:
This PR replaces the query syntax to use select (see e.g. https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#orm-query-unified-with-core-select).


stefanv:
@mcoughlin Can you make a PR with the baselayer changes to that repo, so that I can review it there? Thanks!

PR #2702, opened 2022-02-27
Refactor Tag usage

mcoughlin:
This PR refactors the GcnEvent tags, which are used in both the RecentGcnEvents widget and  GcnEvent page, so this uses the same code both places.


guynir42 (approved):
LGTM

static/js/components/GcnEventPage.jsx
```diff
@@ -39,6 +39,8 @@ import GcnSelectionForm from "./GcnSelectionForm";
 import ObservationPlanRequestForm from "./ObservationPlanRequestForm";
 import ObservationPlanRequestLists from "./ObservationPlanRequestLists";
 
+import RenderTags from "./RenderTags";
```
stefanv, line 42: Since this is specific to Gcn, consider including Gcn in the component name e.g. `GcnTags`.

stefanv (commented):
Suggested a naming change, otherwise looks all good.

static/js/components/GcnEventPage.jsx
mcoughlin: Good call! Fixed!

mcoughlin:
@stefanv please merge if happy?

PR #2727, opened 2022-03-01
DNM: improve shift test

mcoughlin:
Trying to understand why the tests make no sense.


mcoughlin:
No longer relevant given the evolution in shift work.

PR #2731, opened 2022-03-02
update npm/geckodriver

profjsb:
harmonize the geckodriver and npm to that of the GHA in baselayer

### DIFF
--- a/.github/workflows/test_api_and_frontend.yaml
+++ b/.github/workflows/test_api_and_frontend.yaml
@@ -81,7 +81,7 @@ jobs:
 
       - name: Install system dependencies
         run: |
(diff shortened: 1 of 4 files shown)


stefanv:
```
The conflict is caused by:
    The user requested dask==2022.1.1
    distributed 2022.2.1 depends on dask==2022.02.1
```

in the migrations.  Perhaps this is the migration pin failing.

stefanv:
Looks like all the checkouts have `submodules: true`, so *should* not be a problem.

stefanv:
I think we need to merge this to fix the master checkout.  It may require further work, but I want to get this in so long.

PR #2732, opened 2022-03-02
filter sources by annotation

mcoughlin:
This PR enables source filtering by annotations. This is important as it allows one to identify sources by, for example, thresholding on redshift values or other numerical annotations.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -502,8 +502,14 @@ class SourceHandler(BaseHandler):
             explode: false
             style: simple
             description: |
-              Comma-separated string of "annotation: value: operator" pair(s) to filter for sources matching
+              Comma-separated string of "annotation: value: operator" triplet(s) to filter for sources matching
               that/those annotation(s), i.e. "redshift: 0.5: lt"
+          - in: query
+            name: annotationsFilterOrigin
+            nullable: true
+            schema:
+              type: string
+            description: Comma-separated string of who posted the annotations to filter
           - in: query
             name: minRedshift
             nullable: true
@@ -657,6 +663,9 @@ class SourceHandler(BaseHandler):
         )
         classifications = self.get_query_argument("classifications", None)
         annotations_filter = self.get_query_argument("annotationsFilter", None)
+        annotations_filter_origin = self.get_query_argument(
+            "annotationsFilterOrigin", None
+        )
         min_redshift = self.get_query_argument("minRedshift", None)
         max_redshift = self.get_query_argument("maxRedshift", None)
         min_peak_magnitude = self.get_query_argument("minPeakMagnitude", None)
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/source.py
```diff
@@ -491,6 +492,18 @@ def get(self, obj_id=None):
...
+            name: annotationsFilter
+            nullable: true
+            schema:
+              type: array
+              items:
+                type: string
+            explode: false
+            style: simple
+            description: |
+              Comma-separated string of "annotation: value: operator" pair(s) to filter for sources matching
```
guynir42, line 505: pairs -> triplets

skyportal/handlers/api/source.py
```diff
@@ -1366,6 +1389,56 @@ class Validator(Schema):
...
+
+                        # fails the filter if name is not present
+                        if not name_check:
+                            passes_filter = False
+                            break
+                        if len(ann_split) == 3:
+                            index = name_present.index(True)
+                            data_value = annotations[index].data[name]
+                            try:
+                                comp_function = eval(f"operator.{op}")
```
guynir42, line 1432: I'm a little uneasy about this `eval`. Are there really that many operations that we can't just preconstruct a dictionary of keywords -> functions instead?

skyportal/handlers/api/source.py
```diff
@@ -491,6 +492,18 @@ def get(self, obj_id=None):
...
+            schema:
+              type: array
+              items:
+                type: string
+            explode: false
+            style: simple
+            description: |
+              Comma-separated string of "annotation: value: operator" pair(s) to filter for sources matching
+              that/those annotation(s), i.e. "redshift: 0.5: lt"
           - in: query
```
guynir42, line 507: Add `annotationsFilterOrigin` that will only accept specific origin(s) for filtering with `annotationsFilter`. 
(assume some rogue origin is posting unrealistic results that you want to avoid)

skyportal/tests/api/test_sources.py
```diff
@@ -2027,3 +2027,39 @@ def test_token_user_retrieving_source_with_period_exists(
...
+
+    status, data = api(
+        "GET",
+        "sources",
+        params={"annotationsFilter": "period:2.0:ge"},
+        token=view_only_token,
+    )
+    assert status == 200
+    assert data["status"] == "success"
+    assert len(data["data"]["sources"]) == 0
```
guynir42, line 2065: This will fail if any of the other tests happen to post a period annotation. 
One option is to replace the keyword `period` with a uuid string. 
Another option is to make a uuid or other unique string for the origin and use `annotationsFilterOrigin`. 
I'd like to see the test both with and without the origin filter.

guynir42 (commented):
This looks good. I think this will have a lot of usage. I do have a couple of comments:

skyportal/handlers/api/source.py
mcoughlin: i think i got this one.

skyportal/handlers/api/source.py
mcoughlin: added.

skyportal/tests/api/test_sources.py
mcoughlin: hopefully i did what you had in mind..

skyportal/handlers/api/source.py
```diff
@@ -491,6 +492,24 @@ def get(self, obj_id=None):
...
+            style: simple
+            description: |
+              Comma-separated string of "annotation: value: operator" triplet(s) to filter for sources matching
+              that/those annotation(s), i.e. "redshift: 0.5: lt"
+          - in: query
+            name: annotationsFilterOrigin
+            nullable: true
+            schema:
+              type: string
+            description: Comma-separated string of who posted the annotations to filter
```
guynir42, line 512: Origin is not exactly the same as who posted. I would change it to "comma separated string of origins. Only annotations from these origins are used when filtering with the annotationsFilter"

guynir42 (approved):
Minor comment (on the docs, naturally) and good to go.

skyportal/handlers/api/source.py
mcoughlin: Fixed.

guynir42 (approved):

PR #2735, opened 2022-03-04
Display names of users in options when uploading a spectra

jadalilleboe:
Display a user's full name in the option for adding observers/reducers when uploading a spectra
Screenshot (39)
Screenshot (40)

### DIFF
--- a/static/js/components/UploadSpectrum.jsx
+++ b/static/js/components/UploadSpectrum.jsx
@@ -295,7 +295,10 @@ const UploadSpectrumForm = ({ route }) => {
           type: "integer",
           anyOf: users?.map((user) => ({
             enum: [user.id],
-            title: user.username,
+            title:
+              user?.first_name && user?.last_name
+                ? `${user.username} (${user.first_name} ${user.last_name})`
+                : `${user.username}`,
           })),
         },
         uniqueItems: true,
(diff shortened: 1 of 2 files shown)


profjsb (approved):

profjsb:
You might check if there's a need to line wrap in the case of very long names/username

static/js/components/UploadSpectrum.jsx
```diff
@@ -313,7 +316,10 @@ const UploadSpectrumForm = ({ route }) => {
           type: "integer",
           anyOf: users?.map((user) => ({
             enum: [user.id],
-            title: user.username,
+            title:
+              user?.first_name && user?.last_name
+                ? `${user.username} (${user.first_name} ${user.last_name})`
+                : `${user.username}`,
```
stefanv, line 322:
```suggestion
            title: `${user.username} (${first ? first + " " : ""}${last ? last : ""})`,
```

stefanv: Same below.

Otherwise, if either first or last name is missing, the real name won't render.

jadalilleboe: fixed!

jadalilleboe:
@profjsb Doesn't seem like there's an easy way to add a line break with rjsf form? looking at [this issue](https://github.com/rjsf-team/react-jsonschema-form/issues/2562)

stefanv (approved):

mcoughlin:
Given the two approvals, and that react-jsonschema-form does not yet support line breaks, I think we merge this and return to line breaking if we find there are too many long names given trouble here.

PR #2736, opened 2022-03-04
Photometric Series (model)

guynir42:
Add a new table for photometric series, where multiple observations can be saved as one database object. 
The model works by saving a file and keeping that file's name in the DB, while most of the actual measurements are saved to disk storage. 
Searching on these objects can still be done using summary statistics like MJD start/end, mean magnitude, etc.


guynir42:
I will add the migrations when we are happy with how the model looks.

profjsb (commented):
looks good. A few suggestions inline. 

One thing I do think you'll need to change is the requirement that the series is taken at regular intervals. Very few telescopes do this faithfully to the rms < 1 sec level. Instead, I suggest just calculating the start and end times of the series, and then save medium delta t between integrations. Relatedly dont require an exptime since that could also differ in a given series. (exptime could be an optional column in the series)

I think you should be saving this data as hdf5 using `pytables` or `h4py` and not saving it as a np file (using np.save). That way you can store metadata in the file and turn on compression. Think about a sensible naming convention for the file and sensible sub-directory names (dont want everything in one big directory). Allowing the user to set the filename is worrisome/security risk since they could have unsanitized `run_id` screw up our filesystem.

To protect against saving the exact data already uploaded you could index the md5sum hash of the timeseries.

skyportal/models/photometry.py
```diff
@@ -248,3 +252,621 @@ def snr(self):
...
+    continuously from mjd_start to mjd_end.
+
+    To initialize this function user must provide:
+    - data: a dictionary, pandas DataFrame, or structured arry
+            The data must contain an "mjds" column, and either
+            a "fluxes" or a "mags" column.
+            Other columns can include "fluxerr" or "magerr",
+            and any other measurements that are stored but
+            generally not used by the system like raw counts,
+            backgrounds, etc.
```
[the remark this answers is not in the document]
guynir42: The RA/Dec of the series is supposed to be constant for each object, although I agree that for moving objects this could conceivably be a variable.

skyportal/models/photometry.py
```diff
@@ -248,3 +252,621 @@ def snr(self):
...
+    - run_id: a unique identifier of the set of images from
+            which this series was extracted. This is useful
+            for finding other objects that were observed
+            at the same time as this series.
+    - run_obj_id: a unique index or name of the object inside
+            this observation series. Can be an index of the
+            star in the images, or an official designation.
+            Does not have to be the same as the obj_id,
+            and for the same phyical object, it can be
+            a different identifier in different runs.
```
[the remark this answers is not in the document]
guynir42: Maybe series_id?

skyportal/models/photometry.py
```diff
@@ -248,3 +252,621 @@ def snr(self):
...
+        """
+        Load the underlying photometric data from disk.
+        """
+        self.data = np.load(os.path.join(self.folder, self.filename))
+
+    def save_data(self):
+        """
+        Save the underlying photometric data to disk.
+        """
+        np.save(os.path.join(self.folder, self.filename), self.data)
```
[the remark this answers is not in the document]
guynir42: I would, but wouldn't that just raise an exception that's very similar to what I would raise if the folder was inaccessible?

guynir42:
Here are some issues I'm not sure about: 
1) File type. @stefanv indicated he liked saving the numpy array as npz, which I think we can even do with npy because all the data is a single structured array. @profjsb has indicated he prefers using HDF5. I think we should make a decision or leave it as a configurable with some options. 
2) Directory structure. The config defines a root folder for all photometry data. Under than could be other partitions, followed in the end by the `series_id` as a folder and then the filename based on the object id in the series. Do you want other partitions like folder per user? We can also allow the `series_id` be composed of several levels of folders, e.g., `series_id=2022-03-04/ecliptic_run1` is the convention I use in W-FAST. 
3) Should the model also be responsible for deleting (empty) folders when removing a file? I think this is probably safer to leave to some housekeeping chron job. 
4) Require RA/Dec and exposure times to be (generally) uniform across time-series. I think it would get a little weird if we allow these sizes to change dramatically over the course of a series. For example: how to calculate the start/end time of a series when given the mjd of each mid-exposure, if you don't know the exposure time of the first and last exposure. 
One possible solution is either get the exp time directly, or get it from the data array. 
Another option is not save the series start/end but instead save `mjd_first` and `mjd_last` which is just the agnostic mjd of the first/last images, regardless of the exposure time. This also lets the user input the `mjds` of exposure middle or start, as long as they are consistent about it. 
I should mention that it is really important to have __some__ value of exp_time and RA / Dec because we save them on the DB and possibly filter on them or index by them. Maybe I should just check if there is a data field called exp_time and if so, automatically assign the DB exp_time (for the whole series) by calculating the median exposure time (and the same for RA/Dec)? 

Anything else I am missing?

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+
+    def __init__(self, data, series_id, series_obj_id):
+        self.series_identifier = str(series_id)
+        self.series_obj_id = str(series_obj_id)
+
+        if len(data) == 0:
+            raise ValueError('Must supply a non-empty data structure.')
+
+        # verify data has all required fields
+        if 'fluxes' not in data and 'mags' not in data or 'mjds' not in data:
```
profjsb, line 319: makes it seem like `data` has to be dictionary-like with keys and values. Maybe also test for the instance types you allow for data

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+        self.mjd_last_detected = self.mjd[detection_indices[-1]]
+        self.num_exp = len(self.mjds)
+
+        # calculate the mean mag, rms and robust median/rms
+        self.mean_mag = np.nanmean(self.mags)
+        self.rms_mag = np.nanstd(self.mags)
+        (self.robust_mag, self.robust_rms) = self.sigma_clipping(self.mags)
+
+        # if RA, Dec or exposure time are given in the auxiliary data:
+        for key in ['RA', 'ra']:
```
profjsb, line 493:
```suggestion
        for key in ['RA', 'ra', 'Ra']:
```

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+            if key in self._data:
+                self.ra = self._data[key].median()
```
profjsb, line 495: should be nanmedian in case this is missing in some entries

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+                break
+        for key in ['Dec', 'DEC', 'dec']:
+            if key in self._data:
+                self.dec = self._data[key].median()
```
profjsb, line 499: should be nanmedian in case this is missing in some entries

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+        for key in [
+            'exptime',
+            'exp_time',
+            'exposure',
+            'exposure_time',
+            'EXPTIME',
+            'EXP_TIME',
+        ]:
+            if key in self._data:
+                self.exp_time = self._data[key].median()
```
profjsb, line 510: should be nanmedian in case this is missing in some entries

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+        sa.String, nullable=False, default='0', doc='Channel of the photometric series.'
+    )
+
+    filter = sa.Column(
+        allowed_bandpasses,
+        nullable=False,
+        doc='Filter with which the observation was taken.',
+    )
+
+    exp_time = sa.Column(
```
profjsb, line 697: if exp_time is missing in the data then this will be None -- it should be nullable

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+        sa.Float, nullable=False, doc='Median exposure time of each frame, in seconds.'
+    )
+
+    frame_rate = sa.Column(
```
profjsb, line 701: where is this calculated? This maybe None -- it should be nullable

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+
+    time_stamp_alignment = sa.Column(
+        time_stamp_alignment_types,
+        nullable=False,
+        doc='When in each exposure is the mjd timestamp measured: start, middle, or end.',
+        default='middle',
+    )
+
+    ra_unc = sa.Column(sa.Float, doc="Uncertainty of ra position [arcsec]")
+    dec_unc = sa.Column(sa.Float, doc="Uncertainty of dec position [arcsec]")
```
profjsb, line 719: These are calculated right? if ra_unc, dec_unc is missing in the data then this will be None -- it should be nullable

skyportal/models/photometry.py
```diff
@@ -248,3 +263,720 @@ def snr(self):
...
+
+    altdata = sa.Column(JSONB, doc="Arbitrary metadata in JSON format..")
```
profjsb, line 721:
```suggestion
    altdata = sa.Column(JSONB, doc="Arbitrary metadata in JSON format.")
```

profjsb (changes requested):

skyportal/models/photometry.py
guynir42: The xarray median skips NaNs by default.

guynir42:
@profjsb I'll fix the merge conflicts and migration tests along with any code comments.

profjsb:
@guynir42 please fix conflicts and try to get the tests to pass (or attest that failures are unrelated)

profjsb:
Needs migration fix too it looks like

PR #2737, opened 2022-03-05
Enable instrument file uploads

mcoughlin:
This PR adds the ability to upload instrument files from the webpage.

### DIFF
--- a/skyportal/handlers/api/instrument.py
+++ b/skyportal/handlers/api/instrument.py
@@ -10,6 +10,8 @@ from regions import Regions
 from astropy import coordinates
 from astropy import units as u
 import numpy as np
+import pandas as pd
+from io import StringIO
 
 from ..base import BaseHandler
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/instrument.py
```diff
@@ -71,6 +73,21 @@ def post(self):
...
+                )
+
+            if not all(
+                k in field_data
+                for k in [
+                    'ID',
+                    'RA',
+                    'Dec',
+                ]
+            ):
```
guynir42, line 88: Again styling choice but I feel this is more readable (especially after linting).
```suggestion
            if not {'ID', 'RA', 'Dec'}.issubset(field_data):
```

guynir42 (approved):
Very minor change. I assume you've tested these components?

mcoughlin:
Green check mark and very minor edits. Merging.

PR #2738, opened 2022-03-05
Add a follow-up request page

mcoughlin:
This PR adds a page to display follow-up requests.

PR #2739, opened 2022-03-06
Add a basic scheduler download button

mcoughlin:
This PR builds off of https://github.com/skyportal/skyportal/pull/2738 to add a button to download a possible schedule.


skyportal/handlers/api/followup_request.py
```diff
@@ -596,3 +622,406 @@ def delete(self, request_id):
...
+def observation_schedule(
+    followup_requests,
+    instrument,
+    observation_start=Time.now(),
+    observation_end=Time.now() + TimeDelta(12 * u.hour),
+    output_format='csv',
+    figsize=(10, 8),
+):
+
+    """Create a movie to display observations of a given skymap
```
guynir42, line 636: Movie?

guynir42 (approved):
Just a minor documentation comment/question. Looks good overall. This will be very cool when we start using it.

skyportal/handlers/api/followup_request.py
mcoughlin: If we were a movie.... you'd be the right guy!

mcoughlin:
Minor documentation comment, which was addressed. Merging.

PR #2741, opened 2022-03-07
Add a new source from the frontend

mcoughlin:
This PR adds a front-end element to post a source.

### DIFF
--- a/static/js/components/NewSource.jsx
+++ b/static/js/components/NewSource.jsx
@@ -1,0 +1,130 @@
+import React, { useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
+import Form from "@rjsf/material-ui";
+import Paper from "@material-ui/core/Paper";
+import Typography from "@material-ui/core/Typography";
+import PropTypes from "prop-types";
+import { makeStyles } from "@material-ui/core/styles";
+
+import { showNotification } from "baselayer/components/Notifications";
+
+import dayjs from "dayjs";
+import utc from "dayjs/plugin/utc";
+
+import GroupShareSelect from "./GroupShareSelect";
+import { saveSource } from "../ducks/source";
+import { fetchSources } from "../ducks/sources";
+
+dayjs.extend(utc);
+
+const useStyles = makeStyles((theme) => ({
+  header: {},
+  eventListContainer: {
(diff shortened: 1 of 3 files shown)


profjsb:
Please show screenshot/captures of the interaction in the PR. It will be easier to evaluate.

mcoughlin:
Screen Shot 2022-03-07 at 7 12 33 AM

Just a simple frontend element.

profjsb (commented):
I do not see how this page(?) is currently exposed anywhere ... is meant to be embedded in another page perhaps? Or should this be added to the config.default sidebar?

This PR should not be made from SP itself, but from your fork of SP. Please delete this branch.

static/js/components/NewSource.jsx
```diff
@@ -0,0 +1,130 @@
...
+  const handleSubmit = async ({ formData }) => {
+    const result = await dispatch(saveSource(formData));
+    if (result.status === "success") {
+      dispatch(showNotification("Source saved"));
+      dispatch(fetchSources());
+    }
+  };
+
+  function validate(formData, errors) {
+    if (formData.ra < 0 || formData.ra > 360) {
```
profjsb, line 69:
```suggestion
    if (formData.ra < 0 || formData.ra >= 360) {
```

mcoughlin:
@profjsb I had dropped it on the frontpage as a test, I am not really sure where it should go.

profjsb:
I guess I'm stuck on what this PR is trying to accomplish. Do we have a story/discussion about this? 

Also totally unclear what the behavior should be if the source with that name exists already...should it get rejected or somehow appended to that source? What if the new source is at/consistent with the same location of an existing source? What if the end user wants to add more data at creation time (like a note or astrometric errors)?

mcoughlin:
@profjsb Yes these are interesting points. However, currently, if one wants a source, one MUST go through the API currently, which seems kind of crazy. ATLAS or someone detects a transient and wants to run forced photometry using the tools we have... and yet still need to go through the API to make a source? 

Do we have any restrictions on this behavior currently? I.e. what makes a source a duplicate?

mcoughlin:
Closed in favor of https://github.com/skyportal/skyportal/pull/2743

PR #2743, opened 2022-03-07
new source widget

mcoughlin:
This PR adds a new source widget to the frontend.


profjsb (changes requested):
1) Entry for dec and submit is spilling outside the widget for all viewport sizes:
Kapture 2022-03-10 at 11 26 44

Consider making more compact the entries.

2) Missing hamburger icon to move widget elsewhere on the dashboard. See how other widgets are implemented.

3) This widget should be checking the object ID and RA/DEC for potential conflicts. One idea is to use a popup...

4) Popup confirm and take user to the newly formed source as such:
Kapture 2022-03-10 at 12 44 52

mcoughlin:
@profjsb did you mock this up in that gif?? I am having some difficulty with the handling of the checking for existing sources.

mcoughlin:
@profjsb I wrote a custom endpoint to check for existing sources and push those messages back. I hope that suffices.

skyportal/handlers/api/source_exists.py
```diff
@@ -0,0 +1,87 @@
...
+            description: Radius for spatial filtering if ra & dec are provided (in decimal degrees)
+        """
+
+        ra = self.get_query_argument('ra', None)
+        dec = self.get_query_argument('dec', None)
+        radius = self.get_query_argument('radius', None)
+
+        if obj_id is not None:
+            s = Obj.get_if_accessible_by(obj_id, self.current_user)
+            print(obj_id)
```
guynir42, line 57:
```suggestion
```

skyportal/handlers/api/source_exists.py
```diff
@@ -0,0 +1,87 @@
...
+            if len(objs) == 1:
+                return self.success(
+                    f"A source at that location already exists: {objs[0].id}."
+                )
+            elif len(objs) > 1:
+                return self.success(
+                    f"Sources at that location already exist: {','.join([obj.id for obj in objs])}."
+                )
+
+        return self.success(False)
```
guynir42, line 87: Don't we already have this functionality in the source handler using the HEAD request?

static/js/components/NewSource.jsx
```diff
@@ -0,0 +1,104 @@
...
+            <GroupShareSelect
+              groupList={allGroups}
+              setGroupIDs={setSelectedGroupIds}
+              groupIDs={selectedGroupIds}
+            />
+            <Form
+              schema={sourceFormSchema}
+              onSubmit={handleSubmit}
+              // eslint-disable-next-line react/jsx-no-bind
+              validate={validate}
```
guynir42, line 87: why not `liveValidate`?

guynir42 (commented):
My main concern is that you may be able to get the same functionality using the source handler HEAD request.

skyportal/handlers/api/source_exists.py
mcoughlin: We also need to check the nearby RA and Dec per Josh's request.

guynir42:
Also, when playing around with it I noticed a couple of small issues.
1) the first field "Share Data With": the text is too big for the box. 
2) When you add a source, it complains that it cannot find the source. Also the photometry and spectroscopy spinners go on forever. I think these might be related. Maybe we should make sure that a source without photometry/spectroscopy is handled gracefully, and also the spinners should turn into "no photometry" or "no spectroscopy" messages.

skyportal/handlers/api/source_exists.py
guynir42: The source handler doesn't do a cone search?

mcoughlin:
@guynir42 Yeah for now I am throwing an "error" for the source catch. Not sure the best way to catch that. The spinners end for me though... what are you seeing?

static/js/components/NewSource.jsx
mcoughlin: It's a little annoying the way it whines immediately once you input a name. maybe see if you prefer with or without it?

skyportal/handlers/api/source_exists.py
mcoughlin: The head handler doesn't.

guynir42:
Spinners seem to remain forever, I don't know why. 

I urge you to consider adding the cone search to sources HEAD, we have too many handlers as it is. 

I think there must be a simple way to return a source or nothing without popping up an error.

mcoughlin:
@guynir42 HEAD doesn't seem to be an API call available in the duck.

guynir42:
@mcoughlin that's weird. I think this functionality is used for candidates, but maybe it is only called internally. 
Did you find an alternative to returning an error if the source already exists / doesn't exist?

mcoughlin:
@guynir42 Where do you actually see this displayed?

mcoughlin:
@guynir42 maybe we hop on zoom briefly. I don't quite see it but I suspect just I don't know where to see.

guynir42:
sure let's do it around 9

mcoughlin:
@guynir42 can you see if this fix does it for you?

guynir42:
The photometry and spectroscopy look good. The "Source not found" error still pops up.

mcoughlin:
I am not sure how to block this one without screwing the normal behavior up... @guynir42

guynir42:
Is the error produced by the photometry handler?

mcoughlin:
Source handler.

guynir42:
Is it produced in any situation where there is no photometry?

mcoughlin:
I think so?

guynir42:
That doesn't sound like the right behavior to me. Try to see if there's a better criteria for returning such an error. Otherwise maybe we can live without this error? In what situations do we need that to return an error anyway?

mcoughlin:
@guynir42 can i ask if it triggers for you if we remove the push.history call in the NewSource handler?

guynir42:
It is unrelated to the new source. It pops up whenever I refresh the source page (presumably for sources without photometry).

mcoughlin:
And goes away if you add photometry? I can't seem to reproduce this.

mcoughlin:
@jadalilleboe what do you see?

jadalilleboe:
I don't see a source not found error.

guynir42:
hmmm let me dig into this. I thought you were seeing this error as well.

guynir42:
@mcoughlin it seems to only pop up for sources with a space in the name. The handlers that return the error are photometry and spectroscopy. This probably has to do with the parsing of the source ID. I'd say this is not really relevant to this PR.

guynir42:
I did notice, that when there are errors on the submit new source form, it drives the rest of the input fields off the bottom of the component.

mcoughlin:
@guynir42 do you think the benefit of live validate offsets the UI or should we just turn it back off?

guynir42:
If the behavior was ok without the live validate, I think we can live without it (there aren't a lot of inputs and the errors are fairly obvious).

mcoughlin:
@guynir42 Ok removing for now.

guynir42 (approved):

PR #2744, opened 2022-03-08
Add clear spectra and add all spectra buttons to spectroscopy plot

jadalilleboe:
Adds two buttons for clearing the spectra on a spectroscopy plot and adding them back in:
ezgif com-gif-maker


mcoughlin (approved):
@jadalilleboe Played around with this on the demo data, really intuitive. That coupled with the div changes on fritz-marshal should really improve the user experience here. Thanks!

PR #2746, opened 2022-03-09
Ztf forced photometry

mcoughlin:
This PR adds an example of extracting ZTF forced photometry from the ZTF forced photometry service.

### DIFF
--- a/doc/mma.md
+++ b/doc/mma.md
@@ -14,14 +14,14 @@ To evaluate the efficacy of the executed observation plans, we have the `Execute
 
 The vast majority of follow-up instruments will require some form of authentication. All such information is passed through the `altdata` variable of the `Allocation`s API. We briefly describe the authentication form the available telescopes take below:
 
-* ATLAS Forced Photometry: A user account must be made on https://fallingstar-data.com/forcedphot/, at which point the authentication takes the form {"api_token": "testtoken"}.
-* KAIT: A username and password are passed as {"username": "username", "password": "password"}.
-* LCO: A user account must be made on https://lco.global/, at which point the authentication takes the form {"PROPOSAL_ID": "your_proposal_ID", "API_TOKEN": "testtoken"}.
-* LT: A proposal ID, username, and password are passed as {"username": "username", "password": "password", "LT_proposalID": "your_proposal_ID"}.
-* SLACK: As discussed further [here](./slack.html), slack information is pass as {"slack_workspace": "XXX", "slack_channel": "YYY", "slack_token": "ZZZ"}.
-* SWIFT: A username and password are passed as {"username": "username", "secret": "password"}.
-* ZTF Triggering: An API token for an admin user for [Kowalski](https://github.com/dmitryduev/kowalski) can be passed as {"access_token": "your_token"}.
-* ZTF Forced Photometry: A user account must be made with the ZTF forced photometry service (see https://zwicky.tf/vgt). The authentication then takes the form _altdata: '{"ipac_http_user": "http_user", "ipac_http_password": "http_password", "ipac_email": "email", "ipac_userpass": "password"}' where http_user and http_password are provided in the documentation and ipac_email and ipac_userpass are provided for the specific account. Note that IPAC's forced photometry database is updated once per hour, on the hour, and requests will only be available after this update.
+* ATLAS Forced Photometry: A user account must be made on https://fallingstar-data.com/forcedphot/, at which point the authentication takes the form `{"api_token": "testtoken"}`.
+* KAIT: A username and password are passed as `{"username": "username", "password": "password"}`.
+* LCO: A user account must be made on https://lco.global/, at which point the authentication takes the form `{"PROPOSAL_ID": "your_proposal_ID", "API_TOKEN": "testtoken"}`.
+* LT: A proposal ID, username, and password are passed as `{"username": "username", "password": "password", "LT_proposalID": "your_proposal_ID"}`.
+* SLACK: As discussed further [here](./slack.html), slack information is pass as `{"slack_workspace": "XXX", "slack_channel": "YYY", "slack_token": "ZZZ"}`.
+* SWIFT: A username and password are passed as `{"username": "username", "secret": "password"}`.
+* ZTF Triggering: An API token for an admin user for [Kowalski](https://github.com/dmitryduev/kowalski) can be passed as `{"access_token": "your_token"}`.
+* ZTF Forced Photometry: A user account must be made with the ZTF forced photometry service (see https://zwicky.tf/vgt). The authentication then takes the form `{"ipac_http_user": "http_user", "ipac_http_password": "http_password", "ipac_email": "email", "ipac_userpass": "password"}` where http_user and http_password are provided in the documentation and ipac_email and ipac_userpass are provided for the specific account. Note that IPAC's forced photometry database is updated once per hour, on the hour, and requests will only be available after this update.
 
 ## Uploading executed observations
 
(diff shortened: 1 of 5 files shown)


doc/mma.md
```diff
@@ -20,8 +20,8 @@ The vast majority of follow-up instruments will require some form of authenticat
...
 * SLACK: As discussed further [here](./slack.html), slack information is pass as {"slack_workspace": "XXX", "slack_channel": "YYY", "slack_token": "ZZZ"}.
 * SWIFT: A username and password are passed as {"username": "username", "secret": "password"}.
-* ZTF: An API token for an admin user for [Kowalski](https://github.com/dmitryduev/kowalski) can be passed as {"access_token": "your_token"}.
-
+* ZTF Triggering: An API token for an admin user for [Kowalski](https://github.com/dmitryduev/kowalski) can be passed as {"access_token": "your_token"}.
+* ZTF Forced Photometry: A user account must be made with the ZTF forced photometry service (see https://zwicky.tf/vgt). The authentication then takes the form _altdata: '{"ipac_http_user": "http_user", "ipac_http_password": "http_password", "ipac_email": "email", "ipac_userpass": "password"}' where http_user and http_password are provided in the documentation and ipac_email and ipac_userpass are provided for the specific account. Note that IPAC's forced photometry database is updated once per hour, on the hour, and requests will only be available after this update.
```
guynir42, line 24: is it `_altdata` or `altdata`? Also it would look better with backticks on all the code snippets.

skyportal/facility_apis/ztf.py
```diff
@@ -20,22 +28,26 @@
 else:
     ZTF_URL = f"{cfg['app.ztf.protocol']}://{cfg['app.ztf.host']}:{cfg['app.ztf.port']}"
 
+ZTF_FORCED_URL = cfg['app.ztf_forced_endpoint']
+
 bands = {'g': 1, 'ztfg': 1, 'r': 1, 'ztfr': 2, 'i': 3, 'ztfi': 3}
 inv_bands = {v: k for k, v in bands.items()}
```
guynir42, line 34: I'm not sure how you are inverting the dictionary when there are identical values.

skyportal/facility_apis/ztf.py
```diff
@@ -146,6 +185,124 @@ def _build_observation_plan_payload(self, request):
...
+        df.drop(
+            columns=drop_columns,
+            inplace=True,
+        )
+        df['magsys'] = 'ab'
+
+        data_out = {
+            'obj_id': request.obj_id,
+            'instrument_id': instrument.id,
+            'group_ids': 'all',
```
guynir42, line 282: Maybe groups associated with the current user?

skyportal/facility_apis/ztf.py
```diff
@@ -146,6 +185,124 @@ def _build_observation_plan_payload(self, request):
...
+            **df.to_dict(orient='list'),
+        }
+
+        from skyportal.handlers.api.photometry import add_external_photometry
```
guynir42, line 286: Is there a particular reason the import is down here?

skyportal/facility_apis/ztf.py
```diff
@@ -206,6 +363,123 @@ def delete(request):
...
 
+    @staticmethod
+    def get(request):
+
+        """Get a forced photometry request result from ZTF.
+
+        Parameters
+        ----------
+        request : skyportal.models.FollowupRequest
+            The request to add to the queue and the SkyPortal database.
```
guynir42, line 374: I'm not sure this doc string is the right one...

skyportal/facility_apis/ztf.py
```diff
@@ -206,6 +363,123 @@ def delete(request):
...
+        """
+
+        from ..models import (
```
guynir42, line 377: Any reason these imports are here?

skyportal/facility_apis/ztf.py
```diff
@@ -206,6 +363,123 @@ def delete(request):
...
+        params = urllib.parse.urlencode(requestgroup)
+        url = f"{ZTF_FORCED_URL}/cgi-bin/getForcedPhotometryRequests.cgi?{params}"
+
+        r = requests.get(
+            url,
+            auth=HTTPBasicAuth(
+                altdata['ipac_http_user'], altdata['ipac_http_password']
+            ),
+        )
+        r.raise_for_status()
```
guynir42, line 436: Uncaught exception.

skyportal/facility_apis/ztf.py
```diff
@@ -206,6 +363,123 @@ def delete(request):
...
+                        'Light curve not yet available. Please try again later.'
+                    )
+                else:
+                    lightcurve = row['lightcurve']
+                    dataurl = f"{ZTF_FORCED_URL}/{lightcurve}"
+                    IOLoop.current().run_in_executor(
+                        None,
+                        lambda: commit_photometry(
+                            dataurl, altdata, df_request, req.id, instrument.id
+                        ),
```
guynir42, line 468: So this GET request is going to be posting things to the DB? I'm not sure this is kosher. @stefanv what do you think about this?

skyportal/facility_apis/ztf.py
```diff
@@ -221,28 +495,46 @@ def submit(request):
...
+            r = s.send(prepped)
+            r.raise_for_status()
+        elif request.payload["request_type"] == "forced_photometry":
+            r = requests.get(
+                url,
+                auth=HTTPBasicAuth(
+                    altdata['ipac_http_user'], altdata['ipac_http_password']
+                ),
+            )
+            r.raise_for_status()
```
guynir42, line 537: Are these potentially uncaught exceptions?

guynir42 (commented):
Most of the comments are minor. 
I have one concern which is where you are calling GET which posts the results to the DB. I'm not sure if this is the best way to do that. Perhaps this is worth a discussion?

skyportal/facility_apis/ztf.py
mcoughlin: very good point.

skyportal/facility_apis/ztf.py
mcoughlin: The facility_apis have circular imports up top because of the enum stuff.

skyportal/facility_apis/ztf.py
mcoughlin: always tough.

skyportal/facility_apis/ztf.py
mcoughlin: circular imports.

skyportal/facility_apis/ztf.py
mcoughlin: I feel like handling the return codes below is the correct thing here as we do elsewhere in the app right?

skyportal/facility_apis/ztf.py
mcoughlin: The naming is a bit funny because we are getting the data after posting a request to their server.

skyportal/facility_apis/ztf.py
mcoughlin: good idea!

skyportal/facility_apis/ztf.py
guynir42: I'm not sure. Would `raise_for_status` raise an exception when it is called on `r` that failed?

mcoughlin: yeah thats true. https://stackoverflow.com/questions/61463224/when-to-use-raise-for-status-vs-status-code-testing. So maybe we just let the if statement catch it then?

guynir42: Yes. I think the right thing to do is one of:
- if in a handler function, just check the status code of `r` and `self.error` if needed.
- if in a helper function, you can use `raise_for_status` but make sure to catch that in the handler function.

mcoughlin: yes i went through and nixed them.

skyportal/facility_apis/ztf.py
mcoughlin: good call.

guynir42 (commented):
The specific code changes look ok. I am still a bit worried about posting to the database directly from the handler's get function. I think we should get @stefanv to have a look as well.

PR #2749, opened 2022-03-09
Post LegacySurvey photoz's as annotations

mcoughlin:
This PR makes it possible to post LegacySurvey photoz's as annotations.

Blocked by new release post https://github.com/pycurl/pycurl/pull/739.


jadalilleboe (approved):
lgtm :)

guynir42:
Spinners seem to remain forever, I don't know why. 

I urge you to consider adding the cone search to sources HEAD, we have too many handlers as it is. 

I think there must be a simple way to return a source or nothing without popping up an error.

.github/workflows/build-and-deploy-docs.yaml
```diff
@@ -64,6 +64,12 @@ jobs:
                 password: anything
             EOF
 
+      - name: Install system dependencies
+        run: |
+          curl -sL https://deb.nodesource.com/setup_17.x | sudo -E bash -
+
+          sudo apt install -y libcurl4-gnutls-dev libgnutls28-dev
+
```
guynir42, line 72: Remind me again what this is about?

skyportal/tests/api/test_annotation_services.py
```diff
@@ -90,3 +90,42 @@ def test_vizier_quasar(public_group, upload_data_token):
             for d in data['data']
         ]
     )
+
+
+@pytest.mark.flaky(reruns=2)
```
guynir42, line 95: Do we know why this is flaky (if at all?)

skyportal/handlers/api/annotation_services.py
```diff
@@ -277,3 +280,125 @@ def post(self, obj_id):
...
+        content:
+          application/json:
+            schema: Error
+    """
+
+    @auth_or_token
+    def post(self, obj_id):
+        obj = Obj.query.get(obj_id)
+        if obj is None:
+            return self.error('Invalid object id.')
```
guynir42, line 340: shouldn't there be some check if the Obj is accessible to the user?

skyportal/handlers/api/annotation_services.py
```diff
@@ -277,3 +280,125 @@ def post(self, obj_id):
...
+        radius_deg = radius_arcsec / 3600.0
+
+        candidate_coord = SkyCoord(ra=obj.ra * u.deg, dec=obj.dec * u.deg)
+        sql_query = f"""SELECT {catalog}.photo_z.ls_id, z_phot_median, z_phot_std, ra, dec, type, z_phot_l95, flux_z from {catalog}.photo_z
+                      INNER JOIN {catalog}.tractor
+                      ON {catalog}.tractor.ls_id = {catalog}.photo_z.ls_id
+                      where ra > ({obj.ra-radius_deg}) and
+                            ra < ({obj.ra+radius_deg}) and
+                            dec > ({obj.dec-radius_deg}) and
+                            dec < ({obj.dec+radius_deg})"""
```
guynir42, line 368: you know this is not exactly a cone search, right?

guynir42: and also, around the poles this will never match anything because of ra*cos(dec)

skyportal/handlers/api/annotation_services.py
```diff
@@ -277,3 +280,125 @@ def post(self, obj_id):
...
+            annotations.append(annotation)
+
+        if len(annotations) == 0:
+            return self.error("No photo z's available.")
+
+        DBSession().add_all(annotations)
+        try:
+            self.verify_and_commit()
+        except IntegrityError:
+            return self.error("Annotation already posted.")
```
guynir42, line 398: Nice. I'm going to use this in a bunch of places.

guynir42 (commented):
Looks good. Seems like you had to wade through a bunch of installation nightmares to get it to work. 

My main comment is on the edge cases in the cross match query (around the poles and RA=0).

skyportal/tests/api/test_annotation_services.py
mcoughlin: I mark them as such as it is an external query, we could maybe remove but it seems better to give it a few tries.

skyportal/handlers/api/annotation_services.py
mcoughlin: good catch.

skyportal/handlers/api/annotation_services.py
mcoughlin: This is of course why we do things a bit different in SkyPortal ;) But this should still work in the vast majority of cases, and with some tuning of the radius right?

guynir42: yes but if you should try to get it running on the python side before sending the query. 
You could try: 
(1) replacing `radius_deg` with `radius_deg / cos(de)` in the RA clauses (check my calculations / units)
(2) divide the query for low/high RA and for low RA use mod(360) to make sure nothing is on the other side of zero.

mcoughlin: Looks like they had a new cone search function, so replaced with that.

skyportal/handlers/api/annotation_services.py
```diff
@@ -358,23 +358,14 @@ def post(self, obj_id):
...
 
-        candidate_coord = SkyCoord(ra=obj.ra * u.deg, dec=obj.dec * u.deg)
         sql_query = f"""SELECT {catalog}.photo_z.ls_id, z_phot_median, z_phot_std, ra, dec, type, z_phot_l95, flux_z from {catalog}.photo_z
                       INNER JOIN {catalog}.tractor
                       ON {catalog}.tractor.ls_id = {catalog}.photo_z.ls_id
-                      where ra > ({obj.ra-radius_deg}) and
-                            ra < ({obj.ra+radius_deg}) and
-                            dec > ({obj.dec-radius_deg}) and
-                            dec < ({obj.dec+radius_deg})"""
+                      where 't' = Q3C_RADIAL_QUERY(ra, dec, {obj.ra}, {obj.dec}, {radius_deg})"""
```
guynir42, line 364: see, that wasn't so hard, right?

guynir42 (approved):

skyportal/handlers/api/annotation_services.py
mcoughlin: To be fair, it was newish ;)

PR #2751, opened 2022-03-10
WISE colors

mcoughlin:
This PR adds WISE colors as an annotation option.


skyportal/handlers/api/annotation_services.py
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            origin = f"{catalog}-{row['ra']}-{row['dec']}"
+            if (
+                (annotation_data['w1mpro'] - annotation_data['w2mpro'] > 0.6)
+                and (annotation_data['w1mpro'] - annotation_data['w2mpro'] < 1.7)
+                and (annotation_data['w2mpro'] - annotation_data['w3mpro'] > 2.2)
+                and (annotation_data['w2mpro'] - annotation_data['w3mpro'] < 4.5)
+            ):
+                annotation_data['comment'] = 'AGN?'
```
guynir42, line 129: This is super specific. I wonder if this should live in some external service or something like that.

skyportal/tests/api/test_annotation_services.py
```diff
@@ -0,0 +1,39 @@
...
+
+    status, data = api(
+        'GET',
+        f'sources/{public_source.id}/annotations',
+        token=annotation_token,
+    )
+
+    assert status == 200
+
+    assert any(
```
guynir42, line 23: Why is this `any` and not `all`. Also are you sure the public source coordinates are never going to change?

guynir42 (commented):
My main concern is that this is very specific code. Maybe that's ok, if the service is very generic and will be useful to lots of groups doing different sorts of science. If we do leave this very specific request I suggest we just call the handler some thing with WISE and not IRSA.

mcoughlin:
@guynir42 Yeah my intuition here is that since these services house a ton of catalogs, we will be expanding them over time as we grab other catalogs from IRSA (similar with datalab in the other PR). But your comments will help to continue to make it general for sure.

skyportal/handlers/api/annotation_services.py
```diff
@@ -0,0 +1,153 @@
...
+from ...models import (
+    DBSession,
+    Annotation,
+    Group,
+    Obj,
+)
+
+
+class IRSAQueryHandler(BaseHandler):
+    """
```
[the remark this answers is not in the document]
mcoughlin: for some reason i feel extra called out by this one even more than usual.

skyportal/handlers/api/annotation_services.py
```diff
@@ -0,0 +1,153 @@
...
+            schema: Error
+    """
+
+    @auth_or_token
+    def post(self, obj_id):
+        obj = Obj.query.get(obj_id)
+        if obj is None:
+            return self.error('Invalid object id.')
+
+        data = self.get_json()
```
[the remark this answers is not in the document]
mcoughlin: oops.

skyportal/handlers/api/annotation_services.py
```diff
@@ -0,0 +1,153 @@
...
+        group_ids = data.pop('group_ids', None)
+        if not group_ids:
+            groups = self.current_user.accessible_groups
+        else:
+            try:
+                groups = Group.get_if_accessible_by(
+                    group_ids, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible groups.', status=403)
```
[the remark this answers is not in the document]
mcoughlin: I am modifying the response to see "at least some"

skyportal/handlers/api/annotation_services.py
mcoughlin: I am just going to remove this for now.

skyportal/handlers/api/annotation_services.py
```diff
@@ -0,0 +1,153 @@
...
+            'dec',
+            'w1mpro',
+            'w1sigmpro',
+            'w2mpro',
+            'w2sigmpro',
+            'w3mpro',
+            'w3sigmpro',
+            'w4mpro',
+            'w4sigmpro',
+        ]
```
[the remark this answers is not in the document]
mcoughlin: I am renaming it to include WISE. We can reconsider when we add more IRSA catalogs.

skyportal/handlers/api/annotation_services.py
```diff
@@ -120,14 +123,6 @@ def post(self, obj_id):
...
 
             origin = f"{catalog}-{row['ra']}-{row['dec']}"
-            if (
-                (annotation_data['w1mpro'] - annotation_data['w2mpro'] > 0.6)
-                and (annotation_data['w1mpro'] - annotation_data['w2mpro'] < 1.7)
-                and (annotation_data['w2mpro'] - annotation_data['w3mpro'] > 2.2)
-                and (annotation_data['w2mpro'] - annotation_data['w3mpro'] < 4.5)
-            ):
-                annotation_data['comment'] = 'AGN?'
-
```
guynir42: We should think about a way to add this (very useful) functionality in a way that is config-definable (in a separate PR, of course).

skyportal/tests/api/test_annotation_services.py
```diff
@@ -1,39 +1,53 @@
+import uuid
 import pytest
 from skyportal.tests import api
 
 
 @pytest.mark.flaky(reruns=2)
-def test_irsa_wise(annotation_token, public_source, public_group):
+def test_irsa_wise(annotation_token, public_source, public_group, upload_data_token):
```
guynir42, line 7: I think you may be able to call this without `public_source`. It may also be possible to use `upload_data_token` instead of `annotation_token` (but I'm not sure about that last one).

guynir42 (approved):
Minor change to the test, but otherwise looks ok to me.

skyportal/tests/api/test_annotation_services.py
mcoughlin: you were right.

skyportal/handlers/api/annotation_services.py
mcoughlin: maybe something that checks... all annotations from the external services and applies them if they apply? or it could be like config.annotations.agn flag default to false and turns this check on if true?

guynir42: I was thinking more of some generic way to flag objects based on annotation data. 
You'd have to define the name of the flag (e.g., AGN), the set of origins (if not specified, look at all of them) and the data keys you want to look at. 
This is somewhat complex because you'd have to have a whole language to parse which operators and logical combinations are accepted by these definitions. It may be simple enough if you allow only a few operations (e.g., "color" will always look for two keys and subtract them) and always require all the conditions to be true to pass, in which case you can have a relatively simple interface. If you want to do something even more generic, it is probably best to just pull the annotation data for each source, run custom code, and push back whatever comment/annotation you want (i.e., a service call).

mcoughlin: @guynir42 It kind of reminds me what we did for the period(?) stuff... a bit, but these are obviously more intense.

mcoughlin:
Thanks @guynir42. We will be discussing more on these flags.

PR #2752, opened 2022-03-10
Add pyupgrade to pre-commit checks

stefanv:


skyportal/models/instrument.py
```diff
@@ -26,7 +26,7 @@ def bind_expression(self, bindvalue):
         return cast(bindvalue, self)
 
     def result_processor(self, dialect, coltype):
-        super_rp = super(ArrayOfEnum, self).result_processor(dialect, coltype)
+        super_rp = super().result_processor(dialect, coltype)
```
mcoughlin, line 29: Is it surprising that this gives the same result as before?

mcoughlin (approved):
Crazy how many ways there are to do the same thing in python.

stefanv:
> Crazy how many ways there are to do the same thing in python.

```
import this
```

> ...
> There should be one-- and preferably only one --obvious way to do it.
> Although that way may not be obvious at first unless you're Dutch.

guynir42 (approved):
LGTM

PR #2753, opened 2022-03-10
Make it possible to delete allocations from front-end

mcoughlin:
This PR emulates the shift page so that Allocations can be removed.

### DIFF
--- a/static/js/components/AllocationPage.jsx
+++ b/static/js/components/AllocationPage.jsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { useSelector } from "react-redux";
+import { useSelector, useDispatch } from "react-redux";
 import List from "@material-ui/core/List";
 import ListItem from "@material-ui/core/ListItem";
 import ListItemText from "@material-ui/core/ListItemText";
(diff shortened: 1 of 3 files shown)


static/js/components/AllocationPage.jsx
```diff
@@ -21,6 +25,23 @@ const useStyles = makeStyles((theme) => ({
...
+  },
+  allocationDeleteDisabled: {
+    cursor: "pointer",
+    fontSize: "2em",
+    position: "absolute",
+    padding: 0,
+    right: 0,
+    top: 0,
+    opacity: 0,
+  },
```
guynir42, line 44: I'm not sure, but I think when the button is disabled, you should get the properties of both these classes, so you really only need the opacity to be different (please check if that works!).

guynir42 (approved):
Very minor comment.

static/js/components/AllocationPage.jsx
mcoughlin: I dont understand this comment :)

guynir42: I'm not sure: but the way the button is defined, I think it gets the `allocationDelete` class applied as root, and when it is disabled it applies `allocationDeleteDisabled` class on top of the root. If that's true then you don't need to have all the properties copied over from one class to the other, only the opacity which is different. I am not sure if this is how those definitions get inherited but it's worth a try.

mcoughlin: This was very correct. didn't need any of it!

mcoughlin:
Very minor (addressed) comments. Merging.

PR #2754, opened 2022-03-11
Enable ingestion of observations by RA/Dec

mcoughlin:
This PR enables ingestion of observations by RA/Dec (without needing to have field_id assigned yet).


skyportal/handlers/api/observation.py
```diff
@@ -412,11 +427,31 @@ def post(self):
                 'filter',
                 'exposure_time',
             ]
+        ) and not all(
```
guynir42, line 430: Can we replace these with `issubset`? Also the one above it while you're at it.

skyportal/handlers/api/observation.py
```diff
@@ -652,11 +687,31 @@ def post(self):
...
+            k in observation_data
+            for k in [
+                'observation_id',
+                'RA',
+                'Dec',
+                'obstime',
+                'filter',
+                'exposure_time',
+            ]
```
guynir42, line 699: issubset

guynir42 (approved):
Looks good. There are a few instances where we can make it more readable with `issubset` but besides that, it is good to go.

skyportal/handlers/api/observation.py
mcoughlin: yessir.

skyportal/handlers/api/observation.py
mcoughlin: yessir.

mcoughlin:
Thanks @guynir42!

PR #2755, opened 2022-03-11
FOV shape

mcoughlin:
This PR enables setting an instrument FOV by shape.


skyportal/handlers/api/instrument.py
```diff
@@ -333,6 +408,29 @@ def delete(self, instrument_id):
...
+                      type: dict
+                      items:
+                        type: array
+                      description: |
+                        List of ID, RA, and Dec for each field.
+                    field_region:
+                      type: str
+                      description: |
+                        Serialized version of a regions.Region describing
+                        the shape of the instrument field
```
guynir42, line 421: Maybe say something about how you should only input the region or the fov type+attributes?

guynir42 (approved):
Minor comment on the docstring, as always.

skyportal/handlers/api/instrument.py
mcoughlin: Yup added.

PR #2760, opened 2022-03-15
try to catch some errors

mcoughlin:
This PR tries to add some extra catches for the production SkyPortal instance.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -1345,7 +1345,13 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
     else:
         # tabs for mag, flux
         tabs = Tabs(tabs=[p2, p1], width=width, height=height + 90, sizing_mode='fixed')
-    return bokeh_embed.json_item(tabs)
+
+    try:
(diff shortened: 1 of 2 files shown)


skyportal/plot.py
```diff
@@ -1345,7 +1345,13 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
     else:
         # tabs for mag, flux
         tabs = Tabs(tabs=[p2, p1], width=width, height=height + 90, sizing_mode='fixed')
-    return bokeh_embed.json_item(tabs)
+
+    try:
+        return bokeh_embed.json_item(tabs)
+    except ValueError:
+        with pd.option_context('display.max_rows', None, 'display.max_columns', None):
+            print(data)
```
stefanv, line 1353: Maybe some context here so we know what we are seeing in the log?  Otherwise the PR looks good, thanks!

mcoughlin: added another print statement.

mcoughlin:
@stefanv feel free to merge if happy.

PR #2762, opened 2022-03-16
add geojson endpoint for observation plan

mcoughlin:
This PR adds observation plan plotting capabilities.

### DIFF
--- a/skyportal/handlers/api/observation_plan.py
+++ b/skyportal/handlers/api/observation_plan.py
@@ -860,10 +860,6 @@ class ObservationPlanGeoJSONHandler(BaseHandler):
         self.verify_and_commit()
 
         observation_plan = observation_plan_request.observation_plans[0]
-        num_observations = observation_plan.num_observations
-        if num_observations == 0:
-            return self.error('Need at least one observation to produce a GCN')
-
         # features are JSON representations that the d3 stuff understands.
         # We use these to render the contours of the sky localization and
         # locations of the transients.
@@ -877,4 +873,4 @@ class ObservationPlanGeoJSONHandler(BaseHandler):
             else:
                 continue
 
-        return self.success(data=geojson)
+        return self.success(data={'geojson': geojson})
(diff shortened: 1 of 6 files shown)


mcoughlin:
@guynir42 your suggestion won the day.

skyportal/handlers/api/observation_plan.py
```diff
@@ -820,3 +820,57 @@ def delete(self, observation_plan_request_id):
...
+            .joinedload(EventObservationPlan.planned_observations)
+            .joinedload(PlannedObservation.field)
+            .undefer(InstrumentField.contour_summary)
+        ]
+
+        observation_plan_request = ObservationPlanRequest.get_if_accessible_by(
+            observation_plan_request_id,
+            self.current_user,
+            mode="read",
+            raise_if_none=True,
```
guynir42, line 857: Convert to check if no results then `self.error`

static/js/components/ObservationPlanRequestLists.jsx
```diff
@@ -78,11 +81,89 @@ const getMuiTheme = (theme) =>
...
+
+  const [obsList, setObsList] = useState(null);
+  useEffect(() => {
+    const fetchObsList = async () => {
+      const response = await dispatch(
+        GET(
+          `/api/observation_plan/${observationplanRequest.id}/geojson`,
+          "skyportal/FETCH_OBSERVATION_PLAN_GEOJSON"
+        )
+      );
```
guynir42, line 108: Whats the advantage of using this instead of a fetch defined in the js file?

guynir42 (commented):
Some comments below:

static/js/components/ObservationPlanRequestLists.jsx
mcoughlin: That assigns a state variable no?

skyportal/handlers/api/observation_plan.py
mcoughlin: good point

static/js/components/ObservationPlanRequestLists.jsx
guynir42: Yes I think you're right.

guynir42:
@mcoughlin I have some doubts about how we've been dealing with `raise_if_none` recently. But that's maybe an issue to handle separately.

guynir42 (approved):

mcoughlin:
@guynir42 Yeah maybe at the same time with this permissions stuff.

PR #2764, opened 2022-03-17
Identify bad filter in sncosmo errors

guynir42:
This PR adds a try/except around code that calls `sncosmo`, and re-raises the exception only with some more info about the photometry point that caused the error. 

Hopefully this will help us understand what the underlying errors are.


guynir42:
Adresses #2668

skyportal/handlers/api/photometry.py
```diff
@@ -116,58 +116,67 @@ def serialize(phot, outsys, format):
...
-                'zp': corrected_db_zp,
-                'fluxerr': phot.fluxerr,
-            }
-        )
-    else:
+            raise ValueError(
+                'Invalid output format specified. Must be one of '
+                f"['flux', 'mag'], got '{format}'."
+            )
+    except ValueError as e:
```
stefanv, line 175: Won't you be catching the ValueError raised just above now as well?

guynir42: Yes. The goal is not to handle the error, but to figure out where it is coming from, and then fix the underlying issue or at least catch the exception with a more specific error message.

mcoughlin:
@guynir42 Any opposition to merging this? Seems like a good thing to get in.

guynir42:
@mcoughlin I think so. In principle I'm waiting for an OK from @stefanv but if you want to merge this, I think it would be fine.

PR #2772, opened 2022-03-22
add redshift_provenance column

mcoughlin:
This PR adds a "redshift_provenance" column to track source of redshifts.


guynir42 (approved):
This looks fine but I wonder if "redshift_origin" may be a better name for this?

mcoughlin:
Thanks @guynir42

PR #2773, opened 2022-03-23
Add ability to reject based on classifications

mcoughlin:
This PR introduces the ability to filter negatively based on classifications.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -493,6 +493,18 @@ class SourceHandler(BaseHandler):
               Comma-separated string of "taxonomy: classification" pair(s) to filter for sources matching
               that/those classification(s), i.e. "Sitewide Taxonomy: Type II, Sitewide Taxonomy: AGN"
           - in: query
+            name: nonclassifications
+            nullable: true
+            schema:
+              type: array
+              items:
+                type: string
+            explode: false
+            style: simple
+            description: |
+              Comma-separated string of "taxonomy: classification" pair(s) to filter for sources NOT matching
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/source.py
```diff
@@ -1221,6 +1234,56 @@ class Validator(Schema):
...
+                        )
+                    )
+                )
+            )
+            classification_accessible_query = (
+                Classification.query_records_accessible_by(self.current_user).subquery()
+            )
+
+            classification_query = (
+                DBSession()
```
guynir42, line 1264: I would call the first one `classification_accessible_subquery` and the second one `unclassifications_subquery` to make the logic a bit clearer.

skyportal/handlers/api/source.py
```diff
@@ -1221,6 +1234,56 @@ class Validator(Schema):
...
+                .query(
+                    distinct(Classification.obj_id).label("obj_id"),
+                    Classification.classification,
+                )
+                .join(Taxonomy)
+                .filter(Classification.classification.notin_(nonclassifications))
+                .filter(Taxonomy.name.in_(taxonomy_names))
```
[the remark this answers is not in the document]
mcoughlin: This is the purpose of this subquery:
            classification_accessible_subquery = (
                Classification.query_records_accessible_by(self.current_user).subquery()    
            )

skyportal/handlers/api/source.py
mcoughlin: yes good idea

mcoughlin:
Thanks @guynir42 .

PR #2777, opened 2022-03-24
Fix doc requirements

mcoughlin:
This PR restricts jinga2 to be below 3.1.0, which breaks our docs.


guynir42 (approved):

stefanv:
Better to fix the docs to work with latest Jinja.

mcoughlin:
@stefanv we aren't the only ones to see this: https://github.com/sphinx-doc/sphinx/issues/10291.

stefanv:
Ah, okay, it's a bug. Makes sense to pin then.

PR #2778, opened 2022-03-25
Add buttons for clearing/adding photometry

jadalilleboe:
https://user-images.githubusercontent.com/82007190/160194147-ea89c997-ff26-41a2-ae8f-5f89e4d3f454.mp4

@mcoughlin Is this what's desired? Adds a button for clearing photometry, add all, and add ZTF only.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -478,6 +478,50 @@ def add_plot_legend(plot, legend_items, width, legend_orientation, legend_loc):
         )
 
 
+def make_clear_photometry_button(model_dict):
+    button = Button(name="Clear Photometry", label="Clear Photometry", width=112)
+    callback_clear_photometry = CustomJS(
+        args={'model_dict': model_dict},
+        code="""
+        for (const [key, value] of Object.entries(model_dict)) {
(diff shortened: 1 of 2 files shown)


mcoughlin:
@jadalilleboe I like it. Maybe for ZTF, we can add a flag in the config to turn that on or off? What do you think @guynir42 ?

guynir42:
I would be most happy if the "ZTF only" was configurable in a way that lets you choose lists of filters, give them a common name, and have that turn up as buttons.

guynir42:
@jadalilleboe any updates on this?

jadalilleboe:
@guynir42 I looked into what you suggested but I think it would take me a while longer to figure it out. Maybe for now we can merge just the clear photometry and add all photometry buttons and I can work on a different PR for making it configurable

guynir42:
@jadalilleboe yeah that sounds fine.

guynir42 (approved):

mcoughlin:
Thanks @guynir42. Keeping track of the issue here: https://github.com/skyportal/skyportal/issues/2788

PR #2780, opened 2022-03-26
add gcnevent comments

mcoughlin:
This PR adds a comment block to the GcnEvent page.

### DIFF
--- a/skyportal/handlers/api/comment.py
+++ b/skyportal/handlers/api/comment.py
@@ -1,5 +1,6 @@
 import string
 import base64
+from distutils.util import strtobool
 from marshmallow.exceptions import ValidationError
 from baselayer.app.custom_exceptions import AccessError
 from baselayer.app.access import permissions, auth_or_token
@@ -8,7 +9,9 @@ from ...models import (
     DBSession,
     Comment,
     CommentOnSpectrum,
+    CommentOnGCN,
     Spectrum,
+    GcnEvent,
     Group,
     User,
     UserNotification,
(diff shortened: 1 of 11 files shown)


skyportal/handlers/api/comment.py
```diff
@@ -320,17 +355,23 @@ def post(self, associated_resource_type, resource_id):
...
-        if isinstance(comment, CommentOnSpectrum):
-            self.push_all(
-                action='skyportal/REFRESH_SOURCE_SPECTRA',
-                payload={'obj_internal_key': comment.obj.internal_key},
-            )
+            if isinstance(comment, CommentOnSpectrum):
+                self.push_all(
+                    action='skyportal/REFRESH_SOURCE_SPECTRA',
+                    payload={'obj_internal_key': comment.obj.internal_key},
+                )
```
guynir42, line 374: Why do we need this in a separate else clause? If it is an instance of `CommentOnGCN` then it would not have an `obj`, right? Maybe we should just change `if comment.obj.id` to `if hasattr(comment, 'obj')`. In this case the GCN comment `if` can come after the object and spectrum comments. (do this also for PUT and DELETE).

skyportal/models/comment.py
```diff
@@ -167,7 +167,7 @@ class CommentOnGCN(Base, CommentMixin):
     create = AccessibleIfRelatedRowsAreAccessible(gcn='read')
 
     read = accessible_by_groups_members & AccessibleIfRelatedRowsAreAccessible(
-        spectrum='read',
+        gcn='read',
```
guynir42, line 170: good catch!

guynir42 (approved):
Look great. Please make sure to also test all functionality on the mobile comments list.

skyportal/handlers/api/comment.py
mcoughlin: yeah you are right, fixing this.

skyportal/models/comment.py
mcoughlin: ;)

mcoughlin:
Thanks @guynir42.

PR #2785, opened 2022-03-29
Synthetic photometry

mcoughlin:
This PR starts thinking through synthetic photometry.

Known issues:
(i) Spectra need units...
(ii) We need to decide what to do about instruments not having filters... do we maybe not care?

    for packet in rows:
        if packet["filter"] not in instrument_cache[packet['instrument_id']].filters:
            instrument = instrument_cache[packet['instrument_id']]
            raise ValidationError(
                f"Instrument {instrument.name} has no filter " f"{packet['filter']}."
            )


guynir42:
about (i) I don't think we should care about spectra units because we are not going to generate the photometry from the spectrum. This is the job of the user to post photometry in the correct units in a given filter.
about (ii) I think we should just allow all filters on each spectrograph. I'm not sure what the problem would be?

mcoughlin:
@guynir42 but we need units to compute synthetic photometry from a spectrum no?

I am cool with that, so I guess we could bypass this check for spectrographs only?

guynir42:
I don't think we should be producing photometry from a spectrum at all.

mcoughlin:
I must have missed the entire point of the discussion.

guynir42:
What I'm thinking is that a user can produce their own synthetic photometry, but it has to be in one of the filters already present. Then they can upload it, using their spectrograph instrument (on SP) and the filters they chose (also, on SP). 
We'd not be doing any calculation, and we won't be attaching the photometry points to a spectrum (only to the instrument, in the usual way). 
It really isn't a big change to our code, just allowing the spectrographs to be associated with photometry, and allowing each spectrograph to have any of our list of filters.

mcoughlin:
@guynir42 Ok I am with you on the other changes, but this also seems like a nice interface to enable it no?

guynir42:
Spoken like someone whose already implemented it all. 
I guess it doesn't hurt to allow this option, but it does put us in the habit of figuring out the units, which is not trivial. 
I would suggest putting this up for debate with the team.

mcoughlin:
Taking something reasonably normal, i.e. erg/s/cm2/A, we could check it with something like:

>>> import astropy.units as u
>>> units = (u.erg/u.s/u.cm**2/u.AA)
>>> u.equivalencies.spectral_density(1 * units)

We could say that they can enter a string in the form erg/s/cm/cm/AA, and we will break apart the series and unitfy each entry as u.XX and final check u.equivalencies.spectral_density(1 * units) passes?

guynir42:
> Taking something reasonably normal, i.e. erg/s/cm2/A, we could check it with something like:
> 
> > > > import astropy.units as u
> > > > units = (u.erg/u.s/u.cm**2/u.AA)
> > > > u.equivalencies.spectral_density(1 * units)
> 
> We could say that they can enter a string in the form erg/s/cm/cm/AA, and we will break apart the series and unitfy each entry as u.XX and final check u.equivalencies.spectral_density(1 * units) passes?

It seems a little rabbit-holey to me.

guynir42:
> We should not be parsing a free-form user string into a astropy units. What about giving 1 or 2 options (like units=microJy or ABmag) and then we internally convert as appropriate? Fundamentally we need to have sharp elbows with what sort of data goes into the SP database--asking the end user to do some conversions before import isn't too onerous.

After rescaling the flux they can also integrate the synthetic photometry and upload it as photometry instead of bothering us with the units. Just my 2 cents.

mcoughlin:
Makes sense, I updated the PR to restrict the list.

skyportal/handlers/api/spectrum.py
```diff
@@ -790,3 +797,90 @@ def get(self):
...
+            spectrum_id,
+            self.current_user,
+            raise_if_none=True,
+        )
+
+        spec_dict = recursive_to_dict(spectrum)
+        wav = spec_dict['wavelengths']
+        flux = spec_dict['fluxes']
+        err = spec_dict['errors']
+        print(spectrum.astropy_units)
```
guynir42, line 847:
```suggestion
```

skyportal/handlers/api/spectrum.py
```diff
@@ -790,3 +797,90 @@ def get(self):
...
+        """
+
+        data = self.get_json()
+        filters = data.get('filters')
+
+        spectrum = Spectrum.get_if_accessible_by(
+            spectrum_id,
+            self.current_user,
+            raise_if_none=True,
+        )
```
guynir42, line 841: Catch this if no spectrum.

skyportal/handlers/api/spectrum.py
```diff
@@ -790,3 +797,90 @@ def get(self):
...
+        except TypeError:
+            spec = sncosmo.Spectrum(wav, flux * spectrum.astropy_units)
+
+        data_out = []
+        for filt in filters:
+            try:
+                mag = spec.bandmag(filt, magsys='ab')
+                magerr = 0
+            except ValueError:
+                continue
```
guynir42, line 863: This seems kind of reckless...

skyportal/handlers/api/spectrum.py
```diff
@@ -790,3 +797,90 @@ def get(self):
...
+                    'dec': spectrum.obj.dec,
+                    'mag': mag,
+                    'magerr': magerr,
+                    'filter': filt,
+                    'limiting_mag': 25.0,
+                }
+            )
+        df = pd.DataFrame.from_dict(data_out)
+        df['magsys'] = 'ab'
+        data_out = {
```
guynir42, line 878: I'd rename one of the `data_out` variables.

skyportal/models/schema.py
```diff
@@ -1373,6 +1373,12 @@ class SpectrumPost(_Schema):
         },
     )
 
+    units = fields.String(
+        metadata={
+            'description': "Units of the fluxes/errors in astropy notation (e.g. erg/s/cm/cm/AA).",
```
guynir42, line 1378: I'd replace the e.g., with all three options.

skyportal/models/spectrum.py
```diff
@@ -60,6 +61,12 @@ class Spectrum(Base):
         doc="Errors on the fluxes of the spectrum [F_lambda, same units as `fluxes`.]",
     )
 
+    units = sa.Column(
+        sa.String,
+        nullable=True,
+        doc="Units of the fluxes/errors in astropy notation (e.g. erg/s/cm/cm/AA).",
```
guynir42, line 67: Same here, give all three options.

guynir42 (commented):
I have minor suggestions. 
This looks very good. 
I'm definitely warming up to the idea of having this functionality from the front end. 

I do have a couple of questions:
1) What happens if I try to make synth-phot from a spectrum that doesn't have units? 
2) What happens if I try to make synth-phot twice on the same spectrum? (does de-duplication work). 
3) Can I delete these photometry points if I decide I don't like them? (from the front end)
Please make sure all these work on the frontend.

guynir42:
...and also add a test.

skyportal/handlers/api/spectrum.py
mcoughlin: I am adding a log message. What else do you think?

skyportal/handlers/api/spectrum.py
mcoughlin: yup.

skyportal/handlers/api/spectrum.py
mcoughlin: yup.

skyportal/models/schema.py
mcoughlin: yup

skyportal/models/spectrum.py
mcoughlin: yup.

skyportal/handlers/api/spectrum.py
guynir42: I'm not sure. In what cases would this happen? Failure to post one of the photometry points doesn't necessarily mean the whole operation should be stopped... but I also don't want the user to feel like it all worked smoothly and never notice some of the filters were not posted. 
It may be the right thing to return a self.error if any of the filters fail, unless there's a general reason for this to happen often. If there is such a reason, we should be very clear to report that reason to the user.

mcoughlin: Good point, will raise an error.

guynir42:
Reran the tests. Seeing the same failures. Seems to be related...

mcoughlin:
@guynir42 yeah hard to understand exactly what is wrong here.

guynir42:
Yes, but it is consistent, so worth investigating...

mcoughlin:
Indeed, I played around with it a bit and I think might be related to the posted photometry not actually being associated with a real user, definitely a bit strange.

mcoughlin:
Failures now unrelated.

skyportal/tests/api/test_synthphot.py
```diff
@@ -0,0 +1,68 @@
+import datetime
+
+import uuid
+from skyportal.tests import api
+
+
+def test_synthetic_photometry(super_admin_token, public_source, public_group):
+    name = str(uuid.uuid4())
```
guynir42, line 8: name -> telescope name

skyportal/tests/api/test_synthphot.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    assert status == 200
+    assert data['status'] == 'success'
+    instrument_id = data['data']['id']
+
+    status, data = api(
+        'POST',
+        'spectrum',
+        data={
+            'obj_id': str(public_source.id),
```
guynir42, line 45: I think the str conversion here is not needed.

skyportal/tests/api/test_synthphot.py
```diff
@@ -0,0 +1,68 @@
...
+    status, data = api(
+        'POST',
+        f'spectra/synthphot/{spectrum_id}',
+        data={
+            'filters': ['ztfg', 'ztfr', 'ztfi'],
+        },
+        token=super_admin_token,
+    )
+    assert status == 200
+    assert data['status'] == 'success'
```
guynir42, line 68: Is there a way to check that the photometry was actually posted, other than looking at the Bokeh plot (which is tricky)? 

Also, are there obvious failure cases that you can check and verify it returns the error you expect?

guynir42 (commented):
I would make an effort to beef up the test a little bit.

skyportal/tests/api/test_synthphot.py
mcoughlin: yup.

skyportal/tests/api/test_synthphot.py
mcoughlin: yup.

skyportal/tests/api/test_synthphot.py
mcoughlin: yeah adding both a success and failure check here.

guynir42 (approved):

PR #2790, opened 2022-03-30
Add ref_flux and ref_fluxerr columns to photometry

guynir42:
Add the baseline, or reference magnitude (and uncertainty) to the Photometry data model. 

These columns can be used for storing one of these options:
1) the magnitude of the source before/after the transient. 
2) the average magnitude of a variable source. 

Other use cases may emerge later. 

The integration with the plotting tools will be added in a separate PR.


guynir42:
@profjsb This will probably need an updated migration hash before merging. I'll do that after you review.

profjsb (commented):
Looks fine but I'm not sure why this has to be in the photometry table and not at the obj level? Since this is the average mag it shouldn't be different for every new photometry point (it will be different by filter). Adding columns to the photometry table will incur a large hit in DB size since this is our largest table by far. We have in `obj.py` already:

```python
    dist_nearest_source = sa.Column(
        sa.Float, nullable=True, doc="Distance to the nearest Obj [arcsec]."
    )
    mag_nearest_source = sa.Column(
        sa.Float, nullable=True, doc="Magnitude of the nearest Obj [AB]."
    )
    e_mag_nearest_source = sa.Column(
        sa.Float, nullable=True, doc="Error on magnitude of the nearest Obj [mag]."
    )
```
these columns could be extended to lists (including filter info) or perhaps more easily add a `ref` JSONB which looks like: `{'rband': 23.4, 'rband_unc': 0.23, ....}`.

guynir42:
@profjsb two problems with what you suggest:
1) the reference mag must be in a filter from the list, otherwise it may not match any of the photometry points. The solution would be to add `ref_photometry` to each `Obj` and have those objects keep track of filter and so on. 
2) Different photometry for the same `Obj` may have different reference magnitudes (based on different reference images, or based on the mean flux of the star taken until now, etc). Again we can associate each (subtraction) photometry point with a reference photometry point, but that is probably not cheaper than just saving the reference mag. 

So here are a few solutions:
- each photometry would get the reference photometry ID (or list of IDs in case we made the subtraction based on the average magnitude). This seems the most complicated way, but also the most generic. 
- save a reference photometry with one point per object per filter. This is really what most people want to know, i.e., what was the source doing before the transient or what is the average magnitude of this variable. The other photometry points will know how to find the right reference by looking up the `Obj` and filter. Upside: no need to add columns (maybe just a BOOLEAN saying if this point is a subtraction). The downside is that if the user that posted the photometry used different reference images, the reconstructed unsubtracted lightcurve will not be 100% accurate. 
- Just keep the reference mag with each photometry, and leave it empty if no reference is given. 
I think the main consideration is how much does it cost to have a nullable column. If it doesn't add much to the DB then we may consider adding all options and letting the sys admin decide on the strategy.

guynir42:
I've settled on the easiest to implement solution, which I think would also be the most commonly used. 
The reference flux (and error) are saved to each photometry point. That way you can have different reference images assigned to different photometry (when making a subtraction, you need to keep track of which image you used, specifically what was the source magnitude in that image). 

If you want to keep a ref flux on the object itself, then there are a few problems:
1) You need to join the Obj table for every query on the reference/total flux. 
2) You need to keep a JSONB on the Obj that contains the reference flux for each filter/instrument. 
3) If you keep changing the reference images (e.g., LSST plans to make new references each year) you can quickly lose track of that info. 

Finally, I feel like most instances of skyportal will either not use refmag, or use just one method (individual or obj level ref) so it may be easier to add something in the config that lets you define the model based on which method you prefer. I highly doubt anyone would ever go that far just to save a couple of floats off the photometry table...

skyportal/models/photometry.py
```diff
@@ -201,6 +214,145 @@ def e_mag(cls):
...
+        """The magnitude of the photometry point in the AB system."""
+        return sa.case(
+            [
+                (
+                    sa.and_(
+                        cls.ref_flux != None,  # noqa: E711
+                        cls.ref_flux != 'NaN',
+                        cls.ref_flux > 0,
+                    ),
+                    -2.5 * sa.func.log(cls.ref_flux) + PHOT_ZP,
```
profjsb, line 253: is this log10 or log2?

skyportal/models/photometry.py
```diff
@@ -201,6 +214,145 @@ def e_mag(cls):
...
+        return sa.case(
+            [
+                (
+                    sa.and_(
+                        cls.ref_flux != None,  # noqa: E711
+                        cls.ref_flux != 'NaN',
+                        cls.ref_flux > 0,
+                        cls.ref_fluxerr > 0,
+                    ),  # noqa: E711
+                    2.5 / sa.func.ln(10) * cls.ref_fluxerr / cls.ref_flux,
```
profjsb, line 271: Can you put parentheses in this equation to make sure the order of operations is what you expect? I'm not a far of double division without parentheses (even if not needed, useful for readability)

alembic/versions/581774928d27_photometry_magref.py
```diff
@@ -0,0 +1,34 @@
...
+revision = '581774928d27'
+down_revision = '1a43ee08a734'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('photometry', sa.Column('magref', sa.Float(), nullable=True))
+    op.add_column('photometry', sa.Column('magref_unc', sa.Float(), nullable=True))
```
profjsb, line 22: I would have thought you'd need to add ref_flux and ref_fluexerr to the data model, not mag. Mag appears to be a derived property. What am I missing?

skyportal/models/photometry.py
guynir42: According to this it is indeed log10, but I'm glad we made the check: https://www.w3resource.com/PostgreSQL/log-function.php#:~:text=The%20PostgreSQL%20log()%20function,as%20specified%20in%20the%20argument.

skyportal/models/photometry.py
guynir42: will do

alembic/versions/581774928d27_photometry_magref.py
guynir42: I think this is the old migration script. I thought I had removed this but for some reason it is still on the repo. I'll remove it now.

guynir42:
@profjsb Let me know if you want another look before I merge. 
Please do not merge yet, I want to verify the migration hash has not changed since I added the alembic script.

profjsb (approved):
Needs a test but otherwise fine to merge.

guynir42:
@profjsb Tests are going to be added in the next PR, along with some additions to the photometry handler.

profjsb:
Ok. Go ahead and merge when ready.

PR #2791, opened 2022-03-30
Make instrument tiles clickable and add tooltips

jadalilleboe:
@mcoughlin Here's the interaction:
ezgif com-gif-maker (5)
could be annoying to click a bunch of those tiles though...
the button handler just logs the selected instrument field ids as of now


mcoughlin:
@jadalilleboe Looks amazing. Does it also work for the observation plan tiles after making an observation plan at the bottom of the page?

jadalilleboe:
No, the selected state is in GcnSelectionForm right now. I think the state would have to be lifted up to be sent along on submit of the observation plan form

mcoughlin:
But can replicate that same button in the observationplanform or whatever we call it right?

jadalilleboe:
@mcoughlin yes most likely. So the observation plan tiles are different than the instrument tiles? How do I get those to show up? I think I could replicate the same logic for those

jadalilleboe:
ohhh ok. So how do I get the tiles to show up on these ones? they just have the black lines for me
Screen Shot 2022-03-31 at 10 58 53 AM

mcoughlin:
@jadalilleboe super cool. Let's get this merged like this and will deal with using them next.

mcoughlin:
@jadalilleboe when you have a second, could you return the demo file to normal and remove console statements and such? Would like to get this reviewed and in to show off in the demo on Tuesday.

static/js/components/GcnSelectionForm.jsx
```diff
@@ -190,6 +190,11 @@ const GcnSelectionForm = ({ gcnEvent }) => {
     setIsSubmitting(false);
   };
 
+  // const handleSendToObservationPlan = (instrument) => {
```
mcoughlin, line 193: Can you bring this back?

static/js/components/GcnSelectionForm.jsx
```diff
@@ -304,6 +312,14 @@ const GcnSelectionForm = ({ gcnEvent }) => {
         />
       </div>
       <div>
+        <Button
+          variant="outlined"
+          // onClick={() =>
```
mcoughlin, line 317: here too.

static/js/components/ObservationPlanRequestLists.jsx
```diff
@@ -109,6 +109,11 @@ const ObservationPlanGlobe = ({ observationplanRequest, loc }) => {
       setObsList(response.data);
     };
     fetchObsList();
+    if (obsList) {
```
mcoughlin, line 112: This just sets the defaults I guess?

mcoughlin (approved):
Looking overall very good. Looks like a few things commented out that shouldn't be, otherwise looks good to go.

static/js/components/GcnSelectionForm.jsx
jadalilleboe: Can it just stay commented until there's some sort of request being made? the handler doesn't do anything right now.

mcoughlin: Ah of course!

static/js/components/ObservationPlanRequestLists.jsx
jadalilleboe: yes, default for all data points to be unselected.

PR #2792, opened 2022-03-31
Github vs. Black

guynir42:
Try to clear the github cache by changing the cache key
(ref: https://stackoverflow.com/a/64819132/18256949)
--> This didn't work
Tried a few other things but then editing the precommit hooks seems to have done the trick 
(more info about this issue here: https://github.com/psf/black/issues/2964)


guynir42:
Indeed the tests for model logic and api_and_utils fail even though this branch is identical to master.

mcoughlin (approved):
Thanks for figuring this out @guynir42.

Issue #2748, opened 2022-03-09
Spectrum HEAD request


guynir42:
Add the ability to query only the spectrum's metadata without pulling down the whole wavelength-flux columns. 
Should ideally also be able to query the HEAD for all spectra filtered by criteria on their metadata. 
This could be really useful for anyone doing spectrum intensive work.

PR #2793, opened 2022-03-31
Filtering spectrum GET requests

guynir42:
Allow users to send queries filtering spectra by various properties, e.g., the object ID (partial match), the origin, label or type, or by the time the spectrum was observed. 

This is useful for groups doing lots of spectroscopy work (like DESI) where finding a list of spectra is important. 
This also obviates the `ObjSpectrumHandler` and `SpectrumRangeHandler`. We can remove them in the future. 

Also allows users to query if a spectrum exists, and get a bit of metadata about it, without pulling the entire wavelength/flux dataset.


guynir42:
addresses #2748

skyportal/handlers/api/source.py
```diff
@@ -1383,7 +1383,7 @@ class Validator(Schema):
                 comments_filter_author = [comments_filter_author]
             else:
                 return self.error(
-                    "Invalid commentsFilterOrigin value -- must provide at least one string value"
+                    "Invalid commentsFilterAuthor value -- must provide at least one string value"
```
mcoughlin, line 1386: nice

skyportal/handlers/api/spectrum.py
```diff
@@ -165,77 +226,502 @@ def post(self):
...
+              content:
+                application/json:
+                  schema: Error
+        multiple:
+          description: Retrieve multiple spectra with given criteria
+          tags:
+            - spectra
+          parameters:
+            - in: query
+              name: minimal
```
mcoughlin, line 257: maybe minimalMetadata or similar?

skyportal/tests/api/test_spectrum.py
```diff
@@ -9,6 +11,1020 @@
...
+        assert k in minimal_spec  # using multiple spectra, minimal output
+        assert k in full_spec  # using multiple spectra, full output
+        assert k in single_spec  # using single spectra (should be full always)
+        assert minimal_spec[k] == full_spec[k]
+
+    # check that keys of full spec, outside the minimal list, are not included in minimal
+    for k in full_spec.keys():
+        assert k in list_of_keys or k not in minimal_spec
+
+    # make sure full and single are the same
```
mcoughlin, line 1022: really nice test.

mcoughlin (approved):
I think minimal should be a bit more verbose name, otherwise really nice.

PR #2818, opened 2022-04-01
Update Priority form

mcoughlin:
This PR adds functionality to update follow-up priority based on sky-map crossmatch.


skyportal/handlers/api/followup_request.py
```diff
@@ -1027,3 +1033,182 @@ async def get(self, instrument_id):
...
+          description: Filter by instrument ID
+        - in: query
+          name: localizationId
+          schema:
+            type: integer
+          description: Filter by localization ID
+        - in: query
+          name: requestIds
+          schema:
+            type: List[integer]
```
guynir42, line 1076:
```suggestion
          schema:
            type: list
              items:
                type: integer
```

skyportal/handlers/api/followup_request.py
```diff
@@ -1027,3 +1033,182 @@ async def get(self, instrument_id):
...
+          schema:
+            type: string
+          description: Filter by instrument ID
+        - in: query
+          name: localizationId
+          schema:
+            type: integer
+          description: Filter by localization ID
+        - in: query
+          name: requestIds
```
guynir42, line 1074: Usually when you use PUT, it comes with a single path argument that identifies the thing you're updating. Here we specify localization and instrument IDs (one of each, I guess) and then a bunch of followup IDs. Maybe we should instead do this for each followup request in a separate call? 
Or if that is not efficient, and you still want to give a list of followup IDs, do you actually need the instrument and localization IDs, or can you grab them from the followup requests themselves?

skyportal/handlers/api/followup_request.py
```diff
@@ -1027,3 +1033,182 @@ async def get(self, instrument_id):
...
 
         await self.send_file(data, filename, output_type=output_format)
+
+
+class FollowupRequestPrioritizationHandler(BaseHandler):
+    @auth_or_token
+    async def put(self):
+        """
+        ---
+        description: Reprioritize followup requests schedule
```
guynir42, line 1043: Maybe you should say that it automatically sets new priorities, and not that the user gets to assign their own priorities (which is what I thought until I read the code).

skyportal/tests/api/test_followup_requests_reprioritize.py
```diff
@@ -0,0 +1,63 @@
...
+        data=new_request_data,
+        token=upload_data_token,
+    )
+    assert status == 200
+    assert data['status'] == 'success'
+
+    status, data = api('GET', f'followup_request/{id}', token=upload_data_token)
+    assert status == 200
+
+    assert data['data']["payload"]["priority"] == 5.0
```
guynir42, line 63: So the priorities that get assigned really depend on which followup IDs are given? If I specify only one ID that always gets priority 5, regardless of the other requests that aren't specified?

guynir42 (commented):
I am wondering if this is the best way to design the PUT query (see below). Other than that I have only minor comments. 
I'm assuming you checked the front end forms thoroughly...

skyportal/tests/api/test_followup_requests_reprioritize.py
mcoughlin: Yeah my thinking here is that you have done some filtering and limited to the subset you care about (based on the filtering in that page). Then you want to order them by some kind of criteria, in this case the skymaps. The priority range can be changed from 1-5 in the option list, but yeah, if its just 1, it becomes top priority.

skyportal/handlers/api/followup_request.py
mcoughlin: yes adding.

skyportal/handlers/api/followup_request.py
```diff
@@ -1027,3 +1033,182 @@ async def get(self, instrument_id):
...
+        tags:
+            - followup_requests
+        parameters:
+        - in: query
+          name: observationStartDate
+          nullable: true
+          schema:
```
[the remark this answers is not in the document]
mcoughlin: you are right.

skyportal/handlers/api/followup_request.py
mcoughlin: Yeah maybe we have a chat about this.

guynir42 (approved):
Looks good. Did we settle on PUT with everything in the body and no path arguments?

mcoughlin:
Thanks @guynir42. Yeah, we can always change it down the line if we prefer something else of course.

PR #2823, opened 2022-04-04
Updated shift frontend - Shift Calendar

Theodlz:
This PR aims to modify the front-end of the Shift Page to add a Calendar element.

For the calendar, we used the [react-big-calendar](https://github.com/jquense/react-big-calendar) npm package.
Thanks to this PR, you'll find a calendar instead of a lists of shifts. You can still use the form to create a shift but you can also:
- create multiple shifts using the form, that repeat everyday on the selected period.
- create a shift directly on the calendar by clicking and holding the mouse button down over a time period of the calendar, after which you can select a name, description and group. 

#### Creating Shifts with the Form:
Form_Shift

#### Creating Shifts directly on the Calendar:
Calendar_Shift

#### Creating repeating Shifts using the Form:
Form_repeated_shift

#### Joining, Leaving, and deleting Shifts:
Participate_and_delete

#### Using the different Calendar Views:
Calendar_Views


static/js/components/NewShift.jsx
```diff
@@ -27,12 +27,37 @@ const NewShift = () => {
...
-      dispatch(fetchShifts());
+    if (!formData.multipleDays) {
+      formData.start_date = formData.start_date.replace("+00:00", "");
+      formData.end_date = formData.end_date.replace("+00:00", "");
+      delete formData.multipleDays;
+      const result = await dispatch(submitShift(formData));
+      if (result.status === "success") {
+        dispatch(showNotification("Shift saved"));
+        dispatch(fetchShifts());
+      }
```
guynir42, line 38: Shouldn't there be an else with an error message? or is that sent up by the `self.error`?

static/js/components/NewShift.jsx
```diff
@@ -82,6 +107,11 @@ const NewShift = () => {
         type: "string",
         title: "Shift's description",
       },
+      multipleDays: {
+        type: "boolean",
+        title: "Shift repeats everyday",
+        default: false,
+      },
```
guynir42, line 114: This was a little confusing for me to figure out what everyday means exactly. 
You should add in the description (or tooltip) that the shift will repeat during the days specified and that in each day separately the shift starts and ends based on the hh:mm:ss given.

static/js/components/ShiftCalendar.jsx
```diff
@@ -0,0 +1,115 @@
...
+      name,
+      description,
+      start_date,
+      end_date,
+      group_id,
+    })
+  ).then((result) => {
+    if (result.status === "success") {
+      showNotification("Shift saved");
+    }
```
guynir42, line 38: Why do we have this notification in two different components?

static/js/components/ShiftCalendar.jsx
```diff
@@ -0,0 +1,115 @@
...
+      </span>
+    </div>
+  );
+}
+
+function MyCalendar({ shifts }) {
+  currentUser = useSelector((state) => state.profile);
+  dispatch = useDispatch();
+  groups = useSelector((state) => state.groups.userAccessible);
+  console.log(shifts);
```
guynir42, line 61:
```suggestion
```

guynir42 (commented):
This looks really good. Very nice interface. 
My two main comments are
(1) the "everyday" needs a bit of explanation (maybe in a tooltip)
(2) the test should be put into the regular (non-flaky) workflow, unless you've given it a thorough try and couldn't get it to consistently pass. We should not be putting new tests directly into the flaky workflow.

Theodlz:
Hi @guynir42 , thanks for your review. I'll reply to your comments and make the changes you asked for tomorrow. Have a nice day !

static/js/components/ShiftCalendar.jsx
Theodlz: In the NewShift component, this notification triggers when you submit the shift you entered in the Form. In this component, it triggers when you create the Shift directly in the calendar, not with the form.

static/js/components/NewShift.jsx
Theodlz: So, as we are using liveValidate in the form, no error should occur as the user cant even submit the shift if the input is wrong. For the feature where you create a shift directly in the calendar, without having to specify an else, it already displays an error message (lets say you forgot the name, it will print (missing required parameter 'name')

Theodlz:
> This looks really good. Very nice interface. My two main comments are (1) the "everyday" needs a bit of explanation (maybe in a tooltip) (2) the test should be put into the regular (non-flaky) workflow, unless you've given it a thorough try and couldn't get it to consistently pass. We should not be putting new tests directly into the flaky workflow.

Hey @guynir42 , i took care of both of those aspects, but for (1) its hard for me to know for sure if it makes enough sense to the user as i spent so much time on it myself. Maybe you can give me your opinion during your next review ? Thanks.

jadalilleboe:
It would be nice if the start and end time and date were displayed in the upper right hand corner box when you click on a shift. Also I'm seeing this Members: null null thing which should probably not be displayed if there are no members
Screen Shot 2022-04-07 at 12 17 39 PM
Also a way to delete multiple shifts at once would be nice

Theodlz:
> It would be nice if the start and end time and date were displayed in the upper right hand corner box when you click on a shift. Also I'm seeing this Members: null null thing which should probably not be displayed if there are no members
> Screen Shot 2022-04-07 at 12 17 39 PM
> Also a way to delete multiple shifts at once would be nice
>  

Hi @jadalilleboe , you're right for the display of the start and end time of a shift, i'll add it. When it comes to the 'null null' thing, its not because there is no user, its actually because there is one ! But the firstname and lastname of that user in Skyportal is null null. I think that should be changed by the way, it's a bit confusing.

Theodlz:
Hi @guynir42 , all tests passed. I think this is ready for a review if you want to have a look at it.

guynir42 (commented):
The code looks fine. I do have a couple of UX comments:
(1) When clicking on a shift it now shows the start/end time (great!). For daily shifts I would prefer to see the start and end date (no time) and then in a line under that something like "each shift 9:00am to 17:00pm" (or something of that sort). 
(2) When adding shifts by clicking on the calendar, the popup can be improved a bit: 
   - the group choice popup has the commas on the wrong side of the group names
   - if no groups are chosen it returns an error (is there a way to set a default value for this?)
   - when clicking "cancel" the popup continues adding the shift, I would expect the whole process to abort if I click that.

Theodlz:
> The code looks fine. I do have a couple of UX comments: (1) When clicking on a shift it now shows the start/end time (great!). For daily shifts I would prefer to see the start and end date (no time) and then in a line under that something like "each shift 9:00am to 17:00pm" (or something of that sort). (2) When adding shifts by clicking on the calendar, the popup can be improved a bit:
> 
> * the group choice popup has the commas on the wrong side of the group names
> * if no groups are chosen it returns an error (is there a way to set a default value for this?)
> * when clicking "cancel" the popup continues adding the shift, I would expect the whole process to abort if I click that.

Done !

static/js/components/ShiftManagement.jsx
```diff
@@ -30,6 +30,32 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
+function isDailyShift(shiftName) {
+  const regex = /\d+\/\d+/;
+  return regex.test(shiftName)
```
guynir42, line 35: Can we have some data on the shift object (not necessarily in the DB) that tells if it is a daily or not? If you rely on the name alone, you will see some trouble when some user decides to call their shift "not a daily shift 1/1" or something like that.

guynir42 (commented):
Besides the comment below about naming of daily shifts, I also have a couple of UX questions:
1) I seem to not be able to select multiple days. I'm not sure if I could do that before. 
2) Is there some use of the color of the shifts on the calendar? for different users or groups or something like that? It seems like a good display option to use that to tell apart different shifts.

Theodlz:
Hi @guynir42 

(1) What do you mean when you say that you are not able to select multiple shifts ? Do you mean selecting multiple shifts on the calendar to create them (that is not an option we worked on), or at another time ?

(2) Right now, the color indicates you if you are a member of a shift or not. If you are not participating it is in a light blue color, but if you are then it shows up as a bright green.

static/js/components/ShiftManagement.jsx
Theodlz: If we dont store it in the DB, i think the name is the only option. What i can add (i should, thought about it already actually), is checking that the name the user inputs doesnt contain anything that matches that regex condition. If it does, the form tells you its an incorrect name. What do you think ? I think that adding another column in the DB to indicate that a shift repeats itself is too much, especially because the only thing we use it for later on is to change the display slightly.

guynir42: When you save a daily shift to the DB, does it get saved as multiple, different shifts, or as a single shift?

guynir42:
> Hi @guynir42
> 
> (1) What do you mean when you say that you are not able to select multiple shifts ? Do you mean selecting multiple shifts on the calendar to create them (that is not an option we worked on), or at another time ?
> 
> (2) Right now, the color indicates you if you are a member of a shift or not. If you are not participating it is in a light blue color, but if you are then it shows up as a bright green.

(1) I meant that I try to highlight a shift across multiple days, and I can only choose one day at a time. 
(2) good!

Theodlz:
(1) Yes @guynir42, this is the expected behaviour we had in mind. On the form you can either create one shift or repeating shift. And on the calendar you can only create a single shift. The idea is that the form helps you to create repeating shifts mostly, whereas the calendar can be used to create a shift directly on it. 

Also, when selecting existing shifts on the calendar, it is expected that you can select them one by one only. To manage each shift individually.

static/js/components/ShiftManagement.jsx
Theodlz: It creates N shifts in the DB, for the N days you selected. The issue we could have had with creating just one big shift on the whole period + adding a column or some way to say that it is a repeating shift, is that then we could not have different members on each of the daily shifts. We had to share the members for all of the instances/days of a daily shift.

Theodlz: I added the regex check !

guynir42:
OK I think if you are saving the shifts individually and you only generate them as one large daily shift then it is ok to use the name. 

What I had trouble with is that when I tried to add a new shift using the calendar interface, and I try to highlight more than one day, I can't do it. I can only create a shift on a single day (from the week view). I find I am able to mark whole days in the month view or if I mark the top part of the week days, but then I can't choose hours. This may be a limitation of the component, but I think I could do that in previous commits (maybe I am miss remembering).

Theodlz:
I'll have a look. I think it has always been the case for some of the different scenarios you mentionned. At first we did  not even plan to be able to create shifts directly on the calendar (we only had the form as a requirement), but allowing to create it on the calendar even only in certain cases felt like an improvement. 

But, it did work to be able to create a shift that for example spans on 2 days on the calendar before, you are right. But since we first open the PR the package was updated (before we had to run npm i --force even before the having the issues we are all facing with skyportal this days where we have to run that anyway), and it seems that that was removed (at least "out of the box). Maybe making it work again somehow could be the subject of another PR ? As anyway creating shifts directly on the calendar is more of an additional feature even on this one ?

guynir42:
@Theodlz I was just making sure that this is not functionality we've lost because of changes to our code. It seems a shame that their calendar doesn't do that, but that is definitely __not__ your problem to fix.

static/js/components/NewShift.jsx
```diff
@@ -12,6 +12,11 @@ import { fetchShifts } from "../ducks/shifts";
 
 dayjs.extend(utc);
 
+function isDailyShift(shiftName) {
+  const regex = /\d+\/\d+/;
+  return regex.test(shiftName);
```
guynir42, line 17:
```suggestion
function isDailyShift(shiftName) {
  const regex = /\d+\/\d+$/;
  return regex.test(shiftName);
```
The added `$` at the end makes sure it only matches the pattern at the end of the string.

static/js/components/NewShift.jsx
```diff
@@ -62,6 +67,11 @@ const NewShift = () => {
   };
 
   function validate(formData, errors) {
+    if (isDailyShift(formData.name)) {
+      errors.name.addError(
+        'Shift name cannot contain "number/number", please use a different name.'
```
guynir42, line 72: "...cannot contain number/number at the end of the name..."

static/js/components/ShiftCalendar.jsx
```diff
@@ -15,9 +15,14 @@ let currentUser;
 const localizer = momentLocalizer(moment);
 let groups;
 
+function isDailyShift(shiftName) {
+  const regex = /\d+\/\d+/;
```
guynir42, line 19:
```suggestion
  const regex = /\d+\/\d+$/;
```

Theodlz:
Ok ! Other than that, are there any changes you need me to make ?

static/js/components/ShiftCalendar.jsx
```diff
@@ -55,6 +60,13 @@ async function handleSelectSlot({ start, end }) {
     }
   } else if (name === "") {
     dispatch(showNotification("Shift not created, no name given", "error"));
+  } else if (isDailyShift(name)) {
+    dispatch(
+      showNotification(
+        'Shift not created, invalid name (dont use "number/number")',
```
guynir42, line 66: "... at end of name"

guynir42 (approved):
I have a minor improvement. After that ready to merge. Thanks for a very nice feature @Theodlz

Theodlz:
> I have a minor improvement. After that ready to merge. Thanks for a very nice feature @Theodlz

Done 🚀

guynir42 (approved):

profjsb:
@Theodlz can you put a patch in to fix the above? Also, @guynir42 take note of the no carrot requirement.

Theodlz:
Hi @profjsb . I added those changes, should i just open a new pull request so it can be merged ?

guynir42:
@Theodlz either add a new PR or fix this in the next PR you are putting in on this subject.

Theodlz:
Done @guynir42, you can find it here : https://github.com/skyportal/skyportal/pull/2848

PR #2829, opened 2022-04-06
Add ID column to spectrum get

guynir42:
When using spectrum GET for querying multiple spectra, when the parameter `minimalPayload` is `true` the returned data doesn't include the ID of the spectra, which it really should return.


mcoughlin (approved):
Nice @guynir42

PR #2830, opened 2022-04-07
Upgrade react-router-dom from 5.3.0 to 6.3.0

mcoughlin:
This PR upgrades react-router-dom from 5.3.0 to 6.3.0. 

See for example here:
https://stackoverflow.com/questions/63124161/attempted-import-error-switch-is-not-exported-from-react-router-dom


guynir42 (approved):

mcoughlin:
Thanks @guynir42

PR #2831, opened 2022-04-07
Adding missing packages that broke pycurl

Theodlz:
Recently, skyportal couldn't get installed properly as pycurl (package installed with pip) was broken. It seems that adding these 2 packages fixed it. I would suggest to the reviewer to make sure they aren't already installed on his distro before trying it so he can compare and verify their necessity.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -10,8 +10,6 @@ how to install them on MacOS and Debian-based systems below.
 - NGINX (v>=1.7)
 - PostgreSQL (v>=9.6)
 - Node.JS/npm (v>=5.8.0)
-- libcurl4-gnutls-dev
-- libgnutls28-dev
 
 ## Source download, Python environment
 
@@ -36,7 +34,7 @@ These instructions assume that you have [Homebrew](http://brew.sh/) installed.
 1. Install dependencies
 
 ```
-brew install supervisor nginx postgresql node libcurl4-gnutls-dev libgnutls28-dev
+brew install supervisor nginx postgresql node
 ```
 
 2. Start the PostgreSQL server:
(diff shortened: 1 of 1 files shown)


doc/setup.md
```diff
@@ -34,7 +36,7 @@ These instructions assume that you have [Homebrew](http://brew.sh/) installed.
 1. Install dependencies
 
 ```
-brew install supervisor nginx postgresql node
+brew install supervisor nginx postgresql node libcurl4-gnutls-dev libgnutls28-dev
```
mcoughlin, line 39: These don't exist on mac, but I haven't had problems locally, so I expect curl headers are already available.

doc/setup.md
```diff
@@ -10,6 +10,8 @@ how to install them on MacOS and Debian-based systems below.
 - NGINX (v>=1.7)
 - PostgreSQL (v>=9.6)
 - Node.JS/npm (v>=5.8.0)
+- libcurl4-gnutls-dev
```
mcoughlin, line 13: Since I think this comes up mostly on Linux systems, maybe we just discuss it explicitly in that section?

mcoughlin (changes requested):
Some comments on Mac vs Linux.

mcoughlin:
@Theodlz Can you add a few words before the install noting the extra purpose for the Linux packages? (i.e. to install pycurl)?

Theodlz:
> @Theodlz Can you add a few words before the install noting the extra purpose for the Linux packages? (i.e. to install pycurl)?

Done ! I'll let you have a look @mcoughlin, tell me what you think of it.

mcoughlin (approved):
Thanks @Theodlz

PR #2834, opened 2022-04-07
New source should be posted to one of the accessible groups

### LINKED ISSUE #2833: Form to add a source - can be added to single group user
I'm not 100% about this, but from what i understood a source belongs to a proper group, not a single group user which is just one user. Here, in the form to add a source, single group users can be selected

image

mcoughlin:
This PR sets the groups from the front end that are accessible for new source posting.

Closes #2833


Theodlz (approved):

guynir42 (approved):

mcoughlin:
Thanks @guynir42

PR #2835, opened 2022-04-08
Telescope map

Theodlz:
This PR aims to add a world map showing where the Telescopes are, and allowing you to see more information about them and the current time and weather conditions where they are located.

This PR focuses on the frontend, by adding only the map as well as telescope basic informations.

It will be followed by more PRs adding the backend necessary before being able to display telescope availability at a given time

#### Latest version:
telescope_map_better

#### First version:
TELESCOPE_MAP


Theodlz:
@mcoughlin  tagging you here so you'll see it. I used react-simple-maps

mcoughlin:
@Theodlz Could you post a quick screen shot?

Theodlz:
> @Theodlz Could you post a quick screen shot?
here you go @mcoughlin. Please keep in mind that this is a work in progress, the UI needs to be improved by a lot.

TELESCOPE_MAP

Theodlz:
> @Theodlz Pretty cool!

Hey @mcoughlin, i added the "nested" telescope feature to avoid telescope names overlapping each other. What do you think ?

TELESCOPE_MAP_v2

mcoughlin:
@Theodlz Looking nice. Maybe we consider just doing the nickname instead of the whole name? Also, I feel like this map can just go above the existing list of telescopes and add new telescope form?

Theodlz:
@mcoughlin Noted for the nicknames, will do. I didnt understand what you meant for the rest though. Replacing the previous telescope list by the map + the list of selected telescopes on the right -> i completely agree. But for the add new telescope form i didn't get if you mean that we should replace it or not ?

mcoughlin:
@Theodlz I just meant that I think all of this can just appear on the page together? Instead of the click button up top to have to switch back and forth with the button.

Theodlz:
@mcoughlin seems doable, but im not sure we could fit all of those in one page in a way that keeps it clean, maybe a switch just between the  new list of selected telescopes (from the map, the list you see on the right here) and the form to add a telescope ?

mcoughlin:
@Theodlz yeah something to think about, it is probably a good idea to minimize the mess but I also assume people mostly know how to scroll

Theodlz:
@mcoughlin of course, but if most of an element (or all it it) is out of the viewport you dont necessarily notice that its there or that you can scroll . Maybe we can think of a way to minimize/develop the form or the list respectively when you click on its "header" rather than have to click a button to switch between the 2. Its cleaner don't you think ?

mcoughlin:
@Theodlz true, I think it just might be nice to have the map and the add form available so you can see your system appear, for example.

Theodlz:
@ Ok ! I'll do my best to get the 2 of them on the same page in a nice way. Have a nice day, and thanks for your advice !

Theodlz:
Hi @mcoughlin, i changed the map to add some of the changes you asked me to do + i make it a bit prettier and added the frontend tests for it.

Here's a preview:

telescope_map_better

mcoughlin:
@Theodlz Looking really cool.

Theodlz:
@mcoughlin I have to take care of the linting before doing a proper PR and not just a DRAFT, but do you think it will be ready for a review right after that ?

mcoughlin:
@Theodlz definitely.

Theodlz:
Linting is done @mcoughlin , ready for a first review !

data/db_demo.yaml
```diff
@@ -112,8 +112,8 @@ telescope:
     =id: P60
   - name: Nordic Optical Telescope
     nickname: NOT
-    lat: 28.75
-    lon: 17.88
+    lat: 28.7572
```
mcoughlin, line 115: Glad we caught these ;)

package.json
```diff
@@ -23,6 +23,7 @@
     "core-js": "3.17.3",
     "css-loader": "5.2.6",
     "d3": "7.4.0",
+    "d3-array": "^3.1.5",
```
mcoughlin, line 26: No carrots, pin to a release.

package.json
```diff
@@ -33,10 +34,12 @@
     "exports-loader": "1.1.1",
     "geojson-prop-types": "0.2.0",
     "imports-loader": "1.2.0",
+    "install": "^0.13.0",
     "material-ui-phone-number": "2.2.6",
     "mathjs": "10.0.2",
     "moment": "^2.29.1",
     "mui-datatables": "3.8.2",
+    "npm": "^8.6.0",
```
mcoughlin, line 42: Don't we install npm separately? At least the check is done separately?

package.json
```diff
@@ -33,10 +34,12 @@
     "exports-loader": "1.1.1",
     "geojson-prop-types": "0.2.0",
     "imports-loader": "1.2.0",
+    "install": "^0.13.0",
```
mcoughlin, line 37: No carrot.

package.json
```diff
@@ -62,6 +65,7 @@
     "react-responsive": "8.2.0",
     "react-router-dom": "6.3.0",
     "react-simple-dropdown": "3.2.3",
+    "react-simple-maps": "^2.3.0",
```
mcoughlin, line 68: No carrot.

static/js/components/TelescopeMap.jsx
```diff
@@ -0,0 +1,210 @@
...
+import {
+  ComposableMap,
+  Geographies,
+  Geography,
+  Marker,
+  useZoomPan,
+} from "react-simple-maps";
+
+const geoUrl =
+  "https://raw.githubusercontent.com/zcreativelabs/react-simple-maps/master/topojson-maps/world-110m.json";
```
mcoughlin, line 13: If one doesn't have access to the internet and is developing, what does the app do?

static/js/components/TelescopeMap.jsx
```diff
@@ -0,0 +1,210 @@
...
+          </>
+        )}
+      </CustomZoomableGroup>
+    </ComposableMap>
+  );
+};
+
+// prop validation
+
+// SourceSaveHistory.propTypes = {
```
mcoughlin, line 161: Remove comments?

static/js/components/TelescopePage.jsx
```diff
@@ -70,7 +89,7 @@ const TelescopeList = ({ telescopes }) => {
   return (
     <div className={classes.root}>
       <List component="nav">
-        {telescopes?.map((telescope) => (
```
mcoughlin, line 73: Is the ? not a good thing?

mcoughlin (changes requested):
@Theodlz A few comments. This is an objectively cool addition to the app.

static/js/components/TelescopeMap.jsx
Theodlz: i didn't think about this... should we directly add the svg to the data folder ? especially because its a very tiny file.

static/js/components/TelescopePage.jsx
Theodlz: i actually didn't know for sure what it does. I'll put it back if you think that it is better.

data/db_demo.yaml
Theodlz: Yes definitely ! Another strong point for the map i guess :)

static/js/components/TelescopeMap.jsx
mcoughlin: Could you give it a try? Not quite sure about permissioning, but seems like our goal is to limit the things we must be connected to the internet for.

static/js/components/TelescopePage.jsx
mcoughlin: Yeah ? basically won't loop if telescopes is null, in case the variable is reloading and such.

static/js/components/TelescopeMap.jsx
Theodlz: Ok, so it seems to work. I even tried accessing it from another device (my phone) and as a user with the most basic skyportal rights and it seemed to work. One issue i noticed: on smaller screen and portrait mode (like a phone), or even on the computer if i resize the window by a lot, the map doesnt show up anymore.

static/js/components/TelescopeMap.jsx
```diff
@@ -0,0 +1,195 @@
...
+import { useDispatch } from "react-redux";
+import {
+  ComposableMap,
+  Geographies,
+  Geography,
+  Marker,
+  useZoomPan,
+} from "react-simple-maps";
+
+import world_map from "../../images/maps/world-110m.json";
```
mcoughlin, line 12: Is this committed now somewhere?

mcoughlin (approved):
I assume the map needs to be committed somewhere, otherwise, looks good to me.

static/js/components/TelescopeMap.jsx
Theodlz: Yes ! Forgot it in the first commit hahaha. I noticed something, the map doesnt show up in smaller screens. Maybe i could put the normal list back for smaller screens only, otherwise find a way to fix that on the map

mcoughlin: @Theodlz I.e. the behavior is different for the local vs online loading?

Theodlz: Not at all ! Maybe my last message isnt clear enough. Replacing the link by the file itself works perfectly, no issue on that. I was mentionning another issue of how it doesnt display on mobile or small screens, but that has nothing to do with replacing the link by the file directly :).

mcoughlin: Ah got ya. There are other places in the app where desktop vs mobile experience is different, so could just lean into that and explicitly only call the map on desktop and otherwise just show the list form?

Theodlz:
I'll have a look at that at the start of the week then. Have a nice weekend michael!

static/js/components/TelescopeMap.jsx
Theodlz: @mcoughlin  done ! Just used the previous version (simple list) on small screens, and the map on bigger devices. Strange thing, but that issue of the map disappearing only occurs on google chrome, the map resizes and works perfectly on firefox...

mcoughlin (approved):
Thanks for the nice work @Theodlz

static/js/components/TelescopeInfo.jsx
```diff
@@ -0,0 +1,88 @@
...
+              {telescope.name} ({telescope.nickname})
+            </h2>
+            {telescope.robotic ? (
+              <h3 className={classes.h3}>Robotic : Yes</h3>
+            ) : (
+              <h3 className={classes.h3}>Robotic : No</h3>
+            )}
+            <h3 className={classes.h3}>Diameter : {telescope.diameter} </h3>
+            <h3 className={classes.h3}>
+              Location : {telescope.lat}, {telescope.lon}
```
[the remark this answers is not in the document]
Theodlz: 4 digits after the comma should be good for lon lat what do you think ?

static/js/components/TelescopeInfo.jsx
```diff
@@ -0,0 +1,88 @@
...
+            </h3>
+            <h3 className={classes.h3}>Elevation : {telescope.elevation}</h3>
```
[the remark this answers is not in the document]
Theodlz: For the elevation ? Diameter ?

stefanv:
There are quite a few new JS dependencies here; please ensure that each one is justified (`install`?). Then, please see the usage package in, e.g., `VegaPhotometry.jsx` and `ManageDataForm.jsx` for how to lazily import libraries. This will ensure that the JS gets bundled separately, and only gets used when the map is loaded.

Theodlz:
> There are quite a few new JS dependencies here; please ensure that each one is justified (`install`?). Then, please see the usage package in, e.g., `VegaPhotometry.jsx` and `ManageDataForm.jsx` for how to lazily import libraries. This will ensure that the JS gets bundled separately, and only gets used when the map is loaded.

Hi @stefanv, thanks for your review. When using React.lazy() to load the map, should i use it to import the TelescopeMap.jsx component in TelescopeMapDesktop.jsx, or should i rather use it on the TelescopePageDesktop.jsx component directly (in which the map is imported). I think the map is enough, which is what i did for now.

Theodlz:
I had a question regarding the tests, when you have a component that has a mobile and desktop version, do you test both ? I've been looking for similar examples in the tests/frontend directory but i can't find any.

guynir42:
@Theodlz Usually we don't test both, but that's not ideal. If you have two different components come up for mobile vs desktop then a test is even more important. I suggest recycling the previous "telescope list" tests for the mobile. But I am not sure how to get selenium to run the test on a mobile browser (the answer is probably easy to google).

static/js/components/TelescopeMap.jsx
```diff
@@ -37,12 +38,9 @@ function telescopelabel(nestedTelescope) {
...
   } else {
-    for (let i = 0; i < nestedTelescope.telescopes.length; i += 1) {
-      label += nestedTelescope.telescopes[i].nickname;
-      if (i < nestedTelescope.telescopes.length - 1) {
-        label += " / ";
-      }
-    }
+    label = nestedTelescope.telescopes
+      .map((telescope) => telescope.nickname)
+      .join(" / ");
```
guynir42, line 43: Good. I think you may be able to remove the `if length===1` clause altogether.

static/js/components/TelescopeMap.jsx
```diff
@@ -98,12 +90,16 @@ const TelescopeMap = ({ telescopes }) => {
       for (let j = 0; j < nestedTelescopes.length; j += 1) {
         if (
           Math.abs(
-            normalizeLatitude(filteredTelescopes[i].lat) -
-              normalizeLatitude(nestedTelescopes[j].lat)
+            normalizeLatitudeDiff(
```
guynir42, line 93: Is there a difference between `abs` and `Math.abs`?

static/js/components/TelescopePage.jsx
```diff
@@ -1,140 +1,31 @@
...
+  useEffect(() => {
+    const handleResize = () => {
+      if (ref.current !== null) {
+        setWidth(ref.current.offsetWidth);
+      }
+    };
 
-  const classes = useStyles();
+    window.addEventListener("resize", handleResize);
+  }, [ref]);
```
guynir42, line 23: Nice.

guynir42 (commented):
Looks good. I have minor suggestions. And probably should add the mobile test.

Theodlz:
> @Theodlz Usually we don't test both, but that's not ideal. If you have two different components come up for mobile vs desktop then a test is even more important. I suggest recycling the previous "telescope list" tests for the mobile. But I am not sure how to get selenium to run the test on a mobile browser (the answer is probably easy to google).

Hey @guynir42 , i changed what you asked me to + added the 2 different tests for mobile and Desktop. Maybe something worth doing in a next PR would be to do the same for other pages that have a different layout/components between desktop and mobile.

guynir42 (approved):
I'm going to merge this, but I think we should have a separate PR to try to move as many tests as possible from "flaky" to the regular workflow.

Theodlz:
> I'm going to merge this, but I think we should have a separate PR to try to move as many tests as possible from "flaky" to the regular workflow. 

Thanks for merging it @guynir42 ! When it comes to fixing the flakiness of tests I completely agree. I'm thinking about dedicating tomorrow to opening a bunch of PR to reduce/fix the flakiness of those frontend tests, what do you think of that ?

guynir42:
That sounds great. If you can get some of the new tests that were recently added to the flaky workflow into the regular workflow that would be fantastic.

Theodlz:
Perfect. I'll look into those. I think i'll proceed like this : reduce the flakiness of a test when running it n times in a row -> reducing its flakiness when running it along with the all the other flaky tests n times in a row -> checking if it is now consistent when running it n times in a row with the whole tests suite

PR #2837, opened 2022-04-08
Resolved issue where GcnEvent page was crashing

thomasculino:
This PR aims to resolve the issue where the GCN Event page was crashing. It behaved like this:
gcn event page crash
 
Now the page behaves like this
Peek 2022-04-08 18-25

It isn't perfect but the page does not crash anymore.


mcoughlin (approved):
A clear step-up, thanks @thomasculino.

PR #2838, opened 2022-04-08
Small instrument page / observing run page improvements

mcoughlin:
This PR makes small quality of life improvements to the instrument and observing run page. Instrument page adds a space between the filter strings so the string wraps. Observing run page adds checkbox to display near-term runs by default.

### DIFF
--- a/static/js/components/ObservingRunPage.jsx
+++ b/static/js/components/ObservingRunPage.jsx
@@ -70,11 +70,11 @@ const ObservingRunList = ({ observingRuns }) => {
   const [displayAll, setDisplayAll] = useState(false);
 
   const toggleDisplayAllCheckbox = () => {
-    setChecked(!checked);
+    setDisplayAll(!displayAll);
   };
 
   let observingRunsToShow = [];
-  if (!checked) {
+  if (!displayAll) {
     observingRuns?.forEach((run) => {
       const dt = dayjs.duration(dayjs(run.calendar_date).diff(nowDate));
       if (dt.$ms < dt_month.$ms && dt.$ms > 0) {
@@ -88,6 +88,13 @@ const ObservingRunList = ({ observingRuns }) => {
   return (
     <div className={classes.root}>
       <List component="nav">
+        <input
+          type="checkbox"
+          onChange={toggleDisplayAllCheckbox}
+          name="observationRun"
+          data-testid="observationRunCheckbox"
+        />
+        Display all observing runs? &nbsp;&nbsp;
         {observingRunsToShow?.map((run) => (
           <ListItem button component={Link} to={`/run/${run.id}`} key={run.id}>
             <ListItemText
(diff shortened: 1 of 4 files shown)


mcoughlin:
@Theodlz can you also add a review please?

static/js/components/ObservingRunPage.jsx
```diff
@@ -23,16 +34,61 @@ const useStyles = makeStyles((theme) => ({
...
 const ObservingRunList = ({ observingRuns }) => {
   const classes = useStyles();
   const { instrumentList } = useSelector((state) => state.instruments);
   const { telescopeList } = useSelector((state) => state.telescopes);
   const groups = useSelector((state) => state.groups.all);
 
+  const nowDate = dayjs().utc().format("YYYY-MM-DDTHH:mm:ssZ");
+  const dt_month = dayjs.duration(1, "month");
+
+  const [checked, setChecked] = useState(false);
```
jadalilleboe, line 70:
```suggestion
  const [displayAll, setDisplayAll] = useState(false);
```

static/js/components/ObservingRunPage.jsx
```diff
@@ -41,9 +97,17 @@ const ObservingRunList = ({ observingRuns }) => {
                 telescopeList,
                 groups
               )}
+              secondary={observingRunInfo(run, instrumentList, telescopeList)}
             />
           </ListItem>
         ))}
+        <input
```
jadalilleboe, line 104: I think this should be at the top of the list instead of the bottom. So if the user changes their mind and wants to uncheck they don't have to scroll all the way to the bottom.

static/js/components/ObservingRunPage.jsx
```diff
@@ -23,16 +34,61 @@ const useStyles = makeStyles((theme) => ({
...
   const { instrumentList } = useSelector((state) => state.instruments);
   const { telescopeList } = useSelector((state) => state.telescopes);
   const groups = useSelector((state) => state.groups.all);
 
+  const nowDate = dayjs().utc().format("YYYY-MM-DDTHH:mm:ssZ");
+  const dt_month = dayjs.duration(1, "month");
+
+  const [checked, setChecked] = useState(false);
+
+  const toggleCheckbox = () => {
```
jadalilleboe, line 72:
```suggestion
  const toggleDisplayAllCheckbox = () => {
```

jadalilleboe (approved):

static/js/components/ObservingRunPage.jsx
mcoughlin: Good idea!

Theodlz:
It's great @mcoughlin. I'm just wondering whats best on the observing runs page for the date input:
- mm/dd/yyyy
- todays date in that format as a clearer example

It's a very minor detail and i think its more a matter of preferences than an actual improvement. My thinking is that the mm/dd/yyyy is a bit less user friendly for someone that is not used to coding. But considering the people that are using SkyPortal, i don't think it matters that much.

Otherwise, it's a yes for me 🚀

mcoughlin:
Thanks @Theodlz and @jadalilleboe.

PR #2839, opened 2022-04-08
Spectrum PUT without groups

guynir42:
Currently if you use PUT to change a spectrum and do not specify any groups it fails. 
The correct behavior is that it just leaves the existing groups. 
This PR fixes that by catching cases where `group_ids` is None or not given.


skyportal/handlers/api/spectrum.py
```diff
@@ -47,7 +47,9 @@
 class SpectrumHandler(BaseHandler):
     def get_groups(self, group_ids):
         groups = []
-        if group_ids == "all":
```
mcoughlin, line 50: Shouldn't this return an error? Or what is the context where we need this to return None?

skyportal/tests/api/test_spectrum.py
```diff
@@ -11,6 +11,124 @@
...
+    )
+    assert status == 200
+    assert data['status'] == 'success'
+    assert data['data']['label'] == custom_label
+    group_ids = [g['id'] for g in data['data']['groups']]
+    assert group_id1 in group_ids  # PUT is only allowed to remove groups
+    assert group_id2 in group_ids
+    num_groups = len(data['data']['groups'])
+
+    # adding the same group ID doesn't make redundant groups
```
mcoughlin, line 112: Shouldn't this instead tell us that the group already has access?

mcoughlin (changes requested):
The code looks good, just some questions on whether the choice of behavior makes sense (i.e. raising errors vs returning success).

skyportal/handlers/api/spectrum.py
guynir42: one of the usages tests that groups is True, but I think that it may be easier to check if group_id is not None or empty.

skyportal/tests/api/test_spectrum.py
guynir42: I think that in general when you PUT you can post the same data with some of it changed and some as before, without invoking any errors/warnings. I don't know if it gives the user much to know that some of the data is identical to what existed before the PUT. 
For groups there's a bit of a difference in that you can't remove a group from a spectrum with PUT, but that's probably just so people don't accidentally delete spectra from groups they are not even aware of, when they are just meaning to add a group. 
I don't think that has any bearing on which groups already existed.

mcoughlin: ok consistency with existing behavior seems like a good argument.

mcoughlin (approved):
Thanks for helping me understand @guynir42 .

PR #2841, opened 2022-04-12
Add type column to user notification table

### LINKED ISSUE #2245: Add `type` column to `UserNotification` table
Without it, the Slack integration has to parse the notification text to know if it is a mention, e.g.

jadalilleboe:
Closes #2245


mcoughlin:
@jadalilleboe Is it maybe worth using this same PR to modify https://github.com/skyportal/skyportal/blob/master/skyportal/models/user_notification.py#L124 to use your new column and check based on the flag ?

jadalilleboe:
@mcoughlin Yeah for sure! I wasn't sure where this was supposed to be used.

mcoughlin:
@jadalilleboe Ah yeah the main goal is to remove these is_* statements:
https://github.com/skyportal/skyportal/blob/master/skyportal/models/user_notification.py#L90
which I shamefully wrote ;)

mcoughlin:
@guynir42 @jadalilleboe  for the notification types, should we make them camel case maybe?

jadalilleboe:
ye

jadalilleboe:
@mcoughlin Is there anything else that should be added here? Or do you think this is ready for review?

mcoughlin (approved):
Thanks for fixing that @jadalilleboe !

guynir42 (approved):
LGTM. Please make sure the migration hash is up to date before merging.

PR #2847, opened 2022-04-14
Fixing files preview and download when attached to a comment

thomasculino:
The preview and download of files attached to a comment wasn't working.


static/js/components/CommentAttachmentPreview.jsx
```diff
@@ -124,10 +124,23 @@ const CommentAttachmentPreview = ({
   const theme = useTheme();
   const darkTheme = theme.palette.type === "dark";
 
+  function ressourceType(state) {
```
mcoughlin, line 127: resourceType

static/js/components/CommentAttachmentPreview.jsx
```diff
@@ -124,10 +124,23 @@ const CommentAttachmentPreview = ({
...
+    let type = "";
+    if (associatedResourceType === "gcn_event") {
+      type = state.gcnEvent.commentAttachment;
+    } else if (
+      associatedResourceType === "sources" ||
+      associatedResourceType === "spectra"
+    ) {
+      type = state.source.commentAttachment;
+    } else {
+      console.log(`Nothing corresponds to a right ${associatedResourceType}`);
```
mcoughlin, line 137: We probably dont need this else, right? Otherwise I think this should be a showNotification error, because the log here won't help.

static/js/components/CommentAttachmentPreview.jsx
```diff
@@ -149,7 +162,9 @@ const CommentAttachmentPreview = ({
 
   let jsonFile = {};
   try {
-    jsonFile = isCached ? JSON.parse(commentAttachment.attachment) : {};
+    jsonFile = isCached
+      ? JSON.parse(Object.entries(commentAttachment)[1][1])
```
mcoughlin, line 166: What is this [1][1] thing?

mcoughlin (changes requested):
Thanks @thomasculino. A few small questions.

static/js/components/CommentAttachmentPreview.jsx
thomasculino: Object.entries(commentAttachment) is an array of arrays as you can see below
image

and if you access the first array at index 1, you will access the attachment array and if you access this array at index 1, you will have the content of your json file (as this occur only when dealing with json files)

static/js/components/CommentAttachmentPreview.jsx
thomasculino: True, I don't think we need to keep it

guynir42 (approved):

mcoughlin:
@thomasculino can you check and fix the API test failures?

mcoughlin (approved):
Thanks @thomasculino

PR #2848, opened 2022-04-14
Shift Management - Display Admins and handle permissions

Theodlz:
### Here's how it looks:
shift_admin


skyportal/handlers/api/shift.py
```diff
@@ -136,32 +136,66 @@ def get(self, group_id=None):
...
+                    "last_name": gu.user.last_name,
+                    "contact_email": gu.user.contact_email,
+                    "contact_phone": gu.user.contact_phone,
+                    "admin": gu.admin,
+                }
+                for gu in shift.group.group_users
+            ]
+            shift = shift.to_dict()
+            shift["users"] = susers
+            shift.pop('shift_users', [])
```
mcoughlin, line 193: What is this doing?

skyportal/handlers/api/shift.py
```diff
@@ -136,32 +136,66 @@ def get(self, group_id=None):
...
+            shift["group"] = shift["group"].to_dict()
+            shift["group"]["users"] = gusers
+            shift["group"].pop('group_users', [])
```
mcoughlin, line 196: This one too.

mcoughlin (approved):
Just curious about some of the pops. Otherwise looks good to me.

skyportal/handlers/api/shift.py
Theodlz: Hey @mcoughlin , the `shift` is not a dict but a `Shift` object, and i can't assign a new values to it, so i need to convert it in a dict. Then i set the `users` with the new `susers` i created, and i get rid of `shift_users` that is useless now that i have a `users` key. I could just set the `shift_users` to `susers`. I did that because i used the `users` everywhere in the frontend, but if you want me to avoir using pop, i can just replace `users` by `shift_users` in the frontend

mcoughlin: Is that a painful change? If it is more direct, it seems cleaner.

Theodlz: Ok ! I'll do that tomorrow then. Have a nice day Michael

Theodlz:
Hey @mcoughlin . I changed what you asked me to.

mcoughlin (approved):
Thanks for fixing that @Theodlz .

skyportal/handlers/api/shift.py
```diff
@@ -136,32 +136,64 @@ def get(self, group_id=None):
...
+            gusers = [
+                {
+                    "id": gu.user.id,
+                    "username": gu.user.username,
+                    "first_name": gu.user.first_name,
+                    "last_name": gu.user.last_name,
+                    "contact_email": gu.user.contact_email,
+                    "contact_phone": gu.user.contact_phone,
+                    "admin": gu.admin,
+                }
```
guynir42, line 188: Did you try using `user.to_dict()`? Here and above.

skyportal/tests/api/test_shifts.py
```diff
@@ -34,7 +34,7 @@ def test_shift(public_group, super_admin_token, view_only_token, super_admin_use
 
     assert any(
         [
-            len([s for s in s['users'] if s['id'] == super_admin_user.id]) == 1
+            len([s for s in s['shift_users'] if s['id'] == super_admin_user.id]) == 1
             for s in data['data']
```
guynir42, line 38: using `s` as the looping object in both internal and external loop is a little dangerous. Maybe rename one of them?

guynir42 (approved):
Looks good. Minor comments. Let me know if you need me to merge it when you are done.

skyportal/handlers/api/shift.py
Theodlz: Hey @guynir42 ! Thanks for your review. Few questions: What do you mean exactly ? Using user.to_dict() rather than creating a new dict ? I can try of course. The idea here was to be able to create a new user object with only the info we need for shift users, so their name, username, id, contact and if they are an admin or not. If i only use user.to_dict(), i also need to add the admin boolean to it right ? (that is contained in the shift_user, that itself contains user)

skyportal/tests/api/test_shifts.py
Theodlz: You're completely right, github copilot naming of variables makes us lazy 👀

skyportal/handlers/api/shift.py
guynir42: I think each object that comes from baselayer (or from sql alchemy, I'm not sure) has this `to_dict()` method. It just creates a dictionary from the properties of the ORM object that have been loaded. It will just save you a lot of trouble copying many properties manually. You can then delete or add some keys to fit exactly what you want. 
Make sure to verify you are getting all the properties you expect because some things aren't loaded by default from the database row and so they don't get into the dictionary automatically.

Theodlz: Great ! I'll try that and then ask for your opinion ! Thanks again.

Theodlz: Done 🚀

guynir42 (approved):

PR #2836, opened 2022-04-08
Add the ability to filter by origin and color change for attached files in comments

thomasculino:
This PR aims to add the ability to filter photometry by its origin. This PR is an extension of #2825 
Peek_filter_by_origin

There is also a small color change for uploaded files in comments. Their name was in white which was problematic when using the light mode because we couldn't see it. I just changed the color displayed so that we can see it in either mode.

In light mode:
Screenshot from 2022-04-08 17-11-18
In dark mode:
Screenshot from 2022-04-08 17-11-50

PR #2849, opened 2022-04-15
Draft: on top spectra

### LINKED ISSUE #1844: Feature request "in front' spectra
Currently when there are multiple spectra plotted, the "in front" ordering is fixed.  I would like to control the "in front" ordering.  One way to do this is through the order of toggling of spectra visibility.

jadalilleboe:
This is a first pass at letting the user choose which spectra should be on top, which will close #1844. It doesn't work yet, I need to figure out a way to refresh/rerender the plot (similar to the problem in #2836). It works if an on top spectra id is hard-coded though:
Screen Shot 2022-04-14 at 5 24 00 PM
Screen Shot 2022-04-15 at 8 32 52 AM
The main idea is that there will be a dropdown and the user can select which should be on top, this is the part that doesn't work yet.
Screen Shot 2022-04-15 at 8 39 49 AM

### DIFF
--- a/skyportal/handlers/api/internal/plot.py
+++ b/skyportal/handlers/api/internal/plot.py
@@ -47,6 +47,7 @@ class PlotSpectroscopyHandler(BaseHandler):
         device = self.get_query_argument("device", None)
         smoothing = self.get_query_argument("smoothing", False)
         smooth_number = self.get_query_argument("smoothNumber", 10)
+        on_top_spectra_id = self.get_query_argument("onTopSpectraId", None)
         # Just return browser by default if not one of accepted types
         if device not in device_types:
             device = "browser"
@@ -60,6 +61,7 @@ class PlotSpectroscopyHandler(BaseHandler):
                 device=device,
                 smoothing=smoothing,
                 smooth_number=smooth_number,
+                on_top_spectra_id=on_top_spectra_id,
             )
         except Exception as e:
             return self.error(f'Exception in photometry plot: {e}')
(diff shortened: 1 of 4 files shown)


jadalilleboe:
ezgif com-gif-maker (9)

guynir42:
@jada is this ready for review?

jadalilleboe:
@guynir42 yes!

guynir42:
Before looking at the code I pulled down the branch to try it out. For me (on chrome) when I click the dropdown menu it rerenders the component (which pops the menu up to the top of my screen). Then when I click it, it re-renders again. All of this is not great UX. Is there any way to change the order directly in the Bokeh plot without having to re-render the entire spectrum plot? (like when adding lines, etc)

jadalilleboe:
Agreed, it is not great UX. It seems to me that the spectral lines are fixed in their ordering once plotted, and the ordering quite literally depends on the order in which they were plotted in the code. Hence the need to rerender and run the code to plot the lines in a different order. I could be wrong though.

guynir42:
@jadalilleboe what happens if the JS custom code goes and replaces the underlying data from spectrum1 to spectrum2?

jadalilleboe:
not sure what you mean? @guynir42

guynir42:
@jadalilleboe we can do all sorts of modifications to the rendered spectra, using custom JS code. Look at the smoothing for example. When we change the smoothing it doesn't re-render the whole component. Not sure if the same would work for manually switching the arrays of spectra but it is worth a try.

jadalilleboe:
closing due to finding a much clearer approach, I am going to open a new PR so I don't have to fix this branch:)

PR #2855, opened 2022-04-15
Determine instrument tile color based on filters

jadalilleboe:
Determines color for the instrument tiles based on the filters, here's the color for these filters:
Screen Shot 2022-04-15 at 3 58 46 PM
And with just the first 3:
Screen Shot 2022-04-15 at 4 03 12 PM
Also I made the galaxy points green to distinguish from sources

### DIFF
--- a/static/js/components/LocalizationPlot.jsx
+++ b/static/js/components/LocalizationPlot.jsx
@@ -246,6 +246,22 @@ const GeoJSONGlobePlot = ({
       return gdistance < 1.57;
     };
 
+    const filtersToColor = (filters) => {
+      const filterStr = filters.join("");
+      let hash = 0;
+      for (let i = 0; i < filterStr.length; i += 1) {
+        // eslint-disable-next-line no-bitwise
+        hash = filterStr.charCodeAt(i) + ((hash << 5) - hash);
+      }
+      let color = "#";
+      for (let i = 0; i < 3; i += 1) {
+        // eslint-disable-next-line no-bitwise
+        const value = (hash >> (i * 8)) & 0xff;
+        color += `00${value.toString(16)}`.substr(-2);
+      }
+      return color;
+    };
+
     function refresh() {
       svg.selectAll("*").remove();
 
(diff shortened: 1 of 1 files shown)


mcoughlin:
@jadalilleboe what happens if there are multiple filters requested per field? is it just the first one?

jadalilleboe:
@mcoughlin it's a hash that combines all the filters into a string and then computes a hex code from that.

mcoughlin:
@jadalilleboe i am kind of wondering if the ability to cycle through would make sense looking ahead to when we want to add multiple filters per field?

jadalilleboe:
@mcoughlin what do you mean cycle through?

mcoughlin:
Like I want to be able to add any field with any possible filter to it, so I think maybe one intuitive way to do it is by clicking and offering any set of filters. One way to do it would be changing color with every click, another would maybe be like piece of pie with each optional filter in the colors?

jadalilleboe:
I'm gonna close this, I'll open a new PR if I start working on this.

mcoughlin:
@jadalilleboe I think what you have now is better than before for sure, so I would think we should get this cleaned up and merged.

jadalilleboe:
ok!

static/js/components/LocalizationPlot.jsx
```diff
@@ -363,7 +380,7 @@ const GeoJSONGlobePlot = ({
           .data(galaxies.features)
           .enter()
           .append("circle")
-          .attr("fill", (d) => (visibleOnSphere(d) ? "red" : "none"))
+          .attr("fill", (d) => (visibleOnSphere(d) ? "green" : "none"))
```
mcoughlin, line 383: This was just to differentiate galaxies from sources or some other reason?

mcoughlin (approved):
This looks nice @jadalilleboe

static/js/components/LocalizationPlot.jsx
jadalilleboe: yeah!

guynir42 (approved):
LGTM

Issue #2396, opened 2021-10-28
Requests for photometry and spectroscopy plots


erezimm:
Other than the issues presented in [https://github.com/skyportal/skyportal/issues/2395](url) we would also like to ask for adjustments in the way plots are presented in the object page:

1. We think the right pane in the object page should be smaller or preferably collapsable to allow for larger photometry and spectroscopy plots. 
2. The data points in the photometry points are quite small and it would be nice if they could be larger
3. A select all/deselect all button would be much appreciated in the legends of both plots.

guynir42:
number 1 seems to be complicated, but I will leave it here for now. 
number 2 not sure if want to make the points much bigger, we may consider making it customizable. 
number 3 is closed via Skyportal #2778

PR #2857, opened 2022-04-18
Add ability to hide right pane on source page

jadalilleboe:
Adds the ability to hide the right pane on the source page to allow for larger plots. Fixes part of #2396 
ezgif com-gif-maker (8)
Also I made the width of the plots responsive to the size of the div containing them, previously it was hardcoded to 800.


profjsb (approved):

guynir42 (approved):
Looks great. Can you make a PR to fritz to add this code to the source page over there as well?

jadalilleboe:
@guynir42 Yeah for sure!

PR #2863, opened 2022-04-21
Adjustable data point size for photometry plot

### LINKED ISSUE #2396: Requests for photometry and spectroscopy plots

jadalilleboe:
Allows the user to adjust the size of the data points on the photometry plot. With #2857 and #2778, this should close #2396
ezgif com-gif-maker (8)

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -740,10 +740,10 @@ def photometry_plot(obj_id, user, width=600, device="browser"):
         model_dict[key] = plot.scatter(
             x='mjd',
             y='flux',
-            size=4,
             color='color',
             marker=factor_mark('instrument', markers, instruments),
             fill_color=color_dict,
+            fill_alpha=0.1,
             alpha='alpha',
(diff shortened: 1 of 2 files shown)


guynir42 (approved):

jadalilleboe:
Screen Shot 2022-04-21 at 1 04 31 PM
@guynir42 added a fill alpha to make the points opaque when larger

guynir42 (approved):

guynir42:
@jadalilleboe looks great. Let me know if you need me to merge.

PR #2864, opened 2022-04-21
added missing steps to create an administrative user

Theodlz:
Currently in SkyPortal's documentation, if you run the code provided to make a user an administrative user, you'll get an error because there is no DBsession created (no connection to the db). What is missing is the initialization of the connection to the DB being used. 

This PR addresses this issue by adding the missing lines of code.

### EDIT: 
The objective of this PR changed, it uses to be about only adding missing lines of code to the documentation about creating users, but now we decided to replace that by a script/tool to allow users to do so without having to type python code in their terminal, which is not very 'clean'. That new script can be found in `tools/elevate_user.py`, and you can find the associated documentation where the documentation on how to add users used to be.


mcoughlin (approved):
Much clearer. Thanks @Theodlz .

stefanv:
Can we factor this out into a little script that we ship along with `baselayer`? The script can be used to list existing users, and then the user can specify which one should become admin.

Theodlz:
Sounds great! Would be great if that was a command you could run as a a `make` command. Something like `make elevate`, without an argument it lists the users that are not admin yet, and asks you for a username or an id, but if you run `make elevate "the_user_name"` it directly makes that user a super user. 

What do you think ?

stefanv:
That'd be neat, but unfortunately it's very difficult to pass arguments to make targets.  A standalone tool would be fine, though!

Theodlz:
> Can we factor this out into a little script that we ship along with `baselayer`? The script can be used to list existing users, and then the user can specify which one should become admin.

Hi @stefanv , i have a little script ready but it users the `make_super_user` function from `skyportal.model_util`, which from what i saw uses things specific to skyportal. So if we want that tool to be shipped with baselayer rather than SkyPortal, both this tool and `skyportal.model_util` need to be transfered in baselayer. If you want to have a look at the script i added, it is in `tools`, and i added the documentation to use it.

tools/elevate_user.py
```diff
@@ -0,0 +1,65 @@
...
+
+try:
+    username = sys.argv[1]
+except IndexError:
+    username = None
+
+cfg = load_config()
+init_db(**cfg['database'])
+
+# get first argument
```
mcoughlin, line 18: What does this mean?

mcoughlin (approved):
Seems like a nifty idea to me.

mcoughlin (approved):

tools/elevate_user.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            print('\nExiting...')
+            sys.exit()
+
+
+def main():
+    if username is not None:
+        elevate_user(username)
+    else:
+        while True:
```
stefanv, line 54: Instead of an interactive loop, it may be a bit more consistent with our other tools to allow the following arguments:

```
no arguments: print help
--list: list users
--user=foo: use user foo
--role=X: add user to role X; set to System Admin by default
```

stefanv (commented):
Very nicely done @Theodlz; thanks!

tools/elevate_user.py
```diff
@@ -0,0 +1,62 @@
...
+except IndexError:
+    username = None
+
+cfg = load_config()
+init_db(**cfg['database'])
+
+
+def list_users():
+    users = (
+        User.query.options(joinedload(User.roles))
```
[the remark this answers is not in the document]
Theodlz: Noted. Should it be something like `DBSession.query().select_from(User)`, or `...select(User)...` ? I see both on SQLAchemy website and baselayer.

tools/elevate_user.py
Theodlz: @stefanv if the --user or --list are required, that means i remove the list of users that display right ? Or should that be what is displayed when you dont give any argument, or when you use --help. And then you can use the tool (non interactively) by passing the right arguments ?

stefanv:
Very nice, Theo! The help is probably overwritten when you import load_env from baselayer. Maybe try doing the argparsing before that import?

Theodlz:
> Very nice, Theo! The help is probably overwritten when you import load_env from baselayer. Maybe try doing the argparsing before that import? 

Great ! Thanks for the feedback, I'll try that then ! Have a nice weekend.

doc/setup.md
```diff
@@ -148,13 +148,19 @@ setting `debug_login` to `False`.
 ### Creating an administrative user
 
 By default, no user has permission to perform system administration.
-You can add a user with such permissions by running the following from
-Python:
+You can give such permissions to an existing user by running the following from your terminal in the `skyportal` directory:
+```
+PYTHONPATH=. python tools/elevate_user.py --username="<username>" --role="<role>"
```
stefanv, line 153: Can we change the name of the script to something a bit more descriptive, like `set_user_role`?  I think `elevate` always means up!

doc/setup.md
```diff
@@ -148,13 +148,19 @@ setting `debug_login` to `False`.
...
 
+The `username` argument is the name of the user to elevate, and the `role` argument is the role to give them. The role is optionnal, and defaults to `Super admin`.
+
+Also, you can omit those arguments and use:
 ```
-from skyportal.model_util import make_super_user
-make_super_user('your@email.address.org')
+PYTHONPATH=. python tools/elevate_user.py --list
 ```
+to list all users, their roles and the roles that they can be elevated to.
```
stefanv, line 162:
```suggestion
to list all users, their roles and the roles that they can be set to.
```

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+from baselayer.app.models import DBSession
+import sqlalchemy as sa
+
+env, cfg = load_env()
+init_db(**cfg['database'])
+
+parser = argparse.ArgumentParser(
+    description='Elevate user to super admin', add_help=False
+)
+parser.add_argument('--username', help='Username of the user to elevate')
```
stefanv, line 15:
```suggestion
parser.add_argument('--username', help='User to set role for')
```

stefanv: This is not very important, but thinking about it, perhaps we want:

```
./set_user_role.py user role  -> set user role
./set_user_role.py -> list all users
```

But, not a big deal.

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def roles_user_does_not_have(user):
+    return [role for role in role_acls if role not in roles_user_has(user)]
+
+
+def list_users():
+    users = get_users()
+    if len(users) == 0:
+        print('\nNo users left to elevate!')
```
stefanv, line 43:
```suggestion
        print('\nNo users in database')
```

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+    else:
+        # print each user's username and roles
+        print(f'\n{BOLD}List of users with their current roles:{END}')
```
stefanv, line 46:
```suggestion
        print(f'\n{BOLD}List of users and current roles:{END}')
```

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+                print(
+                    f'\n{BOLD}{i+1}. {YELLOW}{user.username}{END} has the following roles: {list_of_roles} and can be elevated to: {list_of_new_roles}'
+                )
+        print('\n')
+
+
+def elevate_user(username=None, role=None):
+    if not role:
+        print(
+            f'{BOLD}{RED}\nNo role given!{END} User will be elevated to {BOLD}{GREEN}Super admin{END}{BOLD} by default{END}.'
```
stefanv, line 71:
```suggestion
            f'{BOLD}{RED}\nNo role provided;{END} setting to {BOLD}{GREEN}Super admin{END}{BOLD}{END}.'
```

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+            if role not in [
+                role.id
+                for role in [
+                    user.roles[0] for user in users if user.username == username
+                ]
+            ]:
+                setup_permissions()
+                add_user(username, roles=[role], auth=True)
+                print(
+                    f'\n{BOLD}{YELLOW}User {username}{END} successfully elevated to role: {BOLD}{GREEN}{role}{END}!\n'
```
stefanv, line 92:
```suggestion
                    f'\nSuccessfully assigned role {BOLD}{GREEN}{role}{END} to {BOLD}{YELLOW}User {username}{END}\n'
```

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+                )
+            else:
+                print(
+                    f'\nUser {BOLD}{YELLOW}{username}{END} {BOLD}{GREEN}already has this role{END}\n'
```
stefanv, line 96:
```suggestion
                    f'\nUser {BOLD}{YELLOW}{username}{END} already has role {BOLD}{GREEN}{role}{END}\n'
```

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+                )
+
+        else:
+            print(f'\n{BOLD}{RED}User does not exist{END}!\n')
```
stefanv, line 100:
```suggestion
            print(f'\n{BOLD}{RED}User{END} {BOLD}{YELLOW}{user}{END} {BOLD}{RED}does not exist{END}\n')
```

tools/elevate_user.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def main():
+    if args.list:
+        list_users()
+    elif args.username:
+        elevate_user(args.username, args.role)
+    else:
+        print(
+            f'\n{BOLD}{RED}No arguments given{END}! Displaying {BOLD}{GREEN}Help{END}:\n'
```
stefanv, line 110:
```suggestion
            f'\n{BOLD}{RED}No arguments given;{END} displaying {BOLD}{GREEN}usage{END}:\n'
```

tools/elevate_user.py
Theodlz: @stefanv maybe we keep the `--list` argument, but if none is given we default to also list the users, rather than show the help ?

Theodlz: I already added it. If you meant something else, let me know and i'll revert that change.

stefanv:
I just realized this PR should probably be on baselayer!

doc/setup.md
```diff
@@ -148,13 +148,19 @@ setting `debug_login` to `False`.
 ### Creating an administrative user
 
 By default, no user has permission to perform system administration.
-You can add a user with such permissions by running the following from
-Python:
+You can give such permissions as well as other permissions to an existing user by running the following from your terminal in the `skyportal` directory:
```
stefanv, line 151:
```suggestion
You can give this and other permissions to an existing user by running the following from your terminal in the `skyportal` directory:
```

doc/setup.md
```diff
@@ -148,13 +148,19 @@ setting `debug_login` to `False`.
...
+```
+PYTHONPATH=. python tools/set_user_role.py --username="<username>" --role="<role>"
+```
 
+The `username` argument is the name of the user that we want to set to a role, and the `role` argument is the role to give them. The role is optionnal, and defaults to `Super admin`.
```
stefanv, line 156:
```suggestion
The role is optional, and defaults to `Super admin`.
```

doc/setup.md
```diff
@@ -148,13 +148,19 @@ setting `debug_login` to `False`.
...
+
+Also, you can omit those arguments and use:
```
stefanv, line 158:
```suggestion
You may also list all users and their roles:
```

doc/setup.md
```diff
@@ -148,13 +148,19 @@ setting `debug_login` to `False`.
...
 ```
-from skyportal.model_util import make_super_user
-make_super_user('your@email.address.org')
+PYTHONPATH=. python tools/set_user_role.py --list
 ```
+to list all users, their roles and the roles that they can be set to. If no arguments are passed, it will also default to listing all users.
```
stefanv, line 162:
```suggestion
```

tools/set_user_role.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def main():
+    if args.list or args.username:
+        if args.list:
+            list_users()
+        if args.username:
+            set_user_role(args.username, args.role)
+    else:
+        print(
```
stefanv, line 110: I thought about this more, and I like your original version: just print the help string here and don't try and connect to the db.

stefanv (commented):
I think this is basically ready. Just some minor issues now.

Theodlz:
> I just realized this PR should probably be on baselayer! 

Hi @stefanv ! Thanks for your last review, I'll make sure to add all the changes you asked for. 

About this, the issue here is that I used methods imported from Skyportal, but i'm sure that i could get rid of them and do it using something I make myself and/or coming from baselayer. Should I try to do that first and if I succeed we move this PR to baselayer rather than merging it with SkyPortal?

doc/setup.md
Theodlz: You want me to replace all of it by just the explanations about the role ? (i don't explain the username argument anymore ?). I made a modification that made more sense to me while you confirm this or not.

doc/setup.md
```diff
@@ -148,14 +148,19 @@ setting `debug_login` to `False`.
...
 
 By default, no user has permission to perform system administration.
-You can add a user with such permissions by running the following from
-Python:
+You can give this and other permissions to an existing user by running the following from your terminal in the `skyportal` directory:
+```
+PYTHONPATH=. python tools/set_user_role.py --username="<username>" --role="<role>"
+```
 
+The `username` argument is the name of the user that we want to set to a role. The `role` argument is optional, and defaults to `Super admin`.
```
stefanv, line 156: Just because the `--user` argument is obvious.

```suggestion
The `role` argument is optional, and defaults to `Super admin`.
```

tools/set_user_role.py
```diff
@@ -0,0 +1,123 @@
+import argparse
+import sys
+
+parser = argparse.ArgumentParser(
+    description='Elevate user to super admin', add_help=True
+)
+parser.add_argument('--username', help='User to set role for')
+parser.add_argument('--list', action='store_true', help='List all users')
+parser.add_argument('--role', help='Role to elevate user to')
```
stefanv, line 9:
```suggestion
parser.add_argument('--role', help='Role to elevate user to')
parser.add_argument('--list', action='store_true', help='List all users')
```

Just swap these for consistency with typical usage.

stefanv (approved):

stefanv:
@Theodlz Thanks very much for this! I'm going to merge, and then would you be willing to move this to baselayer in a new PR?

Theodlz:
Thanks for all the reviews @stefanv ! Sounds great, i'll work on that tomorrow.

PR #2865, opened 2022-04-21
Context Manager Session

guynir42:
There are a couple of places in the skyportal code (i.e., not in baselayer) where we generate sessions. 
Right now they use try/catch/finally to make sure the session gets closed properly. 
I'm not sure if that is good enough so I'm going to replace the finally with a `with Session() as session` block. 
This may help prevent us from slowly losing DB connections from the pool.

### DIFF
--- a/skyportal/facility_apis/ps1.py
+++ b/skyportal/facility_apis/ps1.py
@@ -169,7 +169,10 @@ class PS1API(FollowUpAPI):
             }
 
             url = f"{PS1_URL}/api/v0.1/panstarrs/dr2/detections.csv"
-            r = requests.get(url, params=params)
+            try:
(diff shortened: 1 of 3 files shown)


profjsb (commented):
LGTM but I think you can add a session.rollback() in the exception to be explicit (cf. https://docs.sqlalchemy.org/en/14/orm/session_basics.html#rolling-back). @mcoughlin and @stefanv should weigh in

guynir42:
@profjsb I just found another place where this could come in handy, in the PS1 handler. 
There's also some questionable code in the ZTF handler but I believe we are not using that yet...

profjsb:
LGTM -- @stefanv ? @mcoughlin ?

stefanv (approved):
I think this is a good pattern generally. Thanks!

PR #2868, opened 2022-04-21
Photometry request fix

profjsb:
wrap API in a try except


guynir42 (approved):
Not sure if it will solve the problem but it certainly can't hurt!

guynir42 (approved):

PR #2870, opened 2022-04-22
On top spectra

### LINKED ISSUE #1844: Feature request "in front' spectra

jadalilleboe:
Allows the user to choose which spectra should be displayed on top, closes #1844 
ezgif com-gif-maker (9)
@guynir42 you were right, there was a much easier way to do this within bokeh!


guynir42 (commented):
This looks a whole lot better, thanks! One thing I did notice is that the grid lines get in front of the spectrum. You should probably find their component and push it back to the background explicitly.

jadalilleboe:
@guynir42 Yeah you're right. done

jadalilleboe:
Screen Shot 2022-04-22 at 2 26 23 PM
grid does not go over spectra now

guynir42 (approved):
Looks great. Thanks!

PR #2872, opened 2022-04-23
Add ability to delete observing runs

mcoughlin:
This PR adds ability to delete observing runs.

### DIFF
--- a/skyportal/handlers/api/observingrun.py
+++ b/skyportal/handlers/api/observingrun.py
@@ -16,7 +16,7 @@ from ...models.schema import ObservingRunPost, ObservingRunGetWithAssignments
 
 
 class ObservingRunHandler(BaseHandler):
-    @permissions(["Upload data"])
+    @permissions(["Manage observing runs"])
     def post(self):
         """
         ---
@@ -171,7 +171,7 @@ class ObservingRunHandler(BaseHandler):
         self.verify_and_commit()
         return self.success(data=runs_list)
 
-    @permissions(["Upload data"])
+    @permissions(["Manage observing runs"])
     def put(self, run_id):
         """
         ---
@@ -227,7 +227,7 @@ class ObservingRunHandler(BaseHandler):
         self.push_all(action="skyportal/FETCH_OBSERVING_RUNS")
         return self.success()
 
-    @permissions(["Upload data"])
+    @auth_or_token
     def delete(self, run_id):
         """
         ---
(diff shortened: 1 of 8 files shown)


skyportal/handlers/api/observingrun.py
```diff
@@ -227,7 +227,7 @@ def put(self, run_id):
         self.push_all(action="skyportal/FETCH_OBSERVING_RUNS")
         return self.success()
 
-    @permissions(["Upload data"])
+    @auth_or_token
```
guynir42, line 230: Why is this one not with a "permissions"?

skyportal/utils/offset.py
```diff
@@ -478,6 +478,14 @@ def _calculate_best_position_for_offset_stars(
         log("Uncaught error in ra, dec determination")
         return fallback
 
+    if np.isnan(med_ra) or np.isnan(med_dec):
+        log(
+            "Warning: the median of the positions"
+            " from the photometry data associated with this source "
+            " retured nan, using fallback"
+        )
+        return fallback
```
guynir42, line 487: This seems unrelated, but I guess it's good to have!

static/js/components/ObservingRunPage.jsx
```diff
@@ -115,14 +156,23 @@ const ObservingRunList = ({ observingRuns }) => {
 
 const ObservingRunPage = () => {
   const { observingRunList } = useSelector((state) => state.observingRuns);
```
guynir42, line 158: Are you sure you want to call the variable holding the observing runs `observingRunList` the same as the name of one of the components? maybe just use `observingRuns`?

guynir42 (approved):
Minor suggestions.

skyportal/handlers/api/observingrun.py
mcoughlin: The permissions are done in the handler (we also do this for others)

skyportal/utils/offset.py
mcoughlin: yeah just something.i noticed.

static/js/components/ObservingRunPage.jsx
mcoughlin: It is already used in other places so probably too late to switch now.

skyportal/handlers/api/observingrun.py
guynir42: OK just checking.

static/js/components/ObservingRunPage.jsx
guynir42: Seems very confusing to me, but fine.

PR #2874, opened 2022-04-25
Refactor plot.py

### LINKED ISSUE #2867: Make the photometry/spectroscopy height responsive
Right now when you want to use all the new buttons below one of these plots on the source page (photometry or spectroscopy) you have to scroll down inside the accordion containing the plot. 

I think it would be much nicer to have the height of the accordion fit all the buttons so at least you wouldn't need to scroll inside it. 

It would make the accordions bigger but since they are foldable that shouldn't make much of a difference to the users.

jadalilleboe:
To aid the spring cleaning effort I've refactored `skyportal/plot.py`. Previously there was some duplicated code with creating the photometry panels, I've extracted a lot of that out to some helper functions. Also closes #2867

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -957,6 +957,8 @@ def make_legend_items_and_detection_lines(
                     ),
                     renderers,
                 )
+        else:
+            raise ValueError("Panel name should be one of mag, flux, and period.")
         if panel_name == 'mag' or panel_name == 'flux':
             mark_detections(
                 plot,
@@ -1220,6 +1222,8 @@ def add_widgets(
             layout,
         )
         layout.children.insert(2, period_controls)
+    else:
+        raise ValueError("Panel name should be one of mag, flux, and period.")
 
 
 def make_photometry_panel(panel_name, device, width, user, data, obj_id, spectra):
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -481,6 +481,11 @@ def add_plot_legend(plot, legend_items, width, legend_orientation, legend_loc):
 
 
 def make_clear_photometry_button(model_dict):
+    """Make a button to clear photometry.
+    Returns
+    -------
+    bokeh Button object
+    """
     button = Button(name="Clear Photometry", label="Clear Photometry", width=112)
     callback_clear_photometry = CustomJS(
         args={'model_dict': model_dict},
@@ -495,6 +500,11 @@ def make_clear_photometry_button(model_dict):
 
 
 def make_add_all_photometry_button(model_dict):
+    """Make a button to add photometry.
+    Returns
+    -------
+    bokeh Button object
+    """
     button = Button(name="Add All Photometry", label="Add All Photometry", width=120)
     callback_add_photometry = CustomJS(
         args={'model_dict': model_dict},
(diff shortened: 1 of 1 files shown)


skyportal/plot.py
```diff
@@ -547,57 +567,684 @@ def make_custom_buttons_div(panel_name):
...
+    panel_name : str
+        Name of the panel. One of 'flux', 'mag' or 'period'
+    color_dict : dict
+        Dictionary defining which color the data points should be
+    y_range : tuple(int)
+        Tuple containing the minimum and maximum y range values of the plot respectively
+    data : pandas DataFrame object
+        Photometry data (ungrouped)
+    obsind : pandas Series object
+        ???
```
jadalilleboe, line 811: If someone could help with the documentation here, that would be great. [code here](https://github.com/skyportal/skyportal/blob/master/skyportal/plot.py#L639)

mcoughlin: @jadalilleboe So I see
    obsind = data['hasflux'] & (
        data['flux'].fillna(0.0) / data['fluxerr'] >= PHOT_DETECTION_THRESHOLD
    )
So I would describe this as
   Index of data that exceed the required signal to noise ratio threshold
   
 What do you think @guynir42 ?

guynir42: Maybe "index of points that have been positively detected (observed), i.e., points that are not upper limits."

skyportal/plot.py
```diff
@@ -547,57 +567,684 @@ def make_custom_buttons_div(panel_name):
...
+                period_double_button,
+                period_halve_button,
+                width=width,
+                sizing_mode="scale_width",
+            ),
+            phase_selection,
+            period_selection,
+            width=width,
+        )
+    else:
```
guynir42, line 1145: I would make all these cases where it needs to be one of "flux", "mag" or "period" explicit `elif` for all three options, and add an `else` with a `ValueError`. 

Even though these are internal functions and it is very unlikely that the wrong key would ever be used, it makes it clear that the last `elif` is for "period" and not something generic.

guynir42 (approved):
Looks great! Minor suggestions below.

skyportal/plot.py
jadalilleboe: Good point, updated that.

skyportal/plot.py
```diff
@@ -808,7 +808,7 @@ def make_legend_items_and_detection_lines(
     data : pandas DataFrame object
         Photometry data (ungrouped)
     obsind : pandas Series object
-        ???
+        Index of points that have been positively detected (observed), i.e., points that are not upper limits.
```
guynir42, line 811:
```suggestion
        Index of points that have been positively detected (observed), 
        i.e., points that are not upper limits.
```

skyportal/plot.py
```diff
@@ -1243,7 +1246,7 @@ def make_photometry_panel(panel_name, device, width, user, data, obj_id, spectra
         The source/object's spectra
     Returns
     -------
-    bokeh Panel object
+    bokeh Panel object or None if the panel should not be added to the plot (i.e. no period plot)
```
guynir42, line 1249:
```suggestion
    bokeh Panel object or None if the panel should 
    not be added to the plot (i.e. no period plot)
```

guynir42 (approved):

guynir42:
I can't push to this repo. @jadalilleboe try to pull down the changes, then make any change to the file and commit. The pre-commit hooks should fix the white space.

guynir42:
@profjsb the tests are now :heavy_check_mark:

profjsb (approved):
LGTM (I did not review the code but did try locally).

PR #2878, opened 2022-04-26
short circuit the FollowupRequestSchedulerHandler

guynir42:
Attempt to fix the many queries to followup requests.


profjsb (approved):
Let's give it a shot!

PR #2879, opened 2022-04-26
remove candidate counting from cron in config

guynir42:


stefanv:
I think what you want to do is to set `cron` to an empty list in the fritz config instead.

guynir42:
> I think what you want to do is to set `cron` to an empty list in the fritz config instead.

I couldn't find it in the fritz/config.yaml or anywhere besides the skyportal config.

guynir42:
Actually it does exist in `fritz/config.yaml.default` but I commented it out last time. I should bring it back to be an empty list. 

But do we want to keep this is the normal skyportal?

stefanv:
Sure, why not?

stefanv:
Perhaps once a day is enough, though.

config.yaml.defaults
```diff
@@ -443,8 +443,8 @@ external_logging:
     excluded_log_files: ["log/websocket_server.log"]
 
 cron:
-  - interval: 60
-    script: jobs/count_unsaved_candidates.py
+  # - interval: 60
```
mcoughlin, line 446: This makes sense.

mcoughlin (approved):
Thanks @guynir42.

mcoughlin (approved):

profjsb:
> Perhaps once a day is enough, though.

The query in this cron job brings the (large) DB down essentially. We shouldn't be enabling it, even once a day, until that query is fixed.

profjsb (approved):

PR #2880, opened 2022-04-27
Pagination of followup requests

mcoughlin:
This PR attempts to add pagination of follow-up requests (inspired by the style in the Users table).

### DIFF
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -383,8 +383,8 @@ class FollowupRequestHandler(BaseHandler):
             return self.error("Invalid page number value.")
         try:
             n_per_page = int(n_per_page)
-        except ValueError, TypeError:
-            return self.error("Invalid numPerPage value.")
+        except (ValueError, TypeError) as e:
+            return self.error(f"Invalid numPerPage value: {str(e)}")
 
         if n_per_page > MAX_FOLLOWUP_REQUESTS:
             return self.error(
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -350,7 +350,7 @@ class FollowupRequestHandler(BaseHandler):
             schema:
               type: integer
             description: |
-              Number of followup requests to return per paginated request. Defaults to 10
+              Number of followup requests to return per paginated request. Defaults to 100
           - in: query
             name: pageNumber
             nullable: true
@@ -373,7 +373,7 @@ class FollowupRequestHandler(BaseHandler):
         sourceID = self.get_query_argument('sourceID', None)
         status = self.get_query_argument('status', None)
         page_number = self.get_query_argument("pageNumber", 1)
-        n_per_page = self.get_query_argument("numPerPage", 10)
+        n_per_page = self.get_query_argument("numPerPage", 100)
 
         try:
             page_number = int(page_number)
@@ -440,6 +440,11 @@ class FollowupRequestHandler(BaseHandler):
         )
 
         total_matches = followup_requests.count()
+        if total_matches > 1000:
+            return self.error(
+                f'More than 1000 followup requests ({total_matches}) match your request... please make your query more restrictive and try again'
+            )
+
         if n_per_page is not None:
             followup_requests = followup_requests.limit(n_per_page).offset(
                 (page_number - 1) * n_per_page
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -383,7 +383,7 @@ class FollowupRequestHandler(BaseHandler):
             return self.error("Invalid page number value.")
         try:
             n_per_page = int(n_per_page)
-        except TypeError:
+        except ValueError, TypeError:
             return self.error("Invalid numPerPage value.")
 
         if n_per_page > MAX_FOLLOWUP_REQUESTS:
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -344,6 +344,19 @@ class FollowupRequestHandler(BaseHandler):
               type: string
             description: |
               String to match status of request against
+          - in: query
+            name: numPerPage
+            nullable: true
+            schema:
+              type: integer
+            description: |
+              Number of followup requests to return per paginated request. Defaults to 100
+          - in: query
+            name: pageNumber
+            nullable: true
+            schema:
+              type: integer
+            description: Page number for paginated query results. Defaults to 1
           responses:
             200:
               content:
@@ -359,6 +372,18 @@ class FollowupRequestHandler(BaseHandler):
         end_date = self.get_query_argument('endDate', None)
         sourceID = self.get_query_argument('sourceID', None)
         status = self.get_query_argument('status', None)
+        page_number = self.get_query_argument("pageNumber", None) or 100
+        n_per_page = self.get_query_argument("numPerPage", None)
+
+        try:
+            page_number = int(page_number)
+        except ValueError:
+            return self.error("Invalid page number value.")
+        try:
+            if n_per_page is not None:
+                n_per_page = int(n_per_page)
+        except ValueError:
+            return self.error("Invalid numPerPage value.")
 
         # get owned assignments
         followup_requests = FollowupRequest.query_records_accessible_by(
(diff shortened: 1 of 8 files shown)


skyportal/handlers/api/followup_request.py
```diff
@@ -412,9 +437,20 @@ def get(self, followup_request_id=None):
             joinedload(FollowupRequest.allocation).joinedload(Allocation.group),
             joinedload(FollowupRequest.obj),
             joinedload(FollowupRequest.requester),
-        ).all()
+        )
+
+        total_matches = followup_requests.count()
```
guynir42, line 442: Should there be a hard maximum number of requests? Like for candidates?

static/js/components/FollowupRequestLists.jsx
```diff
@@ -273,8 +278,17 @@ const FollowupRequestLists = ({
     selectableRows: "none",
     enableNestedDataAccess: ".",
     elevation: 0,
-    rowsPerPageOptions: [1, 10, 15],
+    page: pageNumber - 1,
+    rowPerPage: numPerPage,
```
guynir42, line 282:
```suggestion
    rowsPerPage: numPerPage,
```

static/js/components/FollowupRequestLists.jsx
```diff
@@ -84,7 +84,12 @@ const FollowupRequestLists = ({
   followupRequests,
   instrumentList,
   instrumentFormParams,
+  totalMatches,
+  handleTableChange = false,
+  pageNumber = 1,
+  numPerPage = 10,
   showObject = false,
+  serverSide = false,
```
guynir42, line 92: What does `serverSide` do?

static/js/ducks/followup_requests.js
```diff
@@ -10,9 +10,15 @@ const REFRESH_FOLLOWUP_REQUESTS = "skyportal/REFRESH_FOLLOWUP_REQUESTS";
...
 
-// eslint-disable-next-line import/prefer-default-export
-export const fetchFollowupRequests = (params = {}) =>
-  API.GET("/api/followup_request", FETCH_FOLLOWUP_REQUESTS, params);
+export function fetchFollowupRequests(params = {}) {
+  if (!Object.keys(params).includes("pageNumber")) {
+    params.pageNumber = 1;
+  }
+  if (!Object.keys(params).includes("numPerPage")) {
+    params.numPerPage = 10;
```
guynir42, line 18: Unless you are using different defaults for the front end, I don't see any value in putting the page length and page number as defaults here as well. 

That said, for the number of rows per page, it may be good to have a small default (like 10) for displaying on the front end, and a bigger default for API users.

guynir42 (commented):
I wonder if this is really the cause of our problems. It seems to me that no one should be routinely getting all followup requests, but instead they should always be called in the context of one object or one instrument. And even with thousands of requests, it should not be such a long query. 
But even so, it doesn't hurt to page this.

skyportal/handlers/api/followup_request.py
```diff
@@ -344,6 +344,19 @@ def get(self, followup_request_id=None):
               type: string
             description: |
               String to match status of request against
+          - in: query
+            name: numPerPage
+            nullable: true
+            schema:
+              type: integer
+            description: |
+              Number of followup requests to return per paginated request. Defaults to 10
```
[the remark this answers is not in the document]
mcoughlin: So say 50?

skyportal/handlers/api/followup_request.py
mcoughlin: But just counting is ok?

guynir42: I mean that if the user tries to input 10,000 as the number per page (or set `None`) it should not work.

skyportal/handlers/api/followup_request.py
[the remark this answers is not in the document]
guynir42: I would say 100 just because I seem to recall that was the limit for candidates (and it cannot be too heavy, these are just followup requests). 
You may want a smaller default number for the front end elements.

static/js/components/FollowupRequestLists.jsx
mcoughlin: This is checking whether pagination occurs in MUI or in the handler.

static/js/ducks/followup_requests.js
mcoughlin: Ok so keep this but make the aPI default larger?

static/js/components/FollowupRequestLists.jsx
guynir42: ok, got it.

static/js/ducks/followup_requests.js
guynir42: I would get rid of the `pageNumber=1`. That's always going to be the default and there's no point having that defined in two places. 
For `numPerPage` you can leave it, with a value that makes sense for the front end page (I guess 10 is fine).

mcoughlin: good thinking

skyportal/handlers/api/followup_request.py
mcoughlin: Set it to 1000 for now.

skyportal/handlers/api/followup_request.py
```diff
@@ -440,6 +440,11 @@ def get(self, followup_request_id=None):
         )
 
         total_matches = followup_requests.count()
+        if total_matches > 1000:
+            return self.error(
+                f'More than 1000 followup requests ({total_matches}) match your request... please make your query more restrictive and try again'
+            )
+
```
guynir42, line 447: What I meant is that the number per page can never exceed some high number (e.g., 1000). If a user inputs None or >1000 the API returns an error. So you could always send a large query but you won't be able to ask for very long pages.

mcoughlin: @guynir42 ah yeah. makes sense.

stefanv:
While doing this, maybe think about how we can make paging generic between the different endpoints.

skyportal/handlers/api/followup_request.py
```diff
@@ -385,6 +385,9 @@ def get(self, followup_request_id=None):
         except ValueError:
             return self.error("Invalid numPerPage value.")
 
+        if n_per_page > 1000:
+            return self.error('numPerPage should be no larger than 1000.')
+
```
guynir42, line 390: I'd also change the parsing of this variable to catch `TypeError` instead of checking if it is `None` (i.e., when input is `None` also return "Invalid page number value").

skyportal/handlers/api/followup_request.py
```diff
@@ -385,6 +385,9 @@ def get(self, followup_request_id=None):
         except ValueError:
             return self.error("Invalid numPerPage value.")
 
+        if n_per_page > 1000:
```
guynir42, line 388: And I would place this 1000 number as a global defined at top of file.

skyportal/handlers/api/followup_request.py
```diff
@@ -380,13 +382,14 @@ def get(self, followup_request_id=None):
         except ValueError:
             return self.error("Invalid page number value.")
         try:
-            if n_per_page is not None:
-                n_per_page = int(n_per_page)
-        except ValueError:
+            n_per_page = int(n_per_page)
+        except TypeError:
```
guynir42, line 386:
```suggestion
        except ValueError, TypeError:
```

guynir42 (approved):
Minor change suggested.

guynir42 (approved):

Theodlz:
Looks good ! Just one thing i noticed. When going from a gcn event to a source, the source page crashes. In the console, the error is related to the FollowUpRequest Form:

`Uncaught TypeError: Cannot read properties of undefined (reading 'formSchema')`

Theodlz:
> @Theodlz can you try again?

I get this error now:

```
Uncaught TypeError:
    instrumentFormParams[allocationLookUp[selectedAllocationId].instrument_id] is undefined
```
coming from : `FollowupRequestForm FollowupRequestForm.jsx:190`

mcoughlin:
@Theodlz Please try again.

Theodlz:
> @Theodlz Please try again.

Amazing ! It works super well now. This kind off verification should be done on more pages of the app i think. Verifying the type of the props is one thing, but verifying that the data is the one we expect is something else that needs to be done.

PR #2887, opened 2022-04-29
Some followup request handler improvements

mcoughlin:
This PR makes some of the followup request handler improvements discussed on the call today.

### DIFF
--- a/skyportal/facility_apis/sedmv2.py
+++ b/skyportal/facility_apis/sedmv2.py
@@ -123,10 +123,6 @@ class SEDMV2API(FollowUpAPI):
             'skyportal/REFRESH_SOURCE',
             payload={'obj_key': request.obj.internal_key},
         )
-        flow.push(
-            '*',
-            "skyportal/REFRESH_FOLLOWUP_REQUESTS",
-        )
 
     @staticmethod
     def delete(request):
@@ -191,10 +187,6 @@ class SEDMV2API(FollowUpAPI):
             'skyportal/REFRESH_SOURCE',
             payload={'obj_key': request.obj.internal_key},
         )
-        flow.push(
-            '*',
-            "skyportal/REFRESH_FOLLOWUP_REQUESTS",
-        )
 
     @staticmethod
     def update(request):
@@ -256,10 +248,6 @@ class SEDMV2API(FollowUpAPI):
             '*',
             'skyportal/REFRESH_SOURCE',
             payload={'obj_key': request.obj.internal_key},
-        )
-        flow.push(
-            '*',
-            "skyportal/REFRESH_FOLLOWUP_REQUESTS",
         )
 
     form_json_schema = {
--- a/skyportal/facility_apis/sedmv2.py
+++ b/skyportal/facility_apis/sedmv2.py
@@ -120,8 +120,12 @@ class SEDMV2API(FollowUpAPI):
         flow = Flow()
         flow.push(
             '*',
+            'skyportal/REFRESH_SOURCE',
+            payload={'obj_key': request.obj.internal_key},
+        )
+        flow.push(
+            '*',
             "skyportal/REFRESH_FOLLOWUP_REQUESTS",
-            payload={"obj_key": request.obj.internal_key},
         )
 
     @staticmethod
@@ -184,8 +188,12 @@ class SEDMV2API(FollowUpAPI):
         flow = Flow()
         flow.push(
             '*',
+            'skyportal/REFRESH_SOURCE',
+            payload={'obj_key': request.obj.internal_key},
+        )
+        flow.push(
+            '*',
             "skyportal/REFRESH_FOLLOWUP_REQUESTS",
-            payload={"obj_key": request.obj.internal_key},
         )
 
     @staticmethod
@@ -246,8 +254,12 @@ class SEDMV2API(FollowUpAPI):
         flow = Flow()
         flow.push(
             '*',
+            'skyportal/REFRESH_SOURCE',
+            payload={'obj_key': request.obj.internal_key},
+        )
+        flow.push(
+            '*',
             "skyportal/REFRESH_FOLLOWUP_REQUESTS",
-            payload={"obj_key": request.obj.internal_key},
         )
 
     form_json_schema = {
--- a/skyportal/facility_apis/sedmv2.py
+++ b/skyportal/facility_apis/sedmv2.py
@@ -121,6 +121,7 @@ class SEDMV2API(FollowUpAPI):
         flow.push(
             '*',
             "skyportal/REFRESH_FOLLOWUP_REQUESTS",
+            payload={"obj_key": request.obj.internal_key},
         )
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


static/js/components/FollowupRequestPage.jsx
```diff
@@ -35,23 +39,55 @@ const FollowupRequestPage = () => {
...
+      // the new default form fields, so that the allocations list can
+      // update
+
+      const result = await dispatch(instrumentActions.fetchInstruments());
+
+      const { data } = result;
+      setSelectedInstrumentId(data[0]?.id);
+    };
+
+    getInstruments();
```
stefanv, line 60: This is async, so doesn't it require an `await`?

static/js/ducks/followup_requests.js
```diff
@@ -25,9 +27,18 @@ export const prioritizeFollowupRequests = (params = {}) =>
...
   );
 
-messageHandler.add((actionType, payload, dispatch) => {
+messageHandler.add((actionType, payload, dispatch, getState) => {
+  const { source } = getState();
+
   if (actionType === REFRESH_FOLLOWUP_REQUESTS) {
-    dispatch(fetchFollowupRequests());
+    const loaded_obj_key = source?.internal_key;
+    if ("obj_key" in payload) {
```
stefanv, line 35: Probably unnecessary; could just use `payload.obj_key?` below?

static/js/ducks/followup_requests.js
```diff
@@ -25,9 +27,18 @@ export const prioritizeFollowupRequests = (params = {}) =>
...
+      if (loaded_obj_key === payload.obj_key) {
+        dispatch(sourceAction.fetchSource(source.id));
```
stefanv, line 37: If we're going to fetch the source, why not change this to `REFRESH_SOURCE` as used elsewhere?

static/js/ducks/followup_requests.js
```diff
@@ -25,9 +27,18 @@ export const prioritizeFollowupRequests = (params = {}) =>
...
+      }
+    } else {
+      dispatch(fetchFollowupRequests());
```
stefanv, line 40: This doesn't seem to fit with the rest of the pattern here.

stefanv (commented):
This looks great!

static/js/ducks/followup_requests.js
mcoughlin: Yeah that seems better.

static/js/ducks/followup_requests.js
mcoughlin: Yeah reverting to the old way.

static/js/components/FollowupRequestPage.jsx
mcoughlin: I am not sure, the await is in the function?

stefanv: I think it's fine in this case; you just want to make sure the dispatch eventually happens, and it will once the getInstruments function has been called.

`async` functions returns promises, so if you need to *use* the result of an `async` function you need to await it first; but you don't, so no problem.

skyportal/facility_apis/sedmv2.py
```diff
@@ -118,6 +118,11 @@ def submit(request):
...
 
         flow = Flow()
+        flow.push(
+            '*',
+            'skyportal/REFRESH_SOURCE',
+            payload={'obj_key': request.obj.internal_key},
+        )
         flow.push(
             '*',
             "skyportal/REFRESH_FOLLOWUP_REQUESTS",
```
stefanv, line 128: These pushes trigger:

```
messageHandler.add((actionType, payload, dispatch) => {
  if (actionType === REFRESH_FOLLOWUP_REQUESTS) {
    dispatch(fetchFollowupRequests());
  }
});
```

Isn't that what we're trying to avoid?

mcoughlin: Ah yes. Lemme remove those.

stefanv:
How about the duck that handles the now-no-longer-used `REFRESH_FOLLOWUP_REQUESTS`?

stefanv (approved):

skyportal/handlers/api/followup_request.py
```diff
@@ -373,6 +379,7 @@ def get(self, followup_request_id=None):
         start_date = self.get_query_argument('startDate', None)
         end_date = self.get_query_argument('endDate', None)
         sourceID = self.get_query_argument('sourceID', None)
+        instrumentID = self.get_query_argument('instrumentID', None)
```
guynir42, line 382: How was this not here from the start??

skyportal/handlers/api/followup_request.py
```diff
@@ -433,6 +440,15 @@ def get(self, followup_request_id=None):
...
             )
+        if instrumentID:
+            allocation_query = Allocation.query_records_accessible_by(
+                self.current_user
+            ).filter(Allocation.instrument_id == instrumentID)
+            allocation_subquery = allocation_query.subquery()
+            followup_requests = followup_requests.join(
+                allocation_subquery,
+                FollowupRequest.allocation_id == allocation_subquery.c.id,
+            )
```
guynir42, line 451: This seems more complicated than just getting the instrument and all associated followup requests. Can you add a comment explaining why we need the allocations subquery?

static/js/components/FollowupRequestPage.jsx
```diff
@@ -60,17 +96,53 @@ const FollowupRequestPage = () => {
...
       <Grid item md={6} sm={12}>
         <Paper elevation={1}>
           <div className={classes.paperContent}>
             <Typography variant="h6">List of Followup Requests</Typography>
-            {queryInProgress ? (
-              <div>
-                <CircularProgress />
-              </div>
-            ) : (
+            <div>
```
guynir42, line 145: Why get rid of the spinner?

guynir42 (approved):
Minor comments.

skyportal/handlers/api/followup_request.py
mcoughlin: :)

skyportal/handlers/api/followup_request.py
mcoughlin: yeah..

static/js/components/FollowupRequestPage.jsx
mcoughlin: because we got rid of that query style.

skyportal/handlers/api/followup_request.py
```diff
@@ -441,6 +441,8 @@ def get(self, followup_request_id=None):
                 obj_subquery, FollowupRequest.obj_id == obj_subquery.c.id
             )
         if instrumentID:
+            # allocation query required as only way to reach
+            # instrument_id
```
guynir42, line 445: I don't get it. Why not just query all followup requests with this `instrumentID`? Is the allocations subquery supposed to filter only followup requests that have an allocation visible to the user or something like that?

mcoughlin: followup_requests don't have an instrument_id, they have an allocation_id

guynir42: OK that's really surprising. Can you add that to the comment? I'm sure someone like me would run into it at some point and would take a while to figure this one out.

PR #2888, opened 2022-04-29
Remove ResizeObserver on source page

jadalilleboe:
Quick fix for the problem brought up in slack where the source page would freak out due to the ResizeObserver

This reverts back to having some hardcoded plot widths based on if the right pane is hidden or not. I would like to make the plot widths fully responsive but I would have to find a good solution, the ResizeObserver does not seem to work well

Once approved I'll also make a PR on fritz


stefanv (approved):

jadalilleboe:
Will merge with @stefanv 's approval, thanks!

guynir42 (approved):
LGTM

PR #2937, opened 2022-05-03
Telescope Map : Display if telescope in Night (Astronomical)

Theodlz:
### Backend:
For this, we used 2 of the methods defined in the Telescope model: 

- `next_twilight_evening_astronomical`
- `next_twilight_morning_astronomical`

Logically, if the `next_twilight_morning_astronomical` is before the `next_twilight_evening_astronomical`, then it means we are at night (astronomical), and vice versa.

So we added a boolean `is_night_astronomical` to a telescope, so we can use it in the frontend to change the color of the marker depending on that parameter.

### Frontend:

We choose yellow (like the color of the sun) for the color of markers of telescope that are not at night, and a deep purple (like the night sky) for the color of markers of telescope that are at night.

#### Details:
When it comes to nested telescopes (nested telescopes are telescopes that are so close to each other on the map that we grouped them into one marker to avoid overlapping), we show the color associated to that parameter being true if at least one of them has it set to `true`. So a user looking at the map knows that at that location (on this marker), there is at least one of them that is in at night (astronomical).

## Preview:
telescope_is_night

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -183,8 +183,8 @@ telescope:
     =id: Swift
   - diameter: 1.8
     elevation: 4215.0
-    lat: 19.825
-    lon: 204.5317
+    lat: 20.708424
+    lon: -156.257565
     name: PAN-STARRS 1.8m
     nickname: PS1
     skycam_link: null
--- a/skyportal/handlers/api/telescope.py
+++ b/skyportal/handlers/api/telescope.py
@@ -105,9 +105,18 @@ class TelescopeHandler(BaseHandler):
         query = Telescope.query
         if tel_name is not None:
             query = query.filter(Telescope.name == tel_name)
+
         data = query.all()
+        telescopes = []
+        for telescope in data:
+            temp = telescope.to_dict()
+            temp['is_night_astronomical'] = bool(
+                telescope.next_twilight_morning_astronomical().jd
+                < telescope.next_twilight_evening_astronomical().jd
+            )
+            telescopes.append(temp)
         self.verify_and_commit()
-        return self.success(data=data)
+        return self.success(data=telescopes)
 
     @permissions(['Manage sources'])
     def put(self, telescope_id):
--- a/static/js/components/TelescopeInfo.jsx
+++ b/static/js/components/TelescopeInfo.jsx
@@ -20,6 +20,13 @@ const useStyles = makeStyles(() => ({
     justifyItems: "left",
     alignItems: "left",
   },
+  telescope_header: {
+    display: "flex",
+    flexDirection: "row",
+    justifyItems: "center",
+    alignItems: "center",
+    gap: "0.5rem",
+  },
   h2: {
     textAlign: "left",
     fontSize: "1.4rem",
@@ -40,6 +47,18 @@ const useStyles = makeStyles(() => ({
     padding: "0",
     margin: "0",
   },
+  canObserve: {
+    height: "1rem",
+    width: "1rem",
+    backgroundColor: "#0c1445",
+    borderRadius: "50%",
+  },
+  cannotObserve: {
+    height: "1rem",
+    width: "1rem",
+    backgroundColor: "#f9d71c",
+    borderRadius: "50%",
+  },
 }));
 
 const TelescopeInfo = () => {
@@ -57,9 +76,16 @@ const TelescopeInfo = () => {
             className={classes.listItem}
             key={telescope.id}
           >
-            <h2 className={classes.h2}>
-              {telescope.name} ({telescope.nickname})
-            </h2>
+            <div className={classes.telescope_header}>
+              {telescope.is_night_astronomical ? (
+                <span className={classes.canObserve} />
+              ) : (
+                <span className={classes.cannotObserve} />
+              )}
+              <h2 className={classes.h2}>
+                {telescope.name} ({telescope.nickname})
+              </h2>
+            </div>
             {telescope.robotic ? (
               <h3 className={classes.h3}>Robotic : Yes</h3>
             ) : (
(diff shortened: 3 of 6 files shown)


mcoughlin:
@Theodlz this is cool. I also wonder if also adding information on next twilight / dusk in the text info might combine to give the information you need?

Theodlz:
Hi @mcoughlin , do you mean not just saying if its night or day but putting for example the UTC times for the bext twilight and dusk on the list where the telescope info shows up ?

mcoughlin:
@Theodlz Yeah or something like that. To give just a bit more info.

Theodlz:
Ok ! I'll add it tomorrow then. Have a nice day Michael.

guynir42 (commented):
The code looks good, and so does the new map page. 
When loading it myself I get this in the console: `Warning: Each child in a list should have a unique "key" prop.` 
Maybe in some instances the component gets an empty list? 

Also, I was wondering if SWIFT is no longer on the map because it doesn't have a fixed location? If so, should there be some way to load the data for such telescopes?

mcoughlin:
@guynir42 Yeah I asked @Theodlz to not show the satellites because they didn't make much sense on the world map. Maybe @Theodlz has an idea for a specific dropdown for these?

Theodlz:
Hi @guynir42 , @mcoughlin . Yes we didnt show them on the map as their location wouldnt make much sense. As of right now, i can think of anything else then maybe putting a marker (list for the rest of the telescopes) in a specific place on the map (like bottom/top right corner ?), with maybe a bigger size and/or different shape (square not circle) to differenciate it from the rest. So that a user can click that marker and see the list of telescopes with a non fixed location. What do you think ? Easier to implement then yet another drop down list, here we can re use the existing list.

guynir42:
@Theodlz yeah I think a square at the bottom of the map somewhere would be fine. 
Choose a coordinate that puts them somewhere in the ocean where there aren't any other telescopes.

Theodlz:
@guynir42 what do you think of it now ?

guynir42:
Looks a lot better. If you can get those non-fixed telescope to not show longitude/latitude that would be even better...

Theodlz:
> Looks a lot better. If you can get those non-fixed telescope to not show longitude/latitude that would be even better...

How does it work for this btw, is there a way to get their longitude and latitude if they are changing continuously ? I actually don't know how this works.

mcoughlin:
@Theodlz we don't store orbit info, so we won't be able to render anything unfortunately.

Theodlz:
> @Theodlz we don't store orbit info, so we won't be able to render anything unfortunately.

Noted. Would be something pretty cool to implement though. I hid the lon lat of non fixed telescopes as you asked @guynir42

guynir42:
> > Looks a lot better. If you can get those non-fixed telescope to not show longitude/latitude that would be even better...
> 
> How does it work for this btw, is there a way to get their longitude and latitude if they are changing continuously ? I actually don't know how this works.

In principle you could probably query some website like "heavens above" to get their current position, but that seems fairly complicated this...

guynir42:
@Theodlz Looks good. I'm thinking either getting rid of the "elevation" as well, or at least filling it in for SWIFT with the actual mean orbital height or something.

Theodlz:
> @Theodlz Looks good. I'm thinking either getting rid of the "elevation" as well, or at least filling it in for SWIFT with the actual mean orbital height or something.

Hi @guynir42 , thank your for your feedback. Maybe we should, simply hide it here, and add that feature as a seperate PR ? Otherwise, if we add orbital height, should it just be stored in the elevation ? If so, i can just add it in the demo data maybe ? But i'm not sure that just the orbital height is enough to be useful...

Theodlz:
> @Theodlz I think it is really ok to just take off the elevation for space telescopes. It is a cute feature to have that, but no really informative for the user.

Done !

guynir42:
I know I'm being super picky... but can you also remove the sunrise/sunset on the space telescopes, and make the color different maybe? In particular making it yellow when in daytime doesn't make much sense.

Theodlz:
> I know I'm being super picky... but can you also remove the sunrise/sunset on the space telescopes, and make the color different maybe? In particular making it yellow when in daytime doesn't make much sense.

Like this @guynir42 ?

telescopes_non_fixed

Theodlz:
As we noticed with @mcoughlin , some of the lon and lat in the demo data were wrong. One more thing i want to add before merging (important), is to avoid the page from crashing if it receives telescopes with wrong lon lat (out of the range it expects for the angles). So it can be a bit more robust.

Theodlz:
> @Theodlz I've already merged a PR with a fix to the wrong longitudes. No need to do anything else. Please merge master and check that everything works.

Amazing ! Just saw it. So should i remove the verification i added in the handler ?

guynir42:
@Theodlz I think you can remove your verification if it looks like it works with out it (after pulling in the latest). 
As for the demo data, I think we can leave it as it is, if the only "problem" is that some of the longitudes are bigger than 180. That's not really problematic, it only causes problems with the timezone finder.

Theodlz:
> @Theodlz I think you can remove your verification if it looks like it works with out it (after pulling in the latest). As for the demo data, I think we can leave it as it is, if the only "problem" is that some of the longitudes are bigger than 180. That's not really problematic, it only causes problems with the timezone finder.

Great @guynir42. I left the frontend verification for lon lat as a wrong longitude could potentially cause issues with displaying the map. But i removed the modifications i added to the backend and demo data.

Theodlz:
Ok @guynir42, i ended up removing verification i added yesterday from both frontend and backend. Seems that react-simple-maps handles these 'out of range' angles pretty well.

guynir42 (approved):
LGTM

mcoughlin (approved):
Thanks for all the iteration on this @Theodlz

Issue #2876, opened 2022-04-26
Issue with postgresql configuration : pg_hba.conf


Theodlz:
Currently, when adding those 2 lines:
```
host skyportal skyportal 127.0.0.1/32 trust
host skyportal_test skyportal 127.0.0.1/32 trust
```

when using postgres-12 or postgresql-13 (the version provided by ubuntu) or even postgres-14 (having to add the postgresql repo for this one), we can't run skyportal using just the make commands as they can't seem to be able to access the database.

Indeed, i tried creating a db called skyportal manually as the make command couldn't do it, and then the user skyportal can actually access the database, and we run into the same error but for skyportal_test database.

Then, we can't run make db_clear, as the user postgres doesn't seem to have access to the database either.

Here, you'll find a Dockerfile that builds a container that reproduces this error using postgresql-13 (but you can modify it to use postgresql-12 very easily) : [Dockerfile.txt](https://github.com/skyportal/skyportal/files/8561579/Dockerfile.txt)

I think that maybe the problem is the instructions in the documentation, and maybe not postgresql itself. One strange this though, is that when i installed postgresql-12 for the first time on my machine, it did work. But for other people like @thomasculino it did not work(he uses ubuntu 20.04, i use a 21.04). And now that i upgraded postgresql to version 14, it doesn't work anymore. I tried downgrading it but i have the same problem...

EDIT : i think all this makes sense, as in the config file we allow only the skyportal user to access postgres (more specifically skyportal's db) without a password on localhost, but not the 'postgres' user. But then, if i add : 

```
host skyportal postgres 127.0.0.1/32 trust
host skyportal_test postgres 127.0.0.1/32 trust
```

It still doesn't work, but adding: 

```
host skyportal postgres 127.0.0.1/32 trust
host skyportal_test postgres 127.0.0.1/32 trust
```

which basically gives access to all databases to the postgres user without a password. But obviously, for security reasons when deploying the app, giving access to a user to all databases without a password (and that also gives it the right to delete users and databases) is not something we want. 

What im thinking is that maybe it would be best to access postgres to create db and user (when running make db_clear or db_init) using sudo, so allowing only root user to do it. Could be something like: you run make db_clear > it asks for your password ?

# Update: This issue has been fixed by PR https://github.com/skyportal/skyportal/pull/2938

stefanv:
[...]

PR #2938, opened 2022-05-04
Wrong configuration in the Docs for pg_hba.conf

Theodlz:
This PR addresses the issue mentioned here : https://github.com/skyportal/skyportal/issues/2876

In which we talk about how the current lines given in the docs that a user needs to add to the config file of postgresql (pg_hba.conf) does not work because of the changes made to baselayer in this PR: https://github.com/cesium-ml/baselayer/pull/269


stefanv (approved):
I'll merge this for now.

If we still have problems, we may need to add an entry for `::1/128` as well (localhost, when connecting via ipv6).

stefanv:
Thanks, that'd be great!

PR #2940, opened 2022-05-05
Handle Out-of-Footprint cases for PS1

profjsb:
This PR catches the case when the source/obj is not in the PS1 footprint, both for postage stamps and for the finding charts. It also sets a timeout (10 seconds) so that the stamp/image retrieval will not sit in limbo (possibly causing DB idle-in-transactions).


skyportal/utils/offset.py
```diff
@@ -1026,6 +1026,9 @@ def fits_image(
             ra=center_ra, dec=center_dec, imsize=imsize
         )
 
+    if url in [None, ""]:
```
stefanv, line 1029: Should this error hard, or just add an entry in the log?

stefanv (approved):

skyportal/models/obj.py
```diff
@@ -481,15 +483,37 @@ def panstarrs_url(self):
         The cutout service doesn't allow directly querying for an image; the
         best we can do is request a page that contains a link to the image we
         want (in this case a combination of the g/r/i filters).
+
+        If this page does not return without PS1_CUTOUT_TIMEOUT seconds then
```
guynir42, line 487:
```suggestion
        If this page does not return within PS1_CUTOUT_TIMEOUT seconds then
```

skyportal/utils/offset.py
```diff
@@ -42,6 +42,8 @@
 
 _, cfg = load_env()
 
+PS1_CUTOUT_TIMEOUT = 10
```
guynir42, line 45:
```suggestion
from ..models.obj import PS1_CUTOUT_TIMEOUT
```

guynir42 (approved):
Minor suggestions below.

skyportal/utils/offset.py
profjsb: it's ok to error here,  believe.

profjsb:
ok @guynir42. If tests pass, should be good to merge.

Issue #2608, opened 2022-01-28
Tests for Vega components are not really testing


guynir42:
In `tests/frontend/sources_and_followup_etc/test_sources.py` in two tests for Vega components: 
 - `test_centroid_plot` 
 - `test_source_hr_diagram` 
the line that generates the plots is not commented out, so effectively neither test is actually testing anything, because it generates a new plot each time the test is run, and it will always be the same as the current output, so the comparison of the test is trivial. 

When commenting out the line that makes the plot, the tests fail, probably due to differences in the window size or something like that.

PR #2942, opened 2022-05-05
Fix permission tests

guynir42:
The new baselayer has some changes in the way permissions are handled. In consequence, some of the handlers and several tests have to have minor tweaks to pass.


skyportal/handlers/api/photometry.py
```diff
@@ -647,7 +647,8 @@ def get_stream_ids(data, user):
                     f"Invalid format for stream id {stream_id}, must be an integer."
                 )
             stream = Stream.get_if_accessible_by(stream_id, user)
-            if stream is None:
+
+            if not stream:
```
stefanv, line 651: Is this change correct?

stefanv (commented):
Happy to merge this; can you rebase and remove the baselayer commit?

skyportal/handlers/api/photometry.py
guynir42: Yeah you are right, it should return a `None` if failing.

guynir42:
We can only merge this after the new baselayer goes in...

stefanv:
Nice, thanks for these!

guynir42:
@stefanv Should I pin the branch of BL that I am working on, just so the tests pass, and then pin it back to "main" when we merge that?

stefanv:
Sure, that's fine!

guynir42:
@stefanv any idea why the tests would fail to start up?

skyportal/tests/api/test_observation_plan.py
```diff
@@ -1,391 +0,0 @@
-import os
```
mcoughlin, line 1: You will add the file back in?

mcoughlin (commented):
@guynir42 This is clearly a triumph. Congratulations.

I had a few comments as I went through, but mostly I noted some constructs that as part of this PR you could maybe mention briefly in the docs as 'best practice'? If it's just sqlalchemy docs stuff, then fine, but I get the sense you have some knowledge to drop.

(i) This appears to be the way we will do this from now on?
            stmt = Instrument.select(self.current_user).where(
                Instrument.id == data['instrument_id']
            )
            instrument = session.scalars(stmt).first()

(2) self.verify_and_commit() -> Briefly mention how this works now?

(3) lazy='selectin', -> Briefly mention when this might be needed? 

(4) check_success, check_failure -> Are we going to slowly but surely move to use these?

(5) scroll_parent -> Is there a cost to including these? I noticed your deleted a number of them.

skyportal/tests/frontend/test_newsfeed.py
```diff
@@ -85,10 +85,8 @@ def test_news_feed_prefs_widget(
     driver.get(f'/become_user/{user.id}')
     driver.get('/')
     # Default is to not show bot comments; enable for now
-    driver.click_xpath('//*[@id="newsFeedSettingsIcon"]', scroll_parent=True)
```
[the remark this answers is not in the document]
guynir42: I think they were not necessary, and at least on my machine, often got in the way. The screen would scroll down all the way instead of just finding the thing at the top of the screen. 
I think this tools should only be used in drop down menus (or maybe when you know the thing you are looking for is low on the page).

skyportal/handlers/api/annotation.py
```diff
@@ -232,90 +232,94 @@ def post(self, associated_resource_type, resource_id):
...
-        author = self.associated_user_object
-
-        if associated_resource_type.lower() == "sources":
-            obj_id = resource_id
-            data['obj_id'] = obj_id
-            schema = Annotation.__schema__(exclude=["author_id"])
-            try:
-                schema.load(data)
-            except ValidationError as e:
+        with self.Session() as session:
```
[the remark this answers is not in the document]
guynir42: The `session.commit()` is supposed to do the same verification as `self.verify_and_commit()`.

skyportal/handlers/api/internal/token.py
```diff
@@ -41,7 +41,10 @@ def post(self):
                 "User has attempted to grant token ACLs they do not have "
                 "access to. Please try again."
             )
-        if len(user.tokens) > 0 and not user.is_admin:
+        existing_tokens = (
+            DBSession().query(Token).filter(Token.created_by_id == user.id).all()
+        )
+        if len(existing_tokens) > 0 and not user.is_admin:
```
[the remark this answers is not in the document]
guynir42: I'm going to leave that for a future PR.

skyportal/tests/api/test_candidates.py
```diff
@@ -700,7 +702,14 @@ def test_exclude_by_outdated_annotations(
     num_candidates = len(data["data"]["candidates"])
 
     origin = str(uuid.uuid4())
-    t0 = datetime.datetime.now(datetime.timezone.utc)  # recall when it was created
+    t0 = datetime.datetime.utcnow()  # recall when it was created
```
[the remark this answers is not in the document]
guynir42: Some of the tests were failing continually on my machine but passing on GA. The connection to flakiness is not obvious.

skyportal/tests/api/test_observation.py
```diff
@@ -1,349 +0,0 @@
-import os
```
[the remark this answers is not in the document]
guynir42: Sorry I thought I did that!

guynir42:
> @guynir42 This is clearly a triumph. Congratulations.
> 
> I had a few comments as I went through, but mostly I noted some constructs that as part of this PR you could maybe mention briefly in the docs as 'best practice'? If it's just sqlalchemy docs stuff, then fine, but I get the sense you have some knowledge to drop.
> 
> (i) This appears to be the way we will do this from now on? stmt = Instrument.select(self.current_user).where( Instrument.id == data['instrument_id'] ) instrument = session.scalars(stmt).first()
> 
> (2) self.verify_and_commit() -> Briefly mention how this works now?
> 
> (3) lazy='selectin', -> Briefly mention when this might be needed?
> 
> (4) check_success, check_failure -> Are we going to slowly but surely move to use these?
> 
> (5) scroll_parent -> Is there a cost to including these? I noticed your deleted a number of them.

(1) Yes, this is one way to do this. If you are not inside a session and just want to get the objects, you can also use 
```
instrument = Instrument.get(instrument_id, user_or_token, raise_if_none=False)  # get one object
instruments = Instrument.get_all(user_or_token)  # get all instruments accessible by user
```
I've added some of these examples in `baselayer/doc/dev.md` and now also referred to that in the skyportal docs. 

(2) `verify_and_commit()` still works, but you can get the same by doing this inside a handler:
```
   ...
   session.commit()
```
The `lazy="selectin"` is similar to `lazy="join"` or `lazy="subquery"` but regarded as faster by the SQLA docs. 
In all these cases it just means the relationship objects are loaded at the same time as the parent. 
This becomes important when objects are created on a session and then it closes. 
The "detached" objects can no longer fetch child objects unless they were loaded with them. 
This sets those objects to load by default. 

(5) `scroll_parent` is useful sometimes, other times it moves the window away from the button you want. 
My feeling is that it is useful in dropdown menus or in cases where the button is at the bottom of the screen.

mcoughlin (approved):
@guynir42 Thanks for the updates. Looking forward to seeing this in.

doc/dev.md
```diff
@@ -85,6 +85,8 @@ DBSession().add(obj)  # add a new object into the DB
 DBSession().commit()  # commit modifications to objects
```
stefanv, line 85: Is the part about `DBSession` up to date still?

skyportal/handlers/api/annotation.py
```diff
@@ -232,90 +232,94 @@ def post(self, associated_resource_type, resource_id):
...
+            author = self.associated_user_object
+            groups = session.scalars(
+                Group.select(self.current_user).where(Group.id.in_(group_ids))
```
stefanv, line 245: Why is author assigned above but not used here?

skyportal/handlers/api/annotation.py
```diff
@@ -232,90 +232,94 @@ def post(self, associated_resource_type, resource_id):
...
-            )
-        elif associated_resource_type.lower() == "spectra":
-            spectrum_id = resource_id
-            try:
-                spectrum = Spectrum.get_if_accessible_by(
-                    spectrum_id, self.current_user, raise_if_none=True
+            if associated_resource_type.lower() == "sources":
+                obj_id = resource_id
+                data['obj_id'] = obj_id
+                schema = Annotation.__schema__(exclude=["author_id"])
```
stefanv, line 255: Dunno if this was done before, but :+1:

skyportal/handlers/api/annotation.py
```diff
@@ -232,90 +232,94 @@ def post(self, associated_resource_type, resource_id):
...
+            try:
+                session.commit()
+            except IntegrityError as e:
+                return self.error(f'Annotation already exists: {str(e)}')
+
+            if isinstance(
+                annotation, (Annotation, AnnotationOnSpectrum)
+            ):  # annotation on object or object related data
+                self.push_all(
+                    action='skyportal/REFRESH_SOURCE',
```
stefanv, line 314: Do we need both these websocket messages for `AnnotationOnSpectrum`?

skyportal/handlers/api/candidate.py
```diff
@@ -521,43 +521,46 @@ def get(self, obj_id=None):
...
+                for filtr in g.filters
+                if g.filters is not None
             ]
-        else:
-            # If 'groupIDs' & 'filterIDs' params not present in request, use all user groups
-            group_ids = user_accessible_group_ids
-            filter_ids = user_accessible_filter_ids
+            if group_ids is not None:
+                if isinstance(group_ids, str) and "," in group_ids:
+                    group_ids = [int(g_id) for g_id in group_ids.split(",")]
```
stefanv, line 536: No exception handling here?

stefanv: May want to add something like:

`and if set(group_ids) in set(string.digits + ',')`

skyportal/handlers/api/candidate.py
```diff
@@ -521,43 +521,46 @@ def get(self, obj_id=None):
...
+                    )
+                filters = session.scalars(
+                    Filter.select(self.current_user).where(
+                        Filter.group_id.in_(group_ids)
+                    )
+                ).all()
+                filter_ids = [f.id for f in filters]
+            elif filter_ids is not None:
+                if "," in filter_ids:
+                    filter_ids = [int(f_id) for f_id in filter_ids.split(",")]
```
stefanv, line 551: Same check here

skyportal/handlers/api/classification.py
```diff
@@ -194,19 +199,18 @@ def post(self):
...
-            group_ids, self.current_user, raise_if_none=True
-        )
 
         author = self.associated_user_object
 
         # check the taxonomy
         taxonomy_id = data["taxonomy_id"]
         taxonomy = Taxonomy.get_if_accessible_by(
-            taxonomy_id, self.current_user, raise_if_none=True
+            taxonomy_id, self.current_user, raise_if_none=False
```
stefanv, line 210: Is default value True or False?  If False, can remove this flag here and elsewhere.

skyportal/handlers/api/invitations.py
```diff
@@ -102,114 +102,122 @@ def post(self):
...
-                    f"{set(stream_ids).difference({s.id for s in streams})}"
+                    "The following groupIDs elements are invalid: "
+                    f"{set(group_ids).difference({g.id for g in groups})}"
                 )
 
-            # Ensure specified groups are covered by specified streams
-            if not all(
-                [stream in streams for group in groups for stream in group.streams]
-            ):
+            if data.get("streamIDs") not in [None, "", "null", "None"]:
```
stefanv, line 138: We really need a utility function for this check, since it appears so often; not necessary for this PR.

skyportal/handlers/api/spectrum.py
```diff
@@ -154,87 +140,106 @@ def post(self):
                 f'Invalid / missing parameters; {e.normalized_messages()}'
             )
 
-        # always append the single user group
-        single_user_group = self.associated_user_object.single_user_group
+        with self.Session() as session:
+            stmt = Instrument.select(self.current_user).where(
+                Instrument.id == data['instrument_id']
+            )
+            instrument = session.scalars(stmt).first()
```
stefanv, line 147: Maybe we can standardize on `session.execute(stmt).scalars().first()` or `session.execute(stmt).all()` etc. to make it a bit clearer what's going on.  (It's a bit weird that `session.scalars` executes a statement, so may be confusing to readers.)

Can happen in another PR.

skyportal/tests/frontend/test_newsfeed.py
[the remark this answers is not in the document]
stefanv: How can you ever know where it's going to be on the page?  If it scrolls to the wrong position, please file a bug so we remember to track it.

stefanv (approved):
@guynir42 Some minor recommendations, otherwise LGTM.

A big tip-of-the-hat to you; this is a massive changeset!

Question: did you confirm that permissions work the same way they did before?  Everything I saw looked fine, just curious what mechanism we have to ensure that everything's still OK.

doc/dev.md
guynir42: I think so. The DBSession itself still does the same thing. I think I did discuss the case of nested sessions. I think we will probably need to augment the docs once we start updating code to the new scheme.

skyportal/handlers/api/annotation.py
guynir42: It is used in the different if..else cases below.

skyportal/handlers/api/annotation.py
guynir42: Yes because it updates both the annotation list on the source page and the once on the "manage data" form.

skyportal/handlers/api/spectrum.py
guynir42: I think there may be a `scalars_first` and `scalars_all` that can do that as well. I'll check on that.

skyportal/handlers/api/spectrum.py
```diff
@@ -154,87 +140,106 @@ def post(self):
...
+            session.commit()
 
-        return self.success(data={"id": spec.id})
+            self.push_all(
+                action='skyportal/REFRESH_SOURCE',
+                payload={'obj_key': spec.obj.internal_key},
+            )
+
+            self.push_all(
+                action='skyportal/REFRESH_SOURCE_SPECTRA',
```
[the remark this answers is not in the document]
guynir42: I think these are needed to refresh both the source page and the data management page (or maybe other things).

skyportal/models/user_token.py
```diff
@@ -250,9 +261,11 @@ def delete_single_user_group(mapper, connection, target):
     single_user_group = target.single_user_group
 
     # Delete single-user group
-    @event.listens_for(DBSession(), "after_flush_postexec", once=True)
+    @event.listens_for(inspect(target).session, "after_flush_postexec", once=True)
     def receive_after_flush(session, context):
-        DBSession().delete(single_user_group)
+        if single_user_group:
+            new_single_user_group = session.merge(single_user_group)
+            session.delete(new_single_user_group)
```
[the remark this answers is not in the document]
guynir42: Its really very simple (took me only a few hours to figure out):
The `single_user_group` is saved along with the `receive_after_flush` function as part of a "closure". 
That group comes from a different session so it must be merged into the current session. 
The output of the merge must be assigned a different name, otherwise the closure doesn't work, because that name is used locally and overrides the name from external scope (it throws an undefined error for `single_user_group` even if that name is used after the external variable is used). 
Then it deletes that group and everyone is happy. 
Simple, right?

skyportal/tests/__init__.py
```diff
@@ -63,3 +63,57 @@ def api(
         except requests.exceptions.JSONDecodeError:
             data = None
         return response.status_code, data
+
+
+def check_success(status, data):
```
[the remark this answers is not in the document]
guynir42: Yeah that is better.

skyportal/tests/api/test_user.py
```diff
@@ -108,6 +108,7 @@ def test_add_delete_user_adds_deletes_single_user_group(
         "GET", "groups?includeSingleUserGroups=true", token=manage_groups_token
     )
     assert data["status"] == "success"
+
     assert not any(
         [
             group["single_user_group"] and group["name"] == username
```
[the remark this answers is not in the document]
guynir42: That's not the problem. The problem is that *every time* we run the tests we generate new diagrams. Which means we are actually not testing them at all. https://github.com/skyportal/skyportal/issues/2608

guynir42:
Regarding your question: I don't know if the permissions tests are really doing anything. I tried to keep the internal permissions the same as they were before, but I don't know how to test this beyond the specific tests we wrote for individual API endpoints that test permissions.

skyportal/tests/frontend/test_newsfeed.py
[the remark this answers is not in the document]
guynir42: File a bug about the driver scrolling down too much?

skyportal/models/user_token.py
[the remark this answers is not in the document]
stefanv: Err, right ;) Please add that as an inline comment.

skyportal/tests/api/test_user.py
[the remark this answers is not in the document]
stefanv: Ah, woops, thanks for fixing it.

stefanv:
Happy to get this in as soon as tests pass.

skyportal/facility_apis/atlas.py
```diff
@@ -16,7 +16,7 @@
 
 env, cfg = load_env()
 
-if cfg['app.atlas.port'] is None:
+if cfg.get('app.atlas.port') is None:
```
stefanv, line 19: Correct fix is to define ATLAS port in the default configuration file, as per https://github.com/skyportal/skyportal/pull/3091

guynir42: In that case we can just skip the entire conditional.

stefanv: Not really, it's fine to have the port be None.

guynir42: If it is fine to have it None, then you must use `cfg.get()`.

stefanv: No, that's not accurate.  `cfg.get` is for when keys are not defined in the configuration file, just like with dics: you use `.get` when they key may be missing.  A key may have a *value* of None without being missing.

E.g.:

```
server:
  port:
```

`cfg["server.port"]` is valid, and is None.

If you wanted to parse, instead:

```
server:
  url: http://...
```

Then you'd have to use `cfg.get("server.port")`, since `port` is undefined.

guynir42: Yeah I get that. I just feel like this sort of subtlety doesn't help us much. For example, a casual sysadmin will see this config file and delete the empty port and then wonder why nothing works. Whereas we can be nice to the poor guy and give him a reasonable default in the code. 

I'm not sure what I think about having defaults defined in the config and in the code. On the one hand defining it in two places leaves us open to inconsistencies (but this already happens since we have the same things defined in multiple config files). On the other hand, this seems like a ground-floor definition in the code that will be there for you if someone decides to delete stuff they don't care about in the config. But it will also be quietly overridden by any config file(s) above that level.

stefanv: I don't understand: where will they delete it?

guynir42: I don't know if that would help. Is the default config even used once you create a `config.yaml`?

stefanv: It is, by default:

```
    # Always load the default configuration values first, and override
    # with values in user configuration files
    all_configs = [
        Path(basedir / "config.yaml.defaults"),
        Path(basedir / "../config.yaml.defaults"),
    ] + config_files
    all_configs = [cfg for cfg in all_configs if os.path.isfile(cfg)]
    all_configs = [os.path.abspath(Path(c).absolute()) for c in all_configs]

    cfg = Config(all_configs)
```

stefanv: From staging:

```
Config files: baselayer/config.yaml.defaults
                config.yaml.defaults
                config.yaml
                /etc/skyportal/config.yaml
```

guynir42: I see, but that still doesn't help if you delete a sub-key as in `atlas.port`.

guynir42: Besides, having defaults, even easy ones like `None`, are better than throwing an exception in most cases.

stefanv: How do you delete subkeys? I don't think that's possible with our current setup.

I think a default of None is not helpful. As we've seen, it hides errors when you think you are accessing keys when they don't exist (I had to fix instances of this in my PR).

guynir42: If you leave the key `atlas` and in the new config file have `atlas` with no `port` sub-key, then `cfg['atlas.port']` will send an error. If indeed there is no way to use `atlas` without a `port` then I agree there should not be a default, and then there should not be a conditional checking if it is `None`. 
On the other hand, if in some cases you don't need a `port` I don't see any harm allowing the user to either leave an empty `port` sub-key or deleting it altogether. In that case the default is `None`, you check for it (as in the way the code worked until now) and decide to use it or not. 

In other cases like spectrum types, it's not a big problem to have a hard coded default. I think that's better than having an error caused by a missing config key that's not critical to what you are working on. I could be not using spectra at all in my application, and I could delete the `spectrum_type` key and then spend hours figuring out what is `spectrum_types` and why it's complaining. Or I could just have reasonable defaults hard coded so I never have errors there. But that's just me...

stefanv: The configuration works by overriding a dictionary, so if you do not specify the atlas port it is whatever we define it to be in our default config.

The only place I think it's possible to remove entries is where you specify a list. If you specify an empty list, obviously entries from the old list would not be visible.

It is *perfectly fine* to make the code robust against cases where the configuration may be entirely missing.  In those cases (see the weather for example), just use `cfg.get` explicitly.  Again, I think it won't be necessary very often, but the option is there.

guynir42: OK. I think we are basically in agreement. Although I think the missing key will be missing in the atlas case because it is a key under another key. 

I also think using `get` is better unless you really must have something configured (then an error is the right behavior). I just feel like that's more common than you think. In any case, that requires manually moving `cfg[..]` to `cfg.get(..)` in all the places, which I don't want to do in this PR (this already touches way too many files).

stefanv: We do recursive updates. Try it out, maybe?

stefanv:
get is incorrect here since the cookie must be defined.

You can also check for None and error out with a more specific message.

guynir42:
@stefanv as far as I can tell, the field `cookie_secret` was never in `cfg`. It is in `baselayer_settings`, maybe that was the thing you wanted to check against? Right now I am not going to change it unless you can say exactly what the logic should be.

stefanv:
> @stefanv as far as I can tell, the field `cookie_secret` was never in `cfg`. It is in `baselayer_settings`, maybe that was the thing you wanted to check against? Right now I am not going to change it unless you can say exactly what the logic should be.

You should check that this is changed from the default:

https://github.com/cesium-ml/baselayer/blob/main/config.yaml.defaults#L3

guynir42:
@stefanv It makes sense to just test if `secret_key` is `abc01234`. I don't know what `cookie_secret` is supposed to do.

guynir42:
@stefan this is fully green, do you want to have another look or can we merge?

guynir42:
@stefanv @mcoughlin @profjsb Anyone want another look or can I merge this?

PR #2948, opened 2022-05-06
Add general comments to the news feed

thomasculino:
This PR aims to add the ability to comment directly on the Dashboard, on the NewsFeed component as you can see below:

general_comment

The commentEntry form has been added to the NewsFeed widget where you can type your comment. 

The particularity of these comments is that they are not related to an object.

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -14,6 +14,7 @@ from skyportal.handlers.api import (
     ClassificationHandler,
     CommentHandler,
     CommentAttachmentHandler,
+    GeneralCommentHandler,
     EnumTypesHandler,
     AnnotationHandler,
     DatalabQueryHandler,
@@ -168,6 +169,7 @@ skyportal_handlers = [
         r'/api/(sources|spectra|gcn_event)/([0-9A-Za-z-_\.\+]+)/comments(/[0-9]+)/attachment.pdf',
         CommentAttachmentHandler,
     ),
+    (r'/api/comments(/[0-9]+)?', GeneralCommentHandler),
     (r'/api/gcn_event(/.*)?', GcnEventHandler),
     (
         r'/api/localization(/[0-9]+)/airmass(/[0-9]+)?',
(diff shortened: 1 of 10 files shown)


profjsb:
Can you clarify what the use case is here? Since the news feed scrolls, these comments will quickly disappear from view from most end users. It may be searchable but since it's not connected to a source I do not understand how such comments would be used.

As for the UI/UX, I think the extra real estate on the widget is too much given how much it may be used (perhaps a popup would be better?)

thomasculino:
> Can you clarify what the use case is here? Since the news feed scrolls, these comments will quickly disappear from view from most end users. It may be searchable but since it's not connected to a source I do not understand how such comments would be used.
> 
> As for the UI/UX, I think the extra real estate on the widget is too much given how much it may be used (perhaps a popup would be better?)

This feature should enable the addition of comments unrelated to any object but also comments related to multiple ones. Concerning the location of the comments, I was asked to add the comments on the NewsFeed but I didn't realize that they would disappear that quickly. A solution for this would be to select only general_comments on the settings as shown below:

only_general_comments

Maybe adding a shortcut to do this would be a better idea. As for UI/UX part, I completely agree, I was thinking the same but wanted to have your opinion. I can try using a popup instead of the extra real estate on the widget. Maybe we can have a button showing the popup for adding a new general comment and inside this popup have another button showing only general_comments on the NewsFeed. And, to make it clear when there is a new comment, we can attach a notification badge to the first button, displaying the number of general_comments since the last time the user visualized only these comments ? I was thinking of putting this button on the newsfeed widget.

stefanv:
This needs further discussion on the next team meeting.

mcoughlin:
@thomasculino are these comments taggable with multiple like object and gcn event ids and therefore also appear as comments on those pages too?

thomasculino:
@mcoughlin Not yet. About your comment, do you mean to post these comments on the comments_on_gcn table when a gcn is tagged which will show on the gcn event page ? Or simply displaying it  on the gcn event page ? Either way, it is not implemented yet. If we want to do that we also have to think about the way we would tag these object in the general comment.

mcoughlin:
@thomasculino I am thinking then it would serve the purpose (i.e. appearing in places it will last longer term but also sufficient for rapid back and forth?)

thomasculino:
Closing this in favor of two other PRs (#3000 and another one to come). First PR is about comments on shift and the second PR should be about tagging multiple object in a comment.

PR #2949, opened 2022-05-07
Fix abs mag scale on phot plot

jadalilleboe:
I think this will fix the issue with the abs mag scale being upside down on the photometry plot.
fixes fritz-marshal/fritz#412


stefanv:
Thanks @jadalilleboe!

PR #2950, opened 2022-05-08
Fix newsfeed API

mcoughlin:
This PR makes it possible to query the news feed from the API.

### DIFF
--- a/skyportal/handlers/api/news_feed.py
+++ b/skyportal/handlers/api/news_feed.py
@@ -55,10 +55,8 @@ class NewsFeedHandler(BaseHandler):
                 schema: Error
         """
 
-        if hasattr(self.current_user, 'preferences'):
-            preferences = (
-                self.current_user.preferences if self.current_user.preferences else {}
-            )
+        if hasattr(self.current_user, 'preferences') and isinstance(self.current_user.preferences, dict):
+            preferences = self.current_user.preferences
         else:
             preferences = {}
 
--- a/skyportal/handlers/api/news_feed.py
+++ b/skyportal/handlers/api/news_feed.py
@@ -11,6 +11,7 @@ from ...models import (
 )
 
 MAX_NEWSFEED_ITEMS = 1000
+DEFAULT_NEWSFEED_ITEMS = 50
 
 
 class NewsFeedHandler(BaseHandler):
@@ -28,7 +29,7 @@ class NewsFeedHandler(BaseHandler):
             schema:
               type: integer
             description: Number of newsfeed items to return.
-            Defaults to 10. Max is {MAX_NEWSFEED_ITEMS}.
+            Defaults to {DEFAULT_NEWSFEED_ITEMS}. Max is {MAX_NEWSFEED_ITEMS}.
         responses:
           200:
             content:
@@ -55,7 +56,9 @@ class NewsFeedHandler(BaseHandler):
                 schema: Error
         """
 
-        if hasattr(self.current_user, 'preferences') and isinstance(self.current_user.preferences, dict):
+        if hasattr(self.current_user, 'preferences') and isinstance(
+            self.current_user.preferences, dict
+        ):
             preferences = self.current_user.preferences
         else:
             preferences = {}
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/news_feed.py
```diff
@@ -44,13 +54,27 @@ def get(self):
...
-        preferences = (
-            self.current_user.preferences if self.current_user.preferences else {}
-        )
-        if 'newsFeed' in preferences and 'numItems' in preferences['newsFeed']:
-            n_items = min(int(preferences['newsFeed']['numItems']), 50)
+
+        if hasattr(self.current_user, 'preferences'):
+            preferences = (
+                self.current_user.preferences if self.current_user.preferences else {}
+            )
```
guynir42, line 61:
```suggestion
        if hasattr(self.current_user, 'preferences') and isinstance(self.current_user.preferences, dict):
            preferences = self.current_user.preferences
```

skyportal/handlers/api/news_feed.py
```diff
@@ -44,13 +54,27 @@ def get(self):
...
         else:
-            n_items = 10
+            preferences = {}
+
+        n_items = self.get_query_argument("numItems", None)
+        if n_items is None:
+            if 'newsFeed' in preferences and 'numItems' in preferences['newsFeed']:
+                n_items = min(int(preferences['newsFeed']['numItems']), 50)
```
guynir42, line 68: I don't get where `50` comes from.

skyportal/tests/api/test_newsfeed.py
```diff
@@ -0,0 +1,31 @@
...
+            "ra": 235.22,
+            "dec": -23.33,
+            "redshift": 3,
+            "group_ids": [public_group.id],
+        },
+        token=upload_data_token,
+    )
+    assert status == 200
+
+    params = {'numItems': 100}
```
guynir42, line 24: What happens if there are more than 100 sources in the test DB because of previous tests?

guynir42 (commented):
Minor comments/suggestions.

skyportal/handlers/api/news_feed.py
mcoughlin: This was the original default

skyportal/tests/api/test_newsfeed.py
mcoughlin: Then this would fail. I can bump this to 1000.

skyportal/handlers/api/news_feed.py
guynir42: Shouldn't this number be either the new default or the new maximum or something that's somehow documented?

mcoughlin: yeah i am setting this to a default at the top of the file.

guynir42 (approved):

skyportal/handlers/api/news_feed.py
```diff
@@ -44,13 +55,29 @@ def get(self):
               application/json:
                 schema: Error
         """
-        preferences = (
-            self.current_user.preferences if self.current_user.preferences else {}
-        )
-        if 'newsFeed' in preferences and 'numItems' in preferences['newsFeed']:
-            n_items = min(int(preferences['newsFeed']['numItems']), 50)
+
+        if hasattr(self.current_user, 'preferences') and isinstance(
```
stefanv, line 59:
```
preferences = getattr(self.current_user, 'preferences', None) or {}
```

skyportal/handlers/api/news_feed.py
```diff
@@ -44,13 +55,29 @@ def get(self):
...
+            self.current_user.preferences, dict
+        ):
+            preferences = self.current_user.preferences
         else:
-            n_items = 10
+            preferences = {}
+
+        n_items = self.get_query_argument("numItems", None)
```
stefanv, line 66: This feels too complicated.
How about getting:
1. arg n_items
2. preference n_items
3. choose and compare

skyportal/tests/api/test_newsfeed.py
```diff
@@ -0,0 +1,31 @@
...
+            "dec": -23.33,
+            "redshift": 3,
+            "group_ids": [public_group.id],
+        },
+        token=upload_data_token,
+    )
+    assert status == 200
+
+    params = {'numItems': 1000}
+    status, data = api('GET', 'newsfeed', token=view_only_token, params=params)
```
stefanv, line 25: No check for erroneous number of items requested

PR #2954, opened 2022-05-09
TimeZoneFinder telescope longitude range

guynir42:
If any telescope has longitude > 180 the TimeZoneFinder complains that some coordinates are out of range. 
This makes sure the telescope longitude is in range for the finder.

### DIFF
--- a/skyportal/models/telescope.py
+++ b/skyportal/models/telescope.py
@@ -65,7 +65,7 @@ class Telescope(Base):
         except AttributeError:
             tf = timezonefinder.TimezoneFinder(in_memory=True)
             local_tz = tf.closest_timezone_at(
-                lng=self.lon, lat=self.lat, delta_degree=5
+                lng=(self.lon + 180) % 360 - 180, lat=self.lat, delta_degree=5
             )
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Nice catch @guynir42

stefanv:
Is this patch correct? How does the code deal with negative values?

mcoughlin:
@stefanv What do you mean?

>>> (-90) % 360
270

stefanv:
The issue description says this is a patch for the angle being > 180, but in that case:

```
In [4]: (185 + 180) % 360 - 180
Out[4]: -175
```

guynir42:
> The issue description says this is a patch for the angle being > 180, but in that case:
> 
> ```
> In [4]: (185 + 180) % 360 - 180
> Out[4]: -175
> ```
> 
> Hence my question of whether a negative result is OK here.

Yeah, I wasn't totally clear on this, @stefanv

The time zone finder expects to get a range of angles of -180 to 180, so anything above 180 is not accepted.

stefanv:
Cool, thanks!

PR #2956, opened 2022-05-09
[Analysis Epic] #1 Analysis Service Model & API

profjsb:
This is the first PR in the Analysis Epic, described in https://github.com/skyportal/skyportal/discussions/2638. 

Here we create a new model (`Analysis_Service`) and join table with `Group`s that will hold the key information needed to start an analysis on SkyPortal data, including (encrypted) authentication information for the external analysis service and data needed for the analysis.

This PR also creates a new ACL called 'Manage Analysis Services' as well as an API for creating, modifying, and removing analyzes.


skyportal/enum_types.py
```diff
@@ -74,6 +74,29 @@ def force_render_enum_markdown(values):
 
 LISTENER_CLASSES = [getattr(facility_apis, c) for c in LISTENER_CLASSNAMES]
 
+ANALYSIS_TYPES = ('lightcurve_fitting', 'spectrum_fitting')
+ANALYSIS_INPUT_TYPES = ('photometry', 'spectra', 'redshift', 'annotations', 'comments')
```
mcoughlin, line 78: Maybe meta analyses of the observations?

skyportal/handlers/api/analysis.py
```diff
@@ -0,0 +1,423 @@
...
+                    type: object
+                    description: |
+                        Authentication secrets for the service. Not needed if authentication_type is "none".
+                        This should be a string that can be parsed by the python json.loads() function and
+                        should contain the key `authentication_type`. Values of this key will be used
+                        to POST into the remote analysis service.
+                  enabled:
+                    type: boolean
+                    description: Whether the service is enabled or not.
+                  type:
```
mcoughlin, line 96: maybe this should be analysis_type? type is pretty generic.

skyportal/handlers/api/analysis.py
```diff
@@ -0,0 +1,423 @@
...
+                  input_data_types:
+                    type: array
+                    items:
+                        type: string
+                    description: |
+                        List of input data types that the service requires. Zero to many of:
+                        {', '.join(f"'{t}'" for t in ANALYSIS_INPUT_TYPES)}
+                  timeout:
+                    type: float
+                    description: Max time in seconds to wait for the analysis service to complete. Defult is 3600.0.
```
mcoughlin, line 108: Default

skyportal/handlers/api/analysis.py
```diff
@@ -0,0 +1,423 @@
...
+            return self.error(
+                '`authentication_type` is required to add an Analysis Service.'
+            )
+
+        if authentication_type not in AUTHENTICATION_TYPES:
+            return self.error(
+                f'`authentication_type` must be one of: {", ".join([t for t in AUTHENTICATION_TYPES])}.'
+            )
+        else:
+            if authentication_type != 'none':
```
mcoughlin, line 161: It might be nice to summarize in a doc/ how this is expected to work.

skyportal/handlers/api/analysis.py
```diff
@@ -0,0 +1,423 @@
...
+                  input_data_types:
+                    type: array
+                    items:
+                        type: string
+                    description: |
+                        List of input data types that the service requires. Zero to many of:
+                        {', '.join(f"'{t}'" for t in ANALYSIS_INPUT_TYPES)}
+                  timeout:
+                    type: float
+                    description: Max time in seconds to wait for the analysis service to complete. Defult is 3600.0.
```
mcoughlin, line 337: Default

skyportal/handlers/api/analysis.py
```diff
@@ -0,0 +1,423 @@
...
+                  schema: ArrayOfAnalysisServices
+            400:
+              content:
+                application/json:
+                  schema: Error
+        """
+        if analysis_service_id is not None:
+            try:
+                s = AnalysisService.get_if_accessible_by(
+                    analysis_service_id, self.current_user, raise_if_none=True
```
mcoughlin, line 250: I think we usually now do raise_if_none=False and explicitly raise the error if None usually now. But your way seems fine

skyportal/models/analysis.py
```diff
@@ -0,0 +1,154 @@
...
+
+    _authinfo = sa.Column(
+        StringEncryptedType(JSONType, cfg['app.secret_key'], AesEngine, 'pkcs5'),
+        nullable=True,
+        doc=('Contains authentication credentials for the service.'),
+    )
+
+    enabled = sa.Column(sa.Boolean, nullable=False, default=True)
+
+    type = sa.Column(
```
mcoughlin, line 106: analysis_type (or similar)

mcoughlin (commented):
A few comments. It's looking very useful. It will clearly need its own docs/ section, I am not sure how you want to address that.

skyportal/handlers/api/analysis.py
profjsb: There will be more complete description of this in the next PR (it's where we actually implement the log in).

skyportal/models/analysis.py
profjsb: Done

skyportal/handlers/api/analysis.py
profjsb: Done.

skyportal/handlers/api/analysis.py
profjsb: Done.

skyportal/handlers/api/analysis.py
profjsb: Done.

skyportal/enum_types.py
profjsb: Done.

mcoughlin (approved):
@profjsb Looks awesome. Please make sure to update the down_revision in the migration script, otherwise super happy to see this move forward.

skyportal/tests/api/test_analysis.py
```diff
@@ -0,0 +1,424 @@
...
+    analysis_service_id_1 = data['data']['id']
+
+    status, data = api('GET', 'analysis_service', token=analysis_service_token)
+    assert status == 200
+    assert data['status'] == 'success'
+
+    as_ids = [a['id'] for a in data['data']]
+
+    for as_id in [analysis_service_id, analysis_service_id_1]:
+        assert as_id in as_ids
```
guynir42, line 161:
```suggestion
    assert set(analysis_service_id, analysis_service_id_1) == set(as_ids)
```

skyportal/tests/api/test_analysis.py
```diff
@@ -0,0 +1,424 @@
...
+    assert data['status'] == 'success'
+    analysis_service_id = data['data']['id']
+
+    # This token does not belong to public_group2
+    status, data = api(
+        'GET', f'analysis_service/{analysis_service_id}', token=analysis_service_token
+    )
+    assert status == 403
+
+    # Both tokens should be able to view this annotation
```
guynir42, line 389:
```suggestion
    # Both tokens should be able to view this analysis service
```

skyportal/handlers/api/analysis.py
```diff
@@ -0,0 +1,423 @@
...
+
+        DBSession().add(analysis_service)
+        analysis_service.groups = groups
+
+        try:
+            self.verify_and_commit()
+        except IntegrityError as e:
+            return self.error(
+                f'Analysis Service with that name already exists: {str(e)}'
+            )
```
guynir42, line 206: Is this the only integrity error that can come up? Wouldn't it be safer to just propagate the error message from the DB?

skyportal/handlers/api/analysis.py
```diff
@@ -0,0 +1,423 @@
...
+                s = AnalysisService.get_if_accessible_by(
+                    analysis_service_id, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Cannot access this Analysis Service.', status=403)
+
+            analysis_dict = recursive_to_dict(s)
+            analysis_dict["groups"] = s.groups
+            return self.success(data=analysis_dict)
+
```
guynir42, line 258:
```suggestion

# retrieve multiple services: 
```

skyportal/handlers/api/analysis.py
profjsb: done.

profjsb:
Self merged following approvals from Guy and Michael.

stefanv:
Just for my own edification: (1) why is `authinfo` prefixed by an underscore and (2) is the URL the location of the external service? In one place in the code it refers to localhost:5000, making me wonder if I misunderstood.

profjsb:
@stefanv good questions!

> Just for my own edification: (1) why is `authinfo` prefixed by an underscore 

This is the convention we use in other models with sensitive information that needs to be encrypted. In general the underscore version is not accessed directly. Here's the `allocation` model (https://github.com/skyportal/skyportal/blob/c0526e242deb24f9c3c3374cc72febd4d97ac3e1/skyportal/models/allocation.py#L83)

image


> (2) is the URL the location of the external service? In one place in the code it refers to localhost:5000, making me wonder if I misunderstood.

Yes. We decided to access the services via POST requests to external 3rd party sites (hence the need for authentication). We also decided to build a locally-run (SP managed) microservice as an exemplar for others who want to build a standalone service. In the demo data we're calling the local location of the service (https://github.com/skyportal/skyportal/blob/c0526e242deb24f9c3c3374cc72febd4d97ac3e1/data/db_demo.yaml#L919).  The service also needs a callback webhook address to add back the results of the analysis. Much of this functionality will come in PR 2 and 3 soon.

PR #2958, opened 2022-05-10
Single session for external photometry request

mcoughlin:
This PR uses a single session to make external requests.

### DIFF
--- a/skyportal/facility_apis/atlas.py
+++ b/skyportal/facility_apis/atlas.py
@@ -5,6 +5,7 @@ from astropy.time import Time
 import numpy as np
 import pandas as pd
 from io import StringIO
+from sqlalchemy.orm import sessionmaker, scoped_session
 from tornado.ioloop import IOLoop
 
 from . import FollowUpAPI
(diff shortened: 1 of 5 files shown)


skyportal/facility_apis/atlas.py
```diff
@@ -280,6 +273,7 @@ def get(request):
                         req.id,
                         instrument.id,
                         request.requester.id,
+                        session,
```
stefanv, line 276: No, this one is naughty. Can't steal a session for an executor!

stefanv (commented):
Looks good, other than the executor usage of session.

skyportal/facility_apis/atlas.py
mcoughlin: @stefanv makes sense. Removed.

skyportal/handlers/api/photometry_request.py
```diff
@@ -27,10 +29,13 @@ def get(self, request_id):
                 schema: Success
         """
 
+        Session = scoped_session(
```
stefanv, line 32: This should become a bit simpler with @guynir42's refactor.

stefanv (approved):

profjsb (approved):

PR #2959, opened 2022-05-11
Tests flagged as flaky

thomasculino:
Some tests are still flagged as flaky while they do not seem to be. For example, test_new_source.py is marked as flaky but does not seem to be flaky at all. I tried running it multiple times and it never failed.

### DIFF
--- a/skyportal/tests/frontend/test_new_source.py
+++ b/skyportal/tests/frontend/test_new_source.py
@@ -1,7 +1,3 @@
-import pytest
-
-
-@pytest.mark.flaky(reruns=2)
 def test_new_source(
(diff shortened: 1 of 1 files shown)


Theodlz:
Hi @thomasculino, did you try running it only by itself or with the whole test suite it belongs too ? (like all `test_front_part_1` in the GitHub Actions for example). Because they could pass 100+ time locally when you run them alone, but sadly remain flaky when tested along with the other tests.

thomasculino:
You are right, I only ran the test by itself without running the whole test suite it belonged to. After further research, it seems that it is still flaky when running in the test suite. Therefore, I will close this PR as it's lost its purpose.

PR #2960, opened 2022-05-11
Slightly relax line lengths

stefanv:
This slightly relaxes the limit on line length from 80 to 88. This gives black a bit more freedom to reformat long lines reasonably well. Note that we also allow a 10% margin of overshoot. As they say in the black docs: "try to respect --line-length, but don’t become crazy if you can’t".

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -11,6 +11,6 @@
 # for error classes selected below.
 
 [flake8]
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Definitely a good idea.

PR #2964, opened 2022-05-12
Fix issue with converting Local input in form to UTC date

Theodlz:
It seems that a recent modification to either:
- rjsf
- dayjs
- the way the calendar that open on your browser works

affect the format of the date returned by the form (or it could just be that this error was never noticed before).

Here's what happens:

A form has a default value, and when submitting it without modifying a start and end date, all is well, and the date ends in "+00:00" as expected, which we delete. But right now, it seems that modifying the date manually or via the calendar that pops up, modifies the format of the date which then ends int '.000Z'.

This PR takes care of this by also deleting '.000Z'.

### DIFF
--- a/static/js/components/GcnSelectionForm.jsx
+++ b/static/js/components/GcnSelectionForm.jsx
@@ -175,8 +175,12 @@ const GcnSelectionForm = ({ gcnEvent }) => {
 
   const handleSubmit = async ({ formData }) => {
     setIsSubmitting(true);
-    formData.startDate = formData.startDate.replace("+00:00", "");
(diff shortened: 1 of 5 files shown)


mcoughlin (approved):
@Theodlz I tested this locally and it seems to do the trick. Thanks for fixing.

Theodlz:
> @Theodlz I tested this locally and it seems to do the trick. Thanks for fixing.

Thanks for the review @mcoughlin. As this can impact negatively the use of some features by users, i'll merge it now.

guynir42 (approved):
Seems to be ok.

PR #2971, opened 2022-05-14
For Pandas DataFrames, ensure that assignments are made on views, not copies

stefanv:
See https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#why-does-assignment-fail-when-using-chained-indexing

In our case, the code was probably doing the right thing.  But the
SettingWithCopy warning is issued conservatively whenever copies *may*
be returned, and rewriting the statements the way it is done here is
safer overall.


mcoughlin (approved):
Thanks @stefanv. I tested and it seems to do the trick!

PR #2972, opened 2022-05-16
classification shortcuts

### LINKED ISSUE #2858: User defined classification button
User defined button(s) to add to the scanning page that set classification to 1 for commonly used classifications (i.e. artifact or comet for the NEO scanning). Right now you need to click through (many) dropdowns.

jadalilleboe:
Adds classification shortcuts for selecting classifications on the scanning page
Also, I noticed there was some duplicated code with creating the headers and help icons on the preferences page so there's a new component: UserPreferencesHeader.
Hnet-image
closes #2858

### DIFF
--- a/skyportal/handlers/api/internal/profile.py
+++ b/skyportal/handlers/api/internal/profile.py
@@ -200,6 +200,10 @@ class ProfileHandler(BaseHandler):
         if not user_prefs:
             user_prefs = preferences
         else:
+            if "classificationShortcuts" in preferences:
+                user_prefs["classificationShortcuts"] = preferences[
+                    "classificationShortcuts"
+                ]
             user_prefs = recursive_update(user_prefs, preferences)
         user.preferences = user_prefs
 
--- a/skyportal/tests/frontend/test_profile.py
+++ b/skyportal/tests/frontend/test_profile.py
@@ -1,5 +1,9 @@
 import pytest
 import uuid
+from skyportal.tests import api
+from tdtax import taxonomy, __version__
+from selenium.webdriver.common.keys import Keys
+from selenium.webdriver.common.action_chains import ActionChains
 
 
 def test_token_acls_options_rendering1(driver, user):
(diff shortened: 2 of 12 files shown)


mcoughlin:
@jadalilleboe This is nice. Could this be added to the classification table such that you can not only quickly search by classification but classify objects with those same classifications in the table?

jadalilleboe:
@mcoughlin so add similar buttons here on the source page? Why add things to the classification table?
Screen Shot 2022-05-16 at 12 06 09 PM

mcoughlin:
@jadalilleboe Ah no, on that same page in the table with all of the candidates so a source can be set to a particular classification.

jadalilleboe:
Screen Shot 2022-05-16 at 12 37 45 PM
so you want buttons in here somewhere for adding classifications to a source? @mcoughlin

mcoughlin:
@jadalilleboe yeah basically like the buttons you had above that basically auto-assign those classifications for that object.

jadalilleboe:
ok! should we get this merged and that can be a different pr?

guynir42:
A) I think that can be a separate PR, 
B) If you add a bunch of classifications to a "group preset" you may not want to assign all of them together to a candidate.

guynir42:
And C) I was actually rather surprised we don't have a button to add classification on the scanning page.

guynir42:
Maybe the right thing to do is to add multiple drop-down menus on each source, where the first is "all classifications" and the others are to select one from the preset(s) the user has defined in the preferences.

mcoughlin:
@jadalilleboe once you confirm that you don't think it will require much refactor to have such buttons, I think we can merge this and you can open a follow-on PR.

jadalilleboe:
I don't think it would require much refactor, it would just be updating classifications in the db? I mean I'm not sure but I think it would flow naturally from this PR.

mcoughlin (approved):
Really cool step forward @jadalilleboe ! Feel free to merge when green

jadalilleboe:
er actually maybe we should wait, I don't know if we would want to save the classification id to the user preferences as well. and it would probably be bad if we had to change that after users start using the feature. maybe we could merge this but not have the feature available to users yet? soo just take out the part that makes it show up on the preferences page. thoughts? @mcoughlin @guynir42

mcoughlin:
@jadalilleboe What does classification id mean here?

jadalilleboe:
[https://skyportal.io/docs/api.html#tag/classifications/paths/~1api~1classification~1classification_id/get](https://skyportal.io/docs/api.html#tag/classifications/paths/~1api~1classification~1classification_id/get)
the id used in the db for uniquely identifying a classification? right now only the names of the classifications are saved to the user's preferences. I haven't looked into how classifications get added to sources yet, but I would think that would be necessary information? anyhow I realized it makes me nervous to have this implementation in production if it needs to be changed.

mcoughlin:
@jadalilleboe Ah, for posting it is the classification name (string) https://github.com/skyportal/skyportal/blob/main/skyportal/handlers/api/classification.py#L151 and the taxonomy ID (int) https://github.com/skyportal/skyportal/blob/main/skyportal/handlers/api/classification.py#L153. Probability (https://github.com/skyportal/skyportal/blob/main/skyportal/handlers/api/classification.py#L155) will just be set to 1 for these auto-clicks.

jadalilleboe:
ohh ok. So I'd need to change it to save the taxonomy id as well. so should we merge this but not have it available to the users yet?

mcoughlin:
@jadalilleboe I think it's fine to do so.

PR #2973, opened 2022-05-16
Disable classification shortcuts

jadalilleboe:
disable classification shortcuts added in #2972 until the other buttons on the scanning page are added

### DIFF
--- a/skyportal/tests/frontend/test_profile.py
+++ b/skyportal/tests/frontend/test_profile.py
@@ -1,9 +1,10 @@
 import pytest
 import uuid
-from skyportal.tests import api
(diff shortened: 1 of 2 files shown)


mcoughlin:
@jadalilleboe But this short cut still works right? It just doesn't enable setting classifications?

jadalilleboe:
Yes. So we should disable until it can also set classifications? I guess I wouldn't want users to start using it yet

mcoughlin:
@jadalilleboe I think what's there is still useful, but up to you!

mcoughlin (approved):
@jadalilleboe Merge on green.

PR #2975, opened 2022-05-17
WIP: Shift - Ask for replacement

Theodlz:
### This PR addresses a common complaint coming from people on shift, and that I noticed too while being on shift last week. It is VERY common for shifters to not be able to be on shift at a given date. When that happens, we would expect the user to be able to find someone to replace him.
Currently, on Slack (where the GRANDMA campaign that just ended took place), the user has to message everyone on a channel, hoping for an answer.
During the next campaign, where shifts will be managed directly on SkyPortal, we would expect users to ask for a replacement in a better way.
## What is the expected behavior:
A user of a Shift won't be able to cover it, and needs a replacement. He should be able to notify all other users from the group associated to the Shift, and they are expected to be able to see that request and to be able to take its place if they can.

## What does this PR add:
On the menu of a shift, a user can ask for a replacement. Then, every user from the group associated to the shift (expect him of course) receive a notification. If they click on it, they will be redirected to the associated shift, where they will have an additional dropdown list where they can select which user (out of all the users of this shift that asked for a replacement) they want to replace. On submit, the shift users will be updated.

### Additional improvements:
As some of the features mentioned above needed the shift page to be modified, this PR also refactored a lot of it, especially on how frontend elements are updated after a user's input. Now, there is less unnecessary and repetitive fetching of the shifts, as well as no unnecessary update of the store for other 
variables (like the currentShift). The page runs more smoothly, and "gets" less data from the backend.

# PREVIEW:
shift_replacement

## Suggestions:
This PR is a work in progress, the display of the button to ask for a replacement is not definitive. Same for the dropdown list to select who to replace, as well as the notification's display. Any suggestions are more then welcome !

#### Testing ! The tests have been written, and pass locally. However, for some reason, they don't pass on GitHub. What is weird is that the part that does not pass is a copy and paste of the previous tests on the shifts that do pass right now. It is just the part of creating a shift through the API and finding it on the calendar that does not work for some reason.


guynir42:
Just to clarify: If I see a message asking me to replace another user, I can click it and choose that user, and that replaces the user with myself? or is there some way for me to select anyone to replace that user? (the second option is not what we want, I think).

guynir42:
Regarding the github tests, here are some guesses:
1) There are many shifts already posted from other tests, causing the driver to miss the right one. Solution: try to run all shift related tests togther. If not, maybe there are too many users or groups from non-shift-related tests?
2) If the test always fails on github (not just flaky) it may have to do with size of the display, as Stefan mentioned, but also may be due to timezone differences (i.e., a shift gets placed in the "wrong" day). 

I will look at the code and see if anything else pops up.

static/js/components/ShiftManagement.jsx
```diff
@@ -140,23 +148,12 @@ const userLabel = (user) => {
   return label;
 };
 
-function CurrentShiftMenu() {
+function CurrentShiftMenu({ currentShift }) {
   const classes = useStyles();
-  const { shiftList } = useSelector((state) => state.shifts);
   const { permissions } = useSelector((state) => state.profile);
-  const { currentShift } = useSelector((state) => state.shift);
   const currentUser = useSelector((state) => state.profile);
```
guynir42, line 154: I think you may be grabbing the profile twice (just get the permissions from the current user.

static/js/components/ShiftPage.jsx
```diff
@@ -28,30 +29,87 @@ function datestringToDate(shiftList) {
...
+    } else if (currentShift) {
+      if (Object.keys(currentShift).length > 0) {
+        if (
+          events.find((shift) => shift.id === currentShift.id).shift_users
+            .length !== currentShift.shift_users.length
+        ) {
+          setEvents(datestringToDate(shiftList));
+        }
+      }
+    }
```
guynir42, line 54: a lot of these clauses end with the same `setEvents`. Can we compress the logic here a bit?

static/js/components/ShiftPage.jsx
```diff
@@ -28,30 +29,87 @@ function datestringToDate(shiftList) {
...
   }
 
+  useEffect(() => {
+    if (!currentShift.id && route) {
```
guynir42, line 58: is the `currentShift` going to be null or the `id`? Maybe use `currentshift?.id`

static/js/components/ShiftPage.jsx
```diff
@@ -28,30 +29,87 @@ function datestringToDate(shiftList) {
...
+        if (
+          Object.keys(updatedShift).length > 0 &&
+          Object.keys(currentShift).length > 0
+        ) {
+          if (
+            updatedShift.shift_users.some((user) =>
+              currentShift.shift_users.some(
+                (cUser) => cUser.needs_replacement === user.needs_replacement
+              )
+            )
```
guynir42, line 83: I'm not sure what you are checking here, but `needs_replacement` is a boolean so there may be an easier way to check this is true without two `some` calls.

guynir42 (commented):
Some comments below. I have not yet tested this myself. It's really important to try out all the combinations of admin/non admin users and see that it does what you think it does (following the JS logic on some of this is really hard).

Theodlz:
> Just to clarify: If I see a message asking me to replace another user, I can click it and choose that user, and that replaces the user with myself? or is there some way for me to select anyone to replace that user? (the second option is not what we want, I think). 
> 

Hi @guynir42 , thanks a lot for the review ! Plenty of interesting comments. About this, it's the first option, you can only replace a user by yourself on the frontend.

Theodlz:
> Regarding the github tests, here are some guesses:
> 
> 1. There are many shifts already posted from other tests, causing the driver to miss the right one. Solution: try to run all shift related tests togther. If not, maybe there are too many users or groups from non-shift-related tests?
> 2. If the test always fails on github (not just flaky) it may have to do with size of the display, as Stefan mentioned, but also may be due to timezone differences (i.e., a shift gets placed in the "wrong" day).
> 
> I will look at the code and see if anything else pops up.

Tests are passing now !

static/js/components/ShiftPage.jsx
Theodlz: Good idea. I forgot the existence of that syntax entirely... I'll try to use that instead

guynir42:
> > Regarding the github tests, here are some guesses:
> > 
> > 1. There are many shifts already posted from other tests, causing the driver to miss the right one. Solution: try to run all shift related tests togther. If not, maybe there are too many users or groups from non-shift-related tests?
> > 2. If the test always fails on github (not just flaky) it may have to do with size of the display, as Stefan mentioned, but also may be due to timezone differences (i.e., a shift gets placed in the "wrong" day).
> > 
> > I will look at the code and see if anything else pops up.
> 
> Tests are passing now ! I just changed how i wrote them, its all in one test now and i re used the shift already created in the previous test rather than posting a new one, that way i also avoid creating additional data in the db for nothing, which could make some other tests flaky.

This seems to me like the problem may have been with the first shift blocking the second shift or something like that. 
In general, we need to always remember other tests may be generating similar objects and account for that by using randomly generated names, dates, etc.

skyportal/handlers/api/shift.py
```diff
@@ -433,18 +431,17 @@ def patch(self, shift_id, user_id):
             # send a user notification to all members of the group associated to the shift
             # that the user needs to be replaced
             # recover all group users associated to the shift
-            shift = (
-                Shift.query_records_accessible_by(
+            try:
+                shift = Shift.get_if_accessible_by(
+                    [shift_id],
```
guynir42, line 436: you can drop the brackets and then you will get just the one shift, so you can also drop the `[0]` indexing below.

guynir42 (commented):
Minor comment below. Other than that it looks fine. 

I am worried about the change to the shift model, where you let anyone with read access modify the shifts. It isn't a big security risk, but you generally don't want to allow non-admin users to do what they want with the shift (even if the front end blocks them from doing so, the API is still there). Maybe there are API level checks that I've missed that prevent non-admin users from changing shifts?

Theodlz:
> about

@guynir42 thank you very much for the review. There is a check at the handler / api level for the Manage Shift ACL. Issue here is that if I want a user to be able to replace another user, he needs to be able to delete him from the shift. Maybe I cant resteict opérations on the shift, and make it more permissible only on the shift users table ? Thats still better than what it does right now. I could also verify that the user is part of the shift. Could be something like : "a non admin user can delete a shift user only if he needs a replacement and if the user to delete is a member of the shift". But im not sure thats possible for the first part.

guynir42:
@Theodlz I think it is best to just make sure that the specific call to the handler is legal. 
E.g., if a non-shift admin wants to delete a user the whole transaction must be legal (the new user must take the old user's place). Or maybe it is ok for a user to delete himself as well. 
All of this has to be enforced to prevent misuse of the API. Since we are taking off the guard rails for this, it is good to add some tests that make sure that API calls fail when the permissions are not good.

Theodlz:
I'm implementing this right now : a non admin user that has the Manage shift ACL can only delete a shift user if he that user has asked to be replaced. Makes sense no ? And you can only modify yourself unless you are an admin

Theodlz:
@guynir42 just noticed a problem with the permissions i just added on the model. Please ignore the review request for now !

Theodlz:
> @guynir42 i'm implementing this right now : a non admin user that has the Manage shift ACL can only delete a shift user if he that user has asked to be replaced, or if he is himself. Makes sense no ? And you can only modify yourself unless you are an admin.
> 
> EDIT: So i added both specific access logic on update and delete, just commited it.
> 
> Also, as the permission the remove has constraints that the permission to add does not have, if the user is able to remove, it means that he is necessarily able to create. So if he can remove a user, its means that he can also add himself to the shift. Doesnt that imply that as soon as you can remove a user, the whole transaction (removing then adding) is legal ?

Hi @guynir42, i tried different scenarios of basic users with Manage Shift permissions, without the permission, system admins, group admins, shift admins ... and everything seems to make sense and to follow the rules i mentioned above. The tests also pass. I hope its all in order but i'll let you have a look. If you have any suggestions as how to write the permissions more cleanly in the models (the one to check if the current user is an admin of the shift), i'll take it ! I'm really not that knowledgeable in sqlalchemy... I really get confused by the syntax sometimes.

skyportal/handlers/api/shift.py
```diff
@@ -30,7 +30,7 @@ def post(self):
                   - type: object
                     properties:
                       shift_admins:
-                        type: array
+                        type: arrayF
```
guynir42, line 33: This is a typo, right?

guynir42 (approved):
Looks good.

Theodlz:
I caught a little something i would like to try to fix before merging. The frontend test is a bit flaky because: on the form to add and remove users, if you select users to both add and delete, there is a split second where after adding, the remove button will be deactivated (as he rerenders the component and verifies again if users selected are to be removed or added). So the wait_for_xpath finds it, but then it is deactivated when the click_xpath runs.

skyportal/tests/frontend/test_shifts.py
```diff
@@ -112,9 +113,11 @@ def test_shift(
     driver.wait_for_xpath(add_users_button)
     driver.click_xpath(add_users_button)
 
-    # check for button to add and remove users
     remove_users_button = '//*[@id="remove-users-button"]'
-    driver.wait_for_xpath(remove_users_button)
+    # waith for xpath to stay for a second to allow the button to be clickable
+    driver.wait_for_xpath(remove_users_button, timeout=1)
+    time.sleep(1)
+    driver.wait_for_xpath(remove_users_button, timeout=1)
```
guynir42, line 120:
```suggestion
    time.sleep(1)
```
Seems to me like you can just put the wait and then click (I believe click_xpath also does a wait_for_xpath internally). 
Could be worth a shot. 
In any case, add a one-line comment why you add the wait time

guynir42, line 120:
```suggestion
    time.sleep(1)
```
Seems to me like you can just put the wait and then click (I believe click_xpath also does a wait_for_xpath internally). 
Could be worth a shot. 
In any case, add a one-line comment why you add the wait time

guynir42 (approved):

mcoughlin (approved):
@Theodlz another cool use case.

PR #2976, opened 2022-05-17
Add concept of FacilityTransactionRequest

mcoughlin:
This PR starts to add a FacilityTransactionRequest queue for interacting with external APIs.


services/facility_queue/facility_queue.py
```diff
@@ -0,0 +1,195 @@
...
+                            followup_request.id,
+                            instrument.id,
+                            followup_request.requester.id,
+                        )
+                        req.status = 'complete'
+                        session.add(req)
+
+                    elif json_response['starttimestamp']:
+                        followup_request.status = f"Task is running (started at {json_response['starttimestamp']})"
+                        await self.put(req)
```
stefanv, line 99: This means that, if the queue is empty and there is only one request, it will keep pinging the remote server.  Maybe check the difference between current time and last time tried, and ensure that is longer than, what, 2 minutes or so?

services/facility_queue/facility_queue.py
```diff
@@ -0,0 +1,195 @@
...
+        except TypeError:
+            self.set_status(400)
+            return self.write(
+                {
+                    "status": "error",
+                    "message": "Invalid arguments; cannot construct facility request",
+                }
+            )
+
+        for field in ('method', 'endpoint'):
```
stefanv, line 146: This was just my dummy checking to ensure data look OK.  Needs to be matched to whatever you send in.

services/facility_queue/facility_queue.py
```diff
@@ -0,0 +1,195 @@
...
+
+        session = Session()
+
+        current_req = session.execute(
+            sa.select(FacilityTransactionRequest).where(
+                FacilityTransactionRequest.followup_request_id
+                == req.followup_request_id
+            )
+        ).first()
+        if current_req is not None:
```
stefanv, line 164: Should we also add a check for duplicate requests with different IDs (i.e. poking at the same external URL)?

services/facility_queue/facility_queue.py
mcoughlin: Can this just be removed because the lines in :

        # validate data here and return 400 if invalid
        try:
            req = FacilityTransactionRequest(**data)
        except TypeError:
            self.set_status(400)
            return self.write(
                {
                    "status": "error",
                    "message": "Invalid arguments; cannot construct facility request",
                }
            )
       
checks this?

services/facility_queue/facility_queue.py
mcoughlin: Added!

services/facility_queue/facility_queue.py
mcoughlin: Nice idea too.

services/facility_queue/facility_queue.py
```diff
@@ -56,63 +55,74 @@ async def load_from_db(self):
...
-                        await self.put(req)
-                    else:
-                        followup_request.status = f"Waiting for job to start (queued at {json_response['timestamp']})"
-                        await self.put(req)
-                else:
-                    followup_request.status = f'error: {response.content}'
+            dt = datetime.utcnow() - req.last_query
+            if dt < WAIT_TIME_BETWEEN_QUERIES:
+                print(
+                    f"Holding {req.id} for {(WAIT_TIME_BETWEEN_QUERIES-dt).seconds} seconds"
```
stefanv, line 64: Note again that if this is the only entry in the queue you will see a LOT of these messages.

services/facility_queue/facility_queue.py
```diff
@@ -56,63 +55,74 @@ async def load_from_db(self):
...
+                )
+                await self.put(req)
+            else:
+                session = Session()
```
stefanv, line 68: with Session syntax?

services/facility_queue/facility_queue.py
mcoughlin: Should we just not post this?

services/facility_queue/facility_queue.py
```diff
@@ -56,63 +55,74 @@ async def load_from_db(self):
...
+                                altdata,
+                                followup_request.id,
+                                instrument.id,
+                                followup_request.requester.id,
+                            )
+                            req.status = 'complete'
+                            session.add(req)
+                            print(f"Request {req.id} completed")
+
+                        elif json_response['starttimestamp']:
```
stefanv, line 105: With all these clauses, just think about that one item in the queue scenario, so that we don't churn up the DB or the output terminal.

services/facility_queue/facility_queue.py
mcoughlin: This gives me issues like:

ERROR:tornado.application:Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x15e06c040>>, <Task finished name='Task-2' coro=<FacilityQueue.service() done, defined at /Users/mcoughlin/Code/ZTF/skyportal-branches/query_queue/skyportal/services/facility_queue/facility_queue.py:55> exception=DetachedInstanceError('Instance <FacilityTransactionRequest at 0x15e075b50> is not bound to a Session; attribute refresh operation cannot proceed')>)

Theodlz (commented):
No comment! The code looks very good. Is there any way for me to test it by sending fake requests? I haven't worked on any of skyportal's API interactions with follow-up requests so far so I'm not sure how to test it properly !

mcoughlin:
@Theodlz Actually better would be to make a quick real test. Can you follow:

* ATLAS Forced Photometry: A user account must be made on https://fallingstar-data.com/forcedphot/, at which point the authentication takes the form `{"api_token": "testtoken"}`.

i.e. in db_demo

_altdata: '{"api_token": "testtoken"}'

in     proposal_id: ATLAS-001

Theodlz:
Amazing. Let we look at this now.

mcoughlin:
@Theodlz once that is in, just go to a source and trigger an atlas request in the dropdown. then you can click retrieve and it will send that request to the microservice.

services/facility_queue/supervisor.conf
```diff
@@ -0,0 +1,5 @@
+[program:migration]
```
Theodlz, line 1: Is it labeled as migration on purpose ?

mcoughlin: oops

services/facility_queue/facility_queue.py
mcoughlin: I reworked the sessions to make this happen.

services/facility_queue/facility_queue.py
mcoughlin: yeah I made fewer of them.

Theodlz:
Everything works well for me now! Maybe just hiding the `retrieve` button on the frontend when the request is already completed and I think it will be pretty much done !

mcoughlin:
@Theodlz I am actually less clear what to be checking for for that because the facility requests do everything through the transactions. There arent really messages to parse.

Theodlz (approved):
No further comments, everything works well for me !

mcoughlin:
Given Theo's testing and approval and the addressed responses to Stefan comments, I am merging this. I think the best thing for this PR is now testing in the real environment.

PR #2978, opened 2022-05-18
Add classifications on scanning page

jadalilleboe:
Add classifications on scanning page

full interaction:
Hnet-image (1)


mcoughlin (approved):
@jadalilleboe wow that's nifty.

PR #2979, opened 2022-05-18
Remove old stuff on phot plot

jadalilleboe:
remove old stuff on phot plot, the weird form


mcoughlin:
@jadalilleboe can you say more about what this does? Is stuff disappearing off the plot?

jadalilleboe:
@mcoughlin Yeah removes this form off the plots
image
It's useless and takes up a lot of space and will be replaced soon so might as well remove it I think.

jadalilleboe:
not sure, thomas has been working on the new photometry plotting preferences i think
I just wanted to get rid of that old one before deployment because of how much space it takes up. and it is not useful on its own without saving stuff to the user's preferences.

mcoughlin (approved):
Looking forward to the fix!

PR #2983, opened 2022-05-19
tools/set_user_role.py - error when adding a role to a user that has no roles yet.

Theodlz:
Using it in production i discovered an issue i couldn't have noticed when using `make load_demo_data`, is that if a user has no roles to begin with, we can't set a new role for him. A for loop tries to access the roles in user.roles, but as user.roles is empty, it can't iterate.


mcoughlin (approved):
Thanks @Theodlz.

PR #2987, opened 2022-05-19
Telescope bad latitudes

guynir42:
Deal with bad latitudes, longitudes, elevations on telescopes


skyportal/models/telescope.py
```diff
@@ -63,14 +64,33 @@ def observer(self):
         try:
             return self._observer
         except AttributeError:
+            if (
```
stefanv, line 67:
```
try:
    lon = float(self.lon)
    lat = float(self.lat)
    if np.isnan(lon) or np.isnan(lat):
        return None
except TypeError:
    return None
```

stefanv: Anyway, that was the suggestion I was going to make; exception handling version is fine too.

stefanv (approved):
- Comment on warning catch
- Log exception when telescope has invalid lat/lon

skyportal/models/telescope.py
```diff
@@ -63,16 +67,34 @@ def observer(self):
...
+                    longitude=self.lon * u.deg,
+                    latitude=self.lat * u.deg,
+                    elevation=elevation * u.m,
+                    timezone=local_tz,
+                )
+
+            except Exception:
+                log(
+                    f'Telescope {self.id} ("{self.name}") cannot calculate an observer.'
+                )
```
stefanv, line 96:
```suggestion
            except Exception as e:
                log(
                    f'Telescope {self.id} ("{self.name}") cannot calculate an observer: {e}'
                )
```

skyportal/models/telescope.py
```diff
@@ -106,30 +134,57 @@ def next_twilight_morning_nautical(self, time=None):
         if time is None:
             time = ap_time.Time.now()
         observer = self.observer
-        return observer.twilight_morning_nautical(time, which='next')
+        if observer is None:
+            return None
+        with warnings.catch_warnings():
+            # for telescopes above the arcric circle (or below antarctic circle)
```
stefanv, line 140:
```suggestion
            # for telescopes above the arctic circle (or below antarctic circle)
```

skyportal/models/telescope.py
```diff
@@ -106,30 +134,57 @@ def next_twilight_morning_nautical(self, time=None):
...
+            # there is no morning nautical twilight
+            # so it would return a MaskedArray and spew a warning.
```
stefanv, line 142:
```suggestion
            # so this returns a MaskedArray and raises a warning.
```

Issue #2985, opened 2022-05-19
Adding telescope should check for valid latitude and longitudes


mcoughlin:
https://github.com/skyportal/skyportal/blob/main/static/js/components/NewTelescope.jsx

should have a simple check on latitude and longitude.

Theodlz:
Hi @mcoughlin . Should it check if it's null or not, or should it also check that the values are in the correct range ? (i.e -90, 90 for lon, and -180, 180 for lat)

Theodlz:
Should be solved here: https://github.com/skyportal/skyportal/pull/2989

The issue was in fact the elevation, not just lon and lat

PR #2989, opened 2022-05-20
lon lat elevation : validation in NewTelescope.jsx and in the handler if the telescope is fixed

Theodlz:
Solves issue mentionned in https://github.com/skyportal/skyportal/issues/2985, where we would encounter an error because the lon, lat, elevation field are required if the telescope has a fixed location, but that wasn't verified in both frontend and backend.

### DIFF
--- a/skyportal/handlers/api/telescope.py
+++ b/skyportal/handlers/api/telescope.py
@@ -45,7 +45,7 @@ class TelescopeHandler(BaseHandler):
             if data['fixed_location']:
                 if 'lat' not in data or 'lon' not in data or 'elevation' not in data:
                     return self.error(
-                        'Missing latitude, longitude or elevation, required if the telescope is fixed'
+                        'Missing latitude, longitude, or elevation; required if the telescope is fixed'
                     )
                 elif (
                     not isinstance(data['lat'], (int, float))
@@ -53,7 +53,7 @@ class TelescopeHandler(BaseHandler):
                     or not isinstance(data['elevation'], (int, float))
                 ):
                     return self.error(
-                        'Latitude, longitude and elevation must all be numbers'
+                        'Latitude, longitude, and elevation must all be numbers'
                     )
                 elif (
                     data['lat'] < -90
@@ -63,7 +63,7 @@ class TelescopeHandler(BaseHandler):
                     or data['elevation'] < 0
                 ):
                     return self.error(
-                        'Latitude must be between -90 and 90, longitude between -180 and 180 and elevation must be positive'
+                        'Latitude must be between -90 and 90, longitude between -180 and 180, and elevation must be positive'
                     )
         try:
             telescope = schema.load(data)
--- a/skyportal/handlers/api/telescope.py
+++ b/skyportal/handlers/api/telescope.py
@@ -44,8 +44,26 @@ class TelescopeHandler(BaseHandler):
         if 'fixed_location' in data:
             if data['fixed_location']:
                 if 'lat' not in data or 'lon' not in data or 'elevation' not in data:
-                    raise ValidationError(
+                    return self.error(
                         'Missing latitude, longitude or elevation, required if the telescope is fixed'
+                    )
+                elif (
+                    not isinstance(data['lat'], (int, float))
+                    or not isinstance(data['lon'], (int, float))
+                    or not isinstance(data['elevation'], (int, float))
+                ):
+                    return self.error(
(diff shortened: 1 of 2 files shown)


Theodlz:
@guynir42 tagging you as i saw you mentionned the error caused by this on slack

guynir42:
Thanks @Theodlz , I will look into it tomorrow morning. 
In the meanwhile we are also adding protection against bad data already in the telescope list (https://github.com/skyportal/skyportal/pull/2987)

static/js/components/NewTelescope.jsx
```diff
@@ -29,11 +29,39 @@ const NewTelescope = () => {
   };
 
   function validate(formData, errors) {
+    console.log(formData);
```
mcoughlin, line 32: Remove console statement.

static/js/components/NewTelescope.jsx
```diff
@@ -29,11 +29,39 @@ const NewTelescope = () => {
...
     telescopeList?.forEach((telescope) => {
       if (formData.name === telescope.name) {
         errors.name.addError("Telescope name matches another, please change.");
       }
     });
+    if (formData.fixed_location === true) {
+      if (formData.lon === undefined || formData.lat === undefined) {
```
mcoughlin, line 39: Maybe split these up?

static/js/components/NewTelescope.jsx
```diff
@@ -29,11 +29,39 @@ const NewTelescope = () => {
...
+        errors.lon.addError(
+          "Longitude must be specified if telescope is fixed."
+        );
+
+        errors.lat.addError(
+          "Latitude must be specified if telescope is fixed."
+        );
+      } else if (
+        formData.lon < -180 ||
```
mcoughlin, line 48: Similar split here?

mcoughlin (approved):
@Theodlz Looks good. For clarity, would be nice to split up the latitude and longitude reporting / checks.

Theodlz:
> @Theodlz Looks good. For clarity, would be nice to split up the latitude and longitude reporting / checks.

Added it right now @mcoughlin . I think its a good idea for the frontend as it happens all errors in a list to then display all of them. But in the handler, as we return just once. I think they should be grouped by type of error (range of value, undefined), as anyway, if the message is clear (specifying the ranges, or that lon, lat or elevation were undefined) it should be enough for the user to understand.

skyportal/handlers/api/telescope.py
```diff
@@ -40,7 +40,31 @@ def post(self):
         data = self.get_json()
 
         schema = Telescope.__schema__()
-
+        # check if the telescope has a fixed location
+        if 'fixed_location' in data:
+            if data['fixed_location']:
+                if 'lat' not in data or 'lon' not in data or 'elevation' not in data:
+                    return self.error(
+                        'Missing latitude, longitude or elevation, required if the telescope is fixed'
```
stefanv, line 48:
```suggestion
                        'Missing latitude, longitude, or elevation; required if the telescope is fixed'
```

skyportal/handlers/api/telescope.py
```diff
@@ -40,7 +40,31 @@ def post(self):
...
+                    )
+                elif (
+                    not isinstance(data['lat'], (int, float))
+                    or not isinstance(data['lon'], (int, float))
+                    or not isinstance(data['elevation'], (int, float))
+                ):
+                    return self.error(
+                        'Latitude, longitude and elevation must all be numbers'
```
stefanv, line 56:
```suggestion
                        'Latitude, longitude, and elevation must all be numbers'
```

skyportal/handlers/api/telescope.py
```diff
@@ -40,7 +40,31 @@ def post(self):
...
+                    )
+                elif (
+                    data['lat'] < -90
+                    or data['lat'] > 90
+                    or data['lon'] < -180
+                    or data['lon'] > 180
+                    or data['elevation'] < 0
+                ):
+                    return self.error(
+                        'Latitude must be between -90 and 90, longitude between -180 and 180 and elevation must be positive'
```
stefanv, line 66:
```suggestion
                        'Latitude must be between -90 and 90, longitude between -180 and 180, and elevation must be positive'
```

stefanv (approved):

mcoughlin (approved):
Thanks @Theodlz

PR #2990, opened 2022-05-20
Removed unexpected xpassed and xfailed in test_user_profile.py

Theodlz:
Currently, when running GitHub actions, one test from test_user_profile.py consistently fails with the following error: `xfailed` , which is ignored. This PR fixes that test.


stefanv (approved):

stefanv:
Nice work, thank you!

PR #2993, opened 2022-05-21
Fix tests

guynir42:
Some of the frontend tests were failing or flaky locally. 

I've fixed a few of the tests and a couple of underlying issues as well.

### DIFF
--- a/skyportal/tests/frontend/test_newsfeed.py
+++ b/skyportal/tests/frontend/test_newsfeed.py
@@ -1,7 +1,6 @@
 import uuid
 import pytest
 from selenium.webdriver.common.action_chains import ActionChains
-
 from skyportal.tests import api
 
 
@@ -51,7 +50,6 @@ def test_news_feed(driver, user, public_group, upload_data_token, comment_token)
         driver.wait_for_xpath(f'//p[contains(text(),"comment_text_{i}")]')
 
 
-@pytest.mark.flaky(reruns=2)
 def test_news_feed_prefs_widget(
     driver, user, public_group, upload_data_token, comment_token
 ):
@@ -85,10 +83,9 @@ def test_news_feed_prefs_widget(
     driver.get(f'/become_user/{user.id}')
     driver.get('/')
     # Default is to not show bot comments; enable for now
-    driver.click_xpath('//*[@id="newsFeedSettingsIcon"]', scroll_parent=True)
-    driver.click_xpath(
-        '//*[@data-testid="categories.includeCommentsFromBots"]', scroll_parent=True
-    )
+    driver.click_xpath('//*[@id="newsFeedSettingsIcon"]')
+    driver.click_xpath('//*[@data-testid="categories.includeCommentsFromBots"]')
+
     driver.click_xpath('//button[contains(., "Save")]')
     driver.wait_for_xpath('//span[text()="a few seconds ago"]')
     for i in range(2):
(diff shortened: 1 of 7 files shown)


skyportal/tests/frontend/test_shifts.py
```diff
@@ -102,20 +102,25 @@ def test_shift(
...
 
     # check for button to add and remove users
     add_users_button = '//*[@id="add-users-button"]'
     driver.wait_for_xpath(add_users_button)
     driver.click_xpath(add_users_button)
 
     remove_users_button = '//*[@id="remove-users-button"]'
-    # As the component rerenders, the remove button will be deactivated for a bit, so we wait for the xpath to stay for a second to allow the button to be clickable
-    # if we checked for the xpath right now, it might disappear right after when we try to click it. So we add a little delay before clicking the button
+    # As the component rerenders,
```
stefanv, line 117: Why wrap on 54?

skyportal/tests/frontend/test_shifts.py
```diff
@@ -102,20 +102,25 @@ def test_shift(
...
+    # the remove button will be deactivated for a bit,
+    # so we wait for the xpath to stay for a second
+    # to allow the button to be clickable
+    # if we checked for the xpath right now,
+    # it might disappear right after when we try to click it.
+    # So we add a little delay before clicking the button
     time.sleep(1)
```
stefanv, line 124: The correct thing is probably to wait for the button to be disabled, and then again for it to be enabled.

skyportal/tests/frontend/test_shifts.py
```diff
@@ -132,18 +137,23 @@ def test_shift(
...
     driver.wait_for_xpath(deactivated_add_users_button)
+    # see comment above
+    time.sleep(1)
     remove_users_button = '//*[@id="remove-users-button"]'
-    driver.wait_for_xpath(remove_users_button)
-    driver.click_xpath(remove_users_button)
+    button = driver.wait_for_xpath(remove_users_button)
+    ActionChains(driver).move_to_element(button).click().perform()
+    # click once to close the menu, and again to remove the user
+    time.sleep(1)
```
stefanv, line 155: Explicit sleeps, bad.

skyportal/tests/frontend/test_shifts.py
```diff
@@ -259,16 +263,27 @@ def test_shift(
...
+
+    assert title.text == form_name  # we found the right shift
 
     delete_button_xpath = '//*[@id="delete_button"]'
-    driver.wait_for_xpath(delete_button_xpath)
-    driver.click_xpath(delete_button_xpath)
+    driver.wait_for_xpath(delete_button_xpath).click()
+    try:  # try again
+        driver.wait_for_xpath(delete_button_xpath, timeout=1).click()
+    except TimeoutException:
```
stefanv, line 284: Why do we need to wait twice? Can't we set the timeout to be longer?

skyportal/tests/frontend/test_newsfeed.py
```diff
@@ -100,22 +97,22 @@ def test_news_feed_prefs_widget(
         # Comment item
         driver.wait_for_xpath(f'//p[contains(text(),"comment_text_{i}")]')
 
-    driver.click_xpath('//*[@id="newsFeedSettingsIcon"]', scroll_parent=True)
```
stefanv, line 103: Safe to remove these?

stefanv (commented):
Looks good overall; a few questions.

skyportal/tests/frontend/test_shifts.py
guynir42: I was just wrapping based on natural breaks in the text that are less than 80.

skyportal/tests/frontend/test_newsfeed.py
guynir42: Locally it seems to work better without them.

skyportal/tests/frontend/test_shifts.py
guynir42: I've just noticed that sometimes the click doesn't register. So if the button is still around the second time (with a timeout=1) will get it.

profjsb:
What's the status here?

guynir42:
> What's the status here?

I should get back to this after my big baselayer/skyportal update gets merged. I have a lot of bug fixes there, too. 

After that's done I will go over all of these again, see what's still relevant.

guynir42:
Closing as I feel like most of these were addressed in other PRs.

PR #2995, opened 2022-05-22
Add option for default groups for an allocation to share with

mcoughlin:
This PR adds an option for default groups for an allocation to share with.


skyportal/models/allocation.py
```diff
@@ -35,6 +35,9 @@ class Allocation(Base):
     hours_allocated = sa.Column(
         sa.Float, nullable=False, doc='The number of hours allocated.'
     )
+    default_share_group_ids = sa.Column(
+        sa.ARRAY(sa.Integer), comment='List of default group IDs to share date with'
```
guynir42, line 39:
```suggestion
        sa.ARRAY(sa.Integer), comment='List of default group IDs to share data with'
```

guynir42 (approved):

PR #2999, opened 2022-05-25
Health monitor: restart unresponsive app instances

stefanv:


stefanv:
First require one successful ping before monitoring health.

services/health_monitor/health_monitor.py
```diff
@@ -0,0 +1,80 @@
...
+        subprocess.run(supervisorctl + cmd, check=True)
+    except subprocess.CalledProcessError as e:
+        log(f'Failure calling supervisorctl; could not restart app {app_nr}')
+        log(f'Exception: {e}')
+
+
+if __name__ == "__main__":
+    log('Monitoring system health')
+    downtimes = {}
+
```
profjsb, line 64: Wait here until all the backends are up for the first time.

profjsb (changes requested):
Looks good. As discussed you should wait until all the apps are up/responsive before looking to restart them if they go down: this will insure that this service does not interfere with the normal startup process.

services/health_monitor/health_monitor.py
```diff
@@ -0,0 +1,89 @@
...
+        backends_seen = backends_seen | (all_backends - down)
+
+        downtimes = {k: downtimes.get(k, time.time()) for k in (down & backends_seen)}
+
+        for app in list(downtimes.keys()):
+            downtime = time.time() - downtimes[app]
+            if downtime > ALLOWED_DOWNTIME_SECONDS:
+                log(f'App {app} unresponsive: restarting')
+                restart_app(app)
+                del downtimes[app]
```
profjsb, line 89: Do you also need to remove this app from backends_seen so that we wait for it to come back up before monitoring?

profjsb (commented):
See inline comment

services/health_monitor/health_monitor.py
stefanv: Good suggestion; done.

PR #3000, opened 2022-05-25
Comments on shift

thomasculino:
This PR aims to add the ability to comment on a shift by adding a new table called comments_on_shifts. 
comments_on_shift
During an observation campaign, people are on shift to observe anytime they want. If something happens during their shift, they should have a place to report to. The comments on shift aims to be the place they need. You can add a comment by clicking on a shift and then adding a comment using the form.


static/js/components/CommentList.jsx
```diff
@@ -377,6 +403,28 @@ const CommentList = ({
...
+                            size="small"
+                            color="primary"
+                            type="button"
+                            name={`deleteCommentButtonShift${id}`}
+                            onClick={() => deleteCommentOnShift(id)}
+                            className="commentDelete"
+                          >
+                            <CloseIcon fontSize="small" />
+                          </Button>
+                        )}
```
jadalilleboe, line 427: since this code is repeated twice, would it make sense to extract into its own component? not exactly related to this PR, but I find the component hard to read since it's wrapped in a big if based on if the comments are compact or not. I think it would be better to split into two components, compact comment list and regular comment list, and extract out some of the components that would be common to both like the delete button. 

Also, it would be nice if there was less duplicated code for checking the associatedResourceType. I think some of that could be extracted out into helper functions which check the associatedResourceType. So instead of having four delete button declarations, there would just be one. Some thoughts about this component as a whole, but if you don't want to change it in this PR we should make an issue to clean this up.

static/js/components/ShiftManagement.jsx
```diff
@@ -806,4 +810,18 @@ CurrentShiftMenu.propTypes = {
   }).isRequired,
 };
 
-export default CurrentShiftMenu;
+export function CommentOnShift(resourceType) {
```
jadalilleboe, line 813: since it's a CommentOnShift component, is there any reason to pass in the resourceType as an argument? Won't it always be "shift"?

static/js/components/ShiftPage.jsx
```diff
@@ -71,23 +71,29 @@ const ShiftPage = ({ route }) => {
...
+            // check if the users have the same ids, or if they need a replacement when they didnt need one before, and vice versa
+            for (let i = 0; i < updatedShift.shift_users.length; i += 1) {
+              if (
+                updatedShift.shift_users[i].id !==
+                  currentShift.shift_users[i].id ||
+                updatedShift.shift_users[i].needs_replacement !==
+                  currentShift.shift_users[i].needs_replacement ||
+                updatedShift.shift_users[i].modified !==
+                  currentShift.shift_users[i].modified
+              ) {
```
jadalilleboe, line 89: Could you explain this code a bit more in the comments? also would be nice to have some parentheses in long if conditions.

static/js/components/CommentList.jsx
```diff
@@ -283,6 +304,10 @@ const CommentList = ({
     dispatch(gcnEventActions.deleteCommentOnGcnEvent(gcnID, commentID));
   };
 
+  const deleteCommentOnShift = (commentID) => {
+    dispatch(shiftActions.deleteCommentOnShift(commentID));
+  };
+
```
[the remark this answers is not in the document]
thomasculino: I corrected it now it should be fine.

static/js/components/ShiftManagement.jsx
thomasculino: True, I fixed it.

skyportal/handlers/api/comment.py
```diff
@@ -346,20 +362,65 @@ def post(self, associated_resource_type, resource_id):
...
+            if users_mentioned_in_comment:
+                for user_mentioned in users_mentioned_in_comment:
+                    DBSession().add(
+                        UserNotification(
+                            user=user_mentioned,
+                            text=f"*@{self.current_user.username}* mentioned you in a comment on *{shift_id}*",
+                            notification_type="mention",
+                            url="/shifts",
+                        )
+                    )
```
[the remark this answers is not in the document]
Theodlz: Agreed, just the URL should be changed. You could first have your set of if/else/elif to set the url, then add the notifications.

static/js/components/CommentList.jsx
Theodlz: Yeah, it would be nicer to have one component, to which maybe you can pass the name and the onClick function. Maybe even the className.

static/js/components/ShiftPage.jsx
Theodlz: I can explain for Thomas. So basically, when a shift is modified, added, deleted... the list of shift is refreshed. Then, we need to see if the currentShift that has been set before has been updated or deleted. So those are the conditions to check if it was modified (if the users have the same ids, if they need a replacement when they didnt need one before, or simply if the `modified` date has been changed). I think this logic could be simplified and only look at the `modified` value, which should be the result of any modifications anyway.

static/js/components/CommentList.jsx
```diff
@@ -551,6 +629,9 @@ const CommentList = ({
       {permissions.indexOf("Comment") >= 0 && gcnEventID && (
         <CommentEntry addComment={addGcnEventComment} />
       )}
+      {permissions.indexOf("Comment") >= 0 && shift_id && (
+        <CommentEntry addComment={addShiftComment} />
+      )}
```
Theodlz, line 634: Here but also above for gcEvent comments and source comments, you check if the user has the permission and if there is an obj_id, shift_id or gcnEventID, but what you should also check (and it is the most important) is the associatedRessourceType, otherwise you end up having that component duplicated on the page.

Theodlz (changes requested):

skyportal/handlers/api/comment.py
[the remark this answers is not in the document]
thomasculino: I removed the duplicate logic by setting the url with if/else statements

guynir42 (commented):
I have some comments but in general this looks in very good shape. 

As always, take the time to play with this locally. Make sure the compact/regular comments render correctly. Add and delete comments and make sure they're updated in the right places, etc.

skyportal/handlers/api/comment.py
```diff
@@ -622,7 +704,7 @@ def delete(self, associated_resource_type, resource_id, comment_id):
 
         if isinstance(c, CommentOnGCN):
             gcnevent_dateobs = c.gcn.dateobs
-        else:
+        elif not isinstance(c, CommentOnShift):
```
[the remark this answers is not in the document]
thomasculino: In this case I want to make sure that the comment is not a CommentOnShift. Wouldn't this modification make the condition return true if the comment is either an instance of Comment or CommentOnShift? Should I take your suggestion but replacing CommentOnShift by CommentOnSpectrum?

static/js/components/RegularCommentList.jsx
```diff
@@ -0,0 +1,177 @@
...
+
+import UserAvatar from "./UserAvatar";
+
+import CommentAttachmentPreview from "./CommentAttachmentPreview";
+import DeleteComment from "./DeleteComment";
+
+dayjs.extend(relativeTime);
+dayjs.extend(utc);
+
+const CompactCommentList = ({
```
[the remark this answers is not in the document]
thomasculino: True this was a mistake and yes I checked, the compact and regular comments render correctly as you can see below:

regular compact_comments_on_shift

skyportal/handlers/api/comment.py
```diff
@@ -346,18 +363,57 @@ def post(self, associated_resource_type, resource_id):
...
+                author=author,
+                groups=groups,
+                bot=is_bot_request,
+            )
         else:
             return self.error(f'Unknown resource type "{associated_resource_type}".')
 
         users_mentioned_in_comment = users_mentioned(comment_text)
+        username = (
+            self.current_user.created_by.username
```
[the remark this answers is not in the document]
thomasculino: That was really helpful, thanks!

skyportal/handlers/api/comment.py
```diff
@@ -327,6 +343,7 @@ def post(self, associated_resource_type, resource_id):
                 bot=is_bot_request,
                 obj_id=spectrum.obj_id,
             )
+            obj_id = spectrum_id
```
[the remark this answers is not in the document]
thomasculino: True. It was confusing as well so I removed it. It was used below in the code. Because I removed the line, I added this a few lines below  
```
elif associated_resource_type.lower() == "spectra":
            text_to_send = f"*@{self.associated_user_object.username}* mentioned you in a comment on *{spectrum_id}*"
            url_endpoint = f"/source/{spectrum_id}"
```

static/js/components/ShiftPage.jsx
```diff
@@ -124,6 +130,9 @@ const ShiftPage = ({ route }) => {
               <CurrentShiftMenu currentShift={currentShift} />
             ) : null)}
         </Paper>
+        <Paper elevation={1}>
+          <CommentOnShift />
+        </Paper>
```
Theodlz, line 135:
```suggestion
        <Paper elevation={1}>
        {(shiftList && !show && currentShift ? (
          <CommentOnShift />
          ) : null)}
        </Paper>
```

Theodlz: Also above for the `<CurrentShiftMenu/>`, in the if statement the `currentShift` is duplicated, could you also remove that duplicate? Thanks.
You can simply copy and paste the if statement given here in the suggestion.

Theodlz (changes requested):
Looks good. Just one change that is necessary, because right now when clicking on add new shift to show the NewShift form, the comments should disappear along with the ShiftManagementMenu, but they are still displayed.

static/js/components/ShiftPage.jsx
```diff
@@ -69,13 +69,13 @@ const ShiftPage = ({ route }) => {
             let usersHaveChanged = false;
             // check if the users have the same ids, or if they need a replacement when they didnt need one before, and vice versa
             for (let i = 0; i < updatedShift.shift_users.length; i += 1) {
+              const current_user = currentShift.shift_users[i];
+              const updated_user = updatedShift.shift_users[i];
```
Theodlz, line 73: A bit everywhere on skyportal both backend and frontend we have a currentUser object. It's a bit confusing to use that name here. You could mistake it for the currentUser instead of the currentshiftuser. 
i would just call it `shift_user` and `updated_shift_user`, or simpler and maybe clearer: `previous_shift_user` / `new_shift_user`, or `old_shift_user` / `new_shift_user`.

Theodlz (approved):
Looks good!

static/js/components/ShiftPage.jsx
thomasculino: I chose `old_shift_user` / `new_shift_user`

guynir42 (approved):

static/js/components/DeleteComment.jsx
```diff
@@ -0,0 +1,139 @@
...
+          }
+          size="small"
+          color="primary"
+          type="button"
+          name={`deleteCommentButtonGcnEvent${id}`}
+          onClick={() => deleteCommentOnGcnEvent(gcnEventID, id)}
+          className="commentDelete"
+        >
+          <CloseIcon fontSize="small" />
+        </Button>
```
jadalilleboe, line 63: With these buttons, can you set the onClick function and name using if/elseif/else? And also convert the inline style to JSS?

jadalilleboe (approved):
Thank you for refactoring @thomasculino, this is really nice

jadalilleboe:
Also I think some of the imports will have to be changed with #3051 in

Theodlz:
> Also I think some of the imports will have to be changed with #3051 in

Yep. The page is pretty much broken. Everything is out of place if you load the shift page first. I updated the imports too in #3066. You can copy and paste what i changed if needed.

PR #3001, opened 2022-05-25
WIP: Shift summary

Theodlz:
#### This is a work in progress, so any suggestions are more than welcome !
This PR adds a **shift summary** element at the bottom of the shift page. Whenever a user clicks on a shift, he can see the gcn events that occurred during this shift, as well as the sources that are in the localizations associated with the event.
This PR adds a **2nd version** of that same feature, where a user, instead of having the summary of just one shift by clicking it, can select a time period (start date and end date) in a form, to get a summary of the shifts during that period. Here, not just the gcn events will be listed, but also the shifts.

#### This PR adds both backend and frontend. 

Moreover, the backend recovers more elements than those currently used on the frontend (it also recovers the comments and flags which one have been made by shift users), with the goal of using them with the summary frontend component later on, but also providing more detail if used through the API (if you need the comments to know you participated the observations of the event to create a circular for example).

### PREVIEW:
shift_summary


Theodlz:
> @Theodlz This is a good start. I am a bit worried about the speed with the nested for loops and whether this will return within the given time. But maybe it's very fast?

Thanks a lot for the review @mcoughlin . Basically, what i'm trying to do (in order) is:

- Option 1 : Given a specific shift_id:

1. Find the shift
2. Find the gcn events that occured during that shift
3. For each events i also get the comments, and i iterate over them to flag with one were made by shift users, and which were not.
4. Find how many sources are inside of this event's localization (with a last observed within 7 days of the observation).
5. return a report with the shift, the gcn events.

- Option 2: Given a start and end time (time frame):

1. Find the shifts between that start and end date
2. Find the gcn events between that start and end date
3. Iterate over the shifts and find only the gcn events that occured during these shifts
4. For each events i also get the comments, and i iterate over them to flag with one were made by shift users, and which were not.
5. Find how many sources are inside of this event's localization (with a last observed within 7 days of the observation).
6. return a report with the shift, the gcn events.

Option 2 is the one that will generate more loops as there are more shifts. Now i guess that what i could do is change the query that gets the gcn event to keep only those that happened during the shifts that happened between the given start and end date.

Now the questions i'm asking myself but have too little experience with the db in production to answer is:

- What is more likely to take way more ressources from the app and cause issues for all users and not just the one using the feature, expensive for loops or expensive queries ?
- During the GRANDMA GRB campaign, not many events occured, so the feature as it is would work fine in that case. But for O4, how big the difference in number of events will be, and how badly can that affect the feature / the app / the DB ?

You have more knowledge than me on all this, so maybe you can help me figure it out ? Or maybe DB experts can give their opinion too ?

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +522,237 @@ def delete(self, shift_id, user_id):
...
+            .order_by(GcnEvent.dateobs.asc())
+            .all()
+        )
+        gcn_added_during_shifts = []
+        # get the gcns added during the shifts
+        if len(gcns) > 0:
+            for shift in shifts:
+                # get list of user_id who are shift_users
+                for gcn in gcns:
+                    gcn = gcn.to_dict()
```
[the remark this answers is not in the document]
Theodlz: All i know is a new localization means a new localization in which we look for objects in the Source Handler to recover the sources in a gcn event, but on that query i am performing, how can that change it ? Is the row updated / deleted / replaced when a new localization is added ?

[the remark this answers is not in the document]
mcoughlin: It won't affect them at all, it will just increase the number of localizations with that relationship. We should maybe consider doing this update on Localization rather than GcnEvent maybe.

mcoughlin:
@Theodlz O4 will have maybe a GcnEvent per day (see Table 2 of https://arxiv.org/pdf/2108.07277.pdf). I am wondering if instead of doing one big shift level query, we send off a query for each GcnEvent that can be opened in a similar conceptual way to like what happens with the FollowupRequests where the tables are separated by Telescope

skyportal/handlers/api/shift.py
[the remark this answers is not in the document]
Theodlz: Definitely. I suppose that each new localization would be more precise than the previous one ? In that case it could just be replaced ? Or maybe we keep only the n more recent localization, or define metrics to select the *best* one ? I dont know if there is any known way to do all this, you tell me !

mcoughlin:
@Theodlz We should default to the most recent one (which is nominally best), but need to be flexible to use others as sometimes it isn't as simple as most recent is best.

Theodlz:
> @Theodlz We should default to the most recent one (which is nominally best), but need to be flexible to use others as sometimes it isn't as simple as most recent is best.

Maybe on the shift page using the last one makes sense, but on the GCN page defaulting to the last one + allowing the user to select between different localizations to use to retrieve the sources

Theodlz:
> > @Theodlz We should default to the most recent one (which is nominally best), but need to be flexible to use others as sometimes it isn't as simple as most recent is best.
> 
> Maybe on the shift page using the last one makes sense, but on the GCN page defaulting to the last one + allowing the user to select between different localizations to use to retrieve the sources would be the better option.

Theodlz:
> > @Theodlz We should default to the most recent one (which is nominally best), but need to be flexible to use others as sometimes it isn't as simple as most recent is best.
> 
> Maybe on the shift page using the last one makes sense, but on the GCN page defaulting to the last one + allowing the user to select between different localizations to use to retrieve the sources would be the better option @mcoughlin

Theodlz:
But in general @mcoughlin , the gcn page needs some refactoring. The display is pretth sketchy for some elements, and a better and simpler UI would allow us to add these kind of changes

Theodlz:
Yep you can already choose a localization, but i dont think it impacts the menu that has the list of sources

mcoughlin:
@Theodlz ah yeah that would be good to include of course.

Theodlz:
> @Theodlz We should default to the most recent one (which is nominally best), but need to be flexible to use others as sometimes it isn't as simple as most recent is best.

Little update on this @mcoughlin. I had a deeper look at the source handler, and it seems that we are already keeping just one localization with the `.first()` method that keeps only the first row. Right now you can provide a `localizationDateObs` and/or a `localizationName`. If you provide a name, it will find the localization with that name, otherwise use the first row it finds when looking for localizations with the `localizationDateObs`. I guess what i could add before using the `first` row is sort by descending `created_at`, so it defaults to the most recent one. And on the frontend pass a localization Name to the function that calls that handler if the user selected one in the form.

mcoughlin:
@Theodlz Great suggestion! Maybe open an issue unless you plan to just knock it out?

Theodlz:
> @Theodlz Great suggestion! Maybe open an issue unless you plan to just knock it out?

I'll take care of it here. Is the `created_at` the only thing we care about though ? Maybe the `modified_at` should actually matter more ? I don't know. Do we expect people to update a given localization or simply post a new one ?

mcoughlin:
@Theodlz modified_at is maybe safest, but hard to imagine its updated outside of some replacement.

Theodlz:
Considering where this PR is going ? (fixing issues with the gcn page itself and such), i'm thinking that we could start with a lighter version of the ShiftSummary handler (no comments, no count query for the sources) to have something shorter to review, and maybe add more stuff in another PR later ?

skyportal/handlers/api/shift.py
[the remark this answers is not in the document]
mcoughlin: For now, we will just sort by modified_at and roll with that

static/js/components/GcnEventPage.jsx
```diff
@@ -70,11 +70,14 @@ const useStyles = makeStyles((theme) => ({
   columnItem: {
     marginBottom: theme.spacing(2),
   },
-  source: {
+  noSource: {
```
mcoughlin, line 73: What's with this name?!

Theodlz: This is the class for an element that is displayed only when there are no sources to tell the user, seemed more logic that its called noSource then source no ?

skyportal/handlers/api/shift.py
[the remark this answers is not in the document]
Theodlz: Done

static/js/components/GcnEventPage.jsx
Theodlz: the sorting by `modified` for the localization is already done @mcoughlin

mcoughlin: Ah yeah that makes sense!

Theodlz:
> For now, we will just sort by modified_at and roll with that

Hey @mcoughlin, already did that (it's in the source handler, where I select the latest localization by `modified_at`).

Theodlz:
Hi @mcoughlin, should I tag Guy for review on this one? Or do you want me to modify some more stuff?

mcoughlin:
@Theodlz yes go ahead @guynir42

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
             action='skyportal/REFRESH_SHIFTS', payload={'shift_id': int(shift_id)}
         )
         return self.success()
+
+
+class ShiftSummary(BaseHandler):
+    # this handler has a get method that returns a summary of all the activity of shift users on skyportal for a given period
+    # it is used to generate a report
```
guynir42, line 524:
```suggestion
    """
    This handler has a get method that returns a summary 
    of all the activity of shift users on skyportal for a given period.
    It is used to generate a report.
    """
```

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
...
+        ---
+        description: Get a summary of all the activity of shift users on skyportal for a given period
+        tags:
+          - shifts
+          - users
+          - sources
+          - gcn
+        parameters:
+          - in: path
+            name: start_date
```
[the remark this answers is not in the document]
guynir42: Or maybe these two should be query arguments?

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
...
+
+        start_date = self.get_argument("start_date", None)
+        end_date = self.get_argument("end_date", None)
+        if (start_date is None or end_date is None) and shift_id is None:
+            return self.error("Please provide start_date and end_date, or shift_id")
+
+        if start_date is not None and end_date is not None:
+            try:
+                start_date = datetime.datetime.strptime(start_date, "%Y-%m-%dT%H:%M:%S")
+                end_date = datetime.datetime.strptime(end_date, "%Y-%m-%dT%H:%M:%S")
```
guynir42, line 564: I would prefer to use `arrow` to parse the strings (also mention that in the docstring, that these are "arrow parsable strings"). This is more flexible and more consistent with what we usually use.

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
...
+            except ValueError:
+                return self.error("Please provide valid start_date and end_date")
+            if start_date > end_date:
+                return self.error("Please provide start_date < end_date")
+            if start_date.date() > datetime.datetime.today().date():
```
guynir42, line 569: Why do we need `.date()` and not just comparing the datetime objects directly?

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
...
+                return self.error("Please provide start_date < today")
+            # if there is more than 4 weeks, we return an error
+            if (end_date - start_date).days > 28:
+                return self.error("Please provide a period of less than 4 weeks")
```
guynir42, line 573: why do we need this limit?

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
...
+            for shift in shifts:
+                # get list of user_id who are shift_users
+                for gcn in gcns:
+                    gcn = gcn.to_dict()
+                    if (
+                        gcn["dateobs"] >= shift["start_date"]
+                        and gcn["dateobs"] <= shift["end_date"]
+                    ):
+                        if gcn["id"] not in [
+                            gcn["id"] for gcn in gcn_added_during_shifts
```
guynir42, line 646: I'd prefer the inner variable to have a different name in the list comprehension. I would even use `g` if it is local to that one line.

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
...
+            .filter(GcnEvent.dateobs >= start_date)
+            .filter(GcnEvent.dateobs <= end_date)
+            .order_by(GcnEvent.dateobs.asc())
+            .all()
+        )
+        gcn_added_during_shifts = []
+        # get the gcns added during the shifts
+        if len(gcns) > 0:
+            for shift in shifts:
+                # get list of user_id who are shift_users
```
guynir42, line 638: Is this comment in place?

skyportal/handlers/api/shift.py
```diff
@@ -515,3 +517,144 @@ def delete(self, shift_id, user_id):
...
+                        ]:
+                            gcn['shift_ids'] = [shift['id']]
+                            gcn_added_during_shifts.append(gcn)
+
+                        else:
+                            for gcn_added in gcn_added_during_shifts:
+                                if gcn_added['id'] == gcn['id']:
+                                    gcn_added['shift_ids'].append(shift['id'])
+                                    break
+            report['gcns'] = {'total': len(gcn_added_during_shifts)}
```
guynir42, line 656:
```suggestion
            report['gcns'] = {'totalMatches': len(gcn_added_during_shifts)}
```

skyportal/tests/api/test_shifts.py
```diff
@@ -39,3 +41,129 @@ def test_shift(public_group, super_admin_token, view_only_token, super_admin_use
...
+
+def test_shift_summary(
+    public_group,
+    super_admin_token,
+    super_admin_user,
+    upload_data_token,
+    view_only_token,
+    ztf_camera,
+):
+    # add a shift to the group, with a start day one day before today, and an end day one day after today
```
guynir42, line 54:
```suggestion
    # add a shift to the group, with a start day one day before today, 
    # and an end day one day after today
```

skyportal/tests/api/test_shifts.py
```diff
@@ -39,3 +41,129 @@ def test_shift(public_group, super_admin_token, view_only_token, super_admin_use
...
+            'flux': 12.24,
+            'fluxerr': 0.031,
+            'zp': 25.0,
+            'magsys': 'ab',
+            'filter': 'ztfg',
+        },
+        token=upload_data_token,
+    )
+    assert status == 200
+    assert data['status'] == 'success'
```
guynir42, line 140: What do these data do in this test?

skyportal/tests/frontend/test_shifts.py
```diff
@@ -287,3 +284,159 @@ def test_shift(
...
+    with open(datafile, 'rb') as fid:
+        payload = fid.read()
+    data = {'xml': payload}
+
+    status, data = api('POST', 'gcn_event', data=data, token=super_admin_token)
+    assert status == 200
+    assert data['status'] == 'success'
+
+    # wait for event to load
+    time.sleep(15)
```
guynir42, line 349: Here too.

static/js/components/GcnEventPage.jsx
guynir42: Perhaps `noSources` make it a tiny bit more obvious?

skyportal/handlers/api/shift.py
Theodlz: Doesnt pyarrow generate offset aware dates ? Couldnt that be a problem ? We pass UTC dates as query arguments, but I wonder if they are put back as local but not the right one as it creates a mess just like in the frontend. I'll investigate a bit.

skyportal/handlers/api/shift.py
Theodlz: Pyarrow provides offset aware dates and datetime non offset aware dates, so you cant compare them without using `.date()`

skyportal/tests/api/test_shifts.py
Theodlz: We were checking if we managed to correctly recover the sources that are inside of a gcn event. But as I removed that from the handler and only do it on the frontend I can just delete it !

skyportal/handlers/api/shift.py
Theodlz: I replaced the right part of the statement by its equivalent using arrow

skyportal/tests/api/test_shifts.py
```diff
@@ -39,3 +41,129 @@ def test_shift(public_group, super_admin_token, view_only_token, super_admin_use
...
+    with open(datafile, 'rb') as fid:
+        payload = fid.read()
+    data = {'xml': payload}
+
+    status, data = api('POST', 'gcn_event', data=data, token=super_admin_token)
+    assert status == 200
+    assert data['status'] == 'success'
+
+    # wait for event to load
+    time.sleep(15)
```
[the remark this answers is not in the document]
Theodlz: Done !

Theodlz:
Hey @guynir42, I left some comments answering some stuff from your last review and committed the changes you asked for. Also added some titles and tooltips to the frontend to make it more obvious

skyportal/handlers/api/shift.py
guynir42: I just used to see `arrow` everywhere in our code. Maybe that is indeed cause for problems. 
If you think we should be moving in the direction of ISO style dates then just leave it like that.

skyportal/tests/api/test_shifts.py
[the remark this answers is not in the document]
guynir42: I would do a loop to try the API, check for success, and if the loop times out then raise an error. 

If you do a loop it can be 30 seconds or more, as long as most of the time the test runs in only a few seconds. 

We should also be working on smaller datasets, as we discussed last week with @mcoughlin

skyportal/tests/frontend/test_shifts.py
```diff
@@ -346,7 +347,39 @@ def test_shift_summary(
     assert data['status'] == 'success'
 
     # wait for event to load
-    time.sleep(15)
+    n_times = 0
+    while n_times < 30:
```
guynir42, line 351: You know you can just do `for n_times in range(30)`?

guynir42 (approved):
Looks good. 

I might consider changing the `while` loops for `for` loops, but that's just my preference. 

Also, when you are addressing any previous comments mark them as resolved so I can assume you've just implemented the suggestion. Thanks!

skyportal/tests/frontend/test_shifts.py
Theodlz: You're right ! Avoids having to declare and increment the variable. I'll do that then. When all the tests are green and the for loop is implemented, can I merge it or should i ask for you or Michael (a reviewer) to make that decision ? Idk if its ok for the person that created the PR to merge or if it should always be those giving a review.

Theodlz: Done

Theodlz:
> Looks good.
> 
> I might consider changing the `while` loops for `for` loops, but that's just my preference.
> 
> Also, when you are addressing any previous comments mark them as resolved so I can assume you've just implemented the suggestion. Thanks!s
Sorry ! For some reason for one of your suggestionq I can't mark as resolved even if I made the change... For others, as I left some comments after yours I was unsure that I could just mark it as resolved.

Anyway, this is ready for review I think ! All changes have been made.

mcoughlin (approved):
@Theodlz just pulled down the branch. the interface is a nice start, and thanks for the GcnEventPage stuff. Will continue to be good to work on this in future PRs.

PR #3003, opened 2022-05-26
New user (on login or using the API) - Default Groups, ACLs and Role

Theodlz:
Currently, in the user handler, when posting a new user, he is added automatically to the `Sitewide Group` and has the `Full user` role. However, none of this works when a new user is created when login in on the frontend. When a new user joins SkyPortal, the UserHandler is not used, this is where it is done instead: https://github.com/skyportal/skyportal/blob/main/skyportal/onboarding.py#L24-L76
In `onboarding.py`, the user has no role given if he hasn't been invited first, has no groups by default not even sitewide, and has no particular ACLs (except the one associated with a role if he has been invited).

This PR lets us set `default_groups`, `default_acls`, and `default_role` in the `config.yaml` file, and use that new configuration with 3 new methods to set all of these for a given user.
In both `onboarding.py` and `skyportal/handler/api/user.py`, those methods are called to set default groups, roles, and ACL for new users, both when login in for the first time on the frontend, or adding them using the API.


skyportal/handlers/api/user.py
```diff
@@ -19,17 +19,67 @@
...
+                )
+        acls = ACL.query.filter(ACL.id.in_(cfg['default_acls'])).all()
+        DBSession().add_all([UserACL(user_id=user.id, acl_id=acl.id) for acl in acls])
+
+
+def set_default_group(user):
+    default_groups = []
+    if cfg['misc']['public_group_name'] is not None:
+        default_groups.append(cfg['misc']['public_group_name'])
+    if cfg['default_groups'] is not None and isinstance(cfg['default_groups'], list):
```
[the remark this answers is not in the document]
Theodlz: Great! Should we merge the baselayer PR so I can use it then ?

skyportal/handlers/api/user.py
```diff
@@ -45,23 +95,32 @@ def add_user_and_setup_groups(
...
-    if public_group is not None:
-        DBSession().add(GroupUser(group_id=public_group.id, user_id=user.id))
+    if roles == []:
+        set_default_role(user)
+    if group_ids_and_admin == []:
+        set_default_group(user)
+    else:
+        # Add user to specified groups & associated streams
+        for group_id, admin in group_ids_and_admin:
+            DBSession().add(GroupUser(user_id=user.id, group_id=group_id, admin=admin))
```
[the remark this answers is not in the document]
Theodlz: Done @stefanv! Had a bit of trouble getting it to work, but it does work well now, I'll let you have a look!

stefanv:
I think https://github.com/cesium-ml/baselayer/pull/319 is almost in; you can temporarily pin to that branch here if you want to start using the new syntax.  I think it will make your PR a lot cleaner.

Theodlz:
Hi @stefanv, as mentioned in the issue I opened in baselayer, and as @guynir42 said too (we encountered similar problems), baselayer or SKyPortal needs some more work before the last version of baselayer can be pinned. Therefore, I can't use the new syntax. Should we wait for it to work, or should we consider merging this or making some modifications if needed?

skyportal/handlers/api/user.py
```diff
@@ -19,50 +20,122 @@
...
-        oauth_uid=oauth_uid,
-        expiration_date=expiration_date,
-    )
-    DBSession().add(user)
-    DBSession().flush()
+    with DBSession() as session:
+        try:
+            session.execute(
+                f'LOCK TABLE {User.__tablename__} IN SHARE ROW EXCLUSIVE MODE'
+            )
```
guynir42, line 94: If there's an error does this lock ever get released?

skyportal/handlers/api/user.py
```diff
@@ -163,6 +236,7 @@ def get(self, user_id=None):
                     allOf:
                       - $ref: '#/components/schemas/Success'
                       - type: object
+                        #
```
guynir42, line 239:
```suggestion
```

skyportal/onboarding.py
```diff
@@ -18,68 +23,88 @@
...
-
-        if existing_user is None and invite_token is None:
-            raise Exception(
-                "Authentication Error: Missing invite token. A valid invite token is required."
-            )
-        elif existing_user is not None:
-            return {"is_new": False, "user": existing_user}
-
+    with DBSession() as session:
         try:
```
guynir42, line 27: I think that if you use `with DBSession() as session:` we don't need to rollback manually in case of an exception.

skyportal/onboarding.py
```diff
@@ -18,68 +23,88 @@
...
+                }
+                user = strategy.create_user(**fields, **{"oauth_uid": uid})
+
+                set_default_role(user, session)
+                set_default_acls(user, session)
+                set_default_group(user, session)
+
+                session.commit()
+                return {"is_new": True, "user": user}
+            elif existing_user is not None:
```
guynir42, line 101: Is there even a way to reach these `elif`s?

skyportal/handlers/api/user.py
```diff
@@ -19,50 +20,122 @@
...
-    )
-    DBSession().add(user)
-    DBSession().flush()
+    with DBSession() as session:
+        try:
+            session.execute(
+                f'LOCK TABLE {User.__tablename__} IN SHARE ROW EXCLUSIVE MODE'
+            )
+            user = User(
+                username=username.lower(),
```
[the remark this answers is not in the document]
Theodlz: That's how it is done on skyportal/main, so I left it like this.

skyportal/handlers/api/user.py
Theodlz: I followed the same logic as what I found in other handlers. But I'm thinking that this lock might not be necessary in the first place don't you think ?

skyportal/handlers/api/user.py
```diff
@@ -19,50 +20,122 @@
...
-        Group.name == cfg["misc"]["public_group_name"]
-    ).first()
-    if public_group is not None:
-        DBSession().add(GroupUser(group_id=public_group.id, user_id=user.id))
-    return user.id
+            set_default_acls(user, session)
+            session.commit()
+        except Exception as e:
+            session.rollback()
+            return log(e.args[0])
```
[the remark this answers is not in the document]
Theodlz: So i could still do the loggin, but raise the expectation instead of return the raise exception, or simply raise the exception and return nothing ?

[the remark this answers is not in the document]
guynir42: I think you should log, and then either raise an exception or return something, depending on what the calling function would expect.

skyportal/onboarding.py
```diff
@@ -18,68 +23,88 @@
...
+                # No matching user exists; create a new user
+                fields = {
+                    name: kwargs.get(name, details.get(name))
+                    for name in backend.setting("USER_FIELDS", USER_FIELDS)
+                }
+                user = strategy.create_user(**fields, **{"oauth_uid": uid})
+
+                set_default_role(user, session)
+                set_default_acls(user, session)
+                set_default_group(user, session)
```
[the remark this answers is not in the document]
Theodlz: You mean in general in this PR, or just when `onboarding` runs so when a user connects to the platform on the frontend ? All I added is that possibility of setting default groups, acls, and role to a user you add. But it should not change the rest of the code that was already there. When posting through the API, if you specify roles and groups, those will be used over the default ones. When logging in on the frontend: if you received an invitation link, currently on mai it gives you the specific roles that were specified in your invitation, and that's what it still does + the default groups and acls now. When logging in without an invitation link, it gives you default groups, role and acls.

I guess your question here is more through the API ? Well you can just use it in the same way you used to before. Those default values are used if nothing is specified basically. Except the default acls that will always be applied (as no acl can be specified in the request to create a user through the API).

Really the goal of this feature is to allow someone that deploys skyportal to be sure that users are in the sitewide group for example, and that they are full users with a certain acl if needed. If you have some settings that you want all users to have, you can avoid setting them manually for everyone.

skyportal/handlers/api/user.py
Theodlz: I took the syntax of the photometry handler (that uses with DBSession and stuff) as a model and thats where i saw it and thought it might be necessary. Ill remove it then !

skyportal/handlers/api/user.py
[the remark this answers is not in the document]
Theodlz: Usually (i mean if it works and there's not exception)  the function returns the user_id. Maybe I should just return None, and in the function that calls this one, if the user_id is none, I return a self.error ? But then if I just return none I don't know exactly what went wrong. I'm not sure what is the best way to handle all that.

skyportal/onboarding.py
Theodlz: Noted. I'll remove those then!

skyportal/onboarding.py
```diff
@@ -18,68 +23,88 @@
...
+                    raise Exception(
+                        "Authentication Error: Missing invite token. A valid invite token is required."
+                    )
+                elif existing_user is not None:
+                    return {"is_new": False, "user": existing_user}
+
+                try:
+                    n_days = int(cfg["invitations.days_until_expiry"])
+                except ValueError:
+                    raise ValueError(
```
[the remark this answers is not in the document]
Theodlz: You mean just replacing the ValueError by a TypeError ?

guynir42:
@Theodlz if it is not a lot of places I would just use a `or` statement or some other way to get things done. We can always clean that up later.

Theodlz:
> @Theodlz if it is not a lot of places I would just use a `or` statement or some other way to get things done. We can always clean that up later.

That's exactly what I did already! The new syntax was meant to replace them. So if we are not using it in this PR, I think that this is done!

config.yaml.defaults
```diff
@@ -534,6 +534,11 @@ cross_match:
 #    transmission: [0, 1, 1, 0]
 #    description: "basic ATLAS O filter - tophat transmission"
 
+user:
+  default_role: 'Full user'
+  default_groups: ['Grandma']
```
mcoughlin, line 539: Let's choose something less collaboration specific.

skyportal/handlers/api/user.py
```diff
@@ -19,50 +20,119 @@
     Stream,
 )
 
+from skyportal.model_util import role_acls, all_acl_ids
 
+log = make_log("api/user")
 env, cfg = load_env()
 
 
+def set_default_role(user, session):
```
mcoughlin, line 29: Can we have a few short doc strings? I think it might be worth calling out that you need to commit outside of the function.

mcoughlin (approved):
A few small things. otherwise seems like we can get this in.

config.yaml.defaults
Theodlz: Bu default in skyportal there is no other group than the public group "Sitewide group" to which users are added by default thanks to this PR. So in fact i think that the best option is to leave that empty.

skyportal/handlers/api/user.py
Theodlz: Sure thing ! Will add that later today

mcoughlin (approved):
@Theodlz Thanks for fixing that.

skyportal/handlers/api/user.py
```diff
@@ -19,50 +20,131 @@
...
+    '''
+    Set the default acls for a user. The default acls can be set in the config file.
+    This method does not commit the session, so the session needs to be commited after calling this method.
+    '''
+    if cfg['user.default_acls'] is not None:
+        for acl_id in cfg['user.default_acls']:
+            if acl_id not in all_acl_ids:
+                log(
+                    f"Invalid default_acls configuration value: {acl_id} does not exist"
+                )
```
guynir42, line 58: In all these cases why do we want to log this instead of returning an error?

guynir42 (approved):
Looks good. I wonder if we should not raise an exception when one of the groups/roles/acls in the config do not exist in the DB. That should probably be a big enough problem to stop whatever is happening and require a fix before adding more users.

skyportal/handlers/api/user.py
Theodlz: For role and acls i agree that we could raise an error. But for the groups i'm not so sure.
If you do that for example, lets say you add 'Program A' as the default group, if you run `make load_demo_data`, as users are being posted before the groups, you'll just get an error and won't be able to load anything.

Theodlz: Maybe @guynir i can do something like: if the group doesnt exist, i check if there are groups in the db other than sitewide, if there is none (meaning that no groups have been loaded yet), we just log, but if groups have been loaded and the groups in the config don't exist, we raise an error ?

guynir42: Do you expect to run into this situation where the groups aren't loaded yet? It seems like this should never happen...

Theodlz: What about the example i gave before with the demo data ? In the first place, that is why i used logs instead of raising an error, as i couldnt load users because of it.

guynir42: I see. Are the users loaded before the groups? That doesn't sound like the right order. Is there any reason we don't load groups first?

Theodlz: Yes. We load the groups after users so when creating the groups we can pass user ids in the field that allows to set group admins at the same time that we create the group. 
Maybe we can create the groups, then the users, and then set the group users ?

guynir42: I don't know if that would be so simple to do. Maybe give the demo data explicit group lists for each user, so that the handler doesn't use the defaults at all?

Theodlz: Good idea, if setting the group users separately can't be done, i can simply give them the Sitewide Group explicitly so the default group part of the code does not run.

guynir42: That seems like the easiest way to go. And in general if there are non-existent groups in the defaults it should return an error. Make sure to catch those errors and turn them into `self.error` on the handler.

Theodlz:
So here is what i've done @guynir42 :

guynir42:
I don't know about the cesium user, but maybe some stuff gets posted before the error is raised in the handler? 

As for the default group in the demo data: whatever you don't need, remove. We have enough complexity.

Theodlz:
> I don't know about the cesium user, but maybe some stuff gets posted before the error is raised in the handler?
> 
> As for the default group in the demo data: whatever you don't need, remove. We have enough complexity.

Deleted :+1:  Just waiting for @stefanv input on the testuser thingy.

stefanv:
fakeoauth always authenticates, but it's not too tricky to get auth credentials for free under your personal account. Try it out as per the comments in the config file, and let's update the docs if it's no sufficient.

Theodlz:
> fakeoauth always authenticates, but it's not too tricky to get auth credentials for free under your personal account. Try it out as per the comments in the config file, and let's update the docs if it's no sufficient.

Hey Stefan, i tried that an no issue with the authentication using Google. All seems to work just fine for me !

skyportal/onboarding.py
```diff
@@ -91,14 +90,14 @@ def create_user(strategy, details, backend, uid, user=None, *args, **kwargs):
                     for name in backend.setting("USER_FIELDS", USER_FIELDS)
                 }
                 user = strategy.create_user(**fields, **{"oauth_uid": uid})
-
                 set_default_role(user, session)
                 set_default_acls(user, session)
                 set_default_group(user, session)
-
+                print('this should not run if there is an error')
```
guynir42, line 96:
```suggestion

```

guynir42 (approved):

stefanv (approved):

skyportal/onboarding.py
Theodlz: Oops. Thanks !

PR #3009, opened 2022-05-26
Refactor GcnEvent Handler

mcoughlin:
This PR refactors the GcnEvent handler so we can use it in the gcn_service.

### DIFF
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -34,9 +34,9 @@ Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"])
 )
 def handle(payload, root):
 
-    session = Session()
     user_id = 1
-    post_gcnevent(payload, user_id, session)
+    with Session() as session:
+        post_gcnevent(payload, user_id, session)
 
 
 if __name__ == "__main__":
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -5,9 +5,7 @@ from baselayer.app.models import init_db
 from baselayer.app.env import load_env
 
 from skyportal.handlers.api.gcn import post_gcnevent
-from skyportal.models import (
-    DBSession,
-)
+from skyportal.models import DBSession
 
 env, cfg = load_env()
 
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -36,7 +36,7 @@ def handle(payload, root):
 
     session = Session()
     user_id = 1
-    post_gcnevent(payload, user_id, session=session)
+    post_gcnevent(payload, user_id, session)
 
 
 if __name__ == "__main__":
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -1,23 +1,8 @@
-import yaml
-
 import gcn
 
 from baselayer.log import make_log
 
-from skyportal.tests import api
-
-
-def get_token():
-    try:
-        token = yaml.load(open('.tokens.yaml'), Loader=yaml.Loader)['INITIAL_ADMIN']
-        print('Token loaded from `.tokens.yaml`')
-        return token
-    except (FileNotFoundError, TypeError, KeyError):
-        print('Error: no token specified, and no suitable token found in .tokens.yaml')
-        return None
-
-
-admin_token = get_token()
+from skyportal.handlers.api.gcn import post_gcnevent
 
 
 @gcn.include_notice_types(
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/gcn.py
```diff
@@ -33,20 +33,25 @@
 Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
 
 
-def post_gcnevent(payload, user_id):
+def post_gcnevent(payload, user_id, session=Session()):
```
stefanv, line 36:
```suggestion
def post_gcnevent(payload, user_id, session):
```

I don't think a session should be instantiated like that.

mcoughlin: Yes good point. Fixed.

services/gcn_service/gcn_service.py
```diff
@@ -1,23 +1,20 @@
...
-        return token
-    except (FileNotFoundError, TypeError, KeyError):
-        print('Error: no token specified, and no suitable token found in .tokens.yaml')
-        return None
+env, cfg = load_env()
 
+init_db(**cfg['database'])
 
-admin_token = get_token()
+Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
```
stefanv, line 17: Could probably use DBSession directly, curious why this is necessary.

services/gcn_service/gcn_service.py
```diff
@@ -36,9 +33,10 @@ def get_token():
     gcn.NoticeType.ICECUBE_ASTROTRACK_BRONZE,
 )
 def handle(payload, root):
-    response_status, data = api(
-        'POST', 'gcn_event', data={'xml': payload}, token=admin_token
-    )
+
+    session = Session()
```
stefanv, line 37: Use `with Session()` syntax?

skyportal/handlers/api/gcn.py
```diff
@@ -69,82 +162,10 @@ def post(self):
             return self.error("xml must be present in data to parse GcnEvent")
 
         payload = data['xml']
+        session = Session()
```
stefanv, line 165: with Session

services/gcn_service/gcn_service.py
mcoughlin: this was just copying from gcn.py, I removed.

services/gcn_service/gcn_service.py
mcoughlin: added.

skyportal/handlers/api/gcn.py
mcoughlin: added.

stefanv (approved):

services/gcn_service/gcn_service.py
```diff
@@ -1,23 +1,17 @@
...
-        token = yaml.load(open('.tokens.yaml'), Loader=yaml.Loader)['INITIAL_ADMIN']
-        print('Token loaded from `.tokens.yaml`')
-        return token
-    except (FileNotFoundError, TypeError, KeyError):
-        print('Error: no token specified, and no suitable token found in .tokens.yaml')
-        return None
+from skyportal.handlers.api.gcn import post_gcnevent
+from skyportal.models import (
+    DBSession,
+)
```
guynir42, line 10:
```suggestion
from skyportal.models import DBSession
```

skyportal/handlers/api/gcn.py
```diff
@@ -32,6 +33,98 @@
...
+            .filter_by(
+                dateobs=dateobs,
+                localization_name=skymap["localization_name"],
+            )
+            .one()
+        )
+    except NoResultFound:
+        localization = Localization(**skymap)
+        session.add(localization)
+        session.commit()
```
guynir42, line 118: The only place where you are committing is inside an except clause. That doesn't look right...

mcoughlin: This is just checking for a repeat of the localization.

guynir42: Yeah but you are adding and modifying some rows in this function. Are you sure those changes are getting committed at all?

mcoughlin: @guynir42 its behaved that way for a long time now. I tested it repeats the old behavior.

stefanv: Hrm, what needs to be clarified is if each of the following sometimes raises a NoResultsFound exception:

```python
         event = GcnEvent.query.filter_by(dateobs=dateobs).one()
```

and

```python
            Localization.query_records_accessible_by(
                user,
            )
            .filter_by(
                dateobs=dateobs,
                localization_name=skymap["localization_name"],
            )
            .one()
```

mcoughlin: Given that the demo data and the usual flow works, my intuition is this is all fine.

guynir42 (approved):

PR #3012, opened 2022-05-27
Photometry plotting preferences

### LINKED ISSUE #2788: Configurable photometry buttons
Make photometry add and clear buttons configurable in a way that lets you choose lists of filters, give them a common name, and have that turn up as buttons.

jadalilleboe:
coauthored by @thomasculino 

This PR adds photometry plotting preference options to the user settings. There are three main elements:

### Setting automatically visible photometry points
Allows the user to select which filters/origins they would like automatically visible on load of the photometry plot. Hides all other points, unless none of the user's selected filters/origins are on the plot, and in that case all points are visible. Example setting only ztfg visible:
Hnet-image (2)

### Photometry buttons
This lets the select a group of filters/origins, give them a common name, and a button will appear on the photometry plot for showing those filters/origins. The buttons do not hide any existing visible points on the plot, only show the selected ones. Example for adding buttons for showing ztfr & ztfi:
Hnet-image (3)

### Data point size
Saves the user's preferred data point size with the spinner on the plot.
Hnet-image (2)

## *other changes*
I propose moving the legend of the photometry plot to the bottom of the plot so all the interactions with toggling visiblity are in the same area.
Screen Shot 2022-05-27 at 8 27 18 AM

### Specific
- PhotometryPlottingPreferences.jsx - wrapper component for the three forms
-  SetAutomaticallyVisiblePhotometry.jsx - form for setting which photometry should automatically be visible on phot plot
-  PhotometryButtonsForm.jsx - form for adding photometry buttons

### Reusable
- FilterSelect.jsx - select component for all the filters
- OriginSelect.jsx - select component for origins
- SelectWithChips.jsx - general select component where the selected items are rendered as chips
- DeletableChips.jsx - component for rendering a grouping of deletable chips

Apologies for the large PR, it probably should've been split up but it ended up all just going together ...

closes #2788


skyportal/plot.py
```diff
@@ -433,104 +432,94 @@ def annotate_spec(plot, spectra, lower, upper):
...
     )
-    button.js_on_click(callback_clear_photometry)
+    button.js_on_click(callback_hide_photometry)
     return button
 
 
-def make_add_all_photometry_button(model_dict):
-    """Make a button to add photometry.
+def make_show_all_photometry_button(model_dict):
+    """Make a button to show all photometry on the photometry plot.
```
guynir42, line 483: I would add the `Parameters` heading in all the docstrings. And add an empty line before each heading.

skyportal/plot.py
```diff
@@ -433,104 +432,94 @@ def annotate_spec(plot, spectra, lower, upper):
...
-            make_clear_photometry_button(model_dict),
-        ],
-    )
+    buttons = [
+        make_show_all_photometry_button(model_dict),
+        make_hide_photometry_button(model_dict),
+    ]
+    if user.preferences and "photometryButtons" in user.preferences:
+        photometry_buttons = list(user.preferences["photometryButtons"].items())
+        for name, info in photometry_buttons:
```
guynir42, line 515:
```suggestion
        for name, info in user.preferences["photometryButtons"].items():
```

skyportal/plot.py
```diff
@@ -556,6 +545,45 @@ def add_axis_labels(plot, panel_name):
     plot.yaxis.axis_label = axis_label_dict[panel_name]['y']
 
 
+def check_visibility_on_phot_plot(user, show_all_filters, show_all_origins, label):
+    """Function to check if a photometry point should be displayed on the plot in accordance with the user's photometry plotting preferences.
```
guynir42, line 549:
```suggestion
    """Function to check if a photometry point should be displayed 
    on the plot in accordance with the user's photometry plotting preferences.
```

guynir42: I would also break up the long lines in the Parameters block.

static/js/components/ObservabilityPreferences.jsx
```diff
@@ -1,59 +1,38 @@
...
   };
 
   telescopeList?.sort((a, b) => (a.name < b.name ? -1 : 1));
-  const telescopeIDToName = { "-1": "Clear selections" };
   telescopeList?.forEach((telescope) => {
+    telescopeNametoID[telescope.name] = telescope.id;
     telescopeIDToName[telescope.id] = telescope.name;
   });
+  const telescopeNameList = telescopeList
+    .filter((telescope) => telescope.fixed_location)
```
guynir42, line 33: Why filter only on fixed telescopes?

static/js/components/OriginSelect.jsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import PropTypes from "prop-types";
+import SelectWithChips from "./SelectWithChips";
+
+const OriginSelect = ({ onOriginSelectChange, initValue, parent }) => {
+  const origins = ["Clear selections", "Muphoten", "STDpipe"];
```
guynir42, line 6: Are these default values?

static/js/components/SourceTable.jsx
```diff
@@ -426,7 +426,7 @@ const SourceTable = ({
             container
             direction="row"
             spacing={3}
-            justify="center"
+            justifyContent="center"
```
guynir42, line 429: Thanks! That was bugging me for a while...

guynir42 (commented):
The code looks good. Let me pull down to my local version and check that the UX makes sense before I give the :heavy_check_mark:

jadalilleboe:
I agree, it would be better UX to have it directly on the plot. I think that could be accomplished in this PR.

static/js/components/ObservabilityPreferences.jsx
jadalilleboe: It's what [was there previously.](https://github.com/skyportal/skyportal/blob/main/static/js/components/ObservabilityPreferences.jsx#L94) Do you think it should be changed?

guynir42: Wait I didn't notice this was in the context of observability. It makes sense now.

skyportal/plot.py
jadalilleboe: fixed

skyportal/plot.py
jadalilleboe: fixed

static/js/components/OriginSelect.jsx
jadalilleboe: @thomasculino were these placeholder values? and fetching all the origins is a WIP?

thomasculino: Yes, these are default values. We discussed that it could be a good idea to have default values for origin because fetching all the origins would mean fetching all photometry and then we will retrieve all unique origins. We would also add a field for users to add a custom origin if it is not in the list. This way, we can avoid fetching all photometry. Maybe we can find another way to do this but this was my reasoning.

jadalilleboe: @guynir42 thoughts?

guynir42: @jadalilleboe @thomasculino the `origin` column is indexed, so it seems to me it should not be terribly expensive to query the `distinct()` values even on all photometry points. You may want to be selective on the columns you are getting but it may not be so bad. 

Since this is a dropdown menu we don't expect people to use all the time, it would not be a big issue if it takes a couple of seconds to populate it. I suggest first of all trying it with a reasonable query and testing it out on production. If it turns out to be annoyingly slow, we can figure out something smarter to do. 

The smarter thing would probably be to cache all incoming photometry origins, into a set or something. But I have a feeling that this is exactly what the index is already doing, so probably this query is already going to be optimized.

jadalilleboe: @guynir42 should we get this merged and fetching the origins can be a separate PR? (maybe @thomasculino will have time to work on that over the summer, I'll be starting my fulltime job soon)

guynir42: Yeah a separate PR is fine by me.

skyportal/plot.py
```diff
@@ -799,6 +946,26 @@ def make_legend_items_and_detection_lines(
...
+        if "automaticallyVisibleFilters" in user.preferences:
+            for filter in user.preferences['automaticallyVisibleFilters']:
+                if filter in unique_filters:
+                    show_all_filters = False
+                    break
+        if "automaticallyVisibleOrigins" in user.preferences:
+            for origin in user.preferences['automaticallyVisibleOrigins']:
+                if origin in unique_origins:
+                    show_all_origins = False
+                    break
```
guynir42, line 968: I am wondering if it is not better to put this logic into the `check_visibility_on_phot_plot` function? That way we will have all the preferences fetching and hide/display choices done in one place.

guynir42: Is it just that this test is shared by multiple calls to functions?

guynir42: Another option would be to get a list of all the filters and origins to display and pass a list of indices. I don't know if that is better.

guynir42 (approved):
I have one suggestion on relocating the logic of what to display/hide when rendering the plot. If that doesn't make sense you can go ahead an merge.

skyportal/plot.py
jadalilleboe: yes, the test is shared by multiple calls to make_scatter, so I think this is an appropriate spot for it (and avoid doing the check over and over again in `check_visibility_on_phot_plot`). I have extracted the logic into a helper function though, it does stick out a bit in the middle of that function. The purpose of this code is the case where the plot does not contain any of the user's preferred filters/origins, in that case every point should be shown (I guess instead of none being shown, which is what would happen if there were no `show_all_filters` and `show_all_origins` flag and we just toggled visiblity based on if the point is in the user's preferences or not). so I would be inclined to keep the logic like this. are you okay with that @guynir42 ?

guynir42: Yeah this is fine.

jadalilleboe: I think failures are unrelated, ok to merge? @guynir42

guynir42: Go ahead.

PR #3013, opened 2022-05-31
Configuration for the GCN Service

Theodlz:
### Currently, in the gcn_service handler, the list of GCN notice types to be ingested in skyportal is fixed. This PR makes it configurable in the config.yaml

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -526,3 +526,17 @@ spectrum_types:
 #    wavelength: [5580, 5582, 8249, 8250]
 #    transmission: [0, 1, 1, 0]
 #    description: "basic ATLAS O filter - tophat transmission"
+
+gcn_notice_types:
+  - FERMI_GBM_FLT_POS
+  - FERMI_GBM_GND_POS
+  - FERMI_GBM_FIN_POS
+  - FERMI_GBM_SUBTHRESH
+  - LVC_PRELIMINARY
+  - LVC_INITIAL
+  - LVC_UPDATE
+  - LVC_RETRACTION
+  - AMON_ICECUBE_COINC
+  - AMON_ICECUBE_HESE
+  - ICECUBE_ASTROTRACK_GOLD
+  - ICECUBE_ASTROTRACK_BRONZE
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -12,26 +12,15 @@ env, cfg = load_env()
 init_db(**cfg['database'])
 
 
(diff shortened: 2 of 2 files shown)


services/gcn_service/gcn_service.py
```diff
@@ -12,26 +12,15 @@
...
-    gcn.NoticeType.ICECUBE_ASTROTRACK_GOLD,
-    gcn.NoticeType.ICECUBE_ASTROTRACK_BRONZE,
-)
 def handle(payload, root):
-
-    user_id = 1
-    with DBSession() as session:
-        post_gcnevent(payload, user_id, session)
+    notice_type = gcn.get_notice_type(root)
+    if notice_type in [
```
mcoughlin, line 17: Maybe this list should be read in at the same time as the config instead of again for every new event?

mcoughlin (approved):
@Theodlz small comments, just to confirm, you let it run a bit and it filters events as you expect?

Theodlz:
I've let it run for an hour and everything seemed to be in order. I have the same results as what i have in production (when using the same notice types) that is using the current main branch. Lets maybe leave it run some more.

config.yaml.defaults
```diff
@@ -526,3 +526,17 @@ spectrum_types:
 #    wavelength: [5580, 5582, 8249, 8250]
 #    transmission: [0, 1, 1, 0]
 #    description: "basic ATLAS O filter - tophat transmission"
+
+gcn_notice_types:
```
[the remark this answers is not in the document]
Theodlz: No reason for it not to work ! But as we discussed, the display for the very small localization of a swift event is a problem. I think we should/could add it in the config when we open a PR to fix that.

services/gcn_service/gcn_service.py
Theodlz: I'm not so sure what you mean by that. Want me to save that list once in a variable that then i use in this if statement ?

mcoughlin: Yeah that's all I meant.

mcoughlin:
@Theodlz maybe ask Guy to look quick otherwise ok for me to go in.

services/gcn_service/gcn_service.py
Theodlz: I'll take care of it soon then !

Theodlz: Done

Theodlz:
Hey @guynir42, do you think you could have a quick look ?

PR #3043, opened 2022-06-02
Observation plan refactor

mcoughlin:
This PR refactors the GCN events to simplify automated observation request posting.


skyportal/handlers/api/observation_plan.py
```diff
@@ -42,6 +44,95 @@
...
 
+def post_observation_plan(plan, user_id, session):
+    """Post ObservationPlan to database.
+    plan: dict
+        Observation plan dictionary
+    user_id : int
+        SkyPortal ID of User posting the GcnEvent
+    session: sqlalchemy.Session
+        Database session for this transaction
+    """
```
guynir42, line 55:
```suggestion
    """Post ObservationPlan to database.
    
    Parameters
    ----------
    plan: dict
        Observation plan dictionary
    user_id : int
        SkyPortal ID of User posting the GcnEvent
    session: sqlalchemy.Session
        Database session for this transaction
    """
```

skyportal/handlers/api/observation_plan.py
```diff
@@ -42,6 +44,95 @@
...
+
+    user = session.query(User).get(user_id)
```
guynir42, line 57: What prevents me from using someone else's user_id and posting stuff? Is that enforced in the handler that is calling this?

guynir42: Perhaps it is better to pass in a User object rather than an ID, that way the thing passing in the object will be more inclined to not allow a user to input the ID themselves without checking that the user is legal?

guynir42 (commented):
I have only one security concern which is not very crucial as long as the `post_observation_plan()` function is called via the handler. If it later gets imported into other parts of the code I am worried that it may not be very secure, as you may end up getting the `user_id` from outside.

skyportal/handlers/api/observation_plan.py
mcoughlin: @guynir42 This was to simplify the use from, for example microservices, which Stefan had suggested previously. As you intuited, this is to just be called from within handlers, and I would think making a new user in the session is cleaner.

guynir42 (approved):
Are we leaving it with the user id as input rather than the user object?

services/gcn_service/gcn_service.py
```diff
@@ -15,16 +17,49 @@
...
+                allocation = (
+                    session.query(Allocation)
+                    .filter(Allocation.proposal_id == proposal_id)
+                    .first()
+                )
+                event = session.query(GcnEvent).get(event_id)
 
-if __name__ == "__main__":
-    log = make_log("gcnserver")
+                if allocation is not None:
```
Theodlz, line 45: Maybe you could query the GcnEvents after the if statement that checks if there is no allocation? Just a small detail but it avoids querying the gcn events if there is no allocation in the first place.

Theodlz (commented):
Looks very good. Just one minor comment. Also, as we discussed the other day, it would be better to have that 'template' of observation plan per allocation/instrument rather than in the config. But I guess that could be the subject of the following PR not this one.

services/gcn_service/gcn_service.py
mcoughlin: I moved it up just after the event is committed so it is only loaded once actually, which I think is better. What do you think?

Theodlz (approved):

PR #3045, opened 2022-06-03
Add observing run from observation plan

mcoughlin:
This PR creates an observing run from an observation plan.


skyportal/tests/frontend/test_scanning_page.py
```diff
@@ -840,7 +840,7 @@ def test_user_without_save_access_cannot_save(
         f'//button[@name="initialSaveCandidateButton{public_candidate.id}"]'
     )
     driver.scroll_to_element_and_click(save_button)
-    driver.wait_for_xpath("//*[contains(.,'Insufficient permissions')]")
+    # driver.wait_for_xpath("//*[contains(.,'Insufficient permissions')]")
```
Theodlz, line 843: Should this just be deleted before merging if its not used anymore ?

static/js/components/ObservationPlanRequestLists.jsx
```diff
@@ -202,6 +202,14 @@ const ObservationPlanRequestLists = ({ gcnEvent }) => {
     setIsDeletingTreasureMap(null);
   };
 
+  const [isCreatingObservingRun, setIsCreatingObservingRun] = useState(null);
+  const handleCreateObservingRun = async (id) => {
+    console.log("id", id);
```
Theodlz, line 207: remove console.log

skyportal/handlers/api/followup_request.py
```diff
@@ -54,6 +55,60 @@
...
+        Database session for this transaction
+    """
+
+    user = session.query(User).get(user_id)
+
+    try:
+        assignment = ClassicalAssignment(**AssignmentSchema.load(data=data))
+    except ValidationError as e:
+        raise ValidationError(
+            'Error parsing followup request: ' f'"{e.normalized_messages()}"'
```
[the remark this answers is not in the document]
mcoughlin: We don't have a self available when running like this is the issue.

Theodlz:
More general comment on this review. A lot of the logic found in the post methods of handlers are now extracted from the handlers to be refactored as methods that can be used by the handler but also reused in other handlers, which is super useful (thanks for doing that !). In most of them, I don't see any try catch statement, I'm not sure they would be necessary, but could it be useful to add some try catch either around where you call the method or in the method itself, to catch some of the errors and display error messages in the logs ?

skyportal/handlers/api/followup_request.py
```diff
@@ -54,6 +55,60 @@
...
+            ClassicalAssignment.obj_id == assignment.obj_id,
+            ClassicalAssignment.run_id == run_id,
+        )
+        .first()
+    )
+
+    if predecessor is not None:
+        raise ValueError('Object is already assigned to this run.')
+
+    assignment = ClassicalAssignment(**data)
```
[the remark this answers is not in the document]
mcoughlin: https://github.com/skyportal/skyportal/blob/main/skyportal/handlers/api/followup_request.py#L163
I think this is about setting priority correctly.

static/js/components/ObservationPlanRequestLists.jsx
mcoughlin: yes

skyportal/tests/frontend/test_scanning_page.py
mcoughlin: yeah

mcoughlin:
@Theodlz good point!

Theodlz (approved):
No further comments. Would have been nice for the results of the simsurvey to open in another tab to not close skyportal tabs that can take a while to load some times, but its not essential.

skyportal/handlers/api/followup_request.py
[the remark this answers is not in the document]
Theodlz: Oh ! I am just talking about the error message itself (like the string) not the way the error is raised!

thomasculino (approved):
Same remark for the results of the simsurvey, otherwise it looks good

mcoughlin:
@Theodlz @thomasculino Yes I think down the link we might give the simsurvey stuff it's own page (or something) to allow for different injection numbers, light curves, etc. Can revisit maybe with the analysis framework stuff.

PR #3046, opened 2022-06-04
Fix invalid route message

jadalilleboe:
The "Invalid route" message disappeared, this puts it back in and adds a test for invalid routes.


mcoughlin (approved):
@jadalilleboe Really good find. thanks!

PR #3050, opened 2022-06-06
timezonefinder 6.0

mcoughlin:
This PR bumps timezonefinder to 6.0.1.


Theodlz (approved):
Using the frontend (i just looked at the telescope map and airmass plots which could be affected by that change) I can't find anything that is different from the previous version. It looks good to me.

PR #3052, opened 2022-06-07
Fix export to csv button

### LINKED ISSUE #3002: Export to csv button does not work correctly
The export to csv button on the photometry plot does not work correctly, nothing happens when clicked. should fix and also add regression tests if this was working before some of the recent changes to the photometry plot.

jadalilleboe:
fixes #3002 
the export to csv button and period controls have been broken for two months...


guynir42 (approved):
If this has been tested locally it looks fine to me.

PR #3053, opened 2022-06-07
Start adding instrument fields to observation plan globe

mcoughlin:
This PR starts to add fields to the observation plan globe.


Theodlz:
What is the objective of displaying a second skymap in the form to generate observation plans ? Isn't it enough to have it in the GcnSelection Form + a button to apply those settings to the observation plan form ? At first, it thought that you added the fields on the skymap that displays in  the list of observation plans, which I think is useful no ? 
I mean displaying it here: 
image

mcoughlin:
@Theodlz The idea of this one was to inform the making of observation plans with specific fields. It could go on the GcnSelectionForm but would still need to be tied to the observation plan making infrastructure which is why I put it where I did.

static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -47,6 +51,111 @@ const useStyles = makeStyles(() => ({
...
+          <CircularProgress />
+        </div>
+      ) : (
+        <div>
+          <LocalizationPlot
+            loc={loc}
+            instrument={skymapInstrument}
+            options={displayOptionsDefault}
+            height={300}
+            width={300}
```
jadalilleboe, line 105: why so small? I like the size of the one in GcnSelectionForm.

static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -47,6 +51,111 @@ const useStyles = makeStyles(() => ({
...
+  displayOptionsDefault.instrument = true;
+
+  const [skymapInstrument, setSkymapInstrument] = useState(null);
+  useEffect(() => {
+    const fetchSkymapInstrument = async () => {
+      const response = await dispatch(
+        GET(
+          `/api/instrument/${instrument.id}?includeGeoJSONSummary=True&localizationDateobs=${loc.dateobs}&localizationName=${loc.localization_name}`,
+          "skyportal/FETCH_INSTRUMENT"
+        )
```
jadalilleboe, line 77: Wouldn't it be more kosher with the rest of the codebase to put this in a duck file? The long url seems out of place

static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -164,6 +274,9 @@ const ObservationPlanRequestForm = ({ gcnevent }) => {
   };
 
   const handleQueueSubmit = async ({ formData }) => {
+    if (selectedFields.length > 0) {
+      formData.field_ids = selectedFields;
+    }
```
jadalilleboe, line 279: and put the filtering of selected fields here.

jadalilleboe (commented):
A UX note: when I try to get the instrument tiles to show up, there's a delay between selecting the instrument and the tiles showing up on the plot. Would be nice to have some sort of loading icon/message.

jadalilleboe:
> @Theodlz The idea of this one was to inform the making of observation plans with specific fields. It could go on the GcnSelectionForm but would still need to be tied to the observation plan making infrastructure which is why I put it where I did.

If things could be refactored to where GcnSelectionForm could access the observation plan infrastructure, I think that would be best. Having the skymap twice for the selecting fields feature seems a bit redundant to me? But if that's not doable I think this would be fine.

static/js/components/ObservationPlanRequestForm.jsx
Theodlz: Agreed. And honestly, even on the gcn selection form it could be bigger or the skymap could use an initial zoom. It is very small compared to the huge part of the page dedicated to that form. There's a bunch of unused empty space.

static/js/components/ObservationPlanRequestForm.jsx
Theodlz: Agreed

Theodlz:
> > @Theodlz The idea of this one was to inform the making of observation plans with specific fields. It could go on the GcnSelectionForm but would still need to be tied to the observation plan making infrastructure which is why I put it where I did.
> 
> If things could be refactored to where GcnSelectionForm could access the observation plan infrastructure, I think that would be best. Having the skymap twice for the selecting fields feature seems a bit redundant to me? But if that's not doable I think this would be fine.

Couldn't agree more. The redundancy of the fields in the GCNSelectionForm and ObservationPlanForm and the lack of tooltips (and also some useless buttons instead of titles I think ?) made the page already pretty confusing, so having yet another SkyMap doesn't help with that. 

It needs to be more guided or refactored to share some stuff between the 2 forms to avoid redundancy. 

Also, but this could be the suggest of another PR, it would be nice if the observation plan list took the whole width of the page. Right now, its not super user friendly to have to scroll horizontally to see all the fields. Maybe some elements on the right part of the screen could be put in closed by default "accordion" elements of MUI, so that there is enough space for the observation plan list to appear at the bottom of the page without having an empty space in between the GCNSelectionForm and the ObservationPlanForm + list.

mcoughlin:
@Theodlz @jadalilleboe In general I agree with you, some of the issue here is my uncomfortably with React.js and so on that there is some redundancy. But I think the other is also conceptual, as historically we have had separate pages for the observation plan part and the "analysis" part, I.e the observations and GCNs and summaries that Theo is thinking about now. So while the potential merger of the 'what do we plan to do skymap' with the 'what did we actually do' skymap might be nice, I think it would need to be thought through carefully how to make it nice. For now though, I feel like we could just comment out the whole GCnSelectionForm component.

Theodlz:
@mcoughlin I'm thinking that maybe separating the page in 2, with an Analysis Section and below an Observation Plan section. Each being in an "Accordion" from MUI. By default only the Analysis is open, and optionally when opening the Observation Plan section the first section closes? In general separating the 2 would be nice. It's hard to read the ObservationPlan list right now when it takes just half of the page.

static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -60,6 +169,7 @@ const ObservationPlanRequestForm = ({ gcnevent }) => {
   const [selectedLocalizationId, setSelectedLocalizationId] = useState(null);
   const [isSubmitting, setIsSubmitting] = useState(false);
   const [planQueues, setPlanQueues] = useState([]);
+  const [selectedFields, setSelectedFields] = useState([]);
```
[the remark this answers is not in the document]
mcoughlin: I don't quite know how this works given the instrument is changing with the allocation.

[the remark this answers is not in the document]
jadalilleboe: so I am a bit confused with how the 'use selected fields in observation plan' button is working- so if you press that button, the user's selected fields will be sent along with the request? But if the user doesn't press it then no fields will be sent even if the user has some selected? It seems to me like the default should just be if the user has some fields selected on the skymap, it should be sent along with the request.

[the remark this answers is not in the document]
jadalilleboe: agree!

mcoughlin:
@Theodlz @jadalilleboe  Do you know a good way (i.e. a technique that doesn't require a new package) to select multiple fields in a dropdown? Lots of stuff on the internet, but not clear. I am thinking that might be a nice way to do the display.

mcoughlin:
@jadalilleboe @Theodlz maybe see what you think of the implementation as committed.

jadalilleboe:
I got this error after selecting LSST as the allocation on the observation plan request form.
Screen Shot 2022-06-08 at 4 42 16 PM

mcoughlin:
@jadalilleboe I think it would be nice to couple the dropdown with working tooltips. At that point I think it is definitely useful to have both (as the field numbers rise sequentially going around the globe).

jadalilleboe:
@mcoughlin Is it necessary for the minimum viable product to have fields selected in the dropdown be colored in on the skymap? I think it would be clearer but maybe not necessary. 

In general there should some be some more linkage between the state of selected tiles in the skymap and in the form field i.e. hitting the clear fields button should also probably clear all the selected fields on the skymap. Also if you select some items in the dropdown and then use the skymap/button to add more fields, the previously selected fields are lost.

static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -35,18 +38,183 @@ const useStyles = makeStyles(() => ({
...
+    );
+    const selectedIds = theseSelectedFields.map((f) => f?.field_id);
+    setSelectedFields(selectedIds);
+  };
+
+  const fields = [];
+  skymapInstrument?.fields?.forEach((field) => {
+    fields.push(Number(field.id));
+  });
+  fields.sort();
```
jadalilleboe, line 107:
```suggestion
  fields.sort((a, b) => a - b);
```

mcoughlin:
@jadalilleboe As a knower of things, could you outline for me how you would get this interaction going? I don't think it's strictly MVP necessary, but clearly would be nice.

jadalilleboe:
@mcoughlin check this out: [https://github.com/jadalilleboe/skyportal/tree/add_fields](https://github.com/jadalilleboe/skyportal/tree/add_fields)

Works quite well with the dropdowns and making selected tiles light up after being selected on the dropdown. The problem now is that the component is rerendered everytime after a tile is selected on the skymap. I think it's ok to rerender after fields are selected with the dropdown and after clearing fields but obviously not ok while using the skymap. I am thinking this could be solved with [react context](https://reactjs.org/docs/context.html) to only make the select component rerender and update with the correct selected state but I am not sure yet

Hnet-image (3)

mcoughlin:
Perhaps @stefanv's tooled up D3 knowledge knows how to prevent the map from always rotating back?

jadalilleboe:
saving rotations in a state variable seems to work a bit better:
Hnet-image (4)
now just needs to be tweaked to save the previous rotation position more accurately, it is a bit jumpy.

jadalilleboe:
this works nicely now, demo selecting/unselecting fields using the visualization and also the dropdown:

https://user-images.githubusercontent.com/82007190/173237277-24c0c11e-2288-4c89-a10b-f59ecd7a00cb.mov

I also cleaned up the code to use a useState variable for the selected fields instead of directly mutating the state within the instrument's fields

mcoughlin:
@jadalilleboe now I feel like you are just showing off ;)

Theodlz:
Super nice !!! Nicely done @jadalilleboe 🔥

static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -35,18 +38,210 @@ const useStyles = makeStyles(() => ({
...
   },
   localizationSelectItem: {
     whiteSpace: "break-spaces",
   },
+  fieldsToUseSelectItem: {
+    whiteSpace: "break-spaces",
+  },
+  SelectItem: {
+    whiteSpace: "break-spaces",
+  },
```
jadalilleboe, line 55: I'd make all these just one declaration

static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -35,18 +38,210 @@ const useStyles = makeStyles(() => ({
...
+        ))}
+      </Select>
+      <Button
+        id="clear-fieldsToUseSelect"
+        size="small"
+        color="secondary"
+        onClick={() => clearSelectedFields()}
+      >
+        Clear Fields
+      </Button>
```
jadalilleboe, line 112: would be nice if this button color was a little darker. Actually a lot of the buttons on this page look different from the rest of the app, would be good if they were all unified. But maybe that can wait till the shift to mui5

jadalilleboe (approved):

static/js/components/ObservationPlanRequestForm.jsx
mcoughlin: yes good idea.

static/js/components/ObservationPlanRequestForm.jsx
mcoughlin: Yes this is exactly the kind of thing that will change there, so we can prettify the page once that goes in.

mcoughlin:
I have opened issues (https://github.com/skyportal/skyportal/issues/3081, https://github.com/skyportal/skyportal/issues/3082) for Theo's main comments.

PR #3056, opened 2022-06-07
Catch (new) Gaia exception

profjsb:
During the transition from EDR2 to DR3 the main Gaia catalog used by astroquery has become unstable. We look for this but did not catch the new error that is arising. This PR attempts to fix this.

The backup Gaia catalog also changed its units reporting. This PR fixes this.

### DIFF
--- a/skyportal/utils/offset.py
+++ b/skyportal/utils/offset.py
@@ -14,7 +14,7 @@ import matplotlib.pyplot as plt
 import seaborn as sns
 import numpy as np
 import numpy.ma as ma
-from scipy.ndimage.filters import gaussian_filter
+from scipy.ndimage import gaussian_filter
 from joblib import Memory
 
 from astropy import units as u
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

skyportal/utils/offset.py
```diff
@@ -83,6 +83,9 @@ def _connect(self):
         except (HTTPError, ConnectionResetError):
             log("Warning: main Gaia TAP+ server failed")
             self.is_backup = True
+        except:  # noqa: E722
+            log("Warning: main Gaia TAP+ server failed in a way we didn't expect.")
```
stefanv, line 87: Log the exception here, so we can see what type it is and what error message is being returned.

profjsb: Done.

skyportal/utils/offset.py
```diff
@@ -14,7 +14,7 @@
 import seaborn as sns
 import numpy as np
 import numpy.ma as ma
-from scipy.ndimage.filters import gaussian_filter
+from scipy.ndimage import gaussian_filter
```
profjsb, line 17: `filters` is deprecated now in scipy

PR #3057, opened 2022-06-08
Photometric Statistics

guynir42:
This new table keeps a single row for each Obj. 
The table contains data summaries on all the photometric points. 
All the stats are indexed. 
All the stats can be updated using only the stats and a single new photometry point. 

This will make it very easy to query objects by last detected, etc.


skyportal/handlers/api/phot_stat.py
```diff
@@ -0,0 +1,195 @@
...
+        with DBSession() as session:
+            stmt = sa.select(PhotStat).where(PhotStat.obj_id == obj_id)
+            phot_stat = session.scalars(stmt).first()
+
+            if phot_stat is None:
+                return self.error(
+                    f'Could not find a PhotStat for object with id "{obj_id}". '
+                )
+
+            stmt = (
```
profjsb, line 51: Add a comment here that this is a universal query, not permissioned.

skyportal/handlers/api/phot_stat.py
```diff
@@ -0,0 +1,195 @@
...
+                sa.select(Photometry)
+                .where(Photometry.obj_id == obj_id)
+                .order_by(Photometry.created_at.desc())
+            )
+            last_photometry = session.scalars(stmt).first()
+            if last_photometry:
+                phot_stat.last_phot_add_time = last_photometry.created_at
+            else:
+                phot_stat.last_phot_add_time = None
```
profjsb, line 60: Are you also trying to save the update to `phot_stat` here or just update the dictionary and return it? Seems like there are other statistics that could also be updated?

skyportal/models/phot_stat.py
```diff
@@ -0,0 +1,668 @@
...
+        )
+
+        return np.sqrt(new_var)
+
+
+@event.listens_for(Photometry, 'after_insert')
+def insert_into_phot_stat(mapper, connection, target):
+
+    # Create or update PhotStat object
+    @event.listens_for(DBSession(), "before_flush", once=True)
```
profjsb, line 630: why is this before_flush?

skyportal/models/phot_stat.py
```diff
@@ -0,0 +1,668 @@
...
+#         phot_stat = session.scalars(
+#             sa.select(PhotStat).where(PhotStat.obj_id == obj_id)
+#         ).first()
+#
+#         if phot_stat is not None:
+#             all_phot = session.scalars(
+#                 sa.select(Photometry).where(Photometry.obj_id == obj_id)
+#             ).all()
+#             print(f'len(all_phot) = {len(all_phot)}')
+#             phot_stat.full_update(all_phot)
```
profjsb, line 668: remove

profjsb (commented):
Looking good. Thanks for the extensive testing. A few things:

 1) can you briefly explain (in a docstring? or in the docs?) why phot_stat APIs require system admin level permissions? ( I think it's because you need access to all the photometry)
 2) It would be useful to have a short intro to phot stats in the docs. And how a user might get different answers if they were to do this with just the photometry they can see.
 3) A useful stat to add: max/min rise rate (mag/day). This is the maximum/minimum between the successive detection (and the last non-detection before each detection). Could also add a max_rise_rate_signficance which is max rise rate divided by the (quadrature) uncertainty of the different between two data points.

guynir42:
@profjsb So a few things: 
1) max/min rise rate is tricky because it would require querying all photometry on the source with each update. The way I can do it without all that is to always calculate the rise time between the 1st detection and the new point (or the peak and the new point). 
2) I used `before_flush` because it did not work when I tried `after_flush`. I don't know why. 
3) All columns of the `PhotStat` are indexed. I can't see a reason to keep summary statistics if you are not going to filter on them.

skyportal/handlers/api/phot_stat.py
guynir42: The idea is that PhotStats is already updated when the photometry is being inserted (or patched or deleted). In this part of the code I'm just giving the user an opportunity to check that the PhotStat object has been updated close after the last photometry was posted.

skyportal/models/phot_stat.py
guynir42: Before flush has less options for what you can change in the session state, and I could not get it to work at all.

guynir42:
Tests failures seem to be real. I'll try to fix them.

guynir42:
@profjsb tests are fixed. Ready for re-review.

profjsb (approved):
Looks great. You'll need to fix the migration path before merging.

PR #3062, opened 2022-06-09
Retrieving all unique origins to display them in a dropdown menu

thomasculino:
In #3012, we were displaying dropdown menus on the profile page to choose custom filters. Two of these dropdown menus should have been displaying the list of all unique origins. However, this list was hard coded. That is why this PR aims to retrieve all unique origins from the photometry table. In the previous PR, we were discussing how we should retrieve all unique origins in order to display them in these dropdown menus. We concluded that because the `origin` column is indexed it should not be too expensive to query all `distinct()` values.


guynir42 (approved):
LGTM

thomasculino:
> LGTM

I just noticed a little problem in the frontend/test_profile file. It checks if we can find 'Muphoten' in the origin dropdown menu. However, because of my changes there are no hard coded origin anymore. So, when the test is running, we can see that there are no origin in the dropdown menu. I was wondering if I should add an origin when running the test upload_photometry and then check if we see this origin on the frontend. But this would require to run the test uploading photometry before the test_profile. Otherwise, I can just get rid of this part in the tets_profile. What do you think is better ?

guynir42:
> > LGTM
> 
> I just noticed a little problem in the frontend/test_profile file. It checks if we can find 'Muphoten' in the origin dropdown menu. However, because of my changes there are no hard coded origin anymore. So, when the test is running, we can see that there are no origin in the dropdown menu. I was wondering if I should add an origin when running the test upload_photometry and then check if we see this origin on the frontend. But this would require to run the test uploading photometry before the test_profile. Otherwise, I can just get rid of this part in the tets_profile. What do you think is better ?

You cannot rely on having other tests post things. Why not post photometry in the test, or use `public_source` which already has some photometry on it?

static/js/components/OriginSelect.jsx
```diff
@@ -1,18 +1,36 @@
...
 import PropTypes from "prop-types";
+import { useSelector, useDispatch } from "react-redux";
 import SelectWithChips from "./SelectWithChips";
 
+import * as photometryActions from "../ducks/photometry";
+
 const OriginSelect = ({ onOriginSelectChange, initValue, parent }) => {
-  const origins = ["Clear selections", "Muphoten", "STDpipe"];
+  const dispatch = useDispatch();
+  const origins = () => {
```
jadalilleboe, line 10: how about fetchOrigins?

jadalilleboe (approved):

jadalilleboe:
@thomasculino do you think None should be included or no? I think None is a bit confusing to have as an option
Screen Shot 2022-06-12 at 6 59 36 PM

thomasculino:
> @thomasculino do you think None should be included or no? I think None is a bit confusing to have as an option
> Screen Shot 2022-06-12 at 6 59 36 PM
> 

The None comes from the fact that there are no origin specified in at least one photometry. I should remove it after fetching all origins I agree

mcoughlin (approved):
Very useful @thomasculino !

PR #3063, opened 2022-06-09
Format configuration in YAML format

stefanv:


mcoughlin (commented):
@stefanv Are the '' required when doing it like this?

stefanv:
Only for fields that contain special values. I was too lazy to check those two, but I should.

mcoughlin:
@stefanv I downloaded it and removed the '' and it seemed to load ok at least, so hopefully is fine.

stefanv:
Feel free to push to my PR and merge.

PR #3066, opened 2022-06-10
Shift Calendar Enhancement (MUI5, display, options, tooltips)

Theodlz:
## What's new:

- In the calendar element, you now have a section with options and a tooltip to help users understand what is going on with the calendar.
- The first option, allows you to display or not display the shifts that you are not a member of (show all shifts)
- The second option, allows you to filter shifts by groups. If you activate it, a dropdown appears where you can select which groups you want to be able to see on the calendar (as each shift belongs to a group)
- The third element is a tooltip that gives you a bit of help with the new color code of the Shifts on the calendar (which will be mentioned right after). It displays when hovering over the help icon.
- In the DB, shifts now have a new column called `required_users_number` (I'll be happy to discuss its name, I am not 100% convinced by the clarity of this one). This is necessary and a missing feature of the shifts so far. During a campaign, we don't have a random number of people on shift, but a fixed number. This column allows you to set that value, so that later on the frontend, you can display shifts that are missing users differently. **Important to note**, I did not make it required, as I believe that shifts without this users count could be created, especially if used outside of a campaign, where we do not necessarily need to keep track of how many people are on shift (more flexibility). I'll be happy to talk about this too. It is not required if you use the API, but required (or has a default value) in the forms to create shifts on the frontend, so that needs to be changed to either match what I did in the backend, or changing the backend to match what I did in the frontend.
- On the Calendar, shifts have now a distinct and useful color code.
- New color code for the shifts: Before a shift was green if you joined it, blue otherwise. Now, here is the new color code :

1. If a shift has fewer users than it requires, and it will begin in less than 24h, it is displayed in `red`.
2. If a shift has fewer users than it requires, and it will begin in less than 72h but more than 24h, it is displayed in `orange`.
3. If a shift is happening or will happen in less than 72 hours and has all the users he requires, it is displayed in `blue`.
4. If  a shift will happen in more than 72 hours, it is also displayed in `blue`.
5. If a shift  has already happened, it is displayed in `grey`
6. If you join a shift with the color code number 3 or 4 (a shift with all of his users did not happen or is happening, or does not have them but will take place in more than 72 hours), it will be green.
7. Otherwise, if you join a shift that has the color code `1` or `2`, (has fewer users than it requires, and will happen in less than 72 hours), it will be both `green` and `red/orange`, hatched.

Again, the tooltip at the bottom of the calendar reminds you of the color code.

## PREVIEW (last version)
shift_calendar_enhancement


Theodlz:
Hi @mcoughlin, tagging you for review first as we already talked about it earlier.

skyportal/handlers/api/shift.py
```diff
@@ -68,6 +68,9 @@ def post(self):
                             description:
                               type: string
                               description: New Shift's description
+                            required_users_number:
+                              type: integer
+                              description: New Shift's required users number
```
mcoughlin, line 73: number of required users for a complete shift?

data/db_demo.yaml
```diff
@@ -890,6 +890,7 @@ shifts:
     name: "the Night Shift"
     shift_admins:
       - =testadmin
+    required_users_number: 6
```
mcoughlin, line 893: Is required or requested better here?

skyportal/models/shift.py
```diff
@@ -133,6 +133,12 @@ class Shift(Base):
         doc='Elements of a join table mapping Users to Shifts.',
     )
 
+    required_users_number = sa.Column(
+        sa.Integer,
+        nullable=True,
+        doc='The number of users required to join this shift for it to be considered full',
```
mcoughlin, line 139: Yes I think this docstring is better than the one above.

mcoughlin (approved):
@Theodlz The interface is quite slick. I notice that when I add too many folks, you just add the number up until the max. This might be a feature and not a bug (where you just prevent anyone from being added), but I thought I would make note of it depending on what you want.

data/db_demo.yaml
Theodlz: I'm not so sure. It seems to be that it is a requirement, as otherwise the shift is incomplete. I'll let you decide you have more experience than me being on shift.

thomasculino (commented):
The code looks good. I like the idea of using a color code however I think it can be improved slightly. I think we should have all past shifts in the same color (grey?). I think we should also choose another color for the shifts happening in more than 72 hours that you are not part of because it will only create confusion with the past ones if we keep them in grey.

thomasculino (approved):

Theodlz:
@thomasculino you mean that after a shift has passed, ragardless of you being a member of it or not, it is in grey ? And for example for shift in more than 72hours, a lightblue or something alike to look different than the grey ones ?

thomasculino:
> @thomasculino you mean that after a shift has passed, ragardless of you being a member of it or not, it is in grey ? And for example for shift in more than 72hours, a lightblue or something alike to look different than the grey ones ?

That's what I had in mind yes. However, it is only a suggestion. We need to figure out how we want to display shifts. If we want to display shifts depending on whether you are in them or not then your idea was better. But if we want to display shifts depending on whether the shift is coming or passed, then I prefer my suggestion. We need to find what is the best option from the user perspective.

Theodlz:
> > @thomasculino you mean that after a shift has passed, ragardless of you being a member of it or not, it is in grey ? And for example for shift in more than 72hours, a lightblue or something alike to look different than the grey ones ?
> 
> That's what I had in mind yes. However, it is only a suggestion. We need to figure out how we want to display shifts. If we want to display shifts depending on whether you are in them or not then your idea was better. But if we want to display shifts depending on whether the shift is coming or passed, then I prefer my suggestion. We need to find what is the best option from a user perspective. @Theodlz

Hi @thomasculino, i implemented the changes you asked for, or at least some of them. I think both of our ideas can co-exist, and we don't have to choose one or another. Here is what it looks like:

shift_calendar_enhancement

Theodlz:
Note to reviewers: The new upgrade to MUI5 completely destroyed the shift page :sob: In my last commit I fixed it by updating the imports and changing some elements. I think it would also help @thomasculino if this PR could be rewieved / get in quickly ! Thanks a lot for your help.

mcoughlin (approved):
@Theodlz Tested and looks great.

PR #3070, opened 2022-06-11
Update test_quick_search.py

StephenTheCoder:
Resolving my first issue in Skyportal


mcoughlin:
@StephenTheCoder Can you run 
pre-commit install
and then
pre-commit run --files skyportal/tests/frontend/test_quick_search.py

to fix the formatting and push that?

stefanv:
@StephenTheCoder Thank you for your first contribution! 

It'd be great if you can be a bit more specific in the title and description, just so we get a sense of what this change addresses.

PR #3078, opened 2022-06-13
Closes #3073

### LINKED ISSUE #3073: Default 'Observing Run' list not showing imminent runs
The current default is to show observing runs listed in time order, starting from the next that will take place, and hiding those that have already expired. In principle that makes a lot of sense, however the observing runs are deemed to have taken place at exactly midnight at the start of the date on which they occurred. This means that, whenever you click on the page on the day of an observing run, you cannot see that observing run because it is deemed to have already occured. I attach an example, from 2022/06/12, where my run on that day is no longer visible despite not having taken place yet. And the run beginning in 36 hours (evening of 2022/06/13) is listed as beginning in 11 hours.

Since the single most common use case of that page is to check targets just before doing an observing run, I would propose to offset the cut by one day, i.e "to display all observing runs that are scheduled no earlier than 24 hours in the past", rather than the present "to display all observing runs that are scheduled no earlier than now". I think that would be a relatively easy change that would improve user experience. 

Screenshot 2022-06-12 at 12 34 57 PM

StephenTheCoder:
Closes error #3073 (the old one failed)

### DIFF
--- a/static/js/components/ObservingRunPage.jsx
+++ b/static/js/components/ObservingRunPage.jsx
@@ -144,7 +144,8 @@ const ObservingRunList = ({ observingRuns, deletePermission }) => {
   const { telescopeList } = useSelector((state) => state.telescopes);
   const groups = useSelector((state) => state.groups.all);
 
-  const nowDate = dayjs().utc().format("YYYY-MM-DDTHH:mm:ssZ");
+  const nowDate = dayjs().utc().format("YYYY-MM-DDTHH:mm:ssZ")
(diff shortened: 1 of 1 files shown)


guynir42 (approved):

mcoughlin:
@StephenTheCoder For that, just do:
pre-commit install
pre-commit run --files static/js/components/ObservingRunPage.jsx
git commit static/js/components/ObservingRunPage.jsx

and we can get this merged.

PR #3083, opened 2022-06-14
Add airmass calculation to globe

mcoughlin:
This PR adds an airmass calculation to color globe tiles.


static/js/components/ObservationPlanRequestForm.jsx
```diff
@@ -246,6 +256,11 @@ const ObservationPlanRequestForm = ({ gcnevent }) => {
   const [skymapInstrument, setSkymapInstrument] = useState(null);
   const [selectedFields, setSelectedFields] = useState([]);
 
+  const defaultAirmassTime = Date(
+    dayjs(gcnevent?.dateobs).format("YYYY-MM-DDTHH:mm:ssZ")
+  );
+  const [airmassTime, setAirmassTime] = useState(defaultAirmassTime);
+
```
Theodlz, line 263: Something doesn't work well here. You want by default to be the airmass at the dateobs, but for me on the frontend it just shows me the current time. Also, it says it has to be in UTC, but it shows me the current time in my local. I think this could confuse users. Also, maybe its better if the default airmass time is in the middle of the start and end date selected in the form. It might be confusing to see the "observability" of tiles outside of the time frame showed in the form. It could get someone to think they cant observe when in fact they could.

Theodlz (commented):
Hi Michael, a bit of a general comment on this PR. This allows you to show if a tile will be observable at a given time right? But instead, shouldn't it tell you if a tile will be observable in the time slot you select in the form? Not if it will always be available, but if it will be at least once at some point. If that is not easily doable, maybe at least default the airmass time to something that is like at the middle between start and end date of the form.

static/js/components/ObservationPlanRequestForm.jsx
mcoughlin: @Theodlz I understand some of what you mean... could you make what I think are the small mods and show me?

thomasculino (approved):
Looks good to me

Theodlz (approved):
Great. Thanks for addressing my comments ! Works very well for me.

PR #3088, opened 2022-06-15
Add basic queue interface

mcoughlin:
This PR adds a basic queue interface to the observations page, where the queues can be displayed (and deleted if that capability exists).

### DIFF
--- a/skyportal/facility_apis/_base.py
+++ b/skyportal/facility_apis/_base.py
@@ -66,6 +66,7 @@ class _Base:
             "remove": cls._isimplemented('remove'),
             "retrieve": cls._isimplemented('retrieve'),
             "queued": cls._isimplemented('queued'),
+            "remove_queue": cls._isimplemented('remove_queue'),
         }
 
     # subclasses should not modify this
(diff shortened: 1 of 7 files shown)


stefanv:
Do you mind adding a bit of context to the PR description. E.g., what are ZTF queues, how do we interact with them, etc., just to help us review this change.

Theodlz:
Talked about it on Slack with Michael already but here is a preview for the frontend. You can use it in the description if you want @mcoughlin.

#### PREVIEW
observation_page

Also combined the 3 frontend forms to add observations in one "Accordion" type element so the page can be a bit less crowded, and added some spaces in between elements.

thomasculino (approved):
Theo made the modifications we discussed on slack regarding how the components should be displayed and it works well.

stefanv:
@mcoughlin When there is a request for clarification on a PR so that someone can review it (i.e. an expressed interest in reviewing), per our agreed policy you should give a reasonable opportunity for review. This change is fine, but in the future if you need to move faster than reviewers allow, merge the changes into a branch and base future work off of that. If we don't take the time to bring the team up to speed with changes, then very soon it becomes impossible for them to help review, which just worsens the situation.

mcoughlin:
Thanks @stefanv, point well taken. I was iterating with Theo and Thomas in real time to test out functionality, which I want in today per our discussion on the GRB triggering epic at the SkyPortal call on Tuesday.

PR #3089, opened 2022-06-15
WIP: display logging data

mcoughlin:
This PR starts to try to read in log files.


mcoughlin:
@Theodlz For a while, we thought some (subset of) logs appearing on an admin accessible page could be useful for tracking errors without having to go into containers. Still worth thinking about.

mcoughlin:
@Theodlz This one too? Now that you made your endpoint?

Theodlz:
@mcoughlin should we close this? I feel like in between the sentry implementation, and the endpoint to grab the logs, we don't really need a dedicated frontend for this

PR #3091, opened 2022-06-15
Update config to provide defaults where possible, otherwise use `cfg.get`

stefanv:
This requires eyes from @mcoughlin, @guynir42, @profjsb, and @Theodlz, since it touches pieces of code you wrote.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -49,10 +49,12 @@ app:
   atlas:
     protocol: https
     host: fallingstar-data.com
+    port:
 
   kait:
     protocol: http
     host: herculesii.astro.berkeley.edu
+    port:
 
   treasuremap_endpoint: http://treasuremap.space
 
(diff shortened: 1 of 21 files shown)


stefanv:
The usage of `cfg.get` should be extremely limited for SkyPortal, since our configuration file defines defaults in almost all cases.

guynir42:
> The usage of `cfg.get` should be extremely limited for SkyPortal, since our configuration file defines defaults in almost all cases.

I think that really depends. Just because our config file defines a bunch of stuff doesn't mean other people can't delete it and leave it blank. 

In my opinion, whenever you can supply a reasonable default you should do that, even though we define most of that stuff in the config.

stefanv:
We *do* supply reasonable defaults (in `config.yaml.defaults`), and you cannot just delete those values so easily.

stefanv:
What you don't want to do is to provide all defaults twice, once in the config and another time in the code.  I already found instances where they were out of sync.

stefanv:
If you leave a value blank, it can still be accessed using `cfg['key']`, btw.

profjsb (approved):
LGTM

stefanv:
Thanks for that last commit, @guynir42!

PR #3092, opened 2022-06-16
User affiliations

Theodlz:
This PR is a work in progress, that is based on the changes made on this baselayer PR: https://github.com/cesium-ml/baselayer/pull/328

This adds the backend and frontend to create users with affiliations, as well as modifying them (both by an admin and the users themselves). Compared to username, first name and last name that can't be edited by an admin via the frontend, here, in the UserManagement Menu, I added the possibility to edit the affiliations of users. This is because an error in the way they are written could be a problem when using the user's affiliation to automatically generate circulars for example. Allowing the admin to edit it seemed useful to me.


guynir42 (approved):
Minor comments below:

guynir42:
@Theodlz I think strategically we should merge this after this week's deploy. 
The migrations may turn out to be tricky if we need to roll back and I won't be around this weekend to take care of it. 
Also I don't think this is super urgent, so merging after the deploy is my preference.

mcoughlin:
@guynir42 Shall we merge?

guynir42:
@mcoughlin go for it.

PR #3095, opened 2022-06-16
Observation Page - Display of the Add Observation Menu

Theodlz:
Right now, on the right part of the page, you can find a Add New Observation Header, and when clicking on it, it expands and shows you the 3 options that skyportal offers to add new observations.

However, these options are not in the same Paper container as the header, but separated from it, which is confusing.
This PR fixes that.

PREVIEW:

observation_page


mcoughlin (approved):
@Theodlz Nice UI change.

PR #3096, opened 2022-06-16
WIP: reminders

mcoughlin:
This PR starts thinking about reminders, i.e. UserNotifications that happen with some delay and potentially repeating.

Known issues:
(i) How do we refresh the queue here?

@Theodlz @thomasculino I could use some help, lots of overlap with comments and shifts and such.

EDIT 1 (Theo):
I changed some of the logic for the microservice, especially to avoid triggering multiple reminders all at once if for example, the app was down for a while (which would have triggered all the missed reminders `n` times depending on how long the app was down for. Now if a reminder was missed multiple times, it will only trigger one (the last one).

Also, as discussed @mcoughlin, we will add some basic frontend elements to this PR, which besides making it usable by users after merging, allows it to be tested more easily by reviewers.

EDIT 2 (Theo):
I added datatables on the frontend to see the reminders on sources, gcn events and shifts. The last one missing is the reminders on spectra. As both spectra and source appear on the same page, I am unsure on how to implement that. But I'm thinking that displaying both in the datatable, with an extra column showing the resourceType could work well. However, that means adding some stuff in the backend as we do not having the resourceType inside of the reminder currently.
On the datable itself, you'll find a delete button to delete reminders, as well as an ADD button to add a reminder. That button opens MUI Dialog with an rjfs form where you can create a basic reminder (just for yourself).


services/reminders/reminders.py
```diff
@@ -0,0 +1,128 @@
...
+                            url=url_endpoint,
+                        )
+                    )
+
+                    flow = Flow()
+                    flow.push('*', "skyportal/FETCH_NOTIFICATIONS", {})
+
+                    reminder.number_of_reminders = reminder.number_of_reminders - 1
+                    if reminder.number_of_reminders > 0:
+                        reminder.next_reminder = datetime.now() + timedelta(
```
Theodlz, line 111: Instead of `datetime.now()`, I believe you should use the current value of next_reminder. Otherwise, you'll keep accumulating delay.

services/reminders/reminders.py
```diff
@@ -0,0 +1,128 @@
...
+                            text=text_to_send,
+                            notification_type="mention",
+                            url=url_endpoint,
+                        )
+                    )
+
+                    flow = Flow()
+                    flow.push('*', "skyportal/FETCH_NOTIFICATIONS", {})
+
+                    reminder.number_of_reminders = reminder.number_of_reminders - 1
```
Theodlz, line 109: Couldn't we simply delete the reminder from the DB when it reaches 0 ?

services/reminders/reminders.py
```diff
@@ -0,0 +1,128 @@
...
+                        UserNotification(
+                            user=reminder.user,
+                            text=text_to_send,
+                            notification_type="mention",
+                            url=url_endpoint,
+                        )
+                    )
+
+                    flow = Flow()
+                    flow.push('*', "skyportal/FETCH_NOTIFICATIONS", {})
```
Theodlz, line 107: Is it needed to do that here? Shouldn't it be done in the UserNotification model? I think there is a decorator there that calls a function to push notifications to the frontend when something is added a UserNotification is added.

services/reminders/reminders.py
```diff
@@ -0,0 +1,128 @@
...
+                    if reminder.number_of_reminders > 0:
+                        reminder.next_reminder = datetime.now() + timedelta(
+                            days=reminder.reminder_delay
+                        )
+                        await self.put([reminder.id, reminder.__class__])
+
+                    session.add(reminder)
+                    session.commit()
+
+                    time.sleep(10)
```
Theodlz, line 119: I'm not sure of how all of this works. But a delay of 10 seconds between every reminder (of all users combined) seems that i could be a problem. Instead, maybe the service should iterate through the whole queue and then you can have a time.sleep() no ?

skyportal/models/shift.py
```diff
@@ -142,6 +142,15 @@ class Shift(Base):
         doc="Comments posted about this Shift.",
     )
 
+    reminders = relationship(
```
Theodlz, line 145: One very cool thing that the reminders allow us to do is to do things like notify users when their shift is about to start. In general, i believe we could reuse this code to do the following later on: 
Add a reminder for all users 72 hours before a shift, that is triggered only if the shift is not full. That takes care of what we wanted to do which is notify users of a group if one of the shifts is not full yet. Same for 24 hours.
In general remind people that their shift is about to start. To do that we simply add a reminder at the same time as we add a ShiftUser, and remove it when we remove the shift user too.
Anyway, plenty of cool things can be done with this. Nice job !

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,900 @@
...
+            return self.success(data=reminders)
+
+        try:
+            reminder_id = int(reminder_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid (scalar integer) reminder ID. ")
+
+        # the default is to reminder on an object
+        if associated_resource_type.lower() == "sources":
+            try:
```
Theodlz, line 144: You have a try expect for each of the if statement, but can't it just be replaced by one try except around all the if statements ?

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,900 @@
...
+                )
+            reminder_resource_id_str = str(reminder.shift_id)
+        # add more options using elif
+        else:
+            return self.error(
+                f'Unsupported associated_resource_type "{associated_resource_type}".'
+            )
+
+        if reminder_resource_id_str != resource_id:
+            return self.error(
```
Theodlz, line 191: Can't you do that at the same time as you query the reminders instead? I'm not sure that this should be an error. Doesn't it just mean that there is no reminder for the resource ?

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,900 @@
...
+            DBSession().add(
+                UserNotification(
+                    user=user,
+                    text=text_to_send,
+                    notification_type="mention",
+                    url=url_endpoint,
+                )
+            )
+
+        DBSession().add(reminder)
```
Theodlz, line 418: Might be nice to have that logic in a separate function so we can create reminders from other handlers (like automatically creating a reminder when a user is added to a shift).

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,900 @@
...
+        self.verify_and_commit()
+        for user in users:
+            self.flow.push(user.id, "skyportal/FETCH_NOTIFICATIONS", {})
```
Theodlz, line 421: Same comment as in the code of the service, pushing notifications to the frontend is handled in the UserNotification model i believe.

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,900 @@
...
+
```
Theodlz, line 422: Why do we need to do that exactly ? You intend on refreshing stuff on the frontend when it is added

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,900 @@
...
+            self.push_all(
+                action='skyportal/REFRESH_GCNEVENT',
+                payload={'gcnEvent_dateobs': reminder.gcn.dateobs},
+            )
+        elif isinstance(reminder, ReminderOnSpectrum):
+            self.push_all(
+                action='skyportal/REFRESH_SOURCE_SPECTRA',
+                payload={'obj_internal_key': reminder.obj.internal_key},
+            )
+        elif isinstance(reminder, ReminderOnShift):
```
Theodlz, line 439: I'm saying this in the comment above because I am a bit worried about that. Refreshing the shifts everytime a reminder is added on a shift seems like it will happen all the time no?

Theodlz (commented):
All of this seems super useful and reusable for other features that need to check if some conditions are met like X hours before something that is planned happened. Most of it is great, but I have a bunch of comments too. Nicely done.

services/reminders/reminders.py
mcoughlin: yes.

services/reminders/reminders.py
mcoughlin: I think we usually keep DB entries like for completed requests.

services/reminders/reminders.py
mcoughlin: Good point. We should discuss how this fits with your Notification work (Which should probably go in first).

services/reminders/reminders.py
mcoughlin: Yeah removing.

skyportal/handlers/api/reminder.py
mcoughlin: This is a holdover from the comments stuff.

skyportal/handlers/api/reminder.py
mcoughlin: Good point.

skyportal/handlers/api/reminder.py
mcoughlin: Removed!

skyportal/handlers/api/reminder.py
mcoughlin: yeah removed.

skyportal/models/group_joins.py
```diff
@@ -87,6 +96,14 @@
     accessible_by_group_admins & GroupCommentOnSpectrum.read
 )
 
+GroupReminderOnSpectrum = join_model(
+    "group_reminders_on_spectra", Group, ReminderOnSpectrum
+)
+GroupReminderOnSpectrum.__doc__ = "Join table mapping Groups to CommentOnSpectrum."
```
thomasculino, line 102:
```suggestion
GroupReminderOnSpectrum.__doc__ = "Join table mapping Groups to ReminderOnSpectrum."
```

skyportal/models/group_joins.py
```diff
@@ -87,6 +96,14 @@
...
+GroupReminderOnSpectrum.delete = GroupCommentOnSpectrum.update = (
```
thomasculino, line 103:
```suggestion
GroupReminderOnSpectrum.delete = GroupReminderOnSpectrum.update = (
```

thomasculino (commented):
Small changes to the group joins.

Theodlz:
Hey @mcoughlin. I can't ask for a review because you opened the PR. But it would be great if I could have your opinion on the changes I made. Best way to test it is to create some reminders on a source, gcnevent and/or shift. If you're happy with it, and if it works the way you imagined it to do, we can ask a review from Guy. Also, we need to add a frontend element to add reminders on spectra. Totally doable, not very hard but the PR is already really long so it could be added in a following PR. I'll let you and @guynir42 decide once we start getting some reviews on this.

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+                            reminder_id:
+                              type: integer
+                              description: New reminder ID
+        """
+        data = self.get_json()
+
+        reminder_text = data.get("text")
+        next_reminder = data.get("next_reminder")
+        next_reminder = arrow.get(next_reminder).datetime.replace(tzinfo=None)
+        reminder_delay = data.get("reminder_delay", 0)
```
guynir42, line 403: Is zero a reasonable default? What would happen in that case?

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+        with self.Session() as session:
+            group_ids = data.pop('group_ids', None)
+            if not group_ids:
+                group_ids = [g.id for g in self.current_user.accessible_groups]
+            try:
+                groups = session.scalars(
+                    Group.select(self.current_user).where(Group.id.in_(group_ids))
+                ).all()
+            except AccessError:
+                return self.error('Could not find any accessible groups.', status=403)
```
guynir42, line 414: This should not be a try/except, but a test if the set of group IDs is the same as that given. 
And I would rephrase the error to "cannot find some of the requested groups`

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+            user_ids = data.pop('user_ids', None)
+            if not user_ids:
+                users = [self.associated_user_object]
+            else:
+                try:
+                    users = session.query(User).filter(User.id.in_(user_ids)).all()
+                except AccessError:
+                    return self.error(
+                        'Could not find any accessible users.', status=403
+                    )
```
guynir42, line 425: Use `select` instead of `query` and test as for the groups.

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+                resource_id,
+                reminder_text,
+                groups,
+                users,
+                is_bot_reminder,
+                next_reminder,
+                reminder_delay=reminder_delay,
+                number_of_reminders=number_of_reminders,
+            )
+        DBSession().add_all(reminders)
```
guynir42, line 440:
```suggestion
             session.add_all(reminders)
```

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+            text_to_send = f"*@{self.associated_user_object.username}* created a reminder on *shift {resource_id}*"
+            url_endpoint = "/shifts"
+            action = 'skyportal/REFRESH_REMINDER_SHIFT'
+            payload = {'id': resource_id}
+        else:
+            return self.error(f'Unknown resource type "{associated_resource_type}".')
+
+        ws_flow = Flow()
+        for user in users:
+            DBSession().add(
```
guynir42, line 471: Use the `session`

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+                UserNotification(
+                    user=user,
+                    text=text_to_send,
+                    notification_type="mention",
+                    url=url_endpoint,
+                )
+            )
+            ws_flow.push(user.id, "skyportal/FETCH_NOTIFICATIONS")
+
+        self.verify_and_commit()
```
guynir42, line 481:
```suggestion
        session.commit()
```

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+        try:
+            reminder_id = int(reminder_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid (scalar integer) reminder ID. ")
+
+        if associated_resource_type.lower() == "source":
+            schema = Reminder.__schema__()
+            try:
+                c = Reminder.get_if_accessible_by(
+                    reminder_id, self.current_user, mode="update", raise_if_none=True
```
guynir42, line 556: Same comment as in POST.

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+        # add more options using elif
+        else:
+            return self.error(
+                f'Unsupported associated_resource_type "{associated_resource_type}".'
+            )
+
+        data = self.get_json()
+        group_ids = data.pop("group_ids", None)
+        data['id'] = reminder_id
+        attachment_bytes = data.pop('attachment_bytes', None)
```
guynir42, line 607: Do we have attachments?

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+            )
+
+        if group_ids is not None:
+            try:
+                groups = Group.get_if_accessible_by(
+                    group_ids, self.current_user, raise_if_none=True
+                )
+            except AccessError:
+                return self.error('Could not find any accessible groups.', status=403)
+            c.groups = groups
```
guynir42, line 635: Same comments regarding getting the groups as in POST.

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+
+        try:
+            reminder_id = int(reminder_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid (scalar integer) reminder ID. ")
```
guynir42, line 550: Since this only addresses existing reminders, it might be better to use PATCH instead of PUT.

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+        """
+
+        try:
+            reminder_id = int(reminder_id)
+        except (TypeError, ValueError):
+            return self.error("Must provide a valid (scalar integer) reminder ID.")
+
+        if associated_resource_type.lower() == "source":
+            try:
+                c = Reminder.get_if_accessible_by(
```
guynir42, line 714: `get_if_accessible_by` -> `get`

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+        elif isinstance(c, ReminderOnShift):
+            shift_id = c.shift.shift_id
+
+        if reminder_resource_id_str != resource_id:
+            return self.error(
+                f'Reminder resource ID does not match resource ID given in path ({resource_id})'
+            )
+
+        DBSession().delete(c)
+        self.verify_and_commit()
```
guynir42, line 772: Use `self.Session()` and `session.commit()`.

skyportal/models/reminder.py
```diff
@@ -0,0 +1,214 @@
...
+        doc="Boolean indicating whether reminder was posted via a bot (token-based request).",
+    )
+
+    next_reminder = sa.Column(
+        sa.DateTime, nullable=False, index=True, doc="Next reminder."
+    )
+
+    reminder_delay = sa.Column(
+        sa.Float, nullable=False, index=True, doc="Delay until next reminder in days."
+    )
```
guynir42, line 55: Why do we need an index on this?

skyportal/tests/api/test_reminders.py
```diff
@@ -0,0 +1,158 @@
...
+    status, data = api('GET', endpoint, token=token)
+    assert status == 200
+    assert data['status'] == 'success'
+    data = data['data']['reminders']
+    assert data[-1]['text'] == reminder_text
+    assert data[-1]['next_reminder'] == next_reminder.strftime("%Y-%m-%dT%H:%M:%S")
+    assert data[-1]['reminder_delay'] == reminder_delay
+    assert data[-1]['number_of_reminders'] == number_of_reminders
+
+    time.sleep(15)  # wait for the reminder to be sent
```
guynir42, line 38: Having multiple sleep(15) in the tests is not great.

skyportal/tests/frontend/test_reminders.py
```diff
@@ -0,0 +1,192 @@
+import os
+import uuid
+import time
+
+from skyportal.tests import api
+from datetime import date, timedelta, datetime
+from selenium.webdriver.common.keys import Keys
+
+
+def post_and_verify_reminder(endpoint, token):
```
guynir42, line 10: Try to import this from the backend test.

skyportal/tests/frontend/test_reminders.py
```diff
@@ -0,0 +1,192 @@
...
+    status, data = api('GET', endpoint, token=token)
+    assert status == 200
+    assert data['status'] == 'success'
+    data = data['data']['reminders']
+    assert data[-1]['text'] == reminder_text
+    assert data[-1]['next_reminder'] == next_reminder.strftime("%Y-%m-%dT%H:%M:%S")
+    assert data[-1]['reminder_delay'] == reminder_delay
+    assert data[-1]['number_of_reminders'] == number_of_reminders
+
+    time.sleep(20)  # wait for the reminder to be sent
```
guynir42, line 40: even more sleep??

skyportal/tests/frontend/test_reminders.py
```diff
@@ -0,0 +1,192 @@
...
+        driver.wait_for_xpath('//*[@data-testid="AddIcon"]')
+    )
+    reminder_text_2 = str(uuid.uuid4())
+    driver.wait_for_xpath('//*[@id="root_text"]').send_keys(reminder_text_2)
+    next_reminder = (datetime.now() + timedelta(days=1)).strftime("%m/%d/%YT%I:%M %p")
+    driver.wait_for_xpath('//*[@id="root_next_reminder"]').send_keys(
+        next_reminder[0:11]
+    )
+    driver.wait_for_xpath('//*[@id="root_next_reminder"]').send_keys(Keys.TAB)
+    driver.wait_for_xpath('//*[@id="root_next_reminder"]').send_keys('01:01')
```
guynir42, line 76: Is this the right time?

skyportal/tests/frontend/test_reminders.py
```diff
@@ -0,0 +1,192 @@
...
+# frontend for the reminders on spectra is not implemented yet
+
+
+def test_reminder_on_gcn(driver, super_admin_user, super_admin_token):
+    datafile = f'{os.path.dirname(__file__)}/../../../data/GW190814.xml'
+    with open(datafile, 'rb') as fid:
+        payload = fid.read()
+    data = {'xml': payload}
+
+    status, data = api('POST', 'gcn_event', data=data, token=super_admin_token)
```
guynir42, line 168: This is for a different PR, but I would consider making a GCN event fixture, and making its scope be the whole testing suite, so that it only ever needs to be loaded once and used by all relevant tests.

static/js/components/Reminders.jsx
```diff
@@ -0,0 +1,449 @@
...
+    if (formData.number_of_reminders <= 0) {
+      errors.number_of_reminders.addError(
+        "Number of reminders must be greater than 0"
+      );
+    }
+    if (formData.reminder_delay <= 0) {
+      errors.reminder_delay.addError(
+        "Reminder delay must be greater than 0 day"
+      );
+    }
```
guynir42, line 169: If the next reminder is in the future, and `number_of_reminders=1` then this field is actually irrelevant.

static/js/components/Reminders.jsx
```diff
@@ -0,0 +1,449 @@
...
+          <Close />
+        </IconButton>
+      ) : null}
+    </MuiDialogTitle>
+  )
+);
+
+const NewReminder = ({ resourceId, resourceType, handleClose }) => {
+  const dispatch = useDispatch();
+  const defaultDate = dayjs().utc().format("YYYY-MM-DDTHH:mm:ssZ");
```
guynir42, line 135: This default date will always be in the past by the time you finish filling the form. 
Maybe make the reminder in 1 day from now by default?

guynir42 (commented):
Phew. 

I will check out the code and play around with the front end after this round of comments is finished.

services/reminders/reminders.py
```diff
@@ -0,0 +1,141 @@
...
+            if reminder_type == Reminder:
+                text_to_send = (
+                    f"Reminder of source *{reminder.obj_id}*: {reminder.text}"
+                )
+                url_endpoint = f"/source/{reminder.obj_id}"
+            elif reminder_type == ReminderOnSpectrum:
+                text_to_send = (
+                    f"Reminder of spectrum *{reminder.spectrum_id}*: {reminder.text}"
+                )
+                url_endpoint = f"/source/{reminder.spectrum_id}"
```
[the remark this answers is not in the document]
Theodlz: I don't think there is any routing to a spectrum page right? I think that we want the user to be redirected to the source's page that will have that spectrum on it.  But then it might be incorrect to have the spectrum_id, but we should have an obj_id at the end of the URL. I'll look into that.

skyportal/tests/frontend/test_reminders.py
Theodlz: Great idea. I'll try to open a PR doing just that once this one gets in then.

static/js/components/Reminders.jsx
Theodlz: What do you mean by that? That we can calculate the delay based on the difference between the current time and the next reminder? To me, here's how I understood the feature: the user chooses the next_reminder(that is the date of the first time it will be triggered), the number of reminder, and the delay in between the reminders. So I don't see why this field is irrelevant.

static/js/components/Reminders.jsx
Theodlz: Makes sense! I'll do that then.

skyportal/tests/frontend/test_reminders.py
Theodlz: The time doesn't really matter for this part of the test. Here, we don't verify if the reminder gets triggered or not, just if it gets added and that we receive the notification telling us a new reminder was created.

skyportal/tests/frontend/test_reminders.py
Theodlz: I'll try to get rid of it the best I can.

Theodlz: It's just that we have the reminder in 5 seconds from now, + the service that iterates over all reminders every 5 seconds. So in the worse case scenario we need to wait around 10 seconds, best case scenario around 5 seconds. So a sleep will be needed. Maybe what I can do is a 5 seconds sleep, and then query the API `n` times to see if it was sent?

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+                reminders = (
+                    ReminderOnShift.query_records_accessible_by(self.current_user)
+                    .filter(ReminderOnShift.shift_id == resource_id)
+                    .all()
+                )
+            else:
+                return self.error(
+                    f'Unsupported associated resource type "{associated_resource_type}".'
+                )
+            self.verify_and_commit()
```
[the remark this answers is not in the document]
Theodlz: How does that work with the `query_records_accessible_by`. Or more generally, with the new syntax where/when/how do I check if the user has the read permission when querying? @guynir42

skyportal/handlers/api/reminder.py
```diff
@@ -0,0 +1,796 @@
...
+                f'Unsupported associated_resource_type "{associated_resource_type}".'
+            )
+
+        if reminder_resource_id_str != resource_id:
+            return self.error(
+                f'Reminder resource ID does not match resource ID given in path ({resource_id})'
+            )
+
+        if not reminder.attachment_bytes:
+            return self.success(data=reminder)
```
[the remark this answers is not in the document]
Theodlz: Right. I'll remove that.

skyportal/app_server.py
```diff
@@ -220,6 +221,14 @@ def log_request(self, handler):
...
         GcnEventSurveyEfficiencyHandler,
     ),
+    (
+        r'/api/(source|spectra|gcn_event|shift)/([0-9A-Za-z-_\.\+]+)/reminders',
+        ReminderHandler,
+    ),
+    (
+        r'/api/(source|spectra|gcn_event|shift)/([0-9A-Za-z-_\.\+]+)/reminders(/[0-9]+)?',
+        ReminderHandler,
+    ),
```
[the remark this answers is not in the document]
Theodlz: If I have just the first one, i get an error while posting, telling me that the handler received 4 args when it needs just 3. This takes care of that by separating requests with 3 and with 4 args. But if that can be fixed any other way that could be great.

skyportal/handlers/api/reminder.py
Theodlz: Done.

skyportal/handlers/api/reminder.py
Theodlz: No! Removed

skyportal/models/reminder.py
Theodlz: Not needed I believe. I'll remove it.

skyportal/tests/frontend/test_reminders.py
Theodlz: I'm having trouble importing it. This function used the API method, and once it is imported it can't access the API method and it gives me an error:
```
TypeError: 'module' object is not callable
```

Theodlz:
Hi @guynir42. Thanks a lot for the review, this PR is really long sorry about that...
I addressed some of your comments, and I had some questions/answers for some of them. I'll let you have a look.

My main question is how do we handle the read/update permissions when querying things in the handler with the new syntax. I just want to make sure that a user can't access/modify/delete other users reminders that he is not supposed to see

services/reminders/reminders.py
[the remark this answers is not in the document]
guynir42: There's a way to (maybe) direct to a spectrum via the "manage data" page. 
Although there could be a way to navigate directly to the spectroscopy plot (I'm not sure how you'd do that).

skyportal/app_server.py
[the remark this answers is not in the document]
guynir42: What happens if you remove the first line? You may need to add a dummy input variable to some of the handler methods that is `None` by default and never used.

skyportal/tests/frontend/test_reminders.py
guynir42: I'm guessing you import `skyportal/tests/api` as a module, then that name overrides the function `api`. Try doing something like `from skyportal/tests/api import post_and_verify_reminder`.

skyportal/tests/frontend/test_reminders.py
guynir42: I would first off make the reminder 1 second instead of 5, and then do a loop with 1 second intervals. That would probably reduce the wait times in most cases.

static/js/components/Reminders.jsx
guynir42: If I set a reminder with `number=1` and set it to five minutes in the future. The 1 day delay never gets called because after those five minutes the reminder is finished. 
I guess it doesn't matter much for this form, because there's a reasonable default in the `delay` field so if the user ignores this it doesn't hurt.

services/reminders/reminders.py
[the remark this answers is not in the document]
Theodlz: No idea how to directly navigate to the spectroscopy plot. I'll look into that. If I can't figure it out, I'll go with the first option.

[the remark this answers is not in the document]
Theodlz: Adding some anchors means adding additional routing, which can easily get complicated.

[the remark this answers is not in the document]
Theodlz: Same with the manage_data, we need some routing if we want to directly open the right spectra and such. Should this be done in this PR?

skyportal/handlers/api/reminder.py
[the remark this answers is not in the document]
Theodlz: great! Thanks a lot for the explanations. Any differences between the `Reminder.select(user)` syntax and the new `get` you mentioned in one of your comments above this one?

services/reminders/reminders.py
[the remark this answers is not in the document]
guynir42: Not critical. If it isn't very easy to do, just leave it pointing at the source page. Just make sure the link uses the `obj_id` not the `spectrum_id`.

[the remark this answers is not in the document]
Theodlz: Ok! Will do

Theodlz:
Little note for reviewers: all is green but some tests appear to be flaky from time to time. I want to improve them before merging anything.

mcoughlin:
Thanks @Theodlz. Looks really good, thanks for addressing everyone's concerns.

PR #3055, opened 2022-06-07
[Analysis Epic] Analysis results API & Mock Service

profjsb:

PR #3100, opened 2022-06-16
DNM: fix path to discover skyportal services in test config

profjsb:
The discovery path for skyportal services was incorrect. This is fixed. Also turn off some of those services, which should not directly be needed for testing.

> **Note**: it actually was correct and is needed for a lot of the testing. Will fix here but DNM now.

### DIFF
--- a/test_config.yaml
+++ b/test_config.yaml
@@ -55,7 +55,11 @@ testing: true
 services:
   paths:
     - ./baselayer/services
-    - ./skyportal/services
+    - ./services
+  disabled:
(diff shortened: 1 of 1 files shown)


stefanv (approved):

mcoughlin (approved):
Oh the things we overlook.

guynir42 (approved):

profjsb:
DNM

profjsb:
the test config will be updated in #3055 (it's actually ok as is now)

PR #3101, opened 2022-06-17
Telescope Handler - Performance Improvements and Bugfix

Theodlz:
This PR does 2 things:

- reduce the number of calls to the functions to calculate the next morning astronomical and next evening astronomical of a telescope (used to just call the same function multiple times in a row, so here instead we call it once and save its result in a variable, on which we perform operations)
- add a try-except statement if the value returned by those 2 functions is not what we expect, or if nothing can be returned. If that happens, we do the same as what we do for a non-fixed telescope: we set the next morning and evening astronomical to false and continue.

I don't believe that an error should be raised. Calculating these values was an addition to the telescope map, and I think it should return telescopes even if it can't be calculated for one of them, rather than raise an error and return nothing. Not being able to return telescopes just because of some values we need for display purposes seems unnecessary to me.

*This PR has been opened to try to help with some issues encountered in production.*


skyportal/handlers/api/telescope.py
```diff
@@ -135,23 +135,32 @@ def get(self, telescope_id=None):
...
+            ):
+                try:
+                    morning = telescope.next_twilight_morning_astronomical()
+                    evening = telescope.next_twilight_evening_astronomical()
+                    if morning is not None and evening is not None:
+                        is_night_astronomical = bool(morning.jd < evening.jd)
+                        morning = morning.iso
+                        evening = evening.iso
+                except Exception:
+                    morning = False
```
mcoughlin, line 155: Since you set the default above, you do not need to reset morning and such right?

skyportal/handlers/api/telescope.py
```diff
@@ -135,23 +135,32 @@ def get(self, telescope_id=None):
...
+                    evening = False
+                    is_night_astronomical = False
+                    continue
```
mcoughlin, line 158: Did you mean to break the loop with a continue?

mcoughlin (changes requested):
@Theodlz Will the continue mean the list is the wrong length?

skyportal/handlers/api/telescope.py
Theodlz: No ! I meant to just continue without raising an error, should i use `pass` instead ?

skyportal/handlers/api/telescope.py
Theodlz: If the exception happens for the evening, i want the morning to also be false, so i reset it all.

Theodlz:
What i had in mind is that if there is an error, morning evening and is_night are set to false, but it continues iterating through the list.

skyportal/handlers/api/telescope.py
mcoughlin: ah yeah pass is what you want then, but you shouldn't need anything.

stefanv: pass is a no-op, so only used in the context of defining an empty function

stefanv: Michael is correct that the continue should be removed

stefanv: By catching the exception, you are already suppressing it. You can re-raise an exception from a catch clause, if that is needed, otherwise operations will continue without any. Another useful trick is the finally clause, which will execute regardless of whether an exception was raised or not.

Theodlz: Ok ! I just removed the `continue` then.

mcoughlin (approved):

skyportal/handlers/api/telescope.py
```diff
@@ -135,23 +135,31 @@ def get(self, telescope_id=None):
...
-                temp['next_twilight_evening_astronomical'] = False
-
+            morning = False
+            evening = False
+            is_night_astronomical = False
+            if (
+                telescope.fixed_location
+                and telescope.lon
+                and telescope.lat
+                and telescope.elevation
```
profjsb, line 145: if any of these numbers are correct and assigned to 0.0 then this statement will fail. I bet people will put in elevation = 0 if they dont know it. I think you mean:
```suggestion
                and telescope.lon is not None
                and telescope.lat is not None
                and telescope.elevation is not None
```

skyportal/handlers/api/telescope.py
```diff
@@ -135,23 +135,31 @@ def get(self, telescope_id=None):
         telescopes = []
         for telescope in data:
```
profjsb, line 136: Not sure if `telescope` will always have a `.to_dict()` method? 
```suggestion
        for telescope in data:
              if telescope is None:
                   continue
```

profjsb (changes requested):
Looks good. Few small changes suggested inline.

skyportal/handlers/api/telescope.py
Theodlz: Then should I check if it is not None `and` not 0.0 @profjsb ?

guynir42 (approved):
Yeah, I agree plotting without morning/evening is better than not plotting at all. 
Code looks good.

PR #3105, opened 2022-06-17
Paginate GCN events

mcoughlin:
This PR adds pagination to the GCN events.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -185,8 +185,17 @@ class GcnEventHandler(BaseHandler):
                 schema: Error
         """
 
-        page_number = int(self.get_query_argument("pageNumber", 1))
-        n_per_page = int(self.get_query_argument("numPerPage", 100))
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/gcn.py
```diff
@@ -184,6 +184,10 @@ def get(self, dateobs=None):
               application/json:
                 schema: Error
         """
+
+        page_number = int(self.get_query_argument("pageNumber", 1))
+        n_per_page = int(self.get_query_argument("numPerPage", 100))
```
guynir42, line 189: Add try catch with self error in case inputs are not integers.

guynir42, line 189: Add try catch with self error in case inputs are not integers.

guynir42 (approved):
LGTM

skyportal/handlers/api/gcn.py
mcoughlin: done

mcoughlin: done

PR #3107, opened 2022-06-19
Standardized button component/documentation

jadalilleboe:
This adds the infrastructure and documentation for standardized button usage across SkyPortal.

It is a little different from my initial proposal, I've made one general use Button component to be used for all buttons. Boolean props are passed for designating the style of the button, see documentation for more info/code examples. 

I looked into [creating new button variants](https://mui.com/material-ui/customization/theme-components/#creating-new-component-variants) instead of a new component, but that did not work well for the API I was envisioning (pass in a single prop to determine color/variant). With creating new variants, it is difficult to get the style of a contained mui button without explicitly passing that prop in. So I went with the new component solution, i think it is a simpler API (the developer does not fuss with color/variant props, Button.jsx deals with all that)

I haven't changed any buttons in the app yet, I wanted to get thoughts on the API first.


profjsb:
This PR looks good to go. What do you think @guynir42 ? We're not using in the ./Button.jsx yet but once this is merged we can (slowly) replace buttons as such.

static/js/components/Button.jsx
```diff
@@ -0,0 +1,54 @@
...
+const Button = ({ primary, secondary, async, ...muiButtonProps }) => {
+  if (muiButtonProps.startIcon) {
+    throw new Error(
+      "Error: startIcon used in Button props. Please use endIcon as specified in the SkyPortal style documentation: https://skyportal.io/docs/styling.html#buttons"
+    );
+  }
+  if (muiButtonProps.variant || muiButtonProps.color) {
+    throw new Error(
+      "Error: do use variant or color prop in Button, see button documentation for recommended usage https://skyportal.io/docs/styling.html#buttons"
+    );
```
guynir42, line 15: I wonder why we should not allow the user to choose a color. What if there's a good reason to want a customization on top of the defaults offered here?

jadalilleboe: I guess my thought was to have an opinionated set of colors (i.e. the default buttons), and if there was need for a different type of button with a different color, a new button type could be added. So you can define the styles for a certain button type in one place and not have a bunch of one-off button types.

guynir42: That's a great idea in general. I am not sure if we would want to have people making up green buttons or something like that... on the other hand sometimes you just want a different color for a good reason. I'm not sure what the best practice would be here. 
My hunch is that relying on the programmer's laziness to just use the defaults is ok, and let the programmer that really wants their red button to be able to use it.

jadalilleboe: Agreed, if we notice many one-off button types are similar we can add more to variants to this Button.jsx component.

guynir42 (approved):

guynir42:
It looks like there are a few test failures. The analysis page one is failing for everyone but I think it is worth investigating the other ones.

jadalilleboe:
@guynir42 test failures unrelated I think

guynir42:
@jadalilleboe ok you can merge

PR #3110, opened 2022-06-19
Basic plots to track allocation usage

mcoughlin:
This PR makes an endpoint to make some basic plots to track allocation usage.

### DIFF
--- a/skyportal/handlers/api/allocation.py
+++ b/skyportal/handlers/api/allocation.py
@@ -1,8 +1,13 @@
-from ..base import BaseHandler
-from ...models import DBSession, Group, Allocation, Instrument
-from baselayer.app.access import auth_or_token, permissions
 from marshmallow.exceptions import ValidationError
 import sqlalchemy as sa
+import io
+import matplotlib
+import matplotlib.pyplot as plt
+
+from baselayer.app.access import auth_or_token, permissions
+
+from ..base import BaseHandler
+from ...models import DBSession, Group, Allocation, Instrument
 
 
 class AllocationHandler(BaseHandler):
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/allocation.py
```diff
@@ -222,3 +228,139 @@ def delete(self, allocation_id):
         self.verify_and_commit()
         self.push_all(action='skyportal/REFRESH_ALLOCATIONS')
         return self.success()
+
+
+class AllocationAnalysisHandler(BaseHandler):
```
guynir42, line 233: Maybe `AllocationReportHandler` makes it more distinct from "analysis" which is a whole other thing?

skyportal/handlers/api/allocation.py
```diff
@@ -222,3 +228,139 @@ def delete(self, allocation_id):
...
+        def make_autopct_observations(values):
+            def my_autopct(pct):
+                total = sum(values)
+                if np.isnan(pct):
+                    val = 0
+                else:
+                    val = int(round(pct * total / 100.0))
+                return f'{pct:.0f}%  ({val:d} Observations)'
+
+            return my_autopct
```
guynir42, line 314: Is there an easy way to not duplicate all these functions?

guynir42 (approved):
Minor comments, but looks good.

skyportal/handlers/api/allocation.py
mcoughlin: Yes good point.

skyportal/handlers/api/allocation.py
mcoughlin: Done.

PR #3111, opened 2022-06-20
Allow for followup request default

mcoughlin:
This PR allows the user to set a default follow up request.

Closes fritz issue: https://github.com/fritz-marshal/fritz/issues/415


guynir42 (approved):
LGTM

PR #3112, opened 2022-06-20
SourceExist radius check is in degrees, not arcseconds

mcoughlin:
This PR uses a much more reasonable check for nearby sources in the SourceExists handler.


thomasculino (approved):
Looks fine

PR #3113, opened 2022-06-20
Improvement of the comment component

thomasculino:
Currently, when we write a long comment on skyportal, we have to scroll to the side to see all of it:
long_comment_before_changes
 
I fixed it by going to the next line when the text does not have any space left:
image

### DIFF
--- a/static/js/components/CommentList.jsx
+++ b/static/js/components/CommentList.jsx
@@ -86,10 +86,32 @@ const useStyles = makeStyles(() => ({
     marginRight: "1em",
   },
   commentMessage: {
-    maxWidth: "25em",
+    maxWidth: "35em",
     "& > p": {
       margin: "0",
     },
+    wordWrap: "break-word",
+  },
+  commentMessageShift: {
+    maxWidth: "47em",
+    "& > p": {
+      margin: "0",
+    },
+    wordWrap: "break-word",
+  },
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
@thomasculino also working thinking about if there are other places to implement this.

Theodlz (commented):
Looks good. Just one question before approving. I see that this applies just for the comments on shifts correct ? The CSS classes have "shift" in their name that's why. But does this problem occur on other comments as well ? In that case why not fixing them too ? And if its just on the shift comments, maybe the problem has to do with something else, like the styling of the div where the component is called on the Shift Page maybe ?

thomasculino:
> Looks good. Just one question before approving. I see that this applies just for the comments on shifts correct ? The CSS classes have "shift" in their name that's why. But does this problem occur on other comments as well ? In that case why not fixing them too ? And if its just on the shift comments, maybe the problem has to do with something else, like the styling of the div where the component is called on the Shift Page maybe ?

This PR fixes all types of comment. Not only shifts. Some of the styles have the 'Shift' suffix because the length of the comment component on the shift page is bigger than the other ones. Basically, styles without the 'Shift' suffix are used on the source and gcn event page

Theodlz:
@thomasculino i'll merge once green. One set of tests didn't run yet.

PR #3115, opened 2022-06-20
Robustify GitHub Actions by adding some patience

profjsb:
Wait longer for timeouts and for load_demo_data to start.

Note: we're doing a little baselayer patch here just for testing purposes and before the same PR gets merged (https://github.com/cesium-ml/baselayer/pull/329). Once that PR is merged the `sed` command here will have no affect but not break anything.


stefanv:
Easier would be to just give the tools a longer timeout. Opportunistic waiting (polling) is better than implicit waits, like the 60s added here.

guynir42:
@profjsb should we proceed with this?

Theodlz:
It looks like these longer sleeps have been added in another PR. Closing.

PR #3116, opened 2022-06-21
Notifications: More notification types, more settings per notification type

Theodlz:
### PREVIEW
notification_framework_beta


Theodlz:
Hi @mcoughlin, tagging you first to have your opinion on the changes I made since you tested it.

skyportal/models/user_notification.py
```diff
@@ -21,9 +22,28 @@
...
+import gcn
+from sqlalchemy import or_
+
+from skyportal.models import Shift, ShiftUser
 
 _, cfg = load_env()
 
+log = make_log('notifications')
+
+account_sid = cfg["twilio.sms_account_sid"]
```
mcoughlin, line 36: Will this still work for whatever is happening here? https://github.com/skyportal/skyportal/pull/3091

skyportal/models/user_notification.py
```diff
@@ -21,9 +22,28 @@
...
+auth_token = cfg["twilio.sms_auth_token"]
+from_number = cfg["twilio.from_number"]
+client = None
+if account_sid and auth_token and from_number:
+    client = TwilioClient(account_sid, auth_token)
+
+email = False
+if cfg["email_service"] == "sendgrid" or cfg["email_service"] == "smtp":
```
mcoughlin, line 44: Should we have a separate flag in the config like enabling email? I.e. you can have it all setup fine but just turn it off with a email=false?

skyportal/models/user_notification.py
```diff
@@ -90,30 +114,172 @@ def send_slack_notification(mapper, connection, target):
     slack_microservice_url = (
         f'http://127.0.0.1:{cfg.get("slack.microservice_port", 64100)}'
     )
+    ressource_type = None
```
mcoughlin, line 117: resource

skyportal/models/user_notification.py
```diff
@@ -90,30 +114,172 @@ def send_slack_notification(mapper, connection, target):
...
+            headers={'Content-Type': 'application/json'},
+        )
+    except Exception as e:
+        log(f"Error sending slack notification: {e}")
+
+
+@event.listens_for(UserNotification, 'after_insert')
+def send_email_notification(mapper, connection, target):
+
+    if not email:
```
mcoughlin, line 147: should we combine some of these as like a long set in one if?

skyportal/models/user_notification.py
```diff
@@ -90,30 +114,172 @@ def send_slack_notification(mapper, connection, target):
...
+        return
+
+    prefs = target.user.preferences.get('notifications')
+
+    if not prefs:
+        return
+
+    ressource_type = None
+    if "favorite_sources" not in target.notification_type:
+        ressource_type = target.notification_type
```
mcoughlin, line 166: resource :)

skyportal/models/user_notification.py
```diff
@@ -90,30 +114,172 @@ def send_slack_notification(mapper, connection, target):
...
+                body=body,
+            )
+        except Exception as e:
+            log(f"Error sending email notification: {e}")
+
+
+@event.listens_for(UserNotification, 'after_insert')
+def send_sms_notification(mapper, connection, target):
+    if client is None:
+        return
```
mcoughlin, line 221: Can these if blocks be reused between notification types?

mcoughlin (commented):
@Theodlz Really nice work so far, and the interface is really good. My only substantive comment is that if there are ways to simplify the "to notify or not to notify, that is the question" type logic, including making a function to reuse between the different notification types, that might be nice. On the other hand, all this stuff is almost necessarily convoluted, so I thought you did a nice job.

skyportal/models/user_notification.py
Theodlz: Right now it's what the email service is for kinda, if you don't specify one (that's what is done by default) it won't use it, and otherwise, it will. But maybe an extra flag to just deactivate it without having to modify your email config could be a nice addition?

mcoughlin: Yeah that's what I am thinking!

skyportal/models/user_notification.py
Theodlz: What do you mean by that ? I mean it should work, the key exists, but it has no value by default. What should we do, add a default null or ~ value in the config ? My PR aside, that's a problem in general with empty fields in the config.

skyportal/models/user_notification.py
Theodlz: You're right, some of them can be reused. I'll add a separate function then.

skyportal/models/user_notification.py
mcoughlin: Yeah I think it's fine.

skyportal/models/user_notification.py
Theodlz: Hmmm, I'm actually not so sure if it's worth doing. Most of those if are different actually. Not os many of them can be reused. Depending on the notification type we check different stuff.

Theodlz: checking if the user has preferences and such is redundant, so I'll create a function at least for that.

Theodlz: Ok that's done.

guynir42:
I'm going to review now. Thanks for clearing all of this up. And good work on the PR description, this makes it a lot easier for me to know what I'm going to be looking at. 
Let me know if you also want me to test locally with the email and SMS alerts.

skyportal/handlers/api/gcn.py
```diff
@@ -577,3 +579,17 @@ def delete(self, dateobs, localization_name):
...
+    def get(self):
+        '''
+        This method return the list of notices from the config file
+        ---
+        description: Returns the list of notices
+        tags:
+          - gcn_notices
+        '''
+        notices = cfg['gcn_notice_types']
+        return self.success(data=notices)
```
guynir42, line 595: I think you may be able to get this without using a separate handler. Look at `static/js/ducks/config.js` and `skyportal/handlers/api/config_handler.py` .

static/js/components/GcnNoticeTypesSelect.jsx
```diff
@@ -0,0 +1,43 @@
...
+const GcnNoticeTypesSelect = (props) => {
+  const { selectedGcnNoticeTypes, setSelectedGcnNoticeTypes } = props;
+  const dispatch = useDispatch();
+  const gcn_notice_types = useSelector(
+    (state) => state.gcnNoticeTypes.gcnNoticeTypes
+  );
+
+  useEffect(() => {
+    dispatch(fetchGcnNoticeTypes());
+  }, []);
```
guynir42, line 17: If you add this to the `config_handler` it should be loaded into the state when the app is "hydrated".

guynir42 (commented):
Code looks good. Let me know if you would like to close the loop on testing the actual notifications.

skyportal/handlers/api/gcn.py
Theodlz: Oh that's great ! I've completely missed that config_handler. I'll try using that instead.

static/js/components/GcnNoticeTypesSelect.jsx
Theodlz: Sounds good !

Theodlz:
> I'm going to review now. Thanks for clearing all of this up. And good work on the PR description, this makes it a lot easier for me to know what I'm going to be looking at. Let me know if you also want me to test locally with the email and SMS alerts.

Hi @guynir42 . Thanks for the review ! I'll send you the twillio config, slack URL and add you to the slack group i created so you can test all of this without having to do everything from scratch.

skyportal/handlers/api/gcn.py
Theodlz: Done.

guynir42:
I'm getting some console errors and warnings when playing with the user preferences: 
```
 A component is changing the uncontrolled checked state of SwitchBase to be controlled.
Elements should not switch from uncontrolled to controlled (or vice versa).
Decide between using a controlled or uncontrolled SwitchBase element for the lifetime of the component.
The nature of the state is determined during the first render. It's considered controlled if the value is not `undefined`.
```

guynir42:
And also this warning: 
```
You have provided an out-of-range value `undefined` for the select (name="followupRequestAllocationSelect") component.
Consider providing a value that matches one of the available options or ''.
The available values are `-1`, `1`, `2`, `3`, `4`, `5`, `6`, `7`, `8`, `9`, `10`, `11`, `12`, `13`, `14`, `15`, `16`.
```

guynir42:
First impressions from playing with the new notifications: I think this is very cool and so far I've been able to get slack messages. 

I would add tooltips on **everything** in the user preferences page. It is not immediately obvious what the buttons do (especially the notification icon).

Theodlz:
Hi @guynir42, I added the tooltips you asked for.

> First impressions from playing with the new notifications: I think this is very cool and so far I've been able to get slack messages.
> 
> I would add tooltips on **everything** in the user preferences page. It is not immediately obvious what the buttons do (especially the notification icon).

Hi @guynir42, I added the tooltips you asked for.

mcoughlin (approved):
@Theodlz Thanks for fixing all that. This is going to be really good.

PR #3117, opened 2022-06-21
Restore the last and peak table columns

mcoughlin:
This PR adds back in the last detected and peak columns.


skyportal/models/obj.py
```diff
@@ -347,114 +346,6 @@ class Obj(Base, conesearch_alchemy.Point):
...
-        """UTC ISO date at which the object was last detected above a given S/N (3.0 by default)."""
-        detections = [
-            phot.iso
-            for phot in Photometry.query_records_accessible_by(user)
-            .filter(Photometry.obj_id == self.id)
-            .all()
-            if phot.snr is not None and phot.snr > PHOT_DETECTION_THRESHOLD
-        ]
-        return max(detections) if detections else None
-
```
guynir42, line 361: I'm not sure we want to get rid of these. They include the detection stats with permissions, where PhotStats are not.

skyportal/tests/api/test_sources.py
```diff
@@ -771,10 +772,9 @@ def test_object_last_detected(
     )
     assert status == 200
     assert data["status"] == "success"
-    assert (
-        data["data"]["last_detected_at"]
-        == arrow.get((90000.0 - 40_587) * 86400.0).isoformat()
-    )
+    assert arrow.get(
+        Time(data["data"]["photstats"][-1]["last_detected_mjd"], format="mjd").datetime
+    ) == arrow.get((90000.0 - 40_587) * 86400.0)
```
guynir42, line 777: I think you may be able to just compare MJD to MJD.

guynir42 (commented):
Overall looks good, I would wait with merging this until we make sure all sources have PhotStats.

skyportal/tests/api/test_sources.py
mcoughlin: I have a small preference since datetime is what we use in the app.

skyportal/models/obj.py
mcoughlin: I debated this for a bit, but I actually think it's pretty confusing to have these too if the rest of the app will use the other table (even if it is arguably more exact).

skyportal/tests/api/test_sources.py
guynir42: But aren't you just converting both sides of this inequality to MJD? Do you feel this helps make sure the datetime is correct?

mcoughlin: Actually I checked and it is calculating unix time on the right hand side and converting, so this way is probably best.

skyportal/models/obj.py
guynir42: ok

mcoughlin:
@guynir42 thats fine. So if we deploy Monday, can discuss Tuesday with the assumption that we will trigger in filling in such that it would be ok to merge the next week?

guynir42:
@mcoughlin Sounds about right.

mcoughlin:
@guynir42 Can we get this merged now?

guynir42:
@mcoughlin not quite yet. Still need to generate them and I don't have permission to produce phot_stats on production. Will probably be able to merge this tomorrow or next week...

guynir42:
@mcoughlin still running. It got stuck on some bad data during the night, but I put it back at it now. 
It should be finished soon so I think we can probably merge this PR today. 
Is there a serious bug if some sources lack a PhotStats?

mcoughlin:
@guynir42 No since it's just asking for sources which do have these present to make queries on.

guynir42:
In that case I think there's not harm in merging this now.

PR #3130, opened 2022-06-21
Fix "Top Sources Preferences" Popup on Dashboard

leoyang33:
Top Sources contains filters for date of the source as well as sources shown. Previously, it didn't check to see if the number of sources was added to topSourcesPrefs, which would result in the Top Sources Preferences popup on the Dashboard being empty as shown:
Screenshot 2022-06-21 at 4 14 45 PM

Now, it correctly asks the user for an input on number of sources to show:

https://user-images.githubusercontent.com/39569176/174912368-d9f4f15a-6443-462a-919e-2d9fe489bf6b.mp4


mcoughlin:
@leoyang33 Would you mind doing this on a branch instead? This will ensure a clean commit history and that you can work on a few issues at once while they get reviewed and so on.

PR #3131, opened 2022-06-22
Add handler to update batches of objects with new PhotStats

guynir42:
This adds a handler that can perform updates on objects with existing `PhotStat`s or on objects that don't have a `PhotStat` at all. 

Using `GET` will provide a count of how many objects do not have a `PhotStat`, and how many have it (with some filtering arguments like when it was full or quick updated). The handler can also filter on when the objects were created. 

Using 'POST' will let you create missing `PhotStat`s. Again the handler uses filtering on when the objects were created, and is paginated to allow the process to happen in batches. 

Using 'PATCH' will force a re-update of existing `PhotStat`s on objects, with filters for when the objects were created and when the `PhotStat`s were updated, and is also paginated.


skyportal/handlers/api/phot_stat.py
```diff
@@ -187,12 +197,464 @@ def delete(self, obj_id=None):
...
+            if created_at_start_time:
+                created_at_start_time = str(
+                    arrow.get(created_at_start_time.strip()).datetime
+                )
+                stmt = stmt.where(Obj.created_at >= created_at_start_time)
+            if created_at_end_time:
+                created_at_end_time = str(
+                    arrow.get(created_at_end_time.strip()).datetime
+                )
+                stmt = stmt.where(Obj.created_at <= created_at_end_time)
```
profjsb, line 454: Requires `created_at_start_time` and `created_at_end_time` to parse as arrow datetime type -- they might not so you should catch the errors and return appropriately

skyportal/handlers/api/phot_stat.py
```diff
@@ -187,12 +197,464 @@ def delete(self, obj_id=None):
...
+                                type: integer
+                              numPerPage:
+                                type: integer
+            400:
+              content:
+                application/json:
+                  schema: Error
+        """
+
+        page_number = self.get_query_argument('pageNumber', 1)
```
profjsb, line 582: same as above

profjsb (approved):
Couple of small changes but otherwise great.

guynir42:
Sorry, forgot to add that. 
I've also added a migration (a new column) to address the bug in the tests.

profjsb (changes requested):
Consider variable name change as per comments.

alembic/versions/c8063723c76e_phot_stat_predetection_array.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'c8063723c76e'
+down_revision = '5dc838e75a63'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        'photstats', sa.Column('predetection_mjds', sa.ARRAY(sa.Float()), nullable=True)
```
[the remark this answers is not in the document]
guynir42: Yeah but I have MJD in all of these column names already... and it would not be a good idea to use datetimes instead of floats for example...

profjsb (approved):

skyportal/handlers/api/phot_stat.py
```diff
@@ -187,12 +197,521 @@ def delete(self, obj_id=None):
...
+                        arrow.get(created_at_start_time.strip()).datetime
+                    )
+                    stmt = stmt.where(Obj.created_at >= created_at_start_time)
+                if created_at_end_time:
+                    created_at_end_time = str(
+                        arrow.get(created_at_end_time.strip()).datetime
+                    )
+                    stmt = stmt.where(Obj.created_at <= created_at_end_time)
+            except arrow.parser.ParserError:
+                return self.error(
```
profjsb, line 482: I assume leaving this DBSession context will properly shutdown the session?

profjsb:
Just a few suggested changes -^

skyportal/models/phot_stat.py
```diff
@@ -456,12 +467,19 @@ def add_photometry_point(self, phot):
...
+                # check if new point removes some predetections
                 if (
                     self.last_non_detection_mjd is not None
                     and mjd < self.last_non_detection_mjd
                 ):
-                    self.last_non_detection_mjd = None
+                    # keep only the predections that happened before this detection
+                    self.predetection_mjds = [
+                        m for m in self.predetection_mjds if m < mjd
+                    ]
```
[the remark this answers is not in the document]
guynir42: I thought about it, but I don't know if it really matters much.

skyportal/handlers/api/phot_stat.py
guynir42: That's what it is supposed to do, an anyway I don't think we are changing anything in the DB until after these return statements.

PR #3132, opened 2022-06-22
Use GW190814 in demo data

mcoughlin:
This PR uses GW190814 in the demo data in an attempt to keep the memory foot print down.

### DIFF
--- a/data/GW190814.xml
+++ b/data/GW190814.xml
@@ -1,0 +1,114 @@
+<?xml version="1.0" ?>
+<voe:VOEvent xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+xmlns:voe="http://www.ivoa.net/xml/VOEvent/v2.0"
+xsi:schemaLocation="http://www.ivoa.net/xml/VOEvent/v2.0 http://www.ivoa.net/xml/VOEvent/VOEvent-v2.0.xsd"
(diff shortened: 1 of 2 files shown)


Theodlz:
Should we use that also for the tests instead of the one we are using ? I think that one of the problems encountered by people is the app crashing when running tests multiple time when they involve ingesting events.

mcoughlin:
Well the GCNEvent has a given dateobs so ingesting an event a second time doesn't make much sense.

Theodlz:
I tried to do something similar before actually. You can maybe have a look at the branch anf compare with main to see what i needed to change to get it to work. Its here: https://github.com/Theodlz/skyportal/tree/lighter_demodata_gcnevent

mcoughlin:
Personally it does not feel like a good use of time to me.

Theodlz:
> Well the GCNEvent has a given dateobs so ingesting an event a second time doesn't make much sense.

What i meant is that for some reason, maybe because of the fact that it needs to generate thr contour and such, when running tests multiple times in a row it crashes the app, and maybe replacing an event we are using in the tests rights now by this "lighter" one could help ?

Theodlz (approved):
My comments aside (that can be the subject of another PR), this PR does its job. I'll merge once green, or if you think the failures are unrelated.

PR #3134, opened 2022-06-22
[Analysis Epic] (3) Webhook callback

profjsb:
This is the 3rd PR in the analysis epic and should be merged after #3055 and #2956.

Here we add the webhook callback for the analysis service to send results back. It includes the on-disk storage of analysis data, which is properly written, read, and deleted.

This PR also completes the full round-trip of the demo app, showing how the results data are encoded and sent back to SkyPortal.


services/sn_analysis_service/app.py
```diff
@@ -28,33 +31,51 @@
 default_analysis_parameters = {"source": "nugent-sn2p", "fix_z": False}
 
 
-def upload_analysis_results(results, data_dict):
+def upload_analysis_results(results, data_dict, request_timeout=60):
```
mcoughlin, line 34: Are there also size limits on these posts? I guess the same size as for all handlers?

services/sn_analysis_service/app.py
```diff
@@ -85,13 +106,18 @@ async def run_sn_model(dd):
...
     except Exception as e:
-        results.update(
+        rez.update(
             {
                 "status": "failure",
                 "message": f"input data is not in the expected format {e}",
             }
         )
-        upload_analysis_results(results, data_dict)
+        upload_analysis_results(rez, data_dict)
```
mcoughlin, line 115: Doesn't this need to be outside the exception call?

skyportal/handlers/api/analysis.py
```diff
@@ -808,6 +817,7 @@ async def post(self, analysis_resource_type, resource_id, analysis_service_id):
 
         # Now call the analysis service to start the analysis, using the `input` data
         # that we assembled above.
+        log(f'{get_app_base_url()}/{analysis.handled_by_url}/{analysis.token}')
```
mcoughlin, line 820: Maybe add a first few works here to help the logging?

skyportal/handlers/api/analysis.py
```diff
@@ -915,6 +946,18 @@ def get(self, analysis_resource_type, analysis_id=None):
                   schema: Error
         """
         if analysis_resource_type.lower() == 'obj':
+            include_analysis_data = self.get_query_argument(
+                "includeAnalysisData", False
```
mcoughlin, line 950: Should this query be checked outside the if statement?

skyportal/handlers/api/webhook.py
```diff
@@ -0,0 +1,129 @@
...
+                analysis.status_message = f'Analysis timed out before webhook call at {str(datetime.datetime.utcnow())}'
+                analysis.last_activity = datetime.datetime.utcnow()
+                analysis.duration = (
+                    analysis.last_activity - last_active
+                ).total_seconds()
+                session.commit()
+                session.close()
+                return self.error("Token has expired", status=400)
+
+            # lock this token and commit immediately to avoid race conditions, so that the results are not written more than once
```
mcoughlin, line 99: Nice trick!

mcoughlin (approved):
Just a few minor questions here and there!

skyportal/handlers/api/webhook.py
```diff
@@ -0,0 +1,129 @@
...
+            last_active = analysis.last_activity
+            if analysis.status not in ['pending', 'queued']:
+                return self.error(
+                    f"Analysis already updated with status='{analysis.status}'"
+                    f" and message={analysis.status_message}",
+                    status=403,
+                )
+            if (
+                analysis.invalid_after
+                and analysis.invalid_after < datetime.datetime.now()
```
stefanv, line 87: This is correct, butysurprisingly counter-intuitive. "Is the expiry date before today?" Would be easier to parse as "Is today after the expiry date?".

skyportal/handlers/api/webhook.py
stefanv: Should this comment be "lock the analysis", as above?

stefanv:
@guynir42 what's the plan, to rewrite all these PRs in the new Session framework later, or to get them in line now?

services/sn_analysis_service/app.py
stefanv: Need to see when this is None and return error.

services/sn_analysis_service/app.py
```diff
@@ -112,55 +138,81 @@ async def run_sn_model(dd):
...
-                )
-                f.seek(0)
-                plot_data = io.BytesIO(f.read())
-                f.close()
-                os.remove(f.name)
-                log('made lightcurve plot')
+
+        if result.success:
+            f = tempfile.NamedTemporaryFile(
+                suffix=".png", prefix="snplot_", dir="/tmp/", delete=False
```
stefanv, line 144: Cannot guarantee that we can write to /tmp. `./` tmp may be okay, but rather just let Python figure out an appropriate dir.

services/sn_analysis_service/app.py
```diff
@@ -112,55 +138,81 @@ async def run_sn_model(dd):
...
 
     except Exception as e:
-        results.update(
-            {"status": "failure", "message": f"problem running the model {e}"}
-        )
+        log(f"Exception while running the model: {e}")
+        log(f"{traceback.format_exc()}")
+        log(f"Data: {data}")
+        rez.update({"status": "failure", "message": f"problem running the model {e}"})
+    finally:
```
stefanv, line 207: Why prefer this over context managers?

services/sn_analysis_service/app.py
```diff
@@ -112,55 +138,81 @@ async def run_sn_model(dd):
...
-                f.close()
-                os.remove(f.name)
-                log('made lightcurve plot')
+
+        if result.success:
+            f = tempfile.NamedTemporaryFile(
+                suffix=".png", prefix="snplot_", dir="/tmp/", delete=False
+            )
+            f.close()
+            _ = sncosmo.plot_lc(
```
stefanv, line 147: Is this guaranteed to run? If this can fail, how is failure handled.

services/sn_analysis_service/app.py
```diff
@@ -112,55 +138,81 @@ async def run_sn_model(dd):
...
+            inference = az.convert_to_inference_data(
                 {x: post[:, :, i] for i, x in enumerate(result.param_names)}
             )
-            results.update({"data": {"posterior": dataset, "plot": plot_data}})
+            f = tempfile.NamedTemporaryFile(
+                suffix=".nc", prefix="inferencedata_", dir="/tmp/"
+            )
+            f.close()
+            inference.to_netcdf(f.name)
+            inference_data = base64.b64encode(open(f.name, 'rb').read())
```
stefanv, line 172: Why not BytesIO? Are these files very large?

services/sn_analysis_service/app.py
```diff
@@ -71,7 +92,7 @@ async def run_sn_model(dd):
     # the following code transforms these inputs from SkyPortal
     # to the format expected by sncosmo.
     #
-    results = {"status": "success", "message": "", "data": {}}
+    rez = {"status": "success", "message": "", "analysis": {}}
```
stefanv, line 95: Safer to start with failure, perhaps, and only update on success.

services/sn_analysis_service/app.py
stefanv: (For cleaning up files.)

skyportal/app_server.py
```diff
@@ -326,6 +327,11 @@ def log_request(self, handler):
     (r'/api/user(/[0-9]+)/roles(/.*)?', UserRoleHandler),
     (r'/api/user(/.*)?', UserHandler),
     (r'/api/weather(/.*)?', WeatherHandler),
+    # strictly require uuid4 token for this unauthenticated endpoint
+    (
+        r'/api/webhook/(obj)_analysis/([0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})?',
```
stefanv, line 332: This 89ab is part of uuid4 specifically?

skyportal/handlers/api/analysis.py
stefanv: May be worth joining with urljoin to ensure all slashes get placed correctly.

skyportal/handlers/api/analysis.py
```diff
@@ -841,6 +851,9 @@ def analysis_done_callback(
                 try:
                     session = DBSession()
                     analysis = session.query(ObjAnalysis).get(analysis_id)
+                    if not analysis:
```
stefanv, line 854: is None?

skyportal/handlers/api/analysis.py
```diff
@@ -915,6 +946,18 @@ def get(self, analysis_resource_type, analysis_id=None):
                   schema: Error
         """
         if analysis_resource_type.lower() == 'obj':
+            include_analysis_data = self.get_query_argument(
+                "includeAnalysisData", False
+            ) in ["True", "t", "true", "1", True, 1]
+            include_filename = self.get_query_argument("includeFilename", False) in [
```
stefanv, line 952: This is used all over. Can really use a utility function /cc @guynir42

stefanv:
Re: large result posts: note that nginx has limits on upload size. Double check that this fits requirements for analyses.

services/sn_analysis_service/app.py
```diff
@@ -28,33 +31,51 @@
...
+        log("Callback URL is not a POST URL. Skipping.")
+        return
+    url = data_dict["callback_url"]
+    try:
+        result = requests.post(
+            url,
+            json=results,
+            timeout=request_timeout,
+        )
+    except requests.exceptions.Timeout:
```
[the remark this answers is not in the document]
profjsb: If we timeout here then it's precisely because we cannot write back to the SkyPortal instance. So returning something doesn't make sense in this case.

services/sn_analysis_service/app.py
profjsb: I'm a half-full kind of guy but ok.

services/sn_analysis_service/app.py
profjsb: refactored so that we use the tornado callback infrastructure to send our results on success or exception/Error

services/sn_analysis_service/app.py
profjsb: failure is now handled in the callback exception handler

services/sn_analysis_service/app.py
```diff
@@ -112,55 +138,81 @@ async def run_sn_model(dd):
...
+            f = tempfile.NamedTemporaryFile(
+                suffix=".png", prefix="snplot_", dir="/tmp/", delete=False
+            )
+            f.close()
+            _ = sncosmo.plot_lc(
+                data,
+                model=fitted_model,
+                errors=result.errors,
+                model_label=source,
+                zp=23.9,
```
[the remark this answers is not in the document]
profjsb: removed.

services/sn_analysis_service/app.py
profjsb: I tried that first. The problem is that we have to send this data in a JSON packet and I was getting unicode errors with BytesIO as the JSON was trying to get serialized by `requests`. This approach of using b64encoding is what we do for thumbnails and seems to work well.

services/sn_analysis_service/app.py
profjsb: The `with TemporaryFile() as f` proved more finacky then I wanted, since I have to close the file before writing to it.

skyportal/app_server.py
profjsb: Yes, as is the 4 before [0-9a-f]{3}.

skyportal/handlers/api/analysis.py
profjsb: Good catch

skyportal/handlers/api/analysis.py
profjsb: Yes, since it can be used for other `analysis_resource_type`s in the future.

skyportal/handlers/api/webhook.py
profjsb: I will put the blame squarely on Github co-pilot for this! Fixed.

skyportal/handlers/api/analysis.py
profjsb: Fixed.

services/sn_analysis_service/app.py
profjsb: I'm not limiting the size. baselayer/nginx limits the size to 10M via `client_max_body_size`: 
```
baselayer/services/nginx/nginx.conf.template:
    client_max_body_size 10M;
```
In principle we could make this larger just for `/api/webhook` endpoints. But I think 10M is a good size.

skyportal/handlers/api/webhook.py
stefanv: That product has me seriously concerned about licensing issues. Not relevant here, of course, but what if they reproduce GPL snippets?

requirements.txt
```diff
@@ -58,9 +58,7 @@ gwemopt==0.0.76
 humanize==4.0.0
 xarray>=21.1
 simsurvey==0.7.3
-xarray>=21.1
```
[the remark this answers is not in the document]
profjsb: I dont need it for Analysis (like I thought I did) but added it in my previous PR. You can add it back in if you need it for another PR of course.

skyportal/handlers/api/webhook.py
```diff
@@ -0,0 +1,130 @@
...
+)
+
+log = make_log('app/webhook')
+
+_, cfg = load_env()
+
+Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
+
+
+class WebhookHandler(BaseHandler):
```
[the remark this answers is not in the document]
profjsb: Good idea.

services/sn_analysis_service/app.py
```diff
@@ -22,38 +29,48 @@
...
        - fix_z: whether to fix the redshift
        - photometry: the photometry to fit to the model (in csv format)
        - redshift: the known redshift of the object
 
-    TODO in the next PR: send the results back to the webhook
-
+    Other analysis services may require additional keys in the `inputs` dictionary.
     """
-    data_dict = await dd
+    data_dict = dd
```
stefanv, line 73: `dd` is rather obscure; why not just use `data_dict` as the argument name?

services/sn_analysis_service/app.py
stefanv: The b64encoding is correct, but the temporary file is not necessary if `to_netcdf` is happy dumping to `BytesIO`. You'd keep the reading from the buffer the same.

services/sn_analysis_service/app.py
```diff
@@ -112,74 +134,154 @@ async def run_sn_model(dd):
...
+            logger=log,
+            data_dict=data_dict,
+        ):
+            """
+            Callback function for when the sn analysis service is done.
+            Sends back results/errors via the callback_url.
+            """
+            try:
+                result = future.result()
+            except Exception as e:
```
stefanv, line 253: You won't get an exception here; this doesn't do an await.

services/sn_analysis_service/app.py
```diff
@@ -112,74 +134,154 @@ async def run_sn_model(dd):
...
+                logger(f"{str(future.exception())[:1024]} {e}")
+                result = {
+                    "status": "failure",
+                    "message": f"str(future.exception())[:1024] {e}",
+                }
+            finally:
+                upload_analysis_results(result, data_dict)
+
+        runner = functools.partial(run_sn_model, data_dict)
+        x = IOLoop.current().run_in_executor(None, runner)
```
stefanv, line 266: Consider either getting rid of `x` or giving it a more helpful name, like `future_result`.

stefanv:
Overall comment: may be safer to delegate long running analyses to microservice, instead of just hooking it onto IOLoop and hoping app doesn't get killed while running.

profjsb:
?? That's what we're doing here. SP fires off a request to a microservice.

services/sn_analysis_service/app.py
profjsb: It is not happy dumping to a file-like object. It must be a file.

services/sn_analysis_service/app.py
profjsb: Incorrect. This is inside a callback and future.result() will error if the coroutine didn't return something, but errored instead. This is the way to catch errors from coroutines and log them appropriately.

services/sn_analysis_service/app.py
profjsb: done.

.github/workflows/test_models.yaml
```diff
@@ -139,7 +139,7 @@ jobs:
           fi
           sudo tar -xzf ${CACHED_DOWNLOAD_DIR}/geckodriver-v${GECKO_VER}-linux64.tar.gz -C /usr/local/bin
           geckodriver --version
-          pip install --upgrade selenium
+          pip install --upgrade selenium==4.2.0
```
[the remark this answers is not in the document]
profjsb: We need this because the new selenium on pypi (4.3.0) just broke other imports (see slack discussion; michael has a PR in to selenium request project). Even if we fix this on baselayer, the `--upgrade` will try to push selenium to the broken version.

requirements.txt
```diff
@@ -58,10 +58,8 @@ gwemopt==0.0.76
...
 noaodatalab==2.20.0
 black==22.3.0
 click==8.1.3
-quart==0.17.0
```
profjsb, line 65: I refactored the microservice to use tornado instead of quart.

services/sn_analysis_service/app.py
profjsb: fixed.

services/sn_analysis_service/app.py
profjsb: cf.  https://github.com/arviz-devs/arviz/blob/44b908f20372b28b82fffb537c5f69628a99b02e/arviz/data/io_netcdf.py#L46

services/sn_analysis_service/app.py
profjsb: See https://quantlane.com/blog/ensure-asyncio-task-exceptions-get-logged/  ("HOW TO ENSURE TASK EXCEPTIONS GET LOGGED IMMEDIATELY")

stefanv:
> ?? @stefanv: That's what we're doing here. SP fires off a request to a microservice.

I had a terrible dream about this comment of mine last night. Requesting a microservice to handle a microservice to handle a microservice...

Apologies, brain short circuit.

.github/workflows/test_models.yaml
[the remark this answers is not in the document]
stefanv: We should remove the upgrade flag, at least.

stefanv:
What I was after is handling of failures. If the product is valuable, a persistent queue is warranted. If failure is fine, simply recalculating would be good. So,  wanted to hear your reasoning around that.

profjsb:
> What I was after is handling of failures. If the product is valuable, a persistent queue is warranted. If failure is fine, simply recalculating would be good. So, wanted to hear your reasoning around that.

I understand now. 

A failure is "fine" in that the analysis of a particular object can be restarted by just generating another analysis. Some analysis failures might be because the model is simply insufficient to describe the data, there might not be enough data available at the time of the analysis, or the the analysis service might be expecting to get some data from SkyPortal that it does not.  In all of these failure modes, it doesn't make sense to have the microservice restart the analysis on the some input data. (of course a more sophisticated analysis microservice can certainly decide to implement a queue)

When an analysis fails, it's actual useful for the end-user to send back the errors (via a status message) using the webhook.

Hope this helps. Happy to say more if need be.

profjsb:
> > ?? @stefanv: That's what we're doing here. SP fires off a request to a microservice.
> 
> I had a very confusing dream about this comment of mine last night. Requesting a microservice to handle a microservice to handle a microservice...
> 
> Apologies, brain short circuit.

It's quite alright. Can't imagine why your brain might not be at it's usual 110% ...

.github/workflows/test_models.yaml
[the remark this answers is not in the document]
profjsb: Done.

stefanv:
> Hope this helps. Happy to say more if need be.

That makes perfect sense.  The only annoying case is where the analysis is successful, but SkyPortal cannot be reached to upload the result.  But this is a refinement we can add in a future version.

stefanv (approved):

PR #3138, opened 2022-06-23
Observation table query efficiency

mcoughlin:
This PR improves query efficiency on the observation table by calling distinct() as appropriate.


skyportal/handlers/api/observation.py
```diff
@@ -443,10 +447,12 @@ def get_observations(
             )
             query_area = sa.select(area).filter(
                 LocalizationTile.localization_id == localization.id,
+                LocalizationTile.probdensity >= min_probdensity,
```
guynir42, line 450: This seems like a good way to make things go faster. I assume you know what you are doing in terms of the logic of the query...

guynir42 (approved):

PR #3140, opened 2022-06-24
Improve area and probability calculation with subquery

mcoughlin:
This PR improves the area and probability calculation for observation plans with a subquery.

### DIFF
--- a/skyportal/models/observation_plan.py
+++ b/skyportal/models/observation_plan.py
@@ -26,7 +26,7 @@ from baselayer.app.models import (
 )
 
 from .group import Group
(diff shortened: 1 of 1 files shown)


Theodlz (approved):
Looks good. Didn't expect few lines of SQL queries to force me into a 1h reading of healpix stuff hahaha. As discussed, the `.over(order_by=LocalizationTile.probdensity.desc())` shouldn't need the `order_by`. What I mean by that is as we are not keeping just one tile (the last or first one for example), we don't need to sort them, but the result seems to be wrong if we don't do it (maybe the `over` absolutely needs something in it ?). So let's keep it for now.

PR #3141, opened 2022-06-24
Restrict galaxies going into schedule to those within localization

mcoughlin:
This PR restricts the galaxies that go to gwemopt to only those within the localization region.


Theodlz (approved):
Same comment as in #3140 , otherwise this worked well for me. But we need more data in the galaxy catalog to try it properly. Could be done in a separate PR.

mcoughlin:
I opened an issue for adding a new catalog to the demo data. For now, we can merge this.

Issue #3122, opened 2022-06-21
Differentiation between detections and non detections in the object page photometry


mcoughlin:
The markers for detections and non detections are too close visually (on the OBJECT page) , could you choose a different scheme? 
The ones used on the SCANNING page is good and clear (downward arrows for NDs ). 
I know that you're trying to differentiate between telescopes on the object page, however, conservatively, I would keep the downpointing triangles for non detections. 
<img width="438" alt="Screen Shot 2022-05-11 at 13 23 32" src="https://user-images.githubusercontent.com/45436034/167828582-64097bac-7c77-4414-bff4-e0915fb98c34.png">

Copied from https://github.com/fritz-marshal/fritz/issues/414

mcoughlin:
Thanks for resolving this @thomasculino

PR #3143, opened 2022-06-24
Differentiation between detections and non detections in photometry plots

thomasculino:
Currently, we have photometry plots with only circles as markers. This PR aims to differentiate between detections and non detections as requested in #3122.

Currently we have: 
image

After the changes:
image

and when hovering the tooltip you will see additional details: 
image


mcoughlin (commented):
@thomasculino perfect. Could you also add the same thing to SourceMobile.jsx? I know it's funny to keep track of both of these.

guynir42 (approved):

guynir42:
Thanks @thomasculino I was just talking to some users who said this was really needed!

thomasculino:
> @thomasculino perfect. Could you also add the same thing to SourceMobile.jsx? I know it's funny to keep track of both of these.

I was working on it but the tooltip displays information when hovered so it does not work on mobile. Should I change the position of the tooltip and make it display additional information when clicked instead of hovered for mobile? @mcoughlin

mcoughlin:
@thomasculino ah if that's easy sure, but great point on the hover. Don't let it stop the PR from going in if it's more than a few minutes for you.

thomasculino:
@mcoughlin Done !

mcoughlin:
@thomasculino thanks for doing that!

mcoughlin:
@thomasculino if you have a few minutes, could you add the same fix to https://github.com/fritz-marshal/fritz/blob/main/extensions/skyportal/static/js/components/SourceDesktop.jsx? We have a copy because of a few extra buttons. I would like to think about how to re-emerge these for extensions of this kind, as it is annoying to keep them each independently up to date. @Theodlz in case you have ideas there. The diff isn't so bad; but there are more buttons.

Theodlz:
Basically the problem to solve here is being able programmatically to do something similar to a git merge right ? So if there are no conflict, its easier for the dev to update a file that has a copy in the extensions, or even completely automatic (but thats not necessarily a good idea, there will always be conflicts at some point and that could be problematic for deployement)

Theodlz:
I have an idea on how to do something that could help. I'll try that

Theodlz:
Here's a quick example on how we could handle that. Git has a command called `git merge-file`. The arguments are: `current-version` `common-ancestor` `other-version`. So here, lets say you created a modified version of the source page on fritz, with skyportal version A. Lets call that SourcePageFritz, and the original one SourcePageOld. Now there is a new version of SkyPortal, B, in which you'll find SourcePageNew (those are examples, in reality all those files have the same names of course, just a different directory). So we have the `current-version  = SourcePageNewSky`, `common-ancestor = SourcePageOld` and `other-version = SourcePageFritz`.

mcoughlin:
@Theodlz really beautiful. Can we plan to discuss on Tuesday maybe?

profjsb:
Can we promote this merge idea to a SkyPortal discussion topic? I'd like to decouple this important idea from the original PR. Also want to pull in @stefanv ?

mcoughlin:
@profjsb of course! @Theodlz might move the discussion to an issue. Definitely will discuss on Tuesday.

profjsb:
Rather than an issue perhaps a back and forth over a Discussion is better? A la https://github.com/skyportal/skyportal/discussions/2638

PR #3147, opened 2022-06-25
Galaxies Catalog - Added galaxies within event 190814

Theodlz:
Right now, no galaxies are contained in any of the 2 events we have, so we cannot test certain features, such as creating obs plan targeting the galaxies contained in the localization of an event.  This PR extends the CLU_mini catalog with a certain number of galaxies contained in one of the 2 gcn event in the demo data (190814)


mcoughlin:
@Theodlz Can you fix https://github.com/skyportal/skyportal/blob/main/skyportal/tests/flaky/test_galaxy_flaky.py to have the right numbers? Would maybe be good to load 0814 there instead of 0425 given the relevance.

Theodlz:
@mcoughlin on it right now. By the way I think I got why this test is flaky. When I run it once and it passes, I can run it again. But if it fails once, then it will always fail. That happens because there is the following constraint on the galaxy model : you can't have 2 entries with the same galaxy name. So if the test fails once for any reason, as the last bit of code that deletes the catalog doesn't run, then the next time it always fails as a galaxy with that name already exists.

mcoughlin:
@Theodlz what if we start the test by deleting the catalog then? The handler has a catalog option.

Theodlz:
The name of the catalog is a uuid generated at random, so we dont have the name of the catalog that vould have been added previously. Otherwise i was just thinking to replace any assert by if not statements. If the if not is met we delete the catalog anf simply assert False.

mcoughlin:
@Theodlz ah nice.

Theodlz:
Pushed. I'm not sure if its "ok" to do it like this, but that allows me to rerun after it fails and getting it to pass.

mcoughlin:
@Theodlz i mean, it doesn't seem unreasonable to me, I am not sure how else to do it?

Theodlz:
What made me doubt what I did is the fact that I used a seperate function with asserts, and if statements instead of asserts. But if you that's fine i'll untag stefan and we can get this in once green.

mcoughlin:
@Theodlz lets discuss on Tuesday in case others have a better idea, but my guess is something like this will be fine.

Theodlz:
I mean, we can always use a fixed them instead of a uuid, so we know the name of the catalog and we can delete it first thing when we run the test just like you said. Maybe using a fixed name is an easier and cleaner fix ?

mcoughlin:
@Theodlz i am ok with that!

Theodlz:
Hi again @mcoughlin. I just realized that deleting it at the start doesn't solve everything. The thing here is that even though this is a flaky test, it doesn't have a decorator that reruns it `n` times if it fails. So if it fails, that it is and it won't run again anyway. Deleting the catalog (i.e deleting all the galaxies we posted) at the end or the start is not "that" important. However, deleting things after running a test allows tests to be independent, as data doesn't "accumulate" in the DB when some test fails. That way a test's failure is less likely to "cascade". But, one can argue that if it can help to have more tests passing, maybe it gets further away from a real-life scenario (we keep a very small set of data, whereas data accumulating would get the DB "fuller"). One thing to verify if we decide not to delete the posted catalog upon failure is that other tests do not post galaxies with the same names, as that will probably result in those tests not passing. I will have a look at other tests to see if that happens.

Theodlz:
It does not seem to be the case. No tests that will run after that one uses galaxies from this  CLU mini catalog. So keeping it in the DB when the test fails shouldn't be a problem. Deleting it how I did it might be overkill hahaha. It just depends if we want to add a decorator to rerun upon failure, but I don't think we want that so maybe we could just go back to not deleting upon failure, which makes the test simpler.

Theodlz:
In the end, as some tests (api) do use a rerun, I still started every galaxy tests by deleting a catalog. This is not asserted and it is not necessarily to do so, as it is meant to succeed only if there is one. All tests are green. Ready for review.

mcoughlin (approved):
@Theodlz Clearly a big improvement in the quality of the demo data here, and thanks for trying to make the tests more robust in the process. Feel free to merge once the tests pass!

Issue #2970, opened 2022-05-12
Telescope without `.lon` causes uncaught exceptions


stefanv:
```
68:                lng=(self.lon + 180) % 360 - 180, lat=self.lat, delta_degree=5
```

PR #3148, opened 2022-06-25
Telescope model - Issue with missing lon/lat

Theodlz:
This PR addresses https://github.com/skyportal/skyportal/issues/2970. I made some changes to the model and added comments to justify them. I'm tagging @stefanv (that opened the issue) as well as @mcoughlin that assigned me. 
The comments I left in the model have to be removed of course, they are here only for the reviewers to understand the changes I made. But maybe some of them could stay? (for example, I added a comment not to justify any of my changes but to clarify some bits of the code in the `def observer()`)


mcoughlin (approved):
@Theodlz Nice. Hopefully this finishes it off!

Theodlz:
I noticed in the flaky tests that there is also a galaxy tests in the gcn evet flaky tests... Let me fix it before merging !

Theodlz:
Before this goes in, maybe #3101 could go in first ? Its been ready for a some days now.

skyportal/models/telescope.py
```diff
@@ -67,75 +67,83 @@ def observer(self):
...
-                    elevation=elevation * u.m,
-                    timezone=local_tz,
-                )
-
-            except Exception as e:
-                log(
-                    f'Telescope {self.id} ("{self.name}") cannot calculate an observer: {e}'
-                )
+            if self.lat is None or self.lat == "" or np.isnan(self.lat):
+                return None
```
guynir42, line 75: Maybe we should not return anything, but instead set `self._observer = None` and then only at the very end return `self._observer`. That way this logic is only called once, and then the result is cached.

skyportal/models/telescope.py
```diff
@@ -67,75 +67,83 @@ def observer(self):
...
+            self._observer = astroplan.Observer(
+                longitude=self.lon * u.deg,
+                latitude=self.lat * u.deg,
+                elevation=elevation * u.m,
+                timezone=local_tz,
+            )
+
+        except Exception as e:
+            log(
+                f'Telescope {self.id} ("{self.name}") cannot calculate an observer: {e}'
```
guynir42, line 100: If we cache the results of `self._observer` this message will only printout once. I think that's better than having it pop up all the time.

skyportal/models/telescope.py
```diff
@@ -67,75 +67,83 @@ def observer(self):
...
+            )
+            return None
+
         return self._observer
 
+    # in all the methods below that use this self.observer, I checked first if it is None, rather than having an if statement on the "time" value first.
+    # anyway, if the observer is None, we won't need the "time" value, so it seems better to check the observer first.
+    # It will avoid unnecessary calculations if the observer is None.
```
guynir42, line 108: Just so you know, you can also add comments on github (like reviewer comments but on your own code). 
That may be easier that adding code comments and then cleaning them up, but it is totally up to you either way.

guynir42 (approved):
Looks good. I would cache the result of `self.observer` in `self._observer` but otherwise the code looks fine. I am really surprised we are still dealing with this - I hope this takes care of it finally.

skyportal/models/telescope.py
Theodlz: Ok well I did just that, i'll let you have a look @guynir42

skyportal/models/telescope.py
Theodlz: I think all those if statement can be combined into one, where if triggered, we set the `self._observer` to None and return it, so that return will be triggered just once. The reason for keeping the return is that when running for the first time (when there are no observer yet) it will keep it from running the next try expect.

Theodlz:
I hope this will help. As mentionned before, I never managed to reproduced locally the issues encountered in production, so it's hard to try it and give any guarantee that this will fix it. But it should make it more robust for sure.

mcoughlin (approved):
Thanks for all your effort here @Theodlz.

guynir42 (approved):

PR #3151, opened 2022-06-27
Use newer version of postgres in default compose

mcoughlin:
This PR bumps default postgres in docker-compose to 14.4, required for all of the healpix-alchemy infrastructure.


profjsb (approved):
GCP uses 14.2 (and we'd like to stay close to what we run in production) but it looks like there are only bug fixes from 14.2 to 14.4 so I think it's ok to pin 14.4 here.

PR #3155, opened 2022-06-29
Fix ValueError has no attribute normalized_messages error

welterde:
Jesper got that error when trying to create a followup request.
The issue is that ValueError gets thrown in post_assignment, but the exception handler expects a ValidationError only.


mcoughlin:
@welterde Thanks for the PR! Can you please run
pre-commit install
pre-commit run --files skyportal/handlers/api/followup_request.py
git commit skyportal/handlers/api/followup_request.py
git push

to fix the linting errors?

welterde:
Done (after second attempt.. helps to actually add the changes).

mcoughlin:
Thanks @welterde !

PR #3156, opened 2022-06-30
WIP: Gcn summary Handler

Theodlz:
This PR is a work in progress at a very early stage, please do not review!


skyportal/handlers/api/galaxy.py
```diff
@@ -21,6 +21,126 @@
 MAX_GALAXIES = 10000
 
 
+def get_galaxies(
```
mcoughlin, line 24: Nice refactor!

skyportal/handlers/api/gcn.py
```diff
@@ -660,3 +671,491 @@ def delete(self, dateobs, localization_name):
...
+        if event is None:
+            return self.error("Event not found", status=404)
+        if not no_text:
+            group = Group.query.filter_by(id=group_id).first()
+            if group is None:
+                return self.error(f"Group not found with ID {group_id}")
+
+            header_text = []
+
+            trigger_time = astropy.time.Time(event.dateobs, format='datetime')
```
mcoughlin, line 830: Would this be more readable as one f string block?

skyportal/handlers/api/gcn.py
```diff
@@ -660,3 +671,491 @@ def delete(self, dateobs, localization_name):
...
+                    user = User.query.get(user_id)
+                    if user is None:
+                        return self.error(f"User ID {user_id} not found")
+                    users.append(user)
+
+                users_txt = []
+                for user in users:
+                    if user.first_name is not None and user.last_name is not None:
+                        users_txt.append(
+                            f"""{user.first_name[0].upper()}. {user.last_name} (Affiliation)"""  # hardcoded affiliation as it is not implemented yet
```
mcoughlin, line 856: Good.

skyportal/handlers/api/gcn.py
```diff
@@ -660,3 +671,491 @@ def delete(self, dateobs, localization_name):
...
+                )
+                header_text.append(f"""\n{users_txt}\n""")
+
+            header_text.append(f"""\non behalf of the {group.name}, report:\n""")
+            contents.extend(header_text)
+
+        if show_sources:
+            sources_text = []
+            source_page_number = 1
+            source_num_per_page = 100
```
mcoughlin, line 873: Should we grab this number as a default from sources.py?

skyportal/handlers/api/gcn.py
```diff
@@ -660,3 +671,491 @@ def delete(self, dateobs, localization_name):
...
+                sources.extend(sources_data['sources'])
+                source_page_number += 1
+
+                if len(sources_data['sources']) < source_num_per_page:
+                    break
+            if len(sources) > 0:
+                sources_text.append(
+                    f"\nFound {len(sources)} {'sources' if len(sources) > 1 else 'source'} in the event's localization, given the specified date range:\n"
+                ) if not no_text else None
+                ids, aliases, ras, decs, ra_errs, dec_errs, redshifts = (
```
mcoughlin, line 899: I think we can do without the ra/dec uncertainties. redshift uncertainty if it is there though is nice.

skyportal/handlers/api/gcn.py
```diff
@@ -660,3 +671,491 @@ def delete(self, dateobs, localization_name):
...
+                                showindex=False,
+                            )
+                            + "\n"
+                        )
+            contents.extend(sources_text)
+
+        if show_galaxies:
+            galaxies_text = []
+            galaxies_page_number = 1
+            galaxies_num_per_page = 1000
```
mcoughlin, line 1004: Same here for the galaxies loop?

mcoughlin (approved):
A Tour de Force.

I think maybe grabbing the default numbers of objects to loop through instead of hard coding might be best practice. But this is going to be incredibly useful.

mcoughlin:
A few notes from the MMA call (that can be future PRs):
(i) Is it possible to choose the sources that make it into the table using some kind of checkbox list? It's possible that some will already be ruled out by other means hard to specify within the app (an old source, someone else's GCN, etc).
(ii) Can we have a place in the config to add standard acknowledgments at the bottom of the file by default? This will just be a paragraph like: blah blah is funded by ...

Theodlz:
> A few notes from the MMA call (that can be future PRs): (i) Is it possible to choose the sources that make it into the table using some kind of checkbox list? It's possible that some will already be ruled out by other means hard to specify within the app (an old source, someone else's GCN, etc). (ii) Can we have a place in the config to add standard acknowledgments at the bottom of the file by default? This will just be a paragraph like: blah blah is funded by ...

Hi @mcoughlin, and thanks for the review!
For (i), we can add more conditions of course. But we just need to decide exactly what those conditions are, and how we handle them in the frontend. 
For (ii), that can be done without any problem. However should it really be a config thing? That would me that whoever uses that feature, whatever his group, the acknowledgments text is the same?

mcoughlin:
@Theodlz Ah good point, so maybe a group level option in that case?

Theodlz:
Note: the commits are not from my account but someone else's account. I got a new desk and computer at work and apparently vscode kept the account of the person before me...

mcoughlin:
@Theodlz its like we have another contributor ;)

Theodlz:
Hi @mcoughlin, as we discussed last week, (i) would be the subject of another PR, but (we did not discuss about this one yet) (ii) could get in this one. 

If we make it a group level thing, how should we store it, and allow user input. Should we simply add it in the config for now (which I don't think is such a good option), or should we store it in the DB. If we store it in the DB, what seems preferable according to you: 

- Adding a column to the group table
- Adding a table with 2 columns: group_id, acknowledgment_text

Option 2 seems more logical to me, as I don't think that these acknowledgments texts are something so important to the group itself, and not everyone would use them. 
With option 2, we can also easily have multiple texts for one group if needed. But of course, that is also doable with option 1.
On the front-end, we need to add an element to the group page where you can add acknowledgments. I am thinking maybe something similar to what you can find on https://github.com/skyportal/skyportal/pull/3092 to manage affiliations.
Then, on the gcn summary, you can: 
- Option 1: After selecting the group, you can choose one (or more ?) acknowledgments text(s) in a dropdown, ultimately adding them to the summary.
- Option 2: In the DB, we just allow to add one text per group, and it is automatically added to the summary after selecting the group, with no additional user input.

mcoughlin:
@Theodlz I agree that a small table here probably makes the most sense. 

Option 2: In the DB, we just allow to add one text per group, and it is automatically added to the summary after selecting the group, with no additional user input. -> This is also a bit nicer as if we are using multiple telescopes for the same GCN (it could happen) we may want to choose a few acknowledgments to be able to add.

Theodlz:
As discussed ln Slack, considering how big this PR already is, the acknowledgements should be added in a seperate handler.

Theodlz:
Hi @mcoughlin. Besides the photometry stuff that needs to change when https://github.com/skyportal/skyportal/pull/3220 goes in, this is ready for a review. Somehow, all tests are passing, which indicates that later (in a dedicated PR ?) we might be able to put some MMA stuff back in the `tests/api` and `tests/frontend` rather than `tests/flaky`.

skyportal/handlers/api/galaxy.py
```diff
@@ -21,6 +21,126 @@
...
+            query_result.append(
+                {
+                    'catalog_name': catalog_name,
+                    'catalog_count': int(total_matches),
+                }
+            )
+
+        return query_result
+
+    query = Galaxy.query_records_accessible_by(user, mode="read")
```
mcoughlin, line 52: Can we use Guy's session syntax here and below?

skyportal/handlers/api/gcn.py
```diff
@@ -657,3 +668,503 @@ def delete(self, dateobs, localization_name):
...
+              name: showObservations
+              required: true
+              schema:
+                type: bool
+              description: Filter by end date
+            - in: query
+              name: localizationDateobs
+              schema:
+                type: string
+            - in: query
```
mcoughlin, line 734: Add a description here and for any others missing above?

mcoughlin (approved):
@Theodlz Beautiful work. I suggest making sure to use the session syntax on the query and finish off the handler docs and we can get that in.

skyportal/handlers/api/galaxy.py
Theodlz: Of course. However, I'm not sure which syntax to use exactly. Things are done differently in so many handlers that I can't tell which one is the right one to use.

Theodlz:
Thanks for the review @mcoughlin. I replied to one of your suggestions, and I also had a question to ask. Do you think we should delete the gcn summaries in the obsplan and obs handler? I'm not sure they are still useful now. But maybe it's better to have extra stuff rather than maybe missing some.

mcoughlin:
@Theodlz yes we can delete those!

If you grep for stmt, Guy has his nice examples like that!

PR #3165, opened 2022-07-01
Make 24 hour or AM/PM optional

mcoughlin:
This PR makes a 24 hour or AM/PM type clock optional

### DIFF
--- a/static/js/components/UIPreferences.jsx
+++ b/static/js/components/UIPreferences.jsx
@@ -13,6 +13,7 @@ const UIPreferences = () => {
   const currentTheme = preferences?.theme;
   const invertThumbnails = preferences?.invertThumbnails || false;
   const compactComments = preferences?.compactComments || false;
+  const twelveOrTwentyFour = preferences?.twelveOrTwentyFour || false;
   const dispatch = useDispatch();
 
   const themeToggled = (event) => {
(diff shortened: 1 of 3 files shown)


stefanv:
Could you explain the goal of this PR is some more detail? What's the motivation for not always using 24h?

mcoughlin:
@stefanv Right now DateTimePicker defaults to AM/PM, and there is a request for 24 Hours. maybe because I am American I find AM/PM a bit more natural.

static/js/components/UIPreferences.jsx
```diff
@@ -13,6 +13,7 @@ const UIPreferences = () => {
   const currentTheme = preferences?.theme;
   const invertThumbnails = preferences?.invertThumbnails || false;
   const compactComments = preferences?.compactComments || false;
+  const twelveOrTwentyFour = preferences?.twelveOrTwentyFour || false;
```
guynir42, line 16: What are the values going to be? `true` or `false`? Which corresponds to 24h and which to am/pm?

guynir42 (approved):
I would make it a bit more obvious in the code (with a comment or a better variable name) what values represent 24h vs. am/pm. Other than that it seems fine.

static/js/components/UIPreferences.jsx
mcoughlin: yes making this more obvious in the variable name now :)

guynir42 (approved):

Issue #3186, opened 2022-07-01
Photometry distinct origin query very slow


mcoughlin:
I think we need to add an index or something on origin?

Query | Database | Load by total time | Avg execution time (ms) | Times called | Avg rows returned
-- | -- | -- | -- | -- | --
SELECT DISTINCT photometry.origin AS photometry_origin FROM photometry | fritz-1 |   | 326,421.15 | 3 | 8

thomasculino:
I thought that the origin column was already indexed. That is why we decided to use this query in this beginning. However, if the origin column is indexed and is still too slow, we can set a list of commonly used origins so that the execution time to display the list of origins on the profile page will not be long. And if the origin is not in the list, the user can add manually the name of the origin he wants to filter next to the list of hard coded ones.

mcoughlin:
Ah good point @thomasculino, you are right about the index. When is the query triggered? Is it just the user profile page?

thomasculino:
It should only be triggered there yes, just in the OriginSelect.jsx which is used only once in the profile page.

thomasculino:
I can try making it async

PR #3187, opened 2022-07-01
Making the origin select async

thomasculino:
Making the origin select async to try reducing the execution time (related to https://github.com/skyportal/skyportal/issues/3186)

### DIFF
--- a/static/js/components/OriginSelect.jsx
+++ b/static/js/components/OriginSelect.jsx
@@ -7,8 +7,8 @@ import * as photometryActions from "../ducks/photometry";
 
 const OriginSelect = ({ onOriginSelectChange, initValue, parent }) => {
   const dispatch = useDispatch();
-  const fetchOrigins = () => {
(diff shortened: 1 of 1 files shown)


mcoughlin:
@thomasculino I don't think that syntax quite works. I think it has to be like the one in GcnEventPage and others.

thomasculino:
> @thomasculino I don't think that syntax quite works. I think it has to be like the one in GcnEventPage and others.

I am not sure about what syntax I should use then. Should I declare fetchOrigins inside the useEffect?

mcoughlin:
Something like

  useEffect(() => {
    const fetchOrigins = async () => {
      await dispatch(photometryActions.fetchAllOrigins());
    };
    fetchOrigins();
  }, [dispatch]);

It's kind of idiosyncratic.

mcoughlin (approved):
Thanks @thomasculino!

PR #3190, opened 2022-07-02
Pass width and height of fields to simsurvey

mcoughlin:
This PR passes the width and height of fields to simsurvey. It also uses the with DBSession() as session syntax.

### DIFF
--- a/skyportal/handlers/api/observation.py
+++ b/skyportal/handlers/api/observation.py
@@ -820,7 +820,9 @@ class ObservationHandler(BaseHandler):
 
         with self.Session() as session:
             observation = session.scalars(
-                ExecutedObservation.select(self.current_user).where(id=observation_id)
+                ExecutedObservation.select(self.current_user).where(
+                    ExecutedObservation.id == observation_id
+                )
             ).first()
             if observation is None:
                 return self.error("ExecutedObservation not found", status=404)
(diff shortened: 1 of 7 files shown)


skyportal/handlers/api/observation.py
```diff
@@ -1721,153 +1723,179 @@ async def get(self, instrument_id):
...
+                    Localization.query_records_accessible_by(
+                        self.current_user,
+                    )
+                    .filter(Localization.dateobs == localization_dateobs)
+                    .filter(Localization.localization_name == localization_name)
+                    .first()
+                )
+
+            event = (
+                GcnEvent.query_records_accessible_by(
```
guynir42, line 1791: Use select?

skyportal/handlers/api/observation_plan.py
```diff
@@ -1720,120 +1728,147 @@ async def get(self, observation_plan_request_id):
...
 
-        localization = (
-            Localization.query_records_accessible_by(
+            observation_plan_request = ObservationPlanRequest.get_if_accessible_by(
+                observation_plan_request_id,
                 self.current_user,
+                mode="read",
+                raise_if_none=True,
+                options=options,
             )
```
guynir42, line 1758: There's a new `get` method in baselayer that uses the new select tools. And also a `select` method that does permissioning. Since you are already in a session might as well use that.

guynir42 (approved):
I would move all query and get if accessible by to the new methods. Other than that looks good.

skyportal/handlers/api/observation.py
mcoughlin: Indeed!

skyportal/handlers/api/observation_plan.py
mcoughlin: Great.

PR #3191, opened 2022-07-02
Set the group defaults for the allocation

mcoughlin:
This PR sets the group defaults for the default allocation.

### DIFF
--- a/skyportal/handlers/api/allocation.py
+++ b/skyportal/handlers/api/allocation.py
@@ -58,50 +58,61 @@ class AllocationHandler(BaseHandler):
         """
 
         # get owned allocations
-        allocations = Allocation.query_records_accessible_by(self.current_user)
 
-        if allocation_id is not None:
-            try:
-                allocation_id = int(allocation_id)
-            except ValueError:
-                return self.error("Allocation ID must be an integer.")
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/allocation.py
```diff
@@ -58,50 +58,61 @@ def get(self, allocation_id=None):
...
+                else:
+                    return self.error(
+                        f"apitype can only be api_classname or api_classname_obsplan, not {apitype}"
+                    )
 
-        allocations = allocations.all()
-        self.verify_and_commit()
-        return self.success(data=allocations)
+            allocations = session.scalars(allocations).all()
+            self.verify_and_commit()
```
guynir42, line 114:
```suggestion
            session.commit()
```

guynir42: Are you sure we even need a `commit` here?

guynir42 (approved):
Some minor comments below:

skyportal/handlers/api/allocation.py
mcoughlin: Good point!

mcoughlin:
Thanks @guynir42

PR #3192, opened 2022-07-03
Image analysis

mcoughlin:
This PR adds basic image interactions. This feature allows users to upload FITS images from observations of a source, and then use STDPipe to extract the photometry from the image. As this feature requires additional system and python dependencies, we choose to make it optional. It is deactivated by default and can be activated in the config.


thomasculino:
The image analysis is working locally after installing some system dependencies : scamp, sextractor, psfex. Once they are installed, uploading a compressed fits file will result in the posting of a new photometry in the database. Should we add these dependencies in the docs ?  @mcoughlin @Theodlz

mcoughlin:
@thomasculino i think we want to discuss it as an optional dependency of sorts. Just because we know it builds well on Linux but maybe less so on other OS'a

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,409 @@
...
+
+            image_data = data.get("image_data")
+            if image_data is None:
+                return self.error(message='Missing image data')
+
+            filt = data.get("filter")
+            obstime = data.get("obstime")
+            obstime = Time(arrow.get(obstime.strip()).datetime)
+            with tempfile.NamedTemporaryFile(
+                suffix=".fits.fz", mode="wb", delete=False
```
profjsb, line 382: Will need to delete this file later so as not to pollute the disk.

Dockerfile
```diff
@@ -2,6 +2,10 @@ FROM ubuntu:20.04
 
 ARG DEBIAN_FRONTEND=noninteractive
 
+# lines to comment out if you want to use the image analysis feature
```
mcoughlin, line 5: To *uncomment*

Dockerfile
```diff
@@ -20,6 +24,15 @@ RUN python3 -m venv /skyportal_env && \
     bash -c "source /skyportal_env/bin/activate && \
     pip install --upgrade pip==22.2.2 wheel numpy"
 
+
+# lines to comment out if you want to use the image analysis feature
```
mcoughlin, line 28: To *uncomment*

doc/image_analysis.md
```diff
@@ -0,0 +1,73 @@
+# Image Analysis
+
+This feature allows user to upload FITS images from observations a source, and then use STDPipe to **extract the photometry from the image**.
```
mcoughlin, line 3: observations of

doc/image_analysis.md
```diff
@@ -0,0 +1,73 @@
...
+```
+source skyportal_env/bin/activate
+cd stdpipe
+pip install -e .
+```
+
+## Configuration file
+
+By default, the app allows a maximum request size of 10MB (from the client). This is not enough for someone to upload most .fits files (compressed or not).
+To increase the maximum request size, change the `max_body_size: 10` parameter in the `config.yaml` file to a larger value, like a `100` (in MB).
```
mcoughlin, line 36: I would say like `100` (no a)

doc/image_analysis.md
```diff
@@ -0,0 +1,73 @@
...
+
+Also, as mentionned earlier, you need to set `image_analysis` to `True` in the `config.yaml` file. By default, it is set to `False`.
```
mcoughlin, line 38: mentioned.

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,422 @@
...
+            obstime = data.get("obstime")
+            obstime = Time(arrow.get(obstime.strip()).datetime)
+            image_data = image_data.split('base64,')
+            image_name = image_data[0].split('name=')[1].split(';')[0]
+            # if image name contains (.fz) then it is a compressed file
+            if image_name.endswith('.fz'):
+                suffix = '.fits.fz'
+            else:
+                suffix = '.fits'
+            with tempfile.NamedTemporaryFile(
```
mcoughlin, line 392: Should this be delete=True?

mcoughlin (commented):
@Theodlz Really nice. A few comments mostly on documentation.

skyportal/handlers/api/image_analysis.py
Theodlz: So when I've set it to true, I ended up having issues down the line. Instead, I have another piece of code down the line that removes the file. But now that you mention it, I need to make sure that the file is deleted no matter what, which means adding that piece of code in the `expect` of the handler method. Let me take care of that before we merge anything.

Theodlz: Either this or we move the whole creating and opening a file part inside of the reduce_image function (that runs in executor), and out of the handler.

Theodlz: Here's what I'm scared of:
What happens when the app crashes/dies? Currently, even if I add some code to remove the file in the except statement, well naturally if the app is killed and restarted because it is not responsive, none of that code is reached. Therefore the file is not deleted...

Theodlz: Maybe it's best that it put delete to True, and in the run in executor part, I recreate the file also with delete = True. The issue with putting delete=True only in the handler is, as it is asynchronous, when the handler returns it deletes the file and then the method that is running in executor can't open it!

mcoughlin: We can try it and see if we run into issues.

Theodlz: I'll push some code super soon. Basically we were reopening later in some astropy methods to get the header. But we already have it! So I'm changing the reduce_image method to get rid of that.

Theodlz: Turns out that sextractor and psfex also create temp directories and files, which are not deleted if the app dies, or even worse not delete at all if there is an exception...
So I'm adding some code to take care of that. I delete them when an exception occurs, and when the app restarts dies/restarts, it deletes all temp files related to sextractor or pfsex, ensuring that we don't accumulate useless data over time and fill up the disk space.

Theodlz: Should be ok now

Theodlz:
If you need a source and an associated .fits file to try the feature during your review @guynir42, please tell me :)

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+def reduce_image(image, header, obj_id, instrument_id, user_id, detect_cosmics=False):
+
+    try:
+        workdir_sextractor_obj = tempfile.mkdtemp(prefix='sex')
+        workdir_sextractor_obj1 = tempfile.mkdtemp(prefix='sex1')
+        workdir_psfex = tempfile.mkdtemp(prefix='psfex')
+        Session = scoped_session(
+            sessionmaker(bind=DBSession.session_factory.kw["bind"])
+        )
+        session = Session()
```
guynir42, line 106: Is there a reason not to use a context manager here?

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+
+        instrument = session.query(Instrument).get(instrument_id)
+        user = session.query(User).get(user_id)
```
guynir42, line 109: Also we can do this using `select` instead of `query` like in the other handlers.

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+
+        # Parsing its header and print it
+        time = utils.get_obs_time(header, verbose=False)
+        filt = header.get('FILTER')
+
+        wcs = WCS(header)
+        gain = 0.25  # otherwise it is a str
```
guynir42, line 116: Why this value?

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+
+        # Masking saturated stars and cosmic rays
+        # The mask is a binary frame with the same size as the image where True means that this pixel should not be used for the analysis
+        mask = image > 0.9 * np.max(image)
```
guynir42, line 120: This seems fairly arbitrary.

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+
+        if detect_cosmics:
+            cmask, cimage = astroscrappy.detect_cosmics(image, mask, verbose=False)
+            log(f'Done masking cosmics: {np.sum(cmask)} pixels masked')
```
guynir42, line 124: I feel like there are a lot of log statements that may not be needed in production.

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+        log(
+            f'Frame center is {center_ra} {center_dec} radius {center_sr} deg, {pixscale*3600} arcsec/pixel'
+        )
+
+        # ## Reference catalogue
+        # Catalogue name may be any Vizier identifier (ps1, gaiadr2, gaiaedr3, usnob1, gsc, skymapper, apass, sdss, atlas, vsx).
+        # Getting PanSTARRS objects brighter than r=17 mag
+        cat = catalogs.get_cat_vizier(
+            center_ra, center_dec, center_sr, 'ps1', filters={'rmag': '<21'}
+        )
```
guynir42, line 178: Isn't the catalog name an input parameter somewhere?

guynir42: Also it looks like the mag limit is 21, not 17.

guynir42: Also that should probably be a user parameter (at least in the config)

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+            cat_col_mag1='gmag',
+            cat_col_mag2='rmag',
+            max_intrinsic_rms=0.02,
+            order=2,
+            verbose=True,
+        )
+
+        # The code above automatically augments the object list with calibrated magnitudes, but we may also do it manually
+        # obj['mag_calib'] = obj['mag'] + m['zero_fn'](obj['x'], obj['y'])
+        # obj['mag_calib_err'] = np.hypot(obj['magerr'], m['zero_fn'](obj['x'], obj['y'], get_err=True))
```
guynir42, line 216: do we need this?

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+
+        # ## Simple catalogue-based transient detection
+        # Some transients may already be detected by comparing the detected objects with catalogue BUT limited approach
+
+        # Filtering of transient candidates
+        candidates = pipeline.filter_transient_candidates(
+            obj, cat=cat, sr=2 / 3600, verbose=True
+        )
```
guynir42, line 224: I don't know if `verbose=True` is what you'd want in production.

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+        for i, cand in enumerate(filtered):
+            # Create the cutout from image based on the candidate
+            cutout = cutouts.get_cutout(image, cand, 20, mask=mask, header=header)
+            # We may directly download the template image for this cutout from HiPS server - same scale and orientation
+            cutout['template'] = templates.get_hips_image(
+                'PanSTARRS/DR1/r', header=cutout['header']
+            )[0]
+            # We do not have difference image, so it will only display original one, template and mask
+            plots.plot_cutout(cutout, qq=[0.5, 99.9], stretch='linear')
+
```
guynir42, line 245: What is happening here? What is 6238? Why are there hard-coded coordinates for the spherical distance? 
What is XRT enhanced position?

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+        # compute upper limit
+        ra0, dec0, sr0 = astrometry.get_frame_center(
+            wcs=wcs, width=image.shape[1], height=image.shape[0]
+        )
+        pixscale = astrometry.get_pixscale(wcs=wcs)
+
+        zero_fn = m[
+            'zero_fn'
+        ]  # Function to get the zero point as a function of position on the image
+        obj['mag_calib'] = obj['mag'] + zero_fn(obj['x'], obj['y'])
```
guynir42, line 255: I thought this was commented out above?

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+            obj1['mag_calib'] = obj1['mag'] + zero_fn(obj1['x'], obj1['y'])
+
+            # Positional match within FWHM/2 radius
+            oidx, sidx, dist = spherical_match(
+                obj1['ra'],
+                obj1['dec'],
+                sim['ra'],
+                sim['dec'],
+                pixscale * np.median(obj1['fwhm']) / 2,
+            )
```
guynir42, line 316: Why match the object to simulated stars?

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+        psf_model, psf_snapshots = psf.run_psfex(
+            image,
+            mask=mask,
+            checkimages=['SNAPSHOTS'],
+            order=0,
+            verbose=True,
+            _workdir=workdir_psfex,
+        )
+
+        sims = []
```
guynir42, line 271: Add a comment explaining why we are adding simulations here.

skyportal/handlers/api/image_analysis.py
```diff
@@ -0,0 +1,451 @@
...
+            sim['mag_measured'][sidx] = obj1['mag_calib'][oidx]
+            sim['magerr_measured'][sidx] = obj1['magerr'][oidx]
+            sim['flags_measured'][sidx] = obj1['flags'][oidx]
+
+            sims.append(sim)
+
+        sims = vstack(sims)
+
+        # FIXME
+        df = {
```
guynir42, line 329: Certainly not a good habit to call this non-DataFrame `df`. I think it may also be useful to explain what is happening here.

guynir42 (commented):
I have some specific comments on the code. But also I'm wondering in general if this is something we want to do inside skyportal or should this be an external service? I can't imagine a lot of people just uploading their FITS files one by one using the front end. And I can't imagine a skyportal instance replacing people's local pipelines. Maybe I'm not aware of a need for this?

guynir42:
@mcoughlin I should have guessed this is all about the GW. So I understand the use-case at least.

mcoughlin:
@guynir42 Everything is about GW ;)

mcoughlin:
@Theodlz We also discussed this a bit yesterday, and Josh thought that it best if the API call to the backend was internal (i.e.  /api/internal/sources(/[0-9A-Za-z-_\.\+]+)/image_analysis) or something so it isn't potentially spammed by users.

Theodlz:
Hi @mcoughlin . That's a really really good idea actually.

I'm even wondering if we should have a specific permission. Pretty much everyone has the upload data permission so it isn't very restrictive.

mcoughlin:
@Theodlz i think internal should be fine for now, the need to click buttons and so on really slows anyone down from doing a lot here.

static/js/components/ImageAnalysisForm.jsx
```diff
@@ -0,0 +1,215 @@
...
+        type: "string",
+        title: "Astrometric Refinement Method",
+        default: "astropy",
+        enum: astrometric_refinement_method,
+      },
+      matching_radius: {
+        type: "number",
+        title: "Matching Radius",
+        default: 2 / 3600,
+      },
```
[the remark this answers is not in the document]
Theodlz: @mcoughlin what do you think?

skyportal/handlers/api/image_analysis.py
Theodlz: @mcoughlin could you please have a look at Guy's review whenever possible? Some of his comments are about the stdpipe related code which I am not too familiar with. I can't really decide what to do for most of the changes he might be asking for.

skyportal/handlers/api/image_analysis.py
Theodlz: Yes completely. We kept it there for debugging. Are you ok with me removing them during the "last" review? Some changes I or Michael might make to incorporate your changes would benefit from having these logs during testing.

skyportal/handlers/api/image_analysis.py
Theodlz: Same reply as above @guynir42 . I would like to keep this during the review/changes process and remove it later on if that's ok with you of course.

skyportal/handlers/api/image_analysis.py
Theodlz: @mcoughlin

Theodlz:
@mcoughlin I won't be attending today's call as it is during the night, but maybe one thing to discuss is whether or not we should add some tests and GA to try this feature. Perhaps as a separate set of GitHub actions, triggered only when this feature is modified (if that's even possible?).

Theodlz:
@mcoughlin @guynir42 just so nothing get's merged before this is fixed:

Running the exact same code as before, I get an error now with:
```
obj = photometry.get_objects_sextractor(
                image, gain=gain, aper=3.0, edge=10, _workdir=workdir_sextractor_obj
            )
```

I suspect this might be caused by the dependabot that updated astropy to a more recent version. So I will have to look into that...

static/js/components/ImageAnalysisForm.jsx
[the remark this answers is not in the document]
mcoughlin: @Theodlz I agree. Just make sure to divide through by 3600 when passing the radius (so it is in degrees). and add arcseconds to the label here.

skyportal/handlers/api/image_analysis.py
mcoughlin: @Theodlz does with Session() as session: work here?

skyportal/handlers/api/image_analysis.py
mcoughlin: @Theodlz 191.4734, 89.1846 should instead be obj.ra.deg, obj.dec.deg and can remove the XRT comment,

skyportal/handlers/api/image_analysis.py
mcoughlin: So this is just to ensure that you are detecting what you are simulating, as this is a quick and dirty limiting magnitude calculation.

skyportal/handlers/api/image_analysis.py
mcoughlin: @Theodlz See above. It is to make a limiting magnitude assessment.

static/js/components/ImageAnalysisForm.jsx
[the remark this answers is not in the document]
Theodlz: Ok! But should we divide by 3600 in the frontend before sending it to the API call, or tell users: "ok submit this in arcsec, both from the frontend and for API calls" and we divide it in the backend? 

Because if we divide it by 3600 (meaning we expect arsec) only in the frontend, that means that they need to use degree instead for API calls, which seems weird to me right? Submitting 2 different things if you are using the frontend or the API is quite confusing.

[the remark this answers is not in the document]
Theodlz: Oh and also, we don't use most of the params of the frontend form in the backend yet. Like we don't use that radius in the handler for instance.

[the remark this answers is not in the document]
Theodlz: Should we take care of that in that PR?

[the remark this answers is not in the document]
Theodlz: @mcoughlin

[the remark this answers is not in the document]
mcoughlin: I think we should use arcseconds on both front and back-end. And if it will be used by for now-commented code, I think fine to have it.

mcoughlin:
@guynir42 Can you take a look at this?

skyportal/app_server.py
```diff
@@ -472,6 +474,10 @@ def log_request(self, handler):
     (r'/api/internal/notifications/all', BulkNotificationHandler),
     (r'/api/internal/ps1_thumbnail', PS1ThumbnailHandler),
     (r'/api/internal/recent_gcn_events', RecentGcnEventsHandler),
+    (
+        r'/api/internal/sources(/[0-9A-Za-z-_\.\+]+)/image_analysis',
+        ImageAnalysisHandler,
+    ),
```
guynir42, line 480: This means it is not accessible to the users?

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+from ..photometry import add_external_photometry
+from ...base import BaseHandler
+from ....models import Instrument, User, DBSession
+
+_, cfg = load_env()
+
+log = make_log('image_analysis')
+
+try:
+    if cfg['image_analysis'] is True:
```
guynir42, line 31: I'm not sure why we need these "if"s. I think this whole file should not be imported or called if the image analysis is turned off. If we can assume that then we can move the imports to the beginning and get rid of a few conditionals. 
Or maybe I'm missing something?

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+    "usnob1",
+    "gsc",
+]
+
+templates_enum = ["PanSTARRS/DR1/g", "PanSTARRS/DR1/r", "PanSTARRS/DR1/i"]
+
+methods_enum = ["scamp", "astropy", "astrometrynet"]
+
+
+def spherical_match(ra1, dec1, ra2, dec2, sr=1 / 3600):
```
guynir42, line 74: One of the things we could consider putting as a tunable parameter (either in the config or as a user input or both) is the radius for the cross match. This is notoriously hard to pin down to one universal value. It depends on the field density and the survey resolution, among other considerations.

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+    try:
+        idx1, idx2, dist, _ = search_around_sky(
+            SkyCoord(ra1, dec1, unit='deg'), SkyCoord(ra2, dec2, unit='deg'), sr * u.deg
+        )
+
+        dist = dist.deg  # convert to degrees
+
+        return idx1, idx2, dist
+    except Exception as e:
+        raise e
```
guynir42, line 107: I'm not sure what this does...

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+        x *= x
+        y = np.sin(np.deg2rad((dec1 - dec2) / 2))
+        y *= y
+
+        z = np.cos(np.deg2rad((dec1 + dec2) / 2))
+        z *= z
+
+        return np.rad2deg(2 * np.arcsin(np.sqrt(x * (z - y) + y)))
+    except Exception as e:
+        raise e
```
guynir42, line 141: This either

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+    user_id,
+    gain,
+    matching_radius,
+    catalog_name_refinement,
+    catalog_name_crossmatch,
+    template_name,
+    method,
+    detect_cosmics=False,
+):
+    """Reduce an image: Perform astrometric and photometric calibration, and extract photometry to add it to the database.
```
guynir42, line 158: Can we line break this?

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+            user = session.scalars(sa.select(User).where(User.id == user_id)).first()
+
+            time = utils.get_obs_time(header, verbose=False)
+            filt = header.get('FILTER')
+
+            wcs = WCS(header)
+
+            # Masking saturated stars and cosmic rays
+            # The mask is a binary frame with the same size as the image where True means that this pixel should not be used for the analysis
+            mask = image > 0.9 * np.max(image)
```
guynir42, line 199: I would guess you would also want to use "dilate" on this mask before applying it, to get the edges that are not saturated.

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+
+            if detect_cosmics:
+                cmask, cimage = astroscrappy.detect_cosmics(image, mask, verbose=False)
+                log(f'Done masking cosmics: {np.sum(cmask)} pixels masked')
```
guynir42, line 203: Is this log needed or just used for the debugging process?

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+                mask |= cmask
+
+            # ## Detect and measure the objects
+            # The detection is based on building the noise model through (grid-based) background and background rms estimation, and then extracting the groups of connected pixels above some pre-defined threshold
+            # Detecting objects using SExtractor and getting their measurements in apertures with 3 pixels radius
+            # edge argument : control the rejection of objects detected too close to frame edge
+
+            # Astropy Table, ordered by the object brightness
+            obj = photometry.get_objects_sextractor(
+                image, gain=gain, aper=3.0, edge=10, _workdir=workdir_sextractor_obj
```
guynir42, line 213: here are some more tunable parameters. I don't know if you want to address this now, but maybe in a future PR.

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+            )
+
+            if obj is None:
+                raise ValueError(
+                    'No objects detected. If this is unexpected, check source extractor installation.'
+                )
+            log(f'{len(obj)} objects detected')
```
guynir42, line 220: If you are leaving this as a proper log, maybe make the output a little more specific (make it obvious this is image extraction, etc).

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+
+            # First rough estimation of average FWHM of detected objects, taking into account only unflagged ones
+            fwhm = np.median(obj['fwhm'][obj['flags'] == 0])
+            log(f'Average FWHM is {fwhm} pixels')
```
guynir42, line 224: I would consider collecting all of these outputs compressed into one line at the end.

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+            )
+
+            # ## Simple catalogue-based transient detection
+            # Some transients may already be detected by comparing the detected objects with catalogue BUT limited approach
+
+            # Filtering of transient candidates
+            candidates = pipeline.filter_transient_candidates(
+                obj, cat=cat_crossmatch, sr=matching_radius, verbose=True
+            )
+            log(f'{candidates}')
```
guynir42, line 317: I'm definitely getting the feeling these log statements are for debugging and should not go to production as is.

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+
+            # compute upper limit
+            ra0, dec0, sr0 = astrometry.get_frame_center(
+                wcs=wcs, width=image.shape[1], height=image.shape[0]
+            )
+            pixscale = astrometry.get_pixscale(wcs=wcs)
+
+            zero_fn = m[
+                'zero_fn'
+            ]  # Function to get the zero point as a function of position on the image
```
guynir42, line 350: the comment would be better placed before the code.

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+            obj['mag_calib'] = obj['mag'] + zero_fn(obj['x'], obj['y'])
+
+            # We may roughly estimage the effective gain of the image from background mean and rms as gain = mean/rms**2
```
guynir42, line 353:
```suggestion
            # We may roughly estimate the effective gain of the image from background mean and rms as gain = mean/rms**2
```

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+                # Also store measured magnitude, its error and flags
+                sim['mag_measured'][sidx] = obj1['mag_calib'][oidx]
+                sim['magerr_measured'][sidx] = obj1['magerr'][oidx]
+                sim['flags_measured'][sidx] = obj1['flags'][oidx]
+
+                sims.append(sim)
+
+            sims = vstack(sims)
+
+            # FIXME
```
guynir42, line 424: fixme?

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+                return self.error(
+                    message=f'Invalid method, must be once of: {", ".join(methods_enum)}'
+                )
+
+            file_data = data.get("image_data")
+            if file_data is None:
+                return self.error(message='Missing image data')
+
+            filt = data.get("filter")
+            obstime = data.get("obstime")
```
guynir42, line 566: if obstime is none?

skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -0,0 +1,607 @@
...
+            ) as f:
+                file_data = base64.b64decode(file_data[-1])
+                f.write(file_data)
+                f.flush()
+                hdul = fits.open(f.name)
+                header = hdul[0].header
+                image_data = hdul[0].data.astype(np.double)
+                header['FILTER'] = filt
+                header['DATE-OBS'] = obstime.isot
+                # now lets get the file data again, so we can save it later as another file
```
guynir42, line 587: I'm not sure what this comment refers to?

guynir42 (commented):
I have many comments, mostly minor ones. 
I think some of the parameters passed in like matching radius are not used consistently for all the matching that is done, I don't know if that's intended. Same thing for gain calculation. 

Overall it looks good.

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Yes, those along with (a lot more) parameters will be de-hardcoded in a future PR.

skyportal/handlers/api/internal/image_analysis.py
Theodlz: I am not familiar with that process (especially with NumPy), do have any examples in mind? And also, what would you use to do so, a library like openCV?

Theodlz: I have found someone doing something fairly similar at: https://github.com/stubbslab/general_purpose_python_scripts/blob/main/cosmics.py

For instance, they define a method that does:
```
def getdilatedmask(self, size=3):
		"""
		Returns a morphologically dilated copy of the current mask.
		size = 3 or 5 decides how to dilate.
		"""
		if size == 3:
			dilmask = ndimage.morphology.binary_dilation(self.mask, structure=growkernel, iterations=1, mask=None, output=None, border_value=0, origin=0, brute_force=False)
		elif size == 5:
			dilmask = ndimage.morphology.binary_dilation(self.mask, structure=dilstruct, iterations=1, mask=None, output=None, border_value=0, origin=0, brute_force=False)
		else:
			dismask = self.mask.copy()
		
		return dilmask
```
so using scipy

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Agreed. I will for sure add that in the next PR, hard-coding it is not optimal at all. However I guess we need to put some sort of constraints to avoid values that are too large.

skyportal/handlers/api/internal/image_analysis.py
Theodlz: I did that to make sure that someone's app does not crash if they activated the features but did not install all of the dependencies, but that's a little overkill. I will implement your suggestion.

doc/image_analysis.md
```diff
@@ -0,0 +1,89 @@
...
+cd stdpipe
+pip install -e .
+```
+
+## Configuration file
+
+By default, the app allows a maximum request size of 10MB (from the client). This is not enough for someone to upload most .fits files (compressed or not).
+To increase the maximum request size, change the `max_body_size: 10` parameter in the `config.yaml` file to a larger value, like `100` (in MB).
+
+Also, as mentioned earlier, you need to set `image_analysis` to `True` in the `config.yaml` file. By default, it is set to `False`.
```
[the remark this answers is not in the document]
Theodlz: Great idea.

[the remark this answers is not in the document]
Theodlz: However, we don't have tunable parameters to put in that config for now, what you should I do? Should I just put a dummy param like `enabled:true` to make sure that it's clear that it is a dict?

doc/image_analysis.md
```diff
@@ -0,0 +1,89 @@
...
+
+Again, add the same lines mentioned in the Configuration section to the `docker.yaml` file.
+
+## Usage
+
+Go to the page of a source, and add `/image_analysis` at the end of the URL. It should look something like `http://localhost:5000/source/ZTF22aayaehk/image_analysis`.
+Then, fill in the fields required by the form, and upload the image.
+Otherwise, you can find a button labelled `Image Analysis` right under the photometry plot. It will redirect you to the image analysis page.
+
+**The image must be a compressed FITS file (.fits.fz). We recommend compressing it with funpack.**
```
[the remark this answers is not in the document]
Theodlz: Otherwise, it can get way too big for nginx, and it the request doesn't go through. I made sure that the max size was configurable in a PR few weeks ago, but it's still a problem for large images which would require us to allow very large requests (which can be a problem if numerous). 
It's also very easy to compress them, and most libraries (like astropy) can read the compressed file without any problem.

skyportal/app_server.py
Theodlz: Yes. I believe this was the result of a discussion between @mcoughlin and Josh

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Also, that would require more to import the config in the __init__.py file of the internal directory, so that I can conditionally import the handler. Is that alright? @guynir42

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Will remove

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Sure thing

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Will remove

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Hi @mcoughlin @guynir42. Any thoughts about this?

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Hi @mcoughlin. I see different hardcoded values for the aperture throughout the `reduce_image(...)` method. Any reason why:

- There are different values and not just one?
- Why this should be hardcoded?

skyportal/handlers/api/internal/image_analysis.py
Theodlz: @mcoughlin I failed to get an answer from Guy on this one. Any thoughts?

doc/image_analysis.md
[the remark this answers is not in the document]
guynir42: I'm sure you can find something. E.g., photometric aperture radius. But there's no harm in using enable: true.

skyportal/handlers/api/internal/image_analysis.py
guynir42: I think we can import the config anywhere we want and it lazy loads it (if I remember correctly). I don't see why that would be a problem...

doc/image_analysis.md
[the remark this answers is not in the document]
Theodlz: Thanks for your answer guy. I'm changing it as you suggested. Now it just checks if `image_analysis`, has been added to the config, meaning later on it can be something other than `True` if need be

skyportal/handlers/api/internal/image_analysis.py
Theodlz: Of course, it's not a technical problem at all. I was just asking myself if it was good practice / conventional to do so. It has been added

skyportal/app_server.py
mcoughlin: Yeah only through the app.

skyportal/handlers/api/internal/image_analysis.py
mcoughlin: I think we can address this in a follow-on PR.

PR #3194, opened 2022-07-03
Favorites page timeout and scroll_parent

mcoughlin:
This PR adds a timeout and scroll_parent to try to improve reliability of a currently flaky test.

### DIFF
--- a/skyportal/tests/frontend/test_favorites_page.py
+++ b/skyportal/tests/frontend/test_favorites_page.py
@@ -26,11 +26,15 @@ def test_add_remove_favorites(driver, user, public_source):
     # little triangle you push to expand the table
     driver.click_xpath("//*[@id='expandable-button']")
 
+    driver.wait_for_xpath(
+        f"//*[contains(@data-testid, 'favorites-text-include_{public_source.id}')]",
+        timeout=10,
+    )
+
     # click to un-save the source as favorite
     driver.click_xpath(
         f'//*[@data-testid="favorites-text-include_{public_source.id}"]',
         scroll_parent=True,
-        timeout=10,
     )
 
     driver.wait_for_xpath(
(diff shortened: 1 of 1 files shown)


skyportal/tests/frontend/test_favorites_page.py
```diff
@@ -26,8 +26,16 @@ def test_add_remove_favorites(driver, user, public_source):
     # little triangle you push to expand the table
     driver.click_xpath("//*[@id='expandable-button']")
 
+    driver.wait_for_xpath(
+        f"//*[contains(@data-testid, 'favorites-text-include_{public_source.id}')]",
+        timeout=10,
```
guynir42, line 31: Isn't the default timeout 10s? Wouldn't it be better to go straight to 20?

skyportal/tests/frontend/test_favorites_page.py
```diff
@@ -26,8 +26,16 @@ def test_add_remove_favorites(driver, user, public_source):
...
+    )
+
     # click to un-save the source as favorite
-    driver.click_xpath(f'//*[@data-testid="favorites-text-include_{public_source.id}"]')
+    driver.click_xpath(
+        f'//*[@data-testid="favorites-text-include_{public_source.id}"]',
+        scroll_parent=True,
```
guynir42, line 37: I've had to remove a few `scroll_parent=True` statements from various test (I can't remember which ones) but it may be a difference in browser from the GA. Are you sure this actually helps the test pass?

skyportal/tests/frontend/test_favorites_page.py
mcoughlin: Good idea.

skyportal/tests/frontend/test_favorites_page.py
mcoughlin: Good point, removing for now.

guynir42 (approved):
LGTM

Issue #3126, opened 2022-06-21
Increase the size of the spectroscopy plot


mcoughlin:
The new physical size of the plot window shrunk. It should have the same size as the plot of the light curve plot.

From @guynir42 : It may also be useful to make the accordion a bit taller to allow using the controls on the bottom without scrolling.

Copied from https://github.com/fritz-marshal/fritz-beta-feedback/issues/100.

mcoughlin:
Closing for now.

PR #3195, opened 2022-07-04
Increasing size of spectroscopy plot

StephenTheCoder:
Fixes error #3126


mcoughlin:
@StephenTheCoder Please post a screenshot with the updated visual. Also, The upload file is not necessary and should be removed.

StephenTheCoder:
Here is the new spectroscopy. If the changes don't show up, then I think it either didn't load or I would need to make it bigger.
NewSpectro

guynir42 (approved):
I don't see the difference but I'm always in favor of bigger plots.

PR #3197, opened 2022-07-05
Separate observer and observer_timezone

mcoughlin:
This PR fixes a bug where the airmass plots are not reporting UT time, but actually timezone time. This splits out the observer property into one for UT and one that is timezone cognizant, which may be useful for reporting local time on the front-end.


Theodlz:
Looks good ! Some tests are not passing though, but I'll be happy to merge once green.

mcoughlin:
@Theodlz Squashed the bug!

Theodlz (approved):
Great stuff ! I'll merge it now.

PR #3200, opened 2022-07-06
Sources list - Download all sources, not just the ones displayed.

### LINKED ISSUE #3118: Download CSV is ambiguous
**Describe the bug**
Download CSV only downloads the list of alerts that are currently shown rather the full set that meet the current filtering applied.

**To Reproduce**
Filtering on a Sitewide Taxonomy of AGN gives 5333 matches under Nuclear Transient yet Download CSV only downloads whatever the current Rows per page setting is

**Expected behavior**
Download CSV downloads that full current select set (5333 in the example specified) and not just the pagination display setting (a maximum of 200)

Copied from https://github.com/fritz-marshal/fritz/issues/206.

Theodlz:
This PR attempts to fix #3118, where the problem was that when clicking on the download button of the mui datatable of the source list, it only downloads the source currently displayed (so only the current page of the table), and not all the sources matching the filters and/or sorting selected by a user. 

In this PR, the sources are download 10 by 10 which is fine to test the feature. But this could (and probably should) more be done like 100 by 100 to limit API calls.

PREVIEW: (please click on the image to see it fullscreen, makes it easier to read)
download_source_list

### DIFF
--- a/static/js/components/SourceList.jsx
+++ b/static/js/components/SourceList.jsx
@@ -5,11 +5,14 @@ import Paper from "@mui/material/Paper";
 import makeStyles from "@mui/styles/makeStyles";
 import Grid from "@mui/material/Grid";
 import Typography from "@mui/material/Typography";
+import Dialog from "@mui/material/Dialog";
+import DialogContent from "@mui/material/DialogContent";
 
 import * as sourcesActions from "../ducks/sources";
 import UninitializedDBMessage from "./UninitializedDBMessage";
 import SourceTable from "./SourceTable";
 import Spinner from "./Spinner";
+import ProgressIndicator from "./ProgressIndicators";
 
 const useStyles = makeStyles((theme) => ({
   paperDiv: {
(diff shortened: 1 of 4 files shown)


mcoughlin (approved):
I think my favorite part that we start using elsewhere is the progress indicators.

guynir42:
I haven't reviewed yet, but I must admit this is really cool. 

It looks great and I like how you can see exactly what is going on. 

It may become frustrating if someone wants to download a million sources, but I guess that's one of the things we are actually trying to avoid.

static/js/components/SourceList.jsx
```diff
@@ -87,6 +97,65 @@ const SourceList = () => {
...
+          setDownloadProgressCurrent(sourceAll.length);
+          setDownloadProgressTotal(sourcesState.totalMatches);
+        } else if (result && result?.status !== "success") {
+          // break the loop and set progress to 0 and show error message
+          setDownloadProgressCurrent(0);
+          setDownloadProgressTotal(0);
+          if (sourceAll?.length === 0) {
+            dispatch(
+              showNotification(
+                "Failed to fetch some sources. Download cancelled.",
```
guynir42, line 137:
```suggestion
                "Failed to fetch any sources. Download cancelled.",
```

guynir42 (commented):
Looks great. I think this is definitely something we should be testing. 
I would do something like: 
- make a test that posts ~20 sources via API 
- then filter only those sources (e.g., using a unique origin or group or something) 
- then download them into a file and check that the file has the correct number of rows.

Theodlz:
On it !

Theodlz:
Done @guynir42 . Let's just see if it passes on github.

guynir42 (approved):
Looks great, thanks!

mcoughlin:
@Theodlz Is this ready to go in?

Theodlz:
Yep !

PR #3202, opened 2022-07-07
Bring SkyPortal and Fritz telescopes / instruments into alignment

mcoughlin:
This PR adds the missing telescopes / instruments into SkyPortal. The goal will be to merge the fritz/skyportal yamls.


guynir42 (approved):
Seems like a good idea.

PR #3203, opened 2022-07-07
Displaying H0 with full accuracy

StephenTheCoder:
The file should replace core.py in astropy and (hopefully) should fix the error.


mcoughlin:
@StephenTheCoder Is it clear we cannot get away with just changing how we show the results locally? The astropy PR looked a bit a mess :)

stefanv:
Easiest to override the cosmology repr, or write a custom utility function to print the cosmology. Given that we don't have control over what astropy gives us, perhaps the latter.

PR #3204, opened 2022-07-07
Selenium 4.3

mcoughlin:
This PR upgrades selenium to 4.3.


.github/workflows/test_models.yaml
```diff
@@ -139,7 +139,7 @@ jobs:
           fi
           sudo tar -xzf ${CACHED_DOWNLOAD_DIR}/geckodriver-v${GECKO_VER}-linux64.tar.gz -C /usr/local/bin
           geckodriver --version
-          pip install selenium==4.2.0
+          pip install selenium==4.3.0
```
guynir42, line 142: I think we can remove these once we are ok with selenium 4.3.0

guynir42 (approved):

.github/workflows/test_models.yaml
mcoughlin: @guynir42 Should we leave this in until https://github.com/cesium-ml/baselayer/pull/330/files get merged maybe?

guynir42: I'm not sure. If the tests pass without this line we should get rid of it so we don't forget to do it later.

mcoughlin:
@guynir42 Mostly because I am bad at understanding baselayer, I have a merge conflict because I had some pinned version previous to the latest baselayer. I think we want https://github.com/cesium-ml/baselayer/pull/330 in. Could we consider merging that and pinning the new version?

PR #3218, opened 2022-07-12
return self.error instead of raise

mcoughlin:
This PR finishes off replacing raises with return self.error

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -602,7 +602,7 @@ class GroupUserHandler(BaseHandler):
         )
 
         if gu is None:
-            raise AccessError("GroupUser does not exist.")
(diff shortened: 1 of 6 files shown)


guynir42 (approved):

stefanv:
Access errors should be 401, not 403.  We already handle this in baselayer, see base.py:

```
    def write_error(self, status_code, exc_info=None):
        if exc_info is not None:
            err_cls, err, traceback = exc_info
            if isinstance(err_cls, AccessError):
                status_code = 401
        else:
            err = "An unknown error occurred"

        self.error(str(err), status=status_code)
```

stefanv:
Actually, I take that back.

401: "An access token is missing. An access token is expired, revoked, malformed, or invalid for other reasons."

403: Valid access token, but cannot access data requested.

PR #3221, opened 2022-07-12
Make sensitivity_data modifiable from the front-end

mcoughlin:
This PR makes instrument's sensitivity_data modifiable from the front-end.

### DIFF
--- a/skyportal/handlers/api/instrument.py
+++ b/skyportal/handlers/api/instrument.py
@@ -7,6 +7,7 @@ import sqlalchemy as sa
 from tornado.ioloop import IOLoop
 
 import arrow
+import ast
 from healpix_alchemy import Tile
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/instrument.py
```diff
@@ -48,6 +49,9 @@ def post(self):
         )
 
         sensitivity_data = data.get("sensitivity_data", None)
+        if isinstance(sensitivity_data, str):
+            sensitivity_data = ast.literal_eval(sensitivity_data.replace("\'", "\""))
```
guynir42, line 53: If you really only want a dictionary, maybe `json.loads` is more restrictive (in a good way)?

guynir42 (approved):
Minor comment

PR #3225, opened 2022-07-13
Tool to plot access usage by IP address (using nginx logs)

mcoughlin:
This PR adds a small tool to plot IP's for diagnostic purposes.

### DIFF
--- a/tools/nginx_parse.py
+++ b/tools/nginx_parse.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+#
+# Return IP statistics
+#
+
+import sys
+import pandas as pd
+import matplotlib
+import matplotlib.pyplot as plt
+import numpy as np
+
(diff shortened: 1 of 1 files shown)


tools/nginx_parse.py
```diff
@@ -0,0 +1,69 @@
...
+fig = plt.figure(figsize=(30, 30))
+plt.pie(
+    values,
+    labels=labels,
+    shadow=True,
+    startangle=90,
+    autopct=make_autopct(values, label="Requests"),
+)
+plt.axis('equal')
+plt.savefig('requests.pdf')
```
profjsb, line 68: Should make file output name into an optional command-line arguement.

profjsb (commented):
You might look into `https://github.com/skyportal/skyportal/blob/main/tools/direct_insert_to_tables.py` for a recent example of using `fire` to make command-line tools for SP.

profjsb (approved):

PR #3227, opened 2022-07-13
Move the user_expiration test to flaky

mcoughlin:
This PR moves the user_expiration test to flaky. Passes just fine locally.


profjsb (approved):
LGTM

PR #3228, opened 2022-07-13
Wrap source query in try except

mcoughlin:
This PR wraps the source query in a try: except to return a self.error.


skyportal/handlers/api/source.py
```diff
@@ -2020,14 +2020,35 @@ class Validator(Schema):
...
+                        include_spectrum_exists=include_spectrum_exists,
+                        include_period_exists=include_period_exists,
+                        include_detection_stats=include_detection_stats,
+                        is_token_request=is_token_request,
+                        include_requested=include_requested,
+                        requested_only=requested_only,
+                        include_color_mag=include_color_mag,
+                    )
+                except (ValueError, arrow.ParserError) as e:
+                    return self.error(f'Cannot retrieve source: {str(e)}')
```
guynir42, line 2042: Why not grab any kind of exception?

mcoughlin: Yes fixed.

guynir42 (approved):

PR #3229, opened 2022-07-13
Add an endpoint piechart

mcoughlin:
This PR adds a piechart for the endpoints too.

### DIFF
--- a/tools/nginx_ip_and_requests.py
+++ b/tools/nginx_ip_and_requests.py
@@ -8,6 +8,7 @@ import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
 import fire
+import urllib
 
 font = {'family': 'normal', 'weight': 'bold', 'size': 24}
 matplotlib.rc('font', **font)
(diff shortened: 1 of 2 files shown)


stefanv:
I would avoid hardcoding endpoints and give the script a more descriptive name.

mcoughlin:
@stefanv What is the proper way to get endpoints?

stefanv:
I don't see why you need them, you can just parse them out of the log line. But if you really need them you can get them from the openapi spec JSON.

stefanv:
You mean like this?

```
In [4]: import urllib

In [5]: urllib.parse.urlparse('/api/foo?x=3&foo=3')
Out[5]: ParseResult(scheme='', netloc='', path='/api/foo', params='', query='x=3&foo=3', fragment='')
```

mcoughlin:
@stefanv Ah. Are you thinking of just a standard loop here over the frame?

stefanv:
Yes, exactly. Pandas has something for that, `apply` maybe?

mcoughlin:
@stefanv I think a loop is maybe easiest after all.

guynir42 (approved):

PR #3230, opened 2022-07-13
Try decreasing rate limits to improve app stability

mcoughlin:
This PR decreases the rate limits to improve app stability.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -396,7 +396,7 @@ server:
   dedicated_frontend_processes: 2
 
   # Rate limit: number of requests per second (see https://www.nginx.com/blog/rate-limiting-nginx/)
-  rate_limit: 5
+  rate_limit: 1
   # Rate limit burst size (https://www.nginx.com/blog/rate-limiting-nginx/#bursts)
   burst: 10
 


profjsb (commented):
I'm not in favor of this since I think we might have a more fundamental issue -- such a rate limit will make the app very unresponsive for the end user unless we can enforce this to only apply to external/non Kowalski API calls. We might need a `delay` two-stage burst with such a slow rate (www.nginx.com/blog/rate-limiting-nginx/amp/). In any event I'd ask @stefanv to weigh in.

stefanv:
Kowalski is on a whitelist, AFAIR, so it should not be affected.

We should decide what our time limits are on individual requests.  The goal was always to make that time very short.  But, no matter how short you make it, it is possible for someone to try and connect and for a backend not to be available.

That is why we have the fallback status server, but as mentioned before it does not seem to work as intended (I'd investigate that first).

Thereafter, it would be worth adding some dedicated backends for the frontend.  This can be done by, e.g., sending an encrypted header along with JS API requests.

This change could be problematic for some of the brokers out there, but will not necessarily solve the problem: Kowalski likely still occupies the system almost completely when it is active (although most of its requests are short-lived HEAD requests).

A high priority issue to investigate is exactly how GCP determines app health.  If GCP decides we are down, it will refuse to serve any requests for a while.  This should be avoided at all costs, but seems to be what is happening. I suspect it is because the status server gives 503s and not 200s.

mcoughlin:
@stefanv Could you say more about where the change in the JS API request would be (in baselayer somewhere?). Also what the fallback server is?

stefanv:
fallback server is in baselayer: `services/status_server`

API calls are made from the frontend from `API.js`. I have not figured out a good way to tell the backend that we're phoning from the frontend, at least not a method that cannot be abused by scripts to get faster API rates.

guynir42:
@mcoughlin what's the status on this? should we bring it up again in tomorrows meeting?

Issue #3238, opened 2022-07-14
Make GcnEvents table sortable by column


mcoughlin:
Currently sorting does not do anything.

mcoughlin:
@Theodlz Anything more to be done here?

Theodlz:
Hey @mcoughlin. In the last PR i added the frontend and backend necessary to start implementing the sorting. Then, the only one I added is the sorting by dateobs. I'm not sure what we could sort the event by? We can definitely filter them (that's something to add next for sure). We talked about the tags but even then I'm not sure it really works. If events had ONE tag each maybe we could sort by alphabetic order, but here what could we do?

mcoughlin:
@Theodlz ah yes filtering is the clear thing to be doing. Good point on that.

PR #3239, opened 2022-07-15
GCN Events List - Sorting

Theodlz:
This PR addresses #3238. I added the necessary frontend and backend to sort the results. For now, I only implemented the sorting by dateobs, but we could add sorting on the other columns. However, for the other columns we have to define on what to sort.


mcoughlin:
@Theodlz I think the other one that would be nice is the filter function, which should filter on the tags. That could be a separate PR though, what do you think?

Theodlz:
> @Theodlz I think the other one that would be nice is the filter function, which should filter on the tags. That could be a separate PR though, what do you think?

Definitely, I thought the same. Filtering by start and end date on the dateobs and filtering by tags would be super helpful.
However, it makes sense to implement these in 2 separate PRs.
I'll be happy to take care of the 2nd PR.
If we focus on the sorting in this PR, should we add other parameters/columns to sort by? I'm not sure what is useful aside from the dateobs though.

mcoughlin (approved):
@Theodlz LGTM.

PR #3242, opened 2022-07-15
Wrap GCN Service in Try Except to prevent errors from killing the service.

Theodlz:


Theodlz:
I'm wondering if we should even wrap the `gcn.listen(handler=handle)` in a try except too, I'm not sure how leo's listener works and if errors could be encountered at that level.

PR #3246, opened 2022-07-16
Move SourceHandler to the new select syntax

mcoughlin:
This PR moves the SourceHandler to the new select syntax.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -75,6 +75,8 @@ MAX_SOURCES_PER_PAGE = 500
 _, cfg = load_env()
 log = make_log('api/source')
 
+MAX_LOCALIZATION_SOURCES = 50000
+
 
 def get_source(
     obj_id,
@@ -102,7 +104,7 @@ def get_source(
     See Source Handler for optional arguments
     """
 
-    user = session.query(User).get(user_id)
+    user = session.scalar(sa.select(User).where(User.id == user_id))
 
     options = []
     if include_thumbnails:
@@ -187,9 +189,7 @@ def get_source(
                         joinedload(Comment.author),
                         joinedload(Comment.groups),
                     ],
-                )
-                .where(Comment.obj_id == obj_id)
-                .distinct()
+                ).where(Comment.obj_id == obj_id)
             )
             .unique()
             .all()
@@ -226,7 +226,6 @@ def get_source(
             Annotation.select(user)
             .options(joinedload(Annotation.author))
             .where(Annotation.obj_id == obj_id)
-            .distinct()
         )
         .unique()
         .all(),
@@ -270,9 +269,7 @@ def get_source(
             ).first()
             is not None
         )
-    source_query = Source.query_records_accessible_by(user).filter(
-        Source.obj_id == source_info["id"]
-    )
+    source_query = Source.select(user).where(Source.obj_id == source_info["id"])
     source_query = apply_active_or_requested_filtering(
         source_query, include_requested, requested_only
     )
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/source.py
```diff
@@ -669,18 +662,16 @@ def get_sources(
                 )
             )
         )
-        classification_accessible_subquery = Classification.query_records_accessible_by(
-            user
-        ).subquery()
+        classification_accessible_subquery = Classification.select(user).subquery()
```
guynir42, line 665: I'm not sure if this has been defined twice or why?

skyportal/handlers/api/source.py
```diff
@@ -740,6 +731,14 @@ def get_sources(
                 "Invalid commentsFilterAuthor value -- must provide at least one string value"
             )
     if localization_dateobs is not None:
+
+        # This grabs just the IDs so the more expensive localization in-out
+        # check is done on only this subset
+        obj_ids = session.scalars(obj_query).all()
```
guynir42, line 737: What happens if the filters above don't filter anything and you get back *all* the obj ids?

skyportal/handlers/api/source.py
```diff
@@ -740,6 +731,14 @@ def get_sources(
...
+        obj_query = Obj.select(user, options=obj_query_options).where(
+            Obj.id.in_(obj_ids)
+        )
+
         if localization_name is not None:
             localization = (
                 Localization.query_records_accessible_by(user)
```
guynir42, line 744: You forgot one `query_records_accessible_by`

skyportal/handlers/api/source.py
```diff
@@ -1203,7 +1203,11 @@ def post_source(data, user_id, session):
 
     user = session.query(User).get(user_id)
```
guynir42, line 1204: This can also be updated to new style.

skyportal/handlers/api/source.py
```diff
@@ -1232,20 +1236,20 @@ def post_source(data, user_id, session):
...
+        session.add(obj)
 
     if (ra is not None) and (dec is not None):
         # This adds a healpix index for a new object being created
         obj.healpix = ha.constants.HPX.lonlat_to_healpix(ra * u.deg, dec * u.deg)
 
-    groups = (
-        Group.query_records_accessible_by(user).filter(Group.id.in_(group_ids)).all()
-    )
+    groups = session.scalars(Group.select(user).where(Group.id.in_(group_ids))).all()
```
guynir42, line 1252: Should we also test that all groups requested by `group_ids` are accessible (i.e., they were indeed loaded by the select)?

skyportal/handlers/api/source.py
mcoughlin: There is one for classifications and one for nonclassifications.

skyportal/handlers/api/source.py
mcoughlin: I will put a cap and raise an error.

skyportal/handlers/api/source.py
mcoughlin: good idea.

skyportal/handlers/api/source.py
mcoughlin: oops.

skyportal/handlers/api/source.py
mcoughlin: good idea.

skyportal/handlers/api/source.py
```diff
@@ -173,26 +168,30 @@ def get_source(
...
+            session.scalars(
+                Comment.select(
+                    user,
+                    options=[
+                        joinedload(Comment.author),
+                        joinedload(Comment.groups),
+                    ],
+                )
+                .where(Comment.obj_id == obj_id)
+                .distinct()
```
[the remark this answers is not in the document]
mcoughlin: Removed distinct.

guynir42 (approved):

PR #3248, opened 2022-07-17
Use new syntax for Photometry handler

mcoughlin:
This PR uses the new syntax for the Photometry handler. I noticed that with DBSession() as session is still required for handlers that require being able to create a table lock, i.e.
            session.execute(
                f'LOCK TABLE {Photometry.__tablename__} IN SHARE ROW EXCLUSIVE MODE'
            )

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -1367,7 +1367,7 @@ class PhotometryOriginHandler(BaseHandler):
             origins = (
                 session.scalars(sa.select(Photometry.origin).distinct()).unique().all()
             )
-            return self.success(data=[origin[0] for origin in origins])
+            return self.success(data=origins)
 
 
 PhotometryHandler.get.__doc__ = f"""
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/photometry.py
```diff
@@ -1265,35 +1303,45 @@ def get(self):
...
-            mjd = Time(min_date, format='datetime').mjd
-            query = query.filter(Photometry.mjd >= mjd)
-        if max_date is not None:
-            mjd = Time(max_date, format='datetime').mjd
-            query = query.filter(Photometry.mjd <= mjd)
-
-        query = query.join(
-            group_phot_subquery, Photometry.id == group_phot_subquery.c.photometr_id
-        )
+            print(instrument_ids, min_date, max_date)
```
guynir42, line 1316:
```suggestion
```

skyportal/models/phot_stat.py
```diff
@@ -71,9 +71,9 @@ def __init__(self, obj_id):
         self.mag_rms_global = None
         self.mag_rms_per_filter = {}
 
-    read = public
+    read = update = public
 
-    write = update = delete = restricted
+    write = delete = restricted
```
guynir42, line 76: Why is this? Is it because you can't update upon delete/add a photometry point?

guynir42 (approved):
Minor comments:

skyportal/models/phot_stat.py
mcoughlin: Yes.

PR #3251, opened 2022-07-19
Add custom wavelength

thomasculino:
This PR aims to add the ability to add a custom wavelength as a vertical line on the spectroscopy plot as you can see below:
Single_Custom_Wavelength

For now, it only supports one wavelength. Let me know if I should work on adding multiple custom wavelengths. 

I was also thinking that we should change the row with all the sliders because they do not work properly


skyportal/plot.py
```diff
@@ -2329,12 +2329,36 @@ def make_spectrum_layout(obj, spectra, user, device, width, smoothing, smooth_nu
...
+        mode='float',
+    )
 
     # Track elements that need to be shifted with change in z / v
     shifting_elements = []
     renderers = []
     obj_redshift = 0 if obj.redshift is None else obj.redshift
 
+    flux_values = list(np.linspace(ymin, ymax, 100))
+    flux_values[-1] = np.nan
```
guynir42, line 2344: Putting NaN in the last value makes the whole line disappear?

guynir42: I'm not sure where `flux_values` is used?

skyportal/plot.py
```diff
@@ -2386,6 +2410,29 @@ def make_spectrum_layout(obj, spectra, user, device, width, smoothing, smooth_nu
...
+    w_input.js_on_change(
+        'value',
+        CustomJS(
+            args={'input': w_input, 'model_dict': model_dict},
+            code="""
+                    model_dict['custom_line'].data_source.data['x'][0] = input.value;
+                    model_dict['custom_line'].visible = true;
+                    model_dict['custom_line'].data_source.data['wavelength'][0] = input.value;
+                    model_dict['custom_line'].data_source.change.emit();
+                """,
```
guynir42, line 2426: Is there a way to make the line not visible?

guynir42 (commented):
I have a few small questions.

skyportal/plot.py
thomasculino: you mean on the frontend ?

guynir42: Yeah. After I put in a value and see the line, can I make it disappear again or do I have to refresh the page?

skyportal/plot.py
thomasculino: I copied and pasted what was below to make it work but I ended up removing flux_values from el_data, so I should move it back where it was, above the for loop line 2362.

skyportal/plot.py
thomasculino: Currently, there is no button to make this line invisible but I can add it

guynir42: I would advocate that if the user puts in an empty value it should make the line disappear.

skyportal/plot.py
guynir42: OK I guess I just had to expand the view a little more to see it. 
Also I'm not clear on what a NaN on the last value does.

skyportal/plot.py
thomasculino: ok I will try this

mcoughlin:
@thomasculino Any progress on this?

thomasculino:
> @thomasculino Any progress on this?

@mcoughlin no, it still works for one wavelength but not for multiple ones.

thomasculino:
@mcoughlin I think it is a good idea to open that issue

mcoughlin:
@guynir42 Can we get this merged just with the one line done? Then we can open an issue?

guynir42:
@mcoughlin one line is better than none. This is very useful as-is. Multiple lines should be a separate PR.

mcoughlin:
@thomasculino Did you address the nan question?

guynir42:
@thomasculino My only outstanding issue is how to remove that line if I don't want it anymore. I would expect if I input an empty value it should make it disappear.

thomasculino:
@guynir42 I'm going to take a look at this again

thomasculino:
> @thomasculino Did you address the nan question?

The nan was already there in the code and in the end I didn't touch this part. Honestly, I don't really know its purpose

skyportal/plot.py
thomasculino: I followed your advice and it's working well

guynir42:
@thomasculino let me know when ready for a last review.

guynir42 (approved):
Great!

mcoughlin:
@thomasculino ready to merge?

thomasculino:
> @thomasculino ready to merge?

@mcoughlin yes!

PR #3255, opened 2022-07-19
PhotStat minor updates

guynir42:
This PR makes some minor updates to the PhotStat infrastructure. 
- It adds a re-calculation of PhotStats when photometry bulk delete is used. 
- It makes the updates of photometry use all points (regardless of access rights) when calculating PhotStats in all cases. 
- It moves the PhotStats handler to the new select style and uses the verified session. 
- It returns a meaningful error for any failures to generate/update a phot stats through the `PhotStatsUpdateHandler`.

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -1120,6 +1120,8 @@ class PhotometryHandler(BaseHandler):
                     sa.select(Photometry).where(Photometry.obj_id == photometry.obj_id)
                 ).all()
                 phot_stat.full_update(all_phot)
+                for phot in all_phot:
+                    session.expunge(phot)
 
             session.commit()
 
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

guynir42:
@mcoughlin I fixed some tests so probably could do with a re-review.

skyportal/handlers/api/photometry.py
```diff
@@ -1116,16 +1116,12 @@ def patch(self, photometry_id):
...
-                    )
-                    .unique()
-                    .all()
-                )
+                all_phot = session.scalars(
+                    sa.select(Photometry).where(Photometry.obj_id == photometry.obj_id)
+                ).all()
                 phot_stat.full_update(all_phot)
+                for phot in all_phot:
+                    session.expunge(phot)
```
mcoughlin, line 1124: I guess this was the addition?

mcoughlin (approved):
I haven't seen expunge before!

skyportal/handlers/api/photometry.py
guynir42: Among other things. You can see what changed since your last review by looking at the commits picker in review mode.

mcoughlin:
@guynir42 This looks good. Feel free to merge before deploying.

PR #3256, opened 2022-07-20
Routing changes on the shift page

thomasculino:
This PR aims to solve a problem we had on the shift page. Currently, using the routing (i.e /shifts/<shift_id>) gets you to the shift page and automatically opens the calendar on the right week and displays the shift management menu, comments and summary of the shift which id has been passed in the route. However, if you already had selected a shift (which is stored as the current shift in the redux store), this does not work.

This PR adresses that issue.

### DIFF
--- a/static/js/actions.js
+++ b/static/js/actions.js
@@ -15,7 +15,6 @@ import * as telescopesActions from "./ducks/telescopes";
 import * as taxonomyActions from "./ducks/taxonomies";
 import * as favoritesActions from "./ducks/favorites";
 import * as rejectedActions from "./ducks/rejected_candidates";
-import * as shiftsActions from "./ducks/shifts";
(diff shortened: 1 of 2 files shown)


mcoughlin (commented):
@thomasculino If there are actually no shifts, should we have some kind of statement "No shifts available" or similar? Or maybe this is just sufficient. up to you

Theodlz:
> @thomasculino If there are actually no shifts, should we have some kind of statement "No shifts available" or similar? Or maybe this is just sufficient. up to you

Done :heavy_check_mark:

mcoughlin (approved):
LGTM

mcoughlin:
@Theodlz Please review and merge if comfortable.

Theodlz (approved):
LGTM.

PR #3258, opened 2022-07-20
Fix group frontend tests

Theodlz:
This PR addresses a bug introduces by a PR that added logs to the notifications. It seems that some notifications, instead of having a user object as a target, have a user_id. The code added to the user notification model did not take that into account. This fixes that issue and allows tests to pass.


mcoughlin (approved):
@Theodlz Should you raise some kind of error in the else? Otherwise, looks good to me.

Theodlz:
Good idea. Should we log or raise an error ? In both cases, should it be something like: "No user or user_id given as a notification target." ?

mcoughlin:
@Theodlz Let's log so as to not bring down the app... I think. The message seems fine.

PR #3259, opened 2022-07-21
Telescope and Instruments - Permissions to add, modify and delete

Theodlz:
This PR makes sure that the permissions to add telescopes and instruments are the same. Right now in prod, a user has to be a super admin to add its instrument. And has to have the manage source permission to modify and delete a telescope, which seems incorrect?


mcoughlin:
@Theodlz Looks like this one needs a bit more work.

Theodlz:
Hey Michael. Yes this is a complete work in progress. Not done at all I just spent an hour trying to figure out permission stuff, nothing is working yet.

skyportal/handlers/api/instrument.py
```diff
@@ -487,104 +496,137 @@ def put(self, instrument_id):
...
                 return self.error(
-                    'field_region or field_fov_type is required with field_data'
+                    'Invalid/missing parameters: ' f'{e.normalized_messages()}'
                 )
 
-            if type(field_data) is str:
-                field_data = pd.read_table(StringIO(field_data), sep=",").to_dict(
-                    orient='list'
-                )
+            if 'name' in data:
```
mcoughlin, line 569: Maybe we do something similar to the spectrum handler here?

mcoughlin: https://github.com/skyportal/skyportal/blob/main/skyportal/handlers/api/spectrum.py#L811

skyportal/models/instrument.py
```diff
@@ -43,10 +55,21 @@ def process(value):
...
 
 
+def manage_instrument_access_logic(cls, user_or_token):
+    if user_or_token.is_system_admin:
+        return DBSession().query(cls)
+    elif 'Manage allocations' in [acl.id for acl in user_or_token.acls]:
+        return DBSession().query(cls)
+    else:
+        # return an empty query
+        return DBSession().query(cls).filter(cls.id == -1)
```
mcoughlin, line 65: Does this style of empty query appear elsewhere? Kind of interesting.

mcoughlin (approved):
I think the patch should emulate the spectrum handler, which seems cleanest. Otherwise looks good to me.

skyportal/handlers/api/instrument.py
Theodlz: Thanks for the suggestion! I'll have a look at it tomorrow.

skyportal/models/instrument.py
Theodlz: That's the problem, I didn't find anywhere else where that was done. In all other handler it would join the table to another, then to another .... until it reaches somehow the acls to filter on that, or join the table on itself, ... I haven't found a clear example of how to return an empty query. I tried writing it that way as it made sense to me and seemed simpler to understand. But I can change that if it seems incorrect.

mcoughlin: @Theodlz honestly it makes sense to me, I just wasn't sure if there was already precedent of some kind!

mcoughlin (approved):
LGTM!

PR #3262, opened 2022-07-22
Log queries returning a lot of data

mcoughlin:
This PR adds a log to source and candidate handlers for large queries.


profjsb (approved):
Should also capture and return the time it takes to assemble the (large) data. Could just do a start=time.time() at the top of the GET and only return that time if the size is large

profjsb:
are the test failures unrelated?

guynir42 (approved):

PR #3263, opened 2022-07-22
Group admission request: Notify group admins by email

Theodlz:
This PR addresses an issue posted on the beta-testing channel on Slack. Some group admins complained that they were not notified clearly enough that users asked to join their group. This ensures that they are notified by email.

However, in order to work without modifying the current notification framework too much, it adds very little code that simply "bypasses" some of the verification logic to be sure that an email is sent, regardless of the user preferences.
It would be nice for reviewers to test this new feature, as well as existing features of the notification framework to make sure that everything else still works as intended.

### DIFF
--- a/skyportal/handlers/api/group_admission_request.py
+++ b/skyportal/handlers/api/group_admission_request.py
@@ -177,25 +177,6 @@ class GroupAdmissionRequestHandler(BaseHandler):
         )
         DBSession().add(admission_request)
 
-        group_admin_gu = (
-            GroupUser.query.filter(GroupUser.group_id == group_id)
-            .filter(GroupUser.admin.is_(True))
-            .first()
-        )
-        group_admin = (
-            User.query.get(group_admin_gu.user_id)
-            if group_admin_gu is not None
-            else None
-        )
-        if group_admin is not None:
-            DBSession().add(
-                UserNotification(
-                    user=group_admin,
-                    text=f"*@{requesting_user.username}* has requested to join *{group.name}*",
-                    url=f"/group/{group_id}",
-                )
-            )
-
         try:
             self.verify_and_commit()
         except AccessError as e:
@@ -205,8 +186,6 @@ class GroupAdmissionRequestHandler(BaseHandler):
             )
 
         self.push(action="skyportal/FETCH_USER_PROFILE")
-        if group_admin is not None:
-            self.flow.push(group_admin.id, "skyportal/FETCH_NOTIFICATIONS", {})
         return self.success(data={"id": admission_request.id})
 
     @permissions(["Upload data"])
--- a/skyportal/models/user_notification.py
+++ b/skyportal/models/user_notification.py
@@ -188,7 +188,6 @@ def send_slack_notification(mapper, connection, target):
 def send_email_notification(mapper, connection, target):
     resource_type = notification_resource_type(target)
     prefs = user_preferences(target, "email", resource_type)
-    print(prefs)
     if not prefs:
         return
 
@@ -228,7 +227,6 @@ def send_email_notification(mapper, connection, target):
 
     if subject and body and target.user.contact_email:
         try:
-            print("tried sending email")
             send_email(
                 recipients=[target.user.contact_email],
                 subject=subject,
--- a/skyportal/models/user_notification.py
+++ b/skyportal/models/user_notification.py
@@ -109,19 +109,19 @@ def user_preferences(target, notification_setting, resource_type):
             return
         if not target.user.contact_email:
             return
+        # this ensures that an email is sent regardless of the user's preferences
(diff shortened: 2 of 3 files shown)


skyportal/handlers/api/group_admission_request.py
```diff
@@ -205,8 +186,6 @@ def post(self):
             )
 
         self.push(action="skyportal/FETCH_USER_PROFILE")
```
Theodlz, line 188: I'm not sure why there is `self.push(action="skyportal/FETCH_USER_PROFILE")` in the first place.

mcoughlin: I wouldn't be surprised if it was put into there by accident ;)

skyportal/models/user_notification.py
```diff
@@ -100,21 +101,29 @@ def user_preferences(target, notification_setting, resource_type):
         return
     if not target.user:
         return
-
-    if not target.user.preferences:
+    if not notification_setting:
```
guynir42, line 104: If I understand correctly `notification_setting` is a global boolean that turns all notifications on/off?

skyportal/models/user_notification.py
```diff
@@ -100,21 +101,29 @@ def user_preferences(target, notification_setting, resource_type):
...
         if not target.user.contact_email:
             return
-    elif notification_setting == "sms":
+    if notification_setting == "sms":
         if client is None:
             return
         if not target.user.contact_phone:
             return
-    elif notification_setting == "slack":
+
```
guynir42, line 117: Is there a reason this is stuck between sms and slack? Would maybe make more sense right after or even merged in with the `if notification_setting == "email"`.

guynir42 (approved):
Some small comments/questions:

skyportal/models/user_notification.py
Theodlz: No! The method `user_preferences(target, notification_setting, resource_type)` takes a target (that should contain a user, text and URL), a notification_setting (not a boolean, but a string that can be "email", "SMS" or "slack") and a `resource_type` (a string like "sources", "group_admission_request", ...).
This method aims to return the user `preferences` IF there are any for the user in the target, for a given `resource_type`, for a given `notification_setting`. In the methods right below it, that send either slack, emails or SMS notification, we simply pass the associated `notification_setting`.
Then, in those methods, if there are no user preferences for that `resource_type` + `notification_setting`, we do not send a notification. If the method returned preferences (meaning that the user activated that notification type), we send the notification accordingly.

Now, I took the opportunity of adding some verification on the `notification_setting`. As the method is only called in 3 places, and the notification_setting is always either "email", "SMS", or "slack", it should not be None, ever. But I thought that adding it would prevent any errors if the user_notification models are modified later on, let's say to add new notification_settings for example.

skyportal/models/user_notification.py
Theodlz: No you are right, I moved it up :) inside of the `if notification_setting == "email"`. However, the:
```
if not target.user.preferences:
        return
```
will be right between email, and SMS slack, simply because of that feature of "forced" notifications only on emails. Plus, for both slack and SMS, some things are needed in the preferences, down the line. For emails, it's simply about if it is activated in the preferences or not, whereas for the SMS and slack, some values in the preferences are needed in their respective notification methods (slack integration URL, SMS on shift or on a time slot...).

skyportal/models/user_notification.py
guynir42: Ok that makes sense. I would maybe change the parameter name to notification_type just to make that clearer.

skyportal/models/user_notification.py
guynir42: That seems fine, as long as there's a logical reason to be stuck there.

skyportal/models/user_notification.py
guynir42: Also I would either make this conditional `if notification_setting is None` to be explicit, or if you really want to also catch things like empty string, at least comment that this is what you are looking to grab. 
I would also consider just doing `if not isinstance(notification_setting, str)` which really does what you want, I think.

skyportal/models/user_notification.py
```diff
@@ -327,6 +346,26 @@ def receive_after_flush(session, context):
...
+                    GroupUser.admin.is_(True),
+                )
+            ).all()
+            for gu in group_admins_gu:
+                group_admin = (
+                    session.scalars(
+                        sa.select(User).where(User.id == gu.user_id)
+                    ).first()
+                    if gu is not None
+                    else None
```
guynir42, line 363: I'm not sure this is necessary.

guynir42 (approved):
I have some minor comments but overall looks good.

guynir42 (approved):

PR #3264, opened 2022-07-22
fix minor bug in phot stat handler

guynir42:
I accidentally used `last_full_update` instead of `last_update` in one of the filters.


mcoughlin (approved):
LGTM

PR #3265, opened 2022-07-24
verify only one PhotStat per Obj

guynir42:
Turns out there were some cases where a single `Obj` had multiple `PhotStat` associated with it. 
This closes this loophole and uses the PATCH method to delete any additional `PhotStat` rows beyond the one getting a fresh update.


mcoughlin (approved):
Nice catch. LGTM.

PR #3266, opened 2022-07-25
Sources in GCN Events - Confirm or Reject a Source

Theodlz:
### Desciption
This WIP PR allows one to confirm or reject a source found in the localization(s) of a gcn event.
This is done so we can confirm which source is at the origin of a given event, and on the other hand which sources can't be at the origin of an event.

### What is missing?

- Maybe renaming handlers, model and API route.
- Maybe Integrating it with the gcn summary? But that could be the subject of another PR.
- Adding missing display on the Source Mobile Page

### Preview
CONFIRM_REJECT_SOURCE


skyportal/handlers/api/sources_in_gcn.py
```diff
@@ -0,0 +1,183 @@
...
+            return self.error("sourceId must be an integer")
+        if not isinstance(confirmed_or_rejected, bool):
+            return self.error("confirmed_or_rejected must be a boolean")
+
+        if start_date is None:
+            return self.error(message="Missing startDate")
+        if end_date is None:
+            return self.error(message="Missing endDate")
+
+        with self.Session() as session:
```
mcoughlin, line 45: This session should carry down to the end of the handler I expect.

skyportal/handlers/api/sources_in_gcn.py
```diff
@@ -0,0 +1,183 @@
...
+                num_per_page=MAX_SOURCES_PER_PAGE,
+                sourceID=source_id,
+            )
+
+            sources = sources['sources']
+
+        if len(sources) == 0:
+            return self.error("No sources found")
+
+        localization = Localization.query.filter(
```
mcoughlin, line 64: Use the new style query here.

skyportal/handlers/api/sources_in_gcn.py
```diff
@@ -0,0 +1,183 @@
...
+    @auth_or_token
+    def get(self, dateobs):
+        sources_id_list = self.get_query_argument('sourcesIdList', '')
+
+        if not isinstance(sources_id_list, str):
+            return self.error("sourcesIdList must be a comma separated string")
+
+        sources_id_list = sources_id_list.split(',')
+
+        gcn_event = GcnEvent.query.filter(GcnEvent.dateobs == dateobs).first()
```
mcoughlin, line 98: New style query.

skyportal/handlers/api/sources_in_gcn.py
```diff
@@ -0,0 +1,183 @@
...
+                SourcesInGCN.obj_id == source_id,
+                SourcesInGCN.dateobs == dateobs,
+            )
+            .first()
+        )
+        if not source_in_gcn:
+            return self.error("Source is not confirmed or rejected in this GCN event")
+        source_in_gcn.confirmed_or_rejected = confirmed_or_rejected
+
+        DBSession.commit()
```
mcoughlin, line 142: with self.Session() as session:

skyportal/tests/api/test_gcn.py
```diff
@@ -450,10 +459,7 @@ def test_gcn_summary_galaxies(
 def test_gcn_summary_observations(
     super_admin_user,
     super_admin_token,
-    view_only_token,
```
mcoughlin, line 453: Thanks for cleaning these up!

static/js/components/GcnSelectionForm.jsx
```diff
@@ -190,6 +194,8 @@ const GcnSelectionForm = ({ gcnEvent, setSelectedLocalizationName }) => {
       await dispatch(
         sourcesActions.fetchGcnEventSources(gcnEvent.dateobs, formData)
       );
+      console.log(formData);
```
mcoughlin, line 197: Remove console.log

static/js/components/SourceDesktop.jsx
```diff
@@ -230,8 +231,11 @@ const SourceDesktop = ({ source }) => {
   }
   const specIDs = spectra ? spectra.map((s) => s.id).join(",") : "";
 
+  const associatedGCNs = useSelector((state) => state.source.associatedGCNs);
+
   useEffect(() => {
     dispatch(spectraActions.fetchSourceSpectra(source.id));
+    dispatch(sourceActions.fetchAssociatedGCNs(source.id));
```
mcoughlin, line 238: We expect this is still fast enough on a DB with a lot of GCNs because its just that new table right?

static/js/ducks/sourcesingcn.js
```diff
@@ -0,0 +1,54 @@
...
+  );
+
+export const patchSourceInGcn = (dateobs, source_id, data) =>
+  API.PUT(
+    `/api/sources_in_gcn/${dateobs}/${source_id}`,
+    PATCH_SOURCE_IN_GCN,
+    data
+  );
+
+// Websocket message handler
```
mcoughlin, line 32: Do we need this commented code?

mcoughlin (changes requested):
@Theodlz Very nice. The handlers should use Guy's syntax, which is my main comment. Also I just want to be sure the source page query should still be fast even with a lot of objects. Otherwise, looks good to me.

static/js/components/SourceDesktop.jsx
Theodlz: Yep! its just fetching a few rows on that new table.

static/js/ducks/sourcesingcn.js
Theodlz: No! I left it there in case that we need it later on. I'll remove it.

Theodlz:
> @Theodlz Very nice. The handlers should use Guy's syntax, which is my main comment. Also I just want to be sure the source page query should still be fast even with a lot of objects. Otherwise, looks good to me.

Hey @mcoughlin. I'm not sure I understand your concern about the source page query. The query to get associated GCNs is super basic. It looks in the new table added by this PR (which shouldn't have a lot of entries I believe) based on one condition only (by obj_id) which is indexed.

mcoughlin:
@Theodlz Ok good on the query!

static/js/components/SourceTable.jsx
```diff
@@ -1098,6 +1161,19 @@ const SourceTable = ({
     },
   ];
 
+  // if all sources have a status, add a status column
```
guynir42, line 1164: Is this comment line in the right place?

static/js/components/SourcesInGcn.jsx
```diff
@@ -0,0 +1,267 @@
...
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  const findCurrentState = () => {
+    let state = "pending";
+    if (
+      sourcesingcn.filter((s) => s.obj_id === sourceId).length !== 0 &&
+      sourcesingcn?.length > 0
```
guynir42, line 110: Should probably reverse the conditions in this statement.

static/js/components/SourcesInGcn.jsx
```diff
@@ -0,0 +1,267 @@
...
+        sourcesingcn.filter((s) => s.obj_id === sourceId)[0]
+          .confirmed_or_rejected === false
+      ) {
+        state = "rejected";
+      }
+    }
+    return state;
+  };
+
+  const currentState = findCurrentState();
```
guynir42, line 127: Why do we need this to be a function? Can we just calculate the value and assign it?

static/js/components/SourcesInGcn.jsx
```diff
@@ -0,0 +1,267 @@
...
+        </IconButton>
+      ) : null}
+    </MuiDialogTitle>
+  )
+);
+
+const SourcesInGCN = ({
+  dateobs,
+  localizationName,
+  sourceId,
```
guynir42, line 89: I'm not sure what this component does. Does it give me a list of sources associated with the GCN or only checks if this source (given by `sourceId`) is included in the localization? Perhaps a short docstring is in place here. 
Also, `sourceId`->`sourceID` is preferred.

static/js/ducks/source.js
```diff
@@ -277,6 +280,9 @@ export const fetchVizier = (sourceID) =>
 export const fetchPhotoz = (sourceID) =>
   API.POST(`/api/sources/${sourceID}/annotations/datalab`, FETCH_PHOTOZ);
 
+export const fetchAssociatedGCNs = (sourceID) =>
+  API.GET(`/api/source_in_gcns/${sourceID}`, FETCH_ASSOCIATED_GCNS);
```
guynir42, line 284: As discussed above, I don't really like `source_in_gcns`. But if you have good reasons to have this endpoint and the associated logic, a much better name could be `associated_gcns` or something like that.

guynir42 (commented):
Thanks for taking this on, it is a pretty big addition. 

I have some comments below but my main question is do we need the separate `source_in_gcn` handler? What is the reason we want to get "all GCNs associated with a source". Do we expect some sources to have multiple GCNs? I think a more direct question would be "is source confirmed to this GCN" which we can answer with a boolean.

skyportal/app_server.py
```diff
@@ -119,6 +119,8 @@
     WeatherHandler,
     AnalysisWebhookHandler,
     PS1ThumbnailHandler,
+    SourcesInGcnHandler,
+    SourceInGcnsHandler,
```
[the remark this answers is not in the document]
Theodlz: I agree 100%. That's why I said in the PRs description that we may have to change handlers/tables names. I think it makes sense to get your opinion on that, especially as a native English speaker is more likely to find a nicer and neater name for all of this hahaha.

skyportal/handlers/api/sources_in_gcn.py
```diff
@@ -0,0 +1,213 @@
...
+    def post(self, dateobs):
+        data = self.get_json()
+        if 'localizationName' not in data:
+            return self.error("Missing required parameter: localization_name")
+        if 'localizationCumprob' not in data:
+            return self.error("Missing required parameter: localization_cumprob")
+        if 'sourceId' not in data:
+            return self.error("Missing source_id")
+        if 'confirmedOrRejected' not in data:
+            return self.error("Missing confirmed or rejected")
```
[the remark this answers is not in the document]
Theodlz: The issue i have here is that this table keeps track of sources that are confirmed or rejected, but not the sources that are simply 'undefined' (in the localization, obtained with a query on the sources on the fly, those are not stored in the DB as you know). 
So the true or false of this table means that it is confirmed or rejected. In my mind it was 'lighter' in the table to use a boolean to reflect these 2 possible states, rather than a string.
The issue with the second option you give here is that then I need to add 2 columns, one being a boolean for confirmed, and the second for reject. 

Should I just go for a string 'confirmed' or 'rejected' rather than the boolean ?

skyportal/tests/api/test_gcn.py
```diff
@@ -107,15 +107,26 @@ def test_gcn_Fermi(super_admin_token, view_only_token):
...
-    status, data = api('POST', 'gcn_event', data=data, token=super_admin_token)
+    nretries = 0
+    posted = False
+    while nretries < 10 and not posted:
+        status, data = api('POST', 'gcn_event', data=data, token=super_admin_token)
+        if status == 200:
+            posted = True
+        else:
+            nretries += 1
+            time.sleep(3)
```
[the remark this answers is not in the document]
Theodlz: It's a test that is flaky for some reason (test_gcn_IPN). This change was already committed to main btw. I can't tell exactly why the post doesn't work, but its only in the context of the GA, and only for that particular post that gives a 502. Retrying the post here allows the tests to pass at least... It's been flaky on main for a week, with no particular reason. But I would be happy to talk about it to try to find a "cleaner" way to fix it?

skyportal/tests/api/test_gcn.py
```diff
@@ -714,3 +720,231 @@ def test_gcn_summary_observations(
...
+
+    # find gncs associated to source
+    status, data = api(
+        'GET',
+        f"source_in_gcns/{obj_id}",
+        token=upload_data_token,
+    )
+    assert status == 200
+    data = data["data"]
+    assert '2019-08-14T21:10:39' in data['gcns']
```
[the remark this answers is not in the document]
Theodlz: This PR actually adds that to the source page! On the source page, if the source has been confirmed as the one being at the origin of an event, it will display in the InfoList the GCN (or GCNs) associated to the source, as clickable links. It's something users asked me for. That once a source is confirmed as being the one at the origin of a GCN, it is also mentioned on the source page, so they know that if a source already has been confirmed in a GCN when looking at it.
If you think it doesn't make sense, I can simply remove it it's not a problem for me. Maybe @mcoughlin can tell us what he thinks? He knows about that feature.

Now why GCNs and not just one GCN. Well technically in the table I added, a source can be confirmed in different events. So I had to list all of them rather than just one, which might miss a gcn is was confirmed in. If you think that a source can be confirmed for only ONE GCN, then I can add a unique constraint on a `confirmed` source, but not unique on a `rejected` source. Not sure how to do that, but it would be that the source + confirmed being true has to be unique in the DB ? Or at least doing that in the handler.

static/js/components/SourceTable.jsx
```diff
@@ -323,12 +331,26 @@ const SourceTable = ({
 
   const [tableFilterList, setTableFilterList] = useState([]);
   const [filterFormData, setFilterFormData] = useState(null);
+
   const [rowsPerPage, setRowsPerPage] = useState(numPerPage);
   const [queryInProgress, setQueryInProgress] = useState(false);
 
+  const gcnEvent = useSelector((state) => state.gcnEvent);
```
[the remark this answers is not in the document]
Theodlz: Yes, only used on the GCN page.

static/js/ducks/source.js
Theodlz: No good reason! As mentionned in the PR's description I want to change the names, they are definitely not clear enough. Thanks a lot for the suggestion!

Theodlz:
Hi @guynir42. Thanks a lot for the precious review. I'll finish addressing your comments tomorrow, and will start making some of the changes you asked for. Also, some things definitely need to be discussed

skyportal/tests/api/test_gcn.py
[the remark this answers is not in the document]
mcoughlin: I think you can have multiple GCNs in case of a GRB and GW from the same neutron star merger. This gets at the question of aliases that @Theodlz and @leoyang33 are tackling.

skyportal/handlers/api/sources_in_gcn.py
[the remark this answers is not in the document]
guynir42: I think it is fine to have one boolean column called "confirmed". If it is True, it is confirmed, if it is False it is rejected, and if the row doesn't exist then it is neither (in a different scenario you can imagine having a nullable boolean column that does the same). 
The fact that you have a third option should be explained in the docstring for this column (and maybe in other key places) for the developer to understand the subtlety of it... but other than that I think it is fine to have it as a boolean.

skyportal/tests/api/test_gcn.py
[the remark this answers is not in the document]
guynir42: If it works that is great. I would make an effort (at low priority) to figure out why this fails to post. There may be a bug in there that is worth finding.

skyportal/tests/api/test_gcn.py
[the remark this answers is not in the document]
guynir42: Yeah it is true that in principle it can be associated with more than one GCN. We just want to design around thinking it would probably not be a lot of associations. For example, you may have an easier time making something like classifications (where you have space for a few, not a whole table).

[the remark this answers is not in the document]
Theodlz: What do you mean by "something like the classifications"? I'm looking at both frontend and backend of the classifications, and it seems that it is querying and displaying all the classifications associated to a source, not just a few. The only thing that is done for the classifications is that it prevents duplicates in the frontend(for every identical classification, it shows the latest one), but it does show all of them.

skyportal/tests/frontend/test_gcn.py
```diff
@@ -635,3 +635,146 @@ def test_gcn_summary_observations(
...
+    assert status == 200
+    assert data['status'] == 'success'
+
+    driver.get(f'/become_user/{super_admin_user.id}')
+    driver.get('/gcn_events/2019-08-14T21:10:39')
+
+    query_list = driver.wait_for_xpath(
+        '//*[@aria-labelledby="root_queryList-label root_queryList"]', 20
+    )
+    time.sleep(5)
```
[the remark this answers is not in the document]
Theodlz: I noticed that event locally the test was flaky without it. What happens here (and for example also with the favorite source test that is a bit flaky on main) i believe, is that the component rerenders, and sometimes the timing it quite bad. The test finds the xpath, and then when you click its not reachable probably because at the exact moment you tried to click the component rerendered or something similar. Adding a short time.sleep(n) seems to also fix it.

static/js/components/SourcesInGcn.jsx
Theodlz: the name was not clear at all. Will be fixed in the next commit. It is a component to confirm/reject/unconfirm a source from the source table when displayed on a gcn page.

skyportal/handlers/api/sources_in_gcn.py
```diff
@@ -0,0 +1,213 @@
...
+                    first_detected_date=start_date,
+                    last_detected_date=end_date,
+                    localization_dateobs=dateobs,
+                    localization_name=localization_name,
+                    localization_cumprob=localization_cumprob,
+                    page_number=1,
+                    num_per_page=MAX_SOURCES_PER_PAGE,
+                    sourceID=source_id,
+                )
+
```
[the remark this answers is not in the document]
Theodlz: Good idea! In other handlers and models, they use `make_log(<app_name>)` from baselayer to create the logger for a particular app. I'm not sure if I should create a different app for this, or use the same app as the one used in the source handler. What do you think?

guynir42 (commented):
Much better! I think there is still a little bit of work to clean up the validation but this looks in very good shape.

skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -0,0 +1,557 @@
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  localization_name:
+                    type: string
+                    description: The name of the localization of the event
+                  localization_cumprob:
+                    type: string
+                    description: The cumprob of the localization of the event
```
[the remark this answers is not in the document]
Theodlz: Bascially, the reason why I ask for localization_name, localization_cumprob, start_date and end_date to post, is that I want to verify that the source is in the localization, to not end up with sources confirmed in a gcn, when they can't possibly be inside of it. For that, the cumprob is needed. You can see that it is passed to the `get_sources` method.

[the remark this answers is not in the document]
guynir42: But isn't all that information already part of the localization you are adding the source to?

skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -4,11 +4,23 @@
...
 from marshmallow.exceptions import ValidationError
 
 
+class Validator(Schema):
+    source_id = fields.String()
+    dateobs = UTCTZnaiveDateTime(required=False, missing=None)
+    start_date = UTCTZnaiveDateTime(required=False, missing=None)
+    end_date = UTCTZnaiveDateTime(required=False, missing=None)
+    confirmed = fields.Boolean(
+        truthy=['true', 'True', True], falsy=['false', 'False', False]
```
guynir42, line 17: I would add "confirmed" as well

skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -4,11 +4,23 @@
...
-from marshmallow import Schema
+from marshmallow import Schema, fields
 from marshmallow.exceptions import ValidationError
 
 
+class Validator(Schema):
+    source_id = fields.String()
+    dateobs = UTCTZnaiveDateTime(required=False, missing=None)
+    start_date = UTCTZnaiveDateTime(required=False, missing=None)
+    end_date = UTCTZnaiveDateTime(required=False, missing=None)
```
guynir42, line 15: Isn't dateobs required to id the event?

skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -63,7 +75,7 @@ def get(self, dateobs, source_id=None):
         multiple:
           tags:
             - sources_confirmed_in_gcn
-          description: Retrieve all classifications
+          description: Retrieve sources that has been confirmed or rejected in GCN
```
guynir42, line 78:
```suggestion
          description: Retrieve sources that have been confirmed or rejected in a GCN
```

skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -184,20 +205,20 @@ def post(self, dateobs):
...
                     description: The source_id of the source to confirm or reject
-                  confirmed_or_rejected:
+                  confirmed:
                     type: boolean
-                    description: Whether the source is confirmed or rejected
+                    description: Whether the source is confirmed (True) or rejected (False)
                   start_date:
                     type: string
-                    description: The start date (min first detected in photstat) of the source, as an arrow parseable string
+                    description: First detection of source(s) after start_date, as an arrow parseable string
```
guynir42, line 213: I would go with something like "return sources with a first detection after start_date..."

skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -222,18 +243,14 @@ def post(self, dateobs):
...
         if 'localization_name' not in data:
             return self.error("Missing required parameter: localization_name")
         if 'localization_cumprob' not in data:
             return self.error("Missing required parameter: localization_cumprob")
         if 'source_id' not in data:
             return self.error("Missing source_id")
-        if 'confirmed_or_rejected' not in data:
-            return self.error("Missing confirmed_or_rejected")
+        if 'confirmed' not in data:
+            return self.error("Missing confirmed")
```
guynir42, line 253: If you pass `data` directly to the validator it may save you this entire block of code.

skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -430,7 +439,7 @@ def put(self, dateobs, source_id):
     def delete(self, dateobs, source_id):
         """
         ---
-        description: Deletes a confirmed or rejected source in a GCN
+        description: Deletes the confirmed or rejected status of source in a GCN. It's status can be considered as 'undefined'
```
guynir42, line 442:
```suggestion
        description: |
          Deletes the confirmed or rejected status of source in a GCN. 
          Its status can be considered as 'undefined'.
```

guynir42 (approved):
Looks much better! I have a few minor comments and after that it seems good to go.

skyportal/handlers/api/sources_confirmed_in_gcn.py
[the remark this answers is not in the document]
Theodlz: I don't think so. The localization tiles have a probdensity, used to calculate their associate cumprob. Then the `localizationCumprob` basically means: "we consider this source in the `N`th most probable region of the localization", meaning that the `localizationCumprob` >= cumprob of the tiles. This is used in the GCNSelection form of the GCN Event page, which means it is used when querying the sources, galaxies and obs. This is why we also need it here. We couldn't confirm a source without it, as it is a condition to determine if it is in the localization (with a certain "confidence" level) of the event in the first place.

skyportal/handlers/api/sources_confirmed_in_gcn.py
Theodlz: This is in the post method which does not return anything. It's just an argument used to query sources in a localization (we use this time constraint as sources detected way before or way after an event have nothing to do with it).

skyportal/handlers/api/sources_confirmed_in_gcn.py
Theodlz: Done. I decided to use one validation, with added conditional validation (for required / non required values) for the different GET, POST, PATCH, DELETE methods, as they don't required the same parameters.

skyportal/handlers/api/sources_confirmed_in_gcn.py
guynir42: Right, return might be the wrong word but maybe "choose sources with..."

mcoughlin (approved):
Thanks for addressing those issues @Theodlz .

PR #3270, opened 2022-07-27
Bug: Allocation page - List of allocations only showing allocations with an obs plan API

Theodlz:
This PR addresses an issue found by users of the instance of Skyportal I have in production. On the allocation page, the list of allocations only shows the ones with an obs plan API type. This is because in the component to create new default obs plans, a useEffect is fetching allocations with the API type, which end up replacing the list of all allocations in the store.


mcoughlin (approved):
LGTM!

PR #3271, opened 2022-07-27
Issues with Allocations on the Frontend: Full allocation's list replaced by sublists in several components

Theodlz:
This PR continues what has been done in #3267, which only solved the problem caused by one component from the allocations page itself. This adds similar changes but to all pages of SkyPortal that can affect the allocation's list, but in general have a negative effect on each other, causing some components to display incorrect allocations.

Caution: We talked about it with @mcoughlin, so I did my best to fix what I could from what I understood of the instruments, their API, allocations and how they work with Followup requests, Observations and Observation Plans. Now it would make sense to properly run the code and play with it to see if everything still behaves as expect.


guynir42 (approved):
Looks good!

Issue #3257, opened 2022-07-20
GcnEvent Names


mcoughlin:
For now, we track everything by dateobs, which is a convenient ground truth for events. However, events will be much more commonly known by their GRB or GW name. Unfortunately, these names aren't assigned until later and not usually in an easy programmatic way, so we need to:

(i) Add a column to models/gcn.py called display_name or similar.
(ii) Add a new permission called "Manage gcns" or similar
(iii) Add a front-end element where the display name can be set if you have this permission
(iv) Modify the Recent Gcn Events and Gcn Events table to show this name in a new column
(v) Make it possible to filter the GcnEvents table by this name

robertdstein:
I think this directly overlaps with what @leoyang33 was already working on. One key point I would make: I do not think  it is a good idea to bake in the idea that there is a unique name for each event (a la classifications). Instead we should code up something that allows multiple names (GW170817, GRB170817A, AT2017gfo) for a single event.

mcoughlin:
@robertdstein @Theodlz ah that's nice, like aliases or similar? And allow it to be a list?

robertdstein:
Yes exactly. The names can then be used to parse the circulars associated to that name (e.g https://gcn.gsfc.nasa.gov/other/220724A.gcn3)

Theodlz:
Sounds great! Its quite easy to have a column on the gcn event table that is a list of strings.
Unless leo is already working on it, I'll be happy to open a PR.

robertdstein:
@leoyang33 is indeed already working on this, so can we assign this to him instead? I hope it doesn't conflict with any work you already did @Theodlz.

Theodlz:
Not at all! Let's assign him.

Theodlz:
For some reason, I'm not able to assign him. Maybe @mcoughlin can ?

robertdstein:
I think none of the SURF students are in the Skyportal organisation, so we cannot directly assign issues to them. We were doing it exclusively via comments thus far. So maybe we just say "This issue is officially assigned to @leoyang33".

Issue #2528, opened 2022-01-04
GCN event list link


mcoughlin:
Would be nice to link to / display GCNs connected to a given event. I hope there is some kind of API through:
https://heasarc.gsfc.nasa.gov/wsgi-scripts/tach/gcn_v2/tach.wsgi/

robertdstein:
As I said in the meeting, I am happy to do this, but cannot assign myself. I am a member of the Fritz organisation, but maybe different permissions are needed here.

PR #3272, opened 2022-07-28
GCN Events - Add new Names/Aliases from Notices/Circulars

Theodlz:
This PR addresses issue #3257

@leoyang33 you'll find the bases of the handler, model and frontend elements. 
Everything I added here is the bare minimum, it's just to give you the most basic example of the whole `model=>handler=>api=>frontend` workflow example. in general, this is a nice example for all summer students of what implementing a new feature looks like :)

@leoyang33 feel free to add your python code to scrape names and add it in the `post` method of the gcn_aliases handler (I'm not sure if it should be called a name or an alias, I'll let you decide!).


Theodlz:
Hi @leoyang33. As mentioned on Slack I added some code that replaces yours. I'll let you build on top of it. Backend and frontend can be improved, especially the backend. Some fields in the payload of the queries are useless, so you can remove some stuff once you are done with the feature (just so we don't keep unnecessary code).

Theodlz:
Hi @leoyang33, did you get the chance of playing around with TACH and his code to try to parse aliases from the circulars?

leoyang33:
Hi @Theodlz, I added a check for events that may also have a GWYYMMDD name since there are cases where events that are named S YYMMDD and also referenced as GW in the circulars. I'm not sure if there are any other similar cases.

Theodlz:
@leoyang33 what about looking for all the names like `<something>YYMMDD` instead of just GW?

leoyang33:
@Theodlz I added more match cases (GRB, GW, S, IceCube) based on the names for events I found on TACH. Are there any others you think would be necessary?

Theodlz:
> @Theodlz I added more match cases (GRB, GW, S, IceCube) based on the names for events I found on TACH. Are there any others you think would be necessary?

That sounds good!

What about some regex, like: `(\b[^\s\d]+\d{6}\b`, or more generally `(\b[^\s\d]+\d+\b)` to find all of those names with one condition?
(the `\b` are nice but can make us skip some names if they are at the very beginning of a string for example, I'll let you try with and without it)

leoyang33:
Im not entirely sure what you mean by that. Are you saying we should use something like `(\b[^\s\d]+\d+\b)` to broaden what the regex matches with, i.e. any sort of identifying prefix + YYMMDD?

Theodlz:
Yep that's it, any string + number

Theodlz:
Could even be string+number+string to match all cases

leoyang33:
@Theodlz I broadened the regex search criteria to be `String + (day observed) + String`. I still left some constraints in because sometimes the circulars mention other events, and if the constraints were loosened then it would start matching names that do not correspond to the desired event.

Theodlz:
Sounds good!
Did you find, by any chance some events with multiple names to try the code?

To do that you can temporarely (just on your computer) comment the part of the code that verifies if the event exists on SP. That way you can try it with any event. 

It would be nice to be able to properly test it once.

Once that's done, please remove any unnecessary comment in the code and I'll be happy to merge :)

leoyang33:
I tested with various GRB and GW events. It seems that in most cases GW events that are originally named S YYMMDD usually have a GW YYMMDD alias which is picked up by the code, while most GRB events do not have other aliases.

Theodlz:
> I tested with various GRB and GW events. It seems that in most cases GW events that are originally named S YYMMDD usually have a GW YYMMDD alias which is picked up by the code, while most GRB events do not have other aliases.

Sounds great! I'll review your code later today.

mcoughlin:
@leoyang33 Can you please fix the conflicts?

skyportal/handlers/api/gcn_aliases.py
```diff
@@ -0,0 +1,251 @@
...
+            if trigger == dateobs:
+                event_id = event["node"]["id_"]
+
+    if event_id is None:
+        event_ids = []
+        for event in events:
+            if event["node"]["date"] == date and event["node"]["evttype"] in tags:
+                event_ids.append(event["node"]["id_"])
+        if len(event_ids) != 1:
+            # multiple events on the same day, cant figure out the right one. We could look at some circular and notices, event types.... to find which one is the right one.
```
guynir42, line 79: Is there a plan to add some code to make a decision between multiple GCNs in case there are a few with the same date/type?

alembic/versions/809e29e0c0c7_gcn_aliases.py
```diff
@@ -0,0 +1,33 @@
...
+Create Date: 2022-07-28 15:58:41.631666
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '809e29e0c0c7'
+down_revision = 'fccc7e78c7aa'
```
guynir42, line 14: Very likely to have to update the migration hash.

guynir42 (approved):
Looks good! I have minor comments:

leoyang33:
I added a model and display for GCN Circulars to this PR since the code for it is dependent on the aliases code. This should also address issue #2528

alembic/versions/809e29e0c0c7_gcn_aliases.py
mcoughlin: @leoyang33 Did you check that the hash is up to date?

Theodlz: I think this needs additional migrations for the columns you just added @leoyang33

leoyang33: Im not sure how to do migrations

mcoughlin: Remove alembic/versions/809e29e0c0c7_gcn_aliases.py, git push and then commit the artifact migration in its place.

mcoughlin:
@Theodlz https://github.com/skyportal/skyportal/pull/3272/files#diff-1d1d200c4edab5ccb90d35cb43e088be7b531787cb72c300c90e7cf4910bb478R949
What is missing from this test?

Theodlz:
> @Theodlz https://github.com/skyportal/skyportal/pull/3272/files#diff-1d1d200c4edab5ccb90d35cb43e088be7b531787cb72c300c90e7cf4910bb478R949 What is missing from this test?

Leo said he added a new js component for circulars, and I don't think there is anything testing that correct @leoyang33 ?

leoyang33:
@Theodlz That is correct, I can work on making some tests for it

mcoughlin:
@leoyang33 Did these tests get made?

skyportal/handlers/api/gcn_tach.py
```diff
@@ -0,0 +1,333 @@
...
+from astropy.time import Time
+
+from baselayer.log import make_log
+
+
+from sqlalchemy.orm import sessionmaker, scoped_session
+
+Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
+
+log = make_log('api/gcn_aliases')
```
mcoughlin, line 17: Is this the best name for the log?

skyportal/handlers/api/gcn_tach.py
```diff
@@ -0,0 +1,333 @@
...
+
+    aliases = []
+    if response.status_code == 200:
+        data = response.json()
+        if data["data"]["allCirculars"]["totalCount"] > 0:
+            events = data["data"]["allCirculars"]["edges"]
+            aliases.append(events[0]["node"]["evtidCircular"]["event"].replace(" ", ""))
+            new_circulars = {}
+            for event in events:
+                #     aliases.append(event["node"]["evtidCircular"]["event"].replace(" ", ""))
```
mcoughlin, line 189: Maybe remove the comment line?

skyportal/handlers/api/gcn_tach.py
```diff
@@ -0,0 +1,333 @@
...
+            gcn_event.aliases = gcn_aliases
+
+        if gcn_event.circulars is None:
+            gcn_event.circulars = new_gcn_circulars
+        else:
+            gcn_event.circulars = {**gcn_event.circulars, **new_gcn_circulars}
+
+        session.commit()
+
+    push_all(
```
mcoughlin, line 228: This can just be imported from flow. I worry about carrying this dependency on the handler itself.

skyportal/handlers/api/gcn_tach.py
```diff
@@ -0,0 +1,333 @@
...
+                    return self.error(f'No GCN event found for {dateobs}')
+                tach_id = gcn_event.tach_id
+                aliases = gcn_event.aliases
+                circulars = gcn_event.circulars
+
+        except Exception as e:
+            log(f'Error scraping GCN aliases: {e}')
+            return self.error(f'Error: {e}')
+
+        return self.success(
```
mcoughlin, line 331: Maybe best to have the return in the session directly?

static/js/components/EarthquakesPage.jsx
```diff
@@ -9,7 +9,8 @@ const sidebarWidth = 190;
 const EarthquakesPage = () => {
   const ref = useRef(null);
   const theme = useTheme();
-  const initialWidth = window.innerWidth - sidebarWidth - 2 * theme.spacing(2);
+  const initialWidth =
```
mcoughlin, line 12: Thanks for fixing these.

mcoughlin (approved):
Very excited to see this go in. Just a few minor comments.

mcoughlin:
@Theodlz After an upgrade to Ubuntu 22, I have checked that your Dockerfile builds and the binaries inside seem good. I think we should get this in. What do you think?

Theodlz:
@mcoughlin I removed your comment as it was on the wrong PR :) You meant to write it on the Image Analysis one I believe (about the dockerfile).

skyportal/handlers/api/gcn_tach.py
```diff
@@ -0,0 +1,335 @@
...
+        else:
+            gcn_event.circulars = {**gcn_event.circulars, **new_gcn_circulars}
+
+        session.commit()
+
+    flow.push(
+        user_id='*',
+        action_type='skyportal/REFRESH_GCNEVENT',
+        payload={'gcnEvent_dateobs': dateobs},
+    )
```
guynir42, line 233: I'm not sure if this will work. When you send this out to the executor, it will be detached from the `session` and your `user` object may be invalid. You may also run into trouble where the session you open and close inside the executor somehow ends up closing the session outside the executor. It may be safer to pass a session object, but you should check what we do in other instances. For sure passing a `user_id` is safer than a `user` object.

skyportal/handlers/api/gcn_tach.py
```diff
@@ -0,0 +1,335 @@
...
+                              description: The id of the GcnEvent
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+        try:
+            arrow.get(dateobs).datetime
+        except Exception:
+            log(f'Invalid dateobs: {dateobs}')
```
guynir42, line 274: Don't need this log

skyportal/handlers/api/gcn_tach.py
```diff
@@ -0,0 +1,335 @@
...
+
+                IOLoop.current().run_in_executor(
+                    None,
+                    lambda: post_aliases(dateobs, tach_id, session.user_or_token),
+                )
+
+                return self.success(data={'id': gcn_event_id})
+
+        except Exception as e:
+            log(f'Error scraping GCN aliases: {e}')
```
guynir42, line 305: This one either. 
BTW if these logging commands were added just for debugging, I suggest using a simple `print` which is easier to catch and get rid of before moving to production. 
In general we want to reduce the amount of logging we are doing to only what is really necessary.

skyportal/models/gcn.py
```diff
@@ -187,6 +187,28 @@ class GcnEvent(Base):
         doc="Comments posted about this GCN event.",
     )
 
+    aliases = sa.Column(
+        sa.ARRAY(sa.String),
+        nullable=False,
+        server_default="{}",
```
guynir42, line 193: I'm not sure if this is the best way to define the default. Why not `default=[]`?

skyportal/models/gcn.py
```diff
@@ -187,6 +187,28 @@ class GcnEvent(Base):
...
+        sa.String,
+        nullable=True,
+        doc="TACH id associated with a GCN event",
+    )
+
+    circulars = deferred(
+        sa.Column(
+            JSONB,
+            nullable=False,
+            server_default="{}",
```
guynir42, line 207: and here maybe just `default={}`?

skyportal/tests/api/test_gcn.py
```diff
@@ -1057,3 +1057,68 @@ def test_gcn_Swift(super_admin_token, view_only_token):
...
+
+    data = data['data']
+    assert len(data['aliases']) == 0
+
+    status, data = api(
+        'POST', 'gcn_event/2018-01-16T00:36:53/tach', token=view_only_token
+    )
+    assert status == 401
+
+    time.sleep(15)
```
guynir42, line 1093: why do we need to sleep here?

guynir42 (commented):
I have some minor comments below:

skyportal/handlers/api/gcn_tach.py
Theodlz: You're right. For things running in the executor, I'll change to using user_ids whenever possible.

skyportal/models/gcn.py
Theodlz: Good point. I made that change. I also regenerated the migration script, and I'm wondering if I need to add server_defaults or not. It's not super clear to me what that does (does it set the value of new columns of all already existing rows to the server_default when migrating the DB?).

guynir42: I'm not sure either...

Theodlz:
Hi @guynir42 
As you can see from the commit history (ouch) I had some trouble with the migrations. Defaults in the models and server_defaults don't seem to interact so well in certain cases. I guess that this migration updating an existing table with the addition of JSON non null column is one of those cases. Turns out one needs to add a server_default instead of a default in the models as well, and it needs to be "{}".

It works just fine now, and all tests are green. Let me know what you think.

skyportal/handlers/api/gcn_tach.py
```diff
@@ -211,13 +212,13 @@ def post_aliases(dateobs, tach_id, user):
             return
 
         if gcn_event.aliases is None:
-            gcn_event.aliases = new_gcn_aliases
+            gcn_event.aliases = list(set(new_gcn_aliases))
         else:
             gcn_aliases = [alias.upper() for alias in gcn_event.aliases]
             for new_gcn_alias in new_gcn_aliases:
                 if new_gcn_alias.upper() not in gcn_aliases:
                     gcn_aliases.append(new_gcn_alias)
```
guynir42, line 220: You should probably append the `new_gcn_alias` with an `.upper()` call.

guynir42 (approved):
Yeah, if the migration works with the `server_default` then that is fine.

skyportal/handlers/api/gcn_tach.py
guynir42: Make sure that in all cases what you end up appending into the final list is all uppercase.

Theodlz: Hi @guynir42, I made the changes you required, and other modifications as well, as I noticed that the behavior of some if statements to check if we already have something in the DB for the circulars and aliases was incorrect. It should work just fine now. It's a minor change and you already approved the changes, so I'm not tagging you for a new review. Just let me know if you want to check the changes.

skyportal/handlers/api/gcn_tach.py
```diff
@@ -191,47 +191,58 @@ def get_tach_event_aliases(id, gcn_event):
...
+
+            if not gcn_event.circulars:
+                gcn_event.circulars = new_gcn_circulars
+            else:
+                gcn_event.circulars = {**gcn_event.circulars, **new_gcn_circulars}
+
+            if len(new_gcn_aliases) > 0:
+                if (
+                    not gcn_event.aliases
+                ):  # thanks to the implicit booleanness of the empty list in python, this checks if its None or an empty list
```
guynir42, line 227:
```suggestion
                if not gcn_event.aliases:  # empty list or None
```

guynir42 (approved):
Looks good. One minor comment below.

Issue #3249, opened 2022-07-19
Can't assign targets to run on same day


robertdstein:
In the same vein as #3073, you presently cannot assign targets to a run that has already happened, but the run is assumed to have finished at 00:01 on the assigned run date. I would suggest we mimic #3073, and set the default behaviour for assigning a source to be "all observing runs except those which began >24 hours ago".

virajkaram:
This is pretty frustrating for several users. Would really appreciate if this was fixed soon.

mcoughlin:
@virajkaram Please submit a PR.

robertdstein:
I think this would require knowledge of javascript/react/frontend for a fix, right? Not sure about @virajkaram but I definitely am lacking the expertise here.

PR #3275, opened 2022-07-29
Make it possible to submit day-of runs as "upcoming"

mcoughlin:
Closes https://github.com/skyportal/skyportal/issues/3249

### DIFF
--- a/static/js/components/AssignmentForm.jsx
+++ b/static/js/components/AssignmentForm.jsx
@@ -84,8 +84,10 @@ const AssignmentForm = ({ obj_id, observingRunList }) => {
   }));
   const classes = useStyles();
 
+  // the use of integer dates leads to some upcoming runs being
+  // left out depending on the timezone
   const upcomingObservingRuns = observingRunList.filter((run) =>
-    dayjs().isBefore(dayjs(run.run_end_utc))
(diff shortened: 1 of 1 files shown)


static/js/components/AssignmentForm.jsx
```diff
@@ -84,8 +84,10 @@ const AssignmentForm = ({ obj_id, observingRunList }) => {
   }));
   const classes = useStyles();
 
+  // the use of integer dates leads to some upcoming runs being
+  // left out depending on the timezone
   const upcomingObservingRuns = observingRunList.filter((run) =>
-    dayjs().isBefore(dayjs(run.run_end_utc))
+    dayjs().isBefore(dayjs(run.run_end_utc).add(2, "day"))
```
guynir42, line 90: Are you sure this doesn't accidentally add runs that are too far in the future?

guynir42 (commented):
I have one thing I'm not sure about.

static/js/components/AssignmentForm.jsx
mcoughlin: Nope because of how the comparison is done. I checked and you get the day 'before' with this.

mcoughlin:
@guynir42 you can merge it if you are happy.

guynir42 (approved):

PR #3276, opened 2022-07-29
GCN Server - Upgrade to GCN Kafka

Theodlz:
This PR upgrades the gcnserver (microservice to ingest gcn events) to use the newer gcn-kafka instead of pygcn.


mcoughlin (approved):
Thanks @Theodlz. Clearly a step forward for us.

PR #3277, opened 2022-07-30
Add react-transition-group to package.json.

mcoughlin:
This PR adds react-transition-group which was leading to some Docker build issues like https://github.com/mcoughlin/skyportal/runs/7592166725?check_suite_focus=true


profjsb (approved):
LGTM. We should change the pre-commit hooks to disallow ^ carrots in package.json.

PR #3278, opened 2022-07-30
Fix issue with plotting widths

mcoughlin:
This PR fixes an issue seem on fritz where the plot height was not reported properly.

### DIFF
--- a/skyportal/handlers/api/internal/plot.py
+++ b/skyportal/handlers/api/internal/plot.py
@@ -33,9 +33,9 @@ class PlotPhotometryHandler(BaseHandler):
         with self.Session() as session:
             try:
                 json = plot.photometry_plot(
-                    obj_id,
-                    self.current_user.id,
-                    session,
+                    obj_id=obj_id,
+                    user_id=self.current_user.id,
+                    session=session,
                     width=int(width),
                     device=device,
                 )
@@ -60,9 +60,9 @@ class PlotSpectroscopyHandler(BaseHandler):
         with self.Session() as session:
             try:
                 json = plot.spectroscopy_plot(
-                    obj_id,
-                    session,
-                    spec_id,
+                    obj_id=obj_id,
+                    session=session,
+                    spec_id=spec_id,
                     width=int(width),
                     device=device,
                     smoothing=smoothing,
--- a/skyportal/handlers/api/internal/plot.py
+++ b/skyportal/handlers/api/internal/plot.py
@@ -30,14 +30,19 @@ class PlotPhotometryHandler(BaseHandler):
         # Just return browser by default if not one of accepted types
         if device not in device_types:
             device = "browser"
-        json = plot.photometry_plot(
-            obj_id,
-            self.current_user,
-            width=int(width),
-            device=device,
-        )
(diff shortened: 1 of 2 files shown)


profjsb:
Might check that this fixes the plot sizing on mobile (which currently spills past the screen). 

Also, you might update the title of the PR since this also appears to modify the way data is grabbed in a session.

profjsb:
Might check that this fixes the plot sizing on mobile (which currently spills past the screen). 

Also, you might update the title of the PR since this also appears to modify the way data is grabbed in a session.

mcoughlin:
@profjsb yeah it takes care of the move to Guys new syntax. And it's really fixing a problem reported last night about the spectrum plot not showing up, which was that the width / height were not automatically set

profjsb:
Can you check that it fixes the mobile view? Might need other small tweaks for that.

mcoughlin:
@profjsb I am actually a bit unclear on how that works, can you summarize? :)

mcoughlin:
@profjsb https://github.com/react-grid-layout/react-grid-layout/issues/1724 this is the core of that issue.

skyportal/handlers/api/internal/plot.py
```diff
@@ -30,14 +30,19 @@ def get(self, obj_id):
...
-        self.verify_and_commit()
-        self.success(data={'bokehJSON': json, 'url': self.request.uri})
+        with self.Session() as session:
+            try:
+                json = plot.photometry_plot(
+                    obj_id,
+                    self.current_user.id,
+                    session,
+                    width=int(width),
+                    device=device,
```
guynir42, line 40: Both here and in the spectroscopy plot: 
I would explicitly name each parameter I'm sending into the functions. 
There are starting to be a lot of them, and the order is hard to guess.

skyportal/plot.py
```diff
@@ -2041,11 +2061,11 @@ def make_spectrum_layout(obj, spectra, user, device, width, smoothing, smooth_nu
         normfac = np.nanmedian(np.abs(s.fluxes))
         normfac = normfac if normfac != 0.0 else 1e-20
         altdata = json.dumps(s.altdata) if s.altdata is not None else ""
-        annotations = (
-            AnnotationOnSpectrum.query_records_accessible_by(user)
-            .filter(AnnotationOnSpectrum.spectrum_id == s.id)
-            .all()
-        )
+        annotations = session.scalars(
+            AnnotationOnSpectrum.select(session.user_or_token).where(
```
guynir42, line 2065: This is nice. You should be aware that it only works with a handler's `self.Session()` and will not work with a `DBSession()`. I don't know if that is going to be a problem or not. Probably not.

guynir42 (approved):
Looks good!

skyportal/handlers/api/internal/plot.py
mcoughlin: Good idea!

skyportal/plot.py
mcoughlin: Makes sense, for now we pass a handler session (which should hopefully always be the case)

PR #3279, opened 2022-07-31
Simulate efficiencies for skymaps without distance information

mcoughlin:
This PR enables one to simulate efficiencies for skymaps without distance information.

### DIFF
--- a/skyportal/handlers/api/observation_plan.py
+++ b/skyportal/handlers/api/observation_plan.py
@@ -1620,7 +1620,7 @@ def observation_simsurvey(
                 np.max([1, (distmean - 5 * diststd)]) * u.Mpc
             )
             distance_upper = astropy.coordinates.Distance(
-                (distmean + 5 * diststd) * u.Mpc
+                np.max([2, (distmean + 5 * diststd)]) * u.Mpc
             )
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/observation_plan.py
```diff
@@ -1603,21 +1603,32 @@ def observation_simsurvey(
...
+
+            distmean, diststd = parameters_to_marginal_moments(
+                map_struct['prob'], map_struct['distmu'], map_struct['distsigma']
+            )
+
+            distance_lower = astropy.coordinates.Distance(
+                np.max([1, (distmean - 5 * diststd)]) * u.Mpc
+            )
+            distance_upper = astropy.coordinates.Distance(
+                (distmean + 5 * diststd) * u.Mpc
```
guynir42, line 1623: I know it probably never happens, but in principle you could get a `distance_upper` that is less than 1 as well...

guynir42 (approved):

skyportal/handlers/api/observation_plan.py
mcoughlin: Adding a max here too :)

PR #3302, opened 2022-08-01
update webhook-handler to use new SQLA 2.0 methods

guynir42:


skyportal/handlers/api/webhook.py
```diff
@@ -70,63 +65,67 @@ def post(self, analysis_resource_type, token):
...
-                session.close()
-                return self.error("Token has expired", status=400)
-
-            # lock the analysis associated with this token and commit immediately to avoid race conditions,
-            # so that the results are not written more than once
-            analysis.status = 'completed'
-            analysis.last_activity = datetime.datetime.utcnow()
-            analysis.duration = (analysis.last_activity - last_active).total_seconds()
+            except Exception as e:
+                log(f'Troubling accessing Analysis with token {token} {e}.')
```
mcoughlin, line 108: Is this the most likely error and result for this exception? Aren't we checking for tokens above?

mcoughlin (approved):
Just a question on the error returned.

skyportal/handlers/api/webhook.py
guynir42: If I try to interpret what the original author of this error message meant, I would probably guess the idea was "here's an error, here's the token (so you'd have some context) and here's the error message itself. 
Perhaps a better way to make that clearer would be `Trouble posting Analysis results. Token: {token}. Error: {e}`.

mcoughlin: @guynir42 agreed, that seems better to me.

guynir42: I am wondering if we should be returning more info in the `self.error`. Could it be that returning the token/error message reveals internals that we don't necessarily want to expose?

mcoughlin: Presumably it is the token that the user sent, so seems useful to spit it back.

guynir42: how about now?

mcoughlin: Ok by me! Good to merge on green!

PR #3303, opened 2022-08-01
New PhotStat index

guynir42:
Replace the PhotStat index on `mean_color` with a new index using GIN (generalized inverted index). 
This solves the issue that some objects with lots of filters with long names cannot be indexed on `mean_color` because their JSONB tag values are too long.


skyportal/models/phot_stat.py
```diff
@@ -566,42 +574,71 @@ def full_update(self, phot_list):
...
+            mjds.append(phot['mjd'])
+
+            if 'mag' in phot:
+                mags.append(phot['mag'])
+            elif phot['flux'] is not None and phot['flux'] > 0:
+                mags.append(-2.5 * np.log10(phot['flux']) + PHOT_ZP)
+            else:
+                mags.append(np.nan)
+
+            is_detected = (
```
mcoughlin, line 611: Can't the presence of magnitude also indicate detection?

mcoughlin (commented):
@guynir42 only thing I wanted to check on is whether need to also check for magnitudes when checking for detections

skyportal/models/phot_stat.py
guynir42: I think you can have a 2 sigma detection with magnitude that doesn't count as a "detection".

mcoughlin: @guynir42 reasonable! Ok so to me should be ok

PR #3307, opened 2022-08-03
[Analysis Epic] Individual analysis page

profjsb:
Add a linkable page for individual analyses on a given object. Accessed via `/source/source_name/analysis/analysis_id`.

Kapture 2022-08-03 at 19 40 15

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -90,6 +90,8 @@ app:
       component: About
     - path: "/run/:id"
       component: RunSummary
+    - path: "/source/:obj_id/analysis/:analysis_id"
+      component: SourceAnalysisPage
     - path: "/manage_data/:id"
       component: ManageDataForm
     - path: "/filter/:fid"
--- a/skyportal/handlers/api/analysis.py
+++ b/skyportal/handlers/api/analysis.py
@@ -1013,6 +1013,9 @@ class AnalysisHandler(BaseHandler):
                     stmt = ObjAnalysis.select(self.current_user).where(
                         ObjAnalysis.id == analysis_id
                     )
+                    if obj_id:
+                        stmt = stmt.where(ObjAnalysis.obj_id.contains(obj_id.strip()))
+
                     analysis = session.scalars(stmt).first()
                     if analysis is None:
                         return self.error('Cannot access this Analysis.', status=403)
@@ -1028,6 +1031,7 @@ class AnalysisHandler(BaseHandler):
                     analysis_dict[
                         "analysis_service_description"
                     ] = analysis_service.description
+                    analysis_dict["num_plots"] = analysis.number_of_analysis_plots
 
                     if include_filename:
                         analysis_dict["filename"] = analysis._full_name
--- a/skyportal/tests/frontend/sources_and_followup_etc/test_sources.py
+++ b/skyportal/tests/frontend/sources_and_followup_etc/test_sources.py
@@ -1,6 +1,9 @@
 import os
 from os.path import join as pjoin
 import uuid
+import json
+import time
+
 from io import BytesIO
 import pytest
 from selenium.webdriver.common.action_chains import ActionChains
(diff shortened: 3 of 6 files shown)


guynir42 (commented):
The code seems fine. I am still unable to send an analysis successfully to see the results page.

mcoughlin:
@guynir42 i always run the service in another terminal explicitly.

mcoughlin (approved):
@profjsb looks great to me!

guynir42:
@mcoughlin That makes a lot of sense. Is there a documentation somewhere describing how to do that, in testing and in production where it should be managed by supervisord or something like that?

mcoughlin:
@guynir42 i think for now we don't want the demo sn fitter running in production as it doesn't run the full MCMC load. When we get NMMA finished and merged we definitely will though.

I think the docs are ok, although for this particular case it is just running python on app.py in the sn service folder

guynir42:
Where in the docs can I find this?

mcoughlin:
@guynir42 oh that bad is just knowledge from trying it out, but I don't think we need to highlight how to run that service for folks.

profjsb:
@guynir42 The demo fitting microservice should just work after `make load_demo_data` -- it also works in testing. You should have in your `config.yaml`:

```yaml
analysis_services:
  analysis_folder: analysis_data
  sn_analysis_service:
    port: 6801
```

(it's in `config.yaml.defaults`).

guynir42:
OK I managed to get it running as @mcoughlin suggested. It did not automatically start the processes when I used `load_demo_data`.

guynir42:
I can now see the analysis results page which is nice. I can't scroll down to see the entire PDF:
Screenshot from 2022-08-03 14-40-28

guynir42:
Also I have a few suggestions on the display of parameters:
- The `param_names` and `parameters` can just be merged into dictionary with name: value (like for `errors`). 
- The data mask can be converted to an array of `int` to be easier to read. 
- Maybe display a chi2/dof in addition to the two separate values?

guynir42:
On the source page I have two issues that may belong in a separate PR: 
1) The analysis results do not show up when done, only after I refresh the page.
2) The column called "Analysis service" shows number 1 for the demo analysis (I assume it's the ID). Should display the name of the analysis service instead.

profjsb:
> Also I have a few suggestions on the display of parameters:
> 
>     * The `param_names` and `parameters` can just be merged into dictionary with name: value (like for `errors`).
> 
>     * The data mask can be converted to an array of `int` to be easier to read.
> 
>     * Maybe display a chi2/dof in addition to the two separate values?

All of these are not the responsibility of the frontend (or even SkyPortal) but instead are the free-form results from the analysis service. In this case we're just using the results that are generated by `sncosmo`. I think it's beyond the scope for us to try to reformat and rename in a way that's generic to all analyses possible.

profjsb:
> On the source page I have two issues that may belong in a separate PR:
> 
>     1. The analysis results do not show up when done, only after I refresh the page.
> 

Yes, I noted this shortcoming in the original PR for that list/form. Should be changed elsewhere.

>     2. The column called "Analysis service" shows number 1 for the demo analysis (I assume it's the ID). Should display the name of the analysis service instead.

The name and description are given with a mouseover. We show id numbers elsewhere in the app (like in the Assignments, we should Run Id).

profjsb:
> I can now see the analysis results page which is nice. I can't scroll down to see the entire PDF: Screenshot from 2022-08-03 14-40-28

Will need some help on this -- it works on my browser.

guynir42:
I still thing the table should show the name of the analysis. I don't think the ID means much to the user and it is not a big effort on our end to show the name.

mcoughlin:
@profjsb @guynir42 What do we need to get this in?

guynir42 (approved):
Looks good. I will look into the PDFs not scrolling, and we can deal with that in a separate PR.

mcoughlin:
@profjsb Can we merge?

PR #3313, opened 2022-08-07
Update follow-up request handler to use new syntax

mcoughlin:
This PR updates the follow-up request handler to use the new syntax.

### DIFF
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -189,7 +189,6 @@ class AssignmentHandler(BaseHandler):
             if assignment_id is not None:
                 out_json = out_json[0]
 
-            print(out_json)
             return self.success(data=out_json)
 
     @permissions(['Upload data'])
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -96,6 +96,7 @@ def post_assignment(data, user_id, session):
     assignment = ClassicalAssignment(**data)
 
     assignment.requester_id = user.id
+    assignment.last_modified_by_id = user.id
     session.add(assignment)
     session.commit()
 
@@ -171,7 +172,7 @@ class AssignmentHandler(BaseHandler):
                     joinedload(ClassicalAssignment.obj),
                 )
 
-            assignments = session.scalars(assignments).all()
+            assignments = session.scalars(assignments).unique().all()
 
             if len(assignments) == 0 and assignment_id is not None:
                 return self.error("Could not retrieve assignment.")
@@ -188,6 +189,7 @@ class AssignmentHandler(BaseHandler):
             if assignment_id is not None:
                 out_json = out_json[0]
 
+            print(out_json)
             return self.success(data=out_json)
 
     @permissions(['Upload data'])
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/followup_request.py
```diff
@@ -76,25 +77,26 @@ def post_assignment(data, user_id, session):
 
     run_id = assignment.run_id
     data['priority'] = assignment.priority.name
-    run = ObservingRun.get_if_accessible_by(run_id, user, raise_if_none=False)
+    run = session.scalars(
+        ObservingRun.select(user).where(ObservingRun.id == run_id)
```
guynir42, line 81: And the `user` comes from an old style query...

guynir42: Maybe safer to use `session.user_or_token` unless you are specifically allowing people to post assignments with other user's permissions.

skyportal/handlers/api/followup_request.py
```diff
@@ -151,42 +153,44 @@ def get(self, assignment_id=None):
...
+                    joinedload(ClassicalAssignment.obj),
+                )
 
-        assignments = assignments.all()
+            assignments = session.scalars(assignments).unique().all()
 
-        if len(assignments) == 0 and assignment_id is not None:
-            return self.error("Could not retrieve assignment.")
+            if len(assignments) == 0 and assignment_id is not None:
+                return self.error("Could not retrieve assignment.")
```
guynir42, line 178:
```suggestion
                return self.error("Could not retrieve assignment with ID {assignment_id}.")
```

skyportal/handlers/api/followup_request.py
```diff
@@ -151,42 +153,44 @@ def get(self, assignment_id=None):
...
+                json_obj['requester'] = assignment.requester
 
-        if assignment_id is not None:
-            out_json = out_json[0]
+            if assignment_id is not None:
+                out_json = out_json[0]
 
-        self.verify_and_commit()
-        return self.success(data=out_json)
+            print(out_json)
```
guynir42, line 192:
```suggestion
```

skyportal/handlers/api/followup_request.py
```diff
@@ -218,7 +222,7 @@ def post(self):
 
         data = self.get_json()
 
-        with DBSession() as session:
+        with self.Session() as session:
             try:
                 assignment_id = post_assignment(
```
guynir42, line 227: I'm actually not really sure why we need `post_assignment`. It seems somewhat long but it could live happily inside the POST function (in my opinion). 
In any case, there seems no reason to pass the `user_id`, you can just use the `session.user_or_token`.

guynir42 (approved):
Minor comments below. 
I would consider if we really need `post_assignment` as a separate function and not just put that code into the POST handler.

skyportal/handlers/api/followup_request.py
guynir42: Also if you just put that code in the handler you can `self.error` what ever you want instead of catching all these types of exceptions.

skyportal/handlers/api/followup_request.py
mcoughlin: agreed!

skyportal/handlers/api/followup_request.py
mcoughlin: Yeah the issue is the observation_plan handler also calls that function, so we can remove the user stuff but it does need to remain separate.

PR #3315, opened 2022-08-07
Update shift handler to use new syntax

mcoughlin:
This PR updates the shift handler to use new syntax.


skyportal/handlers/api/shift.py
```diff
@@ -141,86 +140,95 @@ def get(self, group_id=None):
...
-                    ],
+
+        with self.Session() as session:
+
+            if group_id is not None:
+                queried_shifts = (
+                    session.scalars(
+                        Shift.select(
+                            session.user_or_token,
+                            mode="read",
```
guynir42, line 151: The default mode is `read`.

skyportal/handlers/api/shift.py
```diff
@@ -141,86 +140,95 @@ def get(self, group_id=None):
...
-                    options=[
-                        joinedload(Shift.group).joinedload(Group.group_users),
-                        joinedload(Shift.shift_users),
-                    ],
+            else:
+                queried_shifts = (
+                    session.scalars(
+                        Shift.select(
+                            session.user_or_token,
+                            mode="read",
```
guynir42, line 168: Here too we don't need `read`.

skyportal/handlers/api/shift.py
```diff
@@ -141,86 +140,95 @@ def get(self, group_id=None):
...
+                                k: v
+                                for k, v in c.to_dict().items()
+                                if k != "attachment_bytes"
+                            },
+                            "author": {
+                                **c.author.to_dict(),
+                                "gravatar_url": c.author.gravatar_url,
+                            },
+                        }
+                        for c in comments
```
guynir42, line 224: I know this is just the way it was, but I feel it would be more readable as a simple loop.

skyportal/handlers/api/shift.py
```diff
@@ -347,57 +354,60 @@ def post(self, shift_id, *ignored_args):
...
-            return self.error(
-                f"User {user_id} is already a member of shift {shift_id}."
-            )
-        if shift.required_users_number:
-            if len(shift.shift_users) >= shift.required_users_number:
+        with self.Session() as session:
+            shift = session.scalars(
+                Shift.select(
+                    session.user_or_token,
+                    mode='read',
```
guynir42, line 361:
```suggestion
```

skyportal/handlers/api/shift.py
```diff
@@ -448,63 +458,66 @@ def patch(self, shift_id, user_id):
...
-        except (ValueError, TypeError):
-            return self.error(
-                "Invalid needs_replacement parameter: unable to parse to boolean"
-            )
-        shiftuser.needs_replacement = needs_replacement
+            admin = data.get("admin", shiftuser.admin)
+            if not isinstance(admin, bool):
+                return self.error(
+                    "Invalid (non-boolean) value provided for parameter `admin`"
+                )
```
guynir42, line 477: Much better!

skyportal/handlers/api/shift.py
```diff
@@ -448,63 +458,66 @@ def patch(self, shift_id, user_id):
...
-                    shift_id,
-                    self.current_user,
-                    options=[
-                        joinedload(Shift.group).joinedload(Group.group_users),
-                    ],
-                    raise_if_none=True,
+                needs_replacement = bool(needs_replacement)
+            except (ValueError, TypeError):
+                return self.error(
+                    "Invalid needs_replacement parameter: unable to parse to boolean"
```
guynir42, line 485: This should also be a test if boolean, not a cast into boolean.

skyportal/handlers/api/shift.py
```diff
@@ -448,63 +458,66 @@ def patch(self, shift_id, user_id):
...
+                    ).where(Shift.id == shift_id)
+                ).first()
+                if shift is None:
+                    return self.error('Could not find shift.')
+                for group_user in shift.group.group_users:
+                    if group_user.user_id != user_id:
+                        session.add(
+                            UserNotification(
+                                user=group_user.user,
+                                text=f"*@{shiftuser.user.username}* needs a replacement for shift: {shift.name} from {shift.start_date} to {shift.end_date}.",
```
guynir42, line 508: Maybe break the string into two lines?

skyportal/handlers/api/shift.py
```diff
@@ -538,24 +551,24 @@ def delete(self, shift_id, user_id):
...
-        DBSession().delete(su)
-        self.verify_and_commit()
-        self.push_all(
-            action='skyportal/REFRESH_SHIFTS', payload={'shift_id': int(shift_id)}
-        )
-        return self.success()
+            session.delete(su)
+            session.commit()
+            self.push_all(
+                action='skyportal/REFRESH_SHIFTS', payload={'shift_id': int(shift_id)}
```
guynir42, line 569: Maybe worth casting in a try catch at the beginning of the function?

skyportal/handlers/api/shift.py
```diff
@@ -623,89 +636,99 @@ def get(self, shift_id=None):
...
-                    ],
+        with self.Session() as session:
+
+            report = {}
+            if start_date and end_date:
+                s = (
+                    session.scalars(
+                        Shift.select(
+                            session.user_or_token,
+                            mode="read",
```
guynir42, line 647:
```suggestion
```

guynir42: Can just get rid of all these `mode="read"` in general.

guynir42 (approved):
I have minor comments, but overall looks good.

skyportal/handlers/api/shift.py
mcoughlin: Yup will do.

skyportal/handlers/api/shift.py
mcoughlin: good idea.

skyportal/handlers/api/shift.py
mcoughlin: ok

skyportal/handlers/api/shift.py
mcoughlin: good idea.

skyportal/handlers/api/shift.py
mcoughlin: I think if we really care, we can do the various instance of these in another PR.

skyportal/handlers/api/shift.py
mcoughlin: Yes.

skyportal/handlers/api/shift.py
mcoughlin: Yes.

skyportal/handlers/api/shift.py
mcoughlin: Yes.

skyportal/handlers/api/shift.py
```diff
@@ -478,12 +475,14 @@ def patch(self, shift_id, user_id):
...
-                    "Invalid needs_replacement parameter: unable to parse to boolean"
-                )
+            # test if bool already
+            if type(needs_replacement) != bool:
+                try:
+                    needs_replacement = bool(needs_replacement)
+                except (ValueError, TypeError):
+                    return self.error(
+                        "Invalid needs_replacement parameter: unable to parse to boolean"
+                    )
```
guynir42, line 485: If it is a boolean you don't need any of this try/except.

guynir42 (approved):

skyportal/handlers/api/shift.py
mcoughlin: good point

PR #3319, opened 2022-08-08
Update spectrum handler to use new syntax

mcoughlin:
This PR updates the spectrum handler to the new syntax.

### DIFF
--- a/skyportal/handlers/api/spectrum.py
+++ b/skyportal/handlers/api/spectrum.py
@@ -7,7 +7,7 @@ import numpy as np
 import pandas as pd
 import sncosmo
 from sqlalchemy.orm import joinedload
-from sqlalchemy import or_, Column
+from sqlalchemy import or_
 
 from marshmallow.exceptions import ValidationError
 from baselayer.app.access import permissions, auth_or_token
@@ -19,7 +19,6 @@ from baselayer.app.custom_exceptions import AccessError
 from .photometry import add_external_photometry
 from ..base import BaseHandler
 from ...models import (
-    DBSession,
     FollowupRequest,
     Group,
     CommentOnSpectrum,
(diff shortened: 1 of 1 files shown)


skyportal/handlers/api/spectrum.py
```diff
@@ -44,62 +43,67 @@
...
+    empty list.
+    """
+    if isinstance(str_list, str):
+        if len(str_list) == 0:
+            return []
+        elif "," in str_list:
+            return [c.strip() for c in str_list.split(",")]
         else:
-            raise TypeError('Must input a string!')
+            return [str_list.strip()]
```
guynir42, line 101: I think you get the same result by removing the last `else` and replacing the `elif` with an `else`.

skyportal/handlers/api/spectrum.py
```diff
@@ -411,59 +415,62 @@ def get(self, spectrum_id=None):
...
+                ).first()
+                if spectrum is None:
+                    return self.error(
+                        f'Could not access spectrum {spectrum_id}.', status=403
+                    )
+                comments = (
+                    session.scalars(
+                        CommentOnSpectrum.select(
+                            session.user_or_token,
+                            options=[joinedload(CommentOnSpectrum.groups)],
```
guynir42, line 432: Do you have any idea why we join the groups here?

skyportal/handlers/api/spectrum.py
```diff
@@ -1149,101 +1158,100 @@ def get(self, obj_id):
...
+                            "author": {
+                                **c.author.to_dict(),
+                                "gravatar_url": c.author.gravatar_url,
+                            },
+                        }
+                        for c in comments
+                    ),
+                    key=lambda x: x["created_at"],
+                    reverse=True,
                 )
```
guynir42, line 1203: As I mentioned in another PR, I'd prefer this as a loop.

guynir42 (approved):

skyportal/handlers/api/spectrum.py
mcoughlin: I had to stare at this comment for a while to try to figure out why it was the case ;)

skyportal/handlers/api/spectrum.py
mcoughlin: Nope.

skyportal/handlers/api/spectrum.py
mcoughlin: I think this is another PR if we go that route.

PR #3328, opened 2022-08-09
Spectroscopy plot - Height of the plot is insufficient, too much scrolling on the Source Page

Theodlz:
On the source page, to see the options below the spectroscopy plot, you have to scroll *quite a lot*. Because of that, the UX is impacted, as you can't see both the options and how they affect the plot without having to scroll up and down every time.


mcoughlin (approved):
Looks a lot better. Thanks @Theodlz!

PR #3329, opened 2022-08-09
Add basic filtering for GcnEvents

mcoughlin:
This PR adds basic filtering for GcnEvents.


guynir42 (approved):
and one more handler pushed to SQLA 2.0. Good!

Issue #2657, opened 2022-02-10
Mac OS Monterey uses port 5000 for Airplay


robertdstein:
Skyportal uses port 5000 by default, but on the new Mac OS Monterey, this port is be default used for Airplay.

Some background: https://anandtripathi5.medium.com/port-5000-already-in-use-macos-monterey-issue-d86b02edd36c 

You can manually disable Airplay receiver, in which case the port is freed up, but a better solution would be to use a different port. 5700 was a popular choice.

leonitousconforti:
I couldn't find anywhere that still hardcodes 5000, additionally there is now a note in the troubleshooting section of the docs :)

PR #3330, opened 2022-08-09
Add troubleshooting line about Macs and Airplay Receiver

bfhealy:
Update troubleshooting documentation to reflect a problem preventing access to localhost:5000 unless Airplay Receiver is disabled.


stefanv:
Alternatively, can reconfigure port number in config.

profjsb:
> Alternatively, can reconfigure port number in config.

@stefanv the port issue was discussed here (https://github.com/skyportal/skyportal/issues/1048) and #2657. I believe that @robertdstein found that 5000 was assumed in a few places and not properly propagated.  We should get clarity on this.

guynir42 (approved):

PR #3331, opened 2022-08-10
Add number of detections as an option for the queries

mcoughlin:
This PR adds the number of detections as an option for the queries.


Theodlz (approved):
LGTM!

PR #3336, opened 2022-08-11
SQLA disable warnings

guynir42:
Pass `overlaps` argument into `join_model` relationships and thus silence (but not fix) the SQLA warnings at app startup. 

I've fixed all the warnings but we should still investigate the underlying causes. 

Also, I've updated the cosmology loader to use the new methods, thus removing the astropy warning. 

This PR pins a branch of baselayer, when the PR on baselayer is merged, it should be changed to pin the main branch instead.


skyportal/models/obj.py
```diff
@@ -249,15 +249,6 @@ class Obj(Base, conesearch_alchemy.Point):
         doc="Internal key used for secure websocket messaging.",
     )
 
-    candidates = relationship(
```
mcoughlin, line 252: Do we not use this anymore?

skyportal/models/spectrum.py
```diff
@@ -385,8 +387,16 @@ def from_ascii(
         )
 
 
-SpectrumReducer = join_model("spectrum_reducers", Spectrum, User)
-SpectrumObserver = join_model("spectrum_observers", Spectrum, User)
+SpectrumReducer = join_model(
```
mcoughlin, line 390: This is clearly a big improvement.

mcoughlin (approved):
Just curious why the removal of candidates from the object table.

skyportal/models/obj.py
guynir42: It is defined after the class code.

skyportal/models/spectrum.py
guynir42: Yeah. I'm really surprised this thing even worked until now. The two relationships were overlapping and probably one killed the other.

mcoughlin: Yeah...

skyportal/models/obj.py
mcoughlin: Ah ok.

guynir42:
Test failures are all unrelated (the usual stuff). I think this could use a final review before merging @mcoughlin.

mcoughlin (approved):
LGTM!

PR #3340, opened 2022-08-12
SourceAnalysisPage: Better display of the plots

Theodlz:
Currently, on the analysis page, plots img width matches the width of the page. However, unless they are ina 16/9 ratio, they are cut and we can't see the whole height of the plots, even if we scroll. It's simply cut. 
To visualize some plots correctly, you are forced to download them.

This PR addresses that issue, by making sure that the plots img height matches the height of the page instead. That way, they might appear smaller, but the user can actually see the plots whole, and doesn't have to download them to see them properly.

That way, we also know that plots stay visible and are displayed properly even when we change the size of the browser's window.

Preview
analysis_page_2

### DIFF
--- a/static/js/components/SourceAnalysisPage.jsx
+++ b/static/js/components/SourceAnalysisPage.jsx
@@ -27,7 +27,6 @@ const useStyles = makeStyles((theme) => ({
   root: ({ size }) => ({
     width: size,
     margin: "0.5rem auto",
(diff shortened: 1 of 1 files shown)


profjsb (approved):

Theodlz:
Thanks! Hopefully not my last PR 😉

PR #3346, opened 2022-08-12
Enable different observation modes for LCO requests

mcoughlin:
This PR enables different observation modes for LCO requests.


guynir42 (approved):
Looks good overall. I am getting a sense that there's a lot of duplicated code. I leave it to you to consider if there's a better way to do this.

PR #3347, opened 2022-08-13
Filter out GcnEvents with a specific tag

mcoughlin:
This PR enables filtering out GcnEvents with a specific tag.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -450,12 +450,19 @@ class GcnEventHandler(BaseHandler):
                 Arrow-parseable date string (e.g. 2020-01-01). If provided, filter by
                 dateobs <= endDate
             - in: query
-              name: tag
+              name: tagKeep
               nullable: true
               schema:
                 type: string
               description: |
                 Gcn Tag to match against
+            - in: query
+              name: tagRemove
+              nullable: true
+              schema:
+                type: string
+              description: |
+                Gcn Tag to filter out
         responses:
           200:
             content:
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/gcn.py
```diff
@@ -450,12 +450,19 @@ async def get(self, dateobs=None):
...
                 type: string
               description: |
                 Gcn Tag to match against
+            - in: query
+              name: tagRemove
+              nullable: true
+              schema:
+                type: string
+              description: |
+                Gcn Tag to filter out
```
guynir42, line 465: We should do something similar to this with source classifications

skyportal/handlers/api/gcn.py
```diff
@@ -547,15 +555,24 @@ async def get(self, dateobs=None):
...
                 )
+            if tag_remove:
+                tag_subquery = (
+                    GcnTag.select(session.user_or_token)
+                    .where(GcnTag.text.contains(tag_remove))
+                    .subquery()
+                )
+                query = query.join(
+                    tag_subquery, GcnEvent.dateobs != tag_subquery.c.dateobs
+                )
```
guynir42, line 575: Could this work with just `.where(~GcnTag.text.contains(tag_remove))`?

guynir42 (approved):
Minor comment below. I am also assuming here that you've actually filtered on GCNs using the new front end keep/reject form options.

skyportal/handlers/api/gcn.py
mcoughlin: I did try that but it was confused!

skyportal/handlers/api/gcn.py
mcoughlin: yeah maybe one of the undergrads can merge the logic in a future PR.

PR #3348, opened 2022-08-13
Query on allocation rather than proposal ID

mcoughlin:
This PR queries on allocation rather than proposal ID for default observation plans, relevant as not all allocations have proposal IDs specified.

### DIFF
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -1,5 +1,6 @@
 from gcn_kafka import Consumer
 from datetime import datetime, timedelta
+import sqlalchemy as sa
 
 from baselayer.log import make_log
 from baselayer.app.models import init_db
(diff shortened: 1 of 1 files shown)


services/gcn_service/gcn_service.py
```diff
@@ -23,12 +24,24 @@
...
-
 log = make_log('gcnserver')
 
+with DBSession() as session:
+    config_gcn_observation_plans = []
+    for config_gcn_observation_plan in config_gcn_observation_plans_all:
+        allocation = session.scalars(
+            sa.select(Allocation).where(
+                Allocation.proposal_id
+                == config_gcn_observation_plan["allocation-proposal_id"]
```
guynir42, line 38: Just making sure the dash and underscore are correct.

guynir42 (approved):
LGTM

PR #3360, opened 2022-08-17
mark analysis test as flaky

mcoughlin:
This PR adds retries for the analysis test, which is flaky.

### DIFF
--- a/skyportal/tests/flaky/test_analysis_flaky.py
+++ b/skyportal/tests/flaky/test_analysis_flaky.py
@@ -1,0 +1,83 @@
+import uuid
+import json
+import time
+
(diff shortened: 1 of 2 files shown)


guynir42 (approved):
I would try to figure out what is the underlying causes for all this flakiness. But for now this is fine.

guynir42 (changes requested):
Actually this is not solving the underlying problem. This test never passes, no matter the retries.

mcoughlin:
@guynir42 Let's move it to flaky? Seems fine locally.

guynir42:
That's fine but make an issue about it. We can't just sweep tests under the carpet.

PR #3361, opened 2022-08-17
Retries for the gcn summary test

mcoughlin:
This PR adds retries to the gcn summary test.


guynir42 (approved):
Good. But I would argue that the right thing to do in the future is to get the GW data in as a fixture, which would make it much faster to run all these tests. Also worth while to get a smaller skymap (maybe with lower resolution?)

PR #3367, opened 2022-08-17
Fix comments query

mcoughlin:
This PR fixes the comment query to be done in the DB.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -724,71 +724,62 @@ def get_sources(
             raise ValueError(
                 "Invalid annotationsFilterOrigin value -- must provide at least one string value"
             )
-    if comments_filter is not None:
-        if isinstance(comments_filter, str) and "," in comments_filter:
-            comments_filter = [c.strip() for c in comments_filter.split(",")]
-        elif isinstance(comments_filter, str):
-            comments_filter = [comments_filter]
-        else:
-            raise ValueError(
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/source.py
```diff
@@ -733,6 +733,38 @@ def get_sources(
             raise ValueError(
                 "Invalid commentsFilter value -- must provide at least one string value"
             )
+
+        comment_query = Comment.select(session.user_or_token).where(
+            Comment.text.in_(comments_filter)
+        )
```
guynir42, line 739: Why not filter on all the comment related stuff in one subquery?

skyportal/handlers/api/source.py
```diff
@@ -1004,49 +1054,6 @@ def get_sources(
                                 break
                 if not passes_filter:
                     continue
-            if (
```
guynir42, line 1007: I'm happy to see this go into the direct query of the DB. 

We should think about making a partial text search of the comments at some point. I'm sure it is something that's pretty commonly done in similar situations.

guynir42 (commented):
Why have multiple subqueries?

skyportal/handlers/api/source.py
mcoughlin: Yeah definitely, just not obvious how to do easily.

skyportal/handlers/api/source.py
mcoughlin: Should be fixed!

PR #3369, opened 2022-08-18
Restrict group sources on download

mcoughlin:
This PR restricts the group sources on download.


guynir42 (approved):
Seems like a good idea.

PR #3373, opened 2022-08-19
Session detached for thumbnails

mcoughlin:
This PR tries to address session detached errors associated with object thumbnails.

### DIFF
--- a/skyportal/handlers/api/candidate.py
+++ b/skyportal/handlers/api/candidate.py
@@ -1118,12 +1118,13 @@ class CandidateHandler(BaseHandler):
                     f"Failed to post candidate for object {obj.id}: {e.args[0]}"
                 )
 
+            obj_id = obj.id
             calling_user_id = self.associated_user_object.id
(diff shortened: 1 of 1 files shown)


guynir42 (commented):
I don't think this solves it. You really want to use the same session inside the function, or make sure the thumbnails are attached to this session when the function finishes. 
This might be a bit dangerous with multiple processes trying to access the same session.

mcoughlin:
@guynir42 But Stefan also always tells us that we can't send sessions into the background process. What's the right move here?

guynir42:
The right thing to do is to figure out which part really needs to be done asynchronously and do only that part in the background. 
What was the error? what was the thing that is detached?

mcoughlin:
I assume the thumbnail stuff is async because of potential delays in the Kowalski query. It has the usual detached obj message, Josh gave the full trace in slack.

guynir42 (approved):
I've changed my mind. I think this will solve the problem.

PR #3375, opened 2022-08-20
Resolve recent test failures

mcoughlin:
This PR resolves recent test failures.


skyportal/models/user_notification.py
```diff
@@ -402,7 +402,11 @@ def receive_after_flush(session, context):
 
         for user in users:
             # Only notify users who have read access to the new record in question
-            pref = user.preferences['notifications']
+            if user.preferences is not None:
+                pref = user.preferences['notifications']
+            else:
+                pref = None
```
guynir42, line 408: If you do `pref={}` you may be able to simplify the following conditionals.

skyportal/models/user_notification.py
```diff
@@ -412,7 +416,9 @@ def receive_after_flush(session, context):
                 is not None
             ):
                 if is_gcnevent:
-                    if "gcn_notice_types" in pref["gcn_events"].keys():
+                    if (pref is not None) and "gcn_notice_types" in pref[
+                        "gcn_events"
+                    ].keys():
```
guynir42, line 421: Can remove `.keys()`

guynir42 (approved):
Minor comments below.

guynir42:
I've restarted the failed test run. If it isn't all green I would spend a little more time investigating why before merging.

mcoughlin:
Down to a test that works very well locally but struggles. Will spend more time this week.

PR #3378, opened 2022-08-21
Update candidate handler to new syntax

mcoughlin:
This PR moves the candidate handler to the new syntax.


guynir42:
The two failing tests here are failing on all PRs. Let's try to get them green this week?

guynir42 (approved):

PR #3380, opened 2022-08-21
Upgrade marshmallow_sqlalchemy

mcoughlin:
This PR upgrades marshmallow-sqlalchemy.


skyportal/models/schema.py
```diff
@@ -136,10 +136,11 @@ def add_schema(schema_class_name, exclude=[], add_to_model=False):
                     {
                         'model': class_,
                         'sqla_session': _DBSession,
+                        'load_instance': True,
                         'ordered': True,
                         'exclude': [],
                         'include_fk': True,
-                        'include_relationships': False,
+                        'include_relationships': True,
```
guynir42, line 143: What do this changes actually do?

guynir42 (approved):

skyportal/models/schema.py
mcoughlin: Not quite sure, but it was one of the required changes as specific in stack overflow to make it happen.

guynir42: Let's make sure the test failures are unrelated then merge.

PR #3382, opened 2022-08-22
Allow retries on endpoints from data loader

mcoughlin:
This PR allows retries on endpoints from data loader (to try to make the app loading more robust).


guynir42 (approved):
Great!

PR #3384, opened 2022-08-22
Function for gcn data loading in tests

mcoughlin:
This PR makes one function for loading gcn data.

### DIFF
--- a/skyportal/tests/api/test_gcn.py
+++ b/skyportal/tests/api/test_gcn.py
@@ -2,27 +2,67 @@ import os
 import numpy as np
 
 from skyportal.tests import api
-from skyportal.utils.gcn import from_url
+from skyportal.utils.gcn import from_url, get_dateobs
 
+import lxml
 import time
 import uuid
 import pandas as pd
 from regions import Regions
 from astropy.table import Table
+from astropy.time import Time
 
 import pytest
+
+
+def load_gcnevent(datafile, token):
+
+    with open(datafile, 'rb') as fid:
+        payload = fid.read()
+    data = {'xml': payload}
+    root = lxml.etree.fromstring(payload)
+
+    dateobs = Time(get_dateobs(root)).isot
+
+    RETRIES = 10
+
+    # wait for the gcnevents to populate
+    ntries = 0
+    gcnevent_loaded = False
+
+    while (ntries < RETRIES) and not gcnevent_loaded:
+        status, data = api('POST', 'gcn_event', data=data, token=token)
+        if status != 200:
+            ntries = ntries + 1
+            time.sleep(5)
+            continue
+        else:
+            gcnevent_loaded = True
+    assert data['status'] == 'success'
+
+    gcnevent_id = data['data']['gcnevent_id']
+
+    # wait for event to load
+    nretries = 0
+    gcnevent_loaded = False
+
+    while (ntries < RETRIES) and not gcnevent_loaded:
+        status, data = api('GET', f"gcn_event/{dateobs}", token=token)
+        if status != 200:
+            nretries = nretries + 1
+            time.sleep(5)
+            continue
+        else:
+            gcnevent_loaded = True
+    assert data['status'] == 'success'
+
+    return gcnevent_id
 
 
 def test_gcn_GW(super_admin_token, view_only_token):
 
     datafile = f'{os.path.dirname(__file__)}/../data/GW190425_initial.xml'
-    with open(datafile, 'rb') as fid:
-        payload = fid.read()
-    data = {'xml': payload}
-
-    status, data = api('POST', 'gcn_event', data=data, token=super_admin_token)
-    assert status == 200
-    assert data['status'] == 'success'
+    load_gcnevent(datafile, super_admin_token)
 
     dateobs = "2019-04-25 08:18:05"
     status, data = api('GET', f'gcn_event/{dateobs}', token=super_admin_token)
@@ -88,13 +128,7 @@ def test_gcn_GW(super_admin_token, view_only_token):
 def test_gcn_Fermi(super_admin_token, view_only_token):
 
     datafile = f'{os.path.dirname(__file__)}/../data/GRB180116A_Fermi_GBM_Gnd_Pos.xml'
-    with open(datafile, 'rb') as fid:
-        payload = fid.read()
-    data = {'xml': payload}
-
-    status, data = api('POST', 'gcn_event', data=data, token=super_admin_token)
-    assert status == 200
-    assert data['status'] == 'success'
+    load_gcnevent(datafile, super_admin_token)
 
     dateobs = "2018-01-16 00:36:53"
     params = {"include2DMap": True}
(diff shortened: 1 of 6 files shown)


guynir42:
Something got broken with the groups here (see the failed tests).

mcoughlin:
@guynir42 I think it was an issue with where I put the helper function. Should be better.

guynir42:
I think you should just make a few fixtures that load these GCNs into memory, and use `scope="session"` so that same GCN is only loaded once for the entire test suite <https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes>.

mcoughlin:
@guynir42 i couldn't figure out how to do a fixture that was a post? Basically we want one that is the whole GcnEvent post framework.

guynir42:
What if you just add the `load_gcnevent()` function into `conftest.py` and in each fixture just call it with the right filename. 
You can grab a token in the definition of each token as we do in the test signatures. 
I don't see why that shouldn't work. 
If you use `scope="session"` that would save you posting the same GCN multiple times, at least.

mcoughlin:
@guynir42 Nice idea. Seems to be working.

skyportal/tests/conftest.py
```diff
@@ -1638,3 +1645,48 @@ def analysis_token(user):
     )
     yield token_id
     delete_token(token_id)
+
+
+@pytest.fixture(scope="function")
```
guynir42, line 1650: If you set `scope='session'` you would not need to re-load them for each test.

skyportal/tests/conftest.py
```diff
@@ -1638,3 +1645,48 @@ def analysis_token(user):
...
+
+    with open(datafile, 'rb') as fid:
+        payload = fid.read()
+
+    user_id = 1
+    with DBSession() as session:
+        event_id = post_gcnevent_from_xml(payload, user_id, session)
+        time.sleep(30)
+        yield event_id
+        GcnEventFactory.teardown(event_id)
```
guynir42, line 1662: I'm not sure how good it is to keep the session open until teardown...
If you are passing an ID it would probably be ok to just close the session, yield the ID, and open a new one to delete the data.

skyportal/tests/fixtures.py
```diff
@@ -228,14 +226,18 @@ class Meta(BaseMeta):
...
-    def teardown(instrument):
-        if is_already_deleted(instrument, Instrument):
-            return
-
-        telescope = instrument.telescope.id
-        DBSession().delete(instrument)
-        DBSession().commit()
-        TelescopeFactory.teardown(telescope)
+    def teardown(instrument_id):
+        instrument_ = (
```
guynir42, line 230: Just curious, why the underscore?

skyportal/tests/fixtures.py
```diff
@@ -530,26 +532,56 @@ def teardown(obj):
...
+        gcnnotice_ = (
+            DBSession()
+            .execute(sa.select(GcnNotice).filter(GcnNotice.id == gcnnotice_id))
+            .scalars()
+            .first()
+        )
+        if gcnnotice_ is not None:
+            # If it is, delete it
+            DBSession().delete(gcnnotice_)
+            DBSession().commit()
```
guynir42, line 559: If you are using a new session to tear down, it is even more important not to have this inside another session. 
And I would use a context manager here as well.

skyportal/tests/fixtures.py
```diff
@@ -530,26 +532,56 @@ def teardown(obj):
...
+            .execute(sa.select(GcnEvent).filter(GcnEvent.id == gcnevent_id))
+            .scalars()
+            .first()
+        )
+        if gcnevent_ is not None:
+            # If it is, delete it
+            gcn_notices = gcnevent_.gcn_notices
+            for gcn_notice in gcn_notices:
+                GcnNoticeFactory.teardown(gcn_notice.id)
+            DBSession().delete(gcnevent_)
```
guynir42, line 583: same here

guynir42 (commented):
Looks good. Thank you for doing this, I know it is a bigger refactor than you thought when you started to work on this... 

We will probably want to do a dedicated PR to move all the fixtures to using context managers at some point, but for now let's make sure the new ones are up to date.

skyportal/tests/conftest.py
mcoughlin: I think we want to reload them when we also delete and add to them. Or maybe that is taken care of?

skyportal/tests/fixtures.py
mcoughlin: Just imitating some other handlers that had this ;)

skyportal/tests/fixtures.py
mcoughlin: So maybe we need to build the session and pass it in for teardown?

skyportal/tests/conftest.py
guynir42: How important is it for the tests if you make some modifications to each dataset? 
How much data is in each file? 
If there's not a lot of data and the changes made by each test are substantial, then leaving it with function scope is right. 
If you load a bunch of data and it takes 10 minutes to load and then the test makes some small changes to it that don't change the outcome of other tests, then session scope.

skyportal/tests/fixtures.py
guynir42: Not necessarily. If the teardown is a logical unit, that doesn't need to get objects connected to a session (you are giving IDs not objects) then just starting a new session inside the function is a good way. Just turn it into a context manager when the function starts, so you don't leave an open session. 
You could get into trouble if you already opened a session outside, and then DBSession grabs it, and then the context manager closes that in the teardown. That's why we need to move the whole concept of session to be more localized, so these sort of things don't happen.

mcoughlin:
@guynir42 https://github.com/skyportal/skyportal/runs/8002016467?check_suite_focus=true#step:15:4371
Can you explain how it should work with these context managers spinning things up? But then you have other things passed:

@pytest.fixture()
def public_group(public_stream):
    group = GroupFactory(streams=[public_stream])
    group_id = group.id
    yield group
    GroupFactory.teardown(group_id)

I tried to open everything in a session manager and commit it and so on but it doesnt seem to behave nicely.

guynir42:
I agree, it is a problem, particularly when you are returning an object, not an ID. 
We should think deeper on whether we want to have one session for the entire test suite, or to figure out a way to open a new session for each test, or what. 
For now, I suggest you use context managers only for the GCN stuff, because you are returning an ID and not an object connected to a session. 
Once that works we can figure out the rest.

mcoughlin:
@guynir42 sounds good; thanks for helping me push on this. Seems to be an ok balance.

mcoughlin:
@guynir42 Can you rereview this?

guynir42 (approved):
LGTM

guynir42:
@mcoughlin need another review or is this ready to go?

mcoughlin:
@guynir42 I think it's breaking things but I am unsure why.

guynir42:
Is this causing new tests that previously passed to fail, on local testing as well?

mcoughlin:
@guynir42 Not locally, but on GA.

mcoughlin:
@Theodlz Never quite got this to work, but clearly would be valuable.

mcoughlin:
@Theodlz we can close this right? Nothing to learn from this probably.

Theodlz:
We added GCN fixtures a little over a month ago, so this isn't an issue anymore. Closing

PR #3385, opened 2022-08-23
Source annotation query in database

mcoughlin:
This PR moves the source annotation query into the database.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -11,6 +11,8 @@ from dateutil.parser import isoparse
 import sqlalchemy as sa
 from sqlalchemy.orm import joinedload
 from sqlalchemy import func, or_, distinct
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.sql.expression import cast
 import arrow
 from marshmallow import Schema, fields
 from marshmallow.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/source.py
```diff
@@ -725,6 +751,78 @@ def get_sources(
...
+                session,
+                annotations_filter_origin=annotations_filter_origin,
+                annotations_filter_before=annotations_filter_before,
+                annotations_filter_after=annotations_filter_after,
+            )
+            annotations_subquery = annotations_query.subquery()
+            obj_query = obj_query.join(
+                annotations_subquery,
+                Obj.id == annotations_subquery.c.obj_id,
+            )
```
guynir42, line 824: Can this be outside the `if annotations_filter is not None:`?

skyportal/handlers/api/source.py
```diff
@@ -962,91 +1060,6 @@ def get_sources(
...
-                            comp_check = comp_function(data_value, value)
-                            if not comp_check:
-                                passes_filter = False
-                                break
-                if not passes_filter:
-                    continue
-
             obj_list.append(obj.to_dict())
 
             if include_comments:
```
guynir42, line 1065: Didn't we already factor out the comments filter into a DB query? or was that for candidates?

guynir42 (approved):
Minor comments:

skyportal/handlers/api/source.py
mcoughlin: Yeah but that is the query. The code below still adds them in.

skyportal/handlers/api/source.py
mcoughlin: Unfortunately no because that is a loop inside of there (you can have multiple filters on those annotations).

PR #3390, opened 2022-08-26
Remove accents from character names in attachment strings

mcoughlin:
This PR removes an error where the app falls over if an attachment with accents is uploaded in the name.


guynir42 (approved):
I'm not a big expert in this, so I assume you've tested this. If so, feel free to merge.

PR #3394, opened 2022-08-26
Add interferometers

mcoughlin:
This PR adds an interferometer model.


guynir42:
First of all, Friday at 4 is not a good time to share an 18 file PR. 

Other than that (I will review it next week probably...) I do think we might find a better name than "interferometer". 
I like that it is one word, but I'm not sure if that really is a good description for a GW detector. 
You may eventually think about other things that are interferometers (like visible light or radio interferometers) and those will probably not fit into the same model as LIGO. 
And you may find other things that should live with GW detectors (like gamma ray detectors, I guess) that are really not interferometers. So I think we should make an effort to choose a better name. 
I don't have a better idea off the top of my head, though.

PR #3397, opened 2022-08-29
Add neutrino tags

mcoughlin:
This PR adds neutrino tags to the ingester.


skyportal/utils/gcn.py
```diff
@@ -101,6 +112,15 @@ def get_tags(root):
     if search is not None:
         yield search.attrib['value']
 
+    # Get Instruments, if if present.
```
guynir42, line 115:
```suggestion
    # Get Instruments, if present.
```

guynir42 (approved):

PR #3403, opened 2022-08-30
Add ability to get LCO images from archive

mcoughlin:
This PR adds the ability to get LCO images from archive.


guynir42 (approved):
LGTM

PR #3404, opened 2022-08-31
Ground velocities

mcoughlin:
This PR adds basic earthquake class to track interferometer "weather." This is important as it will allow us to track ground velocities, the main consideration for detector data quality, to correlate with alert notices.


skyportal/handlers/api/comment.py
```diff
@@ -415,6 +438,26 @@ def post(self, associated_resource_type, resource_id):
...
+            elif associated_resource_type.lower() == "earthquake":
+                earthquake_id = resource_id
+                earthquake = session.scalars(
+                    EarthquakeEvent.select(self.current_user).where(
+                        EarthquakeEvent.id == earthquake_id
+                    )
+                ).first()
+                if earthquake is None:
+                    return self.error(
+                        f'Could not find any accessible gcn events with ID {gcnevent_id}.'
```
guynir42, line 450: should be earthquake id.

skyportal/handlers/api/comment.py
```diff
@@ -687,6 +750,11 @@ def put(self, associated_resource_type, resource_id, comment_id):
                     action='skyportal/REFRESH_SOURCE_GCN',
                     payload={'obj_internal_key': c.obj.internal_key},
                 )
+            elif isinstance(c, CommentOnEarthquake):  # also update the earthquake
+                self.push_all(
+                    action='skyportal/REFRESH_EARTHQUAKE',
+                    payload={'obj_internal_key': c.obj.internal_key},
```
guynir42, line 756: Probably want the earthquake ID here too.

skyportal/handlers/api/comment.py
```diff
@@ -775,6 +843,17 @@ def delete(self, associated_resource_type, resource_id, comment_id):
...
+                c = session.scalars(
+                    CommentOnEarthquake.select(self.current_user, mode="delete").where(
+                        CommentOnEarthquake.id == comment_id
+                    )
+                ).first()
+                if c is None:
+                    return self.error(
+                        'Could not find any accessible comments.', status=403
+                    )
+                comment_resource_id_str = str(c.gcn_id)
```
guynir42, line 856: earthquake id

skyportal/handlers/api/comment.py
```diff
@@ -817,6 +896,11 @@ def delete(self, associated_resource_type, resource_id, comment_id):
                     action='skyportal/REFRESH_GCNEVENT',
                     payload={'gcnEvent_dateobs': gcnevent_dateobs},
                 )
+            elif isinstance(c, CommentOnEarthquake):  # also update the earthquake
+                self.push_all(
+                    action='skyportal/REFRESH_EARTHQUAKE',
+                    payload={'obj_internal_key': obj_key},
```
guynir42, line 902: earthquake id

skyportal/models/earthquake.py
```diff
@@ -0,0 +1,177 @@
...
+        sa.ForeignKey("earthquakeevents.event_id"),
+        nullable=False,
+        comment='Earthquake ID',
+    )
+
+    lat = sa.Column(sa.Float, nullable=False, comment='Latitude')
+
+    lon = sa.Column(sa.Float, nullable=False, comment='Longitude')
+
+    depth = sa.Column(sa.Float, nullable=False, comment='Depth')
```
guynir42, line 51: Is this relative to sea level? I'm considering if negative elevation is better here. But definitely want to define depth more carefully.

skyportal/models/earthquake.py
```diff
@@ -0,0 +1,177 @@
...
+
+    magnitude = sa.Column(sa.Float, nullable=False, comment='Magnitude', index=True)
```
guynir42, line 53: Is this on some specific scale? Not to be confused with astronomical magnitude.

skyportal/models/earthquake.py
```diff
@@ -0,0 +1,177 @@
...
+class EarthquakeNotice(Base):
+    """Earthquake notice information"""
+
+    update = delete = AccessibleIfUserMatches('sent_by')
+
+    sent_by_id = sa.Column(
+        sa.ForeignKey('users.id', ondelete='CASCADE'),
+        nullable=False,
+        index=True,
+        doc="The ID of the User who created this GcnEvent.",
```
guynir42, line 26:
```suggestion
        doc="The ID of the User who created this EarthquakeEvent.",
```

skyportal/models/earthquake.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    sent_by = relationship(
+        "User",
+        foreign_keys=sent_by_id,
+        back_populates="earthquakenotices",
+        doc="The user that saved this GcnEvent",
```
guynir42, line 33:
```suggestion
        doc="The user that saved this EarthquakeEvent",
```

skyportal/tests/api/test_earthquake.py
```diff
@@ -0,0 +1,54 @@
+import os
+import numpy as np
+
+from skyportal.tests import api
+
+
+def test_earthquake(super_admin_token, view_only_token):
```
guynir42, line 7: Sounds like a button a super-villain would have on his dashboard.

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+    EarthquakeNotice,
+    EarthquakePrediction,
+    MMADetector,
+    User,
+)
+
+log = make_log('earthquake')
+
+
+Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
```
guynir42, line 29: Why not use the regular handler session?

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+    ).first()
+
+    if event is None:
+        event = EarthquakeEvent(
+            event_id=event_id, event_uri=event_uri, sent_by_id=user.id
+        )
+        session.add(event)
+    else:
+        if not event.is_accessible_by(user, mode="update"):
+            raise ValueError(
```
guynir42, line 91: Use `from baselayer.app.custom_exceptions import AccessError`

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+        event_id=event_id,
+        lat=origin.latitude,
+        lon=origin.longitude,
+        depth=origin.depth,
+        magnitude=mag,
+        date=origin.time.datetime,
+        sent_by_id=user.id,
+    )
+    session.add(earthquake_notice)
+    session.commit()
```
guynir42, line 106: I think that unless you get a "verified" session from `self.Session` of the handler, this `commit` does not check if the user has permissions to add things.

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+
+    return event.id
+
+
+def post_earthquake_from_dictionary(payload, user_id, session):
+    """Post Earthquake to database from dictionary.
+    payload: dict
+        Dictionary containing date and skymap
```
guynir42, line 114: skymap?

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+                schema: Error
+        """
+
+        with self.Session() as session:
+            statuses = (
+                session.scalars(sa.select(EarthquakeEvent.status).distinct())
+                .unique()
+                .all()
+            )
+            return self.success(data=statuses)
```
guynir42, line 183: This will return everything ever posted. What is this used for? Just to get a list of possible statuses?

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+                )
+
+        with self.Session() as session:
+            if 'xml' in data:
+                event_id = post_earthquake_from_xml(
+                    data['xml'], self.associated_user_object.id, session
+                )
+            else:
+                event_id = post_earthquake_from_dictionary(
+                    data, self.associated_user_object.id, session
```
guynir42, line 238: If you only ever post using `self.Session` you can use the associated user on that session instead of passing it in.

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+        single:
+          description: Retrieve an Earthquake event
+          tags:
+            - earthquakeevents
+          parameters:
+            - in: path
+              name: date
+              required: false
+              schema:
+                type: string
```
guynir42, line 256: This parameter should be `event_id` and an `int`.

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+              description: |
+                Earthquake Status to match against
+            - in: query
+              name: statusRemove
+              nullable: true
+              schema:
+                type: string
+              description: |
+                Earthquake Status to filter out
```
guynir42, line 300: Add page number and num per page.

skyportal/handlers/api/earthquake.py
```diff
@@ -0,0 +1,616 @@
...
+
+            (
+                Dist,
+                Ptime,
+                Stime,
+                Rtwotime,
+                RthreePointFivetime,
+                Rfivetime,
+            ) = compute_traveltimes(notice, detector)
+            Rfamp, Lockloss = 0.0, 0.0
```
guynir42, line 541: This seems strange.

guynir42 (commented):
I am quite "shocked" by the number of new things we have to track to get earthquakes into the mix. 

I am curious to see how this would actually work. Would the earthquake prediction come long enough before the actual earthquake to be useful? Would people really need reminders about earthquakes? We should probably discuss this on the next call. 

Overall the code looks good. I have some comments:

skyportal/handlers/api/comment.py
mcoughlin: yeah.

skyportal/handlers/api/comment.py
mcoughlin: yup.

skyportal/handlers/api/comment.py
mcoughlin: yes.

skyportal/models/earthquake.py
mcoughlin: any suggestions? earthquake folks also use that term.

skyportal/models/earthquake.py
mcoughlin: oops.

skyportal/models/earthquake.py
mcoughlin: oops.

skyportal/tests/api/test_earthquake.py
mcoughlin: We are going to give you this button.

skyportal/handlers/api/earthquake.py
mcoughlin: I am also triggering these in the background currently.

skyportal/handlers/api/earthquake.py
mcoughlin: oops.

skyportal/handlers/api/earthquake.py
mcoughlin: yeah will be used for filtering on the frontend.

skyportal/handlers/api/earthquake.py
mcoughlin: yeah will also be used in the service, so also need a DBSession() style.

skyportal/handlers/api/earthquake.py
mcoughlin: good idea.

skyportal/handlers/api/earthquake.py
guynir42: I guess you'd have to explicitly call `is_accessible` with the 'create' option and the user.

mcoughlin:
@guynir42 Luckily, teleseismic earthquakes can take thirty minutes or more to get around the world, so we can make configuration and controls changes. It's really neat!

mcoughlin:
@guynir42 you make changes to the controls, i.e. filter the frequencies that you seismically monitor and respond to differently.

skyportal/handlers/api/earthquake.py
mcoughlin: ok that's a nifty trick.

guynir42:
@mcoughlin what controls? I didn't know we had filters for different frequencies in skyportal...

mcoughlin:
@guynir42 I meant on the GW detector side, so they can use this to change configurations and follow-up can use it for data quality.

skyportal/handlers/api/comment.py
```diff
@@ -748,17 +748,17 @@ def put(self, associated_resource_type, resource_id, comment_id):
...
                 self.push_all(
                     action='skyportal/REFRESH_EARTHQUAKE',
-                    payload={'obj_internal_key': c.obj.internal_key},
+                    payload={'earthquake_id': c.earthquake.event_id},
                 )
             elif isinstance(c, CommentOnShift):  # also update the shift
                 self.push_all(
                     action='skyportal/REFRESH_SHIFT',
-                    payload={'obj_internal_key': c.obj.internal_key},
+                    payload={'shift_id': c.shift_id},
```
guynir42, line 761: How did this even work until now??

guynir42 (approved):
Looks good. I am still not fully on board with why we need earthquake predictions. 
I suggest we chat about this before merging. Maybe @profjsb wants to weigh in as well.

mcoughlin:
Because this means both the sites and us can use these for data quality, meaning we will be using the same data quality criteria as the sites, simplifying interpretation of data quality metrics when doing go vs no go followup decisions.

skyportal/handlers/api/comment.py
mcoughlin: well we dont have any front-end mechanism for editing comments...

alembic/versions/d0047bf55eac_eq.py
```diff
@@ -0,0 +1,431 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        'earthquakeevents',
+        sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column('created_at', sa.DateTime(), nullable=False),
+        sa.Column('modified', sa.DateTime(), nullable=False),
+        sa.Column('sent_by_id', sa.Integer(), nullable=False),
+        sa.Column('event_id', sa.String(), nullable=False, comment='Earthquake ID'),
+        sa.Column('event_uri', sa.String(), nullable=True, comment='Earthquake URI'),
```
profjsb, line 28: Should probable change this to a sa util `URLType` (https://sqlalchemy-utils.readthedocs.io/en/latest/data_types.html#module-sqlalchemy_utils.types.url) instead of sa.String type.  I checked that URIs are properly parsed:

```python
import furl
furl.furl("ivo://nvo.caltech/voeventnet/catot#1004071150784109051").scheme
'ivo'
```

skyportal/app_server.py
```diff
@@ -201,22 +204,26 @@ def log_request(self, handler):
     (r'/api/galaxy_catalog/ascii', GalaxyASCIIFileHandler),
     (r'/api/galaxy_catalog(/[0-9A-Za-z-_\.\+]+)?', GalaxyCatalogHandler),
     (
-        r'/api/(sources|spectra|gcn_event|shift)/([0-9A-Za-z-_\.\+]+)/comments',
+        r'/api/earthquake/([0-9A-Za-z-_\.\+]+)/ifo/([0-9A-Za-z-_\.\+]+)/predictions',
```
profjsb, line 207: what does `ifo` mean? Add a comment here?

skyportal/models/earthquake.py
```diff
@@ -0,0 +1,183 @@
...
+    r3p5 = sa.Column(sa.DateTime, nullable=False, comment='R-3.5 km/s-wave time')
+
+    r5p0 = sa.Column(sa.DateTime, nullable=False, comment='R-5.0 km/s-wave time')
+
+    rfamp = sa.Column(
+        sa.Float, nullable=False, comment='Earthquake amplitude predictions [m/s]'
+    )
+
+    lockloss = sa.Column(
+        sa.INT, nullable=False, comment='Earthquake lockloss prediction'
```
profjsb, line 100: why is this an INT? Might add some more info in the comment string.

profjsb (commented):
Looks good. A few requests:

1. Add some more comments in the PR entry as to why this is needed, how it will be used etc. I'm completely unaware of the utility of this (but know it must be important).

2. Format the `quakeml.xml` file with line breaks so it's human readable.

3. Need a test for group access (and notification)

alembic/versions/d0047bf55eac_eq.py
mcoughlin: Added!

skyportal/app_server.py
mcoughlin: Clarified that this is an mmadetector.

skyportal/models/earthquake.py
mcoughlin: Yeah changed this and added more info.

mcoughlin:
@profjsb I added a permissions test on the delete (this is the only non-public one, by design) and more info on the PR and cleaned up quakeml.

profjsb (approved):
Should still explain what lockloss is and why it's an INT. But otherwise fine.

mcoughlin:
@profjsb I will get this in the follow-on PR.

PR #3406, opened 2022-09-01
MMADetector spectrum

mcoughlin:
This PR adds spectrum for MMA detectors.


skyportal/models/mmadetector.py
```diff
@@ -74,6 +78,186 @@ class MMADetector(Base):
...
+        doc="MMADetectorSpectra of the object.",
+    )
+
 
 GcnEventMMADetector = join_model("gcnevents_mmadetectors", GcnEvent, MMADetector)
 GcnEventMMADetector.__doc__ = "Join table mapping GcnEvents to MMADetectors."
+
+
+class MMADetectorSpectrum(Base):
+    """Wavelength-dependent measurement of the sensitivity the detector."""
```
guynir42, line 97: I would argue that Frequency dependent is a much better way to look at it.

skyportal/tests/data/aligo_O4high.txt
```diff
@@ -0,0 +1,2736 @@
+1.021659e+01 6.662774e-21
```
guynir42, line 1: Can we name this something more informative (the words "noise" and "spectrum" should be up there with "LIGO").

guynir42 (approved):
I have minor comments. 
Please update your migration hash, as it may have changed since the tests were last run.

skyportal/tests/data/aligo_O4high.txt
mcoughlin: Yeah added!

skyportal/models/mmadetector.py
mcoughlin: OOps.

PR #3433, opened 2022-09-02
make non-classifications inclusive

mcoughlin:
This PR makes it such that if any undesired classification is present in addition to other (potentially desired classifications), the object will be rejected.


profjsb:
Is this always the desired behavior for all SP instances? Might this behavior or the current need to be selectable in a config?

mcoughlin:
I think so, reject should mean reject here I would think. Not just if it's the only classification.

PR #3434, opened 2022-09-03
Add MMADetectorTimeIntervals

mcoughlin:
This PR adds concept of MMADetectorTimeIntervals. Useful for tracking the on/off state for multi-messenger detectors.


guynir42 (commented):
Code looks fine. 
I'm not sure I understand what a "segment" is in this context. In general you should specify more details in the PR comment. Beyond that, I'm not sure it will be obvious to users what a segment is. I went through the code and still don't understand what it does or why it is called a segment (I would imaging a segment to be maybe one of two or three GW detector or one of the arms of an interferometer, so I was very surprised to see it is a segment in time).

mcoughlin:
@guynir42 Good point, I added some more explanation in the model.

guynir42:
@mcoughlin I believe we can find a better name than "segment" if we really put our mind to it.

mcoughlin:
@guynir42 TimeSegments?

guynir42:
@mcoughlin how about TimeInterval?

skyportal/handlers/api/mmadetector.py
```diff
@@ -610,3 +613,396 @@ def delete(self, spectrum_id):
...
+                if {g.id for g in groups} != set(group_ids):
+                    return self.error(
+                        f'Cannot find one or more groups with IDs: {group_ids}.'
+                    )
+
+                if groups:
+                    time_interval.groups = time_interval.groups + groups
+
+            if 'time_interval' in data:
+                setattr(time_interval, 'time_interval', data['time_interval'])
```
guynir42, line 944: Why not `time_interval.time_interval = data['time_interval']` ?

skyportal/models/group.py
```diff
@@ -365,6 +365,15 @@ class Group(Base):
...
     )
 
+    mmadetector_time_intervals = relationship(
+        "MMADetectorTimeInterval",
+        secondary="group_mmadetector_time_intervals",
+        back_populates="groups",
+        cascade="save-update, merge, refresh-expire, expunge",
+        passive_deletes=True,
+        doc='The MMADetector time intervals visible to this group.',
+    )
```
guynir42, line 375: Just wondering, do we really need to associate a time interval with a group? Is there expected to be permissioning on time intervals?

guynir42 (approved):
Looks good. Couple of small comments/questions:

skyportal/models/group.py
mcoughlin: I am kind of envisioning that one might want to post particular data subsets (i.e. detections or something), that might have restrictive access.

skyportal/handlers/api/mmadetector.py
mcoughlin: Good idea.

PR #3437, opened 2022-09-03
Add thumbnail headers

mcoughlin:
This PR adds thumbnail headers (to the images) to more accurately specify their origin.


guynir42 (approved):
Code looks good. 
Would be useful to say what are "thumbnail headers" in the PR comment.

PR #3438, opened 2022-09-05
Loading symbol for group sources

mcoughlin:
This PR uses a loading symbol if either saved or pending sources are still loading.


guynir42 (approved):
An important distinction, no doubt!

PR #3445, opened 2022-09-07
Update button syntax

jadalilleboe:
Update button syntax to follow the new standard documented here: [https://skyportal.io/docs/styling.html#buttons](https://skyportal.io/docs/styling.html#buttons)

Also added the --fix flag to the eslint precommit as I found it helpful for some linting issues. This[ stackoverflow post](https://stackoverflow.com/questions/50289536/is-it-safe-to-apply-autofix-from-eslint#:~:text=Yes%2C%20it's%20safe%2C%20because%20the,your%20JS%20issues%20%5B1%5D.) says that it should be ok

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
     hooks:
       - id: ESLint
         name: ESLint
-        entry: npx eslint -c .eslintrc.yaml
+        entry: npx eslint -c .eslintrc.yaml --fix
         language: system
         files: \.[j]sx?$
         types: [file]
--- a/static/js/components/About.jsx
+++ b/static/js/components/About.jsx
@@ -7,13 +7,13 @@ import makeStyles from "@mui/styles/makeStyles";
 import Card from "@mui/material/Card";
 import CardContent from "@mui/material/CardContent";
 import CardActions from "@mui/material/CardActions";
-import Button from "@mui/material/Button";
 import Typography from "@mui/material/Typography";
 import Box from "@mui/material/Box";
 import Paper from "@mui/material/Paper";
 
 import clsx from "clsx";
 import dayjs from "dayjs";
+import Button from "./Button";
 
 const useStyles = makeStyles((theme) => ({
   root: {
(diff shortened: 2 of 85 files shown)


jadalilleboe:
will fix failures tomorrow

jadalilleboe:
I think this is ready for review, this test seems to be flaky:
```
FAIL: skyportal/tests/frontend/sources_and_followup_etc/test_sources.py:251 in test_classifications

        driver.scroll_to_element(classifications)
    
        del_button_xpath = "//button[starts-with(@name, 'deleteClassificationButton')]"
>       driver.click_xpath(del_button_xpath, wait_clickable=False)

skyportal/tests/frontend/sources_and_followup_etc/test_sources.py:251: 
```
But it's flaky in [another PR's run as well](https://github.com/skyportal/skyportal/runs/8232689556?check_suite_focus=true#step:15:508) so I'm thinking my code changes aren't the issue.
@guynir42 this PR is related to [this PR](https://github.com/skyportal/skyportal/pull/3107)

.pre-commit-config.yaml
```diff
@@ -32,7 +32,7 @@ repos:
     hooks:
       - id: ESLint
         name: ESLint
-        entry: npx eslint -c .eslintrc.yaml
+        entry: npx eslint -c .eslintrc.yaml --fix
```
guynir42, line 35: what does this do?

guynir42 (approved):
LGTM. Good to have you back, Jada! (and also, wow we have a lot of buttons...)

.pre-commit-config.yaml
jadalilleboe: [https://eslint.org/docs/latest/user-guide/command-line-interface#--fix](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix)
Automatically fixes linting problems. I found it helpful when there's like 80+ linting errors that would be a pain to fix manually. I guess one concern would be if the automatic linter breaks the code in some way, but maybe we could try adding the flag for a little bit and see if there are any major issues?

guynir42: Yeah that sounds fine

PR #3446, opened 2022-09-07
Add properties table for ease of filtering events

mcoughlin:
This PR adds a properties table to GcnEvents to simplify filtering on events.


skyportal/handlers/api/gcn.py
```diff
@@ -598,6 +639,54 @@ async def get(self, dateobs=None):
                     tag_subquery, GcnEvent.dateobs != tag_subquery.c.dateobs
                 )
 
+            if properties_filter is not None:
+                for prop_filt in properties_filter:
+                    prop_split = prop_filt.split(":")
+                    if not (len(prop_split) == 1 or len(prop_split) == 3):
+                        raise ValueError(
+                            "Invalid propertiesFilter value -- propotation filter must have 1 or 3 values"
```
profjsb, line 647: propotation?

skyportal/handlers/api/gcn.py
```diff
@@ -598,6 +639,54 @@ async def get(self, dateobs=None):
                     tag_subquery, GcnEvent.dateobs != tag_subquery.c.dateobs
                 )
 
+            if properties_filter is not None:
+                for prop_filt in properties_filter:
+                    prop_split = prop_filt.split(":")
+                    if not (len(prop_split) == 1 or len(prop_split) == 3):
```
profjsb, line 645: the API spec makes it seem like there must be 3 values after split. How can you get one?

skyportal/handlers/api/gcn.py
```diff
@@ -598,6 +639,54 @@ async def get(self, dateobs=None):
...
+                        elif op == "le":
+                            comp_function = operator.le
+                        elif op == "eq":
+                            comp_function = operator.eq
+                        elif op == "ne":
+                            comp_function = operator.ne
+                        elif op == "ge":
+                            comp_function = operator.ge
+                        elif op == "gt":
+                            comp_function = operator.gt
```
profjsb, line 674:
```suggestion
                        comp_function = getattr(operator, op)
```

skyportal/tests/api/test_gcn.py
```diff
@@ -168,13 +185,27 @@ def test_gcn_from_moc(super_admin_token, view_only_token):
...
     data = data["data"]
     assert "2022-06-18T18:31:12" in [event["dateobs"] for event in data["events"]]
 
+    params = {'propertiesFilter': 'BNS: 0.5: gt'}
+    status, data = api('GET', 'gcn_event', token=super_admin_token, params=params)
+    assert status == 200
+    data = data["data"]
+    assert "2022-06-18T18:31:12" in [event["dateobs"] for event in data['events']]
+
+    params = {'propertiesFilter': 'BNS: 0.5: lt'}
```
profjsb, line 203: Test comma-separated filtering criteria?

skyportal/handlers/api/gcn.py
```diff
@@ -598,6 +639,54 @@ async def get(self, dateobs=None):
                     tag_subquery, GcnEvent.dateobs != tag_subquery.c.dateobs
                 )
 
+            if properties_filter is not None:
```
profjsb, line 642: I dont see how comma-separated criteria are parsed and used here

mcoughlin: We do the split and the loop around line 545.

skyportal/tests/api/test_gcn.py
mcoughlin: good idea.

skyportal/handlers/api/gcn.py
mcoughlin: ah nifty.

skyportal/handlers/api/gcn.py
mcoughlin: oops.

skyportal/handlers/api/gcn.py
mcoughlin: adding clarification to the spec.

guynir42:
@mcoughlin Let me know if and when you want me to review this.

mcoughlin:
@guynir42 should be ready for a look

guynir42 (commented):
I wonder what is the advantage of having this as a separate table from the GCNs themselves. I see the similarities to PhotStats but this seems simple enough to just put the properties on the object itself. 
Unless there's some use cases where you can have multiple properties connected to each GCN?

mcoughlin:
@guynir42 Yeah exactly, they update in time.

guynir42:
@mcoughlin Then why not just override the existing properties? If you are always going to look at the most recent? Or is this useful because we want to keep a history of the changes like in redshift?

mcoughlin:
@guynir42 Yeah thats what I am worried about. Plus we know the most recent haven't always been the best due to errors from the source material ;)

guynir42:
@mcoughlin so is there a way to query against older properties? I wonder if this is something that is going to be used?

mcoughlin:
@guynir42 For the moment, we are just building out the capabilities. I think we first will need to make a way to display them and so on, lots of work here to do.

guynir42:
@mcoughlin ok then.

PR #3447, opened 2022-09-07
Add Swift LSXPS catalog query

mcoughlin:
This PR adds an endpoint to query LSXPS (https://www.swift.ac.uk/LSXPS/transients/) and post the sources. The idea is to re-post periodically to grab new candidates.


skyportal/handlers/api/catalog_services.py
```diff
@@ -0,0 +1,223 @@
...
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  telescope_name:
+                    required: false
+                    type: integer
+                    description: |
+                      Name of telescope to assign this catalog too
```
guynir42, line 53:
```suggestion
                      Name of telescope to assign this catalog to
```

skyportal/handlers/api/catalog_services.py
```diff
@@ -0,0 +1,223 @@
...
+                schema: Success
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+
+        data = self.get_json()
+
+        telescope_name = data.get('telescope_name', 'Swift')
```
guynir42, line 68: In what case would the telescope be something other than Swift?

skyportal/handlers/api/catalog_services.py
```diff
@@ -0,0 +1,223 @@
...
+                            attachment_bytes = base64.b64encode(f.read())
+                        comment = Comment(
+                            text='Swift Detection Spectrum',
+                            obj_id=obj_id,
+                            attachment_bytes=attachment_bytes,
+                            attachment_name=attachment_name,
+                            author=user,
+                            groups=groups,
+                            bot=False,
+                        )
```
guynir42, line 217: Is there a reason why we are posting spectra in comments instead of as a Spectrum object?

skyportal/tests/api/test_catalog_services.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+    assert status == 200
+    assert data['status'] == 'success'
+
+    status, data = api(
+        "POST",
+        "catalogs/swift_lsxps",
+        data={'telescope_name': name},
+        token=super_admin_token,
+    )
```
guynir42, line 51: Does this just get ALL the transients in that database?

guynir42 (commented):
The code looks fine. 
Once again I am not sure how this works. Specifically I would expect the query to need more info like RA/Dec, but as it is it only needs to know the telescope.

skyportal/handlers/api/catalog_services.py
mcoughlin: Just because it is an X-ray spectrum :)

skyportal/tests/api/test_catalog_services.py
mcoughlin: All 6 of them ;)

skyportal/handlers/api/catalog_services.py
mcoughlin: This is in case someone used a different name.

mcoughlin:
@guynir42 Yeah this is a (very) small catalog. Maybe if we expand beyond the living transient catalog to the full source thing we need a cone search?

skyportal/handlers/api/catalog_services.py
guynir42: If a user specifies a different name would it still fetch the same sources from that catalog? Would they just be posted under a different name?

skyportal/handlers/api/catalog_services.py
guynir42: Not ideal but I imagine starting out a whole infrastructure to handle that sort of data is out of scope for this PR.

skyportal/handlers/api/catalog_services.py
mcoughlin: Yes exactly. But that instrument will also need the swiftxrt filter, so could not choose randomly.

skyportal/handlers/api/catalog_services.py
mcoughlin: Yes it would be nice to start building out our capabilities to deal with X-ray / radio data in particular.

skyportal/handlers/api/catalog_services.py
guynir42: I don't get it. The catalog that is queried here is strictly sources from Swift.

mcoughlin: Yes. But what if they didnt name it Swift? :)

guynir42: I seriously don't understand when that could happen. Wouldn't it be easier to just assume this handler is for Swift only?

mcoughlin: Well they could still do Neil Gehrels Swift Observatory and such. But I can remove it if that makes sense to you.

guynir42: You mean if a skyportal admin runs their own database and calls the swift telescope something other than `Swift` you would not find that telescope on the DB?

mcoughlin: yes.

guynir42: OK now I understand. I think you can leave it in. Maybe document that in the docstring (something like "use the same name as the nickname given to the Swift telescope").

guynir42 (approved):

PR #3451, opened 2022-09-08
Add the option of phone notifications

mcoughlin:
This PR adds the option of phone notifications.


skyportal/models/user_notification.py
```diff
@@ -291,6 +292,52 @@ def send_sms_notification(mapper, connection, target):
...
+                    current_time.hour >= timeslot[0]
+                    and current_time.hour <= timeslot[1]
+                ):
+                    sending = True
+            else:
+                if current_time.hour <= timeslot[0] or current_time.hour >= timeslot[1]:
+                    sending = True
+    if sending:
+        try:
+            message = f"Greetings. This is the SkyPortal robot. {target.text}"
```
profjsb, line 328: Where is `target.text` defined/set?

profjsb:
Can you think of a way to test this functionality programatically or do we have to do it live? Thinking about the Twilio magic phone numbers (https://www.twilio.com/docs/iam/test-credentials)

mcoughlin:
@profjsb I just tested it locally for safety. And the text is defined back in the notification itself.

skyportal/models/user_notification.py
```diff
@@ -291,6 +292,52 @@ def send_sms_notification(mapper, connection, target):
...
+        if len(timeslot) > 0:
+            current_time = arrow.utcnow().datetime
+            if timeslot[0] < timeslot[1]:
+                if (
+                    current_time.hour >= timeslot[0]
+                    and current_time.hour <= timeslot[1]
+                ):
+                    sending = True
+            else:
+                if current_time.hour <= timeslot[0] or current_time.hour >= timeslot[1]:
```
guynir42, line 324: Isn't this supposed to be `and` as well?

guynir42 (approved):

skyportal/models/user_notification.py
mcoughlin: Actually no because this is the option where you are wanting calls outside of a specific time block.

skyportal/models/user_notification.py
mcoughlin: This is the notification itself.

skyportal/models/user_notification.py
guynir42: Then shouldn't the inequalities be flipped? 
In this case `timeslot[0]` is bigger than `timeslot[1]`. 
Lets assume `timeslot[0]=15` and `timeslot[1]=10`. 
If current time is 9 or 13, it is `True` because `current_time.hour<=timeslot[0]`. 
If current time is 13 or 17, it is `True` because `current_time.hour>=timeslot[1]`. 
I don't think there are any values where it is `False`. 
Unless I missed something this should also be fixed in the SMS notification.

mcoughlin: I want @Theodlz to weigh in here as he thought about it most.

Theodlz: Guy is right! The 0 and the 1 need to be flipped in that else statement, it's a mistake. Please apply the same changes to the sms method so both are fixed.

PR #3453, opened 2022-09-09
PDL service

mcoughlin:
This PR adds a service that enables query of the USGS PDL client for posting live earthquakes.

### DIFF
--- a/services/pdl_service/pdl_service.py
+++ b/services/pdl_service/pdl_service.py
@@ -1,0 +1,48 @@
+from watchdog.observers import Observer
+from watchdog.events import LoggingEventHandler
+
+from baselayer.log import make_log
+from baselayer.app.models import init_db
+from baselayer.app.env import load_env
+
+from skyportal.handlers.api.earthquake import post_earthquake_from_xml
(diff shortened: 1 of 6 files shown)


services/pdl_service/pdl_service.py
```diff
@@ -0,0 +1,48 @@
...
+                        if "quakeml.xml" in event.src_path:
+                            with open(event.src_path) as fid:
+                                payload = fid.read()
+                            post_earthquake_from_xml(payload, user_id, session)
+
+                event_handler = Event()
+                observer = Observer()
+                observer.schedule(event_handler, path, recursive=True)
+                observer.start()  # for starting the observer thread
+                observer.join()
```
guynir42, line 38: Shouldn't there be a small pause between iterations on this endless loop?

guynir42 (approved):
Cool.

services/pdl_service/pdl_service.py
mcoughlin: The watchdog seems to take care of this, and the rate of new events is low enough that I don't expect us to be overwhelmed.

PR #3454, opened 2022-09-09
Fix URL link for thumbnails

mcoughlin:
This PR fixes the URL link for thumbnails (previously pointing just back to the source page).


guynir42 (approved):
Looks good to me.

Issue #3325, opened 2022-08-08
Galaxy Page - Issue with the list of galaxies


Theodlz:
The galaxy page seems to show more galaxies at once than it should (as if it wasn't using the pagination correctly). This results in slowdowns of the frontend (maybe even the backend), and the list not loading correctly.

StephenTheCoder:
It seems to be that the first page is showing a ton of galaxies but when you go to the second page, the problem seems to be resolved (and it seems to fix after going back to the first page). I encountered this bug on Fritz's GCN Events page as well.

PR #3455, opened 2022-09-09
Fixing Galaxy page issue using useEffect (issue reference #3325)

vyom1611:
The Galaxy page on initialization showed more galaxies at once than it should, but on changing the page, it was fixed. The issue only occurred at init, so I used a useEffect hook to change the page when it loads to page 1. Based on issue #3325.


mcoughlin:
Thanks @vyom1611. Could you please use the linter:

pre-commit install
pre-commit run --files static/js/components/GalaxyPage.jsx

and then recommit to fix the linting issue?

vyom1611:
Hey Michael, just using those commands in the terminal does not do anything. Is there anything else I should install?

mcoughlin:
@vyom1611 You can see where it is failing here: https://github.com/skyportal/skyportal/runs/8276632758?check_suite_focus=true#step:13:66

usually the pre-commit then complains about a file.

guynir42 (approved):
LGTM.

PR #3461, opened 2022-09-11
DNM: Merge fritz's skyportal extension with skyportal

mcoughlin:
This PR is a proposal to merge fritz's skyportal extension files into skyportal. The point is that it is infeasible to maintain them separately currently, and therefore will greatly reduce the burden (as well as the effect of having multiple files with the same name).

To deal with this, we add a flag for kowalski.enabled and gloria.enabled in the config (defaulting to False), which turns off the features when un-needed / wanted, therefore keeping the current behavior while making it simple to turn them on.

Companion PR to https://github.com/fritz-marshal/fritz/pull/465


stefanv:
I'm curious how extensively this has been discussed? Keep SkyPortal generic, so it can be used by many different collaborations, has been a priority from the start. Has this changed?

mcoughlin:
@stefanv That's why it is a proposal. I want to discuss it on Tuesday. While the codebase becomes a bit less generic, I think user experience stays generic, which to make seems more important. But maybe some middle ground can be struck. But as I wrote above, what is happening now isn't working at all.

PR #3463, opened 2022-09-12
Legacy Survey DR9

mcoughlin:
This PR moves from Legacy Survey DR8 -> DR9.


alembic/versions/831be4c27aef_ls_dr9.py
```diff
@@ -0,0 +1,38 @@
...
+alter type "public"."thumbnail_types" rename to "thumbnail_types__old_version_to_be_dropped";
+
+create type "public"."thumbnail_types" as enum ('new', 'ref', 'sub', 'sdss', 'dr8', 'dr9', 'ps1', 'new_gz', 'ref_gz', 'sub_gz');
+
+alter table "public"."thumbnails" alter column type type "public"."thumbnail_types" using type::text::"public"."thumbnail_types";
+
+drop type "public"."thumbnail_types__old_version_to_be_dropped";
+"""
+    )
+    pass
```
guynir42, line 31:
```suggestion
```

skyportal/models/obj.py
```diff
@@ -406,8 +406,8 @@ def add_linked_thumbnails(self, session=DBSession):
         """Determine the URLs of the SDSS and DESI DR8 thumbnails of the object,
         insert them into the Thumbnails table, and link them to the object."""
         sdss_thumb = Thumbnail(obj=self, public_url=self.sdss_url, type='sdss')
-        dr8_thumb = Thumbnail(obj=self, public_url=self.desi_dr8_url, type='dr8')
-        session.add_all([sdss_thumb, dr8_thumb])
+        dr9_thumb = Thumbnail(obj=self, public_url=self.desi_dr9_url, type='dr9')
```
guynir42, line 409: Would it be smarter to name the types based on the underlying survey, rather than data release? Wouldn't DESI be a better name than dr9?

guynir42: I realize this is how it was until now, but this is actually a good example why naming things for DR's is not the best idea...

mcoughlin: great idea. we will rename it.

skyportal/models/obj.py
```diff
@@ -425,7 +425,7 @@ def sdss_url(self):
         )
 
     @property
-    def desi_dr9_url(self):
+    def legacysurvey_dr9_url(self):
         """Construct URL for public DESI DR8 cutout."""
```
guynir42, line 429: update docstring as well

guynir42 (approved):

PR #3467, opened 2022-09-13
Add optional plugins for the source and candidate page

mcoughlin:
This PR adds "plugins" for the source and candidate page to simplify extras for those pages (like in extensions).

### DIFF
--- a/static/js/components/CandidateList.jsx
+++ b/static/js/components/CandidateList.jsx
@@ -42,6 +42,8 @@ import Spinner from "./Spinner";
 import AddClassificationsScanningPage from "./AddClassificationsScanningPage";
 import Button from "./Button";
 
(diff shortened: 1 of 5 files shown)


stefanv:
How will these plugins we provided?

mcoughlin:
@stefanv As right now our extensions make files of the same name that are copied over, @guynir42 suggested this model where we do the same (but instead of having the whole file copied, just a "plugin" part).

guynir42 (approved):
Perfect, thanks!

stefanv:
Here are some other ideas you can play around with:

1. As we do with our messagehandlers, you add a Javascript method that a plugin can call to register itself.
2. Copy a *folder* of plugins, and to then use our templating to iterate over the files in the plugin folder and inject them one by one.  This will make the plugins easier to maintain.

PR #3472, opened 2022-09-15
Do not have repeated classifications

mcoughlin:
This PR presents the display of repeated classifications.


static/js/components/ClassificationList.jsx
```diff
@@ -116,6 +117,10 @@ const ClassificationList = () => {
       } else {
         taxname = "Unknown taxonomy";
       }
+      if (usedClassifications.includes(classification)) {
```
profjsb, line 120: It's ok to have the same classification if it's in a different (latest) taxonomy.

profjsb (changes requested):

static/js/components/ClassificationList.jsx
mcoughlin: Yeah but we wouldn't want to redisplay... right?

static/js/components/ShowClassification.jsx
```diff
@@ -37,57 +95,47 @@ export const getSortedClasses = (classifications) => {
...
-                className={classes.chip}
-              />
-            </Tooltip>
-          ));
-        })}
-      </div>
-    );
-  }
-  return <span />;
+  console.log("classificationsGrouped", classificationsGrouped);
```
guynir42, line 126:
```suggestion
```

guynir42 (approved):
There's one console that shouldn't be there but otherwise looks ok.

PR #3473, opened 2022-09-15
Observation source button

mcoughlin:
This PR changes the observation page to remove button for saving a source if there is not an associated target name.


guynir42 (approved):
LGTM

PR #3474, opened 2022-09-15
Public cone search

mcoughlin:
This PR adds a cone search for public ZTF data.


profjsb:
Wait -- why is kowalski token stuff happening in SP? Shouldn't this be in Fritz? Also, why aren't we doing cone search on our own (SP) DB? I dont understand the reason to go to a 3rd party for this... Please advise.

mcoughlin:
I think we draw the line between broker and catalog activities as the concept of streaming data into the app, at least this is what I understood. This is just a service where you can query catalogs and can post from them.

skyportal/handlers/api/catalog_services.py
```diff
@@ -205,6 +206,43 @@ def fetch_transients(allocation_id, user_id, group_ids, payload):
...
+                raise ValueError('Expected an Instrument named ZTF')
+
+            sources = query_fink(
+                jd_trigger, ra_center, dec_center, max_days=dt, within_days=dt
+            )
+
+            obj_ids = []
+            for source in sources:
+                df = source['data']
+                del source['data']
```
guynir42, line 224:
```suggestion
                df = source.pop('data')
```

skyportal/utils/catalog.py
```diff
@@ -370,3 +372,143 @@ def query_kowalski(
...
+
+    drop_columns = list(
+        set(df.columns.values)
+        - {'mjd', 'ra', 'dec', 'mag', 'magerr', 'limiting_mag', 'filter'}
+    )
+
+    df.drop(
+        columns=drop_columns,
+        inplace=True,
+    )
```
guynir42, line 511: Would it be easier to just do 
`df = df[['mjd', 'ra', 'dec', 'mag', 'magerr', 'limiting_mag', 'filter']]`

guynir42 (approved):
Minor comments:

skyportal/utils/catalog.py
mcoughlin: Interesting! I think you had suggested this version for previous ones so we have this form elsewhere. What do you think?

guynir42: I don't remember that, and I don't think there's a difference in runtime or anything like that... 
In this case it looks like you are doing double-negatives and it takes 8 lines instead of one, so the choice in this case is fairly obvious to me.

PR #3476, opened 2022-09-15
Create directory if missing

mcoughlin:
This PR creates a directory for the pdl client if none exists.

### DIFF
--- a/services/pdl_service/pdl_service.py
+++ b/services/pdl_service/pdl_service.py
@@ -24,8 +24,7 @@ def service():
             user_id = 1
             path = "services/pdlclient_service/data/receiver_storage/origin"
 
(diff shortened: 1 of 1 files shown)


services/pdl_service/pdl_service.py
```diff
@@ -22,6 +23,10 @@ def service():
         try:
             user_id = 1
             path = "services/pdlclient_service/data/receiver_storage/origin"
+
+            if not os.path.isdir(path):
+                os.makedirs(path)
```
stefanv, line 28:
```suggestion
            os.makedirs(path, exist_ok=True)
```

guynir42 (approved):
LGTM. I hope this solves the issues I've had locally.

Issue #3398, opened 2022-08-29
Displaying subclasses for slider classification


bfhealy:
On group source pages, the dropdown accessed to the left of each source id shows additional information and allows quick classification of the source using sliders to assign probabilities. Especially for the stellar variable taxonomy branch, there are many subclasses that are all displayed each time a user opens that classification dropdown.

Users could spend less time scrolling through these subclasses, many of which will not be pertinent to a given source, with a modification to the way the subclasses are displayed. One solution without adding more dropdown menus (extra clicking) is to only display the children of a parent class if the parent has a nonzero probability: for an unclassified source, using the current taxonomy, the user might only see Cataclysmic - Eclipsing - Rotating - Eruptive - Pulsating - microlensing under Stellar variable.  The subclasses would become visible if one or more of these sliders is moved to a nonzero probability.

See #2366 for another discussion about the slider classification system.

jadalilleboe:
@vyom1611 this would be a good one for you

PR #3478, opened 2022-09-16
Better sources subclasses slider information classifications ( #3398 )

vyom1611:
The subclasses sliders of a classification would only be visible if the value of the parent slider is greater than 0. This has been implemented for all subclasses for any classification. But, on initial render, the change is not in effect. If you click on the slider label and if the value is 0, then the subclasses get hidden, and if you change the slider again, subclasses come back. But you still need to click the label on 0 for every classification for the first time. Is there any fix for the initial render?


jadalilleboe:
Hi Vyom, adding this condition seemed to have the intended effect:
`classification.class in formState[selectedTaxonomy.id]`
so then the whole condition would be
```
{classification.class in formState[selectedTaxonomy.id] && formState[selectedTaxonomy.id][classification.class]
              ?.probability !== 0 &&
              renderSliders(
                classification.subclasses,
                depth + 1,
                [classification.class].concat(path)
              )}
```
Although I did find some weird behavior, if I change a classification to have a non-zero probability, submit the form, and then reload the page, the probability is saved like normal. But if I change it back to 0 and submit the form, it is not updated correctly after reloading the page. i.e. if I change Galactic Nuclei to have 0.25 probability and then try to change it back to 0, it stays at 0.25. Is that a bug or is it supposed to be like that @mcoughlin ?

mcoughlin:
@jadalilleboe that sounds like a bug to me.

jadalilleboe:
ok, I will open a new issue.

vyom1611:
Hey @jadalilleboe, can you please tell me which line number do I need to add the above condition, in file MultipleClassificationsForm.jsx?

jadalilleboe:
@vyom1611 The same lines where you added the probability check, check if the key exists as well. Let me know if that helps

vyom1611:
Yes, this fix seems to work!

jadalilleboe (approved):

PR #3490, opened 2022-09-20
Add more logging for atlas

mcoughlin:
This PR adds more info returned for atlas queries.

### DIFF
--- a/skyportal/facility_apis/atlas.py
+++ b/skyportal/facility_apis/atlas.py
@@ -246,6 +246,9 @@ class ATLASAPI(FollowUpAPI):
 
         r = requests.post(facility_microservice_url, json=request_body)
         log(f'Response for request {request.id}: {r.text}')
+        data_out = r.json()
+        if 'message' in data_out:
+            request.status = data_out['message']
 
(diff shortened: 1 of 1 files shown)


skyportal/facility_apis/atlas.py
```diff
@@ -246,6 +246,9 @@ def get(request, session):
 
         r = requests.post(facility_microservice_url, json=request_body)
         log(f'Response for request {request.id}: {r.text}')
+        data_out = r.json()
+        if 'message' in data_out:
+            request.status = data_out['message']
```
guynir42, line 251: This seems wrong, usually the status is supposed to be 200 or 400, etc.

mcoughlin: For whatever reason, we have used status as the text.

stefanv: What is this request object that is being passed in? It would be strange to assign to the request if it were a web request so I presume it must be something else.

stefanv (approved):

guynir42 (approved):
OK then.

PR #3498, opened 2022-09-22
Query telescopes based on latitude and longitude

mcoughlin:
This PR makes it possible to query telescopes based on latitude and longitude.


guynir42 (approved):
LGTM

Issue #3163, opened 2022-07-01
Photometry plot failure


mcoughlin:
Traceback (most recent call last):
  File "/skyportal_env/lib/python3.8/site-packages/tornado/web.py", line 1702, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/skyportal/baselayer/app/access.py", line 56, in wrapper
    return tornado.web.authenticated(method)(self, *args, **kwargs)
  File "/skyportal_env/lib/python3.8/site-packages/tornado/web.py", line 3173, in wrapper
    return method(self, *args, **kwargs)
  File "/skyportal/skyportal/handlers/api/internal/plot.py", line 40, in get
    self.success(data={'bokehJSON': json, 'url': self.request.uri})
  File "/skyportal/skyportal/handlers/base.py", line 18, in success
    super().success(*args, **kwargs, extra={'version': __version__})
  File "/skyportal/baselayer/app/handlers/base.py", line 355, in success
    self.write(to_json({"status": "success", "data": data, **extra}))
  File "/skyportal/baselayer/app/json_util.py", line 53, in to_json
    return json.dumps(obj, cls=Encoder, indent=2, ignore_nan=True)
  File "/skyportal_env/lib/python3.8/site-packages/simplejson/__init__.py", line 398, in dumps
    return cls(
  File "/skyportal_env/lib/python3.8/site-packages/simplejson/encoder.py", line 298, in encode
    chunks = list(chunks) 
  File "/skyportal_env/lib/python3.8/site-packages/simplejson/encoder.py", line 704, in _iterencode
    for chunk in _iterencode_dict(o, _current_indent_level):
  File "/skyportal_env/lib/python3.8/site-packages/simplejson/encoder.py", line 658, in _iterencode_dict
    for chunk in chunks:
  File "/skyportal_env/lib/python3.8/site-packages/simplejson/encoder.py", line 658, in _iterencode_dict
    for chunk in chunks:
  File "/skyportal_env/lib/python3.8/site-packages/simplejson/encoder.py", line 724, in _iterencode
    o = _default(o) 
  File "/skyportal/baselayer/app/json_util.py", line 49, in default
    return json.JSONEncoder.default(self, o)
  File "/skyportal_env/lib/python3.8/site-packages/simplejson/encoder.py", line 272, in default
    raise TypeError('Object of type %s is not JSON serializable' %
TypeError: Object of type Tabs is not JSON serializable

[...]

PR #3501, opened 2022-09-24
Catch photometry plot errors

jadalilleboe:
Catch errors fetching plot data and display an error message to the user.
Partially addresses #3163
@mcoughlin @guynir42 please advise if there should be any changes to the photometry plot code so that the badsource data does not cause an exception to be thrown


jadalilleboe:
Screen Shot 2022-09-24 at 2 27 55 PM

mcoughlin:
@jadalilleboe I think this is definitely a step in the right direction. What do you think @guynir42 ?

guynir42 (approved):
Definitely better than just endless circle or crash the page.

PR #3503, opened 2022-09-25
Smoothing and annotations from periods

mcoughlin:
This PR adds light curve smoothing (analogous to what is available for spectroscopy) and annotation posting from periods.

### DIFF
--- a/static/js/plotjs/foldphase.js
+++ b/static/js/plotjs/foldphase.js
@@ -1,3 +1,35 @@
+/* eslint no-undef: "off" */
+const smoothing_func = (values, window_size) => {
+  if (values === undefined || values === null) {
+    return null;
+  }
+  const output = new Array(values.length).fill(0);
+  const under = parseInt((window_size + 1) / 2, 10) - 1;
+  const over = parseInt(window_size / 2, 10);
+
+  for (let i = 0; i < values.length; i += 1) {
+    const idx_low = i - under >= 0 ? i - under : 0;
+    const idx_high = i + over < values.length ? i + over : values.length - 1;
+    let N = 0;
+    for (let j = idx_low; j <= idx_high; j += 1) {
+      if (Number.isNaN(values[j]) === false) {
+        N += 1;
+        output[i] += values[j];
+      }
+    }
+    output[i] /= N;
+  }
+  return output;
+};
+
+// callback inputs: model_dict, n_labels, checkbox, input, slider
+slider.value = input.value;
+
+let binsize = 0;
+if (0 in checkbox.active) {
+  binsize = input.value;
+}
+
 /* eslint-disable */
 if (numphases.active == 1) {
   /* two phases */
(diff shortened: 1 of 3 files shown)


static/js/plotjs/foldphase.js
```diff
@@ -18,6 +50,76 @@ for (let i = 0; i < n_labels; i++) {
...
+    for (j = 0; j < alist.length; j++) {
+      amag_sort[j] = alist[j].mag;
+    }
+    const amag_sort_smooth = smoothing_func(amag_sort, binsize);
+    const amag_sort_smooth_reordered = [];
+    for (var j = 0; j < folda.data.mjd_folda.length; j++) {
+      for (var k = 0; k < folda.data.mjd_folda.length; k++) {
+        if (j == alist[k].index) {
+          amag_sort_smooth_reordered.push(amag_sort_smooth[k]);
+          foldaerr.data.ys[j] = [amag_sort_smooth[k], amag_sort_smooth[k]];
```
guynir42, line 83: What is happening here?

static/js/plotjs/foldphase.js
```diff
@@ -18,6 +50,76 @@ for (let i = 0; i < n_labels; i++) {
...
+        : 1;
+    });
+    const amag_sort = [];
+    for (j = 0; j < alist.length; j++) {
+      amag_sort[j] = alist[j].mag;
+    }
+    const amag_sort_smooth = smoothing_func(amag_sort, binsize);
+    const amag_sort_smooth_reordered = [];
+    for (var j = 0; j < folda.data.mjd_folda.length; j++) {
+      for (var k = 0; k < folda.data.mjd_folda.length; k++) {
```
guynir42, line 80: There's no way to sort a list of tuples in JS? this double loop can't be good for you...

static/js/plotjs/foldphase.js
```diff
@@ -18,6 +50,76 @@ for (let i = 0; i < n_labels; i++) {
     foldb.data.mjd_foldb[m] = folda.data.mjd_folda[m] + 1;
     foldberr.data.xs[m] = [foldb.data.mjd_foldb[m], foldb.data.mjd_foldb[m]];
   }
+
+  if (binsize > 1) {
+    var j;
```
guynir42, line 55: What if you push the whole sorting mechanism into the smoothing function and then you'll just call it for 'a' and 'b' separately?

static/js/plotjs/foldphase.js
mcoughlin: The y-array is just the same value repeated twice... for some reason. Figured it out in console.log.

static/js/plotjs/foldphase.js
mcoughlin: You ask me questions like I know JS ;)

static/js/plotjs/foldphase.js
mcoughlin: I think the main issue here is in the variable names

guynir42: What do you mean? you are doing the exact same logic (sorting then smoothing) on two arrays. 
Why not put each array into a function that returns the sorted+smoothed values?

mcoughlin: Ah got ya. Ok implemented.

static/js/plotjs/foldphase.js
```diff
@@ -22,6 +22,32 @@ const smoothing_func = (values, window_size) => {
...
+  for (l = 0; l < list.length; l += 1) {
+    for (k = 0; k < list.length; k += 1) {
+      if (l === list[k].index) {
+        mag_sort_smooth_reordered.push(mag_sort_smooth[k]);
+        break;
+      }
+    }
+  }
+
+  return mag_sort_smooth_reordered;
```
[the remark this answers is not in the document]
guynir42: Please make sure to test this is correct, my JS is not the best either...

PR #3505, opened 2022-09-27
mount new persistent data directory for analysis results & fixup Docker builds

profjsb:
* Update `.dockerignore` so we do not send more data into a new docker image than is needed
* Update `Dockerfile` to point to the newly mounted `/skyportal/persistentdata` directory
* Update `config.yaml.defaults` for the analysis data to be written to  `/skyportal/persistentdata`
* Make sure that `docker-compose.yaml` is up to date.

To make SP docker images
```bash
make docker-local
```
(On ARM architectures on Mac you may need to run `export DOCKER_DEFAULT_PLATFORM=linux/amd64` first)

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -7,6 +7,9 @@ services:
       - "9000:5000"
       - "9001:5001"
     volumes:
+      # allow the data files to be mounted across containers
+      # if need be with the :z flag
+      # cf. https://stackoverflow.com/a/35222815
       - thumbnails:/skyportal/static/thumbnails:z
       - persistentdata:/skyportal/persistentdata:z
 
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -7,19 +7,22 @@ services:
       - "9000:5000"
       - "9001:5001"
     volumes:
-      - thumbnails:/skyportal/static/thumbnails
+      - thumbnails:/skyportal/static/thumbnails:z
```
mcoughlin, line 10: Maybe add a content that :z is relevant for all shared content?

mcoughlin (approved):
Looks like nice quality of life improvements here.

mcoughlin:
@Theodlz Flagging this for you given your attachments PR.

docker-compose.yaml
profjsb: done

PR #3508, opened 2022-09-28
set group_id for GCN

mcoughlin:
This PR allows one to set a group_id for GCN which will remember the last entry committed from the kafka stream.

### DIFF
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -56,12 +56,12 @@ def service():
         log('No notice_types configured to poll gcn events (config: gcn.notice_types')
         return
 
-    group_id = cfg['gcn.group_id']
-    if group_id is None or group_id == '':
-        group_id = str(uuid.uuid4())
+    client_group_id = cfg.get('gcn.client_group_id')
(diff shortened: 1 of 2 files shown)


services/gcn_service/gcn_service.py
```diff
@@ -53,9 +55,22 @@ def service():
     if notice_types is None or notice_types == '' or notice_types == []:
         log('No notice_types configured to poll gcn events (config: gcn.notice_types')
         return
+
+    group_id = cfg['gcn.group_id']
```
guynir42, line 59: would `get('gcn.group_id')` be better here?

guynir42 (commented):
Why does it need to be named `group_id`? This is a little confusing as this has nothing to do with skyportal groups (if I understand correctly). Is this because the kafka stream expects the config to have a "group_id"?

services/gcn_service/gcn_service.py
mcoughlin: @guynir42 Good point. It is called group.id in kafka, but I am renaming it as client_group_id.

guynir42 (approved):

PR #3509, opened 2022-09-28
Remove indexer to avoid downloading shakemap

mcoughlin:
This PR simplifies the EQ config to not download unnecessary files.


guynir42 (approved):
I mean, if you are sure you don't need all of this, then go ahead.

PR #3510, opened 2022-09-28
Add shifts in either local or UTC time

mcoughlin:
This PR makes it possible to add shifts in either local or UTC time.


guynir42 (approved):
LGTM. Please make sure you manually test all the combinations of inputs just to make sure nothing shakes out.

PR #3513, opened 2022-09-29
Profile picture

mcoughlin:
This PR enables you to upload a profile picture.

### DIFF
--- a/skyportal/models/user_token.py
+++ b/skyportal/models/user_token.py
@@ -29,13 +29,15 @@ def basic_user_display_info(user):
         field: getattr(user, field)
         for field in ('username', 'first_name', 'last_name', 'gravatar_url')
     }
-    if 'avatar' in user.preferences:
+    if getattr(user, 'preferences') and ('avatar' in user.preferences):
         user_info['avatar'] = user.preferences['avatar']
+    else:
+        user_info['avatar'] = None
     return user_info
 
 
 def user_avatar(user):
-    if 'avatar' in user.preferences:
+    if getattr(user, 'preferences') and ('avatar' in user.preferences):
         return user.preferences['avatar']
     else:
         return None
--- a/skyportal/models/user_token.py
+++ b/skyportal/models/user_token.py
@@ -25,10 +25,20 @@ from .invitation import Invitation
 
 
 def basic_user_display_info(user):
-    return {
+    user_info = {
         field: getattr(user, field)
         for field in ('username', 'first_name', 'last_name', 'gravatar_url')
     }
+    if 'avatar' in user.preferences:
+        user_info['avatar'] = user.preferences['avatar']
+    return user_info
+
+
+def user_avatar(user):
+    if 'avatar' in user.preferences:
+        return user.preferences['avatar']
+    else:
+        return None
 
 
 def user_to_dict(self):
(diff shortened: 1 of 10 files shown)


profjsb (changes requested):
We need to insure that the picture is a reasonable size and format. I dont think we should be allowing arbitrary uploads since avatar pictures are used everywhere and images can be large without proper checks.

skyportal/models/user_token.py
```diff
@@ -25,10 +25,22 @@
...
 def basic_user_display_info(user):
-    return {
+    user_info = {
         field: getattr(user, field)
         for field in ('username', 'first_name', 'last_name', 'gravatar_url')
     }
+    if getattr(user, 'preferences') and ('avatar' in user.preferences):
+        user_info['avatar'] = user.preferences['avatar']
+    else:
+        user_info['avatar'] = None
```
guynir42, line 35:
```suggestion
    user_info['avatar'] = getattr(user, 'preferences', {}).get('avatar', None)
```

skyportal/models/user_token.py
```diff
@@ -25,10 +25,22 @@
...
+    return user_info
+
+
+def user_avatar(user):
+    if getattr(user, 'preferences') and ('avatar' in user.preferences):
+        return user.preferences['avatar']
+    else:
+        return None
```
guynir42, line 43: same as above. Actually I would just replace the code above with a call to `user_avatar`.

static/js/components/CompactCommentList.jsx
```diff
@@ -71,6 +71,7 @@ const CompactCommentList = ({
           lastName={author.last_name}
           username={author.username}
           gravatarUrl={author.gravatar_url}
+          avatar={author?.avatar}
```
guynir42, line 74: No sure if the question mark makes sense here. If there is no `author` it would crash on the lines right above it.

guynir42 (approved):
Some comments but overall looks good.

static/js/components/CompactCommentList.jsx
mcoughlin: good point.

static/js/components/NewsFeed.jsx
```diff
@@ -111,6 +111,7 @@ const NewsFeedItem = ({ item }) => {
           lastName={item.author_info.last_name}
           username={item.author_info.username}
           gravatarUrl={item.author_info.gravatar_url}
+          avatar={item.author_info?.avatar}
```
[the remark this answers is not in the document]
mcoughlin: yes

skyportal/models/user_token.py
mcoughlin: ah yes.

skyportal/models/user_token.py
mcoughlin: nifty.

mcoughlin:
Good idea. Updated the check @profjsb

profjsb:
Are we sure we want to do this? gravatar.com manages the avatars we use as the only thing they do. I'm a bit worried about rescaling and storage issues.

mcoughlin:
@profjsb @stefanv Personally I much prefer to be able to cycle through photos, and the gravatar thing implies the same thing everywhere, which I very much dislike.

stefanv:
Gravatar supports multiple identities, fwiw.

Issue #3269, opened 2022-07-26
Default time range for scanning is unclear


robertdstein:
The default times on the candidate scanning page appear to be "from 24 hours ago to now". An example:

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/14312838/181116997-0ec73493-0680-4fce-a7f0-49fcf099e538.png">

While that all looks reasonable, this screenshot comes from my laptop in California, not in a UTC timezone. The scanning page very explicitly states that the times are in UTC.

So either:

- The default end time is not "the time now, in my timezone" but instead "the time now, in the UTC timezone". Obviously that would be very bad/confusing default behaviour.
- OR the default is indeed "Now", in which case the labels on the time boxes are wrong. In that case, we should fix the labels.

jadalilleboe:
@vyom1611 this would be a good one as well

vyom1611:
Would it be better to add a Time zone label for your current Time zone or just remove the "UTC" ?

PR #3520, opened 2022-09-29
Removing "(UTC)" time label from Candidates page

vyom1611:
Removing the "UTC" label from Candidate form, based on issue #3269


mcoughlin:
@vyom1611 i would explicitly say it is local time in the tag

vyom1611:
@mcoughlin added "Local Time" label as suggested.

PR #3522, opened 2022-09-30
Enable plotting with reference magnitudes / fluxes

### LINKED ISSUE #3515: Plotting the total flux/mag
We've recently added a `tot_flux` and `magtot` to the photometry data. 
This would be useful for e.g., variable stars where you want to see the lightcurve of the star's full light, not just the subtraction. 

The handler now lets you input the `ref_flux` or `magref` and when they exist, the handler also outputs `tot_flux` and `magtot` as the addition of the regular flux and the reference flux. 

We want to be able to show this on the photometry plots, at least for sources that have one or more points that have a reference flux. 

I think the best way is to add a button to the Bokeh plot that also ties into a user preference (such that after checking the box it will remain checked for all plots). When checked, it will display the total flux/mag for any sources that have them. 

If some of the photometry points don't have a reference flux they should not appear on this plot, with the expectation that most sources where this is interesting will have reference fluxes for all points (and it would likely be the same reference). 
But that really is up to the users/admins to make sure that data exists. 

I think that it would also be good to have the same behavior on the little vega plots, linking to the same preference in the main bokeh plot. My thinking is that variable star people would always want to see the total flux and transient people will always want to see the subtraction flux. We could make it more complicated if need be later on if there's a requirement from users.

mcoughlin:
This PR enables plotting with reference magnitudes / fluxes.

Closes #3515 .


mcoughlin:
@jadalilleboe the reference flux I added to that file was pretty small, so the effect is subtle for this example. For "real" objects it will be larger for some of them at least. You can confirm by changing that 100 to 1000 or something in the file.

jadalilleboe:
ok, LGTM

data/ZTFrlh6cyjh_phot.csv
```diff
@@ -1,1122 +1,1122 @@
...
-ztfi,58761.25186000019,31506.488329723954,25.0,ab,319.2040309855778,0.25,0.22,342.4596252,51.2633849
-ztfi,58760.24319999991,31102.83668685142,25.0,ab,315.114485042836,0.25,0.22,342.4596298,51.2633971
-ztfi,58745.242999999784,31622.776601683792,25.0,ab,320.38219101336193,0.25,0.22,342.4596214,51.26338
-ztfi,58750.2014299999,31188.895840939353,25.0,ab,315.98638255805866,0.25,0.22,342.459654,51.2634081
-ztfi,58751.199080000166,31017.0149947769,25.0,ab,314.24499334418977,0.25,0.22,342.45962,51.263391
-ztfi,58749.20110000018,30874.504755874485,25.0,ab,312.8011687504012,0.25,0.22,342.4596203,51.2633911
-ztfi,58754.251600000076,30394.84968770113,25.0,ab,307.9416036461826,0.25,0.22,342.45963,51.2633985
-ztfi,58748.2661799998,14777.46706171993,25.0,ab,149.71605227763357,0.25,0.22,342.4596264,51.2633981
+filter,mjd,flux,zp,magsys,fluxerr,ra_unc,dec_unc,ra,dec,ref_flux,ref_fluxerr
+ztfg,58744.20135999983,30988.46042012061,25.0,ab,371.0385501207088,0.25,0.22,342.4596227,51.2633822,100.0,0.1
```
guynir42, line 2: I know it is a pain but would be nice to have a bigger ref flux to make it show up in the plot.

static/js/components/UIPreferences.jsx
```diff
@@ -77,6 +85,14 @@ const UIPreferences = () => {
...
   );
 
+  const useRefMagSwitch = (
+    <Switch
+      value="Plot Total Magnitude / Flux (including reference)"
+      checked={useRefMag}
+      onChange={useRefMagToggled}
+    />
+  );
+
```
guynir42, line 95: This is fine but wouldn't it be even more useful if there was a button in the photometry plot to toggle this on/off?

guynir42 (commented):
I wonder if we can connect this to the bokeh plot directly.

static/js/components/UIPreferences.jsx
mcoughlin: Sure but the bokeh stuff is a nightmare. It's not worth the dev time.

data/ZTFrlh6cyjh_phot.csv
mcoughlin: I made it larger.

static/js/components/UIPreferences.jsx
guynir42: OK, I say we go for it as is and fix it if needed later.

guynir42 (approved):

PR #3530, opened 2022-09-30
Delete icons

mcoughlin:
This PR uses delete icons instead of x's.

### DIFF
--- a/skyportal/tests/frontend/test_reminders.py
+++ b/skyportal/tests/frontend/test_reminders.py
@@ -92,7 +92,15 @@ def post_and_verify_reminder_frontend(driver, reminder_text):
     )
     reminder_text_2 = str(uuid.uuid4())
     driver.wait_for_xpath('//*[@id="root_text"]').send_keys(reminder_text_2)
-    next_reminder = (datetime.now() + timedelta(days=1)).strftime("%m/%d/%YT%I:%M %p")
+    first_of_the_month = int((datetime.now() + timedelta(days=1)).strftime("%d")) == 1
+    if first_of_the_month:
+        next_reminder = (datetime.now() + timedelta(days=14)).strftime(
+            "%m/%d/%YT%I:%M %p"
+        )
+    else:
+        next_reminder = (datetime.now() + timedelta(days=1)).strftime(
+            "%m/%d/%YT%I:%M %p"
+        )
     driver.wait_for_xpath('//*[@id="root_next_reminder"]').send_keys(
         next_reminder[0:11]
     )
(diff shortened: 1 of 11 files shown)


stefanv:
Confirming there are confirmations too?

mcoughlin:
@stefanv Confirmed.

static/js/components/AllocationPage.jsx
```diff
@@ -174,11 +192,25 @@ const AllocationList = ({ allocations, deletePermission }) => {
...
+              </DialogContent>
+              <DialogActions>
+                <Button secondary autoFocus onClick={closeDialog}>
+                  Dismiss
+                </Button>
+                <Button primary onClick={() => deleteAllocation()}>
+                  Confirm
+                </Button>
+              </DialogActions>
+            </Dialog>
```
jadalilleboe, line 213: There's a lot of duplicated code for these dialogs
What about a new component?

mcoughlin: @jadalilleboe How would you imagine refactoring?

jadalilleboe: I would imagine a ConfirmDeletionDialog component which takes the delete function,  dialog state, and name of the resource as arguments

mcoughlin: @jadalilleboe Ok see what you think of the new component? If it makes sense, I can add it into the others.

jadalilleboe: @mcoughlin I like it!

jadalilleboe (approved):

skyportal/tests/frontend/test_reminders.py
```diff
@@ -92,7 +92,15 @@ def post_and_verify_reminder_frontend(driver, reminder_text):
...
-    next_reminder = (datetime.now() + timedelta(days=1)).strftime("%m/%d/%YT%I:%M %p")
+    first_of_the_month = int((datetime.now() + timedelta(days=1)).strftime("%d")) == 1
+    if first_of_the_month:
+        next_reminder = (datetime.now() + timedelta(days=14)).strftime(
+            "%m/%d/%YT%I:%M %p"
+        )
+    else:
+        next_reminder = (datetime.now() + timedelta(days=1)).strftime(
+            "%m/%d/%YT%I:%M %p"
+        )
```
guynir42, line 103: I don't understand why this is needed.

guynir42 (approved):
minor question:

skyportal/tests/frontend/test_reminders.py
mcoughlin: because it defaults to 01 and wont input correctly if you put in a 01 again. It's super weird (and totally artificial as a selenium thing)

stefanv:
Nice suggestion @jadalilleboe, thanks for implementing @mcoughlin!

Issue #3512, opened 2022-09-29
Refactor Thumbnail Directory


profjsb:
The thumbnails are all written into a single directory with no sub-directories. This makes it impossible to list as there are millions of thumbs in one directory.

Suggest writing thumbnails into a structure like:

```pre
thumbnails/
  hash_subdir_1/
      hash_subsubdir_1a/
      ...
      hash_subsubdir_1m/
  hash_subdir_2/
   ... 
  hash_subdir_n/
```
where m and n are large (like 256). 

So if the source name is: ZTFrlh6cyjh and the sha256 hash of ZTFrlh6cyjh is 8e4db823338a42dbb965623be53dc4af. Then the files would be written into:
   thumbnails/8e/4d/ZTFrlh6cyjh_new.png

Could explore the use of https://pypi.org/project/hashfs/ for this or just code it as above using `hashlib`

PR #3534, opened 2022-09-30
use hash of obj id to make subfolders for thumbnail storage

guynir42:
Calculates the sha256 hash of the object ID, then uses the first two characters to make the subfolder and the next two characters to make the sub-subfolder. 

This will split up the thumbnail storage into 256 subfolders with 256 sub-subfolders. 

This way, never get too many files or folders in any level of the filesystem. 

Since the folders are based on the obj_id, there's a (relatively) easy way to find the data even without having the thumbnail row in the DB pointing back to it.


stefanv:
Is there a concern that the directories will fill up? Just trying to understand the rationale of the change.

guynir42:
Right now we have millions of files in one folder. You cannot `ls` that folder because it is so big. 
In the new format each folder will have a small fraction of the total files, so you'd always get a reasonable result from `ls`.

guynir42:
Addresses #3512

stefanv (approved):
Makes sense! Not sure you need two levels, since you can just go with N characters on the first level. Any advantage to having it two deep?

profjsb:
Having 2 levels deep will mean we can actually ls each sub dir. hashing on 4 letters will mean the first subdir could have as much as 2,147,483,647 directories. That doesn't help.

profjsb:
We need a strategy for migrating existing postage stamps to the new directory structure. Not sure if these needs to be here or not. But I believe the static location of thumbnails may be hardcoded in front end pieces so it's worth checking.

stefanv:
> Having 2 levels deep will mean we can actually ls each sub dir. hashing on 4 letters will mean the first subdir could have as much as 2,147,483,647 directories. That doesn't help.

No, we're using hex so 65535.

stefanv:
I suppose what you want to know is total thumbnails anticipated / max size desired.

profjsb:
Sorry, thought we were doing 16 bits per dir

profjsb:
> > Having 2 levels deep will mean we can actually ls each sub dir. hashing on 4 letters will mean the first subdir could have as much as 2,147,483,647 directories. That doesn't help.
> 
> No, we're using hex so 65535.

Actually, I confused here: 1 hex digit is 4 bits, 1 byte. The proposal is 2 byte directories, so two hex digits (possible combinations of 16**2 = 256).

profjsb (changes requested):
Need a way to handle the transition from current structure to new structure:
 1) How will old thumbnails be accessed?
 2) How will old thumbnails be moved to the new structure?

guynir42:
The easiest way would probably be to delete the thumbnails for a source and then re-post them. 
I'm going to test this on one source and see if that works as expected. 
Then we can just run a script to do this for all sources.

profjsb:
Rather than repost you could write a script in tools/ to move the files to the appropriate new directory? Or does the DB also store the location?

guynir42:
The DB also has it saved. The easiest way is to delete/repost in terms of develop time. Might be slow in terms of runtime but I'm guessing it wouldn't be that bad (comparable to phot stats).

stefanv:
Another option is to generate a custom one-off migration script. It should be fairly straightforward to use a script to write out the rename commands for each existing thumbnail?

guynir42:
@stefanv I am not sure how to do that. Are you volunteering to show us how this is done?

stefanv:
I should know better than to open my mouth ;)

Using a Python script, you could generate a text file of the form:

```
UPDATE thumbnails SET file_uri='newly_calculated_file_uri' WHERE id='thumbnail_id';
UPDATE thumbnails SET ...;
```

Then execute the file with all such commands using `psql -f filename.sql`.

The same Python script, you generate `rename.sh`:

```
mv old_thumb_file new_thumb_file
mv ...
```

and run that after the SQL.

profjsb:
Why wouldn't we just write a Python migration script that does something like pseudocode:

```python
ids_with_thumbs = get_all_ids(("sub", "new", "ref"))

# loop over chunks
for chunk in ids_with_thumbs:
   for id in chunk:
      if id not in migrated:
        for suffix in [("sub", "new", "ref")]:
           connect.execute("update thumbnails where obj_id = {id} set file_url = {...} set public_url = {...}")
           try: 
               os.mv(old_file_location, new_file_location)
               save_id_to_migrated_file()
           except:
               ...
```

BTW, we current have 6M rows like this:

```SQL
db=> select count(*) from thumbnails where type in ('new', 'ref', 'sub');
  count  
---------
 6006545
```

stefanv:
That looks just as good.

stefanv:
Why all this work on a handler, instead of just putting the thumbnails where they need to go?

guynir42:
@profjsb I decided to do something similar to the PhotStats update handler. 
This allows us to incrementally move thumbnails to the proper structure, from the API. 
I like this better because I can call this in a loop instead of launching one script that may or may not work.

guynir42:
because
(a) any script I write may fail on some specific file, and then I'd have to start patching it, on production. 
(b) whatever you do to move things around today, you'd be happy to do again down the line when you decide you want a different folder structure.

profjsb (commented):
Looks good. I'm not sure how the paging would work here. If you call this with page_number = 1 then it will fix the first 100 found in the query. When you call with page_number = 2 then you're issuing a new query with those original 100 no longer there, so you effectively miss the first page of unfixed 100. Either you'd need to cache the query or just change `numPerPage` to `maxNumberToFix`.

I think you might as well try to redownload those files of size 0.  We know that they are there.

guynir42:
@profjsb Paging would work by simple re-sending `pageNumber=1`. Unless there's an error and you want to skip that batch. It worked well for `PhotStats`. 

I will try to add the zero size file check.

skyportal/handlers/api/thumbnail.py
```diff
@@ -253,3 +273,345 @@ def delete(self, thumbnail_id):
...
+        required_depth = self.get_query_argument('requiredDepth', 2)
+        try:
+            required_depth = int(required_depth)
+        except ValueError:
+            return self.error('requiredDepth must be an integer')
+
+        if required_depth < 0 or required_depth > 32:
+            return self.error('requiredDepth must be between 0 and 32')
+
+        good_like = f"%thumbnails{'/__' * required_depth}/%"
```
mcoughlin, line 343: Maybe a comment on what this is doing?

skyportal/handlers/api/thumbnail.py
```diff
@@ -253,3 +273,345 @@ def delete(self, thumbnail_id):
...
+        description: |
+          Get information on thumbnails that are
+          or are not in the correct folder/path.
+        tags:
+          - thumbnails
+        parameters:
+          - in: query
+            name: types
+            required: false
+            default: ['new', 'ref', 'sub']
```
mcoughlin, line 292: Maybe the default should be all allowed types?

skyportal/handlers/api/thumbnail.py
```diff
@@ -253,3 +273,345 @@ def delete(self, thumbnail_id):
...
+        description: |
+          Update the file path and file_uri of the database rows
+          of thumbnails that are not in the correct folder/path.
+        tags:
+          - thumbnails
+        parameters:
+          - in: query
+            name: types
+            required: false
+            default: ['new', 'ref', 'sub']
```
mcoughlin, line 387: Same comment as above.

skyportal/handlers/api/thumbnail.py
```diff
@@ -253,3 +273,345 @@ def delete(self, thumbnail_id):
...
+                    session.commit()
+                except Exception as e:
+                    session.rollback()
+                    # make sure to move the files back as well
+                    os.rename(new_file_uri, old_file_uri)
+                    return self.error(f'Could not update database row: {e}')
+
+                num_moved += 1
+
+            stmt = sa.select(Thumbnail).where(Thumbnail.type.in_(types))
```
mcoughlin, line 523: If this is the same syntax as the get, maybe it could be refactored to a single function?

mcoughlin (approved):
Looks reasonable, we can test during deploy.

skyportal/handlers/api/thumbnail.py
guynir42: The default is all types that we save on disk locally...

skyportal/handlers/api/thumbnail.py
guynir42: good idea
let me try that

skyportal/handlers/api/thumbnail.py
mcoughlin: Ah I didn't think about that some thumbnails are not local.

skyportal/handlers/api/thumbnail.py
```diff
@@ -1,16 +1,27 @@
...
 
+# this part will only work on fritz
+# (or any extension of skyportal with alerts)
+try:
+    from .alert import post_alert
+
+    alert_available = True
+except Exception:
+    alert_available = False
+
```
profjsb, line 24:
```suggestion
from .alert import post_alert, alert_available
```

profjsb (commented):
Looks good. I think we need a path to remove the implicit kowalski dependence, however. My suggestion is to write into `skyportal/handler/api/alert.py` a dummy NOOP:

```
alert_available = False
def post_alert(*args, **kwargs):
    return None
```

Fritz extensions will overwrite this file. NB: We need to make a PR to that file on Fritz to add
`alert_available = True` 

You can then remove lines 16-24 in your PR of `thumbnail.py` (see below).

skyportal/tests/api/test_thumbnail.py
```diff
@@ -323,3 +324,119 @@ def test_delete_obj_deletes_thumbnail_file_on_disk(
...
+    )
+    assert_api(status, data)
+    thumbnail_id = data['data']['id']
+    assert isinstance(thumbnail_id, int)
+
+    status, data = api('GET', f'thumbnail/{thumbnail_id}', token=upload_data_token)
+    assert_api(status, data)
+    assert data['data']['type'] == ttype
+
+    # thumbnail = DBSession.query(Thumbnail).filter(Thumbnail.id == thumbnail_id).first()
```
profjsb, line 366:
```suggestion
```

profjsb (approved):
Looks good. Quick question before merging: are the tests non-destructive to a production instance? That is, if run will they interfere where the production thumbnails are stored? Eg. if 14ggr is a real source will this test remove its thumbs and delete its directory?

guynir42:
@profjsb in principle the tests _should_ put back the thumbnails to the original folders. The delete command _should_ only clear up empty folders, not remove any thumbnails. 
The tests will likely fail on production because the API calls only move 500 thumbnails at a time and would probably not move the new thumbnail posted.

profjsb:
Ok. Can the test config (`test_config.yaml`) be used to set a different location for the test thumbnails?

guynir42:
Isn't the test run on a different database? I don't think we move anything that doesn't have an associated row in the DB.

profjsb:
Yes, but you may be hashing the test images into directories with real data and then when you try to delete the (non empty) directory the test will fail.

guynir42:
@profjsb I don't think this would be an issue. 
1) The tests are run on github servers, not on a production (or even staging). So it seems to me like this will never come up. 
2) The thumbnails on the non-test server should stay at their current folder depth because we only move those in the test DB. The thumbnails in the test DB should move to depth 3 and back to depth 2. When I delete the depth 3 folders there would not be anything in there.

PR #3536, opened 2022-09-30
Turn on and off smoothing

mcoughlin:
This PR enables one to turn smoothing on and off on light curves.

### DIFF
--- a/static/js/plotjs/foldphase.js
+++ b/static/js/plotjs/foldphase.js
@@ -74,15 +74,14 @@ for (let i = 0; i < n_labels; i++) {
     foldberr.data.xs[m] = [foldb.data.mjd_foldb[m], foldb.data.mjd_foldb[m]];
   }
 
+  var j;
   if (binsize > 1) {
-    var j;
-
     var alist = [];
     for (j = 0; j < folda.data.mjd_folda.length; j++) {
       alist.push({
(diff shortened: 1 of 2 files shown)


static/js/plotjs/foldphase.js
```diff
@@ -95,13 +94,28 @@ for (let i = 0; i < n_labels; i++) {
...
       foldberr.data.ys[j] = [foldb.data.mag[j], foldb.data.mag[j]];
     }
+  } else {
+    folda.data.mag = folda.data.mag_unsmoothed;
+    for (j = 0; j < folda.data.mag_unsmoothed.length; j++) {
+      foldaerr.data.ys[j] = [
+        folda.data.mag_unsmoothed[j],
+        folda.data.mag_unsmoothed[j],
+      ];
+    }
```
zvanderbosch, line 111: Can reduce redundancy by setting the `foldaerr.data.ys` and `folberr.data.ys` values after the if/else statement is done. Regardless of smoothing or not, the 'ys' values can be set using the folda and foldb mag values, since those are being set within the if/else statement.

mcoughlin: @zvanderbosch Good point. fixed.

guynir42 (approved):

PR #3565, opened 2022-10-02
Thumbnails below photometry and spectra

mcoughlin:
This PR moves thumbnails below the photometry and spectroscopy panels, which makes sense given they are more important for understanding the transient.


guynir42 (approved):
This seems fine to me but I think you should run this by some more people on the dev team (and maybe on the beta testing channel?) to make sure they think this is a good idea.

mcoughlin:
This was never popular with the team.

Issue #3516, opened 2022-09-29
Duplicate rows in group source downloads


bfhealy:
When downloading group sources from Fritz using the frontend or api, the result sometimes contains duplicate rows (although always the correct number of total rows, so some sources are missing). I have not encountered this problem for smaller groups (<~ 1000 sources), but for larger groups I get a variable number of duplicates each time I download. Some additional info:
- There are no duplicates on any single page; they instead appear on separate pages.
- The number of duplicates increases for a large ratio of total sources to number per page.
- I have been unable to replicate this behavior on my local SkyPortal instance, including in scenarios where sources are members of multiple groups or other sources are in the process of being posted.

It is possible that modifying this call of `grab_query_results`:
https://github.com/skyportal/skyportal/blob/91a293fada374b87d7285294199dfbfbf7c25ef2/skyportal/handlers/api/source.py#L1026

by setting `use_cache=True`, see
https://github.com/skyportal/skyportal/blob/03f7c79afbe0bedde859ff1a94355f9644da7298/skyportal/handlers/api/candidate.py#L1234

will solve the problem. We'd also have to decide on how long to maintain the cache.

mcoughlin:
@guynir42 What do you think of trying this and seeing how it goes this week?

guynir42:
You mean adding `use_cache=True` to the download CSV option?

mcoughlin:
@guynir42 yes.

bfhealy:
Unfortunately PR #3569 has not resolved the duplicate row issue with Fritz. I am still unable to replicate the issue locally.

bfhealy:
Testing on a group of ~10,000 sources in preview did not produce any duplicates. It seems this issue is specific to fritz.science.

PR #3569, opened 2022-10-04
Use cache in source.py grab_query_results

bfhealy:
This PR attempts to address #3516 by setting `use_cache=True` in the call to `grab_query_results` in source.py. While I have not been able to replicate the duplicate source download issue in my local instance of SkyPortal, I have tested this change locally and downloads continue to work as expected.


mcoughlin (approved):
As discussed on the call today, we will try this out.

PR #3570, opened 2022-10-04
Use public ZTFRef stars if Gaia servers are down (after retry)

profjsb:
This PR catches errors when the Gaia and backup Gaia DBs are down. It also retries the query just in case it's a temporary issue.

### DIFF
--- a/skyportal/utils/offset.py
+++ b/skyportal/utils/offset.py
@@ -5,6 +5,7 @@ import datetime
 import warnings
 from functools import wraps
 import re
+import time
 
 import pandas as pd
 import requests
@@ -31,7 +32,7 @@ from astropy.io import fits
 from astropy.visualization import ImageNormalize, ZScaleInterval
 from reproject import reproject_adaptive
 import pyvo as vo
-from pyvo.dal.exceptions import DALQueryError
+from pyvo.dal.exceptions import DALQueryError, DALServiceError
 
 from .cache import Cache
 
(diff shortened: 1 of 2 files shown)


skyportal/utils/offset.py
```diff
@@ -751,8 +757,24 @@ def get_nearby_offset_stars(
                            {radius_degrees}))
                 """
 
-    g = GaiaQuery()
-    r = g.query(query_string)
+    try:
```
stefanv, line 760: PR looks good.

I'd probably combine this into a single try loop, but not a big deal.

```python
result = None
for retry in range(2):
    try:
        g = GaiaQuery()
        result = g.query(query_string)
    except Exception as e:
        log(f'Gaia failed [{retry}] ... other info')

if result is None:
    return ...
```

guynir42 (approved):
LGTM

skyportal/utils/offset.py
profjsb: Good idea. Implemented.

skyportal/utils/offset.py
```diff
@@ -378,8 +384,82 @@ def get_ztfref_url(ra, dec, imsize, *args, **kwargs):
 }
 
 
+def get_astrometry_backup_from_ztf(
```
profjsb, line 387: This was Mansi's idea.

guynir42 (approved):
Yeah that's pretty good.

stefanv (commented):
Looks good @profjsb. Would be good to have a test for the `get_astrometry_backup_from_ztf` function, if that's feasible.

mcoughlin (approved):
LGTM

stefanv (approved):

stefanv:
Nicely done, thank you @profjsb.

PR #3575, opened 2022-10-06
Clarify text upon photometry upload

mcoughlin:
This PR clarifies the text upon photometry upload, where it was said instrument_id in the upload file is for multiple instruments, where its really possible for one or more instrument.


stefanv:
I find the old one very confusing and the newer one mildly confusing. Can we be more explicit?

stefanv:
Specifically, what does "for one or more instrument" mean? How should users use that column.

mcoughlin:
@stefanv I.e. you can use different instrument_id's for different instruments if you want.

PR #3578, opened 2022-10-07
Add more token power

mcoughlin:
This PR adds ability to have multiple tokens posted if admin does the posting.

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -427,7 +427,7 @@ skyportal_handlers = [
         r'/api/webhook/(obj)_analysis/([0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})?',
         AnalysisWebhookHandler,
     ),
-    (r'/api/internal/tokens(/.*)?', TokenHandler),
+    (r'/api/internal/tokens(/[0-9A-Za-z-_\.\+]+)?', TokenHandler),
     (r"/api/internal/profile(/[0-9]+)?", ProfileHandler),
     (r'/api/internal/dbinfo', DBInfoHandler),
     (r'/api/internal/source_views(/.*)?', SourceViewsHandler),
--- a/skyportal/handlers/api/internal/token.py
+++ b/skyportal/handlers/api/internal/token.py
@@ -80,7 +80,7 @@ class TokenHandler(BaseHandler):
         single:
           description: Retrieve a token
           tags:
-            - telescopes
+            - tokens
           parameters:
             - in: path
               name: token_id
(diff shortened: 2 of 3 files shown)


skyportal/app_server.py
```diff
@@ -427,7 +427,7 @@ def log_request(self, handler):
         r'/api/webhook/(obj)_analysis/([0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})?',
         AnalysisWebhookHandler,
     ),
-    (r'/api/internal/tokens(/.*)?', TokenHandler),
+    (r'/api/internal/tokens(/[0-9A-Za-z-_\.\+]+)?', TokenHandler),
```
guynir42, line 430: Are you sure you need the dash before the underscore?

skyportal/handlers/api/internal/token.py
```diff
@@ -34,33 +32,108 @@ def post(self):
...
-        )
+
+    @auth_or_token
+    def get(self, token_id=None):
+        """
+        ---
+        single:
+          description: Retrieve a token
+          tags:
+            - telescopes
```
guynir42, line 83:
```suggestion
            - tokens
```

skyportal/handlers/api/internal/token.py
```diff
@@ -34,33 +32,108 @@ def post(self):
...
+              content:
+                application/json:
+                  schema: Error
+        multiple:
+          description: Retrieve all tokens
+          tags:
+            - tokens
+          parameters:
+            - in: query
+              name: userId
```
guynir42, line 105: I prefer `userID`.

guynir42 (approved):
Minor comments:

skyportal/handlers/api/internal/token.py
mcoughlin: Good call.

skyportal/app_server.py
mcoughlin: Yeah its broken without it.

guynir42: ok, just making sure

stefanv: So we limit the token characters to `0-9`, `A-Z`, `a-z`, `-`, `_`, `.`, and `+`. I didn't know we ever needed plusses or underscores?

mcoughlin: I was just following the convention of the IDs, and that seemed to work ok. Users can name their tokens as they like, so I assume we need to allow the flexibility.

stefanv: Odd, I thought we always generate the IDs ourselves (even though users can name the tokens)?  If the users can indeed choose the name that appears in the URL, we should probably add a hint to the frontend of what they are allowed to enter.

mcoughlin: Ah I see what you mean, because the ID and the name is different. Ok I take that back, so we can restrict the input somehow.

stefanv: I suspect we use UUID4, so  `0-9a-e-` should work.

PR #3579, opened 2022-10-07
remove slow (and unhelpful) DB stats queries

mcoughlin:
The queries on candidates times are slow on a large DB, and do not really add anything. This PR removes them.


guynir42 (approved):
This looks fine. I will guess that the slow part is getting the "unsaved" candidates, because the 
"created at" is indexed so it should be fairly fast.

Issue #2966, opened 2022-05-12
Bump to react-device-detect 2.2.2 breaks plots on mobile


jadalilleboe:
The bump to react-device-detect 2.2.2 makes the plots lose their responsiveness on mobile. I checked to make sure #2874 wasn't the cause, the plots looked like this before the bump:
<img width="1051" alt="Screen Shot 2022-05-12 at 10 44 12 AM" src="https://user-images.githubusercontent.com/82007190/168116783-911f22af-183e-4dd9-a99d-331ba18cf830.png">
And after the bump the responsiveness is lost:
<img width="810" alt="Screen Shot 2022-05-12 at 10 48 34 AM" src="https://user-images.githubusercontent.com/82007190/168116871-dda9ca64-95ce-463d-9c2f-40be1478313d.png">

I also think it would be good to add some frontend tests for the bokeh plots to detect stuff like this earlier (I believe as of now there's only backend tests in `skyportal/tests/api/test_bokeh_plots.py`)

stefanv:
Any idea what changed? Ideally we don't want to pin to an old version forever.

jadalilleboe:
I haven't looked into it too much yet, but the `width` returned by react-device-detect's `withOrientationChange` in `SourceMobile.jsx` was different between the version change. It was much smaller before, around 300px, and it now returns 1280px when using the same iPhone SE dimensions in the chrome devtools.

jadalilleboe:
oh interesting, the same 1280 is in react-grid-layout's code here [https://github.com/react-grid-layout/react-grid-layout/blob/master/lib/components/WidthProvider.jsx#L56](https://github.com/react-grid-layout/react-grid-layout/blob/master/lib/components/WidthProvider.jsx#L56)
so perhaps an issue with react-grid-layout and the WidthProvider, not react-device-detect.

mcoughlin:
We removed this component a while ago.

PR #3580, opened 2022-10-07
improve mobile plot experience slightly

jadalilleboe:
Change plot width to fixed size instead of using unreliable react-grid-layout width for the time being, to slighltly improve mobile experience while using plots. 
Screen Shot 2022-10-07 at 3 39 56 PM
Screen Shot 2022-10-07 at 3 40 08 PM
addresses part of #2966

### DIFF
--- a/static/js/components/SourceMobile.jsx
+++ b/static/js/components/SourceMobile.jsx
@@ -240,7 +240,7 @@ export const useSourceStyles = makeStyles((theme) => ({
 }));
 
 const SourceMobile = WidthProvider(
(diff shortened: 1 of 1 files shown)


stefanv:
How does this impact using SP on a tablet?

jadalilleboe:
Screen Shot 2022-10-09 at 3 53 10 PM
Screen Shot 2022-10-09 at 3 53 45 PM
Screen Shot 2022-10-09 at 3 54 24 PM
@stefanv an improvement from before I think.

guynir42 (approved):
LGTM

PR #3581, opened 2022-10-07
Limit token_ids to just letters and numbers

mcoughlin:
This PR limits token_ids to just letters and numbers.


stefanv:
Does it work without the dash? If so, go ahead and merge when green.

PR #3582, opened 2022-10-08
Serialize numpy array as list

mcoughlin:
This PR fixes a bug in the generic analysis serialization for numpy arrays, serializing them as lists.

### DIFF
--- a/skyportal/handlers/api/analysis.py
+++ b/skyportal/handlers/api/analysis.py
@@ -583,7 +583,7 @@ class AnalysisHandler(BaseHandler):
     def generic_serialize(self, row, columns):
         return {
             c: getattr(row, c).tolist()
-            if type(getattr(row, c)) == np.ndarray
+            if isinstance(getattr(row, c), np.ndarray)
             else getattr(row, c)
             for c in columns
         }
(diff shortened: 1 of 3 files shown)


stefanv:
What is this serialization used for? If for JSON, could be okay (depending on dtype), otherwise should use npy format.

mcoughlin:
@stefanv it sends up in a csv file, which renders it with ellipses, so it does not make it with its data to requested analyses.

skyportal/handlers/api/analysis.py
```diff
@@ -580,7 +581,12 @@ def delete(self, analysis_service_id):
 
 class AnalysisHandler(BaseHandler):
     def generic_serialize(self, row, columns):
-        return {c: getattr(row, c) for c in columns}
+        return {
+            c: getattr(row, c).tolist()
+            if type(getattr(row, c)) == np.ndarray
```
guynir42, line 586: I would use `isinstance()` to get sub-classes too.

guynir42 (approved):
LGTM. I think this is a generally good idea. I can imagine other cases where this would blow up but for sure this would catch most of the types of data we are working with.

PR #3584, opened 2022-10-10
Better Slack and Email Notifications: GCN Events and Sources

Theodlz:
This PR adds better-looking notifications on slack and email for:
- Gcn events (new notices, new events)
- Sources (new classification)

The current notifications look like this (identical for email and slack):

image

Aside from the apparent lack of .... esthetics, they give very little information about the source/event; they only inform us that something new happened.

In this PR, we improve on both aspects.

New notification for a GCN Event:
| Slack | Email |
|---|---|
|image | image |

New notification for a Source (new classification):
| Slack | Email |
|---|---|
|image | image |

*Again, the reviewers can contact me so I can provide them with a slack integration URL on a private slack organization where they can try that new feature*


skyportal/handlers/api/gcn.py
```diff
@@ -167,11 +167,28 @@ def post_gcnevent_from_xml(payload, user_id, session):
...
-            }
+            if any([True if 'GRB' in tag.text.upper() else False for tag in tags]):
+                dateobs_txt = Time(dateobs).isot
+                source_name = f"GRB{dateobs_txt[2:4]}{dateobs_txt[5:7]}{dateobs_txt[8:10]}.{dateobs_txt[11:13]}{dateobs_txt[14:16]}{dateobs_txt[17:19]}"
+                source = {
+                    'id': source_name,
+                    'ra': ra,
+                    'dec': dec,
+                }
+            elif any([True if 'GW' in tag.text.upper() else False for tag in tags]):
```
mcoughlin, line 178: Did you see if there was a GW specific name in the gcn event?

skyportal/utils/notifications.py
```diff
@@ -0,0 +1,349 @@
...
+    for c in source.classifications:
+        if c.classification == classification_name and classification is None:
+            classification = c
+        elif (
+            c.classification == classification_name
+            and c.modified > classification.modified
+        ):
+            classification = c
+
+    print(source.photstats[0])
```
mcoughlin, line 223: Remove print

mcoughlin (approved):
A few small comments. If there is any refactoring to repeat code between slack and email and such might be nice, otherwise this is a huge step forward.

skyportal/handlers/api/gcn.py
Theodlz: Not yet! Here's what I'll do: If the VOEvent notice has a **Name** tag, I'll use that name. Otherwise, it will be a name like this one. What do you think?

mcoughlin: I can live with that. I expect we will evolve this as the IGWN stuff settles.

mcoughlin (approved):
LGTM!

PR #3586, opened 2022-10-12
Annotation delete button

mcoughlin:
This PR adds delete button for annotations and fixes errorbars in periodic plot.

### DIFF
--- a/static/js/components/AnnotationsTable.jsx
+++ b/static/js/components/AnnotationsTable.jsx
@@ -1,5 +1,7 @@
-import React from "react";
+import React, { useState } from "react";
+import { useDispatch } from "react-redux";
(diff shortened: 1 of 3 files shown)


guynir42 (commented):
Code looks good. Is this delete button going to appear for every user or only for admins? Or is the permission check only done in the handler because each user can delete their own annotations? What happens if I try to delete someone else's annotation?

mcoughlin:
@guynir42 Every user. Permissions is done in the handler so if you are an admin you can delete everyone's and the rest is your ownership.

guynir42:
And if you don't have permission you just see an error right?

mcoughlin:
@guynir42 yes

guynir42 (approved):

PR #3587, opened 2022-10-13
Reject sources in query based on object ID

mcoughlin:
This PR reject sources in query based on object ID.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -425,7 +425,6 @@ def get_sources(
             func.lower(Obj.id).contains(func.lower(sourceID.strip()))
         )
     if nonsourceIDs:
-        print(nonsourceIDs)
         obj_query = obj_query.where(Obj.id.notin_(nonsourceIDs))
 
     if any([ra, dec, radius]):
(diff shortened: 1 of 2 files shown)


stefanv:
Is this a known astronomy concept? Otherwise, the feature seems quite cryptic. Perhaps update the doc to be more explicit?

mcoughlin:
@stefanv I updated the docs a bit more.

skyportal/handlers/api/source.py
```diff
@@ -354,6 +354,7 @@ def get_sources(
     has_spectrum=False,
     has_followup_request=False,
     sourceID=None,
+    nonsourceIDs=None,
```
guynir42, line 357: Maybe `rejectedSourceIDs` or something like that?

guynir42 (approved):
Seems like a pretty useful thing.

stefanv:
> @stefanv I updated the docs a bit more.

I understand the new version—thanks!

skyportal/handlers/api/source.py
mcoughlin: This is a much nicer name indeed.

PR #3590, opened 2022-10-13
fix circular import

guynir42:
The new thumbnail code requires alerts.py to post a new thumbnail. But that code imports thumbnails.py so there's a circular import (only on fritz).


mcoughlin (approved):
LGTM

PR #3592, opened 2022-10-14
Observation table sorting/filtering

mcoughlin:
This table adds basic observation table sorting / filtering.

### DIFF
--- a/skyportal/handlers/api/observation.py
+++ b/skyportal/handlers/api/observation.py
@@ -460,72 +460,76 @@ def get_observations(
     if sort_by is not None:
         if sort_by == "obstime":
             order_by = (
-                [Observation.obstime]
+                Observation.obstime
                 if sort_order == "asc"
-                else [Observation.obstime.desc()]
+                else Observation.obstime.desc()
             )
         elif sort_by == "observation_id":
             order_by = (
-                [Observation.observation_id]
+                Observation.observation_id
                 if sort_order == "asc"
-                else [Observation.observation_id.desc()]
+                else Observation.observation_id.desc()
             )
         elif sort_by == "exposure_time":
             order_by = (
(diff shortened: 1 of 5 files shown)


skyportal/handlers/api/observation.py
```diff
@@ -443,6 +455,78 @@ def get_observations(
...
+                if sort_order == "asc"
+                else [Observation.validity_window_start.desc()]
+            )
+        elif sort_by == "validity_window_end":
+            order_by = (
+                [Observation.validity_window_end]
+                if sort_order == "asc"
+                else [Observation.validity_window_end.desc()]
+            )
+
```
guynir42, line 525: I would put an `else` with raise value error if it doesn't match any of these strings. 
If not, you can skip the `is not None` condition at the beginning as `None` is not equal to any of these. 
Best is to put the `if order by is None` at the first `if` and the `elif` all the strings and leave the last `else` for a wrong string.

skyportal/handlers/api/observation.py
```diff
@@ -443,6 +455,78 @@ def get_observations(
...
+    if order_by is None:
+        order_by = [Observation.instrument_id.desc()]
+    obs_query = obs_query.order_by(*order_by)
```
guynir42, line 528: Would this work without the `*` if you also don't put the `order_by` in a list?

guynir42 (approved):
Minor suggestions...

skyportal/handlers/api/observation.py
mcoughlin: Yes nice thinking.

skyportal/handlers/api/observation.py
mcoughlin: Good idea.

PR #3593, opened 2022-10-14
patch try/except for missing thumbnails file

guynir42:
Turns out, doing `try, finally` is not the same as `try, except: pass, finally`.


profjsb (approved):

profjsb:
Go ahead and merge when green

PR #3595, opened 2022-10-14
fix interface between thumbnails and post_alert

guynir42:
I have not correctly copied the input names from `post_alerts` on fritz. 
It is very hard to develop across different repos. 
But I think we really got it this time!


mcoughlin (approved):
LGTM

PR #3597, opened 2022-10-15
Update post_alert usage

mcoughlin:
This PR updates the  post_alert usage


guynir42 (approved):
Personally I prefer to call these as keyword arguments. It makes life easier if you add more arguments down the line, and it helps a lot to know what is what, e.g., that `'all'` refers to group ids.

mcoughlin:
@guynir42 It is something I will consider longer term, just not something I am as used to.

PR #3599, opened 2022-10-17
Set default classification probability to 1.0

### LINKED ISSUE #3253: Classifications should always be given a number
Right now, you can post classifications without a number. I think any classification without a number should just be set to 1. Right now, the front-end doesn't display any without a number!

bfhealy:
It is currently possible to post null classification probabilities via the api. This PR sets the default probability to 1.0 in cases where no value is provided. Closes #3253


guynir42 (approved):
This is fine. We should consider how to make this enforced by the database. That would also require we do a migration. Should discuss this tomorrow at the meeting.

mcoughlin:
@guynir42 should we set the default to 1 in the DB?

guynir42:
@mcoughlin that's a good idea, but we need to check that the migration will successfully move all the null probabilities into 1.0.

mcoughlin:
@bfhealy can you see if there is an example migration script in alembic/version that adds a default to a column. Can also google quick.

bfhealy:
@mcoughlin  [This article](https://medium.com/the-andela-way/alembic-how-to-add-a-non-nullable-field-to-a-populated-table-998554003134) seems relevant to the migration we'll be doing

mcoughlin:
@bfhealy I think the galaxy table we also added a default. So can maybe draw from that.

mcoughlin:
@bfhealy anyways, going into the models and setting a default of 1 presumably triggers the right migration artifact.

mcoughlin:
@bfhealy 

In skyportal/models/classification.py , 

    probability = sa.Column(
        sa.Float,
        doc='User-assigned probability of belonging to this class',
        nullable=True,
        index=True,
    )

=>

    probability = sa.Column(
        sa.Float,
        doc='User-assigned probability of belonging to this class',
        nullable=False,
        index=True,
        default=1.0,
    )

or similar.

bfhealy:
@mcoughlin Thanks, will this change automatically update the existing null probabilities?

mcoughlin:
@bfhealy In your migration artifact, we will need to make sure the line reads something like:

op.alter_column('classifications', 'probability', server_default=1)

Or something. So push the change and see what the artifact looks like.

bfhealy:
@mcoughlin Looks like alembic is picking up the change, and I'll just have to add `server_default = 1.0` to the upgrade on my own. I'm confused about the other items below that are appearing in the upgrade and downgrade functions, though (`gcnevents` and `instrumentfieldtiles`). Are there other changes that need to be migrated? I'm running the latest version of SkyPortal locally.

def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    **op.alter_column(
        'classifications',
        'probability',
        existing_type=postgresql.DOUBLE_PRECISION(precision=53),
        nullable=False,
    )**
    op.add_column('gcnevents', sa.Column('trigger_id', sa.BigInteger(), nullable=True))
    op.create_unique_constraint(None, 'gcnevents', ['trigger_id'])
    op.create_index(
        op.f('ix_instrumentfieldtiles_instrument_field_id'),
        'instrumentfieldtiles',
        ['instrument_field_id'],
        unique=False,
    )
    # ### end Alembic commands ###


def downgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.drop_index(
        op.f('ix_instrumentfieldtiles_instrument_field_id'),
        table_name='instrumentfieldtiles',
    )
    op.drop_constraint(None, 'gcnevents', type_='unique')
    op.drop_column('gcnevents', 'trigger_id')
    **op.alter_column(
        'classifications',
        'probability',
        existing_type=postgresql.DOUBLE_PRECISION(precision=53),
        nullable=True,
    )**
    # ### end Alembic commands ###

bfhealy:
@mcoughlin Got it, thanks!

bfhealy:
@mcoughlin The artifact is looking more normal now - only the classification probability change is included. I think I just need to specify the server_default myself. Should I then commit the file to this PR?

mcoughlin:
@bfhealy yes but change the name and the comments inside to indicate what the migration is doing.

bfhealy:
From my local testing I think the migration script still needs a line updating the existing null values.

mcoughlin:
@bfhealy i.e setting server_default to 1 doesn't do it? You can always execute raw sql with op.execute and a line that sets all null values in the probability column to 1?

bfhealy:
@mcoughlin Agreed, I got it to work locally by adding an op.execute line.

mcoughlin:
@bfhealy awesome! I would commit it and then tag Guy for review.

alembic/versions/67109d315f17_make_classification_prob_non_nullable_.py
```diff
@@ -0,0 +1,47 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    classifications = sa.sql.table(
+        "classifications", sa.Column("probability", sa.VARCHAR())
+    )
+    op.execute(
+        classifications.update()
+        .where(classifications.c.probability.is_(None))
+        .values(probability=1.0)
+    )
```
guynir42, line 28: From the errors in the migration test I am going to venture a guess that some probability values are set to some string (it may be "null" or something silly like that) and then it complains it is getting a string and not a float.

guynir42: So if you figure out what those are, you may end up adding another `execute` to change that string to a float.

guynir42 (commented):
It looks fine but you have to make sure the migration tests pass. 
A bad migration would be a big headache when this is deployed.

bfhealy:
Leaving this PR as a draft for now, and pivoting to a frontend fix for #3253.

PR #3607, opened 2022-10-21
Fix plotting in case of multiple instruments

mcoughlin:
This PR fixes the plotting in case of multiple instruments.


guynir42 (approved):
Strange bug but if this works, that is great...

PR #3610, opened 2022-10-24
Simplify selection form

mcoughlin:
This PR simplifies the selection form process by passing form parameters from the original page.


guynir42 (approved):
LGTM

PR #3614, opened 2022-10-27
Add affiliations to GCN and shifts

mcoughlin:
This PR adds affiliations to GCN and shifts.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -1220,7 +1220,7 @@ class GcnSummaryHandler(BaseHandler):
                             self.associated_user_object.affiliations
                         )
                     else:
-                        affiliations = "Affiliation"
+                        affiliations = "..."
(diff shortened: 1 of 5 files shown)


skyportal/handlers/api/gcn.py
```diff
@@ -1211,8 +1211,19 @@ class Validator(Schema):
...
+
+                    if (
+                        self.associated_user_object.affiliations is not None
+                        and len(self.associated_user_object.affiliations) > 0
+                    ):
+                        affiliations = ", ".join(
+                            self.associated_user_object.affiliations
+                        )
+                    else:
+                        affiliations = "Affiliation"
```
guynir42, line 1223: It is a bit strange to have "Affiliation" as a placeholder. Would it be better to just cut off the "at..." and leave just the name?

guynir42 (approved):

PR #3615, opened 2022-10-27
Display group source classifications with null probability

### LINKED ISSUE #3253: Classifications should always be given a number

bfhealy:
This PR modifies the frontend to display classifications with null probability on the group source list. In the attached screenshot, the YSO classification has null probability. It is displayed with a '?', and mousing over the classification shows `P=null`. Classifications with non-null probabilities continue to be displayed as before. Closes #3253

Screen Shot 2022-10-27 at 10 51 00 AM


guynir42 (approved):
This is much better. I think this is a better solution that allows a wider use-case.

PR #3617, opened 2022-10-27
White mode background reverted to normal and dark mode font changes

vyom1611:
Changed the default theme for dark mode to have more darker background colors.

Things which are remaining:

- fix font color since it is black in a black background for dark-mode
- might have to add conditions for different font colors in white and dark mode

Fixed white mode background


vyom1611:
Since we are using MUI Cards, I changed the color of the text using the `styles={{}}` attribute in the <Card> tag. However, I could not figure out how to make it change depending on dark mode. So for now, I picked a color which works best for both modes.

mcoughlin:
@guynir42 What do you think?

guynir42:
@mcoughlin I don't know. The black text on dark grey background is not super legible...

mcoughlin:
@vyom1611 anything better possible?

jadalilleboe:
@vyom1611 so it's very annoying but should be possible to change the font color by adding this to the mui theme:
```
typography: {
      body1: {
        color: dark ? grey[50] : null,
      },
      body2: {
        color: dark ? grey[50] : null,
      },
      subtitle1: {
        color: dark ? grey[50] : null,
      },
      subtitle2: {
        color: dark ? grey[50] : null,
      },
      button: {
        color: dark ? grey[50] : null,
      },
      caption: {
        color: dark ? grey[50] : null,
      },
      overline: {
        color: dark ? grey[50] : null,
      },
      h1: {
        color: dark ? grey[50] : null,
      },
      h2: {
        color: dark ? grey[50] : null,
      },
      h3: {
        color: dark ? grey[50] : null,
      },
      h4: {
        color: dark ? grey[50] : null,
      },
      h5: {
        color: dark ? grey[50] : null,
      },
      h6: {
        color: dark ? grey[50] : null,
      },
    },
```
I tried to see if there was a way to target all the variants at once but it doesn't seem so 
Try that and let me know if it helps

vyom1611:
So now most headings are fixed, but small comments and captions are still un-readable. There are probably some tags which are in the components but not in the new theme typography, so I will try to find them and add them too.

vyom1611:
I found a way to change the smaller texts which were still hard to see, however it includes me changing stuff in each component separately. So at this point of time, I have changed the Homepage (although new source form is still pending), and I will keep pushing more changes to other components.

mcoughlin:
@vyom1611 @jadalilleboe @Theodlz Is there any way to refactor this? I.e. so the same color is used everywhere without it being hard-coded in all components?

mcoughlin:
@Theodlz @vyom1611 Is this converging to anything?

PR #3619, opened 2022-10-28
Update classification slider scaling

### LINKED ISSUE #2366: "updateChildren" part of normalize probabilities in taxonomies (MultipleClassificationsForm.jsx)
Currently if parent value is increased, non-zero values of children are scaled.
They should not be.

In SCoPe taxonomy, for instance, if I decide something is bogus at 0.5,
then drag the "galaxy" child to 0.5,
then drag bogus to 0.75, galaxy gets dragged to 0.75 too.
I may feel that the object is bogus at 0.75, but is a galaxy only at 0.5

const newProbability =
        parseFloat(
          (
            (newValue * currentProbability) /
            totalProbabilityOfSubclasses
          ).toFixed(3)
        ) || 0;

bfhealy:
This PR simplifies the behavior of the classification slider scaling on group source pages and resolves #2366 with a slight variation on the requested behavior:

- Parent classifications are updated such that their probability is ≥ the maximum of their children's probabilities.
- Sibling classification probabilities are not updated.
- Child classifications are updated such that their probabilities are ≤ their parent's probability. This is slightly different from the issue's request not to update children at all. I think it is appropriate to avoid parent classifications having a probability that is less than their children's probabilities.

This new behavior no longer normalizes the probabilities, so the toggle is now named 'Scale parent/child probabilities'. The benefit of these changes is that the slider behavior is more predictable, decreasing the likelihood that the classifying user will experience an unwanted update of probabilities.

skyportal_classification_sliders

### DIFF
--- a/static/js/components/MultipleClassificationsForm.jsx
+++ b/static/js/components/MultipleClassificationsForm.jsx
@@ -110,17 +110,17 @@ const MultipleClassificationsForm = ({
   });
   const sortedClassifications = getSortedClasses(currentClassifications);
 
-  const normalizeProbabilities = useSelector(
-    (state) => state.classifications.normalizeProbabilities
+  const scaleProbabilities = useSelector(
+    (state) => state.classifications.scaleProbabilities
   );
 
-  const [normalizeProbabilitiesChecked, setNormalizeProbabilitiesChecked] =
-    useState(normalizeProbabilities);
+  const [scaleProbabilitiesChecked, setScaleProbabilitiesChecked] =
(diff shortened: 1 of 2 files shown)


guynir42 (approved):
Good! I think this will make sense for people. I suggest we also link to this PR in the beta testing (you can do that after it is merged just to let the know this is the new way).

bfhealy:
Thanks @guynir42! I'll mention the PR on the beta-testing channel once it's merged.

PR #3620, opened 2022-10-28
Try to make test more robust

mcoughlin:
This PR makes a front-end test more robust with scrolling.


guynir42 (approved):
I hope this works. I've seen cases where `scroll_parent` only got us in trouble.

PR #3623, opened 2022-10-31
Calculate plan statistics async

mcoughlin:
This PR calculates plan statistics async.


skyportal/tests/api/test_observation_plan.py
```diff
@@ -10,7 +10,7 @@
 from skyportal.tests import api
 
 
-@pytest.mark.flaky(reruns=2)
+# @pytest.mark.flaky(reruns=2)
```
guynir42, line 13: Did you mean to get rid of this?

guynir42 (approved):
LGTM 🎃

skyportal/tests/api/test_observation_plan.py
mcoughlin: oops

PR #3646, opened 2022-11-01
updating bokeh

mcoughlin:
This PR updates bokeh.


guynir42:
@mcoughlin I've made the buttons appear in the right place. But I don't know how to individually color them.

mcoughlin:
@guynir42 that seems like overkill to me, I suggest we just push it like this. What do you think?

guynir42:
The colors matching the lines is not just a nice feature, it is the only way you can identify which line is which (except if you go and click on each of them). 
I suggest we put a little bit more thought into how to do this. 
I think we can hitch hike on the `callback_toggle_lines` function but I don't know which properties to grab on the element or on the button. 
Maybe @jadalilleboe has some intuition how to do this? 
Another option is to have a look at Stefan's original code, maybe there are hints there. 
Are we in a hurry to update Bokeh? Besides the nice sliders?

mcoughlin:
@guynir42 yes it's mostly solving the sliders and not falling behind the major release.

mcoughlin:
@guynir42 I think I got something reasonable working.

guynir42 (approved):
Great! It looks better than the original in my opinion, but maintains the same feel and functionality. 

I did notice the last column was cut off by by the right end of the accordion. Maybe we should shrink the whole thing or go to 6 columns?

mcoughlin:
@guynir42 do you know why the tests fail? Do they fail for you locally?

guynir42:
@mcoughlin It seems like the tests don't even start up. I'm going to guess it has something to do with the JS prerequisites (I had some issues with that). Maybe this is one of those cases where the cache needs to be cleared? 

I will try to run the tests locally.

mcoughlin:
@stefanv @profjsb We confirmed that tests pass locally on this update, but @guynir42 posited it might be some kind of weird GA caching issue. Any ideas?

stefanv:
The docs failure looks real.

When trying it locally, try removing `node_modules` first to make sure you have a clean run.

I've restarted the tests. If there's still a discrepancy, we should take a closer look. I don't think we cache `node_modules`, just `~/.npm`.

mcoughlin:
@stefanv seems weird since importlib is a python built in :)

mcoughlin:
@stefanv I did just confirm that make docs builds fine for me.

stefanv:
> @stefanv seems weird since importlib is a python built in :)

The problem reported is due to pytest, not importlib. It's a valid error too, in that:

```
AttributeError: module 'importlib' has no attribute 'metadata'
```

Perhaps pytest isn't pinned, and the latest release broke something?

stefanv:
Last release was Oct 25th: https://pypi.org/project/pytest/7.2.0/

mcoughlin:
Ah interesting, thanks @stefanv

PR #3650, opened 2022-11-02
ZTF backup test

mcoughlin:
This PR expands radius to fix ZTF test.


mcoughlin:
Self merging as fixed persistently failing test.

PR #3651, opened 2022-11-03
Default survey efficiencies

mcoughlin:
This PR makes default survey efficiencies available (pairing with default observation plans).


services/gcn_service/gcn_service.py
```diff
@@ -42,6 +47,7 @@
         ).first()
         if allocation is not None:
             config_gcn_observation_plan["allocation_id"] = allocation.id
+            config_gcn_observation_plan["urvey_efficiencies"] = []
```
guynir42, line 50:
```suggestion
            config_gcn_observation_plan["survey_efficiencies"] = []
```

skyportal/handlers/api/observation_plan.py
```diff
@@ -84,6 +85,136 @@
...
+        )
+        status_complete = observation_plan_request.status == "complete"
+
+        if not status_complete:
+            time.sleep(30)
+
+    allocation = session.query(Allocation).get(observation_plan_request.allocation_id)
+    localization = session.query(Localization).get(
+        observation_plan_request.localization_id
+    )
```
guynir42, line 124: Can we use `select` here instead of `query`?

skyportal/handlers/api/survey_efficiency.py
```diff
@@ -164,3 +168,178 @@ def get(self, survey_efficiency_analysis_id=None):
...
+            200:
+              content:
+                application/json:
+                  schema: SingleDefaultSurveyEfficiencyRequest
+            400:
+              content:
+                application/json:
+                  schema: Error
+        multiple:
+          description: Retrieve all default survey efficiencys
```
guynir42, line 248:
```suggestion
          description: Retrieve all default survey efficiencies
```

skyportal/models/observation_plan.py
```diff
@@ -71,6 +71,17 @@ class DefaultObservationPlanRequest(Base):
         overlaps='groups',
     )
 
+    default_plan_name = sa.Column(
+        sa.String, unique=True, nullable=False, doc='Default plan name'
+    )
```
guynir42, line 76: I'm not sure this unique constraint is doing what you think it should. 
Are there not going to be any duplicate default plans at any time on the DB?

guynir42 (approved):
I have minor comments.

skyportal/models/observation_plan.py
mcoughlin: I dont want any with the same name at least. :)

guynir42: Is it a global constraint? Not specific to each user or each localization or something like that?

mcoughlin: Yeah I want people to make different names for their plans.

skyportal/handlers/api/observation_plan.py
mcoughlin: yes

PR #3654, opened 2022-11-03
Swift GRB tags

mcoughlin:
This PR checks for the GRB_identified flag in Swift notices when making tags.

### DIFF
--- a/skyportal/utils/gcn.py
+++ b/skyportal/utils/gcn.py
@@ -68,7 +68,16 @@ def get_tags(root):
         pass
     else:
         if value == 'process.variation.burst;em.gamma':
-            yield 'GRB'
+            # Is this a GRB at all?
+            try:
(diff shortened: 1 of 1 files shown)


skyportal/utils/gcn.py
```diff
@@ -68,7 +68,16 @@ def get_tags(root):
         pass
     else:
         if value == 'process.variation.burst;em.gamma':
-            yield 'GRB'
+            # Is this a GRB at all?
+            try:
+                value = root.find(".//Param[@name='GRB_Identified']").attrib['value']
+            except AttributeError:
+                yield 'GRB'
```
guynir42, line 75: If I understand what you are saying, if it finds nothing you assume it is a GRB?

guynir42 (approved):

skyportal/utils/gcn.py
mcoughlin: Yeah (the current behavior)

PR #3655, opened 2022-11-03
E(B-V) on source page

mcoughlin:
This PR adds E(B-V) to source page.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -257,6 +257,7 @@ def get_source(
     source_info["luminosity_distance"] = s.luminosity_distance
     source_info["dm"] = s.dm
     source_info["angular_diameter_distance"] = s.angular_diameter_distance
+    source_info["ebminusv"] = s.ebminusv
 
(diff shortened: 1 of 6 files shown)


skyportal/handlers/api/source.py
```diff
@@ -257,6 +257,7 @@ def get_source(
     source_info["luminosity_distance"] = s.luminosity_distance
     source_info["dm"] = s.dm
     source_info["angular_diameter_distance"] = s.angular_diameter_distance
+    source_info["ebminusv"] = s.ebminusv
```
guynir42, line 260: I used to call this field `ebv` and it seemed to be fairly well understood. Also "extinctions" may be a good name.

guynir42 (approved):
I would put some thought into the name of the variable. Did you check that the automatic downloading of the data files works out of the box?

skyportal/handlers/api/source.py
mcoughlin: ebv is nice.

PR #3665, opened 2022-11-09
Observation plan error handling

mcoughlin:
This PR adds checks and error handling for observation plans and names.

### DIFF
--- a/static/js/ducks/observationPlanNames.js
+++ b/static/js/ducks/observationPlanNames.js
@@ -1,0 +1,35 @@
+import messageHandler from "baselayer/MessageHandler";
+
+import * as API from "../API";
+import store from "../store";
+
+const FETCH_OBSERVATION_PLAN_NAMES = "skyportal/FETCH_OBSERVATION_PLAN_NAMES";
+const FETCH_OBSERVATION_PLAN_NAMES_OK =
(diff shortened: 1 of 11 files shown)


static/js/ducks/observationPlanNames.js
```diff
@@ -0,0 +1,35 @@
...
+  switch (action.type) {
+    case FETCH_OBSERVATION_PLAN_NAMES_OK: {
+      return action.data;
+    }
+    default:
+      return state;
+  }
+};
+
+store.injectReducer("observationPlanNames", reducer);
```
guynir42, line 35: Can this fit in the same file as the other observation plan reducers?

guynir42 (approved):
Looks good. One question/request:

static/js/ducks/observationPlanNames.js
mcoughlin: I am moving this to a duck called observationPlans.

guynir42: Sounds good. Is this the only observation-plan-related reducer we have?

mcoughlin: yes because the observation plan is always tied to a single gcn event before.

guynir42: OK. If it makes sense to you, then it makes sense to me.

PR #3667, opened 2022-11-11
Observing run dialog box

mcoughlin:
This PR creates a dedicated dialog box for observing run creation.

### DIFF
--- a/static/js/components/AddRunFromObservationPlanPage.jsx
+++ b/static/js/components/AddRunFromObservationPlanPage.jsx
@@ -1,0 +1,98 @@
+import React, { useState } from "react";
+import PropTypes from "prop-types";
+import { useDispatch, useSelector } from "react-redux";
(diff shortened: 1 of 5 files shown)


static/js/components/AddRunFromObservationPlanPage.jsx
```diff
@@ -0,0 +1,98 @@
...
+    const params = { groupIds };
+    await dispatch(
+      Actions.createObservationPlanRequestObservingRun(id, params)
+    );
+    setIsCreatingObservingRun(null);
+    closeDialog();
+  };
+
+  return (
+    <>
```
guynir42, line 40: Is this on purpose?

guynir42 (approved):

static/js/components/AddRunFromObservationPlanPage.jsx
mcoughlin: yeah because it is a fragment.

PR #3670, opened 2022-11-14
Group save power

mcoughlin:
This PR adds check for power to save source to group.


guynir42 (approved):
I'm really surprised this is not checked already!

PR #3671, opened 2022-11-14
Search bar in the sources page

Carla802:
Hi @Theodlz @mcoughlin,
This PR adds a search bar at the top of the sources table, in the sources page. 
You can enter a source ID and it will suggest matching results. When clicking on one of the suggestions, you are redirected to the selected source's page.
This is one of the features Sarah asked us to implement. She already confirmed that it was what she expected, but I am open to any advice !


static/js/components/SourceSearchBar.jsx
```diff
@@ -0,0 +1,176 @@
...
+        }
+      }}
+      onClose={() => setOpen(false)}
+      size="small"
+      noOptionsText="No matching sources."
+      options={options}
+      open={open}
+      loading={loading}
+      clearOnEscape
+      clearOnBlur
```
guynir42, line 142: I'm not sure if we want the search bar to automatically clear when clicking outside of it (I assume that's what `clearOnBlur` does.

static/js/components/SourceSearchBar.jsx
```diff
@@ -0,0 +1,176 @@
...
+
+  const debouncedInputValue = useDebouncer(inputValue, 500);
+  const cache = useRef({});
+
+  useEffect(() => {
+    const get = (val) =>
+      dispatch(
+        GET(
+          `/api/sources?sourceID=${val}&pageNumber=1&totalMatches=25&includeComments=false&removeNested=true`,
+          "skyportal/FETCH_AUTOCOMPLETE_SOURCES"
```
guynir42, line 80: We usually don't call these pages directly, but maybe there's a reason here that I don't understand. Besides, I'm not even sure when this gets called, since it looks like when you choose a source it loads the individual source page, and not the table with these parameters (as it should!).

guynir42 (commented):
I have a couple of small comments below. Otherwise, this looks like a really useful feature, thanks!

static/js/components/SourceSearchBar.jsx
Carla802: You're right, I should delete this option !

static/js/components/SourceSearchBar.jsx
Carla802: Thank you for your feedback ! 
To be honest, I reused the code of the QuickSearchBar component (that has the same utility) and adapted it to the sources page. So I'm not really sure of what this call does, but if you consider it pointless, I can delete it.

Carla802: I tested the search bar without the useEffect get but the search bar didn't return any suggestions anymore... I guess this is the call that returns the matching results of the search, so it is indeed necessary. Do I leave it this way ? Apart from that, my last commit removed the clearOnBlur option !

guynir42: I guess that means it is necessary!

guynir42 (approved):

Theodlz:
Hi @Carla802 . I will be able to review your PR tomorrow. Therefore if you are able to today, please lint the files you modified so I can review the PR :)
The commands to run to do so are given in PR #3679

Carla802:
@Theodlz @mcoughlin I have a question (more of a need of confirmation) concerning this PR and the task 6 (the search bar in the recent sources).
Since the code I implemented for the task 6 is using my file SourceSearchBar (which is in this PR but not yet in the main skyportal), I think I need to wait for this PR to be approved, right ? 
So once this code is merged, I would be able to do a new PR for the task 6, because it is basically just importing the SourceSearchBar component in the RecentSources component.
Sorry if this is unclear, let me know if you understood what I meant !

mcoughlin:
@Carla802 yes, but feel free to open with the title WIP (Work in progress) and can rebase the branch once this is merged.

Carla802:
@mcoughlin Ok I will do that, thank you !

Theodlz:
Hi @Carla802 I have a few questions regarding your PR.

I expected this to add a searchbar within the Sources Table, that would interact with the other filters being applied and such.
Instead, this adds a search **independent** of the table but right next to it, which could be misleading.

Also, this search bar is already present everywhere in the app from the navbar (from any page, including the dashboard), I'm not sure why it is necessary to have a second one.

I can see the point to be able to search for sources by name, but since that feature already exists I'm wondering if the integration we want on the Sources page should be a new feature instead, allowing users to search by name + the other filtering options of the table.

Let me know if you see what I mean by that.

When it comes to task 6, I also think that it should be different from this implementation. It should be able to search sources by name AND by most recent.
Here is an example:
Let's say you have ZTF11, ZTF12, ZTF20 in your DB, which were added in this particular order. That means ZTF20 has been added last, and ZTF11 has been added first. 
If we display them in the most recent sources it would naturally be in that order: ZTF20, ZTF12, ZTF11.
Let's imagine we use a search bar and the user types 'ZTF1'. If it was displayed alphabetically, it would be ZTF11 and then ZTF12. But we want the most recent sources (sorted by date descending). Therefore, we want to have ZTF12 and THEN ZTF11. 

So when implementing Task 6, please make sure that this is the case.

Otherwise, the search bar you added works well! Good job.

One last thing, you can already filter by name (which is what a search bar would do) from the filtering form available in the table.
Since filtering by name AND other parameters is already implemented in the table, please consider using that instead.
You can add a search bar in MUI DataTable very easily, and map it to the same filtering feature that is already implemented.

Feel free to reach out to me. What do you think @mcoughlin ?

Carla802:
We updated this PR with the new version of the search bar, directly in the MUI Datatable. It searches by source name. We deleted the old version that was independant from the datatable.
@Theodlz @mcoughlin

mcoughlin (approved):
LGTM

PR #3675, opened 2022-11-15
Number of candidates to display

mcoughlin:
This PR adds more options for number of candidates to display.

### DIFF
--- a/static/js/components/CandidateList.jsx
+++ b/static/js/components/CandidateList.jsx
@@ -1083,7 +1083,7 @@ const CandidateList = () => {
     selectableRows: "none",
     enableNestedDataAccess: ".",
     rowsPerPage,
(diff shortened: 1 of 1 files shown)


static/js/components/CandidateList.jsx
```diff
@@ -1083,7 +1083,7 @@ const CandidateList = () => {
     selectableRows: "none",
     enableNestedDataAccess: ".",
     rowsPerPage,
-    rowsPerPageOptions: [1, 25, 50, 75, 100, 200],
+    rowsPerPageOptions: [1, 5, 10, 25, 50, 75, 100, 200],
```
guynir42, line 1086: Why do we even need 1?

mcoughlin: great point.

guynir42 (approved):

PR #3676, opened 2022-11-15
Add source in GCN comment

mcoughlin:
This PR adds an explanation column for GCN sources.


guynir42 (approved):
The code looks fine. I would say that calling this a "comment" is confusing because we have a well-defined concept of comments, that is also associated with GCNs and sources, and is in fact very different from this thing. 

I would suggest finding a new name for it. Maybe "explanation" or "description" or "note" or whatever you like. Don't use "annotation" :P

PR #3677, opened 2022-11-16
Task1 sort by default by date saved the most recent

laurall974:


mcoughlin:
@laurall974 I don't think you meant to commit anything outside of source.py

laurall974:
> @laurall974 I don't think you meant to commit anything outside of source.py

Yes sorry! I can't delete them on git. I had some struggle with the `git push` so I pushed all.

mcoughlin:
@laurall974 We can't be committing other things, so I would suggest to either revert that change or instead just make a new branch (this is my usual go-to).

PR #3680, opened 2022-11-16
Explanation GCN column

mcoughlin:
This PR fixes the spelling of the explanation column.


mcoughlin:
This PR just corrects a spelling error. Will self-merge.

PR #3682, opened 2022-11-16
Localization properties

mcoughlin:
This PR adds a localiztion property table to track skymap properties for filtering.


skyportal/handlers/api/gcn.py
```diff
@@ -900,6 +1015,69 @@ def add_tiles(localization_id):
...
+            for text in tags_list
+        ]
+        session.add_all(tags)
+
+        session.commit()
+        return log(f"Generated properties for localization {localization_id}")
+    except Exception as e:
+        return log(
+            f"Unable to generate properties for localization {localization_id}: {e}"
+        )
```
guynir42, line 1076: Return log seems kind of strange...

skyportal/handlers/api/gcn.py
```diff
@@ -900,6 +1015,69 @@ def add_tiles(localization_id):
...
+    finally:
+        Session.remove()
```
guynir42, line 1078: Why `remove`?

skyportal/models/localization.py
```diff
@@ -183,3 +204,61 @@ class LocalizationTile(Base):
...
+    )
+
+    localization_id = sa.Column(
+        sa.ForeignKey('localizations.id', ondelete="CASCADE"),
+        primary_key=True,
+        index=True,
+        doc='localization ID',
+    )
+
+    data = sa.Column(JSONB, doc="Localization properties in JSON format.")
```
guynir42, line 235: If you don't index on this, that query for getting all the keys for all the data could be long.

skyportal/models/localization.py
```diff
@@ -183,3 +204,61 @@ class LocalizationTile(Base):
...
+    )
+
+    localization_id = sa.Column(
+        sa.ForeignKey('localizations.id', ondelete="CASCADE"),
+        primary_key=True,
+        index=True,
+        doc='localization ID',
+    )
+
+    text = sa.Column(sa.Unicode, nullable=False)
```
guynir42, line 264: Probably want to index this, too.

guynir42 (approved):
Some minor questions/suggestions

skyportal/handlers/api/gcn.py
mcoughlin: agreed.

skyportal/handlers/api/gcn.py
mcoughlin: That's what most of our IOloop sessions call.

skyportal/models/localization.py
mcoughlin: makes sense.

skyportal/models/localization.py
mcoughlin: makes sense.

PR #3688, opened 2022-11-17
Limiting candidate re-render with memo

mcoughlin:
This PR tries to use React.memo to specify re-render criteria.


mcoughlin:
@guynir42 @profjsb suggested trying something where we no longer pass all the data to MUI tables but instead try to just grab the data in each row. I am not sure this is the best way, but this tries a mix of that and memo to try to limit updates. Something you want to play around with?

PR #3690, opened 2022-11-18
Speedup observation plan statistics

guynir42:
The query to get the observation plan statistics was super slow. 
The reason has to do with a "cartesian product" in the query done to calculate the overlapping fields. 

So I pulled in the LocalizationTile objects and InstrumentFieldTiles as objects into python. 
Then I calculated the overlapping regions of tiles and fields in an efficient way. 
This leads to about a x100 speedup when tested on the demo data, but we should test it on production as well. 

This is not ready to deploy since I've left the old code, and some debugging statements and timing outputs.


skyportal/handlers/api/observation.py
```diff
@@ -418,41 +425,108 @@ def get_observations(
...
+                    obs_subquery,
+                    InstrumentField.id == obs_subquery.c.instrument_field_id,
+                )
+                .where(
+                    InstrumentFieldTile.instrument_field_id == InstrumentField.id,
+                )
                 .distinct()
-                .subquery()
-            )
+            ).all()
```
guynir42, line 446: I am not 100% sure I got the subqueries correct here, so please double check.

guynir42:
@mcoughlin Test failures show that something in the logic I've written is wrong. Let me know if you see anything. I think it is in api/observation.py.

mcoughlin:
@guynir42 So localization_cumprob disappeared. I guess one thing is that for observations (rather than plans), it is nice to be able to restrict the area calculation to the 95% or something, so we don't report huge area numbers that are accumulating just bits of probability. What do you think?

guynir42:
@mcoughlin I wasn't sure about that. What about the queries for galaxies and sources? Should we also bring back cumprob there?

guynir42:
@mcoughlin The test failures seem to be unrelated.

data/db_demo.yaml
```diff
@@ -317,7 +317,8 @@ instrument:
     field_data:
       file: ../data/ZTF_Fields.csv
     field_region:
-      file: ../data/ZTF_Square_Region.reg
+      file: ../data/ZTF_Region.reg
```
mcoughlin, line 320: We probably still want the Square one for the demo data due to its speed.

skyportal/facility_apis/observation_plan.py
```diff
@@ -23,9 +24,70 @@
 default_filters = cfg['app.observation_plan.default_filters']
 
 
+def combine_healpix_tuples(input_tiles):
+    """
+    Combine (adjacent?) healpix tiles, given as tuples of (lower,upper).
+    Returns a list of tuples that do not overlap.
+    """
+
+    # set upper bound to make sure this algorithm isn't crazy expensive
```
mcoughlin, line 33: Maybe we print something to the logs (or even errors?!) if this range is exceeded?

skyportal/tests/api/test_gcn.py
```diff
@@ -755,7 +755,14 @@ def test_gcn_summary_observations(
...
 
-    obs_summary_text = 'We observed the localization region of LVC trigger 2019-08-14T21:10:39.000 UTC.  We obtained a total of 9 images covering ztfr bands for a total of 270 seconds. The observations covered 715.2 square degrees beginning at 2019-08-17T01:00:00.288 (2 days after the burst trigger time) corresponding to ~9% of the probability enclosed in the localization region.'
+    obs_summary_text = (
+        'We observed the localization region of LVC trigger 2019-08-14T21:10:39.000 UTC.  '
+        'We obtained a total of 9 images covering ztfr bands for a total of 270 seconds. '
+        'The observations covered 26.5 square degrees beginning at 2019-08-17T01:00:00.288 '
+        '(2 days after the burst trigger time) corresponding to ~9% '
+        'of the probability enclosed in the localization region.'
+    )
+    print(data[7])
```
mcoughlin, line 765: Remove print

mcoughlin (approved):
This looks good to try.

data/db_demo.yaml
guynir42: Do we want the square in production or only in the testing? I think we want to make it as slow as possible to make sure the new code is fast enough even with the full region.

mcoughlin: I think only in testing. We can run the real one in production.

skyportal/facility_apis/observation_plan.py
mcoughlin: Seems reasonable.

PR #3692, opened 2022-11-19
Exptime, max_airmass and max_fwhm parameters for SEDM follow-up requests

Yashvi-Sharma:
Added three new parameters for SEDM follow-up request api

    exposure_time : Exposure time in seconds for photometry observations only, single value, float, same for all filters
    maximum_airmass : Maximum airmass constraint for any observation, between 1 and 3, default 2.8
    maximum_fwhm : Maximum FWHM constraint for any observation, between 1 and 10, default 10


mcoughlin:
@Yashvi-Sharma gentle reminder to use branches :)

scizen9 (changes requested):
I recommend that you check the item names with the database column names.  They should match or the request might break the SEDM watcher script.  In the database, the relevant columns are: exptime, maxairmass, and max_fwhm.  If you change the corresponding tag names in the payload that is sent to the sedm website, then all will be well.  Selecting the exposure time might also be tricky because as of now, the watcher generates its own exposure time from the magnitude.  We'll have to add code to the watcher to look for the suggested exposure time and process it properly.

Yashvi-Sharma:
I've already made the relevant changes in the watcher script to handle these payload items here https://github.com/scizen9/sedmpy/pull/93/files
And there's a PR open for SEDMv5 for utilizing these constraints in the scheduler https://github.com/scizen9/SEDMv5/pull/16
Can you take a look at these and let me know if they are okay?

mcoughlin:
@scizen9 Looks like @Yashvi-Sharma had some comments above.

mcoughlin:
@scizen9 Probably lowish priority, but is this something we are going to want to clean up and get in?

scizen9 (changes requested):
On line 150 change the leading 'max_airmass' to 'maxairmass' and it will be fine.

Yashvi-Sharma:
line 297 in the PR you merged on sedmpy accounts for the underscore: https://github.com/scizen9/sedmpy/pull/93/files

scizen9 (approved):
OK to merge.

PR #3706, opened 2022-11-28
pin shapley to 1.8.0

guynir42:
This makes it possible to read the geopandas maps using the latest numpy.


mcoughlin (approved):
Approve when Green!

PR #3710, opened 2022-11-30
Specify colors for Y, J and H band photometry

virajkaram:
J-band points for PGIR are currently plotted in yellow, making them really difficult to read. This PR changes the color to black. 
It also specifies colors for the WINTER filters (Y, J and H) to avoid a similar issue in the future.


mcoughlin (approved):
Good idea to have more specification like this.

PR #3733, opened 2022-12-01
No center for multiple events

mcoughlin:
This PR removes calculation for center in case of multiple events.


guynir42 (commented):
Is there another place where it adds the "center" for cases where there are not multiple events?

guynir42 (approved):

PR #3734, opened 2022-12-02
Add galactic longitude and latitude to group source column options

bfhealy:
This PR adds the galactic longitude and latitude to the list of displayable columns on the group sources page. The new columns are not shown by default. See below for an example of this change:

Screen Shot 2022-12-02 at 1 08 12 PM          Screen Shot 2022-12-02 at 1 13 08 PM

### DIFF
--- a/static/js/components/SourceTable.jsx
+++ b/static/js/components/SourceTable.jsx
@@ -681,6 +681,16 @@ const SourceTable = ({
     return <div key={`${source.id}_dec_sex`}>{dec_to_dms(source.dec)}</div>;
   };
 
+  const renderGalLon = (dataIndex) => {
+    const source = sources[dataIndex];
+    return <div key={`${source.id}_gal_lon`}>{source.gal_lon.toFixed(6)}</div>;
+  };
+
+  const renderGalLat = (dataIndex) => {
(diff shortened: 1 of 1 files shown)


guynir42 (approved):
LGTM

mcoughlin:
@bfhealy Looks like you need to fix the PropTypes a bit.

bfhealy:
@mcoughlin Thanks, it should pass that test now!

mcoughlin:
@bfhealy Could you just confirm the front-end tests pass locally? We are having issues with the tests that download files.

bfhealy:
@mcoughlin Sure, I'll let you know how those tests go.

bfhealy:
@mcoughlin I’m running into errors using [act](https://github.com/nektos/act) to run the frontend tests locally.  I don’t think the errors are related to the SkyPortal frontend, since the local tests don’t get far enough to check that. Seems more like installation issues before the app is running.

It does look like the PropTypes error in the GitHub-hosted Model Logic test is no longer failing, and I also noticed all of the eslint errors for `SourceTable.jsx` go away in my editor after I made the most recent changes.

PR #3735, opened 2022-12-03
DNM: act

mcoughlin:
@profjsb pointed out act, which allows one to run github actions locally through Docker.
https://github.com/nektos/act

I ran into a few things while trying this out today.

First of all, it requires a GitHub Personal Access Token, which can be created here:
https://github.com/settings/tokens

These can be set in the command line with -s GITHUB_TOKEN=$GITHUB_TOKEN

I also noticed that our runners use ubuntu-latest, which allows for Ubuntu 22, when act is only up to 20.04 at this point, so you need to replace ubuntu-latest with ubuntu-20.04.

Another issue is that we don't explicitly install  postgresql, so we either need to do that (that's what I try here) or perhaps use the "full" image (which is very large)

ACTIONS_RUNTIME_URL=http://host.docker.internal:34567/ bin/act -j test -W .github/workflows/test_api_and_frontend.yaml -s GITHUB_TOKEN=$GITHUB_TOKEN --artifact-server-path artifacts/

Can try -P ubuntu-20.04=ghcr.io/catthehacker/ubuntu:full-20.04 if you want to try the giant one (which was very slow to pull down).

CC @guynir42

### DIFF
--- a/.github/workflows/test_migrations.yaml
+++ b/.github/workflows/test_migrations.yaml
@@ -19,7 +19,7 @@ on:
 jobs:
   test:
     name: Test SkyPortal migrations
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     timeout-minutes: 90
     env:
       # We only check migrations from this point onward, since we know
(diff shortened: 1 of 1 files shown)


mcoughlin:
@bfhealy Do you want to try this again to see if any act / SkyPortal changes have made this run a bit easier?

bfhealy:
@mcoughlin Sure, I'll take another look at this.

Theodlz:
@mcoughlin still relevant? I honestly couldn't really ever get `act` to work on my Mac either, and not sure we still need that given how easy it is to run the tests locally already.

PR #3738, opened 2022-12-07
Fix a couple more DBSession instances

guynir42:
I am looking for the last places where we still use the old sqlalchemy methods for interacting with the session. 
There are still a lot of this code in the testing infrastructure but that would be very big refactor. 

For now this just wraps up loose ends.

### DIFF
--- a/skyportal/onboarding.py
+++ b/skyportal/onboarding.py
@@ -1,4 +1,5 @@
 import datetime
+import sqlalchemy as sa
 
 from .models import (
     DBSession,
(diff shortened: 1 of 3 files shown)


skyportal/onboarding.py
```diff
@@ -158,39 +165,42 @@ def setup_invited_user_permissions(strategy, uid, details, user, *args, **kwargs
...
     """Update user details using data from provider."""
     if not user:
         return
-    existing_user = DBSession().query(User).filter(User.oauth_uid == uid).first()
+
+    with DBSession() as session:
+        existing_user = session.scalars(
+            sa.select(User).where(User.oauth_uid == uid)
+        ).first()
     if not (
```
mcoughlin, line 204: The calls below should also be within the session right?

skyportal/onboarding.py
```diff
@@ -105,7 +106,10 @@ def get_username(strategy, details, backend, uid, user=None, *args, **kwargs):
         raise Exception("PSA configuration error: `username` not properly captured.")
     storage = strategy.storage
 
-    existing_user = DBSession().query(User).filter(User.oauth_uid == uid).first()
+    with DBSession as session:
```
mcoughlin, line 109: The below should also be indented to be in the session I think?

mcoughlin (approved):
Good idea. Just should check the indenting to be sure anything in the session that is called later is still in the session.

skyportal/onboarding.py
guynir42: fixed

mcoughlin (approved):
LGTM

Issue #3739, opened 2022-12-07
Group members can only delete their own classifications


bfhealy:
Fulluser members of a group can currently only delete classifications that they themselves have posted. This makes it difficult for these users to work with classifications in a collaborative setting.

For example, if a set of initial classifications is posted by one group member for the purpose of being evaluated and changed, other group members can only add, but not remove classifications for sources in that group. Perhaps we could give fulluser group members permission to delete any classification within their groups?

PR #3742, opened 2022-12-08
Allow group admins to delete classifications

bfhealy:
This PR allows users to delete classifications in groups for which they are an admin, even if they did not author the classification (see #3739).


mcoughlin:
@bfhealy Doesn't it need to be the equivalent of "accessible_by_groups_admins" not "accessible_by_group_admins"?

bfhealy:
@mcoughlin Making that change
(using `accessible_by_groups_admins = AccessibleIfGroupUserIsAdminAndUserMatches('groups.group_users.user'`) does allow classifications to be deleted, but I then sporadically get a crash and the following error messages when reloading group sources:

``Warning: Failed prop type: The prop `sources[0].annotations[0].author` is marked as required in `SourceTable`, but its value is `undefined`.``

``Warning: Failed prop type: The prop `textMode` is marked as required in `FavoritesButton`, but its value is `undefined`.``

``Uncaught TypeError: Cannot read properties of null (reading 'group_users')``

bfhealy:
@mcoughlin Thanks, that just about fixed things. Now the only quirk is that when I delete classifications and refresh the group sources page, I can't delete any more unless I close and reopen my SkyPortal window.

mcoughlin:
Interesting! A bit weird. Nothing else weird in the console?

bfhealy:
There are three additional messages anytime I visit the group sources page (before or after the refresh):

``Failed prop type: The prop `sources[0].annotations[0].author` is marked as required in `SourceTable`, but its value is `undefined`.``

``Warning: Failed prop type: The prop `textMode` is marked as required in `FavoritesButton`, but its value is `undefined`.``

``Warning: Each child in a list should have a unique "key" prop.``

mcoughlin:
And it was working before? But now this is true for even the original classification poster?

bfhealy:
The original classification poster can still delete even after the refresh. It's only the group admin that is affected. Once a classification is deleted, the `currentGroupUser` constant that is used to determine delete permission is getting nullified.

Before deletion:
Screen Shot 2022-12-08 at 5 21 20 PM

After:
Screen Shot 2022-12-08 at 5 19 36 PM


It's also worth noting that I see the three warnings above even on the main branch, so they don't seem to be related to the new code.

mcoughlin:
@bfhealy maybe you can put it in some kind of useEffect then? One that checks if the user is null?

bfhealy:
Ok, I'll give that a shot. I'll experiment with a few other fixes as well.

mcoughlin:
@bfhealy Could be ok. Any better idea @guynir42 ?

bfhealy:
Just to restate the problem, after a page refresh the state changes such that the `group` field (and `group_users` within) become null. This prevents the delete button from displaying for group admins, since the check for their group admin status no longer has access to the necessary values from the initial page render.

guynir42:
@bfhealy does the `useSelector` not work in this case?

bfhealy:
@guynir42 `useSelector` works the first time the page is loaded, but subsequent reloads make the group-related values in the state become null. Other values are unchanged - for example, the line `const currentUser = useSelector((state) => state.profile);` still works after a reload.

Here is the output from `console.log(useSelector((state) => state));` on the first load of the page versus subsequent reloads (this was run on the main branch). It looks like the group information specifically becomes null while most other values are unchanged.

First load:
Screen Shot 2022-12-09 at 1 05 16 PM
Reload:
Screen Shot 2022-12-09 at 1 05 45 PM

guynir42:
@bfhealy this is very strange. I think we should make an effort to understand why the groups are cleared (an I suspect they are actively cleared by our code somehow). Building a work-around if we have no choice, but this is going against the whole concept of React. 
Try looking in the console to find which action actually cleared the groups. Also I'm not sure what happens: when navigating to the page the groups are populated but when refreshing they are null? what happens if you re-navigate to the page?

bfhealy:
@guynir42 This does seem like strange behavior that's worth tracking down. I'll see what I can find in the console.

The group fields are populated upon the first navigation to the group sources page, but they are cleared upon refresh. Re-navigating to group sources brings back the group information.

bfhealy:
@guynir42 After some experimentation I haven't been able to find anything that clears the `group` field. I did find that upon refresh, none of FETCH_GROUP, FETCH_GROUP_ADMISSION_REQUESTS, or FETCH_STREAMS are logged in the console. The `groupAdmissionRequests` and `streams` fields become null upon refresh just like `group` does.

guynir42:
@bfhealy what about the "hydrate"?

bfhealy:
@guynir42 That seems to be the key - all three of the `fetch` calls associated with the above fields are missing from `hydrate` in actions.js. I've already been able to fix the `streams` field, but the other two require a group id as input, which is not always defined depending on which page the user is on.

guynir42:
OK I think I see how that would work. 
Then it may be the right thing to do to add a `useEffect` to get those groups.

bfhealy:
Sounds good, thanks!

bfhealy:
@mcoughlin @guynir42 The delete icon now persists through refresh for group admins. I think the only remaining change to make is to update `test_permissions.py` so it doesn't fail when checking the permissions to update and delete classifications.

Screen Shot 2022-12-12 at 1 17 25 PM

mcoughlin (approved):
LGTM

guynir42 (approved):
This is so hacky but it looks like you don't have any other way to do this, so you have my approval.

PR #3748, opened 2022-12-11
Fix onboarding sessions

guynir42:
The use of context managers in the onboarding process seems to fail because of user being created and then closing the session.


mcoughlin (approved):
LGTM. I wonder if a FIXME at the top related to the DBSession vs session stuff may be in order.

guynir42:
@mcoughlin I will open an issue to fix this later.

PR #3749, opened 2022-12-11
Generic follow-up API

mcoughlin:
This PR creates a generic follow-up API (to save triggers either locally or a second skyportal instance).


guynir42 (commented):
LGTM

PR #3750, opened 2022-12-12
Facility API enum

mcoughlin:
This PR resolves migration from facility enum.


mcoughlin:
Fixes migration - self merging.

PR #3753, opened 2022-12-13
DefaultFollowupRequestHandler Issue: Str instead of Dict (JSONB)

Theodlz:
A bug prevents other features (such as the triggering of a default obs plan based on a new classification) from being triggered correctly because the `source_filter` column receives a string instead of a dict when the DefaultFollowupRequest is created. 

These few lines ensure that it works as intended by converting the string input to dict if necessary. Of course, if the string is not in the correct format in the first place we use self.error to notify whoever submitted the DefaultFollowupRequest that it needs to be changed.


mcoughlin (approved):
LGTM

PR #3756, opened 2022-12-14
Remove position if not specified

mcoughlin:
This PR removes the position if not specified and enable classification deletion.


skyportal/handlers/api/classification.py
```diff
@@ -480,6 +480,55 @@ def get(self, obj_id):
             )
             return self.success(data=classifications)
 
+    @auth_or_token
+    def delete(self, obj_id):
+        """
+        ---
+        description: Delete an object's classifications
```
guynir42, line 487: Emphasize this deletes ALL the classifications on a source.

static/js/components/SourceTable.jsx
```diff
@@ -1247,7 +1395,7 @@ const SourceTable = ({
     serverSide: true,
     rowsPerPage: numPerPage,
     page: pageNumber - 1,
-    rowsPerPageOptions: [10, 25, 50, 75, 100, 200],
+    rowsPerPageOptions: [1, 5, 10, 25, 50, 75, 100, 200],
```
guynir42, line 1398: Why do we need 1 per page?

guynir42 (approved):
LGTM. I have a couple of minor comments. 
I'm not sure when you'd want to clear all the classifications, it seems like a bit of overkill?

PR #3763, opened 2022-12-17
Source scanning and classification voting

mcoughlin:
This PR adds a source scanning checkbox and classification up and down voting.


guynir42:
What is a source scanning checkbox?

mcoughlin:
Screenshot 2022-12-19 at 10 11 06 AM
 @guynir42 Checkbox to indicate someone looked at the source.

guynir42:
a) Why do we need this?
b) I can imagine people scanning tens of objects every day, it seems like a big ask to have them check a box for each source manually. Do you think the users will actually do this?

mcoughlin:
@guynir42 We need to have people check off that they looked at a source (so other's don't have to do it in the same group). Mild disagreement on the box checking, they already have to do multiple clicks for the classification drop down menu so that's currently the rate limiting step.

guynir42:
@mcoughlin I am vert wary of adding extra clicks, even if they are on top of other clicks. 
I seem to remember that usually people just save the source. I guess there are alternative ways of doing scanning where instead of saving they add classifications. Isn't it then enough that there's a classification from the group to mark it as "scanned"? 
I'm sorry to push back against this but it looks to me like the scanning page is already complicated with so many buttons. Can we do anything that would reduce this frankenstein of buttons, checkboxes and dropdowns? 
(unless there are users that actually asked for this, specifically)

mcoughlin:
@guynir42 This is what was discussed at a dedicated meeting of the group who are doing large scale labeling for the source classification stuff. Here, there are already sources and the question is what their classification is. And I think there is benefit to specifically having noted who did what (and for oneself, did I finish labeling this source or not).

guynir42:
@mcoughlin ok so this is coming at the users request, that's good. 
They are going over sources already saved to their groups and only applying classifications. 
What happens if the user only applies classification or vote on an existing one? Can we get that to automatically check the box? I imagine going over tens of sources a day and having to re-affirm that I have seen the source after I've already interacted with it seems very annoying to me. 
Perhaps we should leave this checkbox/button so you will have the option to leave the source as-is but "mark as viewed" in case you don't want to make any vote or change?

guynir42:
and I would name this "source viewed" rather than "source scanned" because scan is for candidates.

mcoughlin:
@guynir42 How about source labelled? I like the idea of auto-checking, I can add that.

guynir42:
@mcoughlin I'm thinking about my podcast app where you can "mark as played" but whatever you think best describes what the users are doing.

mcoughlin:
@guynir42 Makes sense. I renamed it and now the checkbox automatically checks once you classify an object.

bfhealy:
@mcoughlin Thanks for implementing these requests from the meeting last week! I think a checkbox that automatically fills after a vote, classification post or deletion is a good idea.

I'm noticing that the current implementation of classification voting only allows one vote maximum per classification. When two users vote the same way, only one vote is shown. And when they disagree, the vote is changed based on the most recent input. Also, it seems like only the first voter for a classification can undo their vote by clicking on the same button again.

mcoughlin:
@bfhealy Should be resolved. Please check.

bfhealy:
@mcoughlin Thanks, the votes are now working as expected! Just two other questions:
- Should the 'labelled by' box get checked when users vote on or delete classifications? Currently posting a classification is the only behavior that checks the box automatically.
- Should we store the 'labeled by' information for each source? In Mose cases the vote information could be used to determine which sources were labeled, but I could see future utility e.g. in filtering sources by 'labelled/not labelled by user X'.

mcoughlin:
@bfhealy I addressed your comments. We have the users being saved, so user filtering could be added in the future.

bfhealy:
Thanks @mcoughlin, this looks great!

skyportal/handlers/api/source_labels.py
```diff
@@ -0,0 +1,173 @@
...
+                return self.error("Invalid objId")
+
+            source_labels = session.scalars(
+                SourceLabel.select(session.user_or_token, mode="delete").where(
+                    SourceLabel.obj_id == obj_id
+                )
+            ).first()
+            print(source_labels)
+            print(self.associated_user_object.id)
+            print(group_ids)
```
guynir42, line 149:
```suggestion
```

skyportal/handlers/api/source_labels.py
```diff
@@ -0,0 +1,173 @@
...
+                if source_label is not None:
+                    session.delete(source_label)
+            session.commit()
+
+            source_labels = session.scalars(
+                SourceLabel.select(session.user_or_token, mode="delete").where(
+                    SourceLabel.obj_id == obj_id
+                )
+            ).first()
+            print(source_labels)
```
guynir42, line 167:
```suggestion
```

skyportal/handlers/api/source_labels.py
```diff
@@ -0,0 +1,173 @@
...
+                ).first()
+                if source_label is not None:
+                    session.delete(source_label)
+            session.commit()
+
+            source_labels = session.scalars(
+                SourceLabel.select(session.user_or_token, mode="delete").where(
+                    SourceLabel.obj_id == obj_id
+                )
+            ).first()
```
guynir42, line 166: I assume this query is here just for debugging?

skyportal/models/classification.py
```diff
@@ -65,3 +71,53 @@ class Classification(Base):
...
+        doc="ID of the User that made this ClassificationVote",
+    )
+    voter = relationship('User', doc="The User that scanned this source.")
+    created_at = sa.Column(
+        sa.DateTime,
+        nullable=False,
+        default=datetime.utcnow,
+        index=True,
+        doc="UTC timestamp of the vote.",
+    )
```
guynir42, line 117: Isn't this generated automatically by baselayer?

skyportal/models/classification.py
```diff
@@ -65,3 +71,53 @@ class Classification(Base):
...
+    vote = sa.Column(
+        sa.Integer,
+        doc='User-assigned vote for classification',
```
guynir42, line 120: I assume this will be +1 or -1? Maybe the doc should say that.

skyportal/models/source_label.py
```diff
@@ -0,0 +1,57 @@
...
+        back_populates='source_labels',
+        doc='The Group the label is associated with.',
+    )
+    created_at = sa.Column(
+        sa.DateTime,
+        nullable=False,
+        default=datetime.utcnow,
+        index=True,
+        doc="UTC timestamp of the labelling.",
+    )
```
guynir42, line 57: Again, I think this is added to all models.

static/js/components/ShowClassification.jsx
```diff
@@ -80,6 +92,29 @@ const ClassificationRow = ({ classifications }) => {
 
   const classification = classifications[0];
 
+  const upvoterIds = [];
+  const downvoterIds = [];
+  let upvoteValue = 1;
+  let downvoteValue = -1;
+  let upvoteColor = "action";
+  let downvoteColor = "action";
```
guynir42, line 100: I would do a green vs red or something here, it would probably make it easier to click the one you want in a hurry.

static/js/components/ShowClassification.jsx
```diff
@@ -128,6 +163,26 @@ const ClassificationRow = ({ classifications }) => {
...
+                onClick={() => addVote(classification.id, downvoteValue)}
+              >
+                <ThumbDown color={downvoteColor} />
+                <p> {` ${downvoterIds.length} votes`} </p>
+              </Button>
+            </div>
+            <div>
+              <Button
+                key={classification.id}
+                id="add_vote"
```
guynir42, line 179:
```suggestion
                id="up_vote"
```

static/js/ducks/source.js
```diff
@@ -266,7 +289,7 @@ export function getCommentOnSpectrumAttachmentPreview(spectrumID, commentID) {
 
 export function fetchSource(id, actionType = FETCH_LOADED_SOURCE) {
   return API.GET(
-    `/api/sources/${id}?includeComments=true&includeColorMagnitude=true&includeThumbnails=true&includePhotometryExists=true&includeSpectrumExists=true`,
+    `/api/sources/${id}?includeComments=true&includeColorMagnitude=true&includeThumbnails=true&includePhotometryExists=true&includeSpectrumExists=true&includeLabellers=true`,
```
guynir42, line 292: Is there a way to move all these URL parameters into a parameter object to pass like we would do for python? This url is getting a bit long...

guynir42 (commented):
I have some minor comments on the code. Let me checkout this branch and play with it a little.

skyportal/handlers/api/classification.py
```diff
@@ -23,11 +32,12 @@ def post_classification(data, user_id, session):
...
     user = session.scalar(sa.select(User).where(User.id == user_id))
+    obj_id = data['obj_id']
 
+    sources = session.scalars(Source.select(user).where(Source.obj_id == obj_id)).all()
+    source_group_ids = [source.group_id for source in sources]
     user_group_ids = [g.id for g in user.accessible_groups]
-    group_ids = data.pop("group_ids", user_group_ids)
-
-    obj_id = data['obj_id']
+    group_ids = data.pop("group_ids", list(set(source_group_ids) & set(user_group_ids)))
```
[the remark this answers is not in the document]
mcoughlin: Right now it is saved to all groups, not even to those it is currently a source of. This is already narrowing the default (correctly).

skyportal/handlers/api/source.py
```diff
@@ -1793,6 +1862,13 @@ def get(self, obj_id=None):
               type: boolean
             description: |
               Boolean indicating whether to return if a source has a spectra. Defaults to false.
+          - in: query
+            name: includeLabellers
+            nullable: true
+            schema:
+              type: boolean
+            description: |
+              Boolean indicating whether to return marked source labellers. Defaults to false.
```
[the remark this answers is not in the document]
mcoughlin: Yeah fixing.

skyportal/models/classification.py
mcoughlin: yes.

skyportal/models/source_label.py
mcoughlin: good point.

static/js/components/ShowClassification.jsx
mcoughlin: yeah i like that.

static/js/ducks/source.js
mcoughlin: Yeah I think fixed.

guynir42:
A few things I've noticed: 
- there should be a space between the thumbs up/down icon and the number of votes.
- the background of the hover popup is dark and hard to read the text from, at least in light-mode.
- I would actually change the color to be brighter after clicking. The "disabled" color of the thumbs up/down button is a bit confusing since it is not obvious that another click disables it. 
- the new hover/popup blocks the dialog for deleting a classification. 
- I don't see the checkbox anywhere. I looked in the regular sources table and in the group sources. Should I do something else?

mcoughlin:
@guynir42 I fixed some of the font stuff. I am not sure what you have in mind for the "brighter" color, the Icon API is reasonably limited. The order of the dialog can easily be fixed by scrolling, so this seems a minor nuisance.

guynir42:
@mcoughlin the new font looks a lot better. 
The overlapping dialog is going to be annoying to the users, assuming they actually delete classifications very often. I suspect they may not need to, with the voting option. 
I am still thinking that the color scheme is not right. When you click the button it becomes gray (disabled). But you actually want it to light up after being clicked, so it is obvious that you can re-click to cancel the vote. I'm thinking it may even be worth reversing it so the buttons start as gray and get green/red when clicked. I am also open to other ideas but you need to make it obvious that clicking turns it "on".

mcoughlin:
@guynir42 I switched them, and I think this is better. I explicitly set the z-index on the deletion dialog so that helped I think.

guynir42:
@mcoughlin This is much, much better.

guynir42 (approved):

mcoughlin:
Thanks @guynir42 and @bfhealy for very useful back and forth.

PR #3764, opened 2022-12-18
Taxonomy modifications

mcoughlin:
This PR enables modifications to taxonomies.

PR #3765, opened 2022-12-19
WIP: Add API taxonomy patch method

bfhealy:
This PR adds an API patch method to update parts of a taxonomy. The name, groups, provenance and isLatest flag can be updated using this method, but not the version or hierarchy itself. 

This is a WIP because permissions need to be modified to allow the code to work properly. The `verify` code (L77) in `baselayer/app/models.py` discovers `updated_rows` and seems to check for an 'Update taxonomy' ACL that does not currently exist.

Also, the permissions decorator for the patch method currently checks for 'Delete taxonomy' permission, but this may be too stringent given the modifications the patch method can make.


bfhealy:
Closing in favor of #3764.

PR #3769, opened 2022-12-21
Make source and plot handlers async

bfhealy:
This PR makes certain functions in source.py and plot.py asynchronous to address timeout issues. The modified code currently raises the following errors locally:

Screen Shot 2022-12-21 at 3 56 37 PM
Screen Shot 2022-12-21 at 3 56 16 PM

I welcome comments on which additional functions to make asynchronous and how to avoid the above errors.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -348,7 +348,7 @@ def post_gcnevent_from_dictionary(payload, user_id, session):
 
 class GcnEventTagsHandler(BaseHandler):
     @auth_or_token
-    def get(self):
+    async def get(self):
         """
         ---
         description: Get all GCN Event tags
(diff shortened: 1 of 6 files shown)


mcoughlin:
@bfhealy There is the await calls in the front-end for that.

bfhealy:
@mcoughlin The latest push attempts to use await in GroupSources.jsx, but it doesn't seem quite right. The previous errors no longer appear, but the group sources page hangs without displaying any sources.

mcoughlin:
@bfhealy looks like the GCN tests are relevant?

bfhealy:
@mcoughlin Those tests could be relevant, I'll take a look.

bfhealy:
@mcoughlin The GCN tests are passing now; I think the remaining failures are not related to these changes.

PR #3772, opened 2022-12-22
default observation plans 

mcoughlin:
This PR moves observation plans inside gcn event ingestion.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -36,13 +36,19 @@ from ...utils.UTCTZnaiveDateTime import UTCTZnaiveDateTime
 
 from baselayer.app.access import auth_or_token
 from baselayer.log import make_log
+from baselayer.app.env import load_env
 
 from .source import post_source
+from .observation_plan import (
+    post_observation_plan,
+    post_survey_efficiency_analysis,
+)
 from ..base import BaseHandler
 from ...models import (
     DBSession,
     Allocation,
     CatalogQuery,
+    DefaultObservationPlanRequest,
     EventObservationPlan,
     GcnEvent,
     GcnNotice,
@@ -75,6 +81,7 @@ from skyportal.models.gcn import SOURCE_RADIUS_THRESHOLD
 
 log = make_log('api/gcn_event')
 
+env, cfg = load_env()
 
 Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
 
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/gcn.py
```diff
@@ -1033,9 +1056,98 @@ def add_tiles(localization_id):
...
+            .unique()
+            .all()
+        )
+        gcn_observation_plans = []
+        for plan in default_observation_plans:
+            allocation = session.scalars(
+                Allocation.select(user).where(Allocation.id == plan.allocation_id)
+            ).first()
+
+            gcn_observation_plan = {}
```
Theodlz, line 1098: Not very important, I'll let you decide if you change it or not, but here you could just create the object/dict directly, rather than instantiate it and set each key/value one by one.

skyportal/handlers/api/gcn.py
```diff
@@ -1033,9 +1056,98 @@ def add_tiles(localization_id):
...
+        return log(
+            f"Generated tiles / observation plans for localization {localization_id}"
+        )
     except Exception as e:
-        log(f"Unable to generate contour for localization {localization_id}: {e}")
+        log(
+            f"Unable to generate tiles / observation plans for localization {localization_id}: {e}"
+        )
     finally:
         Session.remove()
```
Theodlz, line 1152: Any reason why in the `add_tiles_and_observation_plans(localization_id, user_id)` method you choose to create a session and remove it manually rather than using a context manager as we usually do?

skyportal/handlers/api/observation_plan.py
```diff
@@ -207,40 +205,60 @@ def post_survey_efficiency_analysis(
...
+            maximum_phase=payload['maximumPhase'],
+            model_name=payload['modelName'],
+            optional_injection_parameters=payload['optionalInjectionParameters'],
+        )
 
     return survey_efficiency_analysis.id
 
 
-def post_observation_plans(plans, user_id, session):
+def post_observation_plans(plans, user_id, session, asynchronous=True):
```
Theodlz, line 249: In this method, you do not use the `asynchronous` parameter. You only seem to use it in the `post_observation_plans` one (singular, without the s). Please consider either using it or removing it.

skyportal/handlers/api/observation_plan.py
```diff
@@ -2008,7 +2031,7 @@ def observation_simsurvey(
         order = hp.nside2order(localization.nside)
         t = rasterize(localization.table, order)
 
-        if 'DISTMU' in t:
+        if 'DISTMU' in t.colnames:
```
Theodlz, line 2034: Maybe you also want to check if the other columns you use in the if clause are used, not just this one? (DISTSIGMA, DISTNORM)

Theodlz (approved):
Very minor comments, otherwise it looks good :rocket:

skyportal/handlers/api/gcn.py
mcoughlin: fixed.

skyportal/handlers/api/gcn.py
mcoughlin: We have done this for all our async handlers... not sure about why honestly :)

skyportal/handlers/api/observation_plan.py
mcoughlin: Fixed.

skyportal/handlers/api/observation_plan.py
mcoughlin: implemented.

PR #3773, opened 2022-12-22
Observation plan ToO

mcoughlin:
This PR adds ToO's to observation plan.

### DIFF
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -24,6 +24,7 @@ from astroplan import FixedTarget
 from astroplan import ObservingBlock
 from astroplan.constraints import (
     Constraint,
+    AltitudeConstraint,
     AtNightConstraint,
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/followup_request.py
```diff
@@ -808,6 +809,36 @@ def delete(self, request_id):
             return self.success()
 
 
+class TargetOfOpportunityConstraint(Constraint):
+    """
+    Prioritize target of opportunity targets earlier.
```
Theodlz, line 814: This isn't very clear at least for me. The comment doesn't really help me understand what the method is supposed to do.

Theodlz (approved):
One very minor comment, but LGTM

skyportal/handlers/api/followup_request.py
mcoughlin: Yes added better text.

PR #3776, opened 2022-12-25
Spatial catalog

mcoughlin:
This PR adds spatial catalogs for source cross-matching.


skyportal/handlers/api/spatial_catalog.py
```diff
@@ -0,0 +1,379 @@
...
+            '*',
+            "skyportal/REFRESH_SPATIAL_CATALOGS",
+        )
+
+        end = time.time()
+        duration = end - start
+
+        log(f"Generated catalog with ID {catalog_id} in {duration} seconds")
+    except Exception as e:
+        log(f"Unable to generate catalog: {e}")
```
Theodlz, line 109: Maybe rollback the session here? Imagine you added the `SpatialCatalogEntry`, but then you get an error when adding the corresponding `Tile`, it would be problematic to end up with entries that are missing their tiles. In that case, it would be better to cancel the first operation.

skyportal/handlers/api/spatial_catalog.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+
+        data = self.get_json()
+        catalog_name = data.get('catalog_name')
+        catalog_data = data.get('catalog_data')
+
+        if catalog_name is None:
+            return self.error("catalog_name is a required parameter.")
+        if catalog_data is None:
+            return self.error("catalog_data is a required parameter.")
```
Theodlz, line 179: Please check that the catalog_data is a dict, and that the name is a string, otherwise you'll get plenty of errors right after

skyportal/handlers/api/spatial_catalog.py
```diff
@@ -0,0 +1,379 @@
...
+
+        # check RA bounds
+        if any([(x < 0) or (x >= 360) for x in catalog_data['ra']]):
+            return self.error("ra should span 0=<ra<360.")
+
+        # check Declination bounds
+        if any([(x > 90) or (x < -90) for x in catalog_data['dec']]):
+            return self.error("declination should span -90<dec<90.")
```
Theodlz, line 187: careful here, you might want to first check if `ra` and `dec` are in the dict to avoid an error

skyportal/handlers/api/spatial_catalog.py
```diff
@@ -0,0 +1,379 @@
...
+
+        with self.Session() as session:
+            stmt = SpatialCatalog.select(self.current_user).where(
+                SpatialCatalog.catalog_name == catalog_name
+            )
+            catalog = session.scalars(stmt).first()
+            if catalog is None:
+                catalog = SpatialCatalog(catalog_name=catalog_name)
+                session.add(catalog)
+                session.commit()
```
Theodlz, line 203: and else? I think you want this to end here if there is already a catalog right? So consider using an `else:` statement + `self.error('...')`

skyportal/handlers/api/spatial_catalog.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+
+        catalog_name = self.get_query_argument("catalog_name", None)
+
+        with self.Session() as session:
+
+            if catalog_id is not None:
+
+                stmt = SpatialCatalog.select(self.current_user).where(
+                    SpatialCatalog.id == int(catalog_id)
```
Theodlz, line 238: instead of converting to int in the query, I would do that right after getting the `catalog_name` at the beginning of the handler. In general you could verify the datatype of your query params and data before running the queries.

skyportal/handlers/api/spatial_catalog.py
```diff
@@ -0,0 +1,379 @@
...
+        if catalog_data is None:
+            return self.error("catalog_data is a required parameter.")
+
+        # check RA bounds
+        if any([(x < 0) or (x >= 360) for x in catalog_data['ra']]):
+            return self.error("ra should span 0=<ra<360.")
+
+        # check Declination bounds
+        if any([(x > 90) or (x < -90) for x in catalog_data['dec']]):
+            return self.error("declination should span -90<dec<90.")
```
Theodlz, line 357: Same comment here, please check that the data is indeed a dict and that it does have `ra` and `dec` keys ( you can also do some data type checking for other variables, but this is the obvious one that could cause problems pretty easily.

static/js/components/SourceTableFilterForm.jsx
```diff
@@ -21,6 +21,7 @@ import Button from "./Button";
 import { allowedClasses } from "./ClassificationForm";
 
 import * as gcnEventsActions from "../ducks/gcnEvents";
+import * as spatialCatalogsActions from "../ducks/spatialCatalogs";
```
Theodlz, line 24: I'm putting the comment here but it's about the Component in general, not that line.
Let's say I pick catalog A and entry A to filter on. Then, on top of the table you see the 2 filters being applied and you can click on them to delete one of them. If I delete the entry, well as discussed earlier the handler sends back an error saying the entry doesn't exist, and the table doesn't show up anymore. What could be much cleaner is that on delete of an element, both filtering are removed. 

However, I don't think that catching the deletes of the filters is something we do yet, so we would need to figure it out... It's not critical per say, but it does force you to reload SkyPortal, which can take a while sometimes, or at least change page and come back.

static/js/components/SpatialCatalogPage.jsx
```diff
@@ -0,0 +1,430 @@
...
+        if (result.status === "success") {
+          dispatch(
+            showNotification("Spatial catalog deleting... please be patient.")
+          );
+          dispatch(spatialCatalogsActions.fetchSpatialCatalogs());
+          closeDialog();
+        }
+      }
+    );
+  };
```
Theodlz, line 243: Very minor comment: on delete, the entries of the deleted catalog still show up until your refresh or select another one. That's ok though, and can be fixed in a following PR.

static/js/components/SpatialCatalogTable.jsx
```diff
@@ -0,0 +1,195 @@
...
+      {catalog.entries ? (
+        <Paper className={classes.container}>
+          <StyledEngineProvider injectFirst>
+            <ThemeProvider theme={getMuiTheme(theme)}>
+              <MUIDataTable
+                title={!hideTitle ? "Catalog Entries" : ""}
+                data={catalog.entries}
+                options={options}
+                columns={columns}
+              />
```
Theodlz, line 157: There is an issue with the `download` option. Basically the `entry_data` column just shows `[Object object]` because its a dict inside, and by default MUI tables don't handle that. So it would need a proper downloadCallback to get it right. If its not important yet and won't be fixed in this PR, please consider deactivating the download button of the table to avoid confusion.

Theodlz (changes requested):
I have some comments, some changes are needed I believe.

doc/spatial_catalog.md
```diff
@@ -0,0 +1,87 @@
+
+# Spatial Catalogs
+
+For catalogs where source location uncertainties are large enough that the Source class is insufficient, SkyPortal has the concept of `SpatialCatalog`s. These catalogs enable cross-matching between sources and objects with spatial extent, such as gamma-ray source catalogs.
+
+Catalogs can be uploaded through the API or on the Source Catalogs web page. Files must have the columns name, ra, dec, where name is the source catalog entry (must be unique), ra and dec are right ascencion and declination in degrees. They must also either contain (i) for an ellipse, amaj, amin, and phi, where amaj, amin, and phi are the major and minor ellipse elements with phi the corresponding angle or (ii) for a cone, radius, where radius is the radius of the cone in degrees.
```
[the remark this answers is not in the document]
mcoughlin: nice catch.

skyportal/utils/gcn.py
```diff
@@ -313,6 +316,39 @@ def from_polygon(localization_name, polygon):
...
+    moc = MOC.from_healpix_cells(ipix, depth, fully_covered)
+    ipix = flatten_pixels(moc._interval_set._intervals, int(np.log2(NSIDE)))
+
+    # Convert to multi-resolution pixel indices and sort.
+    uniq = ligo.skymap.moc.nest2uniq(nside_to_level(hpx.nside), ipix.astype(np.int64))
+    i = np.argsort(uniq)
+    ipix = ipix[i]
+    uniq = uniq[i]
+
+    probdensity = 1.0 * np.ones(ipix.shape)
```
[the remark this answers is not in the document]
mcoughlin: good point

skyportal/handlers/api/source.py
```diff
@@ -1025,6 +1030,57 @@ def get_sources(
...
+            SpatialCatalog.select(
+                user,
+            ).where(SpatialCatalog.catalog_name == spatial_catalog_name)
+        ).first()
+
+        catalog_entry = session.scalars(
+            SpatialCatalogEntry.select(
+                user,
+            )
+            .where(SpatialCatalogEntry.entry_name == spatial_catalog_entry_name)
```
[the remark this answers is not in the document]
mcoughlin: good point.

skyportal/handlers/api/source.py
```diff
@@ -2086,6 +2142,18 @@ def get(self, obj_id=None):
...
+            schema:
+              type: string
+            description: |
+                Name of spatial catalog to use.
+          - in: query
+            name: spatialCatalogEntryName
+            schema:
+              type: string
+            description: |
+                Name of spatial catalog entry to use.
```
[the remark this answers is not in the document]
mcoughlin: yup.

skyportal/handlers/api/source.py
```diff
@@ -2199,6 +2267,10 @@ def get(self, obj_id=None):
         localization_reject_sources = self.get_query_argument(
             "localizationRejectSources", False
         )
+        spatial_catalog_name = self.get_query_argument("spatialCatalogName", None)
+        spatial_catalog_entry_name = self.get_query_argument(
+            "spatialCatalogEntryName", None
+        )
```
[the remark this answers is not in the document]
mcoughlin: good idea.

skyportal/handlers/api/spatial_catalog.py
mcoughlin: yes good.

skyportal/handlers/api/spatial_catalog.py
mcoughlin: yes.

skyportal/handlers/api/spatial_catalog.py
mcoughlin: yes.

skyportal/handlers/api/spatial_catalog.py
mcoughlin: No then you are simply appending to a catalog.

skyportal/handlers/api/spatial_catalog.py
mcoughlin: good idea.

skyportal/handlers/api/spatial_catalog.py
mcoughlin: added.

static/js/components/SpatialCatalogPage.jsx
mcoughlin: yeah it's because it's async.

static/js/components/SourceTableFilterForm.jsx
mcoughlin: https://github.com/skyportal/skyportal/issues/3780

static/js/components/SpatialCatalogTable.jsx
mcoughlin: Yeah for now I am turning off download on those other columns. These JSON payload ones don't work well in the CSV stuff.

Theodlz (approved):
LGTM! :rocket:

PR #3777, opened 2022-12-27
WIP - Optional Microservice: Fink Broker

Theodlz:
A year ago, we developed [SkyPortal Fink Broker](https://github.com/skyportal-contrib/skyportal-fink-client), hosted at [skyportal-contrib](https://github.com/skyportal-contrib/skyportal-fink-client) on GitHub. This is a client to poll alerts from Fink broker and post them to any (local or distant) instance of SkyPortal only using API calls.

As much as the model proved itself useful, it also revealed its limitations. By only interacting with the app through API calls, it became obvious that the code would get more and more complex, for simple tasks that could be performed with simple calls to the DB. Moreover, certain operations are not possible through API calls, and having a fink client run within the app makes anything possible.

In [Icare], we defined a microservice that made sure that the service ran on the same machine as SP, and at the same time. However, skyportal-fink-client was unchanged and still did not have access to anything other than API calls (it remained decoupled from the app, simply running synchronously).

For these reasons, and in an effort to avoid diverging from SkyPortal main every time changes have to be made (and to avoid having to maintain several repos) we designed a fink microservice where all the operations (and more much) performed by skyportal-fink-client are done with less latency and way fewer lines of code. It is significantly inspired by Kowalski's workflow and re-uses some of its methods for uniformity's sake.

At the same time, we added elements that were missing: posting of candidates, sources, and photometry to the right streams and filters (streams and filters have been hard coded before), posting of auto-annotations from fink science modules (which was missing, and highly requested), as well as the crossmatch of new candidates with recent gcn events, in an effort to further increase our chances (and decrease the number of actions performed) to discover the optical counterpart of an event using ZTF candidates.

However, we do know that this feature (similar to the one introduced in the image reduction PR that uses STDPipe) is not needed in SkyPortal core. Keeping this constraint in mind, we made sure that it is deactivated by default, and can simply be activated by adding the associated fink broker configuration in the `config.yaml`.

This will allow collaborations and individuals to ingest ZTF candidates - and additional science - in their instance of SkyPortal fairly easily using Fink. It will also prevent having numerous individual implementations (and therefore hardly maintained).
Moreover, in comparison with Kowalski which requires running the broker locally adding an additional level of complexity, using the centralized architecture of Fink which offloads the science and filtering from the Marshall could help individual populate their SkyPortal instance and perform science with less infrastructure needed.

This is obviously a WIP PR, and any comments and suggestions are more than welcome.

_PS: it is not as long as GitHub says, the largest file being an alert schema that can be used for testing_

EDIT 1: I added the crossmatch of the new candidates with the galaxies of SP's database, in an attempt to add a potential host galaxy (name, catalog, distance to it) to the annotations.


services/fink/requirements.txt
```diff
@@ -0,0 +1,5 @@
+fink-client
+fink-filters
+fink-utils
+pyspark
+pyarrow
```
profjsb, line 5: should all have dependency versions made more explicit, like pyarrow>=XX.X

services/fink/fink.py
```diff
@@ -0,0 +1,1420 @@
...
+from astropy.table import Table
+from astropy.visualization import (
+    AsymmetricPercentileInterval,
+    LinearStretch,
+    LogStretch,
+    ImageNormalize,
+)
+import datetime
+import gzip
+import io
```
profjsb, line 20: first set up imports (builtins go first, as per pep8)

services/fink/fink.py
```diff
@@ -0,0 +1,1420 @@
+import os
+
+import sqlalchemy as sa
+from sqlalchemy.orm import joinedload
+import time
+import pandas as pd
+import numpy as np
+import requests
+import yaml
```
profjsb, line 9: first set up imports (builtins go first, as per pep8)

services/fink/fink.py
```diff
@@ -0,0 +1,1420 @@
+import os
+
+import sqlalchemy as sa
+from sqlalchemy.orm import joinedload
```
profjsb, line 4: these go below builtins and above baselayer/SP

services/fink/fink.py
```diff
@@ -0,0 +1,1420 @@
...
+    galaxy_id: int or None
+        The id of the host galaxy
+    galaxy_distmpc: float or None
+        The distance to the host galaxy in Mpc
+    distance_to_galaxy: float or None
+        The distance to the host galaxy in arcsec
+    """
+
+    # POST MANGROVE CATALOG FOR THE CROSSMATCH
+    # DISCUSS WITH JULIEN PELOTON ABOUT ADDIND HIS CROSSMATCH IN THE ALERT PACKET
```
profjsb, line 861: remove before merging

services/fink/fink.py
```diff
@@ -0,0 +1,1420 @@
...
+            #     for group_user in group_users:
+            #         session.add(
+            #             UserNotification(
+            #                 user=group_user.user,
+            #                 text=f"Object {alert['objectId']} was found in GCN event(s): {', '.join(obj_in_events)} (new alert)",
+            #                 notification_type="obj_in_events",
+            #                 url=f"/source/{alert['objectId']}",
+            #             )
+            #         )
+            #     session.commit()
```
profjsb, line 1226: remove

profjsb (commented):
Looks good. If we go this route, there should be some testing (can do another PR for that if need be). Also, there seems to be a lot of baked in ZTF-specific flags/keys. Is there a way to make it more generic so others can ingest other streams with potentially different data schema?

Theodlz:
> Looks good. If we go this route, there should be some testing (can do another PR for that if need be). Also, there seems to be a lot of baked in ZTF-specific flags/keys. Is there a way to make it more generic so others can ingest other streams with potentially different data schema?

Hi @profjsb. Thanks a lot for your review! So I am thinking that in order to make it more general I should start by using Python classes so that I can write a much more generic Mixin class, on which I can build a child class which specific fink methods.
That would allow it to be extended much more easily for other brokers. However, that would mean that to use other brokers other services need to be added to skyportal main. So I would need to make it more generic.
Maybe what I could do is find a way to identify if an alert packet comes from Fink or another source. If it is another broker, I then can use the more general Mixin class and/or generic ZTF methods to post alerts.

Or maybe you meant that the issue here is that it is ZTF specific, and you want it to be more generic so that people can ingest alerts from other instruments or in general with a different schema? Then I guess I need to figure out how one can specify the fields of the alert that we need in SkyPortal, which is trickier.

Maybe I can allow the user to specify which fields to use in the config? And I can have a hardcoded list of fields for ZTF alerts that can be used by default without the user having to specify them.

Tell me what you think!

Theodlz:
Hello @profjsb. I made some suggestions regarding the different approaches I can take to make this more generic as you requested.

Theodlz:
Temporarily closing this PR to open it on my fork and do some testing

PR #3778, opened 2022-12-27
Follow-up request update notifications

mcoughlin:
This PR adds follow-up request update notifications.


skyportal/handlers/api/followup_request.py
```diff
@@ -1694,3 +1705,124 @@ def delete(self, default_followup_request_id):
...
+            if followup_request is None:
+                return self.error("Could not retrieve followup request.")
+
+            watcher = session.scalars(
+                FollowupRequestUser.select(self.current_user, mode="delete").where(
+                    FollowupRequestUser.user_id == self.current_user.id,
+                    FollowupRequestUser.followuprequest_id == followup_request_id,
+                )
+            ).first()
+            session.delete(watcher)
```
Theodlz, line 1820: Please verify that there was a result to this query, i.e that the user is actually a `watcher` of the followup request. If he isn't, it would make sense to return an error like `can't delete user {id} as he is not watching followup request {id} yet`.

skyportal/handlers/api/source.py
```diff
@@ -141,20 +141,25 @@ def get_source(
...
+                    joinedload(FollowupRequest.watchers),
+                ],
+            )
+            .where(FollowupRequest.obj_id == obj_id)
+            .where(FollowupRequest.status != "deleted")
         )
-        .where(FollowupRequest.obj_id == obj_id)
-        .where(FollowupRequest.status != "deleted")
-    ).all()
+        .unique()
```
Theodlz, line 160: Nice catch

skyportal/models/user_notification.py
```diff
@@ -655,20 +660,51 @@ def receive_after_flush(session, context):
...
+                        continue
+                    allocation_id = target.allocation_id
+                    allocation = session.scalars(
+                        sa.select(Allocation).where(Allocation.id == allocation_id)
+                    ).first()
+                    notification_user_ids = [
+                        allocation_user.user.id
+                        for allocation_user in allocation.allocation_users
+                    ] + [watcher.user_id for watcher in target.watchers]
+                    print(notification_user_ids)
```
Theodlz, line 693: remove print

static/js/components/WatcherButton.jsx
```diff
@@ -0,0 +1,125 @@
...
+        disabled={isSubmitting}
+        size="large"
+      >
+        <StarIcon />
+      </IconButton>
+    </Tooltip>
+  );
+};
+
+const ButtonExclude = (requestID, textMode) => {
```
Theodlz, line 48: The naming here is a little confusing for me. Looking at the code I get that exclude is called if the user is not a watcher yet, but it can be easily confused for the opposite: that the button is here to exclude a user that is already watching the followup. I would do it the other way, a button called WatchButton, and why not UnWatchButton

Theodlz: or what you did in the ducks as well, that would be good: addToWatchListButton, removeFromWatchListButton

Theodlz (commented):
Few minor comments. Also please consider wrapping your handler methods in try excepts for good measure and to avoid not catching any unexpected errors (and returning the error message so that we can debug if need be).

skyportal/handlers/api/followup_request.py
mcoughlin: good point.

skyportal/models/user_notification.py
mcoughlin: oops.

static/js/components/WatcherButton.jsx
mcoughlin: yeah I like those button names

mcoughlin:
@Theodlz For these simpler handlers, I prefer not having the try except so to some extent bugs are caught quicker since we aren't necessarily trolling the logs well.

Theodlz (approved):
LGTM :rocket:

PR #3781, opened 2022-12-30
upload photometry form

laurall974:
@Theodlz

### DIFF
--- a/static/js/components/UploadPhotometry.jsx
+++ b/static/js/components/UploadPhotometry.jsx
@@ -18,6 +18,9 @@ import makeStyles from "@mui/styles/makeStyles";
 import withStyles from "@mui/styles/withStyles";
 import { useForm, Controller } from "react-hook-form";
 import PapaParse from "papaparse";
+import { Grid } from "@mui/material";
+// eslint-disable-next-line import/no-unresolved
+import Form from "@rjsf/material-ui/v5";
 import Button from "./Button";
 
 import GroupShareSelect from "./GroupShareSelect";
@@ -52,6 +55,7 @@ const UploadPhotometryForm = () => {
   const [showPreview, setShowPreview] = useState(false);
   const [csvData, setCsvData] = useState({});
   const [successMessage, setSuccessMessage] = useState("");
+  const [selectedInstrument, updateSelected] = useState(1);
   const { id } = useParams();
   const {
     handleSubmit,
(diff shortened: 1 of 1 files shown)


Theodlz:
Hi @laurall974. Before I leave a review, let's start with some more general comments:

1. The form you are adding is quite substantial (pretty long that is). It would be better to create a dedicated jsx file for it. Then, on the current UploadPhotometry component (which consists of the whole page, containing the csv input), you can simply call your new component. That ensures that the code remains readable for both components, and that they do not interfere with each other. For instance here, you use some variables of the existing csv input for your form, when they are supposed to be 2 different things (except of course the instrumentList which is shared). So first, create a new component called something like `NewPhotometry.jsx` which will contain your - and only your - component. That will make it much cleaner and also much easier for development.
2. About your issue about the filters. What you did so far is use the validate method to set the selected instrument, and based on that your filter input has a default value. There are some issues with that (what is done in skyportal already is not always the right answer, you can always do better!). First, when using a state variable, the page/components are reloaded every time it changes. That's the reason why you might have noticed that the content of the Form is deleted when you set the instrument, which is an obvious problem. Luckily for you, rjfs provides a way to set the possible values of a column (so the enum, the default value, and so on) based on the value of another, which is exactly what you want to do here. You want the filters of the selected instrument to appear. To see an example of that, go to the NewShift.jsx component. There, you'll see a `dependencies` key in the schema. Here's of it works: you specify a column that will have other columns depend on it (in NewShift that is the localTime, a bool column that says 2 other inputs should be in UTC or local), and then you can specify the columns that will change (there being the startDate and endDate of a shift). For each columns, you set the values if there is an enum, and/or the default value (in that case, it sets the default of the startDate to the current UTC time if localTime is False, otherwise it sets it to the current local time).
3. You want to convert the date to mjd. You can find how to do that very easily on google.
4. You want to check in the validate similar things (if not exactly the same) to what is checked in the UploadPhotometry.jsx file. 
5. As you might have noticed, the exposure is not a proper column for a photometry in the DB. As discussed few weeks ago, instead of adding that column, you can simply use an existing "all in" column called metadata. It takes a dict of key values. So you can set the metadata as something like `{exposure: blablabla}`. You might want to add some extra verification for the exposure in the validate, to make sure that the string makes sense.
6. Don't hesitate to ask for help :rocket: , and to ask for clarification.

Theodlz:
Just so you can tell how much easier it gets when you create multiple components rather than add your code to an existing one (that is already quite complex): I coded my own basic photometry form, and without the submit and full validate method, it is only 115, and definitely much more readable.

PR #3783, opened 2022-12-30
GCN event test posts

mcoughlin:
This PR enables an event to be posted multiple times (given retries).


skyportal/tests/api/test_gcn.py
```diff
@@ -898,6 +898,7 @@ def test_confirm_reject_source_in_gcn(
         data=params,
         token=super_admin_token,
     )
+    print(data)
```
Theodlz, line 901: remove print

Theodlz (commented):
How will it change the results of the tests? It seems to me that the next line checks if the data['status'] is 'success', equivalent to verifying if the status is 200 right?

Theodlz (approved):
LGTM 🚀

PR #3805, opened 2023-01-01
Observation RA and dec

mcoughlin:
This PR adds download of RA and dec for observations.


Theodlz (approved):
LGTM! :rocket: Thanks for adding that, it's useful for sure. Also thanks for the several test improvements.

PR #3807, opened 2023-01-03
Notifications to active shifters

mcoughlin:
This PR adds notifications for requests to active shifters.

### DIFF
--- a/skyportal/models/user_notification.py
+++ b/skyportal/models/user_notification.py
@@ -8,6 +8,7 @@ from sqlalchemy import func
 
 from sqlalchemy import event, inspect
 import arrow
+import datetime
 import lxml
 import operator  # noqa: F401
 import requests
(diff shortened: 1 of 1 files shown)


skyportal/models/user_notification.py
```diff
@@ -837,3 +862,14 @@ def receive_after_flush(session, context):
...
                                     )
                                 )
+
+
+def users_on_shift(session):
+    users = session.scalars(
+        sa.select(ShiftUser).where(
+            ShiftUser.shift_id == Shift.id,
+            Shift.start_date <= datetime.datetime.now(),
+            Shift.end_date >= datetime.datetime.now(),
```
Theodlz, line 872: Is this where with comma seperated arguments performing an AND or OR operation? If this is an OR, it will notify users even if their shift hasn't started. I'm not sure so I prefer to ask.

Theodlz (commented):
One little question. Once that's clarified, it can get it.

Theodlz (approved):
LGTM

PR #3810, opened 2023-01-04
Sources page resets table if error occurs

KodeurFou:
We manage to prevent the page from crashing when a filter that has to be specified is deleted. The list goes back to default.

What we still have to do on this issue is to also delete all the filters' chips when the list is reset

(done by Adrien & Adrien)


mcoughlin:
@KodeurFou Also need to run
pre-commit install
pre-commit run --files static/js/components/SourceList.jsx
and remove the console.log statement.

and remove the dustmap files, you do not want to commit those.

Adiman007:
the last PR introduced a bug but we have a fix for it, i will make a new PR soon

mcoughlin (approved):
LGTM!

PR #3812, opened 2023-01-04
Filter sources labelled by current user

bfhealy:
This PR allows users to filter sources by whether the user has personally labelled or not labelled a source. This will allow users to easily pick up where they left off in the labelling process regardless of how their progress compares to other users.

A new 'by me' checkbox is added in the 'Which have been...' part of the filter table to facilitate this change (see below). Users will get an error if they select 'by me' without also selecting one of 'labelled' or 'not labelled'. If 'by me' is not checked, the filter will return the sources labelled or not labelled by any user, as it currently does.

Screen Shot 2023-01-04 at 4 11 58 PM

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -607,33 +607,19 @@ async def get_sources(
             followup_request_subquery, Obj.id == followup_request_subquery.c.obj_id
         )
     if has_been_labelled:
+        labels_query = SourceLabel.select(session.user_or_token)
         if current_user_labeller:
-            labels_subquery = (
-                SourceLabel.select(session.user_or_token)
-                .where(SourceLabel.labeller_id == user.id)
-                .subquery()
-            )
-            obj_query = obj_query.join(
-                labels_subquery, Obj.id == labels_subquery.c.obj_id
-            )
-        else:
-            labels_subquery = SourceLabel.select(session.user_or_token).subquery()
-            obj_query = obj_query.join(
-                labels_subquery, Obj.id == labels_subquery.c.obj_id
-            )
+            labels_query = labels_query.where(SourceLabel.labeller_id == user.id)
+        labels_subquery = labels_query.subquery()
+        obj_query = obj_query.join(labels_subquery, Obj.id == labels_subquery.c.obj_id)
     if has_not_been_labelled:
+        labels_query = SourceLabel.select(
+            session.user_or_token, columns=[SourceLabel.obj_id]
+        )
         if current_user_labeller:
-            labels_subquery = (
-                SourceLabel.select(session.user_or_token, columns=[SourceLabel.obj_id])
-                .where(SourceLabel.labeller_id == user.id)
-                .subquery()
-            )
-            obj_query = obj_query.where(Obj.id.notin_(labels_subquery))
-        else:
-            labels_subquery = SourceLabel.select(
-                session.user_or_token, columns=[SourceLabel.obj_id]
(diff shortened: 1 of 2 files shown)


mcoughlin:
@bfhealy Can you add a flag that only shows the "by me" if labelled is checked?

bfhealy:
@mcoughlin Will do, that’s a good idea.

bfhealy:
@mcoughlin I've just about figured it out - the only catch is that if the user checks both 'labelled' and 'not labelled' but then unchecks one of them, the current code will deactivate the 'by me' checkbox. Here's a gif of the current behavior.
labelled_checkboxes

mcoughlin:
@bfhealy cool! What does @guynir42 think?

skyportal/handlers/api/source.py
```diff
@@ -606,13 +607,33 @@ async def get_sources(
...
+                .subquery()
+            )
+            obj_query = obj_query.join(
+                labels_subquery, Obj.id == labels_subquery.c.obj_id
+            )
+        else:
+            labels_subquery = SourceLabel.select(session.user_or_token).subquery()
+            obj_query = obj_query.join(
+                labels_subquery, Obj.id == labels_subquery.c.obj_id
+            )
```
guynir42, line 623: Can we put only the `where` clause in the conditional and leave the rest of the code out of it?

skyportal/handlers/api/source.py
```diff
@@ -606,13 +607,33 @@ async def get_sources(
...
+                SourceLabel.select(session.user_or_token, columns=[SourceLabel.obj_id])
+                .where(SourceLabel.labeller_id == user.id)
+                .subquery()
+            )
+            obj_query = obj_query.where(Obj.id.notin_(labels_subquery))
+        else:
+            labels_subquery = SourceLabel.select(
+                session.user_or_token, columns=[SourceLabel.obj_id]
+            ).subquery()
+            obj_query = obj_query.where(Obj.id.notin_(labels_subquery))
```
guynir42, line 636: same as above. In fact, you can probably get the `if current_user_labeller` conditional after of the has_been/has_not_been conditionals.

skyportal/handlers/api/source.py
```diff
@@ -1756,6 +1777,13 @@ async def get(self, obj_id=None):
               type: boolean
             description: |
               If true, return only those objects which have not been labelled
+          - in: query
+            name: currentUserLabeller
+            nullable: true
+            schema:
+              type: boolean
+            description: |
+              If true and one of hasBeenLabeller or hasNotBeenLabelled is true, return only those objects which have not been labelled/not labelled by the current user. Otherwise, return results for all users.
```
guynir42, line 1786:
```suggestion
              If true and one of hasBeenLabeller or hasNotBeenLabelled is true, return only those objects which have been labelled/not labelled by the current user. Otherwise, return results for all users.
```

guynir42 (approved):
Minor comment. I like the overall design.

skyportal/handlers/api/source.py
bfhealy: Agreed; change made!

skyportal/handlers/api/source.py
bfhealy: I restructured the conditionals and tested to make sure the checkboxes still work as intended.

skyportal/handlers/api/source.py
bfhealy: Thanks!

PR #3813, opened 2023-01-04
remove timeout from find_element

guynir42:


guynir42:
@profjsb between new react and fixing the "find_element" part, it seems we are now green... although I can't rule out that some tests are still flaky, but we will find out in the next PRs.

PR #3814, opened 2023-01-05
WhatsApp alerts

aolrc-vinci:
Allows contact through WhatsApp.
We used Twilio to send the alerts.
The possibility to toggle said notifications in the user parameters was also added.


aolrc-vinci:
Updated code : the front end of the notifications tab is done but the back end part of the on shift/time slot option for the phone calls and whatsapp messages doesn't seem to be connecting on my end (might be due to my occasional issue with loading skyportal with an updated code) so I put the possible back end code as comments.

mcoughlin (approved):
LGTM!

PR #3815, opened 2023-01-05
Instruments page improvement

lydie10:
Hi,
This PR improves the Instruments page by adding a MUI Datatable with some additional columns (instruments type, band, telescopes latitude and longitude).
For now there is no filter functionality but there is a search bar searching by instrument name. However, it only finds exact matching results (we have to include uppercase characters in the search text and write the entire name).
Also, it displays all the rows in one page even though the rowsPerPage is set on 10.

We are currently trying to fix these small issues, but the main part is working.

Thanks for your feedback ! 🚀


lydie10:
We put back the add and modify forms when the user has the right permission.
We removed the pagination options in the table because there isn't one in the backend like for the sources page.

The search bar still only finds exact results without auto complete. The delete feature is still missing because we need to incorporate it in the MUI Datatable. Hopefully, we will do so in the following week.

@Theodlz @mcoughlin

mcoughlin (approved):
Clear improvement. LGTM!

PR #3818, opened 2023-01-06
Telescopes

aolrc-vinci:
We were able to display the entire list of telescopes (shown on the map) by default.
You can still select a telescope on the map and it will only display this one.
To reset the page and go back to the entire list, just click on the "Telescope List" button.

### DIFF
--- a/skyportal/handlers/api/analysis.py
+++ b/skyportal/handlers/api/analysis.py
@@ -281,20 +281,20 @@ class AnalysisServiceHandler(BaseHandler):
         group_ids = data.pop('group_ids', None)
         with self.Session() as session:
             if not group_ids:
-                groups = self.current_user.accessible_groups
-            else:
-                groups = (
-                    session.scalars(
-                        Group.select(self.current_user).where(Group.id.in_(group_ids))
-                    )
-                    .unique()
-                    .all()
+                group_ids = [g.id for g in self.current_user.accessible_groups]
+
+            groups = (
+                session.scalars(
+                    Group.select(self.current_user).where(Group.id.in_(group_ids))
                 )
-                if {g.id for g in groups} != set(group_ids):
-                    return self.error(
-                        f'Cannot find one or more groups with IDs: {group_ids}.',
-                        status=403,
-                    )
+                .unique()
+                .all()
+            )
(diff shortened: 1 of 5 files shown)


mcoughlin:
@aolrc-vinci  Also need to run
pre-commit install
pre-commit run --files static/js/components/Telescope.jsx

Theodlz (commented):
There is a lot of repetition here (all the part where you display the information). It is redundant with what is displayed in the lists that show when you click a marker on the map. It's best you avoid that by creating a method tho which you pass a telescope's data, and it creates the display of the ListItem. After you're done with that, it can go in.

aolrc-vinci:
Problem fixed @Theodlz I add a display function to avoid repeating the code

Theodlz:
Hi @aolrc-vinci, please always look at the github actions on your PRs. Here you can easily see that the linting of the file didn't work, meaning that you did not lint your files. Please do it so that I can merge your PR.

aolrc-vinci:
Hi I'm not sure I know how to do that I'm sorry can you tell me exactly what I need to do ?

mcoughlin:
@aolrc-vinci I left a note above.

Also need to run
```
pre-commit install
pre-commit run --files static/js/components/TelescopeInfo.jsx
```

aolrc-vinci:
I tried running these commands but the 'pre-commit' is not recognized as a valid term

aolrc-vinci:
Hi, I did the pre commit install and everything so the telescope list display should be good now.
I also pushed our attempt at a telescope search bar but as mentioned during our meetings and in the github issue on the subject @ManonFlt and I are unable to run the code since we are having troubles with our skyportal.
We will try to get help from the others but you can still have a look at the code if you want to.

mcoughlin:
@aolrc-vinci It looks like the list of instruments has disappeared. Maybe you need to rebase on the commits? Looks like it is behind main by many commits.

ManonFlt:
We just synced fork. Is it enough or do we need to do something else?
Also, Lydie tested our code this morning and she had the telescope list.
However the search bar is still not working.

mcoughlin:
@ManonFlt I suggest creating a PR with just the pieces you want committed.

PR #3823, opened 2023-01-07
No classifications are also non-classifications

mcoughlin:
This PR fixes the non-classification behavior (i.e. no classification is non-classification too). It also makes some small quality of life improvements (i.e. can specify whether the classification is an AND or OR) and choose to see the classification votes or not.


bfhealy (approved):
LGTM, thanks for fixing all these issues!

Issue #3653, opened 2022-11-03
Sorting by date saved running out after switching few pages in the SOURCES page


Adiman007:
@Theodlz @mcoughlin 
If you sort by most recent **date saved** on page 1 of [Sources](https://skyportal-icare.ijclab.in2p3.fr/sources) , the sorting will be reset after reaching the **3rd page** and if you want to sort from **page 3** you will be push back to **page 1**
![image](https://user-images.githubusercontent.com/80261099/199744825-2b31fd61-fb62-4af8-879b-c6d71ef00caf.png)
![image](https://user-images.githubusercontent.com/80261099/199744986-2eb8236c-8423-436d-8d64-8fb8838a1839.png)

PR #3830, opened 2023-01-10
fix the sorting reset bug Issue#3653

Adiman007:
@Theodlz @mcoughlin 
link to issue : https://github.com/skyportal/skyportal/issues/3653 
Signed-off-by: Adiman007 <adiman.bonnefoy@gmail.com>

### DIFF
--- a/static/js/components/SourceTable.jsx
+++ b/static/js/components/SourceTable.jsx
@@ -637,7 +637,6 @@ const SourceTable = ({
     switch (action) {
       case "changePage":
       case "changeRowsPerPage":
-        setQueryInProgress(true);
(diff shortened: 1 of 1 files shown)


Theodlz (approved):
Hi @Adiman007. Indeed it's not super clear why it fixes it. However, I do have a hypothesis. As you know in React changing a state variable causes a rerender of the page. It's likely that either a rerender or a useEffect is triggered multiple times or while the query is executing, resetting some of the MUI Datatable code that sets the filtering.

Anyway, the queryInProgress variable is just here to display a loading icon while you change page, and it's not very useful. I'd rather lose that minor UX feature and have a working filtering system.

LGTM!

PR #3834, opened 2023-01-11
Vote all classifications

mcoughlin:
This PR adds ability to up or downvote all classifications.

### DIFF
--- a/static/js/components/SourceTable.jsx
+++ b/static/js/components/SourceTable.jsx
@@ -413,20 +413,9 @@ const RenderShowClassification = ({ source }) => {
         disableTouchListener
         title={
           <>
-            <Button
-              key={source.id}
-              id="down_vote"
-              onClick={() => addVotes(downvoteValue)}
-            >
-              <ThumbDown color={downvoteColor} />
-            </Button>
-            <Button
-              key={source.id}
-              id="up_vote"
-              onClick={() => addVotes(upvoteValue)}
-            >
-              <ThumbUp color={upvoteColor} />
-            </Button>
+            <br />
+            <b>All Classifications:</b>
+            <br />
             <Button
               key={source.id}
               id="delete_classifications"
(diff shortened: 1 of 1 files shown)


static/js/components/SourceTable.jsx
```diff
@@ -334,9 +356,49 @@ const RenderShowClassification = ({ source }) => {
     });
   };
 
+  const addVotes = (vote) => {
+    source.classifications?.forEach((c) => {
+      dispatch(sourceActions.addClassificationVote(c.id, { vote })).then(
+        (result) => {
+          if (result.status === "success") {
+            dispatch(showNotification("Vote registered"));
```
bfhealy, line 364: Displaying "Vote registered" for each classification is accurate but can temporarily take up a lot of screen space. Is it worth adding an array to keep track of each "success" and, if votes are uniformly successful, displaying a single "Votes registered" notification?

static/js/components/SourceTable.jsx
```diff
@@ -347,6 +409,20 @@ const RenderShowClassification = ({ source }) => {
...
+            >
+              <ThumbDown color={downvoteColor} />
+            </Button>
+            <Button
+              key={source.id}
+              id="up_vote"
+              onClick={() => addVotes(upvoteValue)}
+            >
+              <ThumbUp color={upvoteColor} />
+            </Button>
```
bfhealy, line 425: I was experimenting with the thumbs up/down box for all classifications. Might the below box reduce unintentional clicks of the delete-all button?
Screen Shot 2023-01-11 at 10 55 34 AM

bfhealy (approved):
Thanks @mcoughlin, this feature works precisely as I hoped it would. Just a few thoughts about what we display on the frontend.

static/js/components/SourceTable.jsx
mcoughlin: @bfhealy It's a bit up to you. I expect you will be a power user of this feature, so whatever you think is best, feel free to push it.

static/js/components/SourceTable.jsx
mcoughlin: @bfhealy I pushed a change for this.

static/js/components/SourceTable.jsx
bfhealy: @mcoughlin Just pushed the new box - I think it will be beneficial to keep the three buttons more separated.

PR #3840, opened 2023-01-15
WIP: API Call tracking handler

mcoughlin:
This PR adds an API Call tracking handler.

Depends on: https://github.com/cesium-ml/baselayer/pull/348


guynir42 (commented):
Should add filtering options for (at least) the `created_at` and `user` and probably the payload size.

skyportal/handlers/api/api_stats.py
```diff
@@ -0,0 +1,99 @@
...
+                    end_date = str(arrow.get(end_date.strip()).datetime)
+                    stmt = stmt.where(APICall.created_at <= end_date)
+                if size:
+                    stmt = stmt.where(APICall.size >= size)
+                if username:
+                    user_query = sa.select(User).where(User.username == username)
+                    user_subquery = user_query.subquery()
+                    stmt = stmt.join(
+                        user_subquery, APICall.user_id == user_subquery.c.id
+                    )
```
guynir42, line 96: I'm not sure how much time is actually spent on just defining the select statement (before executing it) but it may be simpler to just add these filters to `stmt` before the loop on URIs, and only add the `where(APICall.uri==uri)` and the count function inside the loop.

guynir42 (commented):
I like the new addition. Would think about pulling some of that stuff out of the loop:

guynir42:
@mcoughlin do you remember what we got stuck on with this PR?

mcoughlin:
Token sessions... I think? @Theodlz's favorite topic ;)

mcoughlin:
@Theodlz I think some version of this would be also valuable for tracking app health / usage, but this was implemented poorly.

Theodlz:
@mcoughlin does sentry kinda covers these needs already? Could log more, but I think it's already helping us a lot. Maybe we could close this one?

PR #3663, opened 2022-11-07
WIP: spectral cube interactions

mcoughlin:
This PR looks at basic spectral cube interactions.

PR #3843, opened 2023-01-16
WIP: Spectral cube Analysis Service

Theodlz:
This service adds an analysis service running PySEDM, as well as more capabilities to the analysis framework, such as posting files as analysis parameters.

Inspired (heavily) by #3663 from @mcoughlin

image

image

image


Yashvi-Sharma:
@mcoughlin Flux (AU) vs Wavelength for spectrum plot? For the cube image, the alignment should be such that Dec is the Y axis and RA is the X axis but I think the units are in spaxels.

Theodlz:
Thanks @Yashvi-Sharma !
@mcoughlin what about:
image

mcoughlin:
@Theodlz What else is needed before this can go in?

Theodlz:
@Yashvi-Sharma is the title of the last plot I posted correct? I am not sure whether or not the spectrum given in spectrum.data is for the selected centroid, or for the given image in general?

Yashvi-Sharma:
@Theodlz The plot looks good. The spectrum is what's extracted from the whole shaded area around the centroid.

Theodlz:
Thanks! @Yashvi-Sharma 

Then @mcoughlin I guess some tests and it can go in. I can work on that tomorrow so that it can be ready for a last review in your morning?

Theodlz:
Hi @mcoughlin, sorry for the delay. Just to let you know I will commit some extra code to support .fz files today as it can always be useful.

@profjsb what do you think is necessary to test? I found that other analysis services other than the demo one (like `ngsf_analysis_service`) are not tested for now and that the frontend itself is not tested so extensively either. In general, as you built the analysis framework your input would be greatly appreciated (especially about how to handle types of parameters when adding an analysis service, for which I had to modify the analysis handler)!

Theodlz:
Hi @profjsb. Is there any of my changes on the Analysis framework that you think should be done differently or even reversed?

requirements.txt
```diff
@@ -79,7 +79,7 @@ selenium-requests==2.0.0
 afterglowpy==0.7.3
 gcn-kafka==0.3.0
 confluent_kafka==1.9.2 # dependency of gcn-kafka: pinning to understand memory effects on GCNEvents
-iminuit==2.18.0
+iminuit==1.5.4
```
profjsb, line 82: this seems like a large downgrade. Can you explain why this is needed?

profjsb (commented):
Looks good. You've changed some of the front end so probably work looking at updating the analysis tests in `tests/frontend/sources_and_followup_etc/test_sources.py`

If you want to test your code locally (using `make run_testing`) you can add the service port to `test_config.yaml`.

Might be useful to add a section about what services are available out of the box with skyportal in `doc/analysis.md`.

requirements.txt
Theodlz: > Looks good. You've changed some of the front end so probably work looking at updating the analysis tests in `tests/frontend/sources_and_followup_etc/test_sources.py`
> 
> If you want to test your code locally (using `make run_testing`) you can add the service port to `test_config.yaml`.
> 
> Might be useful to add a section about what services are available out of the box with skyportal in `doc/analysis.md`.

Hi @profjsb, thanks a lot for the review. I do agree that some documentation would be very useful, as well as maybe some tests. I'll work on that.

Theodlz:
> Looks good. You've changed some of the front end so probably work looking at updating the analysis tests in `tests/frontend/sources_and_followup_etc/test_sources.py`
> 
> If you want to test your code locally (using `make run_testing`) you can add the service port to `test_config.yaml`.
> 
> Might be useful to add a section about what services are available out of the box with skyportal in `doc/analysis.md`.

Hi @profjsb, thanks a lot for the review. I do agree that some documentation would be very useful, as well as maybe some tests. I'll work on that.

requirements.txt
Theodlz: pysedm unfortunately doesn't pin requirements at the moment, and upgrading iminuit breaks it. What I can do is play around with different versions between 1.5.4 and 2.18.0 to figure out the latest version that works. I opened an issue on their repo and they will add a requirements.txt file soon.

Theodlz:
One more thing @profjsb, other than the frontend tests to test the frontend changes I made, do you think it would make sense to add tests for this particular analysis service? It is the only one that takes a file as input, so it might be worth testing.

profjsb:
Yes, probably a good idea to test the file upload functionality.

profjsb (approved):
Go ahead and merge after confirming that the frontend test failure is unrelated.

PR #3844, opened 2023-01-17
Allocations page improved

Carla802:
Hi,
This PR improves the Allocations page by replacing the Allocations list and the Default Observation Plans list by MUI datatables. 
We didn't modify the third list (Default Survey Efficiency) because in the demo data even with admin permissions, there is no data so we can't test our code. 
The add and modify forms are still the same, and the delete option is inside the datatables, in the last column.
Thanks for your feedback ! 🚀
@mcoughlin @Theodlz


mcoughlin (approved):
Really nice! You should be able to make a default plan just in the drop-down on the allocations page. You can maybe knock that out quick in the next PR.

PR #3848, opened 2023-01-17
Do not propagate ra/dec for ZTF ToO requests

robertdstein:
The RA/Dec values are not required for ToO requests. However, if there is a mismatch between the the ra/dec and field, it will generate errors. Remove these values from the request eliminates the possibility. 

As an additional gain, it means that multiple fields can now be submitted for a triggered payload via a source page without changing the front end at all.


mcoughlin:
I think we want to leave them in (but fix them) right? Just because the ZTF system doesn't need them, it is pretty annoying to not have them in, for example, the json download.

robertdstein:
I wasn't aware of any downsides to this change (e.g I'm not sure what the json you are referring to is), but for sure there are probably downstream consequences I did not consider.  Maybe we could discuss this issue on the next Fritz call?

mcoughlin:
I will work on adding the RA/Dec to pass directly in the case of these source page triggers.

PR #3849, opened 2023-01-17
Filter based on GCN properties

mcoughlin:
This PR enables filtering based on GCN properties.


skyportal/models/observation_plan.py
```diff
@@ -58,6 +58,11 @@ class DefaultObservationPlanRequest(Base):
         doc="Content of the default observation plan request.",
     )
 
+    filters = sa.Column(
+        psql.JSONB,
+        doc="Filters of the default observation plan request.",
+    )
```
guynir42, line 64: what kind of filters are we talking about here?

static/js/components/DefaultObservationPlanTable.jsx
```diff
@@ -127,25 +127,25 @@ const DefaultObservationPlanTable = ({
...
 
-  const renderEnd_Date = (dataIndex) => {
+  const renderFilters = (dataIndex) => {
     const default_observation_plan = default_observation_plans[dataIndex];
 
     return (
       <div>
         {default_observation_plan
-          ? default_observation_plan.payload.end_date
+          ? default_observation_plan.payload.filters
```
guynir42, line 148: You removed the `payload` in line 135 but you use it here? Is that correct?

guynir42 (approved):
I'm not sure what the filters are. Is that just a dictionary that should replace whatever conditions like start_time?

static/js/components/DefaultObservationPlanTable.jsx
mcoughlin: So basically the start_date and end_date don't mean anything in these default plans as they get replaced.

skyportal/models/observation_plan.py
mcoughlin: Adding some extra text here to be more explicit.

skyportal/models/observation_plan.py
```diff
@@ -60,7 +60,7 @@ class DefaultObservationPlanRequest(Base):
 
     filters = sa.Column(
         psql.JSONB,
-        doc="Filters of the default observation plan request.",
+        doc="Filters to determine which of the default observation plan request. get executed for which events",
```
guynir42, line 63:
```suggestion
        doc="Filters to determine which of the default observation plan request get executed for which events",
```

PR #3851, opened 2023-01-18
Recurring API calls

mcoughlin:
This PR adds recurring API calls.


profjsb:
Can you give some more information in the PR what this API/frontend is meant to do? Perhaps some new documentation to explain it and show an example?

Theodlz:
So one thing that is needed is some sort of basic logging of the result of the API calls, at least their status, and possibly the error message if not 200. If users can program these recurring API calls but have no way to track if they are successful or not, it's gonna be impossible for them to fix their own mistakes in the API queries they create. 
Also, maybe consider a counter of successive failures. If the counter reaches a certain maximum you set (i think that maximum should be hardcoded, or at least limited to a certain nb if the user can choose it), the recurring_api call is 'deactivated' (that state can also be saved as a column). This should prevent failing queries to run in repeat, and allow for better debugging from the users that use this new feature.

mcoughlin:
Thanks @Theodlz. I think I implemented things as suggested. Lemme know what you think.

Theodlz (changes requested):
Hi @mcoughlin, it's much better already. It worked well for me, but I have some changes in mind.

- First, not logging on 200 but only on failure, or at least not logging the response of the request. Let's say that someone does a GET request on sources without any params, you'll end up following up the logs. You can log some of the data of the response if you want, but you need to at least limit the nb of char printed.
- Second (and related to the first), what about the GET requests here? I'm not sure it's useful to allow those if the result is anyway not stored and not made available to the user. I think you need to either not allow get, or add a columns that can hold the results. The results of some API requests can be very large, so to address those I would: either consider it as a failure and set the recurring API call to deactivated if the response is too large, sending the user a notification explaining why. Or, maybe better is save the results to disk as joblib files like in the analysis framework. I think this is an important decision, to avoid having to change it later on (meaning having to transition to disk like we needed for the comments).
- What about the PUT requests? Why would a user repeat the same modification of an object over an over? Once should be enough no? In general considering the current feature which is creating unique API calls and not a series of them, I'm not sure that queries other than the POSTS or PUT (for the handlers where the PUT acts as both the POST and PATCH) are relevant.

Let me know what you think.

mcoughlin:
@Theodlz This is a great point. For the use cases we have considered (i.e. grabbing the observations, annotations, catalogs, etc., these are all post requests, so for now, we can maybe have anything that isn't a POST fail?

Theodlz:
I agree. I think the bes thing is to have an ALLOWED_METHODS variable in the handler's file, and in the post method of the handler you send back a failure if the user tried to add anything else than a POST api query. On the frontend, either we don't change anything or if you want to be a lil fancy while anticipating the change of the ALLOWES_METHODS variable, you can add it to the config handler so that you can display a frontend error if the method the user specified in the form is not allowed yet.

mcoughlin:
@Theodlz updated.

static/js/components/NewRecurringAPI.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+import {
+  submitRecurringAPI,
+  fetchRecurringAPIs,
+} from "../ducks/recurring_apis";
+
+dayjs.extend(utc);
+
+dayjs.extend(utc);
+
```
Theodlz, line 18: Already line 15, you can remove line 17 I believe

static/js/components/RecurringAPIPage.jsx
```diff
@@ -0,0 +1,206 @@
...
+  const classes = useStyles();
+  const dispatch = useDispatch();
+
+  const permission =
+    currentUser.permissions?.includes("System admin") ||
+    currentUser.permissions?.includes("Manage Recurring APIs");
+
+  useEffect(() => {
+    const getRecurringAPIs = async () => {
+      // Wait for the analysis services to update before setting
```
Theodlz, line 153: Remove unrelated comment

Theodlz (approved):
Last few comments, but LGTM

PR #3855, opened 2023-01-19
No vote

### LINKED ISSUE #3854: Option in API to avoid upvoting posted classifications
Usually it is ideal to add an automatic upvote when a user posts a classification. However, if a user like myself neutrally posts automated classifications for others to review during active learning, this automatic upvote may be undesirable. I have no a priori disposition for/against these ML classifications, so it would make sense to add a flag in the API that, if set, avoids upvoting them. All posted classifications via the frontend should still be automatically upvoted.

mcoughlin:
This PR makes it possible to not vote.

Closes #3854.

### DIFF
--- a/skyportal/handlers/api/classification.py
+++ b/skyportal/handlers/api/classification.py
@@ -85,10 +85,17 @@ def post_classification(data, user_id, session):
     )
     session.add(classification)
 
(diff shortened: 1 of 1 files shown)


bfhealy (approved):
LGTM!

PR #3857, opened 2023-01-20
Galactic plane exclusion

mcoughlin:
This PR adds a checkbox for galactic plane exclusion.

### DIFF
--- a/skyportal/facility_apis/observation_plan.py
+++ b/skyportal/facility_apis/observation_plan.py
@@ -385,6 +385,10 @@ def generate_plan(
                 [request.payload["exposure_time"]]
                 * len(request.payload["filters"].split(","))
             ),
+            # avoid the galactic plane?
+            'doAvoidGalacticPlane': requests[0].payload.get("galactic_plane", False),
+            # galactic latitude to exclude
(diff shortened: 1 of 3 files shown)


skyportal/facility_apis/observation_plan.py
```diff
@@ -385,6 +385,10 @@ def generate_plan(
                 [request.payload["exposure_time"]]
                 * len(request.payload["filters"].split(","))
             ),
+            # avoid the galactic plane?
+            'doAvoidGalacticPlane': requests[0].payload.get("galactic_plane", False),
+            # galactic latitude to exclude
+            'galactic_limit': request.payload["galactic_latitude"],
```
guynir42, line 391: I'm not a fan of the mixed use of camel case and underscores.

guynir42 (approved):
Minor comment

skyportal/facility_apis/observation_plan.py
mcoughlin: Yeah all inputs are underscores, unfortunately, gwemopt is camel case so we are stuck with that :)

PR #3858, opened 2023-01-20
Spectrum file is required for upload

mcoughlin:
This PR adds validation that you have uploaded a spectrum file.


guynir42 (approved):
that was an easy one!

PR #3861, opened 2023-01-22
Dockerfile - Uncommented SNID and STDPipe dependencies

Theodlz:
As having SNID and STDPipe installed in SP does not cause any problem with existing features, and facilitates SP's deployment and testing (as well as its maintenance by avoiding a duplicate Dockerfile in the extension folder of Fritz), it seems like a good decision to uncomment the dependencies install lines from the Dockerfile.


mcoughlin (approved):
Let's give this a try and see if any issues arise.

PR #3862, opened 2023-01-22
Run Docker GA for PRs with "docker" label

Theodlz:

PR #3863, opened 2023-01-22
Fix issue with Docker workflow on push

Theodlz:
#3862 introduced a bug by only triggering action on label, and not running it on push. This PR attempts to make both conditions work.

### DIFF
--- a/.github/workflows/test_docker_build.yaml
+++ b/.github/workflows/test_docker_build.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


Theodlz:
@mcoughlin I think this will fix it. It did not trigger it on push of my branch (which is expected, I was afraid it would now run it on every push but it does not), only when I added the label. I can't try it out unless we actually merge, so I think we should merge it as soon as possible.

mcoughlin (approved):
Let's try it!

PR #3864, opened 2023-01-23
Photometry annotations

mcoughlin:
This PR adds photometry annotations.


guynir42:
I would advise we put this up for debate to the group tomorrow. I would say that's a good first step before writing all this new code but...

mcoughlin:
@guynir42 You know me, I do before I think ;)

skyportal/tests/api/test_annotations_on_photometry.py
```diff
@@ -0,0 +1,420 @@
...
+        'POST',
+        f'photometry/{photometry_id}/annotations',
+        data={
+            'origin': 'kowalski',
+            'photometry_id': photometry_id,
+            'data': {'ACAI_class': 'type Ia'},
+            'group_ids': [public_group.id, public_group2.id],
+        },
+        token=annotation_token_two_groups,
+    )
```
guynir42, line 172: I assume this is the same behavior as we had for spectra. Do we want to allow users to view an annotation without being able to view the photometry point? This is a little problematic... we generally don't even let a user know if a photometry ID exists if they don't have permission to view it.

guynir42 (approved):
I have minor comments. Looks good overall. 
The main concern I have is about the permissions issue (see below).

skyportal/tests/api/test_annotations_on_photometry.py
```diff
@@ -0,0 +1,420 @@
...
+        f'photometry/{photometry_id}/annotations',
+        data={
+            'photometry_id': photometry_id,
+            'data': {'offset_from_host_galaxy': 1.5},
+            'group_ids': [public_group.id],
+        },
+        token=annotation_token,
+    )
+    assert status in [400]
+    assert 'origin must be specified' in data["message"]
```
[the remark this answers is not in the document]
mcoughlin: Done.

skyportal/tests/api/test_annotations_on_photometry.py
mcoughlin: So right now photometry read implies that you have obj and photometry available for read. Do you have a more particular concern?

guynir42: I'm just worried that if you can read an annotation on a photometry point that is inaccessible to you, you will get information that the photometry point exists.

mcoughlin: But this is impossible because the permissions for read on annotation implies a permission on photometry.

guynir42: I am confused because of the code comment above: `Both tokens should be able to view this annotation, but not the underlying photometry`

PR #3869, opened 2023-01-25
Label when adding classification

mcoughlin:
This PR adds an option to turn off label when adding classification.


bfhealy (approved):
LGTM, thanks!

PR #3870, opened 2023-01-26
GCN notifications should not require notices

mcoughlin:
This PR fixes GCN notifications such that they do not require notices.

### DIFF
--- a/skyportal/models/user_notification.py
+++ b/skyportal/models/user_notification.py
@@ -589,24 +589,19 @@ def add_user_notifications(mapper, connection, target):
                 is not None
             ):
                 if is_gcnevent and (pref is not None):
+                    event = session.scalars(
+                        sa.select(GcnEvent).where(GcnEvent.dateobs == target.dateobs)
+                    ).first()
+                    notice = event.gcn_notices[-1]
                     if "gcn_notice_types" in pref["gcn_events"].keys():
                         send_notification = True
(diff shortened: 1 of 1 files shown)


Theodlz (commented):
Looks good to me. However it seems that we are notifying if any of the notices (not just the new one on which we want to notify) is in the user preferences. Meaning that even if a new notice doesn't have a type selected by the user, he will be notified because a previous one had the right notice type. That seems a little weird to me. I think we should change that piece of code to just notify if the most recent notice of the event (the one we notify on) is in his preferences. I can't highlight it because the code was already there, but its from line 592 to 607. LIne 609 we grab the last notice, so we can move that up and use it in that updated logic.

Here is the code snippet that I think needs to change:
```
if "gcn_notice_types" in pref["gcn_events"].keys():
                        send_notification = True
                        event = session.scalars(
                            sa.select(GcnEvent).where(
                                GcnEvent.dateobs == target.dateobs
                            )
                        ).first()
                        if not any(
                            [
                                gcn.NoticeType(notice.notice_type).name
                                in pref['gcn_events']['gcn_notice_types']
                                for notice in event.gcn_notices
                            ]
                        ):
                            send_notification = False
                            continue
```

mcoughlin:
@Theodlz Great point.

Theodlz (approved):
LGTM 🚀

PR #3872, opened 2023-01-26
explicitly close postgresql sessions

mcoughlin:
This PR explicitly closes postgresql sessions.


guynir42 (approved):
Ahhh yes this is important. I was going to ask you about that because you generate them without a context manager. 

Are these functions all called inside an IOLoop.run_on_executor?

mcoughlin:
Yes exactly. So this should help... I hope.

PR #3873, opened 2023-01-26
astroscrappy

mcoughlin:
This PR pins the astroscrappy requirement.


mcoughlin:
Pinning package. Self merging.

PR #3874, opened 2023-01-26
Observation plan sessions

mcoughlin:
This PR adds session numbers for observation plans.


guynir42 (approved):
This is fine but you can also hot-patch staging if you want.

PR #3877, opened 2023-01-27
Plotting groups

mcoughlin:
This PR fixes bug as groups are already dictionaries.


mcoughlin:
Bug fix. Self merging.

PR #3881, opened 2023-01-27
Upgrade to SQLAlchemy 2.0

mcoughlin:
This PR pairs with https://github.com/cesium-ml/baselayer/pull/349 to upgrade to SQLAlchemy 2.0

### DIFF
--- a/skyportal/facility_apis/lco.py
+++ b/skyportal/facility_apis/lco.py
@@ -509,8 +509,11 @@ def download_observations(request_id, ar):
         Group,
     )
 
-    Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
-    session = Session()
+    Session = scoped_session(sessionmaker())
+    if Session.registry.has():
+        session = Session()
+    else:
+        session = Session(bind=DBSession.session_factory.kw["bind"])
 
     try:
         req = session.scalars(
@@ -539,6 +542,8 @@ def download_observations(request_id, ar):
         session.commit()
     except Exception as e:
         return log(f"Unable to post data for {request_id}: {e}")
+    finally:
+        session.close()
 
 
 class LCOAPI(FollowUpAPI):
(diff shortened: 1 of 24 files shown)


skyportal/facility_apis/lco.py
```diff
@@ -539,6 +542,8 @@ def download_observations(request_id, ar):
         session.commit()
     except Exception as e:
         return log(f"Unable to post data for {request_id}: {e}")
+    finally:
+        session.close()
```
guynir42, line 546: Do we know the difference between `session.close()` and `session.remove()`?

skyportal/handlers/api/analysis.py
```diff
@@ -43,8 +41,6 @@
 
 _, cfg = load_env()
 
-Session = scoped_session(sessionmaker(bind=DBSession.session_factory.kw["bind"]))
```
guynir42, line 46: Just don't need this anywhere...?

skyportal/handlers/api/observation_plan.py
```diff
@@ -2078,7 +2078,7 @@ def observation_simsurvey(
         Optional parameters to specify the injection type, along with a list of possible values (to be used in a dropdown UI)
     """
 
-    session = Session()
+    session = Session(bind=DBSession.session_factory.kw["bind"])
```
guynir42, line 2081: It seems very surprising to me that there would be a difference between setting the `bind` at the `sessionmaker()` level vs. the `Session()` level. Is there any reason given for why this has to be this way in 2.0?

guynir42 (approved):
If the tests pass, and if we really understand where these differences come from, it looks fine. But I think we should be careful about making changes we don't fully understand.

skyportal/facility_apis/lco.py
mcoughlin: You close a session = Session() and remove a Session().

skyportal/handlers/api/observation_plan.py
mcoughlin: best I can tell they are making it more explicit, and our out of session calls (async calls) requires this).

PR #3882, opened 2023-01-29
Fix Gcn Event notification

Theodlz:


mcoughlin (approved):
LGTM. Thanks for cleaning that up.

PR #3887, opened 2023-01-31
End of automated plan end of night

mcoughlin:
This PR makes automated plan finish at end of next night.


guynir42 (approved):
This is only in case where you don't specify an end time, right?

What happens if you need a plan for multiple nights?

mcoughlin:
@guynir42 Yeah so you can always just extend the end time to cover 48 hours, etc., but this is just so the automated plan covers the remaining hours of the night, which is what we want for quick triggered.

guynir42:
OK, that makes sense. Please try to add this context when writing the PR description...

PR #3908, opened 2023-02-01
Only make plan for fields in localization

mcoughlin:
This PR speeds up observation plan by only including fields in localization (otherwise gwemopt has to do that, which is slower).


Theodlz (approved):
LGTM. Anyway these things are better tested on a larger DB i.e in production. You can merged once green. I believe the GAs are delayed because of the dependabot, but this looks good and can be merged as soon as all the GAs are green

PR #3909, opened 2023-02-01
rjsf 4.2.3

mcoughlin:
This PR pins rjsf to 4.2.3


mcoughlin (commented):
Pinning package. Self-merging.

PR #3917, opened 2023-02-04
Delete spectrum annotations

mcoughlin:
This PR fixes delete button for spectrum annotations.

### DIFF
--- a/static/js/components/AnnotationsTable.jsx
+++ b/static/js/components/AnnotationsTable.jsx
@@ -20,6 +20,7 @@ import { getAnnotationValueString } from "./ScanningPageCandidateAnnotations";
 import Button from "./Button";
 
 import * as sourceActions from "../ducks/source";
+import * as spectraActions from "../ducks/spectra";
 
 dayjs.extend(relativeTime);
 dayjs.extend(utc);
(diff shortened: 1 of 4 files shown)


static/js/components/AnnotationsTable.jsx
```diff
@@ -90,15 +91,27 @@ const AnnotationsTable = ({ annotations, spectrumAnnotations = [] }) => {
...
   };
 
   // Curate data
+  annotations?.forEach((annotation) => {
+    annotation.type = "source";
+  });
+  spectrumAnnotations?.forEach((spectrumAnnotation) => {
+    spectrumAnnotation.type = "spectrum";
+    annotations?.push(spectrumAnnotation);
+  });
```
guynir42, line 113: I'm not sure this is the best place to mark these. Maybe in the duck that puts them into "annotations" and "spectrumAnnotations"? That way you will have that "type" in any component.

mcoughlin: @guynir42 to confirm: the handlers right?

guynir42: In either the handlers or in the duck where you change the store's state, if you want this to only exist in the front end.

mcoughlin: I added it to the handlers.

guynir42 (approved):
Yeah, that works too!

PR #3918, opened 2023-02-05
Upload-only Analysis

profjsb:
This PR enabled an upload-only endpoint for analysis services. 

After creation of an analysis service (with the parameter `upload_only`=True), this analysis service id can be used for authenticated uploads of analysis results from an external 3rd party. It is the responsibility of the 3rd party to maintain any provenance for the analysis (ie. what specific data was used to run the analysis), to delete outdated versions of the analysis on a given `obj`, and to avoid leaking sensitive data available to the 3rd party but not available to the groups that a specific analysis is attached to. By default we place a cap of 50 completed analyses per object per user/token.

As an example, a 3rd party analysis on the object named `MySourceName` using service number `service_id` can be uploaded as such:

```python
params={
  "show_parameters": True,
  "analysis": {
    "plots": ...,
    "inference_data": ...,
    "results": {
      "format": "json",
      "data": {
        "external_provenance": {
          "hash": "23baef56",
          "spectrum_ids": [34, 68, 125],
          "model": "jsbFitter23"
        },
        "classification": "DY Per",
        "classification_proba": 0.94
      }
    }
  }
}
url = f"http://<url>:5000/api/obj/MySourceName/analysis_upload/{service_id}"
r = requests.post(url, headers={'Authorization':   'token uuid'}, json=params)
```
here the uuid is the token with `Run Analyses` permission.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "redux-thunk": "2.4.2",
     "regl": "2.1.0",
     "semver": "7.3.7",
+    "sprintf-js": "1.1.2",
     "style-loader": "2.0.0",
     "styled-components": "5.3.5",
     "timezone": "1.0.23",
--- a/skyportal/handlers/api/analysis.py
+++ b/skyportal/handlers/api/analysis.py
@@ -11,6 +11,7 @@ from requests_oauthlib import OAuth1
 from requests.auth import HTTPBasicAuth, HTTPDigestAuth
 from tornado.ioloop import IOLoop
 from marshmallow.exceptions import ValidationError
+from sqlalchemy import func, select
 from sqlalchemy.exc import IntegrityError
 
 from baselayer.app.access import auth_or_token, permissions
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -91,6 +91,7 @@
     "redux-thunk": "2.4.2",
     "regl": "2.1.0",
     "semver": "7.3.7",
+    "sprintf-js": "1.1.2",
```
profjsb, line 94: missing dependency

skyportal/models/analysis.py
```diff
@@ -403,7 +403,7 @@ def check_path_string(string, allow_slashes=False):
         else:
             reg = RE_NO_SLASHES
 
-        if not reg.match(string):
+        if string is None or not reg.match(string):
             raise ValueError(f'Illegal characters in string "{string}". ')
```
mcoughlin, line 407: Maybe update the error message here?

skyportal/handlers/api/analysis.py
```diff
@@ -1310,3 +1346,196 @@ async def get(
                 )
 
             return self.error("No data found for this Analysis.", status=404)
+
+
+class AnalysisUploadHandler(BaseHandler):
```
mcoughlin, line 1351: Should this be somehow named UploadOnly or similar? To differentiate it more somehow

skyportal/handlers/api/analysis.py
```diff
@@ -1310,3 +1346,196 @@ async def get(
...
+
+            status_message = data.get("message", "")
+
+            if analysis_resource_type.lower() == 'obj':
+                obj_id = resource_id
+                stmt = Obj.select(self.current_user).where(Obj.id == obj_id)
+                obj = session.scalars(stmt).first()
+                if obj is None:
+                    return self.error(f'Obj {obj_id} not found', status=404)
+                data["obj_id"] = obj_id
```
mcoughlin, line 1474: Why are both of these needed?

skyportal/handlers/api/analysis.py
```diff
@@ -1310,3 +1346,196 @@ async def get(
...
+                    ObjAnalysis.obj_id == obj_id
+                )
+                stmt = stmt.where(ObjAnalysis.author == author)
+                stmt = stmt.where(ObjAnalysis.status == "completed")
+
+                total_matches = session.execute(
+                    select(func.count()).select_from(stmt)
+                ).scalar()
+
+                if (
```
mcoughlin, line 1490: This is a very good thing.

mcoughlin (approved):
A few small questions, but looks very nice.

skyportal/handlers/api/analysis.py
profjsb: was a cut and paste of legacy code...cleaned it up (not needed)

skyportal/handlers/api/analysis.py
profjsb: Good idea.

PR #3919, opened 2023-02-05
No spaces in names

mcoughlin:
This PR prevents spaces in source names, which break SP.


Theodlz (approved):
LGTM

Theodlz:
Maybe a regex to remove all kind of space chars would be great. But thats already better

guynir42:
I'm not sure but I think I've seen spaces in source names before (maybe this was from the weird names coming out of the asteroid detection program). We should check that we don't have any such examples in the DB already.

PR #3920, opened 2023-02-05
LocalizationTile index

mcoughlin:
This PR adds a multi-column index on the localizationtiles table.


skyportal/models/localization.py
```diff
@@ -235,6 +235,16 @@ class LocalizationTile(Base):
...
 
 
+LocalizationTile.__table_args__ = (
+    sa.Index(
+        'localizationtile_id_healpix_index',
+        LocalizationTile.id,
+        LocalizationTile.healpix,
+        unique=True,
+    ),
+)
```
guynir42, line 245: I think you can have the `__table_args__` added directly into the class definition, without having to call `LocalizationTile` and append it like that.

guynir42 (approved):

skyportal/models/localization.py
mcoughlin: @guynir42 Our other indices are done this way... maybe because they arent defined yet?

guynir42: Usually you define things like this if you need other tables to already exist to refer to them. It looks to me like this can work just as well inside the `class` block.

mcoughlin: Yeah LocalizationTile.id is not defined.

PR #3921, opened 2023-02-05
remove user get

mcoughlin:
This PR removes the "get" call throwing sqlalchemy warnings in a few services.


Theodlz (approved):
LGTM. I did notice those ugly warnings the other day. Thanks for fixing that!

PR #3922, opened 2023-02-05
bump arviz version; fix some marshmallow warnings

profjsb:
fix the `RemovedInMarshmallow4Warning: The 'default' argument to fields is deprecated. Use 'dump_default' instead.` warning


mcoughlin (approved):
Thanks @profjsb! LGTM.

PR #3925, opened 2023-02-07
Set host for recurring API and reminder

mcoughlin:
This PR specified host explicitly in recurring API and reminder.


guynir42 (approved):
LGTM

PR #3926, opened 2023-02-07
Associated resource in comment list

mcoughlin:
This PR specifies associated resource in comment list, such that the correct function can be called for delete.

### DIFF
--- a/static/js/components/CommentList.jsx
+++ b/static/js/components/CommentList.jsx
@@ -267,9 +267,6 @@ const CommentList = ({
 
   if (associatedResourceType === "object") {
     comments = obj.comments;
-    comments?.forEach((comment) => {
-      comment.resourceType = associatedResourceType;
-    });
     if (
       includeCommentsOnAllResourceTypes &&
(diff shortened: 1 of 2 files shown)


static/js/components/CommentList.jsx
```diff
@@ -285,23 +291,35 @@ const CommentList = ({
...
+      comment.resourceType = associatedResourceType;
+    });
   } else if (associatedResourceType === "shift") {
     if (shift_id === null) {
       throw new Error("Must specify a shiftID for comments on shift");
     }
     comments = currentShift?.comments;
+    comments?.forEach((comment) => {
+      comment.resourceType = associatedResourceType;
+    });
```
guynir42, line 312: Would it be easier to add this info in the handler like we did for Annotations?

mcoughlin: Good idea.

mcoughlin: Yes this was a good idea.

guynir42 (approved):

PR #3928, opened 2023-02-07
Let labelling checkbox remain unchecked after classification deletion

bfhealy:
This PR allows the labelling checkbox to remain unchecked after deleting one or all of a source's classifications. This will make it easier to replace classifications in an active learning group without marking the sources as labelled.

### DIFF
--- a/skyportal/handlers/api/classification.py
+++ b/skyportal/handlers/api/classification.py
@@ -462,6 +462,18 @@ class ClassificationHandler(BaseHandler):
             required: true
             schema:
               type: integer
+        requestBody:
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  label:
+                    type: boolean
+                    nullable: true
+                    description: |
+                      Add label associated with classification.
+                required:
         responses:
           200:
             content:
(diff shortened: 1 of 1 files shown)


skyportal/handlers/api/classification.py
```diff
@@ -481,25 +492,36 @@ def delete(self, classification_id):
...
-                if source_label is None:
-                    label = SourceLabel(
-                        obj_id=obj_id,
-                        labeller_id=self.associated_user_object.id,
-                        group_id=group_id,
-                    )
-                    session.add(label)
+            # labelling
+            add_label = True
+            if 'label' in data:
```
mcoughlin, line 504: Maybe best as

add_label = data.get('label', True)

skyportal/handlers/api/classification.py
```diff
@@ -595,28 +628,38 @@ def delete(self, obj_id):
                 .all()
             )
 
+            data = self.get_json()
+
+            # labelling
+            add_label = True
```
mcoughlin, line 634: Here too

mcoughlin (commented):
I think the data.get style is more efficient.

bfhealy:
@mcoughlin I agree that's a better style. Just pushed the changes, although I'm not sure why the model logic test failed twice again (at different points in the 'Install system dependencies' step). We'll see what it does this time around.

mcoughlin (approved):
LGTM

PR #3929, opened 2023-02-07
MAXI events

mcoughlin:
This PR adds MAXI_UNKNOWN to GCN ingestion list (low rate, but interesting objects).


guynir42 (approved):
I don't know what this is but it looks fine to add it.

PR #3930, opened 2023-02-07
Photometric series API

guynir42:
This PR adds the API to manage the photometric series model. 
In this model the photometry points are saved as one big file in persistentdata, instead of individual rows in the DB. 
This is most useful for surveys that make lots of continuous data in a row, like TESS. 
In that case, each sector of TESS would be a photometric series, which could replace 100,000 photometric points with one file and one row in the DB. 

Adding the API (handlers) to the existing models allows us to also add demo data and tests. 

A followup PR will add the plotting of photometric series alongside the regular photometric plot.


guynir42:
I intend to also add a short doc on how to generate the HDF5 files and I also need to make a migration script (after initial review is done).

skyportal/tests/api/test_photometric_series.py
```diff
@@ -0,0 +1,2176 @@
...
+            file_hash = hashlib.md5()
+            file_hash.update(file_buf)
+
+        # assert len(file_buf) == len(mem_buf)
+        assert file_hash.hexdigest() == mem_hash.hexdigest()
+
+        # make sure the same hash is made even when
+        # changing the versions of pandas, etc.
+        # if not, we will need to re-make all the hashes!
+        assert file_hash.hexdigest() == 'daf70e10284a36020af2cc102ae3d32a'
```
profjsb, line 107: this hash string might change with different hdf5 versions. might want to generate this dynamically.

skyportal/tests/api/test_photometric_series.py
```diff
@@ -0,0 +1,2176 @@
...
+from astropy.time import Time
+
+import sqlalchemy as sa
+from sqlalchemy.exc import IntegrityError
+
+from skyportal.tests import api, assert_api, assert_api_fail
+from skyportal.models import DBSession, PhotometricSeries
+
+
+def convert_dataframe_to_bytes(df, metadata=None, encode=True):
```
profjsb, line 20: seems like a broadly useful function here. Can you put it in utils.py and then import it for these tests?

skyportal/tests/api/test_photometric_series.py
```diff
@@ -0,0 +1,2176 @@
...
+        that can be sent over API calls.
+        If encode=False, the bytes are the raw bytes of the HDF5 file.
+    """
+
+    # this store should work without writing to disk
+    # if you open a regular store you'd just need
+    # to delete the file at the end
+    # ref: https://github.com/pandas-dev/pandas/issues/9246#issuecomment-74041497
+    with pd.HDFStore(
+        'test_file.h5', mode='w', driver="H5FD_CORE", driver_core_backing_store=0
```
profjsb, line 47: perhaps better to use a temporarynamed file here since if multiple tests are run in the same directory there could be contention over this file.

skyportal/handlers/api/photometric_series.py
```diff
@@ -0,0 +1,1764 @@
...
+def load_dataframe_from_bytestream(data):
+    """
+    Load the pandas data frame from a byte stream.
+    Will return a DataFrame and a metadata dictionary,
+    if such a dictionary is found in the data file's attributes.
+
+    ref: https://github.com/pandas-dev/pandas/issues/9246#issuecomment-74041497
+    """
+    with pd.HDFStore(
+        "data.h5",
```
profjsb, line 335: should be a temporary named file?

skyportal/handlers/api/photometric_series.py
```diff
@@ -0,0 +1,1764 @@
...
+                ).first()
+                if ps is None:
+                    return self.error('Invalid photometric series ID.')
+                return self.success(data=ps.to_dict())
+
+        # get all photometric series
+        ra = self.get_query_argument('ra', None)
+        dec = self.get_query_argument('dec', None)
+        radius = self.get_query_argument('radius', None)
+        object_id = self.get_query_argument('objectID', None)
```
profjsb, line 1249: how is `objectID` different than `Obj` or `Source`?

skyportal/handlers/api/photometric_series.py
```diff
@@ -0,0 +1,1764 @@
...
+            MAX_SERIES_PER_PAGE,
+        )
+
+        stmt = PhotometricSeries.select(self.current_user)
+
+        if ra is not None and dec is not None and radius is not None:
+            try:
+                ra = float(ra)
+                dec = float(dec)
+                radius = float(radius)
```
profjsb, line 1303: Do some bounds checking here. Dont want to propagate further if they aren't correct.

profjsb (commented):
Looking good. This is different enough that I think it's worth writing up it's own section in the docs for usage.

skyportal/tests/api/test_photometric_series.py
guynir42: Yes, the idea is that when the version changes we would be made aware by this test failure. 
Such a change in version would require that we re-save and re-calculate the hash for all photometric series.

skyportal/handlers/api/photometric_series.py
guynir42: I think in this case it is the id of the skyportal Obj (or Source)

skyportal/handlers/api/photometric_series.py
guynir42: I think in this case the filename doesn't make any difference.

skyportal/handlers/api/photometric_series.py
guynir42: done

skyportal/handlers/api/photometric_series.py
```diff
@@ -0,0 +1,1764 @@
...
+              and the length of each value must be the same.
+              If a string, will be converted to a bytes array
+              and de-serialized by the pandas HDF5 reader.
+              Use the HDFStore to create a file that includes
+              a single key/group with the photometric data.
+              Additional information, including any of the
+              parameters specified below, can be stored in
+              the HDFStore as well, inside the attributes
+              under the key "metadata".
+              In any case the DataFrame must have the following
```
mcoughlin, line 63: Maybe should add explicitly the list of required and allowed optional columns here?

skyportal/handlers/api/photometric_series.py
```diff
@@ -0,0 +1,1764 @@
...
+        """
+
+
+log = make_log('api/photometric_series')
+
+DEFAULT_SERIES_PER_PAGE = 100
+MAX_SERIES_PER_PAGE = 500
+
+
+def get_group_ids(data, user, session):
```
mcoughlin, line 231: Is this different from the photometry one? Can it just be imported from there?

skyportal/handlers/api/photometric_series.py
```diff
@@ -0,0 +1,1764 @@
...
+        if group is None:
+            raise ValidationError(f'Invalid group ID: {group_id}')
+
+    # always add the single user group
+    group_ids.append(user.single_user_group.id)
+    group_ids = list(set(group_ids))
+    return group_ids
+
+
+def get_stream_ids(data, user, session):
```
mcoughlin, line 284: Same for this.

skyportal/handlers/api/photometric_series.py
```diff
@@ -0,0 +1,1764 @@
...
+          - photometry
+          - photometric series
+        parameters:
+          - in: path
+            name: photometric_series_id
+            required: true
+            schema:
+              type: integer
+        requestBody:
+          {body_schema_docstring.replace("required: true", "required: false")}
```
mcoughlin, line 630: This is nifty.

skyportal/models/photometric_series.py
```diff
@@ -0,0 +1,1283 @@
+__all__ = [
```
mcoughlin, line 1: I am glad you put this in its own thing.

mcoughlin (approved):
A few small things. It's a cool product.

skyportal/handlers/api/photometric_series.py
guynir42: We don't require anything except "mjd" and "mag" or "flux", and there could be additional columns that have meanings like "fluxerr" or "magerr" that are used mainly for plotting. 
The users are free to add any column they wish and that is stored just for record-keeping, skyportal doesn't do anything with those.

skyportal/handlers/api/photometric_series.py
guynir42: I've tried to do that, but the existing `get_group_ids` (and the streams one) are a little outdated, e.g., they don't get the session input and I didn't want to touch that code. I suppose we could merge them at some point.

mcoughlin: Ok makes sense, maybe open an issue so we take care of it?

guynir42: done

PR #3931, opened 2023-02-07
Add plugins to SourceAnnotationButtons.jsx

bfhealy:
This PR adds a placeholder component for plugins to `SourceAnnotationButtons.jsx` and imports these plugins to the same. The intent is to allow additional buttons to appear for specific instances of SkyPortal (e.g. a 'Kowalski' button for Fritz).


mcoughlin (approved):
LGTM!

PR #3944, opened 2023-02-13
DNM: Rubin scheduler packages

mcoughlin:
Testing out the Rubin scheduler packages.


profjsb:
Need to add `rubin_sim` (+ version) to requirements.txt?

Theodlz:
@mcoughlin is that something we should consider merging at some point, or closing? If neither, maybe just turn it into a draft?

mcoughlin:
@Theodlz Made it a draft for now.

PR #3946, opened 2023-02-14
Update shape of GROWTH-India FOV

mcoughlin:
This PR updates the shape of GROWTH-India FOV.


harshkumar13:
GIT FOV is a 23 arcmin circle, covering ~0.46 deg^2. The CCD size is bigger than the specified value. However, due to vignetting around the corners, only a 23arcmin radius circle is a usable area and should be used for GIT FOV.

guynir42 (approved):

PR #3951, opened 2023-02-15
Add multiple gcn notification

laurall974:
Need a review on the post method


static/js/ducks/profile.js
```diff
@@ -10,6 +10,10 @@ const GENERATE_TOKEN = "skyportal/GENERATE_TOKEN";
 
 const UPDATE_USER_PREFERENCES = "skyportal/UPDATE_USER_PREFERENCES";
 
+const ADD_USER_GCN_PREFERENCES = "skyportal/ADD_USER_GCN_PREFERENCES";
```
mcoughlin, line 13: This should simply be ADD_USER_PREFERENCES and DELETE_USER_GCN_PREFERENCES right?

static/js/ducks/profile.js
```diff
@@ -20,6 +24,18 @@ export function updateUserPreferences(preferences) {
   });
 }
 
+export function addUserGcnPreferences(preferences) {
```
mcoughlin, line 27: Same thing here.

skyportal/handlers/api/internal/profile.py
```diff
@@ -246,3 +246,70 @@ def patch(self, user_id=None):
...
+                    preferences[k] = {key: val for key, val in v.items() if val != ""}
+            user_prefs = deepcopy(user.preferences)
+
+            if not user_prefs:
+                user_prefs = {}
+            for key, value in preferences.items():
+                if key not in user_prefs:
+                    user_prefs[key] = value
+            try:
+                session.add(user_prefs)
```
mcoughlin, line 306: There isn't actually a user preferences data model is there or maybe I misunderstand?

mcoughlin (commented):
I don't quite get the need for the post. Is the issue that making the GCN preferences a list somehow problematic?

skyportal/handlers/api/internal/profile.py
laurall974: I thought I need to do a post method because with patch I can't create a new gcn events preference in the database. But maybe it is wrong ? Patch works too ?

laurall974: regarding the patch method, I thought there was user preferences data model

mcoughlin: I think patch is fine, as it should be the same as the rest I think.

laurall974: ok I will do that so ! thank you

PR #3952, opened 2023-02-15
delete all instrument fields

mcoughlin:
This PR adds an option to delete instrument fields.


guynir42 (commented):
It looks like you are making the delete instrument do something it is really not supposed to be doing. Since deleting an instrument is a dangerous thing, I would think using the delete instrument handler for other things is not a good idea. I don't know if deleting the associated fields is something we might be doing often (e.g., when playing around with skymaps) so maybe it should go in its own handler?

skyportal/handlers/api/instrument.py
```diff
@@ -975,3 +965,48 @@ def add_tiles(
...
         Session.remove()
         return field_ids
+
+
+class InstrumentFieldHandler(BaseHandler):
+    @permissions(['Delete instrument'])
+    def delete(self, instrument_id):
+        """
+        ---
+        description: Delete an instrument
```
guynir42, line 975:
```suggestion
        description: Delete fields associated with an instrument
```

skyportal/handlers/api/instrument.py
```diff
@@ -975,3 +965,48 @@ def add_tiles(
     finally:
         Session.remove()
         return field_ids
+
+
+class InstrumentFieldHandler(BaseHandler):
+    @permissions(['Delete instrument'])
```
guynir42, line 971: Is this the correct permissions for this action?

guynir42 (approved):
Minor comments:

skyportal/handlers/api/instrument.py
mcoughlin: I think so. We want people with that level to do this.

PR #3955, opened 2023-02-15
CommentList - Missing associated resource type

### LINKED ISSUE #3953: Shifts test is broken
The `test_shifts` test is not working. 
The failure happens when trying to delete a comment from the frontend. 
The delete button only appears when hovering over the comment. 
There's a command to click that comment but that is not the same as hover. 
I don't know how to solve this but I think this is one of the remaining tests that don't work.

Theodlz:
Fixes #3953

### DIFF
--- a/static/js/components/CommentList.jsx
+++ b/static/js/components/CommentList.jsx
@@ -340,9 +340,12 @@ const CommentList = ({
               onFocus={() => handleMouseHover(id, userProfile, author.username)}
               onBlur={() => handleMouseLeave()}
             >
+              {console.log("shiftID", shift_id)}
+              {console.log("associatedResourceType", resourceType)}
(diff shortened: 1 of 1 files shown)


mcoughlin:
@Theodlz I think resourceType is correct. We want to be adding that into all places where comments are returned.

mcoughlin:
@Theodlz I.e. in the handler.

Theodlz:
Ok! Then we should remove the associatedResourceType that is taken as an argument by the frontend methods.
Let me add that in the handlers then.

mcoughlin:
@Theodlz I am fixing it quick in the WINTER PR because its open ;)

Theodlz:
Ok! Let me know if you need a review on any PRs.

PR #3959, opened 2023-02-17
Default Analysis

Theodlz:
This PR aims to add a default analysis table. In this table, one can create a reference to an analysis service, along with a set of default parameters, and a list of filters to apply to sources.

Based on these filters (that for now are limited to source classifications in this first PR), an analysis can be triggered automatically.

What's missing:
- Stats: I believe that an extra column containing basic stats to figure out how many analyses have been triggered using a given set of filters could be very useful for users to estimate the volume of analysis they can expect. On the frontend, we could also display an estimate in a following PR, to guide users in their choice of filters before submitting a default analysis.
- Frontend: This PR is only adding the backend. This is intended

What can change:
- DB trigger: The way I go through the filters is not ideal. For now, it's a JSONB column like `{"classifications: [{"name": "SN", "probability": 0.7}]}`. So far, I haven't found a way to iterate through the list of classifications in a query to filter the default analysis. What I can do is simply treat it as a string and look for "name": "blabla", but for the probability what I want is the query to return a default analysis if the probability of the classification is equal or above the one in the filters. 
So what I do for now is treat it as a string in a query, and then iterate over the results and post an analysis only if the probability value matches the filter. The best is to look at the code in the `models/` which speaks for itself. Reviews can feel free to commit to my branch if they know how to handle JSONB queries!

Thanks!


mcoughlin (approved):
This looks incredibly nice Theo. I like the stats idea you have in particular.

skyportal/models/analysis.py
```diff
@@ -553,3 +657,75 @@ def delete_assoc_analysis_data_from_disk(mapper, connection, target):
...
+                            "classifications"
+                        ]
+                        if classification["name"] == target_data['classification']
+                    ),
+                    None,
+                )
+                if classification_filter['probability'] <= target_data['probability']:
+                    log(
+                        f'Creating default analysis {default_analysis.analysis_service.name} for classification {target.id}'
+                    )
```
profjsb, line 694: This doesn't seem to terminate if the probability is low, as I would have expected.

skyportal/models/analysis.py
```diff
@@ -553,3 +657,75 @@ def delete_assoc_analysis_data_from_disk(mapper, connection, target):
...
+            default_analyses = session.scalars(stmt).all()
+
+            for default_analysis in default_analyses:
+                classification_filter = next(
+                    (
+                        classification
+                        for classification in default_analysis.source_filter[
+                            "classifications"
+                        ]
+                        if classification["name"] == target_data['classification']
```
profjsb, line 687: we may want to do a "like" instead of a "equals" here. I can imagine a case where someone would want to trigger an analysis if the classification matches "supernova".

profjsb (commented):
This looks good. The name default threw me off for awhile. Perhaps a better name could be `triggered analysis` since it's really happening when events transpire on a source (such as classification). 

I didn't see a place an where obvious race condition was considered and stopped: a triggered/default analysis causes another classification to be written.  This is likely one of the main goals of analysis is that data is sent to a 3rd party and the results are not just returned to the webhook but also a classification is made. You could imagine limiting the number of triggered analyses on a source or limiting analyses triggered by those that are not associated to another analysis.

Future ideas after this PR:

  1) Test for classification names in the parent node names. That is, trigger an analysis if the classification is any type of "explosive" event or "variable". Similarly, trigger events based on tags of the classification ("periodic").

  2) Allow for complex triggering with ANDs and ORs. That is, we want to trigger whenever classification P(Ia) > 0.5 and P(Ibc) < 0.8.

Theodlz:
HI @profjsb, thanks for the review! Guy and I talked about it during the iteration meeting today, and it's obvious that I need to implement some sort of limiting rate. We were thinking to have at least something like "max N triggered analysis within a period of K hours", which could be extended in a future PR to have more fine-grained control at the source level, so like "max N analysis within K hours for a source, or simply max N analysis for this source". 

I do agree that the default is not clear enough. I made that choice to follow the current trend of default obs plan, and default follow-up requests.

skyportal/models/analysis.py
Theodlz: What I had in mind is that the probability of a new classification has to be over the value you defined in the triggered analysis to trigger an analysis. Do you mean that it should be the opposite? That it should be a maximum value, rather than a minimum? That seems counter intuitive to me. 

Or maybe you meant that in general we shouldn't trigger the analysis if the value is too low anyway, independently of the value the user picked for his triggered analysis?

mcoughlin (approved):
Looks really nice. LGTM.

skyportal/models/analysis.py
profjsb: My bad: I thought that `with DBSession() as db_session:` was on the same indentation as the `if` statement. I see what you're doing here.

profjsb (approved):
Please confirm that the test failure in unrelated. Go ahead and merge but please make issues for the future PR ideas if you think that they could be useful. also, need some documentation (can do that in another review)

Theodlz:
> Please confirm that the test failure in unrelated. Go ahead and merge but please make issues for the future PR ideas if you think that they could be useful. also, need some documentation (can do that in another review)

Hi Josh, thanks for the review. I will verify that the tests failures are unrelated, and will merge accordingly.

PR #3965, opened 2023-02-18
Colors in Classifications 

KodeurFou:
hi @mcoughlin with @celiune and @Adiman007 we've worked on this long over due task :point_right: :point_left: . 
It's the PR for the colors in classifications tasks, we added only few classifications in our rejected / comfirmed dictionnaries but we can make another update to the PR that modifies the dictionnaries if you want to add more classifications NOR colors.


mcoughlin:
@KodeurFou I think these colors / classes need to go in the config. They are very particular to the instance of SkyPortal you are deploying.

Theodlz (changes requested):
Hi @KodeurFou, I think your last commit belongs to another PR as it differs greatly from the purpose of this one. Please reverse it or simply delete the changes and recommit.

I agree with @mcoughlin, these colors should be configurable somehow. In the config makes sense for a first PR.

static/js/components/ShowClassification.jsx
```diff
@@ -99,6 +112,23 @@ const ClassificationRow = ({ classifications }) => {
 
   const classification = classifications[0];
 
+  const confirmed_classes = ["GO GRANDMA", "GO GRANDMA (HIGH PRIORITY)"];
+  const rejected_classes = ["STOP GRANDMA"];
```
profjsb, line 116: We should not be hard coding anything with GRANDMA here. You can add these via an environment variable in the config OR have these defined in a table.

profjsb (changes requested):

Adiman007:
@KodeurFou you pushed your code on the wrong branch, I think

PR #3967, opened 2023-02-18
WIP: GCN Event notifications - multiple notification preferences

laurall974:
@Theodlz @mcoughlin 
There is an error in on delete method. I do not know why it does not delete the chip when i click on.


mcoughlin:
@laurall974 Printing out in the handler, I see:

'notifications': {'Fermi': {'gcn_tags': [], 'gcn_properties': [], 'gcn_notice_types': ['FERMI_GBM_GND_POS', 'FERMI_GBM_FIN_POS'], 'localization_tags': [], 'localization_properties': []}, 'active': True, 'sources': {'active': True, 'classifications': ['Algol']}, 'gcn_events': {'Fermi': {'gcn_tags': [], 'gcn_properties': [], 'gcn_notice_types': ['FERMI_GBM_GND_POS', 'FERMI_GBM_FIN_POS'], 'localization_tags': [], 'localization_properties': []}, 'active': True}}}


It looks to me like your onDelete function:

    const currentGcnPref = notifications;
    delete currentGcnPref.gcn_events[buttonName];
    const prefs = { 
      notifications: currentGcnPref.gcn_events,
    };

should be more like 

    
    let prefs = notifications;
    delete prefs.gcn_events[buttonName];

or something. As is, you end up not actually deleting what you intend

laurall974:
with the help of @Adiman007

skyportal/handlers/api/internal/profile.py
```diff
@@ -209,7 +209,7 @@ def patch(self, user_id=None):
                     user.contact_email = None
 
             preferences = data.get("preferences", {})
-
+            print(preferences)
```
mcoughlin, line 212: Delete

skyportal/handlers/api/internal/profile.py
```diff
@@ -226,7 +228,7 @@ def patch(self, user_id=None):
                     user_prefs["photometryButtons"] = preferences["photometryButtons"]
                 user_prefs = recursive_update(user_prefs, preferences)
             user.preferences = user_prefs
-
+            print(user.preferences)
```
mcoughlin, line 231: Delete

mcoughlin (approved):
Thanks @laurall974. It looks nice. I would like @Theodlz to give it a review as well.

Adiman007:
You need to remove the print() I added to bugfix your code 😅

Theodlz (changes requested):
Hi @laurall974 @Adiman007.
The feature is nice, being able to create multiple sets of constraints to trigger notifications is very useful.
However, as you might imagine, the backend is what really triggers the notifications. By modifying the frontend like this, you break the backend. 

You need to go the the `user_notifications.py` file in the `models` directory and make a bunch of changes.

Also, keep in mind that if you had notifications on gcn_events before, how do you handle that? You guys added a list of notifications, but if instead of a list someone has one notification setup, you might run into an error. You need to make sure that this won't break the current notification settings.

For that, you need on the frontend to make sure that when loading the current notification settings of the user, you get his current gcnevent notification and put it in a list, so that the next time he submits a profile change it will effectively migrate his old settings to the new format if that makes sense. You can also handle that in the profile.py file if you want. See what works best (if you can only do it in the frontend that's great). Then, in the `user_notifications.py` you need to do the same. If the user `gcn_event` key in his notification preferences is a dict (meaning its the old format), change is to be a list with that dict in it.

These failsafe are very easy to implement and will prevent potential bugs and mostly will insure not to break anyone's notifications. That is the **last** thing to implement. For now, focus on changing the backend to user lists of gcnevent notifications and not just one.

Last comment: the frontend is displayed weirdly for me, the existing gcn event notifications are displayed a little randomly in the middle of the page. Moreover, it would be nice to be able to hover/click on the name to see what settings the user choose when he created it.

Maybe one of you can work on the frontend and the other on the backend?

Theodlz:
By the way, you can easily see that it breaks the notifications by looking at the github actions `frontend_pt2`

laurall974:
@Theodlz

mcoughlin:
@Theodlz any chance you can help resolve conflicts here?

Theodlz:
Sure. I will have a look when I'm done with deploy.

mcoughlin:
Closing in favor of https://github.com/skyportal/skyportal/pull/4045

PR #3970, opened 2023-02-20
Comments attachments: Better fits file preview

Theodlz:
In this PR, we try to provide users with a better preview of fits file, as well as fits.fz files (currently missing)
Here is an example:

| Before | After |
| ------------- | ------------- |
| image | image |

*Details: I used [ZScaleInterval](https://github.com/spacetelescope/stsci.numdisplay/blob/master/lib/stsci/numdisplay/zscale.py) followed by a [SinhStretch](https://docs.astropy.org/en/stable/visualization/normalization.html#combining-stretches-and-matplotlib-normalization) to achieve this result.*

### DIFF
--- a/skyportal/handlers/api/comment.py
+++ b/skyportal/handlers/api/comment.py
@@ -1,6 +1,5 @@
 import string
 import base64
-import io
 from marshmallow.exceptions import ValidationError
 import os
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
Much improved. LGTM.

PR #3977, opened 2023-02-23
Source search bar

mcoughlin:
This PR adds a fix for missing sources in search bar.


Theodlz:
There was also a problem with illegal characters like parentheses and so on. Maybe a try catch that puts the input back to an empty str and displays the error would be nice.

Theodlz (approved):
LGTM

PR #3980, opened 2023-02-26
Onboarding: missing rollbacks

Theodlz:


mcoughlin (approved):
Thanks @Theodlz. Looks like a good thing to try.

PR #4005, opened 2023-03-01
Photometry between duplicate sources

mcoughlin:
This PR enables the user to copy photometry between duplicate sources.


skyportal/handlers/api/source.py
```diff
@@ -3448,3 +3449,127 @@ async def get(self, obj_id):
...
+
+
+class SourceCopyPhotometryHandler(BaseHandler):
+    @permissions(["Upload data"])
+    def post(self, objId):
+        """
+        ---
+        description: Copy photometry from one source to another
+        tags:
+          - notifications
```
guynir42, line 3461:
```suggestion
          - sources
          - photometry
```

skyportal/handlers/api/source.py
```diff
@@ -3448,3 +3449,127 @@ async def get(self, obj_id):
...
+        requestBody:
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  objId:
+                    type: string
+                    description: |
+                      The ID of the Source's Obj the photometry is being copied to
```
guynir42, line 3471:
```suggestion
                      The Obj_id of the target Source (to which the photometry is being copied to)
```

skyportal/handlers/api/source.py
```diff
@@ -3448,3 +3449,127 @@ async def get(self, obj_id):
...
+                  groupIds:
+                    type: array
+                    items:
+                      type: integer
+                    description: |
+                      List of IDs of groups to give photometry access to
+                  duplicateId:
+                    type: string
+                    description: |
+                      The ID of the Source's Obj the photometry is being copied from
```
guynir42, line 3481: Maybe "originID" and "targetID" are better? also, it looks like you are getting the obj_id from the path. Should that be in the docstring?

skyportal/handlers/api/source.py
```diff
@@ -3448,3 +3449,127 @@ async def get(self, obj_id):
             data = io.BytesIO(buf.read())
 
             await self.send_file(data, filename, output_type=output_format)
+
+
+class SourceCopyPhotometryHandler(BaseHandler):
+    @permissions(["Upload data"])
+    def post(self, objId):
```
guynir42, line 3456:
```suggestion
    def post(self, obj_id):
```

skyportal/handlers/api/source.py
```diff
@@ -3448,3 +3449,127 @@ async def get(self, obj_id):
...
+        """
+        ---
+        description: Copy photometry from one source to another
```
guynir42, line 3459:
```suggestion
        description: Copy all photometry points from one source to another
```

guynir42 (commented):
I have some suggestions. And also this could use a test.

skyportal/handlers/api/source.py
mcoughlin: Good points.

skyportal/handlers/api/photometry.py
```diff
@@ -751,7 +751,8 @@ def get_group_ids(data, user):
         )
 
     # always add the single user group
-    group_ids.append(user.single_user_group.id)
+    if hasattr(user, 'single_user_group'):
```
guynir42, line 754: under what circumstances does the user not have a single_user_group? Is it when it is a token? because I feel like that still requires us to add the single user group of the underlying user...

skyportal/handlers/api/source.py
```diff
@@ -3453,36 +3453,40 @@ async def get(self, obj_id):
...
-                      The ID of the Source's Obj the photometry is being copied to
-                  groupIds:
+                  group_ids:
                     type: array
                     items:
                       type: integer
                     description: |
                       List of IDs of groups to give photometry access to
-                  duplicateId:
+                  originId:
```
guynir42, line 3483:
```suggestion
                  origin_id:
```

skyportal/tests/api/test_sources.py
```diff
@@ -2326,3 +2326,93 @@ def test_add_and_delete_source_label(upload_data_token, view_only_token, public_
...
     assert len(data["data"]["labellers"]) == 0
+
+
+def test_copy_photometry_sources(
+    public_group, upload_data_token, ztf_camera, view_only_token
+):
+    obj_id1 = str(uuid.uuid4())
+    obj_id2 = str(uuid.uuid4())
+    ra = 200.0 * np.random.random()
+    dec = 90.0 * np.random.random()
```
guynir42, line 2337:
```suggestion
    ra = 200.0 * np.random.random()
    dec = 89.0 * np.random.random()
```

skyportal/tests/api/test_sources.py
```diff
@@ -2326,3 +2326,93 @@ def test_add_and_delete_source_label(upload_data_token, view_only_token, public_
...
+            "limiting_mag": 22,
+            "magsys": "ab",
+        },
+        token=upload_data_token,
+    )
+    assert status == 200
+
+    status, data = api(
+        "POST",
+        f"sources/{obj_id1}/copy",
```
guynir42, line 2401: I feel like maybe the URL path should say `copy_photometry` just to be more specific.

guynir42 (approved):
I have minor comments.

skyportal/handlers/api/photometry.py
mcoughlin: yes when it is a token.

guynir42: Can we get the underlying user in that case?

skyportal/tests/api/test_sources.py
mcoughlin: yeah this was cause it was using current_user instead of associated_user_or_object. I will strip that out.

mcoughlin: yes this is fixed.

Issue #3846, opened 2023-01-17
Document the JS component tree


guynir42:
It seems like it is getting harder and harder to find my way around the JS/components folder. 
There's a need for two things (maybe three):
1) Add a short code comment at the beginning of each file, saying e.g., "this is in the Source page, under the annotations table, etc.". 
2) Make a readme file that quickly maps the tree of components, starting from main pages (e.g., candidates page, telescopes page). This could just be ascii with indentations for going deeper into the tree. In this file components that are re-used (e.g., SourceTable) could appear twice, under different branches. The goal is to help you figure out which file you need, when all you have is the a part of a page on the front end. 
3) We could also add a graphical representation of this dependency tree. I personally am not sure that's more informative than just a text version but maybe that is more readable for other people. 

NOTE: there could be automated tools to do this. If not, there could be a simple python script one could write to check the imports of each file (keeping only those imports local to the components folder) and building up a representation as e.g., nested dictionaries that are easy to print out later. Again, probably exists tools to do that out there. 

NOTE 2: this could be a good issue for some of the new devs that are working on front end. A good opportunity to get to know the landscape.

mcoughlin:
@Theodlz can we assign one of the students?

Theodlz:
Yes @mcoughlin, I think that's a great idea. Maybe to a group of 2 students. They can easily devide the components folder in between them and do it a bit quicker.

mcoughlin:
@Theodlz Maybe the Adrien's could knock this out?

Theodlz:
Sounds good to me!

mcoughlin:
@Adiman007 @KodeurFou Can you take a look?

KodeurFou:
Here is the progress we've made so far:

![dependency-graph](https://user-images.githubusercontent.com/104773256/219872346-bd1684ca-b026-4438-94f4-35bbd6fb4137.svg)


We used dependency-cruiser to model the architecture of SP, we can see the react components and their relations

The graph is a little messy we tried having it displayed in a reactive way in an html format but so far no success

The problem we've had is it doesn't seem to be fitted to run on the same version of react that SP does


![image](https://user-images.githubusercontent.com/104773256/219872390-7d2434a8-68de-4ada-b86b-7ecb52eae873.png)

but the output we still get seems to be correct




should we try another way to try and automate the process or should we document the way we got it working so it can be re-rendered when some new components are added?

mcoughlin:
[...]

PR #4006, opened 2023-03-02
JS component tree (Dependency-cruiser)

KodeurFou:
Here is the version of dependency cruiser I used

that the site I used to generate the .svg
https://www.npmjs.com/package/dependency-cruiser

lissue we where discussing it on :
https://github.com/skyportal/skyportal/issues/3846


mcoughlin:
@KodeurFou please remove the ^ in the dependencies.

mcoughlin:
@Theodlz I think we should put something like this into the contribution docs

Theodlz:
@mcoughlin though I do love the idea, I feel like how that specific graph looks makes it pretty much unreadable, + we want this to be regenerated automagically along with the docs or smth, we can't commit this as is because it will just be incorrect after one frontend PR.

mcoughlin:
@Theodlz Ok maybe have an issue referencing the idea? Would be great to have it generated at doc creation time...

Theodlz:
Good idea, doing that now

PR #4007, opened 2023-03-02
Edit token acls

mcoughlin:
This PR edits  token acls.

### DIFF
--- a/skyportal/handlers/api/internal/token.py
+++ b/skyportal/handlers/api/internal/token.py
@@ -166,52 +166,51 @@ class TokenHandler(BaseHandler):
         """
 
         with self.Session() as session:
-            token = session.scalars(
-                Token.select(session.user_or_token, mode="update").where(
-                    Token.id == token_id
-                )
-            ).first()
-            if token is None:
-                return self.error(
-                    'Either the specified token does not exist, '
-                    'or the user does not have the necessary '
-                    'permissions to delete it.'
-                )
-            session.merge(token)
-
-            data = self.get_json()
-            data['id'] = token_id
-
-            if 'user_id' in data:
-                user_id = data['user_id']
-                user = session.scalars(
-                    User.select(session.user_or_token).where(User.id == user_id)
-                ).first()
-            else:
-                user = self.associated_user_object
-                user_id = user.id
-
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/internal/token.py
```diff
@@ -135,6 +137,118 @@ def get(self, token_id=None):
             data = session.scalars(stmt).all()
             return self.success(data=data)
 
+    @auth_or_token
+    def put(self, token_id):
+        """
+        ---
+        description: Update token
+        tags:
+          - telescopes
```
guynir42, line 146:
```suggestion
          - tokens
```

skyportal/handlers/api/internal/token.py
```diff
@@ -135,6 +137,118 @@ def get(self, token_id=None):
...
+                token = session.scalars(
+                    Token.select(session.user_or_token, mode="update").where(
+                        Token.id == token_id
+                    )
+                ).first()
+                if token is None:
+                    return self.error(
+                        'Either the specified token does not exist, '
+                        'or the user does not have the necessary '
+                        'permissions to delete it.'
```
guynir42, line 179:
```suggestion
                        'Either the specified token does not exist, '
                        'or the user does not have the necessary '
                        'permissions to update it.'
```

skyportal/handlers/api/internal/token.py
```diff
@@ -135,6 +137,118 @@ def get(self, token_id=None):
...
+                                ACL.select(session.user_or_token).where(
+                                    ACL.id == acl_id
+                                )
+                            ).first()
+                            is not None
+                            for acl_id in new_acl_ids
+                        ]
+                    ):
+                        return self.error(
+                            "Improperly formatted parameter aclIds; must be an array of strings."
```
guynir42, line 225: It seems like you are not just testing that they are strings, you are also checking they are legit ACLs. Maybe the error message should reflect that.

guynir42 (approved):
Very cool. I have some minor comments.

skyportal/handlers/api/internal/token.py
mcoughlin: Good point

PR #4008, opened 2023-03-03
json dumps the analysis results

mcoughlin:
This PR uses json.dumps to be sure result is in sensible download format.

### DIFF
--- a/skyportal/handlers/api/analysis.py
+++ b/skyportal/handlers/api/analysis.py
@@ -1392,6 +1392,7 @@ class AnalysisProductsHandler(BaseHandler):
                             await self.send_file(
                                 output_data, filename, output_type=output_type
                             )
+                            return
                     elif product_type.lower() == "results":
                         if not analysis.has_results_data:
                             return self.error(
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/analysis.py
```diff
@@ -1392,6 +1392,7 @@ async def get(
                             await self.send_file(
                                 output_data, filename, output_type=output_type
                             )
+                            return
```
guynir42, line 1395: This return statement seems out of place...

mcoughlin: It's not. It sends the file and returns.

guynir42: Shouldn't this be a `retutn.success`?

mcoughlin: That returns json, which is not what we want.

guynir42: I don't think we are supposed to have a handler that doesn't return either success or failure.

mcoughlin: I think the file download stuff is an exception here.

guynir42: So if `send_file` replaces this action of passing back `success`, what happens in other cases where there isn't this `return` statement right after it?

guynir42: ok I see how it works. This is fine

guynir42 (approved):

PR #4017, opened 2023-03-07
Classifications colors

celiune:
Hi @mcoughlin, with @KodeurFou and @Adiman007 we've made the changes asked in the previous pull request [Colors in Classifications](https://github.com/skyportal/skyportal/pull/3965). We've opened a new pull request as @KodeurFou's skyportal is broken for the moment and no change could be made.
Hoping that it's good now!

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,0 +1,698 @@
+app:
+  title: SkyPortal
+  logos: # Logos to place in the top-left of the header/login page (zero or more)
+    - src: /static/images/skyportal_logo_dark.png
+      alt_text: Skyportal logo
+  login_message: |
+    For more information about the project, see
+    <a href="https://github.com/skyportal/skyportal">
+      https://github.com/skyportal/skyportal
+    </a>
+
+  secret_key:
+    abc01234 # This secret key can be any random string of
+    # characters.
+    #
+    # You should re-generate this for your application
+    # using:
+    #
+    # base64.b64encode(os.urandom(50)).decode('ascii')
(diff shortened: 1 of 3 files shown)


mcoughlin:
@celiune Can you please add me as a collaborator on your fork?

mcoughlin:
@KodeurFou @Adiman007 @celiune Thanks for this, it is closer:

I think instead of hardcoding these 4 types: confirmed, rejected, not_confirmed, and obs, cleaner and more generic would be:

classifications: {'kilonova': '#03c04a', 'supernova': '#ff0000'}

and then you just need to check for colors in the chip loop. Also please just modify config.yaml.defaults, we do not commit config.yaml to the repo.

Adiman007:
classifications: {'kilonova': '#03c04a', 'supernova': '#ff0000'}, only two classifications are needed ?

> @KodeurFou @Adiman007 @celiune Thanks for this, it is closer:
> 
> I think instead of hardcoding these 4 types: confirmed, rejected, not_confirmed, and obs, cleaner and more generic would be:
> 
> classifications: {'kilonova': '#03c04a', 'supernova': '#ff0000'}
> 
> and then you just need to check for colors in the chip loop. Also please just modify config.yaml.defaults, we do not commit config.yaml to the repo.

mcoughlin:
@Adiman007 My suggestion is just showing a few examples that show up in the demo data, I.e. variable, is fine. Folks can modify the config on deploy to fill out the colors and types they want.

PR #4018, opened 2023-03-07
Observation type for SEDMv2

mcoughlin:
This PR adds observation type to SEDMv2.


mcoughlin:
@Yashvi-Sharma

Yashvi-Sharma:
Looks good to me. We just need to add one more option for stellar science cases for the emccd 'mode' (exposure time per frame in rapid readout).

mcoughlin:
Thanks @Yashvi-Sharma.

PR #4020, opened 2023-03-07
aicommits git hook

Theodlz:
This PR adds the instructions to use aicommits as a git hook to generate commit texts using OpenAI models.

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -14,6 +14,26 @@ conform with our code style standards. We use black to reformat Python
 code, and PrettierJS for JavaScript. We also run ESLint to catch
 several common Redux usage bugs.
 
+## aicommits (optional)
+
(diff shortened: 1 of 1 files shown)


profjsb:
Might be worth trying adding 
```
"aicommits": "github:Nutlope/aicommits#pull/123/head"
```
to `package.json` so that aicommits is installed as part of this codebase?

Theodlz:
Hi @profjsb. After using it for a week, I can say that I'm not impressed by the results. Often, the commit message doesn't really what the changes I made should do, simply because it doesn't really have the context of the previous commit.
As of right now, I don't think it's such a great idea to rely on `aicommits`.

Theodlz:
Closing this for now as the results aren't so good

PR #4022, opened 2023-03-07
pysedm update

Theodlz:
This PR pins a new version of pysedm, which allows us to pin newer versions of iminuit and shapely


Theodlz:
@profjsb here aicommits didn't do so well. Most of it is correct but it clearly shouldn't say all the versions for each package, and it added a weird ".1" at the beginning. It's very probable that switching to gpt3.5 turbo will help with that.

mcoughlin (approved):
Thanks for getting it to this stage @Theodlz.

PR #4023, opened 2023-03-08
Gcn Allocation triggers

Theodlz:
This PR adds a table that can be used to keep track which telescopes triggered or "passed" an event, or didn't do anything with it (simply no entries in that case). The "triggered" state can be modified from the frontend manually by an allocation admin, and is set to True automatically if an observation plan is sent successfully to an instrument.


mcoughlin:
@Theodlz This looks really cool. Some suggestions for this or a future PR:

(i) Add a tooltip over the "Instruments Triggered" text to explain the Grey - Red - Green scheme
(ii) Add Green (triggered) Instruments to the list of tags in the Recent GCN Events panel
(iii) Add a dedicated column to the GCN Events table showing the Green (triggered) and Red (not-triggered) instruments.

Theodlz:
EDIT1: To this PR, I also added a lot of minor frontend modifications here and there to multiple MUI Datatable. That's why there are so many additions and deletions, but the modifications are actually very minor and verbose.

Theodlz:
(sorry when I removed the draft state to make it a normal PR it cancelled the approved state)

mcoughlin (approved):
LGTM!

PR #4025, opened 2023-03-09
Photometric series plotting and more

guynir42:
- add example photometric series from TESS
- add a doc file about uploading and downloading photometric series
- fix some issues in parsing the photometric series
- plot the photometric series alongside the regular photometry
- add a TESS filter using an upgrade of sncosmo
- add a TESS instrument to the demo data
- add the photometric series in the demo data


tools/data_loader.py
```diff
@@ -197,7 +197,18 @@ def inject_references(obj):
...
+                    try:
+                        payload = (
+                            Table.read(filename).to_pandas().to_dict(orient='list')
+                        )
+                    except Exception as e:
+                        # sometimes we save HDF5 files using an HDFStore.
+                        # in this case we read it as a binary file and return it as "data"
+                        if 'values_block_0' in str(e):
+                            with open(filename, 'rb') as fid:
+                                payload = base64.b64encode(fid.read())
```
mcoughlin, line 209: Should this also have its own try: except?

mcoughlin (commented):
Just reading the code a bit. Hopefully the TESS stuff will happen in the next week or so.

tools/data_loader.py
guynir42: Not really. All the exceptions are caught at the end and then there's an error log. 
I put an exception clause here just to be able to fish out the specific h5 files that I was making for the photometric series (although anyone using the `pd.HDFStore` method will be able to use this). 
If anything goes wrong reading that file it would just put an `X` there and report the error at the end.

mcoughlin: Ok!

guynir42:
I should mention that this is still a work in progress.
The code I have now seems to work fine, but I have a bunch more to add...

mcoughlin:
@guynir42 I should say I was just reading with interest, I didn't mean it as a review, just what I noticed.

guynir42:
You can even review what is here, but just be aware that more is coming...

profjsb (approved):
Looks good. You probably need to add photometric_series to the `index.rst` in `doc/` for it to show up? Also can you add a screenshot to this PR? Will make it easier to see what this PR is actually enabling on the front end.

doc/photometric_series.md
```diff
@@ -0,0 +1,169 @@
...
+This format is particularly useful for high-cadence observations,
+where multiple images are taken in the same pointing.
+
+### Posting data
+
+To upload a photometric series use the `POST /api/photometric_series` endpoint.
+The request body must contain a `data` field,
+that can contain either a JSON representation of the photometric series,
+(where the columns of the data table are the keys of the JSON object:
+{'mjd': [1, 2, 3], 'flux': [1, 2, 3], 'fluxerr': [1, 2, 3]}),
```
mcoughlin, line 19: Also can pass in mag and magerr?

doc/photometric_series.md
```diff
@@ -0,0 +1,169 @@
...
+If sending a JSON style dictionary, there is no way to include
+metadata, and that must be sent as request body parameters.
+
+### Downloading photometric series
+
+To download a photometric series, use the `GET /api/photometric_series` endpoint.
+The data will be returned as a JSON style dictionary,
+unless other formats are specified.
+
+... to be continued ...
```
mcoughlin, line 169: Will this be continued?

skyportal/handlers/api/photometry.py
```diff
@@ -1346,17 +1348,39 @@ def get(self, obj_id):
...
+                    .unique()
+                    .all()
+                )
+                series_data = []
+                for s in series:
+                    series_data += s.get_data_with_extra_columns().to_dict(
+                        orient='records'
+                    )
+
+            data = phot_data + series_data
```
mcoughlin, line 1383: Any sorting needed here? Could be nice to sort by date?

skyportal/handlers/api/source.py
```diff
@@ -1368,12 +1407,15 @@ async def get_sources(
                 obj_list[-1]["labellers"] = [user.to_dict() for user in users]
 
             if include_photometry_exists:
-                stmt = Photometry.select(session.user_or_token).where(
-                    Photometry.obj_id == obj.id
+                # stmt = Photometry.select(session.user_or_token).where(
```
mcoughlin, line 1410: Needed?

skyportal/tests/api/test_photometric_series.py
```diff
@@ -2125,3 +2128,278 @@ def test_get_series_paged(
...
     page2_ids = [ps['id'] for ps in data['data']['series']]
     assert set(page1_ids).isdisjoint(page2_ids)
+
+
+def test_allowed_bandpasses():
+    # will remove this test once I add TESS to the bandpasses
+    from skyportal.enum_types import ALLOWED_BANDPASSES
+    from pprint import pprint
+
+    pprint(ALLOWED_BANDPASSES)
```
mcoughlin, line 2138: What is this for exactly?

mcoughlin (approved):
A few small things.

skyportal/tests/api/test_photometric_series.py
guynir42: Ahh yes, I forgot quite a few things in there while I was working... Good catches on all of these. 

I am thinking about leaving this in until we get the TESS filter merged so I can safely add the demo data with the correct filter...

guynir42: Let me make this even more obvious so it doesn't get left in there by mistake.

profjsb (changes requested):
I was able to render the plot in your PR. However a few things stuck out. 

- First, the filter name on the plot seems weird (TESS/kepler). I wasn't aware that TESS used Kepler filters. Perhaps fix or clarify.
- Second, given all the data points, I wanted to change the default binning but click on the binsize slider the browser hangs.
- Third, I'm worried about the unbounded number of datapoints that could be plotted if the timeseries is very large. I asked chatgpt for help here. Here's what I got back: 

**I have more than 5000 data points to plot using bokeh. What's the best way to do this so that I do not slow down my browser?**

<i>When working with a large amount of data points, it's important to consider the performance of your visualization. Bokeh provides several techniques to optimize your visualization and avoid slowing down your browser.

Here are a few tips to get started:

Use downsampling techniques to reduce the number of data points displayed on the plot. Bokeh provides several built-in downsampling methods, such as "auto" and "subsample", which can be used to reduce the number of points displayed based on various criteria. For example, the "subsample" method will randomly select a subset of the data points to display, while the "auto" method will automatically select a downsample method based on the current plot size and the number of data points.

Use glyphs that are optimized for large datasets. Bokeh provides several glyphs that are optimized for displaying large datasets, such as "line" and "circle". These glyphs are designed to be efficient in terms of rendering performance, and can handle a large number of data points without slowing down the browser.

Use webgl backend. By default Bokeh uses canvas backend, but you can enable webgl backend by setting output_backend='webgl' on your output_file or output_notebook. Webgl is faster in general, but can also handle larger datasets, specially when using tools like Scatter or HeatMap.

Use server-side rendering. If downsampling and webgl are not enough and you still need to display a large dataset, you can use Bokeh's server-side rendering capabilities to generate the plot on the server and display it in the browser as a static image or interactive BokehJS document.

By using these techniques, you can create high-performance visualizations that can handle even very large datasets without slowing down your browser.</i>

guynir42:
@profjsb The filter is weird because I need to pin the sncosmo version that has TESS. For now Kepler is just a placeholder. 

Regarding the number of data points, I will give a try to the `auto` feature and ping you to have another look.

profjsb:
copilot:all

profjsb:
copilot:all

profjsb:
@guynir42 can you resolve the conflict so we can merge?

guynir42:
@profjsb I will resolve it, but we should not merge until the sncosmo updates with the TESS filter. Then I can put that in the demo data.

guynir42:
@profjsb @mcoughlin test failures appear to be unrelated. Does any of you want to have another look before I merge?

PR #4014, opened 2023-03-05
Obj Summaries (manual & with ChatGPT) + Embeddings in Pinecone

profjsb:
This PR enables human-written summaries (in markdown) on an object. It also enables AI summarization (using the [ChatGPT API](https://platform.openai.com/docs/guides/chat)) of the comments, classification, redshifts, etc. A user either sets their own OpenAI API key in the profile or uses the key set globally in `config.yaml`. AI summaries can be edited by hand. Summary histories are saved.  See the discussion https://github.com/skyportal/skyportal/discussions/3982 for more details.

Kapture 2023-03-05 at 21 42 04

Summarization is done using the analysis infrastructure and a new model attribute on the `analysis service` model is set: `is_summary` during setup. We enable a new service by default (in `services/openai_analysis_service`) that serves as the wrapper for the ChatGPT API.

In addition, we also enable the ability to save [embeddings](https://platform.openai.com/docs/guides/embeddings) on the AI-generated summaries. This is currently done using `pinecone` (configured in `config.yaml`). 

Next up:
   - documentation PR #4027 
   - testing PR
   - get/show nearest sources by embeddings look up

PR #4029, opened 2023-03-09
Add openai analysis service

profjsb:
Add the services needed to wrap ChatGPT for summarization and embeddings. See #4014


mcoughlin (approved):
LGTM

PR #4032, opened 2023-03-10
Gcn Events: Attempt to reduce skymap ingestion flakiness

Theodlz:
This PR tries to divide each step of a gcnevent ingestion (from XML), and performs the Localization creation a little earlier hoping to avoid an error noticed in prod.

### DIFF
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -23,6 +23,8 @@ notice_types = [
 
 log = make_log('gcnserver')
 
+user_id = 1
+
 
 def service():
     if client_id is None or client_id == '':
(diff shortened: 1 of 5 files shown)


services/gcn_service/gcn_service.py
```diff
@@ -64,7 +66,8 @@ def service():
             for message in consumer.consume():
                 payload = message.value()
                 consumer.commit(message)
-                user_id = 1
+                if payload.find(b'Broker: Unknown topic or partition') != -1:
```
Theodlz, line 69: Btw this is because when a Kafka topic hasn't been pushing notices recently, it says it can't be accessed. However we never caught that, and the payload was still passed to the `post_gcnevent_from_xml` causing errors. The errors themselves aren't a big problem, but we did end up with a lot of ugly prints in the logs.

mcoughlin (approved):
LGTM.

PR #4035, opened 2023-03-15
Webpack 5.76.0

mcoughlin:
This PR updates webpack to 5.76.0. See https://github.com/skyportal/skyportal/security/dependabot/8.


mcoughlin:
Resolves security issue with that package. Self-merging.

PR #4038, opened 2023-03-16
WIP: ZTF depot observation query

mcoughlin:
This PR adds the ability to query ZTF observations from the depot server, which will be the source of truth going forward.

### DIFF
--- a/skyportal/facility_apis/ztf.py
+++ b/skyportal/facility_apis/ztf.py
@@ -11,7 +11,6 @@ import numpy as np
 from sqlalchemy.orm import sessionmaker, scoped_session
 from tornado.ioloop import IOLoop
 import pandas as pd
-import pyvo
 import sqlalchemy as sa
 import urllib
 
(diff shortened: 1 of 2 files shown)


skyportal/facility_apis/ztf.py
```diff
@@ -882,7 +881,64 @@ def custom_json_schema(instrument, user):
...
+        url = f'{depot_url}/{date}/ztf_recentproc_{date}.json'
+        r = session.head(url)
+
+        # file exists
+        if r.status_code == 200:
+            r = session.get(url)
+            obstable = pd.DataFrame(r.json())
+
+            if obstable.empty:
+                log(f'No observations for instrument ID {instrument_id} for JD: {jd}')
```
guynir42, line 912: This could produce quite a bit of logging outputs... are you sure we need this?

skyportal/facility_apis/ztf.py
```diff
@@ -882,7 +881,64 @@ def custom_json_schema(instrument, user):
...
+                continue
+            # only want successfully reduced images
+            obstable = obstable[obstable['status'] == 0]
+
+            obs_grouped_by_obsjd = obstable.groupby('obsjd')
+            for ii, (jd, df_group) in enumerate(obs_grouped_by_obsjd):
+                df_group_median = df_group.median()
+                df_group_median['observation_id'] = ii
+                df_group_median['processed_fraction'] = len(df_group["fieldid"]) / 64.0
+                df_group_median['filter'] = inv_bands[int(1)]
```
guynir42, line 922: What is this `int(1)` about?

guynir42 (approved):

skyportal/facility_apis/ztf.py
mcoughlin: fixed

skyportal/facility_apis/ztf.py
mcoughlin: Should be very rare to trigger I think.

guynir42: OK then

PR #4040, opened 2023-03-17
Gcn event service: Non Async ingestion | Notifications: Async Ingestion

Theodlz:
This PR is a 2nd attempt to reduce (or get rid of) the flakiness experienced in production, where some events have missing localizations.

In production, we notice a warning that says: 

"Attribute history events accumulated on `N` previously clean instances within inner-flush event handlers have been reset, and will not result in database updates."

This happens when we try committing the session right after adding the localization to it. Turns out, that adding notification preferences on events recreates this bug locally. And that both locally and in prod, `N` corresponds exactly to the number of UserNotifications being added to the session. 

Moreover, while testing skymap ingestion locally (with LVC stream catching up to a few days, which means a LOT of alerts at once), we notice that the number of sessions used by the different async child processes spawned by the gcnevent ingestion method creates non desirable behavior, where sessions are being shared between the processes. We've been suspecting this limitation of the connection pool of sqlalchemy for a bit (see https://davidcaron.dev/sqlalchemy-multiple-threads-and-processes/), and there is no obvious way to avoid it, other than avoiding to have too many processes using sessions at once (as they end up overlapping and sharing sessions, meaning that if one commits or closes the other can't proceed and you get an error).
Therefore, we add a simple flag to allow a fully synchronous ingestion of events, which is used by the microservice which risks to ingest multiple events at once.

Last but not least, processing the notifications in separate session - here inside of an async process - prevented us from getting the error mentioned above in this PR's description. Running this async makes a lot of sense, as otherwise whatever commit triggered the creation of notifications might have to wait for it to finis, which can take a while. Here at least it is independent of whatever triggered it, which is a very good thing as it prevents the parent process from having a rollback and being canceled if the child process encounters an error. However one might argue that might cause the same issue as the async  event ingestion that creates too many db connections. The plan for a following PR would be to add a queue (as a microservice) to process user notifications being sent to slack, email, ... instead of DB triggers.

### DIFF
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -69,7 +69,7 @@ def service():
                 if payload.find(b'Broker: Unknown topic or partition') != -1:
                     continue
                 with DBSession() as session:
-                    post_gcnevent_from_xml(payload, user_id, session)
+                    post_gcnevent_from_xml(payload, user_id, session, False)
 
         except Exception as e:
             log(f'Failed to consume gcn event: {e}')
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -102,7 +102,7 @@ Session = scoped_session(sessionmaker())
 MAX_GCNEVENTS = 1000
 
 
-def post_gcnevent_from_xml(payload, user_id, session):
+def post_gcnevent_from_xml(payload, user_id, session, asynchronous=True):
     """Post GcnEvent to database from voevent xml.
     payload: str
         VOEvent readable string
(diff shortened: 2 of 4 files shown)


mcoughlin (approved):
Great explanation. The queue idea for notifications is a great one.

Theodlz:
Thanks for the review @mcoughlin 

I'm curious to see how this kind of "reversed" situation where the notifications are async scales in prod with the many people that have them turned on. At least it is now seperate from the rest, but again nothing is really seperated because of that session pool mess...

I agree that a notification queue would make the whole thing much much cleaner and reliable.

Theodlz:
Failures look unrelated. Can I merge @mcoughlin?

PR #4042, opened 2023-03-17
GcnEvent ingestion: avoid creating more sessions when running async

Theodlz:
This PR builds on top of the previous PR merged on main, which allowed non async ingestion of event. Here, we make sure than when not running async, only one session is used instead of spawning 2 other sessions in the 2 methods we call. That way, we avoid using too many connection from the queue pool.

This PR also contains an attempt to reduce the flakiness of 2 test that are not running well on main.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -231,8 +231,10 @@ def post_gcnevent_from_xml(payload, user_id, session, asynchronous=True):
             )
             IOLoop.current().run_in_executor(None, lambda: add_contour(localization_id))
         else:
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

Theodlz:
For the skymaps I am sure, but not for the generation of obs plans. I will run the code locally in a non testing env to see if that comes from the changes.

Theodlz:
Looks like I can ingest an entire LVC skymap stream with the demo data and default plans are generated without any error (as well as skymaps), so that looks good to me. Merging.

PR #4043, opened 2023-03-18
Notification queue

mcoughlin:
This PR moves notifications from DB triggers to a notification queue service, to try to get sessions under control.


Theodlz:
This is awesome. Clearly a very nice addition. As we discussed, it makes a whole lot of sense to decouple notifying users and actually posting the data that triggers the notifications.

Precisely because of that, I do see one thing to change here. Your DB triggers shouldn't have to wait for the notification queue post method to post the user notifications I think. It should be as simple as : "is the json I sent right? Then I get a 200". Anyway, the DB triggers won't do anything extra/different if it waits for the post method of the queue to create the user notifications fully, or if it encounters an error. I don't think it needs to wait to get a response so to not block the flow of posting data to SP. Similarly, one danger here would be that the notification queue dies (or at least the API part), which would lead to very long DB triggers waiting for a timeout. Because of that, I think that in the DB triggers we need to set a fairly short max timeout, of let's say 1 or 2 seconds (after making the changes suggested above to make sure the response is instant).

mcoughlin:
@Theodlz if you are at your computer, want to quick push that to see if it affects the tests at all?

mcoughlin:
@Theodlz I just made the call run in the background so it wouldn't block. I think that is cleanest.

mcoughlin:
Closing in favor of https://github.com/skyportal/skyportal/pull/4045

PR #4044, opened 2023-03-19
User-specific OpenAI preferences

profjsb:
This PR allows each user to customize which OpenAI chat model and parameters are used in source summaries. The default remains the site-wide parameters set in `config.yaml` but those will be overridden if set by the end user.

Kapture 2023-03-19 at 08 35 10

Small other fixes are in this PR, including removing a console message and cleaning up the view in the start AI summary dialog.

### DIFF
--- a/services/openai_analysis_service/app.py
+++ b/services/openai_analysis_service/app.py
@@ -65,9 +65,7 @@ elif os.path.exists(".secret"):
 else:
     openai_api_key = None
 
-default_analysis_parameters = summary_config.copy()
 default_analysis_parameters = {
-    **default_analysis_parameters,
     **dict(
         model="gpt-3.5-turbo",
         temperature=0.1,
@@ -76,8 +74,8 @@ default_analysis_parameters = {
         frequency_penalty=0.0,
         presence_penalty=1,
     ),
+    **summary_config.copy(),
 }
-
 default_analysis_parameters["openai_api_key"] = openai_api_key
 
 
@@ -146,8 +144,12 @@ def run_openai_summarization(data_dict):
        - redshift: the (known) redshift of the object
 
     """
-    analysis_parameters = data_dict["inputs"].get("analysis_parameters", {})
-    analysis_parameters = {**default_analysis_parameters, **analysis_parameters}
+    tmp_analysis_parameters = data_dict["inputs"].get("analysis_parameters", {})
+    analysis_parameters = {
+        **default_analysis_parameters,
+        **tmp_analysis_parameters["summary_parameters"],
+        **dict(openai_api_key=tmp_analysis_parameters["openai_api_key"]),
+    }
     #
     # the following code transforms these inputs from SkyPortal
     # to the format that will give us a good summary from OpenAI.
@@ -224,7 +226,7 @@ def run_openai_summarization(data_dict):
             messages=[
                 {
                     "role": "system",
-                    "content": "You are the world expert in astronomical time-domain research.",
+                    "content": "You are an authoritative expert in astronomical time-domain research.",
                 },
                 {
                     "role": "user",
@@ -239,6 +241,7 @@ def run_openai_summarization(data_dict):
             presence_penalty=analysis_parameters["presence_penalty"],
         )
     except Exception as e:
+        log(f"OpenAI summarization failed {e}")
         rez.update(
             {
                 "status": "failure",
@@ -249,6 +252,18 @@ def run_openai_summarization(data_dict):
 
     openai_summary = response["choices"][0]["message"]["content"]
 
+    # remove dislaimers & newlines
+    sentences = openai_summary.split(". ")
+    temp_summary = []
+    for s in sentences:
+        if s.startswith("This is an automated summary"):
+            continue
+        if s.find("As an AI language model") != -1:
+            continue
+        if s.find("OpenAI") != -1:
+            continue
+        temp_summary.append(s)
+    openai_summary = ". ".join(temp_summary).replace("\n", " ")
     result = {"summary": openai_summary}
 
     if USE_PINECONE:
--- a/static/js/components/CustomizeOpenAIParameters.jsx
+++ b/static/js/components/CustomizeOpenAIParameters.jsx
@@ -225,6 +225,7 @@ const CustomizeOpenAIParameters = () => {
             schema={formSchema}
             validator={validator}
             uiSchema={uiSchema}
+            data-testid="UpdateOpenAIform"
             onSubmit={handleAISubmit}
             customValidate={validate}
           />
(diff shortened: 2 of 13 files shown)


static/js/components/CustomizeOpenAIParameters.jsx
```diff
@@ -0,0 +1,237 @@
...
+        title: "Temperature",
+        minimum: 0.0,
+        maximum: 1.0,
+        multipleOf: 0.05,
+      },
+      max_tokens: {
+        type: "integer",
+        title: "max_tokens",
+        minimum: 10,
+        maximum: 1000,
```
profjsb, line 93: The models allow for longer outputs than 1000 tokens but we're trying to create succinct summaries.

static/js/components/FollowupRequestPreferences.jsx
```diff
@@ -29,8 +29,10 @@ const FollowupRequestPreferences = () => {
   const defaultAllocationId = useSelector(
     (state) => state.profile.preferences.followupDefault
   );
-  const [selectedAllocationId, setSelectedAllocationId] =
-    useState(defaultAllocationId);
+  // set the default allocation to be -1 if nothing is in the user preferences
+  const [selectedAllocationId, setSelectedAllocationId] = useState(
+    defaultAllocationId || -1
```
profjsb, line 34: fixes unrelated warnings in the console

mcoughlin (changes requested):
Super exciting. A few comments on the form validation and text:

max_tokens: "The maximum number of tokens to generate. Must be between 1 and 2048." But the tokens is allowed to vary from 10-1000 so I suggest using those numbers for the help text.
Temperature: "What sampling temperature to use, between 0 and 2." I think this should be: "What sampling temperature to use, between 0 and 1." or the validation needs to allow up to 2.
frequency_penalty: "Number between -2.0 and 2.0." but validation gives frequency_penalty frequency_penalty must be between 0.0 and 1.0.
presence_penalty: "Number between -2.0 and 2.0." but validation gives presence_penalty presence_penalty must be between 0.0 and 1.0.

static/js/components/CustomizeOpenAIParameters.jsx
```diff
@@ -0,0 +1,238 @@
...
+        "An alternative to sampling with temperature, called nucleus sampling, where the model considers the results of the tokens with top_p probability mass. So 0.1 means only the tokens comprising the top 10% probability mass are considered.",
+    },
+    prompt: {
+      "ui:widget": "textarea",
+    },
+    max_tokens: {
+      "ui:widget": "range",
+      "ui:help":
+        "The maximum number of tokens to generate. Must be between 1 and 2048.",
+    },
```
Theodlz, line 147: As Michael also suggested, I would make this a little more clear. I think that some people (especially non native English speakers) wouldn't be sure if this is the tokens used to generate (how many tokens SP can send to OpenAI), or the ones received. It could be said in a simpler way for the end user, mentioning words rather than token.

static/js/components/CustomizeOpenAIParameters.jsx
```diff
@@ -0,0 +1,238 @@
...
+      "ui:help":
+        "Number between -2.0 and 2.0. Positive values penalize new tokens based on their existing frequency in the text so far, decreasing the model's likelihood to repeat the same line verbatim.",
+    },
+    presence_penalty: {
+      "ui:widget": "updown",
+      "ui:help":
+        "Number between -2.0 and 2.0. Positive values penalize new tokens based on whether they appear in the text so far, increasing the model's likelihood to talk about new topics.",
+    },
+  };
+  // Since we are editing exsiting follow-up requests,
```
Theodlz, line 159: remove these comments in this files

static/js/components/OpenAIPreferences.jsx
```diff
@@ -82,11 +83,12 @@ const OpenAIPreferences = () => {
           }
           label={profile.summary?.OpenAI?.active ? "Active" : "Inactive"}
         />
+        {profile.summary?.OpenAI?.active && <CustomizeOpenAIParameters />}
```
Theodlz, line 86: maybe you want a `?' also between profile and summary in case the user doesn't have any summary settings yet? So that this works whether or not we add empty keys to the user preferences in the backend.

static/js/components/OpenAIPreferences.jsx
```diff
@@ -82,11 +83,12 @@ const OpenAIPreferences = () => {
           }
           label={profile.summary?.OpenAI?.active ? "Active" : "Inactive"}
```
Theodlz, line 84: same comment

Theodlz (commented):
Hi Josh, this is very nice. I would also make the changes Michael suggested so that these settings can be easier to understand for a "non-AI-aware user."

services/openai_analysis_service/app.py
```diff
@@ -249,6 +252,18 @@ def run_openai_summarization(data_dict):
 
     openai_summary = response["choices"][0]["message"]["content"]
 
+    # remove dislaimers & newlines
+    sentences = openai_summary.split(". ")
+    temp_summary = []
+    for s in sentences:
+        if s.startswith("This is an automated summary"):
```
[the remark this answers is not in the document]
profjsb: I found I needed to add this here since gpt-4 seems to like to add this disclaimer.

static/js/components/CustomizeOpenAIParameters.jsx
```diff
@@ -176,7 +172,7 @@ const CustomizeOpenAIParameters = () => {
     if (formData.top_p < 0.0 || formData.top_p > 1.0) {
       errors.top_p.addError("top_p must be between 0.0 and 1.0");
     }
-    if (formData.frequency_penalty < 0.0 || formData.frequency_penalty > 1.0) {
+    if (formData.frequency_penalty < -2.0 || formData.frequency_penalty > 2.0) {
       errors.frequency_penalty.addError(
         "frequency_penalty must be between 0.0 and 1.0"
```
mcoughlin, line 177: Also fix this text.

static/js/components/CustomizeOpenAIParameters.jsx
```diff
@@ -188,7 +184,7 @@ const CustomizeOpenAIParameters = () => {
         "must be an Open AI gpt model. See https://platform.openai.com/docs/models/overview for more information."
       );
     }
-    if (formData.presence_penalty < 0.0 || formData.presence_penalty > 1.0) {
+    if (formData.presence_penalty < -2.0 || formData.presence_penalty > 2.0) {
       errors.presence_penalty.addError(
         "presence_penalty must be between 0.0 and 1.0"
```
mcoughlin, line 189: Also fix this text.

mcoughlin (approved):
Looks like just a small update necessary to the error messages too. Otherwise LGTM.

static/js/components/CustomizeOpenAIParameters.jsx
profjsb: done

static/js/components/CustomizeOpenAIParameters.jsx
profjsb: done

PR #4054, opened 2023-03-22
Add GCN Tags

mcoughlin:
This PR makes it possible to add GCN tags.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -712,18 +712,19 @@ analysis_services:
 comments_folder: persistentdata/comments
 photometric_series_folder: persistentdata/phot_series
 
-classifications:
-  # A few examples that show up in the demo data. You can modify the config on deploy to fill out the colors and types you want.
-  {
-    "variable": "#3a87ad",
-    "Time-domain Source": "#ff0000",
-  }
+colors:
+  classifications:
+    # A few examples that show up in the demo data. You can modify the config on deploy to fill out the colors and types you want.
+    {
+      "variable": "#3a87ad",
+      "Time-domain Source": "#ff0000",
+    }
 
-gcnTags:
-  {
-    "BNS": "#468847",
-    "NSBH": "#b94a48",
-    "BBH": "#333333",
-    "GRB": "#f89406",
-    "AMON": "#3a87ad",
-  }
+  gcnTags:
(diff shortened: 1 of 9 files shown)


config.yaml.defaults
```diff
@@ -718,3 +718,12 @@ classifications:
     "variable": "#3a87ad",
     "Time-domain Source": "#ff0000",
   }
+
+gcnTags:
```
guynir42, line 722: Are these color codes? If so perhaps better to name this config key something more specific?

skyportal/handlers/api/gcn.py
```diff
@@ -199,6 +199,11 @@ def post_gcnevent_from_xml(payload, user_id, session, asynchronous=True):
 
     if skymap is None:
         log(f"No skymap found for event {dateobs}")
+
+        print(tags)
```
guynir42, line 203:
```suggestion
```

skyportal/handlers/api/gcn.py
```diff
@@ -462,6 +467,111 @@ async def get(self):
             tags = session.scalars(sa.select(GcnTag.text).distinct()).unique().all()
             return self.success(data=tags)
 
+    @auth_or_token
+    def post(self):
```
guynir42, line 471: I'm surprised this works without giving null arguments.

static/js/components/GcnTags.jsx
```diff
@@ -63,22 +107,56 @@ const RenderTags = ({ gcnEvent, show_title = false }) => {
   });
   const localizationTagsUnique = [...new Set(localizationTags)];
 
+  const permission = userProfile.permissions.includes("System admin");
```
guynir42, line 110: Only the admin gets to delete these? What about users that posted a tag?

guynir42 (approved):
I have minor comments.

config.yaml.defaults
profjsb: I agree--this whole section could be called "colors". Ie.

```yaml
colors:
   classifications:
        variable: #xxxx
         ....
    gcn_tags:
         BNS: #xxxx
```

profjsb:
Do you need to also update the classification colorization now that the config section has changed structure?

PR #4055, opened 2023-03-22
GcnEvent: SkyMap manual reingestion + avoid repetitive API calls from frontend

Theodlz:
This PR decouples the gcnevent + notice ingestion from the localization ingestion. It adds a handler allowing a user to retry the skymap ingestion from the frontend if the skymap associated with a gcn notice was not posted correctly. This happens for example with FERMI notices, where the localization can become available minutes or hours, or even 1 day after the notice has been emitted...

Also, I noticed a lot of redundant frontend API calls, putting some unnecessary load on the app and the client side. Initially, I wanted to take care of that problem hoping to fix the "page freezing" bug we are experiencing, but that seems to come from somewhere else. I suspect a faulty loop somewhere in one of the components that ends up eating a lot of RAM.

PS: I haven't run any tests locally, so please wait for them to all pass successfully!

Here's a short video showing how that works (sorry for the quality github is very limited size wise):

https://user-images.githubusercontent.com/49785117/227049516-9ba89aa8-963c-4723-b09f-f8ba0dbc4db4.mp4

EDIT: The frontend for the gcn notices changed a little bit, to make it clearer:
image


Theodlz:
Obsplan failures. Dont merge until fixed

mcoughlin (approved):
Lots of nice fixes here. LGTM.

PR #4056, opened 2023-03-22
WIP - Preview for image analysis

lydie10:
Hi @mcoughlin @Theodlz 
With @Carla802 we tried to implement a preview for the image submitted in the image analysis form but we are stuck with this error 
error

First, we tried to follow the example of the code for the finding chart. We created a function image_analysis_preview(image_data) which creates a plot with the uploaded image and saves it in png. The error seems to be when we call self.send_file

We then tried  an other way by doing like in the comment.py file, which uses the bytes output of the get_fits_preview() function and the same error is displayed when calling self.write().

This part of the code is in the image_analysis.py file. 

In the ImageAnalysisForm.jsx file, there is only the code for the new field in the form which allows us the select a preview of the image instead of posting the analysis.

Thank you !


mcoughlin:
@Theodlz Some version of this would be very helpful for the Image Analysis stuff.

PR #4060, opened 2023-03-24
Follow-up requests: missing SEDM defaults + re-submit on failure

Theodlz:
This PR addresses Don and Jesper's comments about the SEDM new fields that needs to be optional and always have default value. 

Moreover, this makes it possible to resubmit if the previous submission has a  "failed to submit" status. This is useful for example if you created a followup requests with an allocation that was missing credentials which you fixed right after, or if the instrument's API wasn't available for a sec... That way, you can retry submitting your request without having to delete it and input all the values again.


mcoughlin (approved):
LGTM

PR #4063, opened 2023-03-24
remove 60s single frame exposure for SEDMv2

broulston:
Remove the 60s single frame option from SEDMv2 follow-up requests. It is not a valid option within the software for the imaging mode.


mcoughlin (approved):
LGTM

mcoughlin:
@broulston Please run the pre-commit to pass the linting requirements

PR #4064, opened 2023-03-25
Bugfix: Incorrect instrument form type showing on GCN Page

Theodlz:
Right now, the normal instrument forms are showing in the obsplan form, not the instrument obsplan forms.


mcoughlin (approved):
Front-end looks good. LGTM.

PR #4065, opened 2023-03-25
Bugfix: Microservices not loaded in production

Theodlz:
Due to how the difference in url to access the app between dev and prod (not ports to specify in prod), we didn't catch an error that caused some microservices to never run in prod 😓. This PR hopefully fixes that.


mcoughlin (approved):
LGTM

PR #4067, opened 2023-03-28
Status updates should also return success

mcoughlin:
This PR fixes a bug where status updates on followup_requests did not return success.


guynir42 (approved):

guynir42:
Do you know what the test failures are about?

mcoughlin:
@guynir42 We have a few flaky tests.

guynir42:
Try to mark them as flaky if you can on the next PR. I really don't want us to slip back into bad habits.

PR #4072, opened 2023-03-30
Observation ingestion async

mcoughlin:
This PR fixes a repeated call to observation ingestion, it should only be async.


Theodlz (approved):
LGTM

PR #4073, opened 2023-03-31
Instruments: has_fields and has_region flag

Theodlz:
This PR adds 2 flags to the instrument table.
These values are updated based on addition of regions and fields on instruments. 
Moreover, we added some temporary code in the instrument handler to avoid having to regenerate all instruments fields and regions in production. Even if temporary, that code isn't incorrect and could even remain in the handler if we want to keep it there.


Theodlz:
Failures are unrelated. Merging

PR #4095, opened 2023-04-01
Number of fields in Instrument table

mcoughlin:
This PR adds the number of fields to the Instrument table.


skyportal/models/instrument.py
```diff
@@ -214,6 +372,16 @@ def api_class_obsplan(self):
...
+    @property
+    def number_of_fields(self):
+        if not self.has_fields:
+            return 0
+        stmt = sa.select(InstrumentField).where(
+            InstrumentField.instrument_id == self.id
+        )
+        count_stmt = sa.select(func.count()).select_from(stmt.distinct())
+        return DBSession().execute(count_stmt).scalar()
+
```
Theodlz, line 384: Do you need a query here? Given that the instrument has a relationship on fields, I'm wondering if you can't just check the `len(self.fields)` or something similar instead. Worth trying.

Theodlz (commented):
Looks good. Just one comment

skyportal/models/instrument.py
mcoughlin: self.fields will load all that data, which I think we want to avoid.

Theodlz: Ok that makes sense. LGTM

Theodlz (approved):
LGTM

PR #4096, opened 2023-04-01
Queuify observation plans

mcoughlin:
This PR moves the observation plan generation to a queue.


Theodlz (changes requested):
Hi @mcoughlin. This is great, but I'm not sure it's exactly what we want. I think that the queue should take care of running the heavier computations that need a DB connections, which is what is problematic to have as a bunch of async processes and so on. 

I think that grabbing the default obsplan from db and config, so just quick db reads can still be done where it is now. What the queue that might help in our case should do, is just receive the json of a plan and whatever other data is necessary, and take care of running gwemopt and submitting whatever the results are back to the db, as well as update the status.

mcoughlin:
@Theodlz Ok this does seem simpler.

Theodlz (approved):
Now that I contributed to that PR I might be a little biased but this looks ready to me! 

LGTM

PR #4098, opened 2023-04-02
Missing type in classification docstring

mcoughlin:
This PR fixes a missing type in the docstring.


mcoughlin:
It was this typo that prevented apispec from upgrading this morning. Merging.

PR #4099, opened 2023-04-04
Custom Engine args

Theodlz:
It is already possible in baselayer to specify the engine args of the database (pool size, overflow, ...) in the config, but a piece of code in SP specified other values than the default ones in a way that creates issues when specifying them in the config. 

This PR addresses that issue by using the SP default values (that are a little higher than the baselayer ones) if not provided in the config, and otherwise uses the ones specified in the config.


mcoughlin (approved):
LGTM

PR #4100, opened 2023-04-05
Shift user permissions

mcoughlin:
This PR lowers the permissions threshold for signing up for shifts.

### DIFF
--- a/skyportal/handlers/api/shift.py
+++ b/skyportal/handlers/api/shift.py
@@ -271,7 +271,7 @@ class ShiftHandler(BaseHandler):
 
 
 class ShiftUserHandler(BaseHandler):
-    @permissions(["Manage shifts"])
+    @auth_or_token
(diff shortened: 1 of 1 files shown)


Theodlz (changes requested):
Hi @mcoughlin, this looks good to me. However as we discussed there is a clear performance issue when you have more than lets say 50 shifts, as there is no pagination right now.

I'm happy to merge this PR, but only if we remove the hydration in action.js until we have a PR addressing the pagination issue, as all users might be affected otherwise.

mcoughlin:
@Theodlz I dont see the shift stuff in the actions.

Theodlz:
Ah! I might have removed it in the past for similar reasons... my bad.

Theodlz (approved):

Theodlz:
LGTM @mcoughlin

PR #4101, opened 2023-04-06
Check if analysis service exists

mcoughlin:
This PR checks if an analysis service is found before retrieving info about it.


Theodlz (approved):
LGTM

PR #4102, opened 2023-04-06
Shifts: Lighter API responses

Theodlz:
The shifts PR was my first ever PR I opened on SkyPortal, and ... some things were questionable at best 👀. 
In this PR, we get rid of a lot of the flaws of the current shift page, while ensuring that the response from the shift handler is much smaller, and we avoid refreshing the whole shifts list as much as possible.

We noticed in the GRB/ZTF GRANDMA campaign last there that the shifts would become almost unusable as you added more and more of them, because the handler would edit each shift to add it the group, group users, shift users, and users. Here, we return the shifts as is, with no joined loads. Instead, we only "enrich" the shift we get when querying a specific one by id. Some modifications were made to the frontend to make it a little more lightweight too.

There's also a little bit of a cleanup of the parameter names used to pass shift ids to other components to follow the naming scheme used with other resourceTypes.

NEXT STEP: In a following PR, I plan to add pagination based on date. That is easy to implement on API side, but I'm unsure how I can make it work with the calendar from react-big-calendar.


skyportal/models/shift.py
```diff
@@ -179,3 +181,9 @@ class Shift(Base):
 ShiftUser.delete = CustomUserAccessControl(shiftuser_delete_access_logic)
 
 ShiftUser.create = ShiftUser.read
+
+Shift.shift_users_ids = column_property(
```
mcoughlin, line 185: Can we add some comment to explain what is happening here?

skyportal/tests/frontend/test_shifts.py
```diff
@@ -223,14 +231,13 @@ def test_shift(
...
 
-    shift_on_calendar = f'//*/span/strong[contains(.,"{form_name}")]'
-    # check for API shift
-    driver.wait_for_xpath(
-        shift_on_calendar,
-        timeout=30,
-    )
+    # scroll to the top of the page
+    driver.execute_script("window.scrollTo(0, 0);")
+    time.sleep(1)
```
mcoughlin, line 236: Why does sleep help here?

mcoughlin (approved):
Just a few small notes.

skyportal/tests/frontend/test_shifts.py
Theodlz: Let me try to remove it, I had issues with that part of the test but I made other chances that might have fixed it and maybe I can remove this one.

skyportal/models/shift.py
Theodlz: Sure! Will do.

Theodlz:
Done @mcoughlin, do you want to have a look or can I merge?

mcoughlin:
@Theodlz Go for it!

PR #4103, opened 2023-04-06
Bugfix: make load_demo_data

Theodlz:
It seems that a recent release of simple-json breaks some of the json compliant verification we do in the data loader method. 

It's very probable that we could just change some code on our side rather than pin a version, but given that its still a little unclear what needs to chance, it is best to pin a version while we modify the code to allow unit tests and the app to still function.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -86,3 +86,4 @@ shapely==1.8.1
 stdpipe==0.1
 openai[embeddings]==0.27.2
 pinecone-client==2.2.1
+simplejson>=3.17.2,<=3.18.4


Theodlz:
@mcoughlin no need to wait for anything other than see if the make load_demo_data works. So I think we can merge quickly once that part of the tests has passed.

guynir42 (approved):
Seems like a good thing to do for now, but please add an issue to make sure we are compliant with the new version.

mcoughlin (approved):
LGTM

Theodlz:
I opened an issue @guynir42, thanks!

Theodlz:
Loading the demo data works again. Merging.

PR #4105, opened 2023-04-07
Candidate Scanning: GCN-based filtering

Theodlz:
This PR adds the same kind of filtering found on the sources list to filter candidate by: event, localization, cumulative probability. But also filtering them by: first detection date after, last detection date before, and min number of detections.


mcoughlin (approved):
LGTM

PR #4106, opened 2023-04-07
GCN Bugfix: Summary won't open

Theodlz:
In the GCN Summary component, we fetch groups and users and until those are fetched, we can't open the component. 
This PR removes this component, as anyway groups are hydrated when loading SP, and we shouldn't fetched all users but only the group_users of the selected group.


Theodlz:
@mcoughlin tagged you again as I added some of the changes we talked about with Mansi in this PR, I thought you might want to have a quick look at the last 2 commits to see if that's fine by you.

mcoughlin (approved):
LGTM

PR #4108, opened 2023-04-08
LocalizationTile: Partitioned Table for performance

Theodlz:
This PR is a draft, where I explore the potential performances benefits of using partitioned tables for the localization tiles (here I tried to do it by partitioning by month). In production, the table is so large that even with the right indexes, the query we need to run to get which instrument field tiles overlap with the LocalizationTiles of a skymap take 45 minutes and the session just times out... Whereas locally with a less populated database, it is pretty much instantaneous (lets say between 0.2 and 0.5 seconds, even with 300 skymaps.

After the changes added in this PR, I experience some improvements, as the same query takes around 0.08 and 0.25 seconds, while ingestion LVC notices from multiple kafka topics, as well as Fermi notices. At the time where I run that code, we are the 8th of April, meaning that I have events from the 28th of March till the 8th of April. So the paritioning isnt really used to its full potential. To test that, I would need to ingest events on a larger time period like 4 to 5 months.

One problem right now, is generating new partitions on the flight. It would be much better than just creating all the tables in a loop. Also, we probably want to have a setup where alembic doesnt need to keep track of every single table that acts as a partition, but just the main table to which they are attached.

I managed to have create partitions automatically when needed in a seperate script but that failed in skyportal, probably due to how we register the schemas?


Theodlz:
@mcoughlin you'll probably want to have a look. This is in a VERY early stages but it seems to work.

mcoughlin:
@Theodlz I am testing your code, which seems to work with https://github.com/growth-astro/growth-too-marshal-gcn-notices which covers a good range of time.

mcoughlin:
@Theodlz We could just commit this and try it out on preview? If it works, we pin the updated main, otherwise we go back to try to understand why?

Theodlz:
> @Theodlz We could just commit this and try it out on preview? If it works, we pin the updated main, otherwise we go back to try to understand why?

well issue is alembic looks back at the tree of migrations everytime so that would break that test forever I believe. Let me see if I can do something about it

Theodlz:
Hi @mcoughlin, the `upgrade` migration is ready. Now, I need to implement the`downgrade` one, which is important to have in case that the migration doesn't work as expected. I'll focus on that tomorrow morning, so that we can hopefully deploy tomorrow afternoon. If we preview still works, and that getting the overlap of tiles works once again in production, we can then modify all the pieces of code that query LocalizationTile to focus the right partition just as I added here in the instrument handler and generate_plan method. 

I'll tag you once the downgrade migration is finished. It should be easier than the upgrade.

mcoughlin (approved):
What a migration script.

PR #4110, opened 2023-04-09
Localizations in persistentdata

mcoughlin:
This PR saves localizations to persistentdata (for ease of download).


skyportal/models/files.py
guynir42, line 1: Maybe this should go in a utils folder?

skyportal/models/localization.py
```diff
@@ -219,6 +230,35 @@ def center(self):
...
+
+    def delete_data(self):
+        """
+        Delete the localizations's data from disk.
+        """
+
+        delete_file_data(self._attachment_path)
+
+        # reset the filename
+        self._localization_path = None
```
guynir42, line 260: Is there a better way to handle errors in this part? For example, if the file is not found, and we remove the path to it, we lose our ability to track it down later. I'm not sure what the right thing to do here.

guynir42 (approved):
I have a couple of small suggestions.

skyportal/models/files.py
mcoughlin: good idea.

skyportal/models/localization.py
guynir42: My first instinct would be to let the delete_file_data pass on any exceptions it has, and catch it outside, where you can put the `self._localization_path = None` line inside the try/except block. That could mean we leave path strings that are dead ends, but you'd have the opportunity to re-check and delete them explicitly. 
But that's just one option, and we should consider what is the best policy.

skyportal/handlers/api/gcn.py
```diff
@@ -261,6 +261,15 @@ def post_skymap_from_notice(dateobs, notice_id, user_id, session, asynchronous=T
...
         localization_id = localization.id
 
+        if url is not None:
+            r = requests.get(url, allow_redirects=True)
+            data_to_disk = r.content
+            urlpath = urlsplit(url).path
+            localization_name = os.path.basename(urlpath)
+            if data_to_disk is not None:
+                localization.save_data(localization_name, data_to_disk)
+                session.commit()
```
Theodlz, line 271: File availability from Fermi or LVC has been a little random. So please make sure that this runs last after we ingest everything else, and wrap the requests.get in a try except so we avoid any problems.

skyportal/models/localization.py
```diff
@@ -219,6 +230,35 @@ def center(self):
...
+        """
+        return self._localization_path
+
+    def save_data(self, filename, file_data):
+        """
+        Save the localization's data to disk.
+        """
+
+        # there's a default value but it is best to provide a full path in the config
+        root_folder = cfg.get('localizations_folder', 'localizations_data')
```
Theodlz, line 245: Please add this to the default config

Theodlz (commented):
I agree with Guy's comments, and thanks for addressing those! I just have 2 comments, but also one suggestion. Now that we keep the file, maybe we should return it when using the handler to download a localization, rather than what we do right now to avoid unnecessary load on the DB and also keep the HEADERS of the original fits file when downloading from SkyPortal.

skyportal/models/localization.py
mcoughlin: yeah its in there.

skyportal/models/localization.py
mcoughlin: done

Theodlz:
@mcoughlin I'll take a better look at the code. The make load_demo_data failed to ingest an event here on git (returned a 504 aka timeout), meaning that something is making it flaky here. I suspect this is simply because we get the file with a request which must have timed out. I'll try to fix that.

Theodlz (approved):
LGTM

PR #4111, opened 2023-04-10
Facility queue logging

mcoughlin:
This PR adds further logging in the facility queue. If you have any way to improve this queue based on the others, please do so @Theodlz.


Theodlz (approved):
Hi @mcoughlin, this looks good. I definitely want to use the same parallel queue/handler model that I implemented for the obsplans here and in the notification queue. I already planned on opening a PR to modify both when I have the time to do so. This one looks good, LGTM

PR #4112, opened 2023-04-11
WIP: Find fields with gwemopt configurable + minor quality of life improvements

Theodlz:
This PR makes it possible to choose which method to use to get the fields to observe when creating an obs plan: Postgres query, or gwemopt code. This is because currently in production, the Postgres query does not return in time, as the queries involving instrument field tiles and localization tiles do not scale well as the localization tiles table grows larger.

Also, we add several minor changes that improve the user experience and make many GCN-related features more practical to use (this is based on multiple astronomers' feedback).

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -20,6 +20,7 @@ app:
 
   observation_plan:
     default_filters: ['ztfg', 'ztfr', 'ztfi']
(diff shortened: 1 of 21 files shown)


mcoughlin (approved):
Thanks for enabling this as an option.

PR #4114, opened 2023-04-11
deepcopy config

mcoughlin:
This PR fixes api_key being popped out never to return.

### DIFF
--- a/services/openai_analysis_service/app.py
+++ b/services/openai_analysis_service/app.py
@@ -58,8 +58,7 @@ if (
 summary_config = copy.deepcopy(cfg['analysis_services.openai_analysis_service.summary'])
 if summary_config.get("api_key"):
     # there may be a global API key set in the config file
-    openai_api_key = summary_config.get("api_key")
-    summary_config.pop("api_key")
+    openai_api_key = summary_config.pop("api_key")
 elif os.path.exists(".secret"):
     # try to get this key from the dev environment, useful for debugging
     openai_api_key = yaml.safe_load(open(".secret")).get("OPENAI_API_KEY")
--- a/services/openai_analysis_service/app.py
+++ b/services/openai_analysis_service/app.py
@@ -1,3 +1,4 @@
+import copy
 import io
 import os
 import functools
@@ -54,7 +55,7 @@ if (
         log(f"index {summarize_embedding_index} created in pinecone")
     USE_PINECONE = True
 
-summary_config = cfg['analysis_services.openai_analysis_service.summary']
+summary_config = copy.deepcopy(cfg['analysis_services.openai_analysis_service.summary'])
 if summary_config.get("api_key"):
     # there may be a global API key set in the config file
     openai_api_key = summary_config.get("api_key")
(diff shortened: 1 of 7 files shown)


services/openai_analysis_service/app.py
```diff
@@ -54,7 +55,7 @@
         log(f"index {summarize_embedding_index} created in pinecone")
     USE_PINECONE = True
 
-summary_config = cfg['analysis_services.openai_analysis_service.summary']
+summary_config = copy.deepcopy(cfg['analysis_services.openai_analysis_service.summary'])
```
guynir42, line 58: I think you can just do `cfg.copy([...])` and you get the same thing. But I'm not against an explicit "deep copy" command either.

skyportal/handlers/api/analysis.py
```diff
@@ -54,6 +57,18 @@
 
 DEFAULT_ANALYSES_DAILY_LIMIT = 100
 
+# check for API key
+summary_config = copy.deepcopy(cfg['analysis_services.openai_analysis_service.summary'])
+if summary_config.get("api_key"):
+    # there may be a global API key set in the config file
+    openai_api_key = summary_config.get("api_key")
+    summary_config.pop("api_key")
```
guynir42, line 65: Can we combine these two lines into just "pop"?

static/js/components/SourceDesktop.jsx
```diff
@@ -298,6 +298,8 @@ const SourceDesktop = ({ source }) => {
     </Tooltip>
   );
 
+  console.log("source", source);
```
guynir42, line 301:
```suggestion
```

guynir42 (approved):
I have minor comments.

skyportal/handlers/api/analysis.py
mcoughlin: We use the summary_config later on.

services/openai_analysis_service/app.py
mcoughlin: I think it needs to be a deepcopy (copy is shallow).

skyportal/handlers/api/analysis.py
guynir42: yeah, do `openai_api_key=summary_config.pop('api_key')`

services/openai_analysis_service/app.py
guynir42: I think dictionary `copy` is deep by default, but I'm going off my memory.

guynir42: Yeah, you are right.

skyportal/handlers/api/analysis.py
mcoughlin: Ah got ya. Done.

services/openai_analysis_service/app.py
guynir42: Although if you are only popping the top level keys I'm not sure it matters. 
In any case, `deepcopy` is more explicit so I think it is fine.

static/js/components/SourceDesktop.jsx
mcoughlin: Done.

PR #4116, opened 2023-04-12
Shifts: Weekly

Theodlz:
There is already an option to split the selected time period by day to create shifts that repeat daily fairly easily. 

This simply adds the possibility to do the same by week.


mcoughlin (approved):
LGTM

PR #4118, opened 2023-04-12
Source Page classification: Avoid errors when config colors aren't fetched yet.

Theodlz:
It seems that if the classifications and source are fetched before the config, the page crashed. This minor change should fix that.


mcoughlin (approved):
LGTM

PR #4121, opened 2023-04-13
Notification Queue: Multi-threading

Theodlz:
In this PR, we make sure that the handler pushing to the queue, and the service processing it run on different threads. Using only asyncio means that right now both are running on the same thread, in the same event loop. Because of that, we may end up having timeouts when pushing a notification, which could impact every feature that needs might trigger notifications (many).

By reducing delays in general, this PR also should help with the flakiness of some tests.

---------------------------------------------------------

Here are 2 tests who's flakiness is more persistent, and should be addressed in later PRs:
- test_analysis_with_file_input -> unknown origin
- test_gcn_tach -> frontend issue, page crashes here but not locally

*Both are frontend tests*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,5 @@ data/gitlog.txt
 # test results
 test-results/
 webdriver-console/
+
+persistentdata/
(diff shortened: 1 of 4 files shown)


mcoughlin (approved):
LGTM

PR #4124, opened 2023-04-14
Python side field picking for obsplan

guynir42:
first attempt at making a python side obs plan


guynir42:
@mcoughlin don't "review" yet, but have a look to see if anything looks like a bug...

Theodlz:
Hi Guy. This looks very nice. I'll play around with the code today to compare the skyportal with db, skyportal with python and gwemopt implementations.

mcoughlin:
@Theodlz Should we fix this one here or open another PR?

mcoughlin:
We got this one in a separate PR.

PR #4126, opened 2023-04-14
Several frontend improvements and bugfixes

Theodlz:
This PR addresses 3 issues:

- When uploading a fits file with special characters in its name (like the comma in LVC skymaps), you would end up generating and storing the localization with an odd name that would break the handler to get a localization).
- When creating weekly shifts, if running skyportal locally all is fine but it seems that in production timezones issues arise. This PR should address that.
- When modifying allocations, the current values wouldn't be loaded in the form. Therefore the user had to re-add everything by hand and would risk modifying some values unintentionally.


mcoughlin (approved):
Bug fixes are good fixes.

PR #4128, opened 2023-04-15
Caching:  Instrument fields & localization overlap query

mcoughlin:
This PR tests out caching of the fields / localization query. Should be updated after https://github.com/skyportal/skyportal/pull/4124.


skyportal/facility_apis/observation_plan.py
```diff
@@ -551,14 +557,25 @@ def generate_plan(
...
             if use_skyportal_fields is True:
                 for request in requests:
-                    field_tiles_query = sa.select(InstrumentField.field_id).where(
-                        LocalizationTile.localization_id == request.localization.id,
-                        LocalizationTile.probdensity >= min_probdensity,
-                        InstrumentFieldTile.instrument_id == request.instrument.id,
-                        InstrumentFieldTile.instrument_field_id == InstrumentField.id,
-                        InstrumentFieldTile.healpix.overlaps(LocalizationTile.healpix),
+                    query_id = (
+                        f"{str(request.localization.id)}_{str(request.instrument.id)}"
```
guynir42, line 561: What happens if between requests to the same instrument the user changed the integrated probability?

mcoughlin:
So the only problem is if the first request was for a small integrated probability, as this selection is just for potential fields to consider. But the first request should always be a default plan, which is a big one.

guynir42:
It sounds like in certain edge cases this will not work (e.g., the default plan had a smaller prob than the new one, or if the cache times out). How complicated would it be to put the important inputs as part of the caching (e.g., in the filename or something)?

guynir42:
Also one more thing to consider: this query should not be a long one. If you get to the point that either the python side or DB side query works in under a minute, wouldn't that make the whole caching thing just an unnecessary complication? I don't feel this part of the calculation is inherently slow (it shouldn't be).

skyportal/facility_apis/observation_plan.py
Theodlz: Agreed. I think we want to cache by localization id, instrument id AND cumprob.

mcoughlin: @Theodlz Sounds good. Let's try this after Guy's PR

Theodlz:
> Also one more thing to consider: this query should not be a long one. If you get to the point that either the python side or DB side query works in under a minute, wouldn't that make the whole caching thing just an unnecessary complication? I don't feel this part of the calculation is inherently slow (it shouldn't be).

Hi Guy. On the other hand I think its pretty much necessary. If it takes less than a minute but more than 30 seconds means a timeout. In general even with partitioning those queries are not very fast, and pretty expensive. Even locally you can have some delays for the API endpoints that return localization tiles overlap, and caching those especially at 0.9 cumprob which is the default would help a lot on the client side.

guynir42:
@Theodlz the 30s is a lot to wait on, but I wonder if the rest of the observation plan scheduling isn't much longer. 
If the `gwemopt` part is longer than ~minute time scale then you'd call this thing asynchronously and probably wouldn't care much about the caching. If the rest of the call is much faster then caching is the right thing to do.

Theodlz:
> @Theodlz the 30s is a lot to wait on, but I wonder if the rest of the observation plan scheduling isn't much longer. 
> If the `gwemopt` part is longer than ~minute time scale then you'd call this thing asynchronously and probably wouldn't care much about the caching. If the rest of the call is much faster then caching is the right thing to do. 

I do agree that for the obs plan scheduling, it might help but its not critical. I think the caching however would be super useful for the frontend queries to see instrument fields on skymaps, which are timing out for now.

guynir42:
@Theodlz ahh I did not think about that. In that case I agree it would be very useful.

Theodlz (approved):
LGTM

PR #4129, opened 2023-04-16
Localization plot: Mollweide projection and bugfix

Theodlz:
This PR adds:
- a projection dropdown to switch between mollweide and orthographic, which can be easily extended to more projections.
- Default to mollweide in the obsplan list for better visibility.
- Fixed bug where changing localization wouldn't load the newly selected one.


mcoughlin (approved):
Trippy.

PR #4131, opened 2023-04-17
Edit shifts

mcoughlin:
This PR makes it possible to edit shifts.

### DIFF
--- a/skyportal/handlers/api/mmadetector.py
+++ b/skyportal/handlers/api/mmadetector.py
@@ -183,7 +183,7 @@ class MMADetectorHandler(BaseHandler):
 
             schema = MMADetector.__schema__()
             try:
-                schema.load(data)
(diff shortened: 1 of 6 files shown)


guynir42 (commented):
Maybe add a small test (on its own or as part of one of the other shift tests)?

Theodlz:
@mcoughlin, the save button is deactivated for me, even as super admin. I'm on it.

Theodlz:
It's the "isInvalid" logic that is incorrect. A shift doesn't need a description or a required user number. So if a shift doesn't have one, you can't modify it here.

mcoughlin:
@Theodlz I just removed that logic.

Theodlz:
Great! I'm fixing some more stuff right now

Theodlz (approved):
LGTM. Failures are unrelated. Merging

PR #4132, opened 2023-04-17
GCN Table and GCN Candidate filtering: Search by dateobs

Theodlz:
Previously, it was impossible to search events by date in the gcn event list, and we could only filter candidates by the last 10 events. In this PR, we make it possible to search events by typing partial dateobs strings in both components.

This will be followed by a (less urgent) PR where the code added to the candidate list will be reused in all other components where we need a dropdown to select events.


mcoughlin (approved):
LGTM

PR #4136, opened 2023-04-18
Add origin cache

mcoughlin:
This PR caches origins, to make those queries faster.


mcoughlin:
@Theodlz It seems to break GA photometry tests, but passes locally.

mcoughlin:
@Theodlz Can we close this?

Theodlz:
Yep we should close for now

PR #4137, opened 2023-04-18
Rejection tag query performance

mcoughlin:
This PR tries to improve tag query performance by following classification style.


Theodlz (approved):
LGTM. I guess that the only way to assess the performance gain from these changes is to try in production.

PR #4139, opened 2023-04-18
Send observation plan by slack and email

mcoughlin:
This PR enables sending observation plans by slack and email.

### DIFF
--- a/skyportal/facility_apis/observation_plan.py
+++ b/skyportal/facility_apis/observation_plan.py
@@ -5,6 +5,7 @@ import humanize
 import json
 import pandas as pd
 from regions import Regions
+import requests
 from datetime import datetime, timedelta
 import numpy as np
 import os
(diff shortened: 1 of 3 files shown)


skyportal/facility_apis/observation_plan.py
```diff
@@ -22,16 +23,23 @@
...
 env, cfg = load_env()
 
+SLACK_URL = f"{cfg['slack.expected_url_preamble']}/services"
+
 default_filters = cfg['app.observation_plan.default_filters']
 
 use_skyportal_fields = cfg['app.observation_plan.use_skyportal_fields']
 
+email = False
+if cfg["email_service"] == "sendgrid" or cfg["email_service"] == "smtp":
```
guynir42, line 40: Should these be a `get` in case `email_service` is not defined in the config?

guynir42 (approved):

skyportal/facility_apis/observation_plan.py
mcoughlin: Fixed

PR #4140, opened 2023-04-20
Instrument column uploads

mcoughlin:
This PR allows for instrument column uploads.


guynir42 (approved):
Small adjustment required.

PR #4141, opened 2023-04-20
Follow-up requests to email/slack

mcoughlin:
This PR adds email/slack request when making follow-up requests.

### DIFF
--- a/static/js/components/EditFollowupRequestDialog.jsx
+++ b/static/js/components/EditFollowupRequestDialog.jsx
@@ -38,6 +38,8 @@ const EditFollowupRequestDialog = ({
       obj_id: followupRequest.obj_id,
       payload: formData,
     };
+    console.log("json", json);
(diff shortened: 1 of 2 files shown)


static/js/components/EditFollowupRequestDialog.jsx
```diff
@@ -38,6 +38,8 @@ const EditFollowupRequestDialog = ({
       obj_id: followupRequest.obj_id,
       payload: formData,
     };
+    console.log("json", json);
+    console.log("followupRequest", followupRequest);
```
guynir42, line 42: Can we remove these?

guynir42 (approved):
Looks good. Only comment is some console outputs.

static/js/components/EditFollowupRequestDialog.jsx
mcoughlin: Thanks for flagging this.

PR #4142, opened 2023-04-20
Plotting preferences: Temporarely deactivate all origins fetching

Theodlz:
In this PR, we temporarely prevent the frontend from fetching ALL unique origins from the photometry table, which times out and puts too much load on the database.

This PR will be followed by a second PR, where we change the feature entirely to avoid running that query.


mcoughlin (approved):
LGTM

Theodlz:
@mcoughlin I'll double checking if there is a test for it that I should comment too, so please wait before merging!

Theodlz:
Merging now as we have a long migration to run in prod today. This is a frontend only change, and the tests concerned pass locally.

PR #4143, opened 2023-04-21
ObservationPlanRequest query parameters

mcoughlin:
This PR adds query parameters for ObservationPlanRequests.

### DIFF
--- a/skyportal/handlers/api/observation_plan.py
+++ b/skyportal/handlers/api/observation_plan.py
@@ -1,3 +1,4 @@
+import arrow
 from datetime import datetime, timedelta
 import functools
 import io
@@ -24,6 +25,7 @@ import jsonschema
 import requests
 from marshmallow.exceptions import ValidationError
 import sqlalchemy as sa
+from sqlalchemy import func
 from sqlalchemy.orm import joinedload, undefer
 from sqlalchemy.orm import sessionmaker, scoped_session
 import urllib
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/observation_plan.py
```diff
@@ -575,6 +579,54 @@ def get(self, observation_plan_request_id=None):
...
+                type: string
+              description: |
+                String to match status of request against
+            - in: query
+              name: numPerPage
+              nullable: true
+              schema:
+                type: integer
+              description: |
+                Number of observation plan requests to return per paginated request. Defaults to 100. Can be no larger than {MAX_FOLLOWUP_REQUESTS}.
```
guynir42, line 623: I'm not sure this is an f-string. 
I'm also surprised you did not notice that, since modern IDEs will highlight a value in an f-string and not highlight curved brackets in a regular string...

skyportal/handlers/api/observation_plan.py
```diff
@@ -625,16 +700,64 @@ def get(self, observation_plan_request_id=None):
...
+                    .limit(n_per_page)
+                    .offset((page_number - 1) * n_per_page)
+                )
+            observation_plan_requests = (
+                session.scalars(observation_plan_requests).unique().all()
             )
-            return self.success(data=observation_plan_requests)
+
+            info = {}
+            info["observation_plan_requests"] = [
```
guynir42, line 756: Maybe change the key from `observation_plan_requests` to `request` just to make it easier for the users?

guynir42 (approved):
Minor comments.

skyportal/handlers/api/observation_plan.py
mcoughlin: The salt is strong with this one.

skyportal/handlers/api/observation_plan.py
mcoughlin: great idea.

PR #4146, opened 2023-04-22
InstrumentFieldTile: no index on instrument id

Theodlz:
Monitoring the google cloud console query planing page, we can see that for the same query, good query planning doesn't use that index, while bad query planning does. The difference in speed to return is over x150 times, and impacts negatively all MMA related features.

This PR is an attempt to avoid such horrible query planning by removing this index that it seemed to be using. Locally with over 50 localizations and 5 instruments with fields, this index is never used and removing it did not modify the good query planning.


mcoughlin (approved):
LGTM

PR #4147, opened 2023-04-24
Summary Search API

profjsb:
Facilitate the search for sources using natural language. This PR makes use of the summary embeddings, leveraging `langchain` to identify the top k sources that match a query.  From the CLI:

```bash
curl -s -X GET -H 'Authorization: token <user_token>' http://...:5000/api/summary_query \
     -d '{"q":"What sources are associated with NGC galaxies?", "k": 1, "z_min": 0.004}'
```
returns something like:
```json
{
  "status": "success",
  "data": {
    "query_results": [
      {
        "id": "ZTF23.....",
        "score": 0.830897391,
        "metadata": {
          "class": [
            "Type II"
          ],
          "redshift": 0.002,
          "summary": "Based on the information provided, it seems that a Type II supernova (SN II) has been observed in NGC XXX with a redshift of 0.005. The supernova peaked fainter than -15 and has an absolute magnitude of -14.5. There are also HST data and STIS long slit spectra covering the SN position,...."
        }
      }
    ]
  },
  "version": "..."
```


services/openai_analysis_service/app.py
```diff
@@ -273,11 +273,14 @@ def run_openai_summarization(data_dict):
         )
         pinecone_index = pinecone.Index(summarize_embedding_index)
         metadata = {}
-        if z is not None:
-            metadata["redshift"] = z
+        metadata["redshift"] = z if z is not None else -1
```
mcoughlin, line 276: Is this strictly necessary? Should it only be used in certain circumstances?

skyportal/handlers/api/query.py
```diff
@@ -0,0 +1,264 @@
+import os
```
mcoughlin, line 1: Maybe a more descriptive name for the .py file?

skyportal/handlers/api/query.py
```diff
@@ -0,0 +1,264 @@
...
+
+setattr(Pinecone, 'search_sources', search_sources)
+
+# Preamble: get the embeddings and summary parameters ready
+# for now, we only support pinecone embeddings
+summarize_embedding_config = cfg[
+    'analysis_services.openai_analysis_service.embeddings_store.summary'
+]
+USE_PINECONE = False
+if (
```
mcoughlin, line 73: Maybe we log if something is missing to help admins?

mcoughlin (approved):
Very cool. A few small recommendations.

services/openai_analysis_service/app.py
profjsb: I always want to have redshift in the metadata but might be able to get away without it.

mcoughlin: I guess my worry is whether there will be confusion about a negative default redshift.

skyportal/handlers/api/query.py
profjsb: Done.

PR #4148, opened 2023-04-24
GCN Event ingestion: Filter by tags

Theodlz:
In this PR, we make it possible to filter which events to ingest based on the tags that can be extracted from their notice.
That way can filter out LVC test events that are pushed to the same alert streams as the real events. Events we should start to receive on the 26th of April (beginning of the O4 engineering run).


mcoughlin (approved):
Nice.

PR #4151, opened 2023-04-26
TNS Submission bugfix

Theodlz:
In an f-string, one needs to double the curly brackets if they are part of the string and not just to add variables in the string. This seems to be missing in the handler to submit sources to TNS.


mcoughlin (approved):
LGTM

PR #4154, opened 2023-04-28
Observations in Localization query: use caching

Theodlz:
In this PR, we add 2 features:
- Use cached instrument and localization overlap if available when fetching the observations of an instrument in a localization. In a following PR, we can imagine caching it here as well, but that is not a priority as common values of cumprob like 0.90 and 0.95 will be cached by other features already.
- Specify which instrument to get the observations for when using the GCN Summary feature (optional, all if not specified)


skyportal/handlers/api/gcn.py
```diff
@@ -1918,8 +1918,9 @@ def add_gcn_summary(
     show_observations,
     no_text,
     photometry_in_window,
+    stats_method,
```
mcoughlin, line 1921: Maybe these should have defaults?

skyportal/handlers/api/observation.py
```diff
@@ -57,6 +58,14 @@
 
 Session = scoped_session(sessionmaker())
 
+cache_dir = "cache/localization_instrument_queries"
+cache = Cache(
+    cache_dir=cache_dir,
+    max_items=cfg.get("misc.max_items_in_localization_instrument_query_cache", 100),
+    max_age=cfg.get("misc.minutes_to_keep_localization_instrument_query_cache", 24 * 60)
```
mcoughlin, line 65: Just maybe comment that this is 1 day?

static/js/components/GcnSummary.jsx
```diff
@@ -174,6 +177,36 @@ const GcnSummary = ({ dateobs }) => {
     label: `${user.first_name} ${user.last_name}`,
   }));
 
+  let sortedInstrumentList = [...instrumentList];
+  sortedInstrumentList.sort((i1, i2) => {
+    // if the instrument is ZTF, put it first
+    if (i1.name === "ZTF") {
```
mcoughlin, line 183: Does ZTF need hard-coding here?

mcoughlin (approved):
Small suggestions.

static/js/components/GcnSummary.jsx
Theodlz: Will delete

skyportal/handlers/api/observation.py
Theodlz: ok!

skyportal/handlers/api/gcn.py
Theodlz: Agreed, will add

mcoughlin (approved):
LGTM

PR #4179, opened 2023-05-01
Notification queue: Prevent error

Theodlz:
There is no try catch in the queue service. Therefore, if an error occurs no notifications will be processed until we restart the service.


mcoughlin (approved):
LGTM

PR #4180, opened 2023-05-02
Annotations info caching

Theodlz:
In this PR, we cache the result of the annotations info caching.

The annotation page is heavily used, and this query is expected to pretty much always return the same output. Especially now that it has a tendency to time out, it makes sense to cache it given that the results will almost always be the same. We cache it for 6 hours by default, as the only time the output might change is overnight when a new Kowalski filter has been added.

Also, we modify the frontend to remove repetitive calls to that endpoint.

In this PR, we also add a `dict_to_bytes` method to the cache utils, which might come in handy when one needs to cache something other than a list/array.

### DIFF
--- a/skyportal/handlers/api/internal/annotations_info.py
+++ b/skyportal/handlers/api/internal/annotations_info.py
@@ -1,8 +1,26 @@
 from collections import defaultdict
+
+import numpy as np
 from sqlalchemy import func, literal
+
 from baselayer.app.access import auth_or_token
-from ...base import BaseHandler
+from baselayer.app.env import load_env
+from baselayer.log import make_log
+
 from ....models import Annotation
+from ....utils.cache import Cache, dict_to_bytes
+from ...base import BaseHandler
+
+_, cfg = load_env()
+
+cache_dir = "cache/annotations_info"
+cache = Cache(
+    cache_dir=cache_dir,
+    max_age=cfg.get("misc.minutes_to_keep_annotations_info_query_cache", 360)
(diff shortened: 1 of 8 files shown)


skyportal/handlers/api/internal/annotations_info.py
```diff
@@ -32,29 +50,46 @@ def get(self):
...
-        with self.Session() as session:
-            # Objs are read-public, so no need to check that annotations belong to an unreadable obj
-            # Instead, just check for annotation group membership
-            results = session.execute(
-                Annotation.select(session.user_or_token, columns=[Annotation.origin])
-                .add_columns(
-                    annotations.c.key,
-                    func.jsonb_typeof(annotations.c.value).label("type"),
+
+        try:
```
mcoughlin, line 54: Given the number of times we are doing this construction, how hard would it be to turn this into a function? not necessarily for this PR, but maybe worth thinking about.

mcoughlin (approved):
LGTM. Just one comment on thinking about how to maybe refactor this construction in the future if possible.

skyportal/handlers/api/internal/annotations_info.py
Theodlz: To load things? Yeah I thought about the exact same thing while writing this 😅. Given that it will be helpful to get this in before deploy, I suggest we merge it as is, and I'll be happy to open a separate PR this week to add load methods to the cache utils.

mcoughlin:
@Theodlz Deal.

PR #4181, opened 2023-05-02
Delete object photometry

mcoughlin:
This PR makes it simpler to delete an object's photometry.

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -1381,7 +1381,7 @@ class ObjPhotometryHandler(BaseHandler):
 
             return self.success(data=data)
 
-    @permissions(["Upload data"])
+    @permissions(["Delete bulk photometry"])
     def delete(self, obj_id):
         """
         ---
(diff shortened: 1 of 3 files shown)


profjsb (commented):
This action pretty major and subject to abuse. Perhaps we need special permission for bulk delete? Could also allow this delete if the user is the owner of the data. 
Most liberally we should only allow data from a given telescope or filter to be deleted per call. If someone deletes data that others uploaded it could be bad. 

Also, we should rerun phot stats after this since some of the non-deleted data may persist.

mcoughlin:
@profjsb I agree on a "Bulk Delete" permission. This automatically already ensures users have the ability to delete the photometry by the permissions checker (same as deleting individual photometry). But I will add this extra hurdle.

Theodlz:
@mcoughlin should this be by instrument maybe? I'm trying to think of ways to avoid people from deleting photometry that they didn't want to delete/aren't supposed to. I guess my point is what happens if someone deletes photometry and needs to cancel their action if let's say they don't have a copy for an instrument they didn't intend to delete stuff from. Of course, as developers, we can't prevent everyone's mistakes. But it seems reasonable to constrain the bulk delete in some ways.

mcoughlin:
@Theodlz Well the permissions are still at the photometry point level, so they can still only delete what they could have with individual points, just more conveniently.

Theodlz:
> @Theodlz Well the permissions are still at the photometry point level, so they can still only delete what they could have with individual points, just more conveniently.

Ok that makes sense then.

Theodlz (approved):
LGTM

PR #4186, opened 2023-05-03
GCN Notification queue refactor: use less sessions

Theodlz:
In this PR, we try to open/close less sessions in the notification queue to avoid various errors experienced in production.

### DIFF
--- a/skyportal/models/telescope.py
+++ b/skyportal/models/telescope.py
@@ -312,6 +312,6 @@ class Telescope(Base):
 
         return {
             "is_night_astronomical": is_night_astronomical,
-            "next_twilight_morning_astronomical": morning,
-            "next_twilight_evening_astronomical": evening,
+            "morning": morning,
+            "evening": evening,
         }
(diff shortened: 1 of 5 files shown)


skyportal/models/telescope.py
```diff
@@ -312,6 +312,6 @@ def current_time(self):
 
         return {
             "is_night_astronomical": is_night_astronomical,
-            "next_twilight_morning_astronomical": morning,
```
mcoughlin, line 315: Can you confirm the below is the only place this naming is used?

mcoughlin (approved):
Just confirm that the name change in the telescope morning / evening does not come up anywhere else. But I hope this resolves the session stuff!

skyportal/models/telescope.py
Theodlz: Thanks for the review Michael. I added that method yesterday and its only used in the telescope handler, map and obsplan queue  which I corrected as well if necessary. So that should be fine.

PR #4187, opened 2023-05-04
Show Similar sources

profjsb:
Use the AI summaries to show similar sources on the Source page. Updates the API so that a search can also be done directly on objID, not just via natural language.
Kapture 2023-05-04 at 14 49 37


skyportal/handlers/api/summary_query.py
```diff
@@ -122,10 +122,17 @@ def post(self):
           name: q
```
mcoughlin, line 122: maybe we can name this something better than q if we are going to use it as a variable we reference?

skyportal/handlers/api/summary_query.py
```diff
@@ -213,8 +220,13 @@ def post(self):
 
         data = self.get_json()
         query = data.get('q')
```
mcoughlin, line 222: So I guess q->query?

skyportal/tests/api/test_summary_query.py
```diff
@@ -7,7 +7,7 @@ def test_bad_queries(view_only_token):
     query_data = {}
```
mcoughlin, line 7: What happened to test_query.py?

static/js/components/SourceDesktop.jsx
```diff
@@ -431,6 +433,13 @@ const SourceDesktop = ({ source }) => {
               </div>
             </div>
           )}
+          {source.summary_history?.length > 0 ? (
```
profjsb, line 436: There's a (good) chance that there's an AI summary in the DB for this, so try to show this component

mcoughlin (changes requested):
Small comment on variable name. Also I assume you did not mean to delete that test?

profjsb:
> Small comment on variable name. Also I assume you did not mean to delete that test?

I did mean to delete that test -- it was a duplicate of the current test.

skyportal/tests/api/test_summary_query.py
profjsb: it's the same as another test file.

skyportal/handlers/api/summary_query.py
profjsb: I wanted to avoid calling it `query` in the openapi spec since there's another meaning of `query`.
```yaml
 - in: query
          name: query
          schema:
              type: string
```
feel awk.

skyportal/handlers/api/summary_query.py
profjsb: yes, see above for rationale.

mcoughlin (approved):
Thanks @profjsb. Makes sense.

Theodlz (approved):
Super cool! Looking forward to seeing people use all of these features soon, really exciting.
LGTM

PR #4200, opened 2023-05-10
Standards in observation plan

mcoughlin:
This PR adds ability to retrieve standards in observation plan.

### DIFF
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -1,4 +1,5 @@
 import arrow
+import ast
 from datetime import datetime, timedelta
 import healpy as hp
 import jsonschema
@@ -15,6 +16,7 @@ import sqlalchemy as sa
 from sqlalchemy import func
 from sqlalchemy.orm import joinedload
 import time
+import uuid
 
 from astropy import units as u
 from astropy.coordinates import SkyCoord
@@ -59,6 +61,7 @@ from ...models import (
     cosmo,
 )
 from ...models.schema import AssignmentSchema, FollowupRequestPost
+from ...utils.offset import get_formatted_standards_list
 
 log = make_log('api/followup_request')
 
@@ -943,6 +946,7 @@ def observation_schedule(
     instrument,
     observation_start=Time.now(),
     observation_end=Time.now() + TimeDelta(12 * u.hour),
+    standards=pd.DataFrame(),
     output_format='csv',
     figsize=(10, 8),
 ):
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/followup_request.py
```diff
@@ -954,10 +958,12 @@ def observation_schedule(
...
     instrument : skyportal.models.instrument.Instrument
         The instrument that the request is made based on
-    observation_start: astropy.time.Time
+    observation_start : astropy.time.Time
         Start time for the observations
-    observation_end: astropy.time.Time
+    observation_end : astropy.time.Time
         End time for the observations
+    standards : pandas.DataFrame
+        Standard stars for inclusion in the observation plan.
```
guynir42, line 966: What needs to be the columns of this table?

skyportal/handlers/api/followup_request.py
```diff
@@ -1276,6 +1310,38 @@ async def get(self, instrument_id):
           description: |
             Arrow-parseable date string (e.g. 2020-01-01). If provided, end time
             of observation window, otherwise 12 hours from now.
+        - in: query
+          name: includeStandards
+          nullable: true
+          required: false
```
guynir42, line 1316: Do we need nullable and required? I thought they were just opposite of each other...?

skyportal/handlers/api/followup_request.py
```diff
@@ -1365,17 +1411,64 @@ async def get(self, instrument_id):
...
-            followup_requests = session.scalars(followup_requests).unique().all()
-            if len(followup_requests) == 0:
+                # get owned assignments
+                followup_requests = FollowupRequest.select(self.current_user).join(
+                    allocation_subquery,
+                    FollowupRequest.allocation_id == allocation_subquery.c.id,
+                )
+
+                if start_date:
+                    start_date = str(arrow.get(start_date.strip()).datetime)
```
guynir42, line 1427: Not sure if you need to cast to string, I think `created_at` is a datetime object.

skyportal/utils/offset.py
```diff
@@ -728,22 +734,34 @@ def get_formatted_standards_list(
...
         ]
 
+    if standard_type == "ESO":
+        mag = []
+        for _, row in df.iterrows():
+            commentSplit = row["comment"].split(" ")
+            mag.append(float(commentSplit[1].replace("V=", "")))
+        df["mag"] = mag
+
+        df = df[(df["mag"] <= magnitude_range[0]) & (df["mag"] >= magnitude_range[1])]
```
guynir42, line 744: Is magnitude range always going to be [faint, bright] or is it possible that the order would be [low, high]?

guynir42 (approved):
I have mostly minor comments.

skyportal/handlers/api/followup_request.py
mcoughlin: good idea.

skyportal/handlers/api/followup_request.py
mcoughlin: good point.

skyportal/handlers/api/followup_request.py
mcoughlin: ok.

skyportal/utils/offset.py
mcoughlin: I am adding some checking to be sure they always go faint, bright

guynir42: I would prefer that you just order them as you want, and let the user input them either way.

mcoughlin: Ok I fixed that.

PR #4201, opened 2023-05-11
Galaxy in Localization Query: Use specific partition

Theodlz:
We've transitioned to a partitioned table for the LocalizationTiles. We added code to use the correct partition in other handlers, but not in this one yet.

Also, we add a few frontend lines to prevent resetting the GcnSelectionForm values upon submission and re-rendering.


mcoughlin (approved):
LGTM

Theodlz:
Failures are unrelated. Merging.

PR #4202, opened 2023-05-11
Source Page: optimizations

Theodlz:
In this PR we:
- Match python's and JS' bokeh versions.
- Avoid repetitive API calls
- Get rid of the red errors in the console while using the source page

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -1,68 +1,64 @@
+import collections
 import copy
 import itertools
-import math
 import json
-import collections
+import math
+import os
 
+import bokeh.embed as bokeh_embed
 import numpy as np
 import pandas as pd
-from sqlalchemy.orm import joinedload
-
-from bokeh.layouts import row, column
+from astropy.time import Time
+from bokeh.events import DocumentEvent
+from bokeh.layouts import column, row
 from bokeh.models import (
-    CustomJS,
-    HoverTool,
-    Range1d,
-    Slider,
     Button,
-    LinearAxis,
-    RadioGroup,
     CategoricalColorMapper,
+    CustomJS,
+    Dropdown,
+    HoverTool,
     Legend,
     LegendItem,
-    Dropdown,
+    LinearAxis,
+    RadioGroup,
+    Range1d,
+    Slider,
     Spinner,
     TabPanel,
     Tabs,
 )
 from bokeh.models.widgets import (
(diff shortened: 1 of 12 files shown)


Theodlz:
@profjsb as I suspected the source page fetches some data multiple times. On initial rendering, looks like it fetched the photometry plot multiple times

profjsb:
Looks good. Do you have a sense of the absolute number of api calls for both before this PR and after?

Theodlz:
> Looks good. Do you have a sense of the absolute number of api calls for both before this PR and after?

So that depends if the user goes to a source page directly (by opening in a new tab) or from any other page, as it will need to hydrate the rest of the data from`actions.js`. It might be that because of how things are rendered and prefetched (at least before this PR0), the nb of API calls can vary. Without the hydration, I believe it's around 16 or so. After this PR it should always be 12.

Theodlz:
Test failures unrelated. Merging.

profjsb:
Yeah good catch. Worth checking our caching mechanism. I noticed this only recently (<1 month) but thought it was from Chrome's new RAM saver mode. 

As you know, 12 is still more than our nominal number of DB connections per app. Perhaps we need to think about a sequential loading on the source page.

Theodlz:
> Yeah good catch. Worth checking our caching mechanism. I noticed this only recently (<1 month) but thought it was from Chrome's new RAM saver mode.
> 
> As you know, 12 is still more than our nominal number of DB connections per app. Perhaps we need to think about a sequential loading on the source page.

So I did just compare chrome and firefox. Firefox behaves as expected, so I'm wondering if it's a chrome/chromium issue rather than a SkyPortal one. However putting some efforts in optimizing the webpack (if there is room for improvement) could be beneficial.

The source page loading is already somewhat sequential. For instance, the photometry and spectra plot wait for other things to be fetched before they run. I'll take a look but that might already ensure that we get less than 10 calls at once. But even 12 should be fine right? Like taking the queue pool size + overflow into account we should have more than 10 connections per app available.

profjsb:
Well we're right on the edge and this obviously means that a single app can host only a single user frontend session. So I think we're warranted in bumping connections.

Theodlz:
@profjsb tagging you here just to make sure you know that: though this fixed it with the demo data, it seems to still happen in production, so I'm trying to figure out why.

PR #4204, opened 2023-05-11
Manage data is really share data

mcoughlin:
This PR changes the name of "ManageData" to "ShareData" since Manage implies more than what it really is.


guynir42 (approved):
I am happy whenever we name thing what it actually is.

PR #4207, opened 2023-05-13
cosmetic improvements

mcoughlin:
This PR makes a few cosmetic improvements.

### DIFF
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -974,9 +974,9 @@ def mark_detections(plot, detection_dates, ymin, ymax):
     first_r = plot.line(
         x=np.full(5000, first),
         y=y,
-        line_alpha=0.1,
+        line_alpha=0.175,
         line_color=first_color,
-        line_width=2,
+        line_width=3,
     )
     plot.add_tools(
         HoverTool(
--- a/skyportal/plot.py
+++ b/skyportal/plot.py
@@ -974,7 +974,7 @@ def mark_detections(plot, detection_dates, ymin, ymax):
     first_r = plot.line(
         x=np.full(5000, first),
         y=y,
-        line_alpha=0.5,
+        line_alpha=0.1,
         line_color=first_color,
         line_width=2,
     )
@@ -987,7 +987,7 @@ def mark_detections(plot, detection_dates, ymin, ymax):
     last_r = plot.line(
         x=np.full(5000, last),
         y=y,
-        line_alpha=0.5,
+        line_alpha=0.1,
         line_color=last_color,
         line_width=2,
     )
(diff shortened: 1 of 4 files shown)


static/js/components/AnnotationsTable.jsx
```diff
@@ -224,15 +271,51 @@ const AnnotationsTable = ({ annotations, spectrumAnnotations = [] }) => {
     jumpToPage: false,
     pagination: true,
     tableBodyMaxHeight: "20rem",
+    customToolbar: () => (
+      <>
```
Theodlz, line 275: I might be wrong but not sure these `<> </>` are needed here right?

static/js/components/AnnotationsTable.jsx
```diff
@@ -75,6 +83,40 @@ const getMuiTheme = (theme) =>
     },
   });
 
+const dialogTitleStyles = (theme) => ({
+  root: {
```
Theodlz, line 87: These isn't many annotations in the demo data to see how it renders when there is a lot of them. Did you try it by any chance while adding this styling? I am wondering if it doesn't overflow or anything, as it doesn't seem to be constrained in width and height.

static/js/components/CommentList.jsx
```diff
@@ -306,6 +309,10 @@ const CommentList = ({
 
   comments = comments || [];
 
+  if (!includeBots) {
```
Theodlz, line 312: Awesome! Thanks for adding that it's super nice.

Theodlz (approved):
Few minor comments, but thanks a lot for taking care of these so quickly.

skyportal/plot.py
```diff
@@ -974,7 +974,7 @@ def mark_detections(plot, detection_dates, ymin, ymax):
     first_r = plot.line(
         x=np.full(5000, first),
         y=y,
-        line_alpha=0.5,
+        line_alpha=0.1,
```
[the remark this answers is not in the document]
mcoughlin: added.

static/js/components/AnnotationsTable.jsx
mcoughlin: good.

static/js/components/CommentList.jsx
mcoughlin: great.

static/js/components/AnnotationsTable.jsx
mcoughlin: I just tested it with what is in there. We don't have all that many columns, and the height seems OK. I suggest we move forward like this for now and then see how it works in production.

PR #4208, opened 2023-05-13
DECam u-band

mcoughlin:
This PR pins sncosmo github (which has DECam u-band).


guynir42 (commented):
I understand that you are now able to use sncosmo (or any package) based on the git branch, but why do you also need to add the bandpass?

mcoughlin:
@guynir42 The DECam GW follow-up program has plans to use u-band.

guynir42 (approved):
I'm not sure what the difference is but let's talk about it later.

PR #4209, opened 2023-05-13
Diagnostic imaging plot

mcoughlin:
This PR adds a diagnostic plot when doing reduction.

### DIFF
--- a/skyportal/handlers/api/internal/image_analysis.py
+++ b/skyportal/handlers/api/internal/image_analysis.py
@@ -5,9 +5,9 @@ import os
 import arrow
 from astropy.io import fits
 from astropy.wcs import WCS
-from astropy.table import vstack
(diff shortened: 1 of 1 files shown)


skyportal/handlers/api/internal/image_analysis.py
```diff
@@ -197,15 +214,24 @@ def reduce_image(
...
 
             time = utils.get_obs_time(header, verbose=False)
             filt = header.get('FILTER')
 
             wcs = WCS(header)
 
+            # Fill value for image
+            v, c = np.unique(image, return_counts=True)
+            fill = v[c == np.max(c)][0]
+            mask = image == fill
```
guynir42, line 229: Why are we masking the most common value?

guynir42 (approved):
Looks very cool. I have only one question about the image mask.

PR #4210, opened 2023-05-14
User-defined wavelengths

mcoughlin:
This PR adds user-defined wavelengths as options for spectroscopy plot.

### DIFF
--- a/skyportal/tests/frontend/test_upload_spectroscopy.py
+++ b/skyportal/tests/frontend/test_upload_spectroscopy.py
@@ -51,7 +51,7 @@ def test_upload_spectroscopy(
     submit_button_xpath = '//button[contains(.,"Upload Spectrum")]'
     driver.click_xpath(submit_button_xpath, scroll_parent=True, timeout=30)
 
-    driver.wait_for_xpath('//*[contains(.,"successful")]')
+    driver.wait_for_xpath('//*[contains(.,"successful")]', timeout=10)
 
     # Go to "Share data" page to look for the spectrum, since we can't easily
     #  look into the Bokeh <canvas> tag on the Source page.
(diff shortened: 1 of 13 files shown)


skyportal/tests/frontend/test_upload_spectroscopy.py
```diff
@@ -51,7 +51,7 @@ def test_upload_spectroscopy(
     submit_button_xpath = '//button[contains(.,"Upload Spectrum")]'
     driver.click_xpath(submit_button_xpath, scroll_parent=True, timeout=30)
 
-    driver.wait_for_xpath('//*[contains(.,"successful")]')
+    driver.wait_for_xpath('//*[contains(.,"successful")]', timeout=10)
```
Theodlz, line 54: Nice! That should help the test be less flaky

static/js/components/SpectroscopyButtonsForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+          <div>
+            <TextField
+              label="Wavelengths"
+              {...register("spectroscopyButtonWavelengths", {
+                required: true,
+              })}
+              name="spectroscopyButtonWavelengths"
+              id="spectroscopyButtonWavelengthInput"
+            />
+            <TextField
```
Theodlz, line 80: So I would add a little tooltip/help icon to redirect users to the bokeh page that has the list of available colors. As is, hard for people to figure out what to type. If that color list is available in bokeh js somehow, might even want to replace by a dropdown that contains all unused colors?

Theodlz: Here is the URL btw: https://docs.bokeh.org/en/latest/docs/reference/colors.html#bokeh-colors-groups

Theodlz: If free form, we definitely want some sort of verification upon profile update that the colors used are valid.

Theodlz (approved):
Few comments about the user profile component. Otherwise looks great! Worked like a charm for me.

PR #4214, opened 2023-05-15
Skymap distance_error

mcoughlin:
This PR adds distance_error upon skymap ingestion.


Theodlz (approved):
LGTM. I thought we already grabbed the pipeline as we had MDC as a tag, but maybe that was a seperate bit of code. Merge once green or gcn tests have passed.

PR #4217, opened 2023-05-15
NED galaxy columns

mcoughlin:
This PR adds a few more columns to the Galaxy table (present in NED).


Theodlz (approved):
LGTM. I'm wondering if we should have designed the galaxy table to be like "catalog, name, altnames, data" with `data` as a JSONB to store all kinds of things. Otherwise this looks good to me.

mcoughlin:
@Theodlz that could work, I am not sure if jsonb filtering of that kind is slower though.

Theodlz:
> @Theodlz that could work, I am not sure if jsonb filtering of that kind is slower though.

Oh yeah it's definitely slower. I wasn't sure if we filtered on anything other than the redshift/distmpc (these can be proper columns). Something to consider in the feature, but super low priority for now.

Issue #4219, opened 2023-05-16
First/last detection photometry is hard to see


robertdstein:
In Fritz, the lightcurve plots have these vertical bars to indicate first and last detection. They make it nearly impossible to actually see the first and last detections at all, and especially to make out the error bars. It would be helpful if these lines could just be removed entirely (I'm not aware of anyone using them) or made very substantially less visible (a very faint/transparent colour).

Here's an example:

<img width="703" alt="Screenshot 2023-05-16 at 1 54 53 PM" src="https://github.com/skyportal/skyportal/assets/14312838/85036af2-2d09-4efb-82dd-358b1af5bca4">

PR #4220, opened 2023-05-16
Reduce opacity in vertical plot lines

mcoughlin:
This PR reduces alpha for vertical plot lines. Closes https://github.com/skyportal/skyportal/issues/4219.


Theodlz (approved):
Users probably all got 20/10 on their last eye exams. LGTM

Issue #4226, opened 2023-05-16
Plot gcn event times in sources


robertdstein:
Relaying request from MMA call:

Right now we have sources which are associated to gcn events. The GCN events have times. It would be great if we could plot these event times in the source lightcurves, perhaps with a vertical dashed line.

Theodlz:
@robertdstein just to be sure. These 2 concepts (sources you mark as associated to an event from the source list in the GCN Event page) should be linked? Or do you just want that on any source page, you can pick any gcnevent to display a line at its trigger time on the photometry plot?

I am asking because if its the first, that means you already know that source happened after an event and marked it as such, so not sure why that line on the plot is useful there.

robertdstein:
So previously when I was opening up fritz pages for some sources, it would say "associated to xxxx event", though clicking through some random ones I don't see it anymore.

Anyway, what we specifically want: 

If a source is associated to an event, i.e in a GCN event page source table (perhaps excluding rejected sources but definitely including both confirmed and undefined), then we want to be able to see easily when the event happened relative to the detections. We are not using it to see whether an event was detected beforehand (that's covered by filtering). Instead, it is very important to easily how much later after an event detections have happened. A detection at +5 hours is different than +50 hours, but on the scale of the plots it's not easy to work out which category it is.

Theodlz:
Ok gotcha. So yes this is what I imagined @mcoughlin. They wanna see events on the plot in general. Because the challenge here @robertdstein is that we query sources on the fly right. Unless a user sets or unsets a status there is no record that a source is in an event. If we did, the database would be flooded by any random sources saved to any group that happens to be inside of any event + that would add a lot of overhead everytime someone saves a source and said sources gets new photometry or when we get new events...

What about, a dropdown menu just like the on the candidate scanning page. There you can pick an event and then we display it on the plot. Otherwise, it can be something like a button where when you click it regenerates the plot + queries events that occurred up to N days before first detection and that contain the source, and then plots lines/markers for each.

Anyway I don't think this can be or should be an automatic thing, as spatial queries can be pretty expensive. For anyone doing MMA stuff waiting a little might be reasonable, but we don't want to end up impacting the user experience of everyone that uses a source page.

Theodlz:
@robertdstein the dropdown menu thingy is basically the existing feature of the spectrum plot to show a custom line. There you set a wavelength, but here you would pick an event and a line would show at the mjd when the event happens. 

That's just an idea, but this seems to be the most reasonable one.

PR #4233, opened 2023-05-16
GCN Event in confirmed sources plot

mcoughlin:
This PR notes GCN Event in confirmed sources plot. (closes https://github.com/skyportal/skyportal/issues/4226)

### DIFF
--- a/skyportal/handlers/api/sources_confirmed_in_gcn.py
+++ b/skyportal/handlers/api/sources_confirmed_in_gcn.py
@@ -199,11 +199,11 @@ class SourcesConfirmedInGCNHandler(BaseHandler):
 
                 if len(sources_id_list) == 0:
                     stmt = SourcesConfirmedInGCN.select(session.user_or_token).where(
-                        Localization.dateobs == dateobs
+                        SourcesConfirmedInGCN.dateobs == dateobs
                     )
                 else:
                     stmt = SourcesConfirmedInGCN.select(session.user_or_token).where(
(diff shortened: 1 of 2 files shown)


Theodlz:
Hi @mcoughlin. This works just fine. Just not sure people want to show a line for the GCN trigger on photometry plots only if the source has been highlighted/confirmed. I might have misunderstood the user request (i think it was Dan?), but I believe this was a thing they should be able to see on sources without having to manually query for the sources and highlight interesting ones. I mean if they queried it and got it, that already means it was detected within X days after the event, so I'm not sure I see why they need that on the plot then.

mcoughlin:
@Theodlz I mean it's always nice to see the GRB time relative to the first detection. So I disagree.

Theodlz:
Ok that makes sense. Just want to know if we expect people to actively confirm/highlight a source before being able to see that line/marker. Btw little suggestion but I would add a dotted line with a low opacity, as the current implementation isn't that easy to read. Probably can just reuse the code for the detections but use the `dash` (or dashed) method of bokeh.

mcoughlin:
@Theodlz Can you give the syntax or just push it?

Theodlz:
@mcoughlin talked with Robert and we agreed on what to do (for which sources to show, and what to show). So its for highlighted (old confirm) and ambiguous (have a row but not confirmed or rejected), and we show a dotted line + the G marker.

mcoughlin:
@Theodlz Ok.

Issue #4231, opened 2023-05-16
GCN event page - source tab


tahumada:
1. GCN source table does not allow for download, there is a button but doesn't work
2. (minor) when you use the filter and there is no match, you can't go back to the table and have to query again

Theodlz:
Hi @tahumada. Just to be clear, what you mean is that you are applying filters to the sources table displayed on the gcn page, other than just the localization and so on? Like you add constraints on other parameters?

mcoughlin:
@Theodlz can we close this one?

Theodlz:
@mcoughlin I will go over each issue after deploy today and closed if users confirm that this is addressed.

Issue #4230, opened 2023-05-16
Default sources on GCN event page should be increased


robertdstein:
Relaying from MMA call:

The default number of sources to display on the GCN event page (10) is too low, and it would make scanning easier if the default was 100.

(related to #4229 and #4228 )

Theodlz:
Sure! Any reason why people don't want to just... change page? I agree that 10 is not enough however, so will increase for sure.

robertdstein:
I am just relaying what was said on the MMA call here, and unfortunately cannot provide you with any more detailed justification on this one

Issue #4228, opened 2023-05-16
Rename "confirm" button for GCN events


robertdstein:
Relaying from MMA call:

Right now we have three icons for sources in the GCN event source table: confirm, reject or undefined. The confirm button does not seem super useful (we are unlikely to forget which source is a kilonova), so it would make sense to rename this "highlight" where it highlights promising sources. If those sources are subsequently ruled out, then we can recategorise.

Screenshot example:
<img width="1509" alt="Screenshot 2023-05-16 at 2 32 07 PM" src="https://github.com/skyportal/skyportal/assets/14312838/22bd0ee8-f9a5-414d-aa63-4ca8a7bd7e3a">

Theodlz:
Hi @robertdstein. Agreed! Will take care of that issue. In the backend the table is already called "sourcesconfirmedingcn" so that won't change, but I'll just change what's written on the frontend component then.

mcoughlin:
@Theodlz can we close this?

PR #4235, opened 2023-05-17
Sources in GCN Table: Various user requests

Theodlz:
In this PR we:
- Fix the download of sources from the source table on the gcn page [issue #4231]
- Stop hiding the table when there are no sources, so that additional filters can be reset without having to requery sources [issue #4231]
- Increase the default nb of sources in the table to 100 [issue #4230]
- Rename "confirmed" to "highlight" in the component to confirm/reject sources in gcn [issue #4228]


mcoughlin (approved):
LGTM

PR #4236, opened 2023-05-17
Minor Planet Center query

mcoughlin:
This PR enables Minor Planet Center queries.


skyportal/handlers/api/mpc.py
```diff
@@ -0,0 +1,197 @@
...
+            except (TypeError, arrow.ParserError):
+                return self.error(f'Cannot parse time input value "{date}".')
+
+        limiting_magnitude = data.get('limiting_magnitude', 24.0)
+        try:
+            limiting_magnitude = float(limiting_magnitude)
+        except Exception:
+            return self.error('Cannot read in limiting magnitude.')
+
+        search_radius = data.get('search_radius', 60.0)
```
guynir42, line 101: The default is one minute but if the user inputs 1 or 2 that would not be in minutes...

skyportal/handlers/api/mpc.py
```diff
@@ -0,0 +1,197 @@
...
+            ra = Angle(obj.ra, unit='degree')
+            ra_hms = ra.to_string(
+                unit='hourangle', sep=' ', precision=2, pad=True
+            ).split(" ")
+            dec = Angle(obj.dec, unit='degree')
+            dec_dms = dec.to_string(unit=u.deg, sep=' ', precision=1, pad=True).split(
+                " "
+            )
+
+            url = f"{mpcheck_url}?year={year}&month={month}&day={day}&which=pos&ra={ra_hms[0]}+{ra_hms[1]}+{ra_hms[2]}&decl={dec_dms[0]}+{dec_dms[1]}+{dec_dms[2]}&TextArea=&radius={search_radius}&limit={limiting_magnitude}&oc={obscode}&sort=d&mot=h&tmot=s&pdes=u&needed=f&ps=n&type=p"
```
guynir42, line 132: Can split this string out to a few short rows.

skyportal/handlers/api/mpc.py
```diff
@@ -0,0 +1,197 @@
...
+        if pre is None:
+            log(f'Error checking MPC for {obj_id}: {str(soup)}')
+        else:
+            if re.findall("The following objects,.*", response.text)[0]:
+                responseSplit = response.text.split("(1)")[-1].split(" ")
+                mpc_name = list(filter(None, responseSplit))[0]
+
+                log(f'{obj_id}: identified MPC name {mpc_name}')
+
+                obj.is_roid = True
```
guynir42, line 183: What does `roid` stand for? is that short for asteroid...?

skyportal/models/obj.py
```diff
@@ -266,6 +266,10 @@ class Obj(Base, conesearch_alchemy.Point):
         default=False,
         doc="Boolean indicating whether the object is a moving object.",
     )
+    mpc_name = sa.Column(
+        sa.String,
+        doc="Minor planet center name(s).",
```
guynir42, line 271: Is there a situation where multiple names exist? If so, are they just going to be saved as a comma separated list in a string?

guynir42 (commented):
I have some comments/questions, mostly small stuff.

skyportal/handlers/api/mpc.py
mcoughlin: good catch

skyportal/handlers/api/mpc.py
mcoughlin: Yeah... that was historical...

skyportal/models/obj.py
mcoughlin: Changing this to be singular

skyportal/handlers/api/mpc.py
mcoughlin: Good point.

guynir42 (approved):

PR #4237, opened 2023-05-17
TNS Obj Handler: Only submit for ZTF

Theodlz:
For now we haven't found a nice way to find the id of any instrument on TNS. Until we do, we limit the submission of ZTF data (which is anyway the most important use case) for which we know the TNS instrument id.


Theodlz:
@virajkaram tagging just to make sure you know this will be available after the next deploy. Worked just fine for me locally. Let's see how it goes in production.

mcoughlin (approved):
LGTM

PR #4241, opened 2023-05-18
Instrument configuration

mcoughlin:
This PR adds instrument configuration JSON column.


skyportal/facility_apis/observation_plan.py
```diff
@@ -451,6 +451,8 @@ def generate_plan(
                 request.payload.get("ra_slice_min", 0),
                 request.payload.get("ra_slice_max", 360),
             ],
+            # use only primary grid
+            'doUsePrimary': request.payload.get("use_primary", False),
```
Theodlz, line 455: I think that what we wanted is that we can:
1. First create a plan only with the primary, which means being able to turn off the secondary
2. Target only the secondary when we managed to cover the primary
3. Maybe do both.

Not sure what @tahumada wanted that now the primary is an option, but rather the opposite that the secondary is turned off by default.

Theodlz: Or are you telling me that this doUsePrimary means that you ONLY use the primary??? This is confusing hahaha

mcoughlin: Yes it just uses the primary. It is a quick way to avoid overlapping tiles to save time. This brings us to consistency with O3, the second requires gwemopt changes.

Theodlz: Just wondering if in this PR we can also add the doUseSecondary arg and make it so that on the frontend you can pass both to True (all grid), just primary, or just secondary

mcoughlin: @Theodlz yes added.

Theodlz: Do you want me to add the frontend for that too?

mcoughlin: https://github.com/skyportal/skyportal/pull/4241/files#diff-8c6d218804fafa52e37ebbea006afc0200ecff0d2c71d95d76df6176462496c2R992

It is one of the options on the front-end. Feel free to make changes.

Theodlz: Right my apologies I forgot about the rjsf schemas. Looks good.

static/js/components/CommentList.jsx
```diff
@@ -208,7 +208,10 @@ const CommentList = ({
     (state) => state.profile.preferences?.compactComments
   );
   const { currentShift } = useSelector((state) => state.shift);
-  const [includeBots, setIncludeBots] = useState(false);
+
+  const { showBotComments } = useSelector((state) => state.profile.preferences);
```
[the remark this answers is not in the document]
Theodlz: Might just be that it does get the user profile but after the rendering of the comments list. If so, an empty useEffect on the user profile added somewhere on the source page should do the trick by forcing it to rerender.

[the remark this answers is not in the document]
mcoughlin: @Theodlz can you add / try this quick?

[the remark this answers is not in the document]
Theodlz: Sure!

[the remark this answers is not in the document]
mcoughlin: Ok let's merge after tests then. Are we deploying today?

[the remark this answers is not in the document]
Theodlz: Yep! Let's deploy these 3 billion migrations lol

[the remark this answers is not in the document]
Theodlz: Will do around 2pm if all is green

Theodlz:
Failures seem to be unrelated. GraceDB is down, so all the tests involving a GW notice can't grab the skymap. I tested this PR locally and everything worked just fine.

LGTM

PR #4242, opened 2023-05-18
Field picking fix: use field_id instead of id

Theodlz:
It seems that the fields selection was using the ids in the DB and not the actual field ids. With these changes, I can successfully get gwemopt to target specific fields 👍 

TODO: In a following PR, fix the field selection from the map. Right now it works for the primary and not the secondary somehow. Probably that the fields plotted first (that end up underneath) can't be clicked. This is not very important and requires some D3 magic.


mcoughlin (approved):
LGTM!

Theodlz:
Failures are unrelated. GraceDB is struggling and the skymaps can't be downloaded. Merging.

PR #4246, opened 2023-05-19
TNS submission: Make sure last non-detection is before first detection

Theodlz:
Make sure last non-detection is before first detection


mcoughlin (approved):
Thanks for tracking that down!

Theodlz:
This allowed me to submit to TNS with a source which failed in production.

With this, we should ensure that the photometry available can conform to the TNS requirements (if not will throw an error).

virajkaram:
Sorry for the delay, LGTM

PR #4248, opened 2023-05-20
GCN Ingestion: Skip retractions if the event doesn't exist in the DB

Theodlz:
Skip retractions ingestion if the event doesn't exist in the DB


mcoughlin (approved):
LGTM

mcoughlin (approved):
LGTM

PR #4250, opened 2023-05-20
In-out for an individual source

mcoughlin:
This PR begins to add handler to check in-out on events for an individual source.

@Theodlz the syntax must be wrong because it doesn't find a tile, and we know the demo data /api/sources/ZTFinGCN/gcn_event is in the GRB. Can you think of a better way to do this / why it returns empty?


Theodlz:
Hi @mcoughlin. Looking at the code, no obvious issue here. I would suggest running first the query without the healpix.contain() to see if at least you are getting any localization tiles. If yes, then its that bit of code that is faulty. If that's the case, I could suggest creating a healpix range out of that single healpix index, and using the overlap method maybe?

mcoughlin:
@Theodlz yes I do get tiles out without the overlap call. I was wondering if maybe it needs to be specifically type cast as a BigInt or whatever, but maybe your range idea is enough.

Theodlz:
@mcoughlin what's weirder is that this exact same call happens in multiple handlers on objs and galaxies and works just fine... Next coding session I'll try debugging this if you don't find a solution until then.

mcoughlin:
@Theodlz I just made it look like the other queries and it works. I don't understand the difference.

Theodlz (approved):
LGTM. Worked perfectly fine locally, we'll see how it works in production and I'll be happy to help optimize if necessary.

Theodlz:
We're starting to have a lot of those querying options on the frontend, moving the page content towards the bottom more and more. I'll take care of fixing that when I work on a new version of the source page one of these days

PR #4251, opened 2023-05-20
GcnTag notification

mcoughlin:
This PR triggers notification check when GcnTag is added.


Theodlz:
Hi @mcoughlin. What's the idea behind this feature? We want people to be notified everytime a tag is added by a user?

mcoughlin:
@Theodlz we want to be notified if a user adds a tag to an event and that tag is in our notification criteria. I.e if ex post facto I add BNS when it originally was BBH in the GCN I want the notification to go out.

Theodlz (approved):
LGTM. I agree that this will be useful, thanks for adding that!

PR #4254, opened 2023-05-21
Photometry form instrument

mcoughlin:
This PR fixes weirdness with setting instrument in photometry form.


Theodlz (approved):
Thanks for fixing that. This broke with an rjsf or another dependency update I guess, as it used to work just fine. Noticed it didn't work in production the other day.

PR #4255, opened 2023-05-22
Significant tag

mcoughlin:
This PR gets "Significant" tag from the events.

### DIFF
--- a/static/js/components/GcnTags.jsx
+++ b/static/js/components/GcnTags.jsx
@@ -191,9 +191,10 @@ const GcnTags = ({ gcnEvent, show_title = false }) => {
             key={tag}
             // if there is a class for this tag, apply it, otherwise use the basic grey color
             style={{
-              backgroundColor: gcn_tags_classes
-                ? gcn_tags_classes[tag]
-                : "#999999",
+              backgroundColor:
+                gcn_tags_classes && tag in gcn_tags_classes
+                  ? gcn_tags_classes[tag]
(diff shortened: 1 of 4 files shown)


static/js/components/GcnTags.jsx
```diff
@@ -190,10 +172,14 @@ const GcnTags = ({ gcnEvent, show_title = false }) => {
...
             ) : (
               <Chip
-                className={styles[tag]}
                 size="small"
                 label={tag}
                 key={tag}
+                style={{
+                  backgroundColor: gcn_tags_classes
+                    ? gcn_tags_classes[tag]
+                    : "#999999",
```
Theodlz, line 181: What if gcn_tags_classes exists but the tag isn't in it? It looks like this will still get it from gcn_tags_classes and yield null, using the default MUI color rather than the "#999999" you've set. Same comment for GcnEvents.jsx

Theodlz (commented):
One comment to address before merging. It might work out anyway because MUI has a default color, but I think that this piece of code isn't doing what we want it to.

static/js/components/GcnTags.jsx
mcoughlin: @Theodlz Good point. I think I fixed it.

Theodlz: Yep looks like it works! Feel free to merge whenever you want.

Theodlz (approved):
LGTM

PR #4256, opened 2023-05-22
Gwemopt: incorrect parameter name

Theodlz:
Yesterday we had issues creating a plan that didn't keep alternating filters. It looks like there is a typo in the observation_plans.py file, which meant it wasn't using the block strategy.

### DIFF
--- a/skyportal/facility_apis/observation_plan.py
+++ b/skyportal/facility_apis/observation_plan.py
@@ -405,7 +405,12 @@ def generate_plan(
         params = {
             # gwemopt filter strategy
             # options: block (blocks of single filters), integrated (series of alternating filters)
-            'doAlternativeFilters': request.payload["filter_strategy"] == "block",
+            'doAlternatingFilters': True
+            if request.payload["filter_strategy"] == "block"
+            else False,
+            'doBlocks': True
+            if request.payload["filter_strategy"] == "block"
+            else False,
             # flag to indicate fields come from DB
             'doDatabase': True,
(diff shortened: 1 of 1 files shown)


Theodlz:
@mcoughlin let's wait till all the tests are done running. This might change the nb of observations in some tests, getting them to fail.

Theodlz:
@mcoughlin I added more stuff to this PR (visualization especially). Feel free to have a look or rereview if you think is necessary.

mcoughlin:
@Theodlz feel free to merge once the tests behave ;)

Theodlz:
> @Theodlz feel free to merge once the tests behave ;)

They're always so mean 🙄

Theodlz:
@mcoughlin to keep you in the gwemopt loop. Looks like some gwemopt methods called when using `doBlocks` look for the catalog_struct, which is not passed separately as the methods want it too. When not using doBlocks it works fine, as anyway it looks like the catalog struct is refactored and added to the tile_structs, but that is not used by the methods called in doBlocks. 

So for now, I force it to False when doing galaxy targeting until I get to fix it in gwemopt (and with all the changes roberts made, we probably will have to change few things here anyway so doing a quick release now wont help).

Also, it looks like changing the min_diff_filter time doesn't do anything. Something to look into later on.

Otherwise this PR is pretty much done. Maybe you can have a quick look at the generated plans through the new frontend to see if they don't alternate filters too much, and we can merge?

Theodlz:
Failures are unrelated. Merging.

PR #4264, opened 2023-05-23
Check for existing photometry when deleting instrument filters

mcoughlin:
This PR makes it possible to delete instrument filters with check for existing photometry.


Theodlz (approved):
Thanks for taking care of that. It was indeed pretty annoying to not be able to delete filters other than by accessing the DB.

PR #4266, opened 2023-05-23
For mixed, make sure trigger is false

mcoughlin:
This PR fixes the trigger for mixed.


Theodlz (approved):
I just created a branch locally to take care of it too 😉. LGTM

PR #4267, opened 2023-05-23
Include comments when requesting follow-up

mcoughlin:
This PR adds ability to include comments when requesting follow-up.

### DIFF
--- a/skyportal/facility_apis/generic.py
+++ b/skyportal/facility_apis/generic.py
@@ -69,7 +69,7 @@ def validate_request(request, filters):
         raise ValueError('priority must be within 1-5.')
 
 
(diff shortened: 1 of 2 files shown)


skyportal/facility_apis/generic.py
```diff
@@ -127,6 +129,13 @@ def create_target_text(request, request_type='email'):
         f"Username: {target['username']}"
     )
 
+    if comments:
```
Theodlz, line 132: maybe check that the obj has comments too, to not have `Comments:` empty when there is no comments, but just skip it or say there is no comments.

Theodlz (approved):
LGTM. One minor comment

PR #4270, opened 2023-05-23
Object host index

mcoughlin:
This PR adds an index for object host.


Theodlz (approved):
LGTM. Out of curiosity, is this because you plan on querying which objects are next to a specific galaxy? (more generally, which do we need on index here)

mcoughlin:
@Theodlz Actually this is what the internet tells me I need for efficient deleting :)

Theodlz:
> @Theodlz Actually this is what the internet tells me I need for efficient deleting :)

Stack overflow is always right, that's the number one rule 🐿

PR #4272, opened 2023-05-24
GCN event getter

mcoughlin:
This PR tries to refactor the GCN event getter to make it useful for the crossmatch page. Something about the state is funny.. I wonder if you can take a quick look @Theodlz?

### DIFF
--- a/static/js/components/CrossmatchGcnEvents.jsx
+++ b/static/js/components/CrossmatchGcnEvents.jsx
@@ -57,7 +57,7 @@ const Crossmatch = () => {
   };
 
   return (
-    <form className={classes.root}>
+    <div className={classes.root}>
       <div className={classes.gridItem}>
         <h6> 1: </h6>
         <FindGcnEvents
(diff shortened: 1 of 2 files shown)


Theodlz:
On it @mcoughlin

mcoughlin:
@Theodlz The second GCN doesn't load in for me at least.

Theodlz:
Ohh ok I see, yeah there were two bugs then and I fixed the other one 😅. Let me have another look.

Theodlz:
What about now? Works just fine for me on my end.

mcoughlin:
@Theodlz Magic. Can we review and get it in?

Theodlz (approved):
LGTM

Theodlz:
Oh @mcoughlin, we will need a second PR to add this getter to the skymap queue submission btw.

mcoughlin:
@Theodlz If you can do it in a follow-on PR, then great. Let's get this in so it works for deploy tomorrow.

PR #4273, opened 2023-05-24
Allow requester to be none when returning allocation

mcoughlin:
This PR checks if the follow-up requester is None.


Theodlz (approved):
LGTM

PR #4275, opened 2023-05-24
TNS queue for submission

mcoughlin:
This PR starts process of moving to a TNS queue system. First in hopefully a series of PRs that, for example, track TNS submission success.

### DIFF
--- a/skyportal/handlers/api/sources_confirmed_in_gcn.py
+++ b/skyportal/handlers/api/sources_confirmed_in_gcn.py
@@ -6,9 +6,10 @@ from tornado.ioloop import IOLoop
 from baselayer.app.access import auth_or_token, permissions
 from ..base import BaseHandler
 from .source import get_sources, MAX_SOURCES_PER_PAGE
-from .tns import tns_submission
+
 from ...models import GcnEvent, Localization, SourcesConfirmedInGCN, TNSRobot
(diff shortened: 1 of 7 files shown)


skyportal/handlers/api/sources_confirmed_in_gcn.py
```diff
@@ -684,14 +685,16 @@ async def post(self, dateobs):
...
+
                 IOLoop.current().run_in_executor(
                     None,
-                    lambda: tns_submission(
-                        [obj.obj_id for obj in sources_in_gcn],
-                        tnsrobot.id,
-                        self.associated_user_object.id,
-                        reporters=reporters,
-                    ),
+                    lambda: post_tns(request_body, timeout=30),
```
Theodlz, line 697: Given that we have a post_tns method, I would create the request body there and just pass it objs, tnsrobot_id, user_id, and reporters. Just to avoid that recurring bit of code.

Theodlz (approved):
LGTM. Just one minor comment, but looks very good overall.

skyportal/handlers/api/sources_confirmed_in_gcn.py
mcoughlin: Great idea.

PR #4277, opened 2023-05-25
Update baselayer (No social auth alchemy dependency)

Theodlz:
An attempt to avoid auth errors, and more generally make them easier to debug, while not relying on a package that isn't maintained anymore.


mcoughlin (approved):
LGTM!

PR #4286, opened 2023-05-28
Field_id of the instrument, not database id

Theodlz:


mcoughlin (approved):
LGTM

mcoughlin:
I think current behavior is fine. We can reopen if it turns out not to be.

Theodlz:
Discussed with @mcoughlin, reopening.

What I did is rename 'field_id' (the database one) to 'field_db_id', and add the actual field_id named 'field_id' so we have both, but people aren't confused. Tell me what you think.

PR #4289, opened 2023-05-28
Split API tests in 2 parts

Theodlz:
We've reached a point where the API tests take almost 2 hours to run. This makes it increasingly hard to iterate quickly, and clearly discourages us to tests new features thoroughly as it will increase the time needed even further.

This PR is an attempt to split said tests suite in 2 parts, just like the frontend tests. The first part contains the most "critical" elements for O4, which all interact with each other, in an attempt to replicate somewhat realistically the heavy load that MMA features can sometimes put on the app.


mcoughlin:
@Theodlz Nice!

Theodlz:
> @Theodlz Nice!

All green! Feel free to balance the 2 tests suite in terms of time if you want. One of them takes 50 minutes the other around 30. The latter is the most likely to have tests added to it however, so we can also keep it as is.

mcoughlin:
@Theodlz lets start with this

mcoughlin (approved):
Definitely a helpful development.

PR #4290, opened 2023-05-28
Make all GCN tables updateable

mcoughlin:
This PR makes it such that all GCN tables are updateable


Theodlz (approved):
Much cleaner, thanks! LGTM, merge once green.

PR #4292, opened 2023-05-29
Underflow from small skymap values

mcoughlin:
This PR prevents underflow from small skymap values.

### DIFF
--- a/skyportal/utils/gcn.py
+++ b/skyportal/utils/gcn.py
@@ -545,11 +545,9 @@ def from_url(url):
     filename = os.path.basename(urlparse(url).path)
 
     skymap = ligo.skymap.io.read_sky_map(url, moc=True)
-    minval = 1e-20
-    idx1 = np.where(skymap['PROBDENSITY'] < minval)[0]
-    idx2 = np.where(skymap['PROBDENSITY'] > 0)[0]
-    idx = np.intersect1d(idx1, idx2)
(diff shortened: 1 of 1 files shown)


Theodlz (commented):
2 questions before I approve:
- We shouldn't have negative values at all right? Why do we need this intersection here?
- Is 1e-20 small enough? Looking at the database most skymaps had a lot of values under 1e-20. Especially large ones. Could we maybe lower that number to something much smaller, like maybe 1e-200?

Theodlz:
> 2 questions before I approve:
> 
> * We shouldn't have negative values at all right? Why do we need this intersection here?
> * Is 1e-20 small enough? Looking at the database most skymaps had a lot of values under 1e-20. Especially large ones. Could we maybe lower that number to something much smaller, like maybe 1e-200? It's not very risky, and we avoid changing the skymap too much.

mcoughlin:
@Theodlz Good point, let's just make it 1e-200 and set them all to 0. Wanna take care of that?

Theodlz:
@mcoughlin sure! Let me try right now.

Theodlz:
Feeling like taking risks today. Tried 10^-300, which was enough to get the 2 buggy skymaps I had to work (i did it manually in postgres last time) just fine so it seems like a good number 😉. Let's see if the tests are still happy with that.

PR #4295, opened 2023-05-30
TNS archival submission

mcoughlin:
This PR enables TNS "archival" submission (those without upperlimits).

### DIFF
--- a/services/tns_queue/tns_queue.py
+++ b/services/tns_queue/tns_queue.py
@@ -47,7 +47,7 @@ def tns_submission(
     user_id,
     reporters="",
     archival=False,
-    archivalComment="",
+    archival_comment="",
     parent_session=None,
 ):
     """Submit objects to TNS.
@@ -61,8 +61,8 @@ def tns_submission(
         Reporters to appear on TNS submission.
     archival : boolean
         Reporting the source as an archival source (i.e. no upperlimit).
-    archivalComment : str
-        Comment on archival source. Required if archival is True:
+    archival_comment : str
+        Comment on archival source. Required if archival is True.
     parent_session : `sqlalchemy.orm.session.Session`
         Database session.
     """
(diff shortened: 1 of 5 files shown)


services/tns_queue/tns_queue.py
```diff
@@ -41,7 +41,15 @@
...
 
 
-def tns_submission(obj_ids, tnsrobot_id, user_id, reporters="", parent_session=None):
+def tns_submission(
+    obj_ids,
+    tnsrobot_id,
+    user_id,
+    reporters="",
+    archival=False,
+    archivalComment="",
```
Theodlz, line 50: let's use snake_case for the arguments, to keep it python friendly.

services/tns_queue/tns_queue.py
```diff
@@ -51,6 +59,10 @@ def tns_submission(obj_ids, tnsrobot_id, user_id, reporters="", parent_session=N
         SkyPortal ID of User posting to TNS
     reporters : str
         Reporters to appear on TNS submission.
+    archival : boolean
+        Reporting the source as an archival source (i.e. no upperlimit).
+    archivalComment : str
```
Theodlz, line 64: same thing here in the docstring

skyportal/handlers/api/tns.py
```diff
@@ -382,6 +382,8 @@ def post(self, obj_id):
             data = self.get_json()
             tnsrobotID = data.get('tnsrobotID')
             reporters = data.get('reporters', '')
+            archival = data.get('archival', False)
+            archivalComment = data.get('archivalComment', '')
```
Theodlz, line 386: here I would do `archival_comment = data.get('archivalComment', '')` instead, to have that snake_case

Theodlz (approved):
One minor comment, but otherwise LGTM

services/tns_queue/tns_queue.py
mcoughlin: yes

services/tns_queue/tns_queue.py
mcoughlin: good.

PR #4299, opened 2023-05-31
Use follow-up allocations for catalog queries

mcoughlin:
This PR uses the follow-up allocations for catalog queries (which makes more sense).


Theodlz (approved):
Makes a lot of sense. LGTM

PR #4327, opened 2023-06-02
Enable photometry form to ingest MJD

mcoughlin:
This PR enables the photometry form to ingest MJD.


Theodlz:
@mcoughlin I added few lines to not create altdata for rows that should have any. If that looks good to you, let me know and I'll merge.

Theodlz (approved):
LGTM

PR #4329, opened 2023-06-02
TNS photometry

mcoughlin:
This PR enables retrieval of TNS photometry.

### DIFF
--- a/skyportal/handlers/api/tns.py
+++ b/skyportal/handlers/api/tns.py
@@ -211,8 +211,7 @@ def tns_retrieval(
 
     user = session.scalar(sa.select(User).where(User.id == user_id))
 
-    # try:
-    if True:
+    try:
         obj = session.scalars(Obj.select(user).where(Obj.id == obj_id)).first()
         if obj is None:
             raise ValueError(f'No object available with ID {obj_id}')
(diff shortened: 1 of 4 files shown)


skyportal/utils/tns.py
```diff
@@ -50,7 +55,8 @@
     'ztfr': 111,
     'ztfi': 112,
 }
-INSTRUMENT_TNS_IDS = {v: k for k, v in TNS_INSTRUMENT_IDS.items()}
+
+FILTER_TNS_IDS = {v: k for k, v in TNS_FILTER_IDS.items()}
```
Theodlz, line 59: Maybe we should rename those: `SNCOSMO_TO_TNSFILTER`, and `TNSFILTER_TO_SNCOSMO`, or something similar to make it a little more clear. Or keep the current names but just add a `_TO_` in between.
Very minor comment however, feel free to ignore.

skyportal/handlers/api/tns.py
```diff
@@ -254,6 +263,25 @@ def tns_retrieval(obj_id, tnsrobot_id, user_id, include_spectra=False):
...
+                            data_out = {
+                                'obj_id': obj_id,
+                                'instrument_id': instrument_id,
+                                'group_ids': group_ids,
+                                **df.to_dict(orient='list'),
+                            }
+                            add_external_photometry(data_out, user)
+                        except Exception as e:
+                            log(
+                                f'Cannot read TNS photometry {str(photometry)}: {str(e)}'
```
Theodlz, line 282: Here I'd use flow to send a frontend messages. You can maybe store failed ingestions info in a list to concatenate into one error message you can send back to the frontend after iterating. That way we can display a message on the frontend (well of course, if this was triggered from the frontend). This is to avoid us from having to check the logs manually every time someone doesn't get the data they expected to ingest.

Theodlz (approved):
2 minor comments, but LGTM!

mcoughlin:
@Theodlz can you add these changes quick? :)

Theodlz:
> @Theodlz can you add these changes quick? :)

Made these changes. I would suggest trying the code before merging, but LGTM.

PR #4330, opened 2023-06-02
Profile page safety: Followup requests preference

Theodlz:
This PR adds a little more type safety to the profile page's followup requests preferences components.


mcoughlin (approved):
This is clearly better. LGTM.

PR #4340, opened 2023-06-06
Forced photometry Delete Bugfix

Theodlz:
Compared to "triggered" followup, there is no queue_name for forced phot, or anything to delete from the server in general it seems.

This PR skips deleting from server for FP.


mcoughlin (approved):
LGTM

PR #4341, opened 2023-06-06
InstrumentSummary start and end form

mcoughlin:
This PR adds form to Instrument Summary to select start and end data.


skyportal/plot.py
```diff
@@ -3092,11 +3093,18 @@ async def instrument_log_plot(
...
+        InstrumentLog.instrument_id == instrument_id
+    )
+    if start_date is not None:
+        stmt = stmt.where(InstrumentLog.end_date >= start_date)
+
+    if end_date is not None:
+        stmt = stmt.where(InstrumentLog.start_date <= end_date)
+
+    instrument_logs = session.scalars(stmt).all()
+    if len(instrument_logs) == 0:
```
Theodlz, line 3106: This causes frontend issues I think. The frontend should receive a dictionary with the Bokeh expected structure, but empty data. Not a list of 3 null values (when inspecting you can see a json error).

mcoughlin: I fixed Plot.jsx to check for ! null.

PR #4345, opened 2023-06-07
WIP - GCN Notification: No DB Trigger

Theodlz:
This PR tries to change how we trigger notifications on events. Instead of relying on new Localization being added and triggering a db trigger, we simply push to the queue in the method to ingest events. That way, we can handle missing localizations as well as wait for tags (of both events and localizations) to be added.


mcoughlin (approved):
This will be best tested in production. Here we gooooo

Theodlz:
> This will be best tested in production. Here we gooooo

Terrifying

PR #4346, opened 2023-06-07
altdata list

mcoughlin:
This PR makes altdata a list if it is not already.


mcoughlin:
Fixes issues in production. Self-merging.

PR #4347, opened 2023-06-07
Object delete with more checks

mcoughlin:
This PR restores the  object delete with more checks.


Theodlz (commented):
Hi @mcoughlin, I would also make sure there is: no classifications, no annotatis, no analysis, and maybe even no confirmed/rejected in gcn status.
What do you think?

Also, I think you need to run these queries not with the user that hit the endpoint, but was user_id=1, or any other way that actually has access to all of the data. Otherwise, we might allow deletion of an object even if there is data that this user doesn't have access to.

mcoughlin:
@Theodlz please see revised version!

Theodlz (approved):
LGTM. As suggested maybe consider a check on analysis, but other than that it looks good.

mcoughlin:
@Theodlz i checked but the way the mixin works makes that weird or something.

Theodlz:
> @Theodlz i checked but the way the mixin works makes that weird or something.

Ah ok got it. That's fine then, it can go in as is.

PR #4357, opened 2023-06-09
Annotation buttons: Styling

Theodlz:
Very small change to support another change in Fritz (annotation buttons plugins)


mcoughlin (approved):
Frontend!

PR #4358, opened 2023-06-10
Fetch Executed Observations in Localization: Better performance

Theodlz:
It seems that the subquery created when using the cache slows the query by at least a factor of 2 (but sometimes more than that as the time that a query takes in postgres varies).

That happens only when specifying an instrument to query for. Without it's somehow quicker. 

I also noticed that the frontend to fetch observations wasn't passing the selected instrument as query param, which is the behavior that is expected from the form.

So this PR might slow down fetching observations from the frontend, while accelerating/fixing the treasuremap submission. If it turns out that after these changes even if the query (with specified instrument id) is faster, fetching observations is now too slow and the treasuremap still takea too long, we can easily reverse the frontend change and make the treasuremap submission async.

Testing it as is in production is the best way to assess performance.

### DIFF
--- a/skyportal/handlers/api/observation.py
+++ b/skyportal/handlers/api/observation.py
@@ -308,7 +308,6 @@ def get_observations(
         area: number
             Area covered in square degrees
     """
-
     if return_statistics and localization_dateobs is None:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
Progress ;)

PR #4364, opened 2023-06-15
mean / error from map header

mcoughlin:
This PR modifies the map tags to read distmean / standard deviation from map header.


Theodlz (approved):
LGTM

PR #4366, opened 2023-06-15
TNS Monitoring

Theodlz:
In this PR, we add a TNS name retrieval queue that runs in the same place as the submission queue, only in a different thread.
This system will query for the TNS names from the last `X` days, and run 2 arcseconds cone searches to determine whether or not an object in the database might be associated with it. If that is the case, it will add a TNS name to it.

In a following PR, we want to extend that system to add as much `tns_info` as we can to the objects in the DB. Here we used bs4 to get the TNS data instead of the API, simply because the API does not return the coordinates of objects when fetching multiple objects at once, while the frontend does. Given that we get 50 objects per page, even with just 50 objects added in a day or 2, that brings us to 1 GET request instead of 1+50x1=51 requests, gaining speed while actually letting TNS take a breath (especially considering its very restrictive rate limits). This however does not get the full tns_info. We can imagine calling TNS' API object endpoint ONLY if there is a TNS crossmatch (found with bs4), and not for each object.

🚨 CONFIG CHANGE 🚨: there is now a `tns` key under which one can specify the TNS endpoint, bot_name, bot_id and api_key that can be used by the backend, instead of just **one** key called `tns_endpoint`

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -76,7 +76,7 @@ app:
     bot_id:
     bot_name:
     api_key:
-    look_back_days: 2
+    look_back_days: 1
 
   gracedb_endpoint: https://gracedb.ligo.org/api/
   # See https://computing.docs.ligo.org/guide/auth/x509/ for example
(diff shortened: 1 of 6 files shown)


Theodlz:
@mcoughlin let me know if you need credentials to try it out.

mcoughlin (approved):
This is going to be really nice!

PR #4368, opened 2023-06-16
Telescope observability

mcoughlin:
This PR updates check for telescope observability.


Theodlz (approved):
One way to make this a little shorter would be to just create all the variables set to None, and then go through the if statements without needing an else every time.

Other than minor comment, LGTM

PR #4369, opened 2023-06-16
Frontend: Minor user requests

Theodlz:
- Order the allocation by alphabetically by: telescope name => instrument name => pi
- New Source form: better validate to avoid some issues + better default layout.
- Minor bugfix with ingesting GCNs manually


mcoughlin (approved):
Other than the

if notify 

LGTM

PR #4371, opened 2023-06-17
Forced photometry queue

mcoughlin:
This PR modifies forced photometry such that the retrieval is automated.

### DIFF
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -35,7 +35,7 @@ request_session.trust_env = (
     False  # Otherwise pre-existing netrc config will override auth headers
 )
 
-WAIT_TIME_BETWEEN_QUERIES = timedelta(seconds=10)
+WAIT_TIME_BETWEEN_QUERIES = timedelta(seconds=120)
 
 queue = []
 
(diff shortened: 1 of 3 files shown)


Theodlz:
Hi @mcoughlin. 

I have one question, why an API + Service in this scenario where we already have the concept of transactions/followup requests with a status? From the frontend of skyportal's API, we could directly post the request to the database with a status like "processing", and this microservice could just poll from the DB and go through the requests to process, rather than add the complexity of running a multithreaded service like we do here right?

mcoughlin:
@Theodlz that could work, although I think it is easier to keep track of things queue style, if only because our other services seem to handle things well this way. I think this also will likely make for easier user polling of status; we will need some better way for users to be able to track the status of their requests in general.

Theodlz:
Let's try this then. I would try not to add to many multithreaded services in the future however. So let's get this one in, but we should think about single threaded model for this one at some point.

Theodlz (approved):
LGTM. Only comment is what about the frontend now that you removed the GET methods? People might be confused if they still see the button. At least we should make sure to display a decent error message that makes it clearer that the GET might be handled automatically.

mcoughlin:
@Theodlz shouldnt be a problem since we deleted the GET

Theodlz:
Ah you mean that the frontend checks if these methods are available before showing (or not) the associated buttons? In that case feel free to merge.

mcoughlin:
@Theodlz Yeah exactly.

PR #4375, opened 2023-06-19
WIP: publish event

mcoughlin:
This PR starts thinking about saving information to put on a dedicated table that will represent saved events. This piece will need a bunch added (including making survey efficiency analyses selectable).

My thinking is that we basically have users save the info in JSON to a dictionary from which the data is derived. This does mean it is basically time stamped and would need to be updated with a new JSON if more data comes through or something, however this means there will not be any dynamic queries against the database other than to load the JSON that have been flagged for publish, which would be done from that page where they are made.

What do you think @Theodlz ?


skyportal/models/gcn.py
```diff
@@ -128,13 +139,133 @@ class GcnPublication(Base):
...
+            html += "<div style='display: flex; flex-direction: row; justify-content: space-between; align-items: center;'><h2>Observations</h2>"
+
+        if len(data.get("observations", [])) == 0:
+            html += "</div><p>No observations</p>"
+        else:
+            html += "<button type='button' onclick='tableToCSV(\"observations\")' style='max-height: 30px;'>download as CSV</button></div>"
+            html += (
+                "<table style='width:100%' border='1' align='center' id='observations'>"
+            )
+            html += "<tr><th>Observation ID</th><th>Time</th><th>Instrument ID</th><th>Field ID</th><th>Filter</th><th>Exp. Time</th><th>Limiting Mag.</th><th>Airmass%</th><th>Processed %</th></tr>"
```
mcoughlin, line 188: @Theodlz Super minor, but this is not technically a %, just a fraction. Could multiply by 100 to get to %.

PR #4376, opened 2023-06-20
FacilityTransaction queue: bugfixes

Theodlz:
@mcoughlin feel free to add to this based on what we discussed on slack.

### DIFF
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -14,8 +14,8 @@ from requests.auth import HTTPBasicAuth
 
 import sqlalchemy as sa
 
-from baselayer.app.models import init_db
 from baselayer.app.env import load_env
+from baselayer.app.models import init_db
 from baselayer.log import make_log
 from skyportal.models import (
     DBSession,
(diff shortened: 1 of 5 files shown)


mcoughlin:
@Theodlz Seems to be ingesting alright after your suggested changes. Please lemme know.

services/facility_queue/facility_queue.py
```diff
@@ -237,29 +257,39 @@ def service(queue):
...
-                            except Exception:
-                                status = 'In progress: Light curve not yet available. Waiting for it to complete.'
-                                followup_request.status = status
-                                log(f'Job {req.id}: {status}')
-                                req.last_query = datetime.utcnow()
+                            exitcode = row['exitcode']
+                            exitcode_text = ZTF_PHOTOMETRY_CODES[exitcode]
+
+                            if exitcode in [63, 64, 65, 255]:
+                                req.status = f'No photometry available: {exitcode_text}'
```
Theodlz: @mcoughlin is that what we want, or do we want that be the followup request status is set to 'no photometry available', while the facility transaction is set to complete? That way its not fetched anymore when we restart the queue.

skyportal/facility_apis/atlas.py
```diff
@@ -312,7 +319,19 @@ def delete(request, session):
...
+
+        transaction = (
+            session.query(FacilityTransactionRequest)
+            .filter(FacilityTransactionRequest.followup_request_id == request.id)
+            .first()
+        )
+        if transaction is not None:
+            if transaction.status == "complete":
+                raise ValueError('Request already complete. Cannot delete.')
+            session.delete(transaction)
```
Theodlz, line 332: Nice. Thanks!

services/facility_queue/facility_queue.py
mcoughlin: yeah I had fixed that in the last commit.

PR #4377, opened 2023-06-21
More mods for HA

mcoughlin:
This PR ensures HA is between 0-24 before converting to negative HAs.


mcoughlin:
Bug-fix to account for hour angle -12-12 calculation issues. Self-merging.

PR #4378, opened 2023-06-22
Require groupIds for bulk TNS reporting

mcoughlin:
This PR ensures groupIds are specified for bulk TNS reporting.


Theodlz (approved):
LGTM

PR #4380, opened 2023-06-22
Source's GCN event status: set obj.gcn_crossmatch

Theodlz:


mcoughlin (approved):
This is really nice. It also makes me wonder if we made a mistake with dateobs as the type it is ;)

Theodlz:
> This is really nice. It also makes me wonder if we made a mistake with dateobs as the type it is ;)

You mean having it as a string and not a date in the array? I think it makes sense actually, that allows for stuff like querying by crossmatch to be easier if we ever need that.

mcoughlin:
@Theodlz Oh good! That makes me feel better ;)

PR #4381, opened 2023-06-22
Observations in GCN: Min obs per field

Theodlz:
This PR adds an (optional) parameter to specify a minimum of observations per field required when fetching observations overlapping with a localization at the Nth credible region. 
This is useful as some folks only care about reporting observations of fields that have been observed at least N times (especially when generating summaries, which has been added here too).

When None or set to 1, this will simply be ignored and the query will be unchanged.


mcoughlin (approved):
This will be a useful parameter to have.

PR #4382, opened 2023-06-22
Photometry upload bugfix: avoid issue with obj_ids converted to int64

Theodlz:
Pandas converts obj_ids that are just made of numbers to int64, which breaks the query that checks if the object exists as the datatype doesn't match the one of the column. This is a quick bugfix.


mcoughlin (approved):
LGTM

PR #4383, opened 2023-06-23
HourAngleConstraint transposes

mcoughlin:
This PR catches corner cases in the HourAngleConstraint.


mcoughlin:
Adds robustness to the HourAngleConstraint code. Self-merging.

PR #4386, opened 2023-06-24
Reduce dimensions in HA constraint

mcoughlin:
This PR reduces dimensions if necessary in HA constraint.


mcoughlin:
Bugfix on corner case for HA scheduling constraint. Self-merging.

PR #4388, opened 2023-06-26
sncosmo fit limits

mcoughlin:
This PR removes the limits between first and last detection when doing sncosmo fit.


Theodlz (approved):
LGTM

PR #4390, opened 2023-06-26
TNS 2 arcsec crossmatch: update all sources

Theodlz:
Add TNS to all sources within 2 arcsec, and not just the closest one.


mcoughlin (approved):
LGTM

PR #4393, opened 2023-06-27
No need for observation plans queried with GcnEvent

mcoughlin:
This PR removes observation plans queried with GcnEvent, as they are not used.


Theodlz (approved):
Very nice, always like PR removing stuff 💯. LGTM

PR #4394, opened 2023-06-27
numberObservations to TreasureMap submission query

mcoughlin:
This PR adds the number of observations to the treasuremap submission query.


Theodlz (approved):
LGTM! Could you maybe add it to the docstring here too:
https://github.com/mcoughlin/skyportal/blob/tmap_ndet/skyportal/handlers/api/observation.py#L1598-L1604

Otherwise, this can go it.

PR #4397, opened 2023-06-28
GcnEvent: Always return True or ambiguous in list

mcoughlin:
This PR adds source query option to always return sources in True or ambiguous.


Theodlz (approved):
LGTM

PR #4398, opened 2023-06-28
TNS instrument nickname error

mcoughlin:
This PR fixes the TNS instrument nickname error.


mcoughlin:
Fixes mistake in an error message returned. Self-merging.

PR #4400, opened 2023-06-29
TNS submission queue port: Change to something an api instance wouldn't use

Theodlz:
After deploying an app kept dying caus it used the port I specified (app 10), as the app ports are in that range.
I added a hotfix, but good to know.


Theodlz:
self merging

PR #4403, opened 2023-06-29
Expanded Annotation table: Fit screen

Theodlz:
We already had the possibility to open a dialog with the annotation table (on the source page) so that it could be easier to read. However, this didn't work, as its height was still constrained and basically showed the same table. Also, the button to expand the table showed even in the table in the dialog, resulting in an annotations-ception if used incorrectly 🕳️ .

This PR solves both of these problems.


mcoughlin (approved):
LGTM

PR #4404, opened 2023-06-30
Photstats: added stats ignoring forced photometry

Theodlz:
This PR adds 7 stats:
- first_detected_no_forced_phot_mjd
- first_detected_no_forced_phot_mag
- first_detected_no_forced_phot_filter
- last_detected_no_forced_phot_mjd
- last_detected_no_forced_phot_mag
- last_detected_no_forced_phot_filter
- num_det_no_forced_phot_global

These are useful, so when running queries on sources and candidates with constraints on first/last detections and number of detections, we can specify whether or not to ignore forced photometry. This PR also adds that.

It will be followed by another PR adding the frontend for users to ignore or not forced photometry, as we first need to update all of the photstats.


mcoughlin (approved):
LGTM

guynir42 (approved):

PR #4426, opened 2023-07-01
Update workflows from python 3.8 to 3.9

mcoughlin:
This PR updates workflows from python 3.8 to 3.9.


mcoughlin:
Python 3.9 seems to be working well. Merging.

PR #4427, opened 2023-07-01
Bump to node.js 18

mcoughlin:
This PR bumps Dockerfile to use node.js 18.


mcoughlin:
Need to trigger Docker build. Merging.

PR #4428, opened 2023-07-02
Follow-up request allocation info

mcoughlin:
This PR adds more more info for follow-up request allocations.


Theodlz (approved):
LGTM. I would maybe just rename `Allocation PI` to just `PI`, as it's the allocation table anyway. That would help not have a table where we need to scroll horizontally.

PR #4429, opened 2023-07-03
galaxy table filter

mcoughlin:
This PR adds a filter on galaxy table.


Theodlz (approved):
@mcoughlin I tried to fix some weird behaviors of the form. Worth trying it again before merging. But otherwise, LGTM!

PR #4430, opened 2023-07-04
Add telescope summary page

mcoughlin:
This PR adds a telescope summary page with associated instruments and allocations tables.


Theodlz (approved):
Just added the weather widget there as it might be useful. Otherwise, LGTM

PR #4433, opened 2023-07-06
Use jd instead of jdendhist to account for schema change in Fink

JulienPeloton:
The table behind the conesearch had a change in its schema -- and `jdendhist` is now `jd`.

In addition, conesearch performances should be much better.


mcoughlin (approved):
LGTM

PR #4435, opened 2023-07-06
Frontend - Candidate scanning and sources in GCN: exclude forced photometry

Theodlz:
This PR is the frontend counterpart of #4404, added separately to not expose the frontend in prod before all the photstats are updated.


mcoughlin (approved):
LGTM

PR #4436, opened 2023-07-07
Dockerfile: add screen

Theodlz:
add screen to the docker image, which is very useful when we need to run long running scripts to update entries of the db in bulk for example, which has proven challenging and suboptimal which dedicated endpoints (both for development and when running the code).

With screen, we can create screens in which to run the scripts, detach (when leaving the container) and re-attach to verify the progress of a script.


mcoughlin (approved):
LGTM

PR #4441, opened 2023-07-19
ML-based classifications

Theodlz:
This PR aims to make it much cleared what is a real/definitive human classification of a source/object, and what is a score from an ML classifier.

DO NOT MERGE YET: I am waiting for some more feedback


bfhealy (commented):
@Theodlz thanks for implementing this! I have two comments:

- I think it would be best to preface ML classifications with "ML: " instead of "ML-". There are some hyphenated labels that currently look a bit strange, e.g. `ML-non-variable`.
- I wasn't able to get the "Hide ML Classifications" option to work. My test ML classifications were still displayed after toggling, both on the group sources page and for an individual source.

Theodlz:
> @Theodlz thanks for implementing this! I have two comments:
> 
> * I think it would be best to preface ML classifications with "ML: " instead of "ML-". There are some hyphenated labels that currently look a bit strange, e.g. `ML-non-variable`.
> * I wasn't able to get the "Hide ML Classifications" option to work. My test ML classifications were still displayed after toggling, both on the group sources page and for an individual source.

Hi @bfhealy, thanks for the review!
- I addressed your first comment. 
- When it comes to the second, I wasn't able to reproduce this bug. When you load the demo data, you should see a source (`ZTF18aabcvnq`) that has both ML and non-ML classifications. Button works there for me, and the slider in the user preferences does enforce a default behavior. Could you try again by flushing the cache of the browser first?

bfhealy:
Hi @Theodlz, thanks for implementing that change! I tried flushing my chrome browser's cache and still can't get the demo ML classification to be hidden by checking the option in my preferences. 

Not sure if this is helpful, but inspecting my browser console shows that `skyportal/FETCH_SOURCES_OK` is displayed before `skyportal/FETCH_USER_PROFILE_OK`. Perhaps my `hideMLClassifications` preference is not provided to the frontend until after the sources/classifications have been fetched and displayed?

Theodlz:
just to be sure @bfhealy, are you talking about the classifications being hidden from the list on a single source page, or are you saying that they do not disappear from the sources list?

Theodlz:
Because the feature to hide it is just meant for the list of classifications on an individual source page (they will display anywhere else, including on top of the source page. It's just for the list).

If you think it should be extended to other places where this appears, just let me know.

bfhealy:
@Theodlz Ah I see! The code is working as intended then. I was expecting the ML classifications to be hidden from top of a source page and group sources as well, but I think your current approach is preferable. It would seem too much to allow one checked option to withhold all mentions of ML classifications on Fritz. As you said, I think the way the "Hide" option currently works combined with prefacing each classification with 'ML: ' is sufficient, at least until we get more feedback.

bfhealy (approved):

Theodlz:
> @Theodlz Ah I see! The code is working as intended then. I was expecting the ML classifications to be hidden from top of a source page and group sources as well, but I think your current approach is preferable. It would seem too much to allow one checked option to withhold all mentions of ML classifications on Fritz. As you said, I think the way the "Hide" option currently works combined with prefacing each classification with 'ML: ' is sufficient, at least until we get more feedback.

I agree. Let's get this in once green, and I'll make any changes if necessary.

PR #4442, opened 2023-07-20
Obsplan queue refactor

Theodlz:
This PR's main goal is to refactor the obsplan queue service, to create requests as they are submitted and not only when they start processing, to better keep track of what will be processed or not.

TODOs:
- [x] The frontend tries to grab a plan as soon as it finds a request. Modify this behavior to only do so once successfully processed.
- [x] Migrations.
- [x] Try it thoroughly :)

Following PR:
- When creating combined plans, I always got one plan with observations and one without. To fix
- What about combined plans on different allocations/instruments? What is the behavior we want here? To test


Theodlz:
> I see logs like:
> 
> [09:48:46 observation_plan_queue] Error occured prioritizing the observation plan queue: 'NoneType' object is not subscriptable [09:48:46 observation_plan_queue] Error occured processing the observation plan queue: list index out of range
> 
> This seems to occur on ingestion of new events, so something funny with the default plans, although everything seems to finish eventually.

The index stuff should be fixed by now. Haven't seen the other error however, but the fix for nb 2 could have also fixed nb 1.
Could you give it a try? Frontend should be fixed as well (at least what I knew needed to change). I also added a default survey efficiency to the demo data, so that can be tested as well. Best is to add an LVC event manually once the demo data is ingested to try that too, and add some plans manually from the frontend (combined and single plan).

But I believe this queue has the same behavior as the previous version, which was the goal of this PR.

mcoughlin:
@Theodlz Tested both individual and combined and both ran and I dont see that same error message.

Theodlz:
> @Theodlz Tested both individual and combined and both ran and I dont see that same error message.

Awesome. Do you know anything about this?
`When creating combined plans, I always got one plan with observations and one without. To fix`

Basically combined plans didn't work at all already looks like. I had to add an extra parameter to what is given to gwemopt as gwemopt said it was missing. Any suggestion on what to set it to? (True or False).

and for: 
`What about combined plans on different allocations/instruments? What is the behavior we want here? To test`
what do you expect that behavior to be?

Theodlz:
I agree. This PR is about improving the current queue behavior, not fixing gwemopt. So if you are happy with the changes, I would consider merging once green.

PR #4443, opened 2023-07-20
Annotation sorting test: fix

Theodlz:


mcoughlin (approved):
Thanks for trying to make this more robust.

PR #4445, opened 2023-07-27
Frontend caching fix: routes lazy loading

Theodlz:
This PR is an attempt to fix the issues with chrome and firefox not caching our monolithic webpack that's over 50MB, by splitting it in a smaller main bundle, along with individual assets for each route. 

This seems to fix all caching issues on chrome and firefox, and here's why:

Currently, our webpack is over 50MB. This seems to prevent browsers like chrome and firefox from caching it once it has been loaded. For firefox, this is imply because there is a hard limit to how big one file can be to be cached. For chrome it is still unclear, as they claim there is no limit, even though it behaves like there is one.
Most websites don't send the client all of the assets as soon as they access any page, but rather split their website in smaller chunks, that will be lazy loaded in the background, or only when the client goes to a page where they're needed.

In SkyPortal, there is a `Main.template.jsx` file which generates `Main.jsx`, using values from the config file. This `Main.jsx` file is what will contain references to all of the valid routes (webpages) of our website, and it is the entry point when a client accesses any page.

So far, Main.jsx was importing all of the pages of SkyPortal without any lazy loading, meaning that as soon as you access any page, it will download the whole website. Because of that, webpack was also generating one big file for our whole website, meaning it couldn't be split in several smaller assets.

In this PR, the main change we make is to use `React.lazy` in `Main.template.jsx` in the for loop that generates all of the import based on the routes found in the config. That way, webpack will split the assets in multiple ones:
- one or more asset(s) with shared ressources/components, used accross several pages/routes.
- assets for each route of the website.

On the client side this means:
- only the assets you need for the page you are trying to access will be loaded, which speeds up initial loading of the website.
- As you navigate to more and more pages, the full webpack will effectively be downloaded.
- As each asset is now smaller, they are cached as expected!

Also, since each assets are named with a hash generated based on their content, even after deploy, if some pages haven't changed, it is likely that the asset will be the same, and won't need to be redownloaded!

**BUT**, this did create some issues with the existing code. Mainly, this has to do with redux. As the order in which things are loaded changed, it might be that a page is now looking for a ressource in a redux state before it has been loaded. Here is an example:

let's say you had code like `const origins = useSelector((state) => state.photometry.origins)` but before it is hydrated, the photometry state might be undefined, or an empty dict without any `origins` key. Now that things are lazy loaded and split by routes, it might throw an error like `state.photometry is undefined`. Luckily, there are plenty of ways to fix this. For instance, you can replace your line with `const photometry = useSelector((state) => state.photometry)`, and later when you need the origins do something like `const origins = photometry?.origins || []` which will just give you an empty list while what you need is being fetched, and will rerender once it's done fetching.

### DIFF
--- a/skyportal/tests/frontend/test_user.py
+++ b/skyportal/tests/frontend/test_user.py
@@ -3,6 +3,6 @@ def test_user_info(driver, super_admin_user):
     driver.get(f'/become_user/{user.id}')
     driver.get(f'/user/{user.id}')
     driver.wait_for_xpath(f'//div[contains(.,"{user.username}")]')
-    pg_src = driver.page_source
     for acl in user.permissions:
-        assert acl in pg_src
+        permission = f'//ul/li[contains(.,"{acl}")]'
+        driver.wait_for_xpath(permission)
--- a/static/js/components/CommentEntry.jsx
+++ b/static/js/components/CommentEntry.jsx
@@ -30,7 +30,9 @@ const useStyles = makeStyles(() => ({
 
 const CommentEntry = ({ addComment, editComment }) => {
   const styles = useStyles();
+  const users = useSelector((state) => state.users);
   const { userAccessible: groups } = useSelector((state) => state.groups);
+  const { instrumentList } = useSelector((state) => state.instruments);
   const [textValue, setTextValue] = useState("");
   const [textInputCursorIndex, setTextInputCursorIndex] = useState(0);
   const [autosuggestVisible, setAutosuggestVisible] = useState(false);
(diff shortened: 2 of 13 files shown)


Theodlz:
I'm working on fixing the tests, but worth it for the reviewers to test it locally, especially to verify the caching behavior.

stefanv:
I only see two lazily loaded additions now, but in principle these changes LGTM.

Theodlz:
> I only see two lazily loaded additions now, but in principle these changes LGTM.

Hi @stefanv, the changes are in the template which generates all of the import, so the 20+ routes we have are now all lazy loaded once the Main.jsx file is generated.

stefanv:
Excellent, that's perfect!

Here's the line I missed: https://github.com/skyportal/skyportal/pull/4445/files#diff-dff53895f7170d810d629971ec586989e6b299461fcadabd826b419d61081288R41

profjsb (approved):
This looks great. I like how the meat of the changes are small and straightforward. Might be good to document in the PR comments precisely what the changes do that plays better with browser caching. 

Note: I've not pulled the PR to test caching behavior locally.

Theodlz:
> This looks great. I like how the meat of the changes are small and straightforward. Might be good to document in the PR comments precisely what the changes do that plays better with browser caching.
> 
> Note: I've not pulled the PR to test caching behavior locally.

Thanks for taking a look at the changes Josh. 

I updated the PR's description to explain a bit better how this works.

PR #4446, opened 2023-07-27
Spectrum: Fix patching reducers and observers

Theodlz:
It looks like the method to update a spectrum isn't doing what the documentation is claiming it does.
This gets us closer to that by allowing one to update the list of reducers and observers of a spectrum.


mcoughlin (approved):
LGTM

PR #4447, opened 2023-07-28
WIP - ZTF cutouts for all objects

Theodlz:
This adds a super basic version to grab ZTF thumbnails for all objects that aren't coming from ZTF alerts, by grabbing images from IPAC.

TODO:
- [ ] Grab the right image(s), to decide for which filters, and which type.
- [x] Find more recent images? So far, what I got from IPAC looks a little old. Maybe there is a better source for the images.
- [ ] Migrations, as usual. **SOLVED: grab all images that cover a 63x63 arcsec area centered on the object, sort by date, and grab the most recent.**
- [ ] Frontend improvement? If we get cutouts for different bands/filters, we might want to make it so you see only one by default, but can switch between bands with a slider/menu element.

Tagging you @AshishMahabal here so you can give you help figure out what's needed.

So far, it simply grabs the science image that is closest to the object's position, and requests a cutout of size 63x63 from that image, and uploads it. However it can be of any band (as it just grabs the closest img), and the date at which the image  was taken isn't saved anywhere. We might want to add that. 

It simply looks like this, following the current standard we have for cutouts:
image

### DIFF
--- a/services/thumbnail_queue/thumbnail_queue.py
+++ b/services/thumbnail_queue/thumbnail_queue.py
@@ -1,24 +1,26 @@
-from threading import Thread
+import asyncio
+import io
+import json
+import re
 import time
+import traceback
+from threading import Thread
 
+import pandas as pd
+import requests
+import sqlalchemy as sa
+import tornado.escape
 import tornado.ioloop
 import tornado.web
-import asyncio
-import tornado.escape
-import json
+from sqlalchemy.orm import scoped_session, sessionmaker
 
-import sqlalchemy as sa
-from sqlalchemy.orm import sessionmaker, scoped_session
-
-from baselayer.app.models import init_db
 from baselayer.app.env import load_env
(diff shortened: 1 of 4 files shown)


Theodlz:
@mcoughlin feel free to jump in an weight on how we should handle this best.

AshishMahabal:
@Theodlz ideally there should be an image nearest to a particular date that may be of interest, plus the oldest and the latest image. All this for each filter. If there is more "space" an image corresponding to a brightest point in the time series, and a faintest point in the time series (any would do in case of redundancy). A slider if multiple cutouts are available would indeed be useful.

mcoughlin:
@Theodlz Is this still relevant?

Theodlz:
@mcoughlin I haven't feel a strong need to get that in the app for a couple of months, so we can close the PR for now and we can always re-open later

PR #4448, opened 2023-07-28
Basic Sun and Moon markers on the Localization plot

Theodlz:
This PR simply adds 2 yellow and grey circles where the sun and moon are on the skymap, with labels and some basic info.

Question for @mcoughlin: Do you think we want to display their position at the dateobs, or current time? We could do both with a slider or smth as well.

Oh and dont review the code yet, I need to cleanup this old branch. EDIT: done, lots of commits for nothing but its synced with main


mcoughlin:
@Theodlz default should be the current time, although adding a slider would be awesome!

Theodlz:
> @Theodlz default should be the current time, although adding a slider would be awesome!

Great! Are you ok with me adding that in a seperate PR? That way I could deploy this today.

mcoughlin:
@Theodlz of course!

PR #4449, opened 2023-07-28
Spectrum upload: Pre-fill form using URL arguments

Theodlz:
This PR allows us to pre-fill the upload spectrum page's form by passing arguments directly in the URL.
When it comes to uploading the file itself, the file can be passed as a URL from where it can be downloaded.

These are the arguments that can be passed:
-observed_by (list of ints of user ids)
- reduced_by (list of ints of user ids)
- spectrum_type (source, host, or host_center)
- instrument_id (int)
- mjd (float)
- group_ids (list of ints of group ids)
- has_fluxerr (Yes or No)
- fluxerr_column (int, nb of the column in the ascii file. To pass only if has_fluxerr is Yes)
- flux_column (int, nb of the column in the ascii file)
- wave_column (int, nb of the column in the ascii file)
- file_name (string)
- file_url (string)

String values should NOT be in quotes, though we do recommend it for the file_url as it might confuse the router otherwise.

### DIFF
--- a/static/js/components/UploadSpectrum.jsx
+++ b/static/js/components/UploadSpectrum.jsx
@@ -1,7 +1,7 @@
 import React, { Suspense, useEffect, useState } from "react";
 import PropTypes from "prop-types";
 import { useDispatch, useSelector } from "react-redux";
-import { Link } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Cool

PR #4450, opened 2023-07-28
Phot plot height fix when legend is too long

Theodlz:
If there are too many items in the legend, those are cropped and it makes it impossible to hide/show items on the photometry plot. This is a quick fix as this is impacting user experience, but we might want to handle this better in the future.

Basically, here, we just add more height as needed, which already the case on other deviced, but didn't work on desktop as we forced the height to 500.


mcoughlin (approved):
LGTM

Theodlz:
Tested the source page locally (ran all the tests), and it is working. Merging

PR #4473, opened 2023-08-01
langchain 0.0.229

mcoughlin:
This PR updates langchain to the latest on pypi.


mcoughlin:
Package update fixing the docs. Self-merging.

PR #4474, opened 2023-08-01
Skymap trigger: Search for event

Theodlz:
We can already search for an event to search (by dateobs or name) on multiple pages, but it was missing here.


mcoughlin (approved):
LGTM

PR #4482, opened 2023-08-10
WIP: Add script to run analysis on candidates

bfhealy:
This WIP PR adds a script to `jobs` that runs an analysis service on transient candidates. The intent is for this script to be run as a cron job to facilitate regular automated analyses of the latest data.

Currently the analysis service and model can be specified in this script's arguments. It may be useful to add a configuration file that contains more specific parameters for a given service (e.g. `tmin`, `tmax` and `dt` for NMMA_Analysis).

The script still needs code to identify the candidates on which to run the analysis. One way of doing this could be to keep a running list of recent candidates along with number of detections, and any time a new candidate is saved or an existing one gets more photometry, that candidate is queued for a new analysis. [This script](https://github.com/ZwickyTransientFacility/scope/blob/4112cb426e536b62936707d2d8816d69d58c0fda/gcn_cronjob.py#L4) offers an example of a running list of candidates that is checked each run (for a different workflow).

### DIFF
--- a/jobs/candidate_analysis_cronjob.py
+++ b/jobs/candidate_analysis_cronjob.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+import os
+import urllib
+import requests
+import time
+from typing import Optional, Mapping
+from requests.exceptions import InvalidJSONError, JSONDecodeError
+from urllib3.exceptions import ProtocolError
+import argparse
+
+token = os.getenv('TOKEN')
+protocol = os.getenv('PROTOCOL')
+host = os.getenv('HOST')
+base_url = f"{protocol}://{host}/"
+
+MAX_ATTEMPTS = 10
+SLEEP_TIME = 5
+
+
+def api(
+    method: str,
+    endpoint: str,
+    data: Optional[Mapping] = None,
+    token: str = token,
+    base_url: str = base_url,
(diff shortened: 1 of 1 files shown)


bfhealy:
Currently testing with a local setup of the NMMA API and encountering an error when making a second NMMA_Analysis API call while the first is still running:
```
At line 152 of file /Users/bhealy/nmma/MultiNest/src/nested.F90
Fortran runtime error: Attempting to allocate already allocated variable 'pwrap'
Trace/BPT trap: 5
```

The script will need to gracefully handle many NMMA API calls corresponding to each candidate's analysis run. This could involve some kind of multi-threaded solution on the NMMA side of things or a code in the script to ensure only one analysis run is called at a time.

mcoughlin:
@bfhealy Does dynasty behave better? Or ultranest assuming it has been fixed by Peter.

bfhealy:
@mcoughlin I don't get the fortran error using dynesty or ultranest, although making more than one API call seems to slow things down significantly. I can get both samplers to finish for a single source outside of the API, so I'll continue testing to see if there's some difference in inputs that's partially responsible for the API slowdown.

Theodlz:
I see how this could be useful to test some stuff, but this is hardcoded with a specific source, and not of any use in production. + we already have the concept of recurring API calls that can be used here to trigger periodic jobs. Closing

PR #4485, opened 2023-08-24
Photometry updates

mcoughlin:
This PR enables photometry updates from the photometry table.

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -1,3 +1,4 @@
+import copy
 import uuid
 import datetime
 import json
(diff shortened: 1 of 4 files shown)


Theodlz (approved):
LGTM! This looks great. Thanks for adding that feature.

Theodlz:
Any reason why its only these 3 quantities?

mcoughlin:
@Theodlz Good point. I just added limiting magnitude. Instruments and filters are more subtle, so I think we should leave those alone. Folks can always delete and re-add points for complicated uploads.

Theodlz:
@mcoughlin yes I was precisely thinking about limiting_mag. For the rest, it can remain the same. Why not allowing to change the metadata, but not essential either. This is more than enough for this PR. 

LGTM

PR #4486, opened 2023-08-24
FollowupRequestForm - frontend crashes fix

Theodlz:
This is an attempt to prevent an issue a user (yes, just one) encounters when going to the source page.


mcoughlin (approved):
The user is me???

Theodlz:
> The user is me???

Surprisingly, no 👀

PR #4487, opened 2023-08-29
instrument and stream IDs for TNS submission

mcoughlin:
This PR adds instrument and stream ID options for TNS submission.


Theodlz:
@mcoughlin just to keep you up to date: I have 2 commits that I'll take care of tomorrow before this is merged:
- migrations
- on the frontend form on the source page, only show instruments which submission to TNS is supported. This is done by simply passing the mapper we have in `skyportal/utils/tns.py` to the config handler.

PR #4488, opened 2023-08-31
Observation planning - Use references, better frontend.

Theodlz:
This PR allows us to pass reference filters to gwemopt, but also improves on a few things:
- only show sub-parameters when the parameter they depend on it set on the right value, which helps make the frontend form WAY more usable.
- allows to change the airmass threshold used when plotting fields, from the frontend directly.
- fixes a bug where downloading an airmass chart on browsers other than firefox wouldn't save it as PDF.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -685,21 +685,6 @@ gcn:
   reject_tags: # reject notices with these tags (optional)
     - MDC
 
-  observation_plans:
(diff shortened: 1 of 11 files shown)


mcoughlin (approved):
Great thanks @Theodlz

Theodlz:
I've got some tests to fix first. Will do tomorrow so we can hopefully deploy in the afternoon before this weekend.

PR #4509, opened 2023-09-02
Faster airmass, rise and set time calculations

Theodlz:
This PR duplicates some of the methods found in telescopes, instruments, and observation_plans models, and refactors them to process multiple fields add once. That way, fetching the instrument fields that overlap with a skymap (where we calculate airmass), and fetching a single observation plan (where we calculate rise and set times) is **much** faster.

The speed up is more than x100 times, and even better than that for large maps and for plans with a lot of observations. It takes us from a few seconds to virtually nothing, removing some of the delay experienced when going to an event's page.

Latency aside, the goal here is to reduce the load put on the machine running the app.

### DIFF
--- a/skyportal/handlers/api/instrument.py
+++ b/skyportal/handlers/api/instrument.py
@@ -1,24 +1,26 @@
-from marshmallow.exceptions import ValidationError
-from baselayer.app.access import permissions, auth_or_token
-from baselayer.log import make_log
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
This calculations code is definitely nice for the future package

Theodlz:
> This calculations code is definitely nice for the future package

yes for sure. I'm thinking about starting that soon btw.

PR #4510, opened 2023-09-04
Modify Instruments; add/update references

Theodlz:
Make it possible to add references from the frontend, and also fixes the has_fields flag which had a tendency to stay false.


skyportal/handlers/api/source.py
```diff
@@ -1563,7 +1563,7 @@ async def get_sources(
                 count_stmt = sa.select(func.count()).select_from(stmt.distinct())
                 total_spectrum = session.execute(count_stmt).scalar()
                 obj_list[-1]["spectrum_exists"] = total_spectrum > 0
-            if include_spectrum_exists:
+            if include_comment_exists:
```
mcoughlin, line 1566: Thanks for fixing this ;)

mcoughlin (approved):
LGTM

PR #4511, opened 2023-09-05
TNS id mapper: lowercase

Theodlz:
Change the instrument names in the `TNS_INSTRUMENT_IDS` mapper to be lowercase, and make sure we also compare it to lowercased values anywhere it is used.

This is just to make sure that this works regardless of how the instrument name is written, useful for instances other than Fritz.


mcoughlin (approved):
Seems straight-forward enough. LGTM.

PR #4512, opened 2023-09-05
Modify instruments: update regions

Theodlz:
This makes it possible to update the contour and contour summary of fields of an instrument.


mcoughlin (approved):
LGTM

PR #4513, opened 2023-09-05
Frontend: tns instruments lowercase

Theodlz:
Missing frontend changes, to make sure we still show the tns instruments correctly


mcoughlin (approved):
LGTM

PR #4514, opened 2023-09-06
TNS prefix

mcoughlin:
This PR adds a TNS prefix to the query. It also adds the classification to the news feed.

### DIFF
--- a/skyportal/handlers/api/news_feed.py
+++ b/skyportal/handlers/api/news_feed.py
@@ -76,7 +76,9 @@ class NewsFeedHandler(BaseHandler):
 
         with self.Session() as session:
 
-            def fetch_newest(model, include_bot_comments=False):
+            def fetch_newest(
+                model, include_bot_comments=False, include_ml_classifications=False
+            ):
                 query = model.select(self.associated_user_object)
                 if model == Photometry:
                     query = query.where(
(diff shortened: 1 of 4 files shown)


static/js/components/NewsFeed.jsx
```diff
@@ -172,7 +172,13 @@ const NewsFeedItem = ({ item }) => {
         <div className={styles.entryIdent}>
           <span className={styles.entrySourceId}>
             <Link to={`/source/${item.source_id}`}>
-              Source: {item.source_id}
+              {item.classification ? (
```
Theodlz, line 175: maybe the usual `?` between item and classification?

skyportal/handlers/api/news_feed.py
```diff
@@ -100,6 +100,23 @@ def fetch_newest(model, include_bot_comments=False):
 
                 return newest
 
+            def latest_classification(obj):
```
Theodlz, line 103: Nice, I would just wrap the whole logic (right after `classification = None`) in a try except with a pass, so it returns no classification instead of an error and the rest of the code can work just fine.

Theodlz (approved):
LGTM. Minor suggestions

static/js/components/NewsFeed.jsx
Theodlz: Done

PR #4515, opened 2023-09-06
Remove ..._exists flags when fetching multiple sources

Theodlz:
An attempt to fix or at least understand where the performance issues come from when querying list of sources. It seems not to come from querying the sources itself, but all the subqueries we run for each source. At least, this is what appears when running it locally on a small database.

There are many more subqueries, so this might not be enough of course, but worth trying.

### DIFF
--- a/services/tns_retrieval_queue/tns_retrieval_queue.py
+++ b/services/tns_retrieval_queue/tns_retrieval_queue.py
@@ -232,13 +232,13 @@ def tns_retrieval(
             'User-Agent': f'tns_marker{{"tns_id":{tnsrobot.bot_id},"type":"bot", "name":"{tnsrobot.bot_name}"}}'
         }
 
-        _, tns_name = get_IAUname(
+        tns_prefix, tns_name = get_IAUname(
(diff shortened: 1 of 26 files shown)


mcoughlin (commented):
@Theodlz I suspect Dan and AMPEL are the main folks who might use this. Maybe check with them quick?

Theodlz:
I'm trying to fix the react memo stuff, looks like rerendering triggers another API call from the VEGA plots. There was no memo for the photometry which was already an issue, but adding it doesnt seem to fix that. Once fixed, we can merge and try that on preview to see how the performance changes.

Theodlz:
> @Theodlz I suspect Dan and AMPEL are the main folks who might use this. Maybe check with them quick?

Of course. This is mostly to assess performance. If it does prove useful on preview, then I'll ask around.

mcoughlin (approved):
I am ok with giving it a try then!

Theodlz:
Merging so we can try it on preview. Will revert if needed.

PR #4519, opened 2023-09-08
Fix source query cache

Theodlz:
Right now, the `use_cache` flag is forced to true, no query_id can actually be provided. This results in a broken cache feature, running a longer query for every single query and affecting performance.

In a following PR, we will make sure to use the cache from the frontend.


mcoughlin (approved):
LGTM

PR #4520, opened 2023-09-08
P200 Starlist formatting

Theodlz:
Fixes formatting issue with the P200 starlist, which prevents observers from using it as is.


mcoughlin (approved):
LGTM!

PR #4521, opened 2023-09-09
TNS Watcher service: keep TNS prefix

Theodlz:
I was wondering why we had the TNS prefix on the TNS objects we create but not the existing objects. This is because the watcher service that takes care of that didn't get updated like the other service.


mcoughlin (approved):
LGTM

PR #4522, opened 2023-09-11
TNSRobotHandler: change permission

Theodlz:
Set the required permission to `Manage groups`, given that robots are attached to groups it should be the same permission.


mcoughlin (approved):
LGTM

PR #4524, opened 2023-09-11
Facility service: avoid duplicate notifications

Theodlz:
RIght now, we update a Followup request's status every time we query the external service for the results. This causes us to spam some users.

This PR makes sure that we don't update the status if it hasn't changed. But we still update the associated facility transaction request as needed by the service.


mcoughlin (approved):
LGTM

PR #4525, opened 2023-09-11
TNS watcher: avoid unnecessary operations

Theodlz:
Avoid committing to the DB when nothing has been changed. It's mostly so we don't log that the TNS name has been updated when it remains the same


mcoughlin (approved):
LGTM

PR #4527, opened 2023-09-13
Frontend: cleanup warnings, share redux state to new tabs

Theodlz:
This PR does 2 major things:
- Share the current tab's redux state to new tabs, avoiding a lot of API queries. A lot of said queries were cached anyway, but they still queried the server. It is simply that the server would give a 304, avoiding sending the data back to the client. But the query would still go through if I understand correctly. But also, some repeated queries did hit the server and get results back. This should help those who like opening a bunch of tabs:) Worth giving it a shot. Also, `redux-sync-state` allows to share the dispatch actions between tabs. Naturally, we don't want to share any of the fetching actions as this would cascade into a ton of queries, but we can share things like notifications, and also most of the `..._OK` actions, with the results from a fetching. I'll whitelist some of those either in this PR or a separate PR. I want to be careful about that.
- Get rid of a LOT of the warnings. Few of them remain: yellow warnings on the candidate page, due to the complex behavior of some co-dependent mui selects, and red warnings on all the pages with maps due to a proptype thing I couldn't fix.

But also, this PR improves the fetching behavior of the GCN event page, to hopefully reduce the number of repetitive API queries. Also fixes some issues with creating scanning profiles, and other minor bug fixes. Last but not least, fixes some of the DatePicker to set/edit users' expiration dates, which right now crashes some pages and makes it impossible to edit those.

One thing to keep exploring, is how to decide whether to hydrate or not. Right now, I have a delay or 1.5 sec where I wait to see if the state was populated from the previous tab's state. If there's no previous tab then the answer is no, and we hydrate. Otherwise we don't. I wonder if this could be done in a cleaner way, like add a callback on the library's action that tries to GET from the previous tab, and hydrate if that fails. Anything that would allow us to remove that 1.5 sec delay would be great. We can try shorter delays, but this seemed robust.


Theodlz:
@mcoughlin really worth running this, to make sure it doesn't break the frontend. Best way to try it is to open a tab, just on the dashboard. From there, navigate on the source table, click on a source which should open a new tab. First of course verify that it loaded correctly, and then inspect->console, and check if there is no `FETCH_...` actions that we normally hydrate, but only source-specific stuff.

mcoughlin (approved):
@Theodlz confirmed that it works nice. Thanks!

Theodlz:
> @Theodlz confirmed that it works nice. Thanks!

Awesome! Thanks for taking a look at it. I'll try it on preview now.

PR #4528, opened 2023-09-14
Favorite notifications: ignore (optional) bot/ml classifications and comments

Theodlz:
Some users complained about getting spammed with a lot of notifications on their favorite objects. This is a first step towards fixing that, by ignoring bot comments and ML classifications.

In a following PR, modifying either the frontend or the backend to deduplicate the notifications that are displayed would help further reduce the spamming.

If there's 2 of the same notification, we can imagine replacing "new comment on..." with "2 new comments on..."

The issue with doing that on the backend, is what about non-frontend notifications (emails, slack, text message,... do we want to avoid new notifications there?) We can't edit an email or text once sent to reflect any change, so we might just want to keep doing the same and de-duplicate on the frontend side of things.

*PS: Also noticed that a .jsx file wasn't used, so I removed it.*


mcoughlin (approved):
LGTM

PR #4529, opened 2023-09-14
GCN Event page: avoid some repetitive API calls

Theodlz:
Avoid repeating some API calls. The tricky thing is that the useEffect is triggered on changes on the event (for good reason), but the fetch methods called in the useEffect modify the event, retriggering it.

To prevent that, we just add some `fetching...` state variables, to prevent starting an API call while one is already running.


mcoughlin (approved):
LGTM

PR #4530, opened 2023-09-15
WIP - Faster sources query

Theodlz:
The source handler right now has really poor performance. The queries generated by sqlalchemy (and how we use it) aren't great.
I've been experimenting with "hand-made" queries, where I just write the raw sql code and use sqlalchemy simply to compile the text query with the parameters and execute it.

I've had great results. This is a script showing what features I've reproduced so far. This is looking promising, and we might be able to just use that in prod if I can get all the feature recreated that way.

Or at least, based on these results, we could rewrite the current code and try to get sqlalchemy to generate better queries. But that might end up being even harder.

This is *absolutely not* a PR to merge (at least for now), but a way for me to keep track of things and share it with others. I run this script locally and in prod to assess performances.

### EDIT 1:

This is looking **really promising**. I reimplemented most of the features we have in the current source handler method to get_sources. **Performance is extremely good even in production**. Queries that used to never return now return very quickly.

### TODOs:
- add the taxonomy in the classification filtering 
- make sure sql injection isn't possible. This is mostly me looking into the tools offered by sqlalchemy or psycopg2 to solve that. Though I already use name parameters.
- Move the code from that script to the proper handler, to see how much of a drag and drop replacement it can be. I expect no major issues there.

### DIFF
--- a/skyportal/source_queries_script.py
+++ b/skyportal/source_queries_script.py
@@ -23,6 +23,8 @@ from skyportal.models import (
     Group,
     Galaxy,
     Allocation,
+    Localization,
+    LocalizationTile,
 )
 
 _, cfg = load_env()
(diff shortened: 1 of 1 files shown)


Theodlz:
Closing this in favor of another PR

PR #4531, opened 2023-09-17
WIP - Faster sources querying

Theodlz:
This rather long PR is an attempt to speedup the sources query as much as possible.
There is another PR on main with the script I developed before I could replace the current handler's method: https://github.com/skyportal/skyportal/pull/4530

The idea here is to merge this on another branch, so I can deploy that on preview and play around with it to assess performance, other then by just running the script, but rather using it along with the frontend.

Things that I still need to implement before we can consider prod:
- sorting by classification, if still useful
- sorting by sourcesingcn status, necessary for sure
- double, triple, if not QUADRUPLE check that the permissions are enforced as expected :)


mcoughlin (approved):
Let's try this in preview.

Theodlz:
> Let's try this in preview.

Thank you good sir

PR #4533, opened 2023-09-17
WIP - Faster sources querying: bugfixes

Theodlz:
Added some things that were missing, fixed one bug:
- page information, used by frontend
- fixed default sorting order and type


Theodlz:
self merging, this is for testing

PR #4534, opened 2023-09-18
WIP - Faster sources querying: Sort by gcn_status, bindparams for type safety.

Theodlz:
This PR adds 2 things: 
- When querying sources in an event, sort them (optionally, does it by default if not sorting specified) by:
  - has not been vetted
  - has been highlighted
  - ambiguous
  - rejected
- passes all parameters using bindparams. This proves to be very challenging for any `IN (1,2,3,4)` kind of query, as the current implementation would cast to an `ARRAY[]` which requires an `ANY()` query instead. Both as pretty much equivalent under the hood, but I did see multiple stackoverflow issues complaining that with an ANY query, an index wasn't being used. Overall, using `IN` seems like a decent option anyway. One thing to note, is that for large sets/arrays for an `IN` query, we might want to add a threshold over which we transform the queries to use a `VALUES ()` statement to speed it up.

I'd like to deploy this to preview, and see if these changes slowdown any of the existing queries.


mcoughlin (approved):
Glad this is testing on preview!

PR #4535, opened 2023-09-18
WIP - Faster sources querying: quicker gcn status queries

Theodlz:
I noticed that when using the "include_sources_in_gcn" flag, which basically does: "sources in the localization OR sources have an entry in the sourcesconfirmedingcn table that is not false".

This `EXISTS OR EXISTS` query is very slow. Given that we constrain sources by first and last detected date (not optional when querying in a localization), and usually query for the group in which the scanners saved the sources, we can afford to just split that query in 2 queries without the sorting -> concatenate and deduplicate the results -> run a basic query to reapply the sorting (which would have been lost if applied to the 2 queries separately.


Theodlz:
self-merging, this is for testing purposes. Still tagging you @mcoughlin so you can keep track of these changes if you want.

PR #4536, opened 2023-09-18
Header: refresh button

Theodlz:
Users took a habit of refreshing the webpage fully to refresh the data displayed, for things as simple as the navbar. 
First, it's unnecessary to overload the server with API calls just for that. In a previous PR I already made sure that new tabs fetch data from existing tabs to hydrate, and not the server. But that does prevent users from updating the data on lets say the dashboard when they already have other tabs open. 
Second, we can already click on "dashboard" again to rehydrate. However that should only re-fetch dashboard data.

In this PR, we:
- make sure clicking on dashboard again only fetches dashboard data
- add a dedicated refresh icon, so users can rehydrate fully if they want. It's located on the header, next to the user profile and notifications.
- on refresh of a tab (via browser, not our own refresh icon), skip hydrating EXCEPT the dashboard data, to always update that.


mcoughlin (approved):
LGTM

PR #4537, opened 2023-09-19
Source notifications: notify on new spectra

Theodlz:
Makes it possible to get notified on new spectra on any source that the user has access to, not just favorite sources.

Also, the user can specify groups for which to restrict the notifications on sources (effectively become notifications on group sources), which will be applied both to notifications on new classifications and new spectra.

This PR also enforces data access restrictions for these 2 notifications. If the user isn't an admin, we make sure to only notify if the classification/spectrum has been shared with user-accessible groups.

TODO:
- slight frontend improvement
- write a unit test

### DIFF
--- a/services/notification_queue/notification_queue.py
+++ b/services/notification_queue/notification_queue.py
@@ -39,6 +39,7 @@ from skyportal.models import (
     Localization,
     ObjAnalysis,
     ObservationPlanRequest,
+    Source,
     Shift,
     ShiftUser,
     Spectrum,
@@ -93,12 +94,15 @@ def notification_resource_type(target):
     if (
         "favorite_sources" not in target["notification_type"]
         and "gcn_events" not in target["notification_type"]
+        and "sources" not in target["notification_type"]
     ):
         return target["notification_type"]
     elif "favorite_sources" in target["notification_type"]:
         return "favorite_sources"
     elif "gcn_events" in target["notification_type"]:
         return "gcn_events"
+    elif "sources" in target["notification_type"]:
+        return "sources"
 
 
 def user_preferences(target, notification_setting, resource_type):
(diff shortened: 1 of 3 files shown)


services/notification_queue/notification_queue.py
```diff
@@ -666,25 +667,44 @@ async def post(self):
...
+                            )
                         elif is_spectra:
                             users = session.scalars(
                                 sa.select(User).where(
-                                    User.preferences["notifications"][
-                                        "favorite_sources"
-                                    ]["active"]
-                                    .astext.cast(sa.Boolean)
-                                    .is_(True)
+                                    sa.or_(
```
mcoughlin, line 680: Is this better as a get with a default?

mcoughlin (approved):
One comment. Otherwise looks nice.

services/notification_queue/notification_queue.py
Theodlz: What do you mean by that?

mcoughlin: This assumes all those keys exist.

Theodlz: I don't think sqlalchemy cares about defaults, I'm not sure it makes a difference. Would be worth looking at the query it generates.
But in this case, where we need the key to be set to a certain value, we do want the user to have all of them anyway.

Maybe your question was just whether or not it yields an error, but we know it doesn't. By default user profiles don't have any key for the notifications altogether, and that hasn't been a problem so far.

Theodlz: Here we're looking into the user class, not an actual user returned by a query, so preferences isn't a real dictionary where looking at a missing key creates an exception.

PR #4538, opened 2023-09-19
Github actions: pin postgres version

Theodlz:
We haven't pinned a specific dockerhub image for postgres. It looks like the latest one (version 16.0) creates some issues.

This is an attempt to pin the latest available image for version 14 (the one we support), to see if that helps.

Of course we can consider fixing it for version 16, but we need to get the Github actions to work again first.


Theodlz:
Seems to be working again!

Theodlz:
Still the same issue (not this new one, older one) with the API tests part 2. Something to solve too if we can.

PR #4540, opened 2023-09-20
WIP - Faster sources handler: use new handler with useExperimental flag

Theodlz:
This PR adds:
- sorting by favorites
- use the new code from https://github.com/skyportal/skyportal/pull/4539 only when setting the `useExperimental` flag to True. This ensures everything still uses the current code, while allowing us to try the new code on production, as suggested by @profjsb.
- Fixes the name of a relationship between 2 tables. No migrations needed, it's just the python-side name that changes here. It's not called anywhere (but required by sqlalchemy), so that won't be a problem.


Theodlz:
@profjsb I tagged you so you know about this PR. I might still merge it today so I can try it in production and compare performance, as its possible that these changes helped on preview but won't help as much in production with a much bigger DB. 
Frontend and backend still use the existing code. It can be activated just using the useExperimental flag.

skyportal/handlers/api/source.py
```diff
@@ -2740,12 +2753,19 @@ class Validator(Schema):
...
         origin = self.get_query_argument('origin', None)
         has_tns_name = self.get_query_argument('hasTNSname', None)
+        has_no_tns_name = self.get_query_argument('hasNoTNSname', None)
         has_been_labelled = self.get_query_argument('hasBeenLabelled', False)
         has_not_been_labelled = self.get_query_argument('hasNotBeenLabelled', False)
         current_user_labeller = self.get_query_argument('currentUserLabeller', False)
         total_matches = self.get_query_argument('totalMatches', None)
         is_token_request = isinstance(self.current_user, Token)
 
+        use_experimental = self.get_query_argument('useExperimental', False)
```
mcoughlin, line 2763: Maybe this name could be a little more explicit?

skyportal/handlers/api/source.py
```diff
@@ -569,6 +572,15 @@ async def get_sources(
     See Source Handler for optional arguments
     """
 
+    if has_no_tns_name:
+        raise NotImplementedError(
+            "hasNoTNSName is not implemented here yet, use experimental handler with useExperimentalSources=True"
```
mcoughlin, line 577: useExperimental vs useExperimentalSources. I prefer the latter.

mcoughlin (approved):
A few small comments.

skyportal/handlers/api/source.py
Theodlz: Ah you are right! Thanks for the suggestion

skyportal/handlers/api/source.py
Theodlz: Yes. This is just for testing for a few days, I don't intend on anyone using that flag other than me while I benchmark, but you're right it won't hurt to change that name

PR #4542, opened 2023-09-20
Gcn summary: bugfix attempt

Theodlz:
Some summaries fail in prod, and I believe it might be because of that line. The error has to do with 'SSL connection closed', so maybe using the session we just created rather than the old sqlalchemy syntax might fix it.


mcoughlin (approved):
LGTM

PR #4543, opened 2023-09-20
Legacy get_sources method: fix default sort order

Theodlz:
fix the default sort order when no sort by is specified.


Theodlz:
self merging

PR #4544, opened 2023-09-22
Sources: don't save if already saved to group_ids, avoid duplicate comments

Theodlz:
This PR adds 3 features:
- (1) When using the POST endpoint to save a source to group(s), cancel saving to some group(s) if they've already been saved as sources to some defined groups. Here's an example: you want to save an obj as a source to group A, but only if it's not already saved to group B. So you pass an extra argument called `ignore_if_in_group_ids` equal to `{group_A_id: [group_B_id]}`.
- (2) When triggering a follow-up request on an object, pass a list of group IDs to which the object needs to be saved as a source already. Example: you want to trigger a follow-up with allocation A, but only if the obj is saved as a source to group A, so you pass `source-group_ids` as an argument, which value will be `[group_A_id]`.
- (3) When posting a comment with the same author, text, attachments, and groups as an existing comment, return 200 with the existing comment id and a message like "comments already exists", without creating a new - duplicated - comment. 

The idea being that kowalski's autosave feature can use (1) to not autosave to group X if a source has for example been saved in Group X junk. And for (2), this will be useful once we have the auto_followup feature running in Kowalski, to prevent triggering followup requests if the candidate hasn't been saved as a source to the group associated to the filter, maybe because of (1). For both autosave and auto followup, we also want to (optionally) post associated comments on the source.

### DIFF
--- a/skyportal/handlers/api/catalog_services.py
+++ b/skyportal/handlers/api/catalog_services.py
@@ -230,7 +230,7 @@ def fetch_transients(allocation_id, user_id, group_ids, payload):
                 if s is None:
                     log(f"Posting {source['id']} as source")
                     source['group_ids'] = group_ids
-                    obj_id = post_source(source, user_id, session)
+                    (obj_id,) = post_source(source, user_id, session)
                     obj_ids.append(obj_id)
 
                 if alert_available:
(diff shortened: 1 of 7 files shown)


mcoughlin (approved):
LGTM

PR #4546, opened 2023-09-27
TNS - various improvements

Theodlz:
- add the missing API Key fields when adding/editing a bot
- always display the TNS name on the sources list
- better sources list display
- display sources TNS names on the gcn summary, instead of the alias field which is always empty anyway.
- if the robot has autoreporters, it will use that as the default in the submission form instead of the user's name


mcoughlin (approved):
LGTM

PR #4547, opened 2023-09-27
More follow-up constraints 

Theodlz:
add more followup constraints (which if met, cancel triggering):
- not_if_classified
- not_if_spectra_exists
This will be used by the auto-followup feature of Kowalski, to avoid triggering if a source is already classified or already has spectra data.

Also, if constraints are met and we don't trigger, still send a 200 instead of an error (as this cancelation has been requested by the user), along with an ignored flag set to true, and a message containing the reason why it wasn't triggered (which constraint was met).

### DIFF
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -1,68 +1,63 @@
-import arrow
 import ast
-from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM. Please fix WIP title before merging.

PR #4570, opened 2023-10-03
GroupUser without stream access: better error message

Theodlz:
Igor wasn't able to add another user to his group, because the user did not have access to one of your group's stream(s).

Stream access control is limited to super admin (for a good reason), so this is something that a user needs to contact a super admin to take care of it. 
The current error message isn't that clear and isn't very helpful in understanding what to do.

This should help.


mcoughlin (approved):
LGTM

PR #4571, opened 2023-10-04
FollowupRequestHandler: return target groups

Theodlz:
add missing target group ids in the response when fetching follow-up request(s), which is useful information in general, but especially when you want to update the request, so you have the list and you do not risk to change it without knowing when using the PUT handler.


mcoughlin (approved):
LGTM

PR #4573, opened 2023-10-06
Follow-up requests GET: more filtering options

Theodlz:
This PR:
- fixes the filtering by start_date and end_date of the requests, and makes it optional by default
- adds filtering by allocation
- adds filtering by requester
- display the created_at date of each request on the followup requests page
- adds sorting on created_at, modified, and obj_id
- sort by created_at desc (newer first) by default on the followup requests page


mcoughlin (approved):
Thanks for fixing this Theo.

PR #4575, opened 2023-10-07
Hydration from existing tabs - allow partial hydration

Theodlz:
If users open multiple tabs one after the other, with the current code, new tabs might think they are fully hydrated when really the previous tab was only partially hydrated.

This should help with that, by better tracking how complete hydration is, and partially hydrating if needed.


mcoughlin (approved):
Looks useful.

PR #4578, opened 2023-10-11
Source: Quick save

Theodlz:
This PR adds a quick save button on the source page. Most scanners save sources to the same groups, and complain that saving a source to the same groups every time is not practical, as they need to go click save -> which opens a dialog window -> they need to go through their list of groups and select them -> confirm. When really, they could just set which groups they want to always save to, and click a "quick save" button instead.

This PR adds that, by adding a section in the user preferences to select the groups to "quick save" to, and the button to do so on the source page.

### DIFF
--- a/static/js/store.js
+++ b/static/js/store.js
@@ -12,10 +12,10 @@ import {
 
 const syncConfig = {
   whitelist: [
-    "baselayer/SHOW_NOTIFICATION",
+    // "baselayer/SHOW_NOTIFICATION",
     "baselayer/HIDE_NOTIFICATION",
     "baselayer/HIDE_NOTIFICATION_BY_TAG",
-    "skyportal/FETCH_INSTRUMENTS_OK",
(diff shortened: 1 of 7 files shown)


static/js/store.js
```diff
@@ -12,10 +12,10 @@ import {
 
 const syncConfig = {
   whitelist: [
-    "baselayer/SHOW_NOTIFICATION",
+    // "baselayer/SHOW_NOTIFICATION",
```
mcoughlin, line 15: Needed?

mcoughlin (approved):
LGTM. We should probably make a short video on this once it goes in.

static/js/store.js
Theodlz: Yes! This is to leverage the new shared tab data stuff. We can also share all the dispatch actions. The issue is, that if tab A had some error notifications, and you open tab B, they appear and don't go away until you click on them. A pain for some users, and not very useful anyway.

PR #4582, opened 2023-10-12
Experimental source queries: fixed pagination datatypes

Theodlz:
Make sure that the pageNumber and NbPerPage, once returned in the query response, are integers and not strings.


mcoughlin (approved):
LGTM

PR #4583, opened 2023-10-14
Telescopes: next day/night time caching

Theodlz:
Cache telescope current time info (next day time, next night time, is it day or night right now), and only refresh the cache when any of these dates are past the current UTC date. This ensures that we don't trigger astroplan's slow calculations every time we hit the telescopes endpoint from the frontend, which currently takes ~7 seconds to return in production. This should get that back to ~100 ms, except of course when one or more telescopes' current time info needs to be updated.


mcoughlin (approved):
LGTM

PR #4584, opened 2023-10-15
WIP - User handler GET: limit info returned to non admin users

Theodlz:
Before this gets merged, I need to test the frontend well to make sure it doesn't break anything there.

### DIFF
--- a/services/health_monitor/health_monitor.py
+++ b/services/health_monitor/health_monitor.py
@@ -30,7 +30,7 @@ class DownStatus:
 def migrated():
     try:
         r = requests.get(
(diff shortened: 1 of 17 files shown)


Theodlz:
@mcoughlin I'm thinking we might want to not share personal info either (email and phone number) unless:
- the user querying is a super admin
- I think it's fine to provide email for the users that are in groups the user that is querying has access to.

What do you think? Just trying to think what's the best logic when it comes to sharing user data. Might be very important if we really want to consider public facing pages at some point.

mcoughlin (approved):
Yes I agree that less is more here. Just enough info for folks to tag.

PR #4586, opened 2023-10-17
SEDM interactions: handle new status message

Theodlz:
Handle the cases where a user tries to edit or delete a request that the instrument is already processing.


mcoughlin (approved):
LGTM

PR #4588, opened 2023-10-18
Follow-up requests GET: pagination info

Theodlz:
It looks like the handler currently implements pagination, but does not return pagination info.

It does not "break" anything, but makes it confusing if you are using it without knowing about the pagination.


mcoughlin (approved):
Thanks for dealing with this so quickly.

PR #4589, opened 2023-10-19
Frontend: Follow-up request vs FP

Theodlz:
This FP is the second in a series of 2 PRs to differentiate Follow-up requests from Forced photometry request.


mcoughlin (approved):
2 heads are clearly better than 1 in this case :)

mcoughlin:
@Theodlz I have tried this locally and it works intuitively.

Theodlz:
> @Theodlz I have tried this locally and it works intuitively.

Great! Thanks a lot for testing it out. I'll clean up some stuff here and there and deploy soon

PR #4590, opened 2023-10-24
Allocations: types

Theodlz:
This PR adds types to allocations.

This will be useful to simply another PR: https://github.com/skyportal/skyportal/pull/4589 (and avoid some complex if else tricks to get it to work) as well as future PRs.


mcoughlin (approved):
LGTM

PR #4592, opened 2023-10-27
Analysis Frontend: update list after adding,deleting, and updating

Theodlz:
Currently, the list of analyses isn't responsive at all to any changes. This PR ensures that:
- we refresh the list to show a new analysis when a user submits one
- we refresh the list correctly after deleting an analysis
- we refresh the list when an analysis gets updated (as completed or failed) through the associated webhook

This should help prevent users from refreshing the whole page every time they perform any actions there, or when the results are uploaded back to SkyPortal.


mcoughlin (approved):
LGTM

PR #4596, opened 2023-10-31
Implement search feature for dropdown menus in NewObservations.jsx

wkw8402:


Theodlz (changes requested):
Hi @wkw8402, this works very well!

However, I believe that part of this frontend overall we are looking for is to package this kind of search bars as their own reusable component. There are multiple pages that also need that instruments searchbar, here are few of them for reference:
- GCN event page, analysis section
- Observing runs page (form to add a new run)
- Allocation page (almost all the forms there, visible on the right by admins or those with the manage allocations ACL).
- Instruments page (form to modify an instrument)

I think that identifying what pages all use the same thing to simplify a lot of bits of code by providing one reusable component should be the priority here. It will make your work orders of magnitude easier as time passes, while making the codebase much cleaner and maintainable.

Some suggestions on how to do that:
- make a generic search bar component, customizable as much as possible (or at least, as much as necessary. One thing for sure, is that this needs to take an optional "serverSide" boolean along with optional methods to iterate through the pagination of the API endpoint if any. This will be super useful to deduplicate a lot of code here and there where the full list isn't returned in one go, and querying the backend is needed. You can see an example of that in: FindGcnEvents.jsx (which to be clear, is not perfect and would benefit from such a logic).
- then, you'll be able to reuse that component to create N number of components for whatever needs to be searched.
- that, or if generic enough to work for most use cases, the generic search bar can directly be called where needed instead of creating N number of components. You might still need to create a few dedicated components when some more complex operations are needed. Your choice!

Again this already good, but I think the idea isn't just to make the frontend better for the users, but better for you to develop, which in term means for the users as well.

Theodlz:
Hi @wkw8402, just wanted to know how things were going, and if you needed to discuss any of that?

wkw8402:
Hi @Theodlz! Thank you for your comment and feedback! I have tried to make a generic dropdown with search bar component named "NewDropDownSearchBar.jsx" which takes in 5 parameters:

1.  List of items in dropdown menu
2. Currently selected item (state var)
3. Function to handle 3, the selected item change
4. Currently typed input on search bar (state var)
5. Function to handle 5, the search input change

and then imported this new component into NewObservation.jsx. Could you please review the changes and provide feedback?

Theodlz:
Hi @wkw8402,

Thanks a lot for the changes! Before I review, make sure that you pass the linter tests. As you can see, all the github actions are bright red. Clicking on any of them you can easily see why.

We use several linters through `pre-commit` to standardize how the code looks and is formatted. 
So just go back to your terminal window where you usually do your git commit -> make sure you're also in your python env -> `pip install pre-commit` -> `pre-commit install`. 

When that's done, anytime you try to commit some changes to files you will see it run and give your the errors you have to address before you can commit. Here because the files are already committed (so there's no pending changes), you'll have to specify it while files to look at. For that you can just do `pre-commit run --all-files`. 

See the documentation here if needed: https://pre-commit.com/. We also detail that here in our docs: https://pre-commit.com/

Please share that with the rest of the team as well!

Tag me here when you're done and I'll re-review. Looks like you are mostly just missing the propTypes.

PR #4620, opened 2023-11-03
Follow-up requests: various improvements

Theodlz:
This PR adds various improvements to the follow-up requests:
- makes sure they all refresh the frontend on add/modify/delete
- optionally take 2 arguments to change the refresh behavior: refresh_source and refresh_requests. Useful to refresh the associate source page, and/or the full list of requests if one is on the follow-up requests page.
- fix the PS1 forced photometry which was broken.
- improves the follow-up request table download on the follow-up request page, with a nice downloader like the sources page, and adds a little more info than what is seen on the frontend table.
- Clarifies what the buttons under the follow-up request query form are for (scheduling with Astroplan)

This PR also adds a new source annotations query handler: PS1QueryHandler. This can be used from the front end as well with a dedicated button and will fetch the closest PS1 source within 2 arcseconds.

Also, this PR fixes all the broken tests found in flaky (which I believe haven't worked properly in a while, due to frontend changes and dependencies upgrades). Essential to make sure that all of the other changes do not break anything. To be clear, the tests can now pass, but that doesn't mean they are not flaky. Locally, I do not encounter **any** issues running them, but it seems to not work as well when running on GitHub. Something to look into.

TODO in a following PR:
- fix that scheduling thing under the form. There are errors with the code there.

Here's a little GIF to show the downloader thing in action, just for reference:


mcoughlin (approved):
LGTM

Theodlz:
> LGTM

We need to publish a paper about how @mcoughlin can review faster than the speed of light.

Theodlz:
Well as this hasn't been complicated enough with the nodejs breaking changes + dependabot, GitHub is having some connectivity issues right now, so the tests are failing... Will have to wait until this is resolved. Yikes.

@nabeelre

Theodlz:
It is GREEN

PR #4622, opened 2023-11-03
Ingest observations in batches

Theodlz:
Quick PR while another PR's tests are running :)

Monitoring production, we can see a large peak in DB load around 12 p.m. daily, which is exactly when we insert ZTF observations from the night before using the recurring_api service.

The existing method runs a lot of queries, some duplicated, some not. This is an attempt to deduplicate and run these queries on a batch of items at once (100, but we could try more).

Then, when inserting, this is also an attempt to ingest not everything at once but in smaller batches (1000).

TODO in a following PR:
There is **no pagination whatsoever** on the observations page. This overloads the backend and the frontend, so we need to fix that. Will help reduce the number of reads we run on that table, which might also help speed up the ingestion as well.

*PS: This is labelled as WIP because I still need to try it locally.*


mcoughlin (approved):
Neat idea.

PR #4629, opened 2023-11-10
Follow-up request constraints: don't trigger if saved to group(s)

Theodlz:
In the context of BTSBot (but it could very well bu useful for other applications), we want Kowalski to be able to send a request without having access to more info about the source. However, we want to cancel triggering if the source is saved to the junk groups. This offloads the check to Fritz, like we already do with other constraints.


mcoughlin (approved):
LGTM

PR #4630, opened 2023-11-10
Download/Preview attachments: bugfix

Theodlz:
Recently folks have been uploading attachments that are png screenshots. However, looks like their system includes some non standard space chars, which can't be encoded with the default latin-1 that tornado uses.

To get around that, and prevent any other weird names from breaking the download process, we make sure that non-ascii characters are all replaced by simple spaces. We can imagine something a little smarter, but having a first bugfix is important.

Right now that failure leads to the "system provisioning" error for the client for some reason, preventing any other API call from running for a few seconds.


mcoughlin (approved):
LGTM

PR #4636, opened 2023-11-18
external_photometry method - different behaviors regarding duplicates

Theodlz:
This PR extends on the current add_external_photomety method to make sure that we can either:
- as we do currently, error when any new rows already exist, and not even add the non-duplicates
- ignore duplicates, and only add new ones
- update duplicates if needed, and then add new ones

I made it so that the update is the default behavior, as I believe most places in the code that do call that method want to be doing that, but we could keep it to "error" by default, and specify that we allow an "update" where needed.

### DIFF
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -153,6 +153,7 @@ def service(queue):
                                         instrument.id,
                                         followup_request.requester.id,
                                         parent_session=session,
+                                        duplicates="update",
                                     )
                                     req.status = 'complete'
                                     session.add(req)
@@ -220,7 +221,12 @@ def service(queue):
                             ),
                         )
 
-                        if response.status_code == 200:
+                        if "Zero records returned" in str(response.text):
+                            log(
+                                "Found no records yet for this ZTF forced photometry account."
+                            )
+                            continue
+                        elif response.status_code == 200:
                             df_result = pd.read_html(response.text)[0]
                             df_result.rename(
                                 inplace=True,
@@ -300,13 +306,17 @@ def service(queue):
                                         instrument.id,
                                         followup_request.requester.id,
                                         parent_session=session,
+                                        duplicates="update",
                                     )
                                     req.status = 'complete'
                                     session.add(req)
                                     session.commit()
                                     log(f"Job with ID {req.id} completed")
-                                except Exception:
-                                    status = 'In progress: Light curve not yet available. Waiting for it to complete.'
+                                except Exception as e:
+                                    if 'Failed to commit photometry' in str(e):
+                                        status = f'error: {str(e)}'
+                                    else:
+                                        status = 'In progress: Light curve not yet available. Waiting for it to complete.'
                                     if followup_request.status != status:
                                         followup_request.status = status
                                         session.add(followup_request)
--- a/skyportal/facility_apis/atlas.py
+++ b/skyportal/facility_apis/atlas.py
@@ -99,7 +99,6 @@ def commit_photometry(
         DBSession,
         FollowupRequest,
         Instrument,
-        User,
     )
 
     if parent_session is None:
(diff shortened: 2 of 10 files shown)


Theodlz:
Let me know what you think @mcoughlin

services/facility_queue/facility_queue.py
```diff
@@ -220,7 +221,12 @@ def service(queue):
                             ),
                         )
 
-                        if response.status_code == 200:
+                        if "Zero records returned" in str(response.text):
```
mcoughlin, line 224: Should status also be updated here?

skyportal/facility_apis/atlas.py
```diff
@@ -196,7 +202,11 @@ def commit_photometry(
         from skyportal.handlers.api.photometry import add_external_photometry
 
         if len(df.index) > 0:
-            add_external_photometry(data_out, request.requester)
+            ids, _ = add_external_photometry(
+                data_out, request.requester, duplicates=duplicates
+            )
+            if ids is None:
+                raise ValueError('Failed to commit photometry')
```
mcoughlin, line 209: Similarly, should this be a status?

skyportal/handlers/api/photometry.py
```diff
@@ -872,12 +877,112 @@ def add_external_photometry(json, user, parent_session=None):
...
+                        for id in stream_ids_update:
+                            session.add(
+                                StreamPhotometry(
+                                    photometr_id=duplicate.id, stream_id=id
+                                )
+                            )
+                        log(
+                            f'Adding streams {stream_ids_update} to photometry {duplicate.id}'
+                        )
+
```
mcoughlin, line 944: Should we also log here at the end that photometry did not need updating?

mcoughlin (approved):
Just some questions about status vs logging vs raises. Just wondering if more of those statuses should also be put in the request so the users see it on the frontend (or not!).

skyportal/handlers/api/photometry.py
Theodlz: Might end up with a **lot** of logs that way. I think it's fine without it. Any specific reason why you'd like to see that info?
I think what could be ok is keeping track of which were not updated in a list, and then posting just one single log with that list.

mcoughlin: Ok that seems good. Just want to simplify debugging if needed!

skyportal/facility_apis/atlas.py
Theodlz: I think it's good to raise that error, and let the method upstream of that decide what status should be set when failing to add the photometry. For some application you might want to just set it as failed and move on, for others you might want to retry again.

services/facility_queue/facility_queue.py
Theodlz: No, this happens if your FP account has nothing in it yet. Which always happens if you dont send any requests for a bit to IPAC, and then start sending stuff. It's something to ignore, just their webpage showing a string instead of an empty list...

skyportal/facility_apis/atlas.py
Theodlz: The queue will use that to set the status.

mcoughlin: ok!

services/facility_queue/facility_queue.py
mcoughlin: Makes sense.

skyportal/handlers/api/photometry.py
Theodlz: Agreed! I'll add that one line log thing now.

Theodlz: Just added something. Let me know if that's what you expected.

mcoughlin: Yes ok!

Theodlz:
@mcoughlin just for bookkeeping (and so I'm not the only one to know about this change), I also pushed a few lines that make the photometry data table a little better. I have one more commit that I want to add (some points show the full list of all groups and users, we want to fix that it looks horrendous), and then I'll ask for another review.

mcoughlin (approved):
Excited to try this! Thanks @Theodlz !

Issue #4633, opened 2023-11-15
add acknowlegments to GCNs


tahumada:
For ZTF, it would be ideal for the GCN summary to include the usual acknowledgement:

ZTF and GROWTH are worldwide collaborations comprising Caltech, USA; IPAC, USA; WIS, Israel; OKC, Sweden; JSI/UMd, USA; U Washington, USA; DESY, Germany; MOST, Taiwan; UW Milwaukee, USA; LANL USA; Tokyo Tech, Japan; IITB, India; IIA, India; LJMU, UK; TTU, USA; SDSU, USA and USyd, Australia. ZTF acknowledges the generous support of the NSF under AST MSIP Grant No 1440341. GROWTH acknowledges generous support of the NSF under PIRE Grant No 1545949. Alert distribution service provided by DIRAC@UW (Patterson et al. 2019). Alert database searches are done by AMPEL (Nordin et al. 2019), Kowalski (Duev et al. 2019), and SkyPortal (van der Walt et al. 2019; Coughlin et al. 2023). GROWTH India telescope is located at the Indian Astronomical Observatory (Hanle), operated by the Indian Institute of Astrophysics (IIA). GROWTH-India project is supported by SERB and administered by IUSSTF, under grant number IUSSTF/PIRE Program/GROWTH/2015-16 and IUCAA.  The Fritz and SkyPortal projects acknowledge the generous support of The Gordon and Betty Moore Foundation.

profjsb:
slightly edited above, including the SkyPortal refs.

Theodlz:
Wondering what's the best way to deal with these acknowledgments. I think it makes sense to have these at the group level, and you can pick which one to use while generating the summary. Either a many to one (a list of possible acknowledgements added to a group) or one to one relationship (just one acknowledgment text by group).

Could also be at the instance level, but that's not very practical if let's say DECAM wants their own version which is different from this one (I know some of them used the GCN summary we have here).

mcoughlin:
@Theodlz what about the frontend form allowing you to choose from the templates in the config? So you will choose from ZTF or DECam style when creating the summary.

Theodlz:
@mcoughlin sounds reasonable. Just wondering if it's not easier to have that outside of the config, so we can make edits easily. However, I'm really ok with a first PR that uses the config. That would be very quick to implement while we think about something a little more flexible for the users.

mcoughlin:
@Theodlz I trust your intuition here!

[...]

PR #4638, opened 2023-11-20
GCN summary - acknowledgments in config

Theodlz:
Takes care of https://github.com/skyportal/skyportal/issues/4633


mcoughlin (approved):
Thanks for doing this. Code looks good.

PR #4639, opened 2023-11-27
NewDropDownSearchBar.jsx created and NewObservation.jsx modified

wkw8402:
Generic component for search bar inside dropdown menu created with name: NewDropDownSearchBar.jsx.
Tested one dropdown menu for instruments in Observations page and modified the file: NewObservation.jsx.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@rjsf/validator-ajv8": "5.5.0",
     "autosuggest-highlight": "3.3.4",
     "buffer": "6.0.3",
-    "check-dependencies": "^1.1.0",
+    "check-dependencies": "1.1.0",
     "choices.js": "10.2.0",
     "clsx": "1.2.1",
     "convert-css-length": "2.0.1",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@rjsf/validator-ajv8": "5.5.0",
     "autosuggest-highlight": "3.3.4",
     "buffer": "6.0.3",
-    "check-dependencies": "1.1.0",
+    "check-dependencies": "^1.1.0",
     "choices.js": "10.2.0",
     "clsx": "1.2.1",
     "convert-css-length": "2.0.1",
--- a/static/js/components/NewDropDownSearchBar.jsx
+++ b/static/js/components/NewDropDownSearchBar.jsx
@@ -1,0 +1,55 @@
+// NewDropDownsearchBar.jsx
+import React from "react";
+import PropTypes from "prop-types"; // Import PropTypes for prop validation
+import Autocomplete from "@mui/material/Autocomplete";
+import TextField from "@mui/material/TextField";
+
+const NewDropDownSearchBar = ({
+  instrumentList,
+  selectedInstrument,
+  onInstrumentChange,
+  searchValue,
+  onSearchChange,
+}) => (
+  <Autocomplete
+    options={instrumentList}
+    getOptionLabel={(instrument) =>
+      `${instrument.telescopeName} / ${instrument.name}`
+    }
+    value={selectedInstrument}
+    onChange={onInstrumentChange}
+    renderInput={(params) => (
+      <TextField
+        {...params}
+        label="Select an instrument"
+        variant="outlined"
+        onChange={(e) => onSearchChange(e.target.value)}
+        value={searchValue}
+      />
+    )}
+  />
+);
+
+// Define prop types for the InstrumentSearchBar component
+NewDropDownSearchBar.propTypes = {
+  instrumentList: PropTypes.arrayOf(
+    PropTypes.shape({
+      telescopeName: PropTypes.string.isRequired,
+      name: PropTypes.string.isRequired,
+      // Add other properties as needed
+    })
+  ).isRequired,
+  selectedInstrument: PropTypes.shape({
+    id: PropTypes.number.isRequired,
+    // Add other properties as needed
+  }),
+  onInstrumentChange: PropTypes.func.isRequired,
+  searchValue: PropTypes.string.isRequired,
+  onSearchChange: PropTypes.func.isRequired,
+};
+
+NewDropDownSearchBar.defaultProps = {
+  selectedInstrument: null,
+};
+
+export default NewDropDownSearchBar;
--- a/static/js/components/NewDropDownSearchBar.jsx
+++ b/static/js/components/NewDropDownSearchBar.jsx
@@ -1,27 +1,26 @@
-// NewDropDownsearchBar.jsx
+// NewDropDownSearchBar.jsx
 import React from "react";
-import PropTypes from "prop-types"; // Import PropTypes for prop validation
+import PropTypes from "prop-types";
 import Autocomplete from "@mui/material/Autocomplete";
 import TextField from "@mui/material/TextField";
 
 const NewDropDownSearchBar = ({
-  instrumentList,
-  selectedInstrument,
-  onInstrumentChange,
+  optionsList,
+  selectedOption,
+  onOptionChange,
   searchValue,
   onSearchChange,
+  label,
 }) => (
   <Autocomplete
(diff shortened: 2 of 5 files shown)


Theodlz:
please @pronav5000 update your PR's branch so the github actions can run (I just fixed them)

static/js/components/NewDropDownSearchBar.jsx
```diff
@@ -0,0 +1,55 @@
+// NewDropDownsearchBar.jsx
+import React from "react";
+import PropTypes from "prop-types"; // Import PropTypes for prop validation
+import Autocomplete from "@mui/material/Autocomplete";
+import TextField from "@mui/material/TextField";
+
+const NewDropDownSearchBar = ({
```
Theodlz, line 7: As discussed on the PR that was closed, can you make that component generic, i.e. not call the parameters instrumentList and selectInstrument, but rather something like optionsList and selectOption? As well as make the label of the TextField a parameter and not hardcode it to "Select an instrument".

Once you are done with that, and it works as expected, happy to merge.

Theodlz (changes requested):

package.json
```diff
@@ -27,7 +27,7 @@
     "@rjsf/validator-ajv8": "5.5.0",
     "autosuggest-highlight": "3.3.4",
     "buffer": "6.0.3",
-    "check-dependencies": "1.1.0",
+    "check-dependencies": "^1.1.0",
```
Theodlz, line 30: Also, please remove that change from the PR. node makes that change when you install dependencies, but it's not actually needed.

Theodlz (changes requested):

static/js/components/NewDropDownSearchBar.jsx
```diff
@@ -0,0 +1,54 @@
+// NewDropDownSearchBar.jsx
```
Theodlz, line 1: Maybe a more generic name here? It's basically a `SearchableDropDown`, not necessarily so select anything `New` as this name suggests.

static/js/components/NewObservation.jsx
```diff
@@ -34,29 +54,26 @@ const NewObservation = () => {
...
 
+  // Render the observation form with the NewDropDownSearchBar component
   return (
     <Form
       schema={observationFormSchema}
       validator={validator}
       onSubmit={handleSubmit}
-    />
+    >
+      <NewDropDownSearchBar
```
Theodlz, line 68: Please move that outside of the form (like above it), it does not do anything to embed it in the form. There is actually a proper way to embed components in a rjsf fromSchema, but not worth doing that here. When you move it up, make sure that you also add a little space (where this NewObservation component is called) between the component and the `Add new observations` typography that's above it.

static/js/components/NewDropDownSearchBar.jsx
```diff
@@ -0,0 +1,54 @@
...
+  optionsList,
+  selectedOption,
+  onOptionChange,
+  searchValue,
+  onSearchChange,
+  label,
+}) => (
+  <Autocomplete
+    options={optionsList}
+    getOptionLabel={(option) => `${option.telescopeName} / ${option.name}`}
```
Theodlz, line 17: The component can't be generic if hardcoded to display this text specific to the telescopes. Also, I currently see this because of it:
image

because `telescopeName` doesn't not exist in the instrumentList you passed as a parameter. 

What you want, is to build the right option list before passing it to this component.

Here for instance, instead of passing the plain instrumentList, you can create an instrumentOptionList array, where each element has: an id, a label.

Does that make sense? You want the component to take a generic list of options with an id and a label you build before hand.

Here, to add the telescopeName, in the NewObservations.jsx you'll need to either get the telescopeList to and find the match for each by id, or check if the instrument contains a telescope object which does have the name. A good old console.log will do the trick :)

Theodlz (changes requested):
Much better! But still some things are hardcoded in the generic component, and even table a variable that does not exist! I added a lot of instructions, it should be easy to fix.

static/js/components/NewObservation.jsx
wkw8402: I moved SearchableDropdown component outside of the form. Could you clarify how I should add a little space between the NewObservation component and the typography in the image below?

image

static/js/components/NewDropDownSearchBar.jsx
wkw8402: Thank you for the feedback!

static/js/components/NewDropDownSearchBar.jsx
wkw8402: I am not 100% sure but I am finding this failure on other pages such as Allocations page as shown below where I have not modified the original codebase. Do you think I am not successfully loading instrument lists inside my environment?

image

Theodlz: What do you mean by not loading successfully? In your frontend code, you call your newly added component and pass it a list of options. In your component, as I mentioned in my review, you are looking for a `telescopeName` that doesnt even exist. 

Locally, with the code on main, that does not happen at all. That is introduces by your changes.

As I mentioned in the last review, you want to not just pass `instrumentList` that you just fetched from the redux with `useSelector`, but you want to create a new list which element can have the instrument id, and name/label. But you need to build that label there and not hardcode that you need instrument name in the component you created, which effectively does not make it generic at all if you build the name in it.

and to get the telescopeName, well as it is already done in many places of the code you get the list of telescopes from the redux with `useSelector` and you can build a simple mapper from the telescope id to its name, so you can easily figure out what's the name of the telescope associated to an instrument.

Doest that make sense?

Theodlz: However, I do not understand why you would see that on existing pages where you do not use your component. There could be another problem on top of the one I mentioned in my latest review. Please check in the console if the list of telescopes in the redux is there or missing.

static/js/components/NewObservation.jsx
```diff
@@ -1,30 +1,50 @@
...
 import Form from "@rjsf/mui";
 import validator from "@rjsf/validator-ajv8";
 import dataUriToBuffer from "data-uri-to-buffer";
 import { showNotification } from "baselayer/components/Notifications";
 import { fetchObservations, uploadObservations } from "../ducks/observations";
+import SearchableDropDown from "./SearchableDropDown";
 
 const NewObservation = () => {
   const { instrumentList } = useSelector((state) => state.instruments);
-  const { telescopeList } = useSelector((state) => state.telescopes);
```
Theodlz, line 12: Look, its here. We grabbed the telescopeList to build that mapper and get the name of the telescope associated to each instrument, but you removed it.

static/js/components/NewObservation.jsx
```diff
@@ -34,29 +54,28 @@ const NewObservation = () => {
         title: "Observation file",
         description: "Observation file",
       },
-      instrument_id: {
-        type: "integer",
-        oneOf: instrumentList.map((instrument) => ({
-          enum: [instrument.id],
-          title: `${
-            telescopeList.find(
```
Theodlz, line 42: And this is where we find the instrument name, which you replaced by a variable called `telescopeName` that doesnt exist at all in your generic component (again, you have to build the option list and then pass it, and not hardcode the label in the generic component that defeats its purpose).

Theodlz (changes requested):
I highlighted some of the things you deleted that you should not have

Theodlz:
@wkw8402  any progress?

wkw8402:
Hi @Theodlz ! Hope you've been doing well 🙂

I wanted to revisit this PR and see what you think about potentially picking it back up. I recently restored the branch and thought it might be a nice side project to wrap up and contribute again to SkyPortal.

Do you think this dropdown component would still be useful? Or is there another UI improvement you think might be particularly more helpful to work on instead?

Either way, would love to hear your thoughts! Thanks as always 🙏

PR #4642, opened 2023-11-29
GA bugfix: pycurl system dependencies

Theodlz:
Looks like the GAs are not running because we're not managing to install a few system dependencies we need for pycurl.

Opening a PR just so I can run said actions and try a few potential bugfixes.


mcoughlin (approved):
LGTM. Just delete WIP.

PR #4643, opened 2023-11-29
WIP - Revert new sqlalchemy syntax when creating sessions

Theodlz:
Looks like this new syntax is causing us trouble again.

Unfortunately, the PS1 API is down right now, so I can't test these changes properly. I will have to wait until it's back up to verify everything. The way we create the sessions recently (with the bindings to DBSession instead of directly using DBSession after import) is the only different between places where it works and places where it doesn't work, so I'd like to try to revert that and test it on preview.

### DIFF
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -151,7 +151,6 @@ def service(queue):
                                         altdata,
                                         followup_request.id,
                                         instrument.id,
-                                        followup_request.requester.id,
                                         parent_session=session,
                                         duplicates="update",
                                     )
@@ -304,7 +303,6 @@ def service(queue):
                                         altdata,
                                         followup_request.id,
                                         instrument.id,
-                                        followup_request.requester.id,
                                         parent_session=session,
                                         duplicates="update",
                                     )
--- a/skyportal/facility_apis/atlas.py
+++ b/skyportal/facility_apis/atlas.py
@@ -72,7 +72,6 @@ def commit_photometry(
     altdata,
     request_id,
     instrument_id,
-    user_id,
     parent_session=None,
     duplicates="error",
 ):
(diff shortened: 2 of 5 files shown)


mcoughlin:
Is there a workflow where we periodically restart idle apps to clear weirdness?

Theodlz:
> Is there a workflow where we periodically restart idle apps to clear weirdness?

Yes, but:
- we need to assess app weirdness and that's not straightforward
- that still means we have some bad code somewhere, but we are still ignoring it and it will still cause issues between the time the problem occurs and when we restart the app.


We should just make code changes to fix this problem I think. But yeah, clearly something new since we changed the sqlalchemy syntax with that new thing to create and close sessions (which I'm not sure we even need).

Theodlz:
@mcoughlin!

```
Just one time, somewhere in your application’s global scope. It should be looked upon as part of your application’s configuration. If your application has three .py files in a package, you could, for example, place the [sessionmaker](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.sessionmaker) line in your __init__.py file; from that point on your other modules say “from mypackage import Session”. That way, everyone else just uses [Session()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session), and the configuration of that session is controlled by that central point.

If your application starts up, does imports, but does not know what database it’s going to be connecting to, you can bind the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) at the “class” level to the engine later on, using [sessionmaker.configure()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.sessionmaker.configure).

In the examples in this section, we will frequently show the [sessionmaker](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.sessionmaker) being created right above the line where we actually invoke [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session). But that’s just for example’s sake! In reality, the [sessionmaker](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.sessionmaker) would be somewhere at the module level. The calls to instantiate [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) would then be placed at the point in the application where database conversations begin.
```

so yes, it's just for their examples, apparently we should **never** use session maker in multiple places, but only once, which is what the `DBSession` object we import does! According to this, the syntax we had before was the right one!

Theodlz:
Just tried that. Let's see how the tests go, and I'll play around with the app locally later today

Theodlz:
Encountering other issues after these changes. To discuss on Tuesday's call.

Issue #3532, opened 2022-09-30
Source page collapse functionality


jadalilleboe:
Make the source page collapse functionality use standard collapsing mechanisms (i.e. an arrow to press instead of a button, and then a sliver of the right pane should be visible for expanding again)
![2022-09-30_10_06_14 (1)](https://user-images.githubusercontent.com/82007190/193334982-7d8a297e-e709-41de-bc52-318a1c16e805.png)

PR #4664, opened 2023-12-01
Fixed Issue #3532 - created drawer for collapsable right pane on the source page

syh379:
For issue #3532, in the SourceDesktop file, I replaced the original collapse component for the right pane with the MUI drawer component so the right pane transitions in and out more smoothly. When the drawer is closed, a sliver of it is shown. The "show/hide right pane" button is replaced with arrows indicating directions. One thing that is a bit different from using the collapse component is that the updated right pane has its own scroll bar separate from the scroll bar for the main content.  
image
image

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -442,13 +442,13 @@ server:
   # From https://console.developers.google.com/
   #
   # - Create Client ID
-  # - Javascript origins: https://localhost:8000
-  # - Authorized redirect URLs: http://localhost:8000/complete/google-oauth2/
+  # - Javascript origins: https://localhost:5000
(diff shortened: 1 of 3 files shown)


Theodlz:
This looks good on the screenshots! Good work.

Some of the unit tests are not passing. Did you try those locally to assess whether or not your changes broke anything?

syh379:
Hi, I have been trying to run the tests locally but I'm having a lot of trouble loading them from my computer. I double checked the SourceDesktop file I edited and reran the website with test data, and I still can't seem to figure out where things might have been messed up.

Theodlz:
@syh379 being able to run the tests is essential. Can you walk me through the issues you are facing so we can fix those?

PR #4667, opened 2023-12-06
Remove duplicate timestamps from analysis photometry

### LINKED ISSUE #4666: Drop duplicate photometry points before running analyses
SkyPortal currently sends all of a source's photometry points to analysis services, which leads to duplicate timestamps (e.g. from alert and forced photometry). A reasonable solution is to drop all duplicate photometric timestamps before passing the data to the analysis service.

bfhealy:
This PR removes duplicate-timestamp photometry points before passing the remaining data to an analysis service. The first of each set of duplicate points is retained.

Resolves #4666


bfhealy:
@Theodlz Correct - the initial PR only considered mjd when dropping duplicate points. I agree it's worth considering the filter as well to avoid the potential scenario you mentioned, and I've pushed a new commit to do that.

Theodlz:
Closing this issue for now as it was solved. If we want some fancier deduplication later on, we'll simpy open a dedicated one.

Theodlz:
Oh crap I thought this was the issue! Reopening

Theodlz (approved):
LGTM

PR #4668, opened 2023-12-06
Add commented NMMA analysis to db_demo.yaml

bfhealy:
This PR adds the NMMA analysis service to `db_demo.yaml` to facilitate easier local testing.

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -1295,20 +1295,20 @@ analysis_service:
     group_ids:
       - =program_A
       - =program_B
-  - name: "NMMA_Analysis"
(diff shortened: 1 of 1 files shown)


Theodlz:
I think it's fine. Very nice to have that live here and not have to search it in an old PR every time you do a db_clear.

The only thing is, I think it might create confusion to add a reference to a service which doesn't actually run by default on the machine. Maybe we do add it but as comment? @mcoughlin

mcoughlin:
@Theodlz Commented out is fine.

bfhealy:
@Theodlz That sounds good to me too. I'll push the change.

PR #4669, opened 2023-12-07
Analysis - Input Filters

Theodlz:
This PR adds the necessary code to specify an extra layer of filtering to the input data of an analysis request before sending it to the service.

The first application of this system is constraining what photometry data (per instrument and/or per filter) gets sent to any service relying on it.

On the frontend, we report/display the input data filters used, for bookkeeping.


Theodlz:
@mcoughlin @bfhealy

PR #4671, opened 2023-12-07
GA: labeler v5 update

Theodlz:
In our GitHub actions, we're not specifying a version of the labeler to use. They recently released a new version which seems to break ours.

This is an attempt to update to that newer syntax, which honestly isn't as clear as the previous version in my opinion.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -2,54 +2,65 @@
 
 # Add 'config-change' label to any change to default config files
 config-change:
-  - "config.yaml.defaults"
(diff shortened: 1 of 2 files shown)


Theodlz:
Many issues for many people across git. Looks like their last release might just not work, or the documentation is incorrect. Not sure

Theodlz:
https://github.com/actions/labeler/pull/669/files 

and looks like we might not see the changes in action before we merge

mcoughlin (approved):
LGTM

PR #4673, opened 2023-12-10
Multi-threaded microservices: thread health check

Theodlz:
We had an issue in production where the thumbnails queue kept filling up and didn't seem to get processed at all, as if the API running on one thread was doing just fine, but the queue processing running on the other wasn't there anymore. Nothing in said queue processing service seems to be able to just get stuck like that, so it's likely that the thread was just not alive anymore for some reason. The app had been up for more than a week which rarely happens, so that might be the reason.

This should check if the threads are still up and restart them if necessary.

### DIFF
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -415,6 +415,14 @@ if __name__ == "__main__":
         while True:
             log(f"Current facility queue length: {len(queue)}")
             time.sleep(120)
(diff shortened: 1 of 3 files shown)


Theodlz:
still need to fix this stupid labeler action... their update just breaks everything yikes. I followed their new syntax but that didn't fix the problem we're having...

mcoughlin (approved):
LGTM

PR #4674, opened 2023-12-10
Labeler action: fix version number

Theodlz:
Numerous issues with the version of the action that's on its main branch. Following the new documentation to fix breaking changes still leaves us with an issue that prevents anything from running. This is an attempt to just pin the previous (working) version while we figure out the v5 or while it gets fixed.

EDIT: The version number I pinned is incorrect. Let's try to fix that first to see if the v5 actually works now


Theodlz:
self merging

PR #4675, opened 2023-12-12
Finderchart - PS1 fallback

Theodlz:
Right now, the fallback for the finder chart (when the requested image_source does not work) points to a service likely hosted on the same server/network as the default image_source: ps1.

The result is that often when the ps1 API is acting up, that fallback doesnt seem to be available either.

Based on user feedback and suggestion, this changes the fallback to another service for PS1 cutouts hosted at CDS Starsbourg.

I tested it on a few targets, and I got identical images to the ones from the original PS1 service.

We might want to consider implementing a similar fallback for the cutouts?


mcoughlin (approved):
LGTM

PR #4677, opened 2023-12-13
Forced Photometry Form - dynamic date range

Theodlz:
Right now, the start and end date values of our instrument forms are calculated in the backend when they are fetched. However, this means that the values aren't updated in real time. So for example, if you have a Fritz tab left open, the date will never update.

This PR addresses that problem for both forced photometry and follow-up requests:
- for the forced photometry, we simply use 30 days with now as the end date.
- for follow-up requests, we calculate the time range used from the default values, and recalculate new values using now as the start date and now + calculated range as the end date.


mcoughlin (approved):
LGTM

PR #4678, opened 2023-12-14
GitHub actions: PostgreSQL 15

Theodlz:
Migrate to PostgreSQL 15 for the GitHub actions.

If possible, we'll also want to open a PR where we document how to upgrade the DB in our docs.


stefanv (commented):
Looks good, thanks @Theodlz

PR #4679, opened 2023-12-14
Insert photometry bugfix - no error when no rows are uploaded

Theodlz:
A variable (upload_id) ended up being undefined if there is no new photometry to upload. This should fix it.


mcoughlin (approved):
LGTM

PR #4680, opened 2023-12-15
Thumbnails ingestion - Simpler & (hopefully) better

Theodlz:
Truth is we do not need a queue for all thumbnails types. Only PS1 is challenging, because we can't simply format a string and BOOM we've got the URL to add in the DB. Instead, we have to run an API call which proved to be more and more unstable.

So here, we do a few things:
- Synchronously post thumbnails for sdss and ls as this doesn't require any external requests, wherever we need. This effectively replaces all the bits of code where we pushed the obj_id to the thumbnails queue's internal API endpoint.
- Remove the internal API endpoint from the queue, making it much simpler to maintain.
- Instead of getting the list of objects to process from an internal API queue, the service now runs a simple DB query which does: "give me the most recent object for which we don't have all three thumbnail types: sdss, ls, and ps1. Generate what is missing". We repeat that DB query in a loop to hopefully generate missing sdss and ls thumbnails for all objects, and generate ps1 thumbnails.

### DIFF
--- a/services/observation_plan_queue/observation_plan_queue.py
+++ b/services/observation_plan_queue/observation_plan_queue.py
@@ -1,10 +1,8 @@
 import itertools
-import requests
 import time
 
 import arrow
 import sqlalchemy as sa
-from sqlalchemy.orm import scoped_session, sessionmaker
 
 from baselayer.app.env import load_env
 from baselayer.app.models import init_db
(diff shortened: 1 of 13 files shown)


services/observation_plan_queue/observation_plan_queue.py
```diff
@@ -22,6 +24,26 @@
 
 Session = scoped_session(sessionmaker())
 
+REQUEST_TIMEOUT_SECONDS = cfg['health_monitor.request_timeout_seconds']
```
mcoughlin, line 27: Maybe this could be a get with a default?

services/thumbnail_queue/thumbnail_queue.py
```diff
@@ -26,118 +21,115 @@
...
+
+REQUEST_TIMEOUT_SECONDS = cfg['health_monitor.request_timeout_seconds']
 
-def service(queue):
+host = f'{cfg["server.protocol"]}://{cfg["server.host"]}' + (
+    f':{cfg["server.port"]}' if cfg['server.port'] not in [80, 443] else ''
+)
 
+
+def is_loaded():
```
mcoughlin, line 33: since we use this over and over again, could it be put into a microservices utils file?

mcoughlin (approved):
LGTM

services/observation_plan_queue/observation_plan_queue.py
Theodlz: I would not actually. This is a basic baselayer config argument we should also have in multiple places of the app, and we must be using the right value. You can imagine a user not passing the config right and we want it to fail if that happens I think.

services/thumbnail_queue/thumbnail_queue.py
Theodlz: Done!

mcoughlin (approved):
LGTM

PR #4682, opened 2023-12-19
Alert-based forced photometry - photstats

Theodlz:
We aren't ingesting ZTF alert-based forced photometry in SkyPortal yet, but before we do we want to make sure that we take that forced photometry into account when calculating photstats, to reject them from the stats that exclude forced photometry.


mcoughlin (approved):
LGTM

PR #4686, opened 2023-12-28
PS1 Thumbnails: https -> http

Theodlz:
Recent commits on main handled the SSL error we get when querying PS1 with Python `requests`, but the problem is this keeps adding unavailable PS1 thumbnails in production while slowing down the thumbnails queue as we are waiting for the error to occur every time, meaning we take 10 seconds to add thumbnails to any object.

This prevents this error from occurring, even if moving to http isn't ideal. However, let's remember that this first URL is used to get the final image URL, which does have https in it, so it's not a problem for the end user. The URL saved in the DB for the thumbnails does have https in it.

### DIFF
--- a/skyportal/models/obj.py
+++ b/skyportal/models/obj.py
@@ -517,7 +517,7 @@ class Obj(Base, conesearch_alchemy.Point):
         we assume that the image is not available and return None.
         """
         ps_query_url = (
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #4687, opened 2023-12-29
scanner board

mcoughlin:
This PR adds a component to track scanning statistics.


Theodlz:
Just wondering, the numbers you have are sources, but you can have multiple sources per object, e.g. one for every group the obj is saved to. 

Should we rather have just one "point" per object saved regardless of how many groups a scanner saves it to?

mcoughlin:
@Theodlz Yeah great point. One per object is good.

Theodlz:
Also, the display on the dashboard isn't quite right. This will require changes to the config default to have proper placement for each screen size.

Theodlz:
What is the left column of the table supposed to represent? Who's first? To make the display a little cleaner (the star can be very small here), maybe we can simply have one star right next to the name in the Name column, to not have another column for that (seems unnecessary, given that we display the star only for one person in the list).

Theodlz:
I'm a litttttttle bit worried about performance of the query once in prod, but only trying it will tell us. 

For that reason, maybe we can consider separating frontend and backend in 2 PRs, so we can get the backend in and play around with it before we have that API endpoint called by everyone when they open a page on Fritz.

Theodlz:
> @Theodlz I was hoping to have stars for the top 3 people (gold, silver, bronze), and fine for me with the API!

Ok! Yeah stars or medals icons sound good, then can be lame and boring for the others so that they feel BAD.

mcoughlin:
@Theodlz That's always the right choice. ;)

Theodlz:
@mcoughlin feel free to review your PR with the changes I made

A little test wouldn't hurt either 😉. At least an API one. I can take care of it later if you can't.

Theodlz (approved):
LGTM! 🥇

Theodlz:
oh wait test to fix

PR #4707, opened 2024-01-05
Quick save source to only save updated groups

mcoughlin:
This PR fixes the quick save source to only save updated groups.


Theodlz (approved):
LGTM! Thanks a lot for the bugfix.

PR #4711, opened 2024-01-09
QR token embedding

mcoughlin:
This PR adds a QR code embedding the token.


Theodlz:
@mcoughlin I made some changes:
- removed the dependency you added, as we already had a qrcode one for the source and alert page sharing.
- edited the existing SharePage component, so we can use that nice button + dialog+ qrcode thing not just with the current URL but any value if specified.

mcoughlin:
Thanks @Theodlz !!!

PR #4714, opened 2024-01-12
Photometry handler bugfix: avoid deduplication error if there is no photometry

Theodlz:
Right now, if a source has no photometry the handler throws a pandas-related error as it tries to access a column that doesn't exist.


mcoughlin (approved):
LGTM

PR #4715, opened 2024-01-15
Query options for savers and times

mcoughlin:
This PR fixes the savers query to enable different lookback times and number of savers.


Theodlz (approved):
LGTM!

PR #4716, opened 2024-01-16
Frontend V2: Navbar and Appbar

Theodlz:
This PR implements some of the design changes discussed with the UX/UI team, with the addition of some of my own design choices which I think will help us get a lot of screen space back without sacrificing functionality.

Here, we:
- The side menu is now permanent on Desktop, but minimized. By default, only the icons are shown (with tooltips) and the menu can be expanded to see the full name of each option. Thanks to that, we can also make the default to NOT show the expanded menu, getting back a lot of width to display more info without forcing the user to close the menu every time they open a tab. The menu is also thinner even when extended. On mobile, the menu remains more or less the same. We did remove the search bar from said menu.
- The Appbar (the header of the webpage) is now thinner, using less unnecessary height. Also, the search bar has been moved to the app bar, and can search for both sources and GCN events with the category selector. The App bar is now more responsive and will hide/show logos, and switch between showing the full search bar and an icon to show it when the screen size gets smaller.

Overall, this should give us a cleaner look while increasing functionality.

I strongly encourage reviewer(s) to run the code here, and play around with different screen sizes and browsers.


mcoughlin (approved):
@Theodlz I clicked through a bunch of pages in both big and small window, looks really neat. Thanks for doing the Source/GCN Event search, should also think about opening an issue for Comments once this is merged. Really nice job.

Theodlz:
> @Theodlz I clicked through a bunch of pages in both big and small window, looks really neat. Thanks for doing the Source/GCN Event search, should also think about opening an issue for Comments once this is merged. Really nice job.

Thanks for the review Michael! Yes, I thought about the comments (as discussed), but I thought it might be nice to get that frontend PR as is and leave the necessary backend changes to support comments search for future PR.

mcoughlin:
@Theodlz Understood.

skyportal/tests/frontend/test_newsfeed.py
```diff
@@ -84,8 +102,26 @@ def test_news_feed_prefs_widget(
...
     driver.get(f'/become_user/{user.id}')
     driver.get('/')
+
+    driver.wait_for_xpath('//span[text()="a few seconds ago"]')
+    driver.wait_for_xpath('//*[@id="newsFeedSettingsIcon"]')
+
     # Default is to not show bot comments; enable for now
-    driver.click_xpath('//*[@id="newsFeedSettingsIcon"]')
+    n_retries = 0
+    while n_retries < 3:
```
stefanv, line 111: Not fond of this pattern.

static/js/components/EarthquakePage.jsx
```diff
@@ -97,7 +97,7 @@ DownloadXMLButton.propTypes = {
   }).isRequired,
 };
 
-const sidebarWidth = 190;
+const sidebarWidth = 170;
```
stefanv, line 100: Meh, these things really belong in CSS, don't they? Perhaps not possible here.

static/js/components/HeaderContent.jsx.template
```diff
@@ -1,92 +1,138 @@
...
+    if (isDesktop) {
+      setOpenSearch(false);
+    }
+  }, [isDesktop]);
+
   return (
-    <div className={classes.topBannerContent}>
-      <div style= {%- raw -%} {{display: "inline-flex", flexDirection: "row" }} {%- endraw -%} >
+    <div className={classes.topBannerContent} style={{"{{justifyContent: openSearch ? 'flex-end' : 'space-between'}}"}}>
+      <Link onClick={() => dispatch(hydrate())} to="/" style={{"{{ display: openSearch ? 'none' : 'inline-flex', flexDirection: 'row', alignItems: 'center', textDecoration: 'none' }}"}}>
```
stefanv, line 86: Best practice may be to use logic like this to set a class, but not style, and specify the latter in CSS.

stefanv (commented):
Overall this PR looks fine; certainly a big UI improvement, thanks!

The inlining of so many styles doesn't look quite right to me; do you have thoughts on that? I would think logic would determine classes, and CSS would determine how those are rendered?

Theodlz:
> Overall this PR looks fine; certainly a big UI improvement, thanks!
> 
> The inlining of so many styles doesn't look quite right to me; do you have thoughts on that? I would think logic would determine classes, and CSS would determine how those are rendered?

So we rely on that MUI CSS style methods, where in a given file we create a useStyle method that we then call to get a `classes` object that we use to apply styling to different components.

That's where some things could move for sure. ALSO, we have `Theme.jsx` where we can apply some global CSS rules that then become the defaults for all of the MUI components.

But no, at least so far, SklyPortal doesn't use a single CSS stylesheet like some styling frameworks do. But that's pretty standard now, things like MUI or tailwind all use that global + .jsx file level styling.

What we can remove when relevant are the `style={{ some_css_stuff }}` that we've accumulated (a lot of it is on me!). Sometimes it's just a pain to keep track of what style is applied to what element with the MUI stuff, and can create more boilerplate than a simple `style=` would have.

static/js/components/EarthquakePage.jsx
Theodlz: Removing it from all the pages that have it but the sidebar component itself. We don't need it at all anywhere else. I'm removing our current hacky listeners in favor of the cleaner MUI `useMediaQuery` that relies on the more standard breakpoints.

static/js/components/HeaderContent.jsx.template
Theodlz: This style is controlled by the `openSearch` variable defined in this file, so not sure how we could define it elsewhere like a global css thing.

PR #4719, opened 2024-01-19
Apply new linter rules to codebase

Theodlz:
Earlier today I thought "hey let's update the versions of the linter rules to match the baselayer ones!".

I regret everything.


Theodlz:
looks like new linter rules just HATE any extra space and can't tolerate not having a comma on every single frontend line 🤡

mcoughlin (approved):
LGTM

PR #4721, opened 2024-01-20
fix dependabot config (schedule.time format)

Theodlz:
Looks like dependabot's time formatting is more strict than it used to be :)


mcoughlin (approved):
LGTM

PR #4737, opened 2024-01-21
Source Page V1.5

Theodlz:
Maybe not a V2, but I think this is a solid clean-up and refactoring of the source page. Most of the "one-line" info and buttons we had on the top left that were eating a lot of space are condensed in some kind of info card that fits in a paper component like anything else on the page.

Also, we all know how complicated it became to keep the mobile page in sync with the desktop one. Here, we remove the separate mobile page completely, in favor of a single page that relies on a MUI grid. The tricky part here is having that side panel that we want to show/hide whereas on mobile we have only one column. That doesn't work that well with the MUI grid. But, with a few changes and tricks, we can get something clean that works very well.

Also, the minimized menu sidebar can now open (without moving the content) on hover, showing the text next to the icons without having to open it with the menu icon (that makes it persistent, moving the content).

I also adjusted a lot of the spacing and made some changes to the dashboard which I hope makes it look cleaner. I made similar changes to some of the components on the source page to minimize the whitespace we kept accumulating.

TODOs:
- [ ] I modified a bunch of components that we also use on other pages. For those, I need to make sure they are still displayed decently everywhere else.
- [x] I'm not sure about the thumbnail display. Here I opted for a one-line image gallery, which is the best way I could find to keep the layout consistent with any number of thumbnails added to a source.
- [x] Fix the tests that I bet will break :)
- [x] On the source page, lots of `style` arguments to refactor into some reusable MUI styles. Very important to take care of that before merging.

Any feedback is more than welcome! It might look better to me, but be less practical for an astronomer, so I'd benefit from the reviewers running the PR and spending some time using this source page.


Theodlz:
@mcoughlin I still have things to take care of before merging, but ti's more than ready for a first review

mcoughlin (approved):
I pulled down and played around on the page, both in desktop and mobile format. Very neat, very concise.

Theodlz:
> I pulled down and played around on the page, both in desktop and mobile format. Very neat, very concise.

Thanks for the review! Nothing to add/delete/modify? And do you think that the thumbnail stuff is fine like this?

mcoughlin:
@Theodlz no I think it looks nice, including the thumbnails.

PR #4740, opened 2024-01-24
New source page: thumbnails and user preferences feedback

Theodlz:
We got a batch of users to try the new source page before deployment to production. Overall they are happy with it, but they do have some suggestions. This PR implements those + some of the suggestions from @profjsb brought up during the weekly SkyPortal meeting.

Now, if there's enough space, we get the thumbnails in a 6x1 layout, and 3x2 thereafter. Users can decide to hide summaries from the source page's header by default, or only hide the AI-generated ones. Also, they can hide the "similar sources" feature (which is based on said AI summaries embeddings) which is now hidden by default as it created some amount of confusion. The edit/pen icons for most editable fields are now hidden except on hover.

Other items addressed here:
- brought back the background greyed-out spectra when smoothing
- minimized a lot of the margins around the plots
- changed the orientation of the hide/show right panel on source page
- rearranged some of the items on the source page
- source plugins, which we use to add ztf-specific buttons in Fritz, are now in the same row/block as the normal buttons, which makes it more compact
- summaries can now be hidden from the source page header, can be turned on from the user preferences. Defaults to showing the summaries.
- AI-based summaries are now hidden from the source page header by default, showing only human summaries. Can be reactivated from the user preferences.
- fixed the margin issue between the additional logos (like the growth logo) and the basic skyportal one.

### DIFF
--- a/static/js/components/PhotometryPlot.jsx
+++ b/static/js/components/PhotometryPlot.jsx
@@ -336,11 +336,13 @@ const PhotometryPlot = ({
         point.origin !== "" &&
         point.origin !== null
       ) {
-        key += `/${point.origin}`;
-      }
-      // limit the key to 30 characters
-      if (key.length > 30) {
-        key = `${key.substring(0, 27)}...`;
+        // the origin is less relevant, so we crop it to not have more than 23 characters + 3 x ...
+        const remaining = 23 - key.length;
+        if (remaining < point.origin.length) {
+          key += `/${point.origin.substring(0, Math.max(remaining - 3, 3))}...`;
+        } else {
+          key += `/${point.origin}`;
+        }
       }
       if (!acc[key]) {
         acc[key] = [];
@@ -400,7 +402,7 @@ const PhotometryPlot = ({
             text: upperLimits.map((point) => point.text),
             mode: "markers",
             type: "scatter",
-            name: key,
+            name: `${key} (UL)`,
             legendgroup: `${key}upperLimits`,
             marker: {
               line: {
@@ -612,8 +614,8 @@ const PhotometryPlot = ({
             text: upperLimitsText,
             mode: "markers",
             type: "scatter",
-            name: key,
-            legendgroup: key,
+            name: `${key} (UL)`,
+            legendgroup: `${key}upperLimits`,
             marker: {
               line: {
                 width: 1,
@@ -931,7 +933,7 @@ const PhotometryPlot = ({
         sx={{
           display: {
             maxWidth: "95vw",
-            width: "100&",
+            width: "100%",
             "& > button": { lineHeight: "1.5rem" },
           },
         }}
@@ -944,7 +946,7 @@ const PhotometryPlot = ({
       <div
         style={{
           width: "100%",
-          height: plotStyle?.height || "65vh",
+          height: plotStyle?.height || "70vh",
           overflowX: "scroll",
         }}
       >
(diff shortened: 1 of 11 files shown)


mcoughlin (approved):
Onwards and upwards.

static/js/components/PhotometryPlot.jsx
```diff
@@ -955,14 +957,39 @@ const PhotometryPlot = ({
             legend: {
               orientation: mode === "desktop" ? "v" : "h",
               yanchor: "top",
-              y: mode === "desktop" ? 1 : -0.25,
-              x: mode === "desktop" ? 1.15 : 0,
+              y: mode === "desktop" ? 1 : plotData?.length > 10 ? -0.4 : -0.3, // eslint-disable-line no-nested-ternary
```
stefanv, line 960: A comment explaining these decisions would be useful here

static/js/components/ShowSummaries.jsx
```diff
@@ -36,11 +36,22 @@ export const useStyles = makeStyles((theme) => ({
...
 
-const ShowSummaries = ({ summaries = [] }) => {
+const ShowSummaries = ({ summaries = [], showAISummaries = true }) => {
   const styles = useStyles();
   const renderCommentText = () => {
-    if (summaries?.length > 0) {
-      return summaries[0].summary;
+    let filteredSummaries = [...summaries].filter(
+      (summary) =>
+        summary?.summary &&
```
stefanv, line 44: our is-null test comes up again!

static/js/components/Source.jsx
```diff
@@ -208,6 +207,74 @@ export const useSourceStyles = makeStyles((theme) => ({
   },
 }));
 
+const SourcePageThumbnails = ({
+  ra,
+  dec,
+  thumbnails,
+  rightPanelVisible,
+  downSm,
```
stefanv, line 215: downSmall, downLarge

And somewhere a comment is needed to signify what these mean

stefanv: Double checking that this cannot be handled by CSS alone

static/js/components/Source.jsx
```diff
@@ -258,7 +332,26 @@ const SourceContent = ({ source }) => {
   const noSummary =
     source.summary_history?.length < 1 ||
     !source.summary_history ||
-    source.summary_history[0].summary === null;
+    source.summary_history.filter(
+      (summary) => summary.summary !== null && summary.summary.trim() !== "",
+    )?.length < 1;
+
+  const noHumanSummary =
+    source.summary_history?.length < 1 ||
```
stefanv, line 340: lots of null checking!

static/js/components/Source.jsx
```diff
@@ -231,15 +298,22 @@ const SourceContent = ({ source }) => {
   const [copyPhotometryDialogOpen, setCopyPhotometryDialogOpen] =
     useState(false);
   const [tnsDialogOpen, setTNSDialogOpen] = useState(false);
-  const [thumbnailDialogOpen, setThumbnailDialogOpen] = useState(false);
-  const [anchorEl, setAnchorEl] = React.useState(null);
-  const open = Boolean(anchorEl);
+
+  const [anchorElFindingChart, setAnchorElFindingChart] = React.useState(null);
```
stefanv, line 302: Comment would be helpful

stefanv (commented):
Looks good;

static/js/components/ShowSummaries.jsx
Theodlz: I know 😓. I promise I will have another huge refactor PR dedicated to this 😉

static/js/components/Source.jsx
Theodlz: Definitely we could, but since we are creating some of these grid and grid items with MUI + passing these size elements to the components themselves, that would require its own refactor.

static/js/components/Source.jsx
Theodlz: done!

static/js/components/PhotometryPlot.jsx
Theodlz: done!

stefanv (approved):

PR #4742, opened 2024-01-27
sinceDaysAgo float

mcoughlin:
This PR makes fractional days for sinceDaysAgo possible.

### DIFF
--- a/skyportal/handlers/api/internal/source_savers.py
+++ b/skyportal/handlers/api/internal/source_savers.py
@@ -19,8 +19,10 @@ class SourceSaverHandler(BaseHandler):
         top_savers_prefs = {**default_prefs, **top_savers_prefs, **user_options}
 
         max_num_savers = int(top_savers_prefs['maxNumSavers'])
-        since_days_ago = int(top_savers_prefs['sinceDaysAgo'])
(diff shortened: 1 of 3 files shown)


Theodlz:
@mcoughlin nothing else needed for it to work? I noticed that it didn't work in prod yet, just wanted to make sure this was related.

mcoughlin:
I honestly can't quite tell, but for sure this was a problem.

Theodlz:
Got it, let's merge this then.

Theodlz (approved):
LGTM

Theodlz:
@mcoughlin it's a frontend refresh issue. I can see that the right data is loaded in the console, but it doesnt refresh the dashboard. Probably just needs a nice useEffect

mcoughlin:
Can you tell where?

Theodlz:
let me take a look

Theodlz:
genuinely confused, as this does work just fine locally, hmmm. Maybe an issue with the existing user preferences in prod (that were missing those new keys) that I don't have on a brand new database. Will keep looking

Theodlz:
yeah an odd rendering issue. Using the search bar rerenders the list hmmmm

Theodlz:
@mcoughlin I think this should help

PR #4744, opened 2024-01-31
WINTER Follow-up requests

Theodlz:
This PR adds the code in `skyportal/facility_apis/winter.py` to trigger follow-up requests on specific sources with WINTER.

Here, we wanted to have different default values of n_dither and exposure_time when using different filters. Unfortunately, `react-json-schema-form` initializes fields with their defaults on initial render, and changing the value of a field can't overwrite the content of another to set a new default. To get around that, we simply have different fields as dependencies of each filter in the enum, which we rename later on when submitting. That might get trickier if we implement EDIT later on, but that's the only solution to get that up and running right now without re-inventing the wheel.


skyportal/facility_apis/winter.py
```diff
@@ -0,0 +1,405 @@
...
+env, cfg = load_env()
+
+
+WINTER_URL = f"{cfg['app.winter.protocol']}://{cfg['app.winter.host']}"
+if cfg['app.winter.port'] is not None and int(cfg['app.winter.port']) not in [80, 443]:
+    WINTER_URL += f":{cfg['app.winter.port']}"
+
+log = make_log('facility_apis/winter')
+
+WINTER_USERNAME = cfg.get('app.winter.username', None)
```
mcoughlin, line 23: I think this belongs in the allocation secrets, which is where we put such things encrypted.

skyportal/facility_apis/winter.py
```diff
@@ -0,0 +1,405 @@
...
+        """
+
+        if WINTER_USERNAME in [None, ''] or WINTER_PASSWORD in [None, '']:
+            raise ValueError("WINTER username and password not set in config.")
+
+        from ..models import FacilityTransaction
+
+        req = WINTERRequest()
+
+        altdata = request.allocation.altdata
```
mcoughlin, line 185: Yeah in the altdata here.

mcoughlin (changes requested):
I think the secrets should go in altdata. Also yes to migrations.

skyportal/facility_apis/winter.py
Theodlz: So as Robert explained to me, users that will trigger aren't expected to have a WINTER api key/account, only credentials for a program (which is what I've put in altdata already). So it would make sense to have these credentials (that are used regardless of the program/allocation) available instance-wide.

Theodlz: What I can do is have the config credentials, and if they aren't there (so if the instance hasn't been configured to use WINTER by sysadmin), then you can still pass those in the allocation as a fallback?

Theodlz: Just to be clear @mcoughlin I'm not saying this is ABSOLUTELY necessary, just that it made sense here, but @robertdstein and I have no problem putting that in the altdata if you think that's better.

mcoughlin: Personally I think it is more consistent with how we have used this. I don't really like the secrets in the config.

Theodlz: done

mcoughlin (approved):
LGTM

PR #4745, opened 2024-01-31
Winter followup bugfix - max airmass + store defaults

Theodlz:
* for optional fields, enforce the defaults before saving the payload to the DB, for bookkeeping
* fix the airmass field, which had an incorrect name


mcoughlin (approved):
LGTM

PR #4749, opened 2024-02-01
Follow-up requests constraints: cancel if already classified in TNS

Theodlz:
This PR adds an optional argument when posting follow-up requests, that allows the system to cancel triggering if there are any sources within 0.5 arcsec that have a TNS SN designation already.

### DIFF
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -14,6 +14,7 @@ import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+import conesearch_alchemy as ca
 import sqlalchemy as sa
 from astroplan import FixedTarget, Observer, ObservingBlock
 from astroplan.constraints import (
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

skyportal/handlers/api/followup_request.py
```diff
@@ -429,6 +430,23 @@ def post_followup_request(
...
+            # don't trigger if there is any source within 0.5 arcsec of the target
+            # that has a tns_name that contains "SN"
+            obj = session.scalars(
+                Obj.select(session.user_or_token).where(Obj.id == data['obj_id'])
+            ).first()
+            if obj is None:
+                raise ValueError(f'Could not find source with ID {data["obj_id"]}.')
+            stmt = Obj.select(session.user_or_token).where(
+                Obj.within(ca.Point(ra=obj.ra, dec=obj.dec), 0.5 / 3600),
+                Obj.tns_name.contains("SN"),
```
nabeelre, line 443: Does this fail if the TNS name is something like, e.g., AT 2024asn?

Theodlz: Nop, my understanding was that we only want to cancel triggering if its classified on TNS (has SN in its name), and otherwise we let it trigger.

@nabeelre is that correct? Or do you want the bot to not do anything as soon as there is an entry on TNS (even if not with SN)?

Theodlz: this query just says "look for objects within 0.5 arcsec and with a TNS name with SN in it", if we find any results we cancel the trigger.

nabeelre: Sorry, let me clarify. I was asking what the behavior would be if there's an unclassified source that just happens to have the letters "sn" in the randomized part of its name, for example, AT 2024asn. Unclassified but contain "sn." Is .contains() case sensitive?

Theodlz: oh gotcha @nabeelre. It shouldn't be, but let me test that ASAP (haven't deployed yet anyway).

Theodlz: @nabeelre answer is yes, it is case sensitive. Trying it locally, I can't trigger if its SN2023ab, but I can if it's sn2023ab. Maybe we want it to not be case sensitive but only check if the string starts with SN. Might make more sense in case people add tns name with an endpoint without capitalizing that

Theodlz: But as is, this works

Theodlz: Opening a PR now to replace `contains` by `startswith` while keeping it case sensitive, seems like the safest way to go

Theodlz: `sa.func.lower(Obj.tns_name).startswith("sn")`

PR #4751, opened 2024-02-01
Python requirements update

Theodlz:
- dependabot struggling with updating openai[embeddings]. This happens because we have a very old one pinned, that has an optional dependency that doesn't exist anymore (now part of the core package).
- dependabot strulling with numpy, because of the version of numba that is pinned, so here we update both.
- overall update of some packages that dependabot didn't update very recently.
- relax the version pinned for tables and stdpipe. The latest versions of these 2 packages we barely use are hard to install on mac silicon.


mcoughlin (approved):
LGTM

PR #4755, opened 2024-02-02
photometry table download

mcoughlin:
This PR fixes the photometry table download csv.


Theodlz (approved):
Other than the comment that I left on the issue, this looks good. LGTM

PR #4756, opened 2024-02-07
GCN Observations Survey - Bugfixes and frontend improvements

Theodlz:
- have the table on the left and form on the right, using the whole width of the page (unless the screen is too small, in which case they are on top of each other).
- use websockets to get the frontend to fetch the updated list of survey analysis when adding and deleting, and when a survey that's running finishes.
- custom search and download logics.
- fix the path to the dustmaps


Theodlz:
@mcoughlin the tooltips/docs will be added in a separate PR tomorrow, with Tomas' help.

mcoughlin (approved):
LGTM. Thanks for fixing this.

PR #4757, opened 2024-02-07
Brotli support in nginx

Theodlz:
See https://github.com/cesium-ml/baselayer/pull/379

This PR adds the necessary lines in the dockerfile to install nginx with brotli support, and the documentation needed.

This is a draft, and thinks need to change before reviewing and/or merging. For example, I would like to move the `nginx.service` file to baselayer's nginx service.


Theodlz:
@stefanv just so you know that this exists. I added the brotli unit test there to test it as well, but we can keep/remove it from SP too. One thing I wanted to try here is building the dockerfile, but the GA didn't trigger it when the docker label was added which is surprising, I remembered that this was fixed before.

Theodlz:
Test are passing, and this is a minor/optional feature. We need to merge on main to trigger the docker building GA anyway, so self merging

PR #4758, opened 2024-02-09
Followup - WINTERAPI: implement DELETE + add target name in POST payload

Theodlz:
- add a delete method to the WINTERAPI, that grabs the schedule name from the POST response (from facility transactions) and calls the `/too/delete` endpoint.
- add a method to grab said schedule name from the POST response
- send the target_name (obj_id) to WINTER when submitting a request (max 30 chars).

### DIFF
--- a/skyportal/facility_apis/winter.py
+++ b/skyportal/facility_apis/winter.py
@@ -97,9 +97,6 @@ class WINTERRequest:
         # the WINTERAPI POST response has a "msg" key,
         # that contains Schedule name is <schedule_name> if successful
         if "msg" in content:
-            # just to that we are not sensible to the exact format of the message
-            # we look for the string "Schedule name is" and get what's after that
-            # then split by space and take the second element
             # it's in quote, so remove them and strip
             schedule_name = (
                 str(content["msg"].split('Schedule name is')[1])
--- a/skyportal/facility_apis/winter.py
+++ b/skyportal/facility_apis/winter.py
@@ -1,10 +1,11 @@
+import json
 import requests
+import urllib
 from requests.auth import HTTPBasicAuth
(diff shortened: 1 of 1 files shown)


Theodlz:
@robertdstein just so you know that exists

mcoughlin:
@Theodlz just confirming that you see the failed tests.

skyportal/facility_apis/winter.py
```diff
@@ -1,10 +1,11 @@
+import json
 import requests
+import urllib
 from requests.auth import HTTPBasicAuth
 from datetime import datetime, timedelta
 from astropy.time import Time
-import urllib
 
-from . import FollowUpAPI
+from . import FollowUpAPI, FacilityTransaction
```
mcoughlin, line 8: This likely needs to be imported inside.

skyportal/facility_apis/winter.py
```diff
@@ -89,6 +91,28 @@ def _build_payload(self, request):
 
         return target
 
+    def schedule_name(request):
+        content = request.transactions[0].response["content"]
+        content = json.loads(content)
+        # the WINTERAPI POST response has a "msg" key,
+        # that contains Schedule name is <schedule_name> if successful
+        if "msg" in content:
+            # just to that we are not sensible to the exact format of the message
```
mcoughlin, line 100: sensible is a weird word.

mcoughlin (approved):
Probably just need to move that import.

skyportal/facility_apis/winter.py
Theodlz: @mcoughlin that's the french in me sorryyyyy. Will fix ASAP. I also fixed the import already

PR #4759, opened 2024-02-10
SourceHandler: search both my id and tns_name

Theodlz:
- single source fetch: can now specify an additional variable called tns_name, which is an exact match.
- multiple sources fetch: the existing sourceID (capable of partial match) now also looks at tns_name

This makes a lot of sense because folks want to look for objects by TNS name, and not just the survey's internal name which is our objectId in most cases here).

This new feature is also supported in the frontend for both source table and quicksearch bar.


mcoughlin (approved):
LGTM

PR #4760, opened 2024-02-11
Remove unused NGINX service config

Theodlz:
Remove unused config file added in previous commit, when attempting to get brotli installed from source and not from ppa


Theodlz:
self merging

PR #4766, opened 2024-02-13
Follow-up requests constraints: conesearches

Theodlz:
In practice, we noticed how many duplicates of a source you might get with a survey like ZTF.

So it makes more sense to apply these constraints to nearby sources and not by exact name.

Also, we add a "no_duplicates" constraints to cancel triggering if there's already a follow-up request with the same allocation that is submitted or completed. We already have that logic in Kowalski but this will add redundancy and help with concurrency issues.


Theodlz:
@nabeelre

PR #4768, opened 2024-02-15
Follow-up requests constraints: not_if_duplicates as cone search

Theodlz:
Just like for the other conditions, use a cone search instead of an exact match.

Also, change the queries from scalars().all() to scalars().first() to only get one match, as anyway our conditions are of the form "if any", meaning we don't need more than one.


mcoughlin (approved):
LGTM!

PR #4769, opened 2024-02-20
TNS V2

Theodlz:
The idea here is that we want better TNS interactions, with fine grain control over who can use a TNS robot to report anything to TNS, who should always be mentioned as coauthor, and who is allowed to trigger auto-reports when saving to a group that we allow auto-report for.

This means better SQLAlchemy models:
- a TNSRobot model not bound to a single group
- a TNSRobotGroup model that binds N robots to N groups: a subset of these groups can be owners of the robot, and can activate auto reporting, add/remove auto-reporters, add/edit/remove groups...
- a TNSRobotGroupAutoreporter model that binds N robots to N group users allowed to trigger an automatic report when a source is saved to a group.
- a TNSRobotCoauthor model that lists users that will always be listed as coauthor of a TNS report made with this group.
- a TNSRobotSubmission model binding an obj to a tns_robot (and a user "submitter") for submission with a submission status. Used by the microservice to process requests instead of sending internal API calls and losing requests on restart/crash.
- A brand new microservice submission service, feeding from the submission table, and sending error messages to the frontend user with websockets.

Naturally, this comes with API endpoints to create/modify/delete entries in said tables, and:
- a dedicated frontend page, to create/edit/delete all TNS robots, instead of one table on each group page
- a submission table page, where one can see all the submission jobs created for a robot, along with their status and payload if submitted to TNS.

NEW ADDITIONS:
* redesigned the TNS retrieval microservice, to not have these 2 duplicate logics running in parallel to query TNS for data of new and existing objects, but only one service which queue is fed by a TNS watcher (looking for new objects on TNS to add to SkyPortal) and the API (frontend users ask the app to find the TNS name of an existing object).
*  deprecated the spectra report submission, which does not work with the current code on main anyways, and is not used by anyone. We might want to bring this feature back in the future based on user request(s).


Theodlz:
@nabeelre so you know that this exists

Theodlz:
@mcoughlin tests won't pass because I need to merge a PR in baselayer, but 99% of the features are implemented here.

static/js/components/TransferList.jsx
```diff
@@ -0,0 +1,197 @@
...
+import Card from "@mui/material/Card";
+import CardHeader from "@mui/material/CardHeader";
+import ListItemButton from "@mui/material/ListItemButton";
+import ListItemText from "@mui/material/ListItemText";
+import ListItemIcon from "@mui/material/ListItemIcon";
+import Checkbox from "@mui/material/Checkbox";
+import Button from "@mui/material/Button";
+import Divider from "@mui/material/Divider";
+
+function not(a, b) {
```
mcoughlin, line 14: Something for a .js utility file?

mcoughlin (approved):
A few ideas here and there, small edits.

skyportal/handlers/api/tns.py
```diff
@@ -241,203 +389,513 @@ def put(self, tnsrobot_id):
...
+        # the PUT handler is used to add or edit a group
         data = self.get_json()
+        autoreport = data.get('autoreport', None)
+        owner = data.get('owner', None)
+        if group_id is None:
+            group_id = data.get('group_id', None)
+
+        if autoreport is not None:
+            # try to convert the autoreport to a boolean
+            if str(autoreport) in ['True', 'true', '1', 't']:
```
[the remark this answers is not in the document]
Theodlz: Maybe, but no idea where it is...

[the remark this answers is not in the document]
Theodlz: Yeah I only find different places implementing that, though I agree that just one method re-used accross the app would be so much better

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -35,425 +42,579 @@
...
-            required_keys = {'obj_ids', 'tnsrobot_id', 'user_id', 'reporters'}
-            if not required_keys.issubset(set(data.keys())):
-                self.set_status(400)
-                return self.write(
-                    {
-                        "status": "error",
-                        "message": "TNS requests require obj_ids, tnsrobot_id, user_id, and reporters",
-                    }
+                # DEBUG ONLY, NOT SENDING
+                status_code = 0
```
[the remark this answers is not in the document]
Theodlz: the comment no, the status code to 0 yes

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -35,425 +44,613 @@
...
-        if 'api_key' not in altdata:
-            raise ValueError('Missing TNS API key.')
-
-        if archival is True:
-            if len(archival_comment) == 0:
-                raise ValueError(
-                    'If source flagged as archival, archival_comment is required'
+@check_loaded(logger=log)
+def service(*args, **kwargs):
+    while True:
```
stefanv, line 50: This is quite a long piece of code. Would it make sense to break it into functions? Something along the lines of:

```
while True:
    next_task = fetch_task()
    if next_task is None:
        time.sleep(5)
    else:
        service_task(next_task)
```

skyportal/handlers/api/tns.py
```diff
@@ -40,90 +48,199 @@
...
-
-        group_id = self.get_query_argument("groupID", None)
-
-        with self.Session() as session:
-            try:
-                # get owned tnsrobots
-                stmt = TNSRobot.select(session.user_or_token).options(
-                    joinedload(TNSRobot.auto_report_streams),
-                    joinedload(TNSRobot.auto_report_instruments),
+def create_tns_robot(
```
stefanv, line 51: Type annotate or add docstrings, even for internal functions. It's surprising how hard it can be to guess later on what we, ourselves, expected as input at the time.

stefanv (commented):
Looks sensible! I won't go to deep, since there's a warning not to review ;)

Watch out for very long pieces of code---they can make it hard to follow logic, so it's helpful to break them into functions (as you already do for, e.g., update_tns_robot).

Those functions can also often be tested by themselves, so that we need to lean less heavily on web-based testing.

services/tns_submission_queue/tns_submission_queue.py
Theodlz: Yes! Exactly what I've been doing this morning. I'm trying to avoid redundant pieces of code too.

skyportal/handlers/api/tns.py
Theodlz: Agreed, will add now!

services/tns_submission_queue/tns_submission_queue.py
Theodlz: Done!

Theodlz:
@stefanv you can rereview now if you'd like!

skyportal/models/tns.py
```diff
@@ -43,30 +38,43 @@ class TNSRobot(Base):
...
+        nullable=False,
+        server_default="",
+        doc="Acknowledgments to use for this robot.",
+    )
+
+    report_existing = sa.Column(
+        sa.Boolean,
+        nullable=False,
+        server_default="false",
+        doc="Whether to still report objects that are already in the TNS, but not reported with this object internal name (i.e., reportedby another survey).",
```
mcoughlin, line 70: reported by

mcoughlin (approved):
Only other comment is whether you want to do some logging when you put the request in the db. But looks good

Theodlz:
addressed your comments @mcoughlin 👍

mcoughlin:
@Theodlz thanks!

static/js/components/TNSRobotsPage.jsx
```diff
@@ -1038,6 +1040,20 @@ const TNSRobotsPage = () => {
     );
   };
 
+  const renderAcknowledgments = (dataIndex) => {
+    const tnsrobot = tnsrobotList[dataIndex];
+    return (
+      <Tooltip
+        title={`Added at the end of the author list, e.g. 'First Last (University), ${
+          tnsrobot?.acknowledgments || "<insert_acknowledgments_here>"
```
stefanv, line 1048: Won't this be confusing, to display the existing acknowledgements in the tooltip? Also, here and above, is the `<` expected as part of the string? If not, you may simply want to replace it with an example: "Jane Doe (CalTech), added spectroscopy" (or whatever makes sense).

services/thumbnail_queue/thumbnail_queue.py
```diff
@@ -80,7 +80,7 @@ def service(*args, **kwargs):
                 obj, err = fetch_obj(session)
                 if err is not None:
                     log(f"Error fetching object with missing thumbnails: {str(err)}")
-                    time.sleep(5)
+                    time.sleep(1)
```
stefanv, line 83: Let's go :rocket: :sweat_smile:

static/js/components/CommentAttachmentPreview.jsx
```diff
@@ -153,6 +153,26 @@ const CommentAttachmentPreview = ({
   const isCached = commentId === cachedAttachmentCommentId;
 
   const [open, setOpen] = useState(false);
+
+  const getURLs = () => {
+    const type = filename.includes(".") ? filename.split(".").pop() : "";
+    let baseUrl = "";
+    // The FilePreviewer expects a url ending with .pdf for PDF files
+    if (associatedResourceType === "gcn_event") {
```
stefanv, line 161: This may be a good place for a [switch statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch).

static/js/components/TNSRobotsPage.jsx
Theodlz: So the idea of that tooltip is explain a user what this "acknowledgments" field adds to the report. So if there is already a value, I felt like showing what the actual string send to TNS will look like was useful.

The "<insert_acknowledgments_here>" was just to make it clear that it's a placeholder where the acknowledgments will be.

I can change the format for sure, but the idea is just for the user to understand that it's basically a string appended to the end of the other list after a comma

static/js/components/CommentAttachmentPreview.jsx
```diff
@@ -272,12 +279,28 @@ const CommentAttachmentPreview = ({
...
-                  hideControls
-                />
-              )}
+              {supportedType &&
+                ["jpeg", "jpg", "png", "gif"].includes(fileType) && (
+                  <img
+                    src={previewUrl}
+                    alt={filename}
+                    style={{
+                      maxHeight: "68vh",
```
stefanv, line 288: Where do these values come from? 68vh seems rather specific.

services/thumbnail_queue/thumbnail_queue.py
Theodlz: These errors never occur, but if they do there's no point in sleeping really. I felt like a 1 second sleep was enough to not just absolutely destroy the logs if we do get stuck in an error, but no need for 5 seconds here.

static/js/components/CommentAttachmentPreview.jsx
Theodlz: Agreed! Will add in a following PR. This was a quick fix I added here based on user feedback I got last night, but I want to have a dedicated PR. `react-file-preview` is a dead package, we should not use it anymore. And the PDF display is plain crap.

static/js/components/CommentAttachmentPreview.jsx
Theodlz: I could have a `calc(do_something)` instead, but it's basically what's enough to exclude the app bar, the dialog header, and dialog footer, and the margins between the dialog and the page on top and bottom. It's pretty arbitrary, and again in that PR I want to open that I mentioned in a comment above, I want to have something that figures out the exact height that's available.

skyportal/handlers/api/source.py
```diff
@@ -1943,6 +1943,9 @@ def post_source(data, user_id, session, refresh_source=True):
                 auto_submission=True,
             )
             session.add(submission_request)
+            log(
+                f"Added TNSRobotSubmission request for obj_id {obj.id} saved to group {group.id} with tnsrobot_id {tnsrobot_group_with_autoreporter.tnsrobot_id} for user_id {user.id}"
```
stefanv, line 1947: Not necessary to use here, but fun feature of f-strings: `f"{obj.id=}"` -> `obj.id=3`.

static/js/components/CommentAttachmentPreview.jsx
stefanv: Yes, there is a replacement now, linked from the related issue.

skyportal/handlers/api/source.py
Theodlz: oh!!! Where has this been all my life 💀 💀 💀 💀 💀 💀 

Thanks! That will be super super useful to shorten a lot of f strings!

static/js/components/CommentAttachmentPreview.jsx
stefanv: It's probably a good idea to file issues for items you think you may address in the future. With checklists, if the issue covers multiple items. I'm amazed at how often I intend to fix something and then get distracted along the way.

Theodlz: Yes! I agree 100% I used to do it but I should use that again. I'm taking care of deploy right now, but I will have time to open a bunch of issues tomorrow afternoon. I got a batch of feature request that will require a significant number frontend and backend changes, so it will come in handy.

static/js/components/CommentAttachmentPreview.jsx
Theodlz: Oh I didn't see that the last issue on that package was yours hahaha

skyportal/utils/tns.py
```diff
@@ -501,6 +603,22 @@ def get_objects_from_page(
...
 def get_tns_objects(headers, discovered_period_value=5, discovered_period_units='days'):
+    """Get all objects from TNS
+
+    Parameters
+    ----------
+    headers : dict
+        Headers to use for the request
+    discovered_period_value : int, optional
+        Discovered period value, by default 5
+    discovered_period_units : str, optional
```
stefanv, line 619: Maybe make this an enum:

```
discovered_period_units : {'days', 'hours', 'minutes', 'seconds'}, optional
```

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -52,6 +52,618 @@ class TNSReportError(Exception):
     pass
 
 
+def find_accessible_tnsrobot_groups(tnsrobot, user, submission_request, session):
+    """Verify that the user is allowed to submit to TNS with this robot, and return the TNSRobotGroup to use for submission if so.
```
stefanv: Is this a permission check on the robot, on TNS, or on both?

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -52,6 +52,618 @@ class TNSReportError(Exception):
...
+        tnsrobot_group = tnsrobot_groups_with_autoreport[0]
+    else:
+        # we use the first remaining group to submit the object
+        tnsrobot_group = tnsrobot_groups[0]
+
+    return tnsrobot_group, tnsrobot_groups
+
+
+def apply_existing_tnsreport_rules(tns_headers, tnsrobot, submission_request, session):
+    """Apply the rules for existing TNS reports to the submission request.
```
stefanv, line 131: What are these rules defined? Or is this the logic in the function to prevent duplicate submissions?

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -52,6 +52,618 @@ class TNSReportError(Exception):
...
+            raise TNSReportError(
+                f'No source {obj_id} saved to any group with TNSRobot {tnsrobot_id}.'
+            )
+        else:
+            raise TNSReportError(
+                f'No source {obj_id} saved to group with TNSRobot {tnsrobot_id}.'
+            )
+
+
+def build_reporters_string(submission_request, source, tnsrobot, session):
```
stefanv, line 216: This is the type of logic that is quite easy to unit test outside of the web framework.

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -52,6 +52,618 @@ class TNSReportError(Exception):
...
+            )
+        ).all()
+
+        if len(photometry) == 0:
+            raise TNSReportError(
+                f'No photometry available for {obj_id} with instruments {instrument_ids}.'
+            )
+
+        # FILTER THE PHOTOMETRY BY STREAMS
+        # we keep the photometry that is attached to no streams or to at least one of the streams that are in the list of all_streams
```
stefanv: Out of curiosity, are you not concerned about data permissions when submitting to TNS?

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -52,6 +52,618 @@ class TNSReportError(Exception):
...
+            raise TNSReportError(
+                f'No acknowledgments found for tnsrobot {tnsrobot_id}, cannot report {obj_id} to TNS.'
+            )
+
+        reporters += f' {str(tnsrobot.acknowledgments).strip()}'
+
+    return reporters
+
+
+def process_submission_request(submission_request, session):
```
stefanv, line 658: This function seems to still do generation + submission of the report; worth separating?

services/tns_submission_queue/tns_submission_queue.py
Theodlz: on the TNS robot groups of the TNS robot. THe docstring is misleading, I will update it

services/tns_submission_queue/tns_submission_queue.py
Theodlz: I could describe them in greater detail. here too, but it's described at the DB model schema, as this relies on the `report_existing` flag.

services/tns_submission_queue/tns_submission_queue.py
Theodlz: Good point, will do!

PR #4770, opened 2024-02-21
PUT photometry deduplication: optionally ignore flux in logic + optional update

Theodlz:
Here we allow PUT photometry to not account for flux when deduplicating (meaning we consider a duplicate anything with the same mjd, instrument_id, obj_id, and origin), and to (optionally) update the duplicates if that's the case. BUT, for safety we limit the update to only photometry rows that have an origin specified already. That will avoid accidental updates of photometry from different sources that forgot to specify and origin value.

This will be very useful for 2 use cases:
- for those who upgrade their photometry pipeline, to update datapoints from the same observation and pipeline (meaning, hopefully, same mjd, instrument_id, and origin) without running a risky bulk delete and then re-adding the data.
- when pushing data from a broker, where deduplication or updates of the data happen upstream, but Fritz is naturally not aware of that, but simply replacing old photometry by new photometry means using the most recent and deduplicated values anyway.

In production, the first application of this will be pushing the alert-based forced photometry from ZTF to Fritz. That alert-based FP is recomputed for each detections/alerts, and upstream the broker has its own logic to decide what FP to keep at a given JD: old or new. So the deduplication happens upstream, but if we push FP to Fritz for alert 1, and do it again for alert 2 without this new code from this PR, we would re-duplicate the data. This allows us to prevent that.

TODO: add an ACL to use this option or limit it to superusers only for now. We don't want any user to be allowed to update existing photometry data.


mcoughlin (approved):
LGTM

Theodlz:
Thanks for the review @mcoughlin I'll just add some tests before merging tomorrow or on Thursday

skyportal/handlers/api/photometry.py
```diff
@@ -1418,12 +1456,51 @@ def put(self):
                                 f'Adding streams {stream_ids_update} to photometry {duplicate.id}'
                             )
 
+                    # if ignoring flux deduplication, do we update the value(s) of the existing photometry?
```
stefanv, line 1459: Not quite clear what these two lines of comments mean.

skyportal/handlers/api/photometry.py
```diff
@@ -664,6 +665,8 @@ def get_values_table_and_condition(df):
     df: `pandas.DataFrame`
         Dataframe with the columns 'obj_id', 'instrument_id', 'origin',
         'mjd', 'standardized_fluxerr', 'standardized_flux'.
+    ignore_flux: bool
```
stefanv, line 668:
```suggestion
    ignore_flux : bool
```

skyportal/handlers/api/photometry.py
```diff
@@ -1418,12 +1456,51 @@ def put(self):
...
+                        and ignore_flux_replace
+                        and str(duplicate.origin).strip().lower()
+                        not in ['none', '', 'nan', 'null']
+                    ):
+                        # we might have more than one datapoint at a given: jd, instrument_id, filter, and origin
+                        # that do have different flux values. We can't update the flux/fluxerr for all of them,
+                        # because this will conflict with the deduplication index we have and raise an error
+                        # so we only want to update the first one we find
+                        # for that, we compute a hash of the jd, instrument_id, filter, and origin everytime we update a row
+                        # store it, and make sure we don't update a row with the same hash twice
```
stefanv, line 1472: Also unclear; can we better document when this situation arises?

skyportal/handlers/api/photometry.py
```diff
@@ -1327,6 +1339,11 @@ def put(self):
 
         refresh = self.get_query_argument('refresh', default=False)
 
+        if refresh is not None and str(refresh).lower() in ['true', 't', '1']:
```
stefanv, line 1342: Are we missing a doc update to the PUT handler?

skyportal/handlers/api/photometry.py
```diff
@@ -1418,12 +1456,51 @@ def put(self):
...
+                        and str(duplicate.origin).strip().lower()
+                        not in ['none', '', 'nan', 'null']
+                    ):
+                        # we might have more than one datapoint at a given: jd, instrument_id, filter, and origin
+                        # that do have different flux values. We can't update the flux/fluxerr for all of them,
+                        # because this will conflict with the deduplication index we have and raise an error
+                        # so we only want to update the first one we find
+                        # for that, we compute a hash of the jd, instrument_id, filter, and origin everytime we update a row
+                        # store it, and make sure we don't update a row with the same hash twice
+                        duplicate_hash = hashlib.sha256(
```
stefanv, line 1473: How about rather a few calls to `hashlib.update`?

stefanv (commented):
Since this is a bit tricky, it's worth getting the documentation very clear. Otherwise, the code looks fine.

Theodlz:
Thanks a lot for the review @stefanv! Yes, documentation and comments could be better. Let me update that now!

skyportal/handlers/api/photometry.py
Theodlz: Why is that space needed for this argument but not the others?

skyportal/handlers/api/photometry.py
Theodlz: added

skyportal/handlers/api/photometry.py
Theodlz: removed hash calculation in favor of a simple string to avoid unnecessary computation

skyportal/handlers/api/photometry.py
Theodlz: I updated that comment, though the docstring should now explain what's happening (might not even need another comment there)

skyportal/handlers/api/photometry.py
Theodlz: updated, and as mentioned in another comment this is also explained in the docstring now

skyportal/handlers/api/photometry.py
stefanv: It should be everywhere, according to the numpydoc standard.

stefanv (approved):

PR #4771, opened 2024-02-21
Langchain, openai, and pinecone update

Theodlz:
All of these APIs have changed a ton since the summary and summary search features have been implemented, and recent attempts at updating these packages along with dependabot created some issues.

This PR upgrades the dependency we need and the code needed to use them.

I tested both summarization and summary search with pinecone, and everything now works as expected 👍

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -732,7 +732,6 @@ analysis_services:
     embeddings_store:
       summary:
         location: pinecone
-        environment: us-east1-gcp
         api_key:
         index_name: openai-skyportal-summary
         index_size: 1536
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 astroscrappy==1.1.0
 numpy==1.23.4
 scipy==1.11.3
-pandas==2.1.4
+pandas==2.2.0
 dask==2023.4.1
 geopandas==0.13.2
 intervals==0.9.2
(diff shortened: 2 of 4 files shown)


profjsb (changes requested):
It looks like the new API version of pinecone requires a `spec` keyword during `create_index` (this happens the first time a given API key references a pinecone instance that does not have an index yet.

```bassh
Traceback (most recent call last):
  File "services/openai_analysis_service/app.py", line 51, in <module>
    pinecone_client.create_index(
TypeError: create_index() missing 1 required positional argument: 'spec'
```
See
https://docs.pinecone.io/reference/create_index

Theodlz:
> It looks like the new API version of pinecone requires a `spec` keyword during `create_index` (this happens the first time a given API key references a pinecone instance that does not have an index yet.
> 
> ```shell
> Traceback (most recent call last):
>   File "services/openai_analysis_service/app.py", line 51, in <module>
>     pinecone_client.create_index(
> TypeError: create_index() missing 1 required positional argument: 'spec'
> ```
> 
> I think we're using the (free) `pod` version instead of the `serverless` version, which requires an `environment`, `pod_type`, etc. See https://docs.pinecone.io/reference/create_index

@profjsb thanks for the feedback. I'm running the pod version and created the index manually before running SP, which is probably why I didn't have that error. I'll look into it.

Theodlz:
@profjsb these changes worked for me. We should now also support serverless if one wants to do that (though that part is not tested).

I made sure to delete my index on my personal pinecone account, ran SP and it built the index just fine

profjsb (changes requested):
With a fresh installation, without adding the openAI key to the user, when trying the search_summary page I get an error `local variable 'user_openai_key' referenced before assignment`. Looks like this is happening in `handlers/api/summary_query.py", line 236

Theodlz:
@profjsb should be fixed now

profjsb (approved):

PR #4773, opened 2024-02-23
Remove image analysis (STDPIPE) feature

Theodlz:
We've seen how this just does not work, as STDPIPE grabs all of the resources of the machine and impacts the app's performance very negatively.

In the future, we might want to simply package that into an external analysis service that can POST photometry back to the app long with metadata about the results. But for now, we remove the integration of STDPIPE from SkyPortal.

A backup of all the STDPIPE-related code is available here: https://github.com/Theodlz/skyportal-stdpipe


mcoughlin (approved):
LGTM

Theodlz:
This is enough green tests to know that removing that feature didn't break everything. Merging

PR #4776, opened 2024-02-27
More flexibility in xml file ingestion

mcoughlin:
This PR makes voevent xml file more flexible.


Theodlz (approved):
LGTM. Very cool! I didn't notice that TestCoords thing, this is great, thanks!

PR #4778, opened 2024-02-28
2nd CandidateHandler: return candidates and not objs

Theodlz:
The current CandidateHandler is designed to return the objects associated with candidates that passed a set of filters + a bunch of constraints on candidates, associated source/obj, ...

For some applications, we want a handler that does not necessarily implement all of these parameters, but that returns the actual Candidate entries from the DB.

Though I'm not convinced about the handler's name and endpoint (open to suggestions), this handler allows us to do that.

An example usage of this API endpoint is figuring out what candidates/alerts passed a Kowalski filter, not just what ObjectID. Useful to map back to the exact alert packet and compute some stats.


skyportal/app_server.py
```diff
@@ -17,6 +17,7 @@
     AssignmentHandler,
     BulkTNSHandler,
     CandidateHandler,
+    CandidateCandidHandler,
```
mcoughlin, line 20: CandidateFilterHandler?

mcoughlin (approved):
Just a suggestion on the name, but it is up to you.

skyportal/app_server.py
Theodlz: Good idea, makes a little more sense

PR #4779, opened 2024-02-29
Pin baselayer to 56480544d9895a263782e8a0cf52e1e86c4785f1

Theodlz:
Pin new baselayer, that comes with some bugfixes with the query_accessible_rows() and related methods, to support sqlalchemy 2.0 syntax


mcoughlin (approved):
LGTM

PR #4794, opened 2024-03-05
Candidates Filter Form V2 Bugfix: program selection

Theodlz:
I spotted an issue on preview, where searching for groups with the search bar would cause incorrect groups to be selected. 

This PR stops using the form to handle that selected group state so we don't need to hack our way around using indexes to map to group ids which is a mess.


Theodlz:
Tested extensively locally (both manually and running the unit tests. Self merging

PR #4796, opened 2024-03-05
Candidates Scanning: Bring back pagination

Theodlz:
After adding the "infinity scroll" we saw an impact on performance (because of RAM usage) because the DOM renders everything even if its not in the view port.

To fix that, we used [react-viewport-list](https://www.npmjs.com/package/react-viewport-list) to only render what is in view.

However, the latter seems to prevent the "infinity scroll" by introducing a bug when appending new candidates to the existing list ( the order of the elements does not make sense anymore).

Until that is resolved, this PR removes the intersection observer API used to load more candidates in the background, and adds back some pagination with previous/next buttons at the bottom of the list.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "react-text-loop": "2.3.0",
     "react-to-print": "2.14.13",
     "react-transition-group": "4.4.5",
-    "react-viewport-list": "^7.1.2",
+    "react-viewport-list": "7.1.2",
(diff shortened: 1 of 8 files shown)


mcoughlin (approved):
LGTM

Theodlz:
SDSS's thumbnails service is down and this is affecting SkyPortal. Looking into it

PR #4797, opened 2024-03-05
GCN Observation plans: missing changes to existing plan name handler

Theodlz:
Missing backend changes, where the frontend has been merged already


Theodlz:
self merging

PR #4798, opened 2024-03-05
Candidates List: better grid proprotions

Theodlz:
* fix the annotations which origin text did not wrap properly, stealing width from the other elements of the grid
* more space for the photometry plot and annotations
* load 50 candidates at a time instead of 25, because this is cached anyway, and the photometry is fetched separately, lazy loaded when the candidate gets into the screen
* slightly bigger size for the detections. in the vega plot To make them even bigger I need to figure out how to also scale the error bars so their longer and still visible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "react-text-loop": "2.3.0",
     "react-to-print": "2.14.13",
     "react-transition-group": "4.4.5",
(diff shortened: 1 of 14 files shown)


mcoughlin (approved):
LGTM

PR #4799, opened 2024-03-06
FilterCandidateList: reset button and better css

Theodlz:
This removes the need for awkward pixel values when matching the height of the left and right sections of the form.

Also adds a nice reset button to reset form values to the defaults or the selected profile's values.


mcoughlin (approved):
LGTM

PR #4801, opened 2024-03-06
Candidates Filter Form: fix redshift input, fixed program selection height

Theodlz:
Turns out its really hard to have the left section of the form match the right section's height without letting it grow forever, which means more program = a longer form and not a scrollable one. This PR adds a fixed height.

Also fixes the redshift min and max inputs


mcoughlin (approved):
LGTM

PR #4802, opened 2024-03-08
Sources Handler: make get_sources_experimental the new default

Theodlz:
I've been playing around with both the existing version and the "experimental" version of the sources handler in the past couple of weeks.

I see no difference in the output except PURE SPEED ⏩ ⏩ ⏩ ⏩ ⏩ ⏩ .

I would like to try using the new code as the default when calling the sources GET API endpoint, to see how positively/negatively that impacts using the webapp.

Some things are still slow, and will always be slow, especially as a super admin that has access to everything, there's just a ton of data to look at. But overall, this should be much quicker. Especially when fetching group sources, which is super slow on Fritz right now.


mcoughlin (approved):
Let's gooooooooooo.

PR #4804, opened 2024-03-08
Empty altdata in follow-up triggering

mcoughlin:
This PR checks for empty altdata in the generic API allocations.


Theodlz (approved):
LGTM

PR #4806, opened 2024-03-08
Patch photometry ID

mcoughlin:
This PR fixes the patch photometry call.


Theodlz (approved):
oupsssss this one's on me 💩

PR #4807, opened 2024-03-09
Front-end AB and Vega

mcoughlin:
This PR adds a switch between AB and Vega.


Theodlz (changes requested):
Looks good. As discussed on Slack button placement needs to be fixed first, and I'll take care of it later today or tomorrow

Theodlz (approved):
LGTM!

Theodlz:
Ah actually wait, I see something wrong in the new component

PR #4812, opened 2024-03-11
Edit Source Coordinates: update thumbnails

Theodlz:
Make sure we delete old thumbnails (ps1, sdss, ls) from an object when updating its coordinates. 

The queue will regenerate the thumbnails.

@mcoughlin editing coordinates seems pretty risky to me in general. Wondering if we should have a logic where we do not allow that to happen if the source has corresponding candidates (i.e. comes from alert data where the position can't really be wrong and most definitely shouldn't be edited!).


mcoughlin (approved):
LGTM! And yeah, if you want to add the candidate logic here, I would be ok with that.

mcoughlin:
Thanks @Theodlz!

Theodlz:
Preventing the update of coordinates for objects with candidates seems to make a bunch of tests fail somewhat randomly (they pass locally, so I guess that another test somewhere adds a candidate that is not deleted.
Might want to add a fixture like "public_group_no_candidates" that I can use knowing that other tests wont add stuff to the source I am using

PR #4813, opened 2024-03-12
Information on front-end pages

mcoughlin:
This PR adds information on front-end pages in the app.


Theodlz (approved):
That’s a great idea! LGTM. Just wondering if this new file needs to referenced somewhere for it to appear on the menu.

PR #4814, opened 2024-03-12
ZTF Forced Photometry: Flux weighted position

Theodlz:
Adds a dropdown to pick what position to use: the object's, flux weighted alerts only (ignoring FP), and flux weighted ZTF alerts only (ignoring FP).

### DIFF
--- a/skyportal/facility_apis/ztf.py
+++ b/skyportal/facility_apis/ztf.py
@@ -148,7 +148,7 @@ class ZTFRequest:
         }
 
         ra, dec = None, None
-        position = request.payload.get("position", "Source")
+        position = request.payload.get("position", "Source (object's coordinates)")
         if "Flux weighted centroid" in position:
             # here we query the database to get the all the photometry rows for the object that:
             # 1. belong to an instrument with data stream(s)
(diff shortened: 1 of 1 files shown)


skyportal/facility_apis/ztf.py
```diff
@@ -130,19 +130,98 @@ def _build_forced_payload(self, request):
...
+                    and not np.isnan(p.ra)
+                    and not np.isnan(p.dec)
+                    and len(p.streams) > 0
+                    and all(constraint(p) for constraint in additional_constraints)
+                ]
+                if len(photometry) > 0:
+                    # calculate the weighted centroid based on the object's photometry
+                    ras, decs, flux, fluxerr = np.array(photometry).T
+                    snr = flux / fluxerr
+                    ra, dec = np.sum(ras * snr) / np.sum(snr), np.sum(
```
mcoughlin, line 204: abs(snr)?

mcoughlin (approved):
Just one question on SNR.

skyportal/facility_apis/ztf.py
Theodlz: Ah yes you are right, plus I said I would use that too, yikes.

Also wondering if I should put a cut on SNR to remove detections of low significance

mcoughlin: Ah yeah abs(snr) > 3 at least.

mcoughlin (commented):
Yeah SNR cut good idea.

skyportal/facility_apis/ztf.py
Theodlz: added

PR #4824, opened 2024-03-14
TNS Retrieval: bugfixes

Theodlz:
* if the TNS obj exists in the DB but isn't saved as a source to the public group, save it
* fix issue created by one of the last commit in the TNS V2 PR, where I exchange the name of 2 variables for clarity but forgot to do so for 2 lines.
* use websockets to update the frontend when adding a TNS name to an existing source or it's TNS source counterpart.
* keep the start_date of the TNS query for new objects identical until we do get new objects.


mcoughlin (approved):
LGTM

PR #4825, opened 2024-03-14
TNS Submission: bugfixes and stricter submission rules

Theodlz:
* make it mandatory to specify instruments when adding a robot, and must be specified too before being able to submit to TNS
* validate the source obj_id before submitting using the source_group_id on TNS (basically, what survey the source comes from) and the survey's naming convention.


mcoughlin (approved):
LGTM

PR #4830, opened 2024-03-15
Photometry Plot: show nearby sources (possible duplicates)

Theodlz:
On Fritz, we'll soon have both ZTF and WINTER alerts ingested in our system, and we need tools to easily show the photometry of multiple nearby sources in one plot.

This PR is an attempt at that, where we leverage the redux capability to store photometry from different objects to concatenate photometry and show multiple objects at once on the plot.

I guess this is what the PhotometryCopyHandler tried to achieve, but there is also value in merging the photometry only for display purposes, and not always requiring duplicated photometry in the DB. But, being able to copy the photometry from one object to another in the DB is useful too as it also updates the photometry statistics used for querying.

Also just so the reviewer knows, this PR will be followed by an associated Fritz PR, where we add another annotation or source plugin that let's users ingest ZTF sources close to a source in the database, from Kowalski. The idea here is that a source from WINTER could have a ZTF counterpart but that never made it through a Kowalski filter, so being able to add ZTF sources around an object from the click of a button is very useful, as you can then plot them on one object thanks to this new feature.


mcoughlin:
@Theodlz the original purpose of the copy handler was in cases where you had a ZTF and a WINTER source, which clearly have different permissions, to easily "merge" the two sources in some sense when useful (in case of the GW searches).

Theodlz:
> @Theodlz the original purpose of the copy handler was in cases where you had a ZTF and a WINTER source, which clearly have different permissions, to easily "merge" the two sources in some sense when useful (in case of the GW searches).

Ok I see, yes when phot stats are needed that makes sense. But the reality is that in most cases folks just want to see the data, and also the current copy photometry handler does not work most of the time, because of data access restrictions precisely. It requires you to copy the photometry to specific groups that I guess the other source is saved to, and most of the times said groups don't have the data access to have that photometry so the transfer fails (without showing error or warning).

I think we can have both features, show it on the fly or properly "merge" the photometry of 2 sources. But the handler to copy photometry and its frontend need some love. What I noticed (aside from the fact that it did not work as intended most of the time), is that users (or myself) couldn't really figure out if it was copying TO or FROM the source.

mcoughlin:
@Theodlz sounds good.

skyportal/handlers/api/source.py
```diff
@@ -289,10 +290,22 @@ async def get_source(
     duplicates = session.scalars(
         Source.select(user).join(duplicate_objs, Source.obj_id == duplicate_objs.c.id)
     ).all()
-    if len(duplicates) > 0:
-        source_info["duplicates"] = list({dup.obj_id for dup in duplicates})
-    else:
-        source_info["duplicates"] = None
+    # ironically, we might have duplicates of duplicates because we can have multiple sources for the same obj_id
+    # so, deduplicate to only keep one source per obj_id, for that we take advantage of dictionaries and their unique keys
+    # TODO: replace this query by something better where we have only one entry per obj_id
```
mcoughlin, line 295: Anything to do with the TODO here?

static/js/components/PhotometryPlot.jsx
```diff
@@ -851,7 +895,7 @@ const PhotometryPlot = ({
       setLayouts(newLayouts);
       setInitialized(true);
     }
-  }, [photometry, dm, filter2color, defaultVisibleFilters]);
```
mcoughlin, line 854: Do you still want filter2color since it's in the useEffect?

mcoughlin (approved):
A few small questions.

skyportal/handlers/api/source.py
Theodlz: Not really the query is fine as is, not worth spending more time on it (unless prod goes bananas). Just updated the comment to look not so crazy

static/js/components/PhotometryPlot.jsx
Theodlz: Good point. Just added it back

PR #4836, opened 2024-03-18
Starlist bugfix: source_id col

Theodlz:
Looks like maybe one of the API services we grab data from when generating starlists has changed the 'source_id' column to 'SOURCE_ID'. 

This just makes sure that we are not dependant on capitalization anymore, to fix the current problem and maybe avoid another one.

it was hot-patched in production.


mcoughlin (approved):
Hot patch for the win.

PR #4840, opened 2024-03-22
TNS V2: photometry options

Theodlz:
Here we add the concept of photometry options, and polish some of the reporting behavior.

Namely, we make it possible to send reports with only one detection (as some other survey might do, or ZTF when only one detection is available). The default behavior can be sent on the robot (and will be what's used for auto reporting), and overwritten from the frontend when manually submitting to TNS from the source page.

We also make it mandatory to specify instrument(s) when submitting (to restrict the photometry that is sent), avoid sending forced photometry, and make it possible to overwrite the streams that the robot is set to use for auto-reporting.

With these changes we can enforce some sensible defaults when creating reports automagically, but allow enough flexibility from the frontend to make exceptions when needed.

TODO: migrations


services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -306,33 +338,66 @@ def build_reporters_string(submission_request, source, tnsrobot, session):
...
+            # sort alphabetically (A -> Z) to ensure consistent ordering
+            affiliations = sorted(affiliations)
+
+            affiliations = ", ".join(affiliations)
+            reporters.append(f'{author.first_name} {author.last_name} ({affiliations})')
+
+        reporters = ", ".join(reporters)
+
+        # acknowledgments are added to the end of the reporters string if they exist (optional)
+        if tnsrobot.acknowledgments not in [None, ""]:
```
mcoughlin, line 391: This should be ' ' or '\n'?

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -764,11 +855,13 @@ def process_submission_request(submission_request, session):
...
         )
 
-        # FETCH THE PHOTOMETRY (FILTERED BY INSTRUMENTS)
+        # FETCH THE PHOTOMETRY (FILTERED BY INSTRUMENTS, NO FP)
         photometry = session.scalars(
             Photometry.select(user).where(
                 Photometry.obj_id == obj_id,
                 Photometry.instrument_id.in_(instrument_ids),
+                # make sure that the origin does not contain 'fp' (for forced photometry)
+                ~Photometry.origin.ilike('%fp%'),
```
mcoughlin, line 864: Maybe add a note why no forced photometry here.

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -862,6 +984,7 @@ def service(*args, **kwargs):
                     log(f"Error rolling back session: {str(e)}")
                 else:
                     try:
+                        traceback.print_exc()
```
mcoughlin, line 987: Useful?

static/js/components/TNSRobotsPage.jsx
```diff
@@ -1050,7 +1094,7 @@ const TNSRobotsPage = () => {
     const tnsrobot = tnsrobotList[dataIndex];
     return (
       <Tooltip
-        title={`Added at the end of the author list, e.g. 'First Last (University) ${
+        title={`Added at the end of the author list, e.g. 'First Last (Affiliations) ${
```
mcoughlin, line 1097: Affiliation(s)

mcoughlin (approved):
Small comments.

services/tns_submission_queue/tns_submission_queue.py
Theodlz: You mean on the join? Should be a comma and a space, there is no new line on TNS reports as far as I know

services/tns_submission_queue/tns_submission_queue.py
Theodlz: Yes! I want to see exactly where errors come from when we run this in prod. might delete later

static/js/components/TNSRobotsPage.jsx
Theodlz: Thanks!

services/tns_submission_queue/tns_submission_queue.py
Theodlz: added

services/tns_submission_queue/tns_submission_queue.py
Theodlz: Ah maybe you meant the warning that's a few lines under that. If so, just fixed it

PR #4843, opened 2024-03-26
TNS Robots: Including testing in unique constraint

Theodlz:
That way we can have a copy of a TNS robot as long as that copy is in testing mode and not the existing bot (or vice versa).

Useful to have a copy of a bot in testing mode to test the auto-reporting while the real bot can still be used for manual reporting only.

TODO: try to improve the TNS to existing source matching, because the conesearch results from TNS contain no distance information or position, and we've matched some older TNS sources with newer ones. Without running a billion extra API calls, this could be as simple as just using a smaller conesearch radius, from 2 to 1 arcsec or less.


mcoughlin (approved):
LGTM

PR #4844, opened 2024-03-27
TNS interactions: various improvements

Theodlz:
* don't show sources from the public group in the "most recent sources" widget. This makes sense now that we save TNS sources to that group, but makes sense in general for all of the instances of SkyPortal we know of. Folks have their different groups and do not save sources to the public group for everyone to see. Sources are saved to individual groups that want to see it to not force everyone else to see sources they don't care about, but data products can of course be shared to the public group. A checkbox in the options of the widget can be set to True to show public sources again if necessary.
* Fix issue with the TNS retrieval service, erroring out and stopping after looping once.
* add missing DB session commit after creating auto reports from the source handler. From the source page it already worked because it uses the source_groups handler, but it didn't work when saving directly from the candidate scanning page.
* remove the download lightcurve button from the photometry plot, to avoid having to maintain this + the download button on the photometry table (just good to avoid duplicated features in general). Folks are now used to download data from these MUI tables, so keeping this one (with the addition of flux space information) makes more sense.
* when looking for the TNS name of an object using it's coordinates with a TNS cone search, loop over all of the results to get their RA/DEC, and compute a great circle distance to use the closest match as the name. This is because TNS conesearches do not return spatial information for some reason, and we ended up matching incorrect names with a 2 arcsec cone search.
* Differentiate some of the errors as warnings. They are still technically "errors" that prevent creating or sending the report, but we want to display some of these in orange on the frontend, not bright red, as not being able to send an automatic report is not critical if there is a good reason for it.
_Might have some tests to fix because of the widget change, we'll see_


mcoughlin (approved):
LGTM

Theodlz:
Nightfall is near, deploy is required, merging I will

mcoughlin:
Feel the rhythm of the night.

PR #4845, opened 2024-03-28
Sources widgets TNS name chip + archival auto reporting

Theodlz:
Absolutely no one seems to use the quick view option. The window that opens simply does not have enough information to make it useful, the display isn't great as is, and if a user wants the information that they need, they always open a source page.

Instead, we can retrieve that space to display TNS names of sources on the dashboard, very useful to know at a glance which of the recently saved sources from your groups have been TNS-ed or even classified as SN on TNS already.

I did not remove the quick view component itself, so we can easily bring it back in an improved format later on if we'd like.

This PR also adds a little regex to prevent adding objects with weird names. Talking with someone who had installed SkyPortal for the first name, the first thing they tried was adding a name with special characters which broke the source page instantaneously 😓. With these changes we only allow letters, numbers, underscores, dashes, colon and semi-colon, which should be more than enough and shouldn't cause any trouble.

UPDATE:
In production BTS folks want auto-reports to switch to archival reports when no non-detections before the first detection can be reported.  However, that might not be the case for everyone using TNS robots to submit data from Fritz. So, just like for the minimum number of detections required, we make that an option that can be specified when editing the TNS robot from the TNS robots page. The archival comment is auto-generated to say something like: "No non-detections before first detection", or "No non-detections before first detection in ZTF Public, ... alert streams" if the bot is set to use specific alert streams for auto-submission. Manual reporting from the source page still requires to user to pick archival or not, not automatic switching here


mcoughlin (approved):
LGTM

Theodlz:
@mcoughlin re-requesting review because the PR changed quite a bit as I added this auto archival report stuff

mcoughlin (approved):
LGTM

PR #4847, opened 2024-03-28
TNS Retrieval: Watcher lifetime

Theodlz:
Looks like some TNS responses (from both success and failures) came back as non-JSON compliant..., which would kill our TNS watcher service and stop us from ingesting TNS sources in the app after any error is encountered. 

This PR wraps all `respoonse.json()` statements in a try except to use the `response.text` if we can't read the JSON, and also wraps the entire TNS watcher service's white loop content in a try except to avoid killing the service if we get any error.


mcoughlin (approved):
TNS is good.

Theodlz:
> TNS is good.

👁️ 👁️ 
  👄

PR #4848, opened 2024-03-29
Observation plan queue bugfix:

Theodlz:
I noticed on the frontend that all the plans that have been generated recently where stuck in "running" state. What looks like happened is that we marked the plan as complete, and the request too but for some reason marking the request as complete in the queue did not work.

This PR fixes that with the pretty simple addition of a session.merge(), but also adds code in the facility_apis/observation_plans to reprocess plans or mark them as failed or complete when needed.


mcoughlin (approved):
LGTM

PR #4849, opened 2024-03-29
TRT API

mcoughlin:
This PR adds TRT as an API option.


alembic/versions/468435875f30_trt.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+
+    op.execute(
+        """
+alter type "public"."followup_apis" rename to "followup_apis__old_version_to_be_dropped";
+
+create type "public"."followup_apis" as enum ('MMAAPI', 'GENERICAPI', 'SLACKAPI', 'ATLASAPI', 'GROWTHINDIAMMAAPI', 'KAITAPI', 'SEDMAPI', 'SEDMV2API', 'IOOAPI', 'IOIAPI', 'SPRATAPI', 'SINISTROAPI', 'SPECTRALAPI', 'FLOYDSAPI', 'MUSCATAPI', 'NICERAPI', 'PS1API', 'UVOTXRTAPI', 'UVOTXRTMMAAPI', 'TESSAPI', 'TRTAPI', 'WINTERAPI', 'ZTFAPI', 'ZTFMMAAPI');
```
Theodlz, line 26: I know it works but... do we really need to do that to migrate??? I remember that we could simply add the new API in the correct location in the existing enum, there's a bunch of examples of that in previous migrations, seems cleaner and less risky than dropping the existing one everytime.

Theodlz (commented):
Just one comment about the migrations, not a huge fan of dropping that table/enum when we can just add the new API class in the existing enum

alembic/versions/468435875f30_trt.py
mcoughlin: @Theodlz Ooops, I forgot. Should be updated now.

Theodlz (approved):
LGTM good sir

PR #4850, opened 2024-03-29
TNS submission: confirm report processing on TNS

Theodlz:
When sending a report, TNS gives us a report_id. Reports are processed in their server async (meaning after we sent them), so we can use that report_id (here in our data model called `submission_id`) to query TNS and validate that the report resulted in the creation of a new object, or was posted on an existing one.

This PR adds a little service that loops over submitted reports and tries to query TNS for the report's status. The API docs is not clear enough for me to catch failures to send reports, so for now this only figures out if it was successful. I'm hoping that some reports fail at some point so I can query their system and figure out what a failed response looks like...

When a report is "validated" we make the tns_retrieval_queue fetch the information of the newly added TNS object, adding the correct TNS name and information to the object that we just submitted.

### DIFF
--- a/services/tns_retrieval_queue/tns_retrieval_queue.py
+++ b/services/tns_retrieval_queue/tns_retrieval_queue.py
@@ -310,13 +310,18 @@ def process_queue(queue):
                     if tns_source in [None, "None", ""]:
                         raise ValueError(f'{task.get("obj_id")} not found on TNS.')
                     tns_name = f"{tns_prefix} {tns_source}"
-                elif (
-                    task.get("tns_source") is not None
-                    and task.get("tns_prefix") is not None
-                ):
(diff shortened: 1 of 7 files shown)


services/tns_retrieval_queue/tns_retrieval_queue.py
```diff
@@ -367,6 +372,25 @@ def process_queue(queue):
                     log(f"Error getting TNS data for {tns_name}: no reply in data")
                     continue
 
+                try:
+                    msg = (
+                        tns_source_data.get("name", {})
+                        .get("110", {})
```
mcoughlin, line 378: Maybe comment what this 110 is?

mcoughlin (approved):
LGTM

PR #4851, opened 2024-03-29
Source Views Dashboard Widget: don't count token source views

Theodlz:
Some folks keep querying sources for months using scripts (at like a 1-hour cadence).

Though maybe I should just BAN ALL OF THEM USERS, given that we know how much frontend is king for SP, it makes total sense to only show frontend source views in the frontend, and not those registers by API calls.

This PR doesn't just that by ignoring token source views when querying for views in the internal API used by the dashboard widget.


mcoughlin (approved):
BAN ALL USERS

PR #4857, opened 2024-04-01
TNS Robots management: sort group users, groups, and coauthors alphabetically

Theodlz:
as the title suggests we make sure that items in lists of things to add/delete/edit are all ordered alphabetically to make it easier to find what you are looking for in prod.


mcoughlin (approved):
LGTM

PR #4858, opened 2024-04-02
TNS submission service(s) bugfix

Theodlz:
* move imports into their respective and not at the top of the file, as models are initialized to a session, so we can't share them between threads (which somewhat randomly crashes the service that sends reports or the one that verifies them right now). But we should be able to move imports in the methods where they are needed which will use the session that's in scope for a given thread calling the method. Looks like this fixed these issues locally.
* add the tns_name of each object in the TNS submissions lists


mcoughlin (approved):
LGTM

PR #4861, opened 2024-04-02
MUI data table refactoring of analysis services component

jinglin-l:


Theodlz:
Awesome @jinglin-l 🚀 . I see you've git added baselayer though, which is likely a mistake (we only git add new versions, but my guess is you pulled changes from SkyPortal, which doesn't update baselayer as it is a submodule). Can you remove that change? To do it you can just check what's the version of baselayer that's pinned on `skyportal/main`, git checkout to it, and git add this one and commit. Thanks!

jinglin-l:
> Awesome @jinglin-l 🚀 . I see you've git added baselayer though, which is likely a mistake (we only git add new versions, but my guess is you pulled changes from SkyPortal, which doesn't update baselayer as it is a submodule). Can you remove that change? To do it you can just check what's the version of baselayer that's pinned on `skyportal/main`, git checkout to it, and git add this one and commit. Thanks!

baselayer is removed! thanks!

Theodlz:
I see that you replaced the analysis service list by this mui datatable. Maybe it's just git not showing me this, but I just wanted to make sure that this `AnalysisServiceList` component is still used somewhere, in which case we might want to delete it. **OR**, maybe that's the one that should really be modified to use a MUI datatable?

jinglin-l:
> @jinglin-l I see that you replaced the analysis service list by this mui datatable. Maybe it's just git not showing me this, but I just wanted to make sure that this `AnalysisServiceList` component is still used somewhere, in which case we might want to delete it. **OR**, maybe that's the one that should really be modified to use a MUI datatable?

Good point. It looks like the AnalysisServiceList component has specific logic for handling user interactions (like deletion with confirmation) and custom styling.

Given this, integrating the MUIDataTable into the AnalysisServiceList would probably be the better thing to do to ensure that these customizations are preserved. I'll work on that now.

Theodlz (commented):
LGTM @jinglin-l ! This is great. I'm very very happy with the code. I do have one more request if that's ok with you. Part of the refactoring here, and why we want to use a MUI datatable is readability. Sadly keeping the form to create a new analysis service next to the table means very little width for the table to use and look good on smaller screens.

One thing we've been doing for most of these (you can see it on the list of sources, and GCN events), is add a `+` icon in the custom toolbar of the datatable, that opens a MUI Dialog window that contains the form. On - successful - submit of the form, the dialog closes.

Do you think that is something you could add in this PR? That would wrap up the issue completely which would be awesome!

Theodlz:
Hi @jinglin-l! Any updates on this? Not sure you saw my comment earlier. Once again, I'm happy merging this already, and adding the `+` button in a following PR. Just let me know what works best for you!

jinglin-l:
Does this look ok? I put the form in a dialogue, added a title, and added the "+" icon to the MUIDataTable toolbar. 

Screenshot 2024-07-30 at 5 53 44 PM

Screenshot 2024-07-30 at 5 55 20 PM

Theodlz:
Hi @jinglin-l, unfortunately this has been open for a little too long, so we ended up making these changes ourselves while refactoring other pages. You can see them on the main branch.

Closing this one for now. Would you like to help with another issue? I have a couple in mind

jinglin-l:
@Theodlz Sure, send them my way! Sorry about that. I was abroad for quite a while, and realized that I didn't push.

PR #4862, opened 2024-04-02
GitHub actions: back on track (part 1)

Theodlz:
Not sure how many PRs I'll need to get this back on track (some of the flaky tests give me no information whatsoever as to why they are really failing, which makes it hard to fix them), but this is a first set of fixes for a few tests.

The tests on comments are the most annoying, looks like the comments on GCN fail (again, don't know why really) and that cascades into the app not being able to run a single test. Finishing the ThreadSession PR on baselayer to run these GCN-related processes in a separate session to avoid potentially killing the same session as the one used by the API could help for sure, but not guaranteed either.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -68,7 +68,7 @@ humanize==4.8.0
 xarray>=21.1
 simsurvey==0.7.5
 noaodatalab==2.20.0
-black==24.2.0
+black==24.3.0
 click==8.1.6
 arviz==0.16.1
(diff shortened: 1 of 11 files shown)


mcoughlin (approved):
Fight that fight.

Theodlz:
Ok so this fixed all of the broken tests. Some are still flaky but I'll tackle the flakiness in a dedicated PR (part 2 of this one).

PR #4867, opened 2024-04-09
Source Page Coordinates: adjusted using photometry

Theodlz:
This PR adds a new handler called when accessing a source page. This handler recomputes a source's position using it's photometry (data access agnostic, like the photometry statistics). The improved position will replace the original one on the source page if they are not the same. A tooltip lets the user know which position he is seeing on hover of a source's coordinates.

Just wondering if I call it "ObjPositionHandler" or "ObjCoordinatesHandler", but that aside the implementation shouldn't change too much.

It re-uses Josh's code to compute the best position to use for a finding chart or starlist, with the addition of some extra filters removing forced photometry and low SNR detections. Filters that we might want to add to the underlying method one of these days.


mcoughlin (approved):
LGTM

PR #4868, opened 2024-04-09
TNS submissions DBSession bugfix

Theodlz:
This is an attempt at scoping the DBSession appropriately to avoid issues with shared sessions between 2 threads. As DBSession is scoped with a contextvar that is None, it looks like it ends up having one session shared between threads. So here, we defined the relevant contextvar with a random UUID (just like we do for an individual API request) to try to see if sessions are different for each thread which is what we want

Before I tried to import DBSession in each thread which didn't help, but I kept that here which means that the contextvar is defined before the DBSession import, which I'm not sure works. 

If this does not work, I might want to switch to one import of DBSession on top of the file as usual, and simply set the context in the main method/logic of each thread before opening a session with a contextmanager

### DIFF
--- a/services/tns_submission_queue/tns_submission_queue.py
+++ b/services/tns_submission_queue/tns_submission_queue.py
@@ -2,6 +2,7 @@ import json
 import re
 import time
 import traceback
+import uuid
 from threading import Thread
 from urllib.parse import urljoin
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Cool!

Issue #4832, opened 2024-03-15
Photometry Plot: Forced photometry toggle


Theodlz:
On the candidate scanning page, we already have a MUI switch to hide/show forced photometry from the plots. We would like to have the same feature on the source page's photometry plot, right next to the existing switch to hide/show non-detections.

Given that almost identical code exists on the candidate page, this should be very straightforward. The only addition there is to make sure that we keep hiding the FP data when clicking on the legend. There is custom logic there for the existing non-detections toggle, so again very straightforward to do the same for forced photometry (maybe even easier, as this is based on origin and not mag or flux).

Theodlz:
@theobonzi can't assign you on this for some reason, so doing it in a comment here

PR #4869, opened 2024-04-11
Photometry Plot: Forced photometry toggle #4832

theobonzi:
This Pull Request fix this issue:  Photometry Plot: Forced photometry toggle #4832

Introduces a toggle switch on the Source page to control the visibility of forced photometry data on the photometry graph. This allows the user to hide or show forced photometry on photometry graph.


Theodlz (changes requested):
I think there's some kind of confusion. Upper limits are non-detections, that has nothing to do with forced photometry. So the feature here is exactly the toggle we already have...

Theodlz (changes requested):
Double-clicking a trace in the legend to hide everything works just fine but when double-clicking it again to show every the deactivated forced phot trace comes back. I think you are missing some code in the legend double-clicking logic (towards the end of the file).

Theodlz:
@theobonzi any updates?

theobonzi:
> @theobonzi any updates?

It's okay, I fixed it.

static/js/components/PhotometryPlot.jsx
```diff
@@ -1005,6 +1015,22 @@ const PhotometryPlot = ({
     }
   }, [showNonDetections]);
 
+  useEffect(() => {
```
Theodlz, line 1018: There is still a bug @theobonzi. If the datapoint is an upper limit with forced photometry origin, if the upper limits are deactivated and the forced photometry too, reactiving only the forced photometry should NOT show the upper limits again (since they are still deactivated).

See that example here:
image

Theodlz (changes requested):
There is still a bug @theobonzi. You didn't account for what happens if a datapoint is an upper limit with forced photometry as the origin.

I know you love your ML project but make sure to test your code well! Otherwise it's time and momentum lost for both you and I. That can happen of course, but try your best to test things a bit.

Theodlz (approved):
LGTM!

Theodlz:
@theobonzi you've pinned another version of baselayer (by accident maybe?), can you just remove that super quick before I can merge? Thanks!

PR #4871, opened 2024-04-12
NewPhotometry ISOT date input - bugfix

Theodlz:
Looks like on some browsers it was using the locale and not the UTC, by assuming the date string was in local time.


Theodlz:
minor change, self merging

PR #4873, opened 2024-04-14
update Vega vs AB in photometry table

mcoughlin:
This PR improves handling of columns / Vega vs AB in photometry table.


Theodlz:
@mcoughlin can you tell me a little more about what this does exactly? I mean I understand the code, but not sure why some of it is there (like sending the magsys from server to client, and the thing with the `openColumns`)?

mcoughlin:
Columns would reset after every interaction. And Magsys was not respected when editing

Theodlz (approved):
LGTM boss

Theodlz:
@mcoughlin looks like a test is failing with these changes: `skyportal/tests/frontend/sources_and_observingruns_etc/test_sources.py:783 in test_show_photometry_table`

PR #4874, opened 2024-04-14
Astropy 6.0 breaking changes: MaskedArray -> MaskedNDArray

Theodlz:
Looks like astropy is using it's own datatypes for some of the masked arrays now, so we need to look for those too when we want to convert missing values to None, otherwise the response can't serialize that astropy masked dtype.


mcoughlin (commented):
LGTM

mcoughlin (approved):

PR #4875, opened 2024-04-14
Astropy 6.0 - Observing run ephemeris bugfix

Theodlz:
Just like for the instruments and telescopes, we check that the results from ephemeris calculations are not MaskedArray OR MaskedNDArray.


mcoughlin (approved):
LGTM

PR #4876, opened 2024-04-15
ObservingRunHandler GET: handle rise/set vectorized masked array

Theodlz:
https://github.com/skyportal/skyportal/pull/4875 stopped the handler from crashing, but looks like as soon as one target is not visible, since we reshape the masked array we end up with all the value being 0-dimensional masked arrays, even those that might have a value that's usable.

This PR deals with those appropriately, checking if the 0-dimensional array (which is just a scalar) doesn't have its only value masked.


mcoughlin (approved):
LGTM

PR #4880, opened 2024-04-17
GCN notes table

mcoughlin:
This PR adds a table with GCN notes to Source page.


static/js/components/Source.jsx
```diff
@@ -537,6 +538,31 @@ const SourceContent = ({ source }) => {
           </AccordionDetails>
         </Accordion>
       </Grid>
+      <Grid item xs={12} lg={6} order={{ md: 4, lg: 3 }}>
```
Theodlz, line 541: 4 and 3 seem pretty high up on the source page for this feature, is that intended?

mcoughlin: @Theodlz Not necessarily. I think it could be moved elsewhere. It was just flagged as an issue at the MMA call that the notes were not made (highly) visible on the source page.

mcoughlin:
@Theodlz Could you help finish this one off? It was a request after 0422 that we see notes written in the GCN scanning also on the original source page.

Theodlz:
> @Theodlz Could you help finish this one off? It was a request after 0422 that we see notes written in the GCN scanning also on the original source page.

Thanks for reminding me, sorry I've been trying to catch up with the accumulated workload after these 2 weeks of travels. Will review and fix what needs to be fixed tomorrow late afternoon.

Theodlz (approved):
LGTM :) Though I'm a little biased now you can't trust me

PR #4881, opened 2024-04-17
TNS Submission: "report not found"

Theodlz:
Looks like in production, we sent a report to TNS which gave us a submission ID back, but using it to look for the submission status we get a "404" error, as if they acknowledged receiving it and then... lost it?

Whatever the reason for that, these changes set such a failed submission back to pending after a minute, so we can retry submitting to TNS.


mcoughlin (approved):
It's a black hole.

PR #4882, opened 2024-04-20
GaiaQuery class: verify that server is accessible

Theodlz:
Currently, we launch a job to the server and if it times out we use the backup server. Except that with the python package we are using to query Gaia, the timeout is one minute, which of course exceeds our max timeout expected from our webapp: 30s.

I do not see a way to use a different timeout with that package, so this PR adds a quick line that runs a HEAD HTTP call to the server URL to see if it's up and running, with no more than 5 seconds.

The issue is that we can't verify that this works until their server is up and running again, but **what I can say for sure is that with this addition we connect to the backup quickly enough to get starlists and finding charts, which are all failing right now.**

### DIFF
--- a/skyportal/utils/offset.py
+++ b/skyportal/utils/offset.py
@@ -90,7 +90,7 @@ class GaiaQuery:
             requests.head(
                 "http://" + str(g._Tap__connHandler.get_host_url()), timeout=5
             )
-            
+
             q = f"SELECT TOP 1 ra, dec from {self.main_db}.gaia_source"
             job = g.launch_job(q)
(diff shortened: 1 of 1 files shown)


Theodlz:
It's not the best solution however (though the best solution could be that + something else), because we still need to wait 5 sec everytime. Maybe we should "cache" that the main server failed and avoid it for like an hour or smth, because right now getting the starlist is still very slow because of that.

mcoughlin (approved):
LGTM

Theodlz:
Not sure this is the best way to do this. Closing for now

Issue #4884, opened 2024-04-21
Updating coordinates leads to Thumbnail permission deletion error


mcoughlin:
![image](https://github.com/skyportal/skyportal/assets/2291947/5a203d16-df93-4280-b5c3-c2d483e0957d)

When you are not an admin, updating the coordinates of a source you made still leads to a Thumbnail permission deletion error. This should be bypassed somehow.

Theodlz:
On it

PR #4885, opened 2024-04-22
Obj Coordinates update: thumbnails delete permission fix

Theodlz:
Addresses #4884, where a user without delete permissions for thumbnails couldn't remove let the handler remove the public_url ones (sdss, ls, ps1) when updating the coordinates of a source.

This PR fixes that by adding a new decorator named `run_async`. This decorator can take any method and its argument, and will run them in an async process where it will set a unique context for the DBSession created in the method passed to the decorator. This ensures that we can run async processes which DBSessions aren't shared with the main thread or other threads, and therefore don't get killed in the middle of an async operation when the main thread or other threads use the same session. Here, each thread has it's own scope for its session, avoiding this problem entirely.

If this new decorator proves to work as expected, we'll want to add it everywhere we start an async thread, which is a lot of places but here are those that come to mind as a priority:
- follow-up requests
- GCN events

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -4,14 +4,12 @@ import io
 import operator  # noqa: F401
 import time
 import traceback
-import uuid
 from json.decoder import JSONDecodeError
 
 import re
 import arrow
 import astropy
 import astropy.units as u
-import asyncio
 import conesearch_alchemy as ca
 import healpix_alchemy as ha
 import matplotlib.pyplot as plt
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/source.py
```diff
@@ -114,6 +118,44 @@
 Session = scoped_session(sessionmaker())
 
 
+def run_async(func, *args, **kwargs):
```
mcoughlin, line 121: Can we put this in utils/ or models/ or somewhere where it might be useful beyond source.py?

mcoughlin (approved):
Might be useful to put that new function somewhere where it can be used easily again.

skyportal/handlers/api/source.py
Theodlz: Yes, I was only trying to figure out where to put it! Will move it somewhere better

Theodlz: done @mcoughlin, also fixed some stuff here and there, and made sure the position shown on the frontend updates when we modify it in the DB.

PR #4887, opened 2024-04-23
field ID in manual plan

mcoughlin:
This PR queries DB for field ID in case using different skyportal for plan.


mcoughlin:
Trivial fix. Self-merging.

PR #4889, opened 2024-04-23
No detections in the source queries

mcoughlin:
This PR allows for no detections in the source queries.

### DIFF
--- a/skyportal/handlers/api/sources.py
+++ b/skyportal/handlers/api/sources.py
@@ -505,6 +505,7 @@ async def get_sources(
     include_labellers=False,
     include_hosts=False,
     exclude_forced_photometry=False,
+    require_detections=False,
     is_token_request=False,
     include_requested=False,
     requested_only=False,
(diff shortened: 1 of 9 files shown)


static/js/API.js
```diff
@@ -77,12 +77,11 @@ export const filterOutEmptyValues = (params, removeEmptyArrays = true) => {
     // Also, the number 0 may be a valid input but evaluate to false,
     // so just let numbers through
     if (
-      (!(
+      !(
         Array.isArray(params[key]) &&
         params[key].length === 0 &&
         removeEmptyArrays
-      ) &&
-        params[key]) ||
```
mcoughlin, line 85: @Theodlz I was super confused, but this line was preventing passing false to a handler?!!?

mcoughlin:
@Theodlz Just a first pass to turning off the photstats stuff. I am sure you have better ideas for how to do this safely.

static/js/API.js
Theodlz: hmmmm that would be super weird! Is that something I added??? There must be a reason why that's here

skyportal/handlers/api/sources.py
```diff
@@ -789,136 +790,141 @@ async def get_sources(
...
-                    'first_detected_mjd'
-                    if not exclude_forced_photometry
-                    else 'first_detected_no_forced_phot_mjd'
-                )
-                query_params.append(
-                    bindparam(
-                        'first_detected_date',
-                        value=Time(arrow.get(first_detected_date).datetime).mjd,
-                        type_=sa.Float,
+        if require_detections:
```
Theodlz, line 793: do we really need an extra boolean? Can't we just not make it mandatory to specify first and last detection dates and pop it out of the form in the frontend? Seems like a very confusing way to handler parameters when we have one parameter that prevents other parameters from being used. Especially confusing for someone using the API. I would just not make the arguments mandatory when querying a localization or smth.

Theodlz (commented):
Just some questions

skyportal/handlers/api/sources.py
mcoughlin: I am ok with that as well. I do think the nominal behavior should be to encourage checking photometry and this does do that. But maybe a better frontend interface might suffice

Theodlz: Let me make some changes to that PR through this week, and I'll ask for a review!

But thanks a lot for all your help with the different requests from Mansi's group

Theodlz: Ah and also the changes need to be made to the `get_sources` method from the `sources.py` file, not this one which is being deprecated (and not the one used by the frontend anymore). I'm planning on deleting the old one soon.

Theodlz: You do make a point about encouraging folks to include first and last detection, unless they pass an argument to explicitly not constrain on photometry. I'll keep that behavior and just mirror it in the new `get_sources` method too.

Theodlz: Ah but you did that too!!! One week off and my brain cells are dead. Then I'm happy with this, just pushing a minor frontend change and we can merge

Theodlz:
@mcoughlin added the same changes to the candidates endpoint (though it's less important there, except maybe to not reject these AGN candidates or those with sparse (and maybe bogus) older detections. Relaxed the max time window constraint on the first and last detections, and modified the frontend a little bit.

static/js/API.js
Theodlz: Though it might not be done properly, I think this is important. Because when deleting it you make it possible to send values like null, undefined, or empty strings that might break the API calls. Might want to look for anything BUT false.

mcoughlin: Ok that would make sense. Right now you literally can't send false, so some exception needs to be made!

Theodlz: Ok should work now, I'll let you take a look?

mcoughlin:
@Theodlz It works!

Theodlz:
@mcoughlin that sending false or not stuff is creating issues. Looks like a bunch of frontend stuff then sends false when we used to just not send anything, and for some of the stuff that's usually either True or not sent, that breaks some things. 

I'll try to see if changing some of the frontend can fix that.

PR #4890, opened 2024-04-23
SourcesConfirmedInGCN: don't validate sources spatially

Theodlz:
The current handler verifies if a source being confirmed as inside of a GCN handler is spatially coincident with the localization. Aside from being unnecessary for all use cases we had so far, this extra validation step comes with some assumptions as to what the first and last detection to use (and min nb detections) when querying the sources in the localization are, which can come with some issues.

In this PR we simply remove this extra check.


mcoughlin:
LGTM

PR #4893, opened 2024-04-24
priority threshold

mcoughlin:
This PR adds a priority threshold option to followup request queries.


Theodlz (approved):
Looks great. Does the time resolution basically mean that it gives you the best solution found before that timeout is reached? LGTM

mcoughlin:
@Theodlz It's the duration of time they break the night up into to solve, i.e. what resolution can they break individual events into.

PR #4894, opened 2024-04-24
TNS submission - 504 (gateway timeout)

Theodlz:
Looks like every now and then TNS gives us a 504 error, where their server is inaccessible. Weirdly enough that seems to happen even when a report gets received by their system.

Here we re-use the "report confirmation" queue to look for reports with a gateway timeout error and figure out if they were submitted or need to be reprocessed again.


Theodlz:
@nabeelre @mcoughlin this is a WIP, I'll try to simulate all different cases locally before merging anything

nabeelre:
504 no more 🙌🏽

PR #4897, opened 2024-04-25
Allocation observation plan table

mcoughlin:
This PR adds a table for plan status for individual allocations.

### DIFF
--- a/skyportal/app_server.py
+++ b/skyportal/app_server.py
@@ -14,6 +14,7 @@ from skyportal.handlers.api import (
     UserACLHandler,
     AllocationHandler,
     AllocationReportHandler,
+    AllocationObservationPlanHandler,
     AssignmentHandler,
     BulkTNSHandler,
     CandidateHandler,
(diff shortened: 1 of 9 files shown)


Theodlz:
Feels a bit weird to have a different displays for the payload and the summary statistics:
image

That aside, can you summarize super quick what is the goal of the PR?

mcoughlin:
@Theodlz The JSONTree version is nice because it enables generic objects, which look differently for the different plans. But the main goal is to have a webpage to show for each allocation how the observation plan generation is going. Right now, it's hard to figure out systematic failures (or successes) because we only see them at an event level.

Theodlz (approved):
LGTM this is great, really helping making the list on the GCN page more compact too. I have a bunch of minor improvements (which all revolve in grouping the different buttons in a better way to save width and make the options clearer), but that can wait for another PR. I just pushed few minor changes to the "send to queue" button(s) to only show the buttons when required.

Feel free to merge once green!

PR #4898, opened 2024-04-26
Facility requests: various bugfixes

Theodlz:
* ZTF: for FP lightcurves with an overall non-zero procstatus, make sure to filter out individual datapoints that don't have a procstatus=0, to avoid bad substractions and things of that nature.
* WNTR: fix the class method call to retrieve the schedule_name from the response from the POST method that runs when sending a target to WNTR.


mcoughlin (approved):
LGTM

PR #4899, opened 2024-04-26
Source name validation: accept periods

Theodlz:
PERIOD


mcoughlin (approved):
LGTM. Is there something similar in the get_sources function call too? Just thinking of the navbar on the front-end.

Theodlz:
I answered on the issue. The problem there is completely different. I'll work on that tomorrow.

PR #4901, opened 2024-04-27
Observing run API GET: dtype bugfix

Theodlz:
Earlier this week we had issues where runs where a target would never rise/set would not return rise/set for any of the targets, because all of the individual rise/set would be masked arrays. When fixing it, it ended up causing troubles for runs where the targets are ALL visible, and none of the items end up having masked arrays to look at.

This PR fixes that


mcoughlin (approved):
LGTM

PR #4903, opened 2024-04-28
TNS submission job validation: remove continue statement

Theodlz:
Removes unnecessary continue statement that would prevent a following block of code from running, and only commit the session if we had any changes to make.


mcoughlin (approved):
LGTM

PR #4924, opened 2024-05-01
Observation plans status=complete frontend refresh

Theodlz:
When marking an observation plan as complete, make sure to refresh the list of plans in the frontend for the users that are looking at that GCN event's page.


mcoughlin (approved):
LGTM

Theodlz:
Tried locally and it worked. Will merge once green-ish.

PR #4925, opened 2024-05-01
Observation plan request send/delete: various bugfixes

Theodlz:
- When sending a plan to the scheduler, don't show an error due to a unique index on the GCNTrigger table
- When deleting queued observations from ZTF, based on the queue name, mark any associated observation plans as "deleted from telescope queue" to avoid having plans on a GCN page look like they are still in the queue when they aren't anymore.


mcoughlin (approved):
Thanks @Theodlz !

PR #4931, opened 2024-05-06
Period plot: units

Theodlz:
Pick units from: minutes, hours, days when plotting a period-folded photometry plot. Defaults to the current option with it per day.

image


mcoughlin (approved):
LGTM

PR #4934, opened 2024-05-14
DNM: doParallel in gwemopt

mcoughlin:
This PR turns on parallelization in gwemopt.


Theodlz:
@mcoughlin did you have a chance to share the exact error msg this gives you? Maybe you already did, just didn't remember where I could find it.

PR #4935, opened 2024-05-15
Galaxy Queries: various improvements

Theodlz:
* sort by and sort_order added to the `get_galaxies()` method as well as the API handler
* added option cut on mass to the `get_galaxies()` method as well as the API handler
* when querying galaxies in skymaps, default sorts by mass * probdensity (descending)
* querying galaxies in skymaps from the frontend now requires galaxy catalog section
* galaxy targeting with gwemopt uses the `mstar_prob_weighted` (mass * probdensity mentioned earlier) ordering by default. Also uses it with `equal` ordering to keep the results consistent and not random (which happens with no sorting), though the galaxies are not weighted by mass when sent to gwemopt in that case, only prob.

TODO: 
- [x] resolve issues where plans that fail (because there are no galaxies or smth like that) are still marked as complete. For example this can happen if we pick `sfr_fuv` as the ordering strategy, but the catalog does not have this metric.


skyportal/utils/observation_plan.py
```diff
@@ -0,0 +1,892 @@
...
+import sqlalchemy as sa
+from astropy.time import Time
+from regions import Regions
+from sqlalchemy.orm import sessionmaker, scoped_session
+
+from baselayer.log import make_log
+from baselayer.app.flow import Flow
+from baselayer.app.env import load_env
+
+from ..utils.cache import Cache, array_to_bytes
```
mcoughlin, line 17: from .cache

mcoughlin (approved):
It looks great. Can also run it once I can get the galaxy file.

Theodlz:
@mcoughlin thanks for the review! I still want to play around with it and maybe try to figure out why gwemopt is consistently giving 0 observations here

Theodlz:
also migrations are needed but well that's the usual 😉

Theodlz:
@mcoughlin I added some frontend and fixed some backend, feel free to rereview if you'd like but these are minor changes so I can merge anyway I think.

Theodlz:
(after I've fixed some of the tests)

mcoughlin:
@Theodlz good to merge when ready.

PR #4939, opened 2024-05-20
WINTERAPI: dither_distance default 30 -> 90

Theodlz:
Update the WINTERAPI dither_distance default from 30 to 90, as suggested by @robertdstein.

Also update the end_date to be +7 days from the start_date, not just +1


mcoughlin (approved):
LGTM

PR #4940, opened 2024-05-21
TNS reporting: bot user logic

Theodlz:
* allow or don't allow bot users to trigger auto reports (don't allow by default). This adds a column in the model of the TNSRobotGroup
* bot users cannot be added as permanent co-author.
* bot users are ignored as the first saver if they do not have a bio (that we'll use as the TNS report remark).
* report fails if one or more co-authors that are also bot users with a missing bio.
* concatenate the bios from all bot reporters as the remark for the TNS report.
* add a custom_remarks_string column in the TNSRobotSubmission model, similar to the custom_reporters_string. Used for manual reporting and to skip re-computing the remark if re-processing a submission request. Testing in an API test.
* allow users to add a remark to a TNS report from the frontend.

TODO: test the code locally, along with a testing instance of Kowalski (and the Fritz code that allows SP to specify the user account to use as the saver for a given filter, which is what is missing right now for me to test the full workflow).


Theodlz:
@mcoughlin migrations are missing I know :)

services/tns_submission_queue/tns_submission_queue.py
```diff
@@ -393,15 +427,30 @@ def build_reporters_string(submission_request, source, tnsrobot, session):
             affiliations = ", ".join(affiliations)
             reporters.append(f'{author.first_name} {author.last_name} ({affiliations})')
 
+            if author.is_bot:
+                # if the author is a bot, we use the bio as the remark
+                bio = str(author.bio).strip()
+                # if it doesn't end in a period, add one
+                if bio[-1] != ".":
```
mcoughlin, line 434: Would maybe also be good to enforce this when folks write and submit the description too.

mcoughlin (approved):
LGTM

services/tns_submission_queue/tns_submission_queue.py
Theodlz: Will do!

PR #4942, opened 2024-05-21
Followup APIs: priorityOrder

Theodlz:
add a priorityOrder field to all followup requests API, to (internall…y) keep track of the priority/urgency order. Used by Fritz to tell Kowalski what priority order to use when updating automated triggers


mcoughlin (approved):
LGTM!

PR #4943, opened 2024-05-22
Obsplan list: display fix

Theodlz:
Show the localization plot and statistics which are not visible right now once sent to queue


Theodlz:
pretty important for ToOs, self merging

PR #4944, opened 2024-05-29
TNSRobot bot user reporting: migration fix

Theodlz:
add missing server default


Theodlz:
self merging to fix prod

PR #4945, opened 2024-05-29
latest gwemopt

mcoughlin:
This PR updates code to use latest gwemopt.


alembic/versions/43be265c03c0_auto_report_allow_bots.py
```diff
@@ -0,0 +1,31 @@
...
+# revision identifiers, used by Alembic.
+revision = '43be265c03c0'
+down_revision = '94c30396a122'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.get_context().autocommit_block():
```
Theodlz, line 21: Why do we need to drop the default?

Theodlz (commented):
some comments

alembic/versions/43be265c03c0_auto_report_allow_bots.py
Theodlz: apologies, that was my mistake (had to merge quick to fix production last night)

mcoughlin: Yeah I was just doing my usual and copying the migra.sql in there. It seems to pass but feel free to do something different of you think it is better.

requirements.txt
```diff
@@ -63,7 +63,7 @@ geojson==3.1.0
 typing-extensions==4.8.0
 swifttools==3.0.21
 pip==23.3.1
-gwemopt==0.2.2
+git+https://github.com/skyportal/gwemopt.git@main#egg=gwemopt
```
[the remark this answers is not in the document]
mcoughlin: Yeah seems pretty quick!

alembic/versions/43be265c03c0_auto_report_allow_bots.py
Theodlz: trying to fix it now. Because that migration won't break our Fritz (that has already been migrated so the column is there) but could break something like Icare that hasn't been updated yet. Caus effectively thing goes back to the incorrect migration, which works if the table is empty but breaks the migration server otherwise

PR #4949, opened 2024-05-30
Add rust & cargo to Dockerfile for mocpy

Theodlz:
Add cargo and rust install to the Dockerfile, necessary for python dependencies that have a rust-based core


Theodlz:
merging to main so the docker container can be built

PR #4950, opened 2024-05-30
enable gwemopt parallel

mcoughlin:
This PR enables parallel gwemopt from config.


mcoughlin:
@Theodlz I modified the way I do parallelization in gwemopt and that seems to have fixed it. There is some issue with the MOC stuff and multiprocessing (https://github.com/cds-astro/mocpy/issues/132).

Theodlz:
> @Theodlz I modified the way I do parallelization in gwemopt and that seems to have fixed it. There is some issue with the MOC stuff and multiprocessing ([cds-astro/mocpy#132](https://github.com/cds-astro/mocpy/issues/132)).

Ok! Unfortunate, from my experience multiprocessing is really much better than threading for most use-cases but it's already much better to use threading than to not use anything at all!

I guess we are planning to turn this on in prod, right?

In which case maybe we should set it to True and maybe like 2 cores in the test.yaml config so that the parallelization is tested on git?

mcoughlin:
@Theodlz added!

Theodlz:
The test plans seem to be somehow even flakier with this option turned on. Just unlucky? @mcoughlin

mcoughlin:
@Theodlz Could be. Also presumably is tougher on the worker...

mcoughlin:
@Theodlz things seem to be passing, locally at least.

Theodlz:
@mcoughlin passing here too! Let me just see the changes out of curiosity and we can merge after that.

Theodlz (approved):
LGTM 🚀

PR #4951, opened 2024-05-31
Centroid Plot V2 (with plotly)

Theodlz:
Migrating the vega centroid plot to a much more customizable one with plotly, implementing some of the feed back from the October 2023 ZTF team meeting.

I still don't really know for sure:
- how I should compute the centroid: I implemented the hopefully smarter method we've been using the starlists, finding charts, ... in the frontend too which in theory is better than using a simple median/avg
- what the current red and blue circle represent. I think that the blue one is a 1-sigma circle (so basically a circle around the photometry. Only modification I made here is to only include the photometry that's not further than the 0.5" max offset (same as enforced while calculated the centroid), effectively applying a 0.5" upper bound on that circle. For the red circle, I honestly don't know yet, still need to figure that one out. The current code really lacks comments!
- maybe find a smarter way to have some kind of centroid plugins to add the archival stuff in Fritz without having to overwrite the whole component. Probably possible, but rules out useEffects given how this is written I think, but that might be fine.

**TODO**: a good clean-up/re-write to simplify and duplicate some of the logic to speed it up 🚀

### DIFF
--- a/static/js/components/CentroidPlot.jsx
+++ b/static/js/components/CentroidPlot.jsx
@@ -1,18 +1,41 @@
-import React from "react";
-import { useSelector } from "react-redux";
+import React, { useState, useEffect } from "react";
+import { useSelector, useDispatch } from "react-redux";
 import PropTypes from "prop-types";
-import { useTheme } from "@mui/material/styles";
 import makeStyles from "@mui/styles/makeStyles";
-import embed from "vega-embed";
 import * as d3 from "d3";
-import convertLength from "convert-css-length";
+import Typography from "@mui/material/Typography";
+
+import Plotly from "plotly.js-basic-dist";
+import createPlotlyComponent from "react-plotly.js/factory";
+
+import { PHOT_ZP } from "../utils";
+
+import * as archiveActions from "../ducks/archive";
+
+const Plot = createPlotlyComponent(Plotly);
 
 const useStyles = makeStyles(() => ({
-  centroidPlotDiv: (props) => ({
-    width: props.width,
-    height: props.height,
-  }),
+  plotContainer: {
+    width: "100%",
+    height: "100%",
+  },
 }));
+
+const catalogColors = {
+  AllWISE: "#2f5492",
+  Gaia_EDR3: "#FF00FF",
+  PS1_DR1: "#3bbed5",
+  PS1_PSC: "#d62728",
+  GALEX: "#6607c2",
+  TNS: "#ed6cf6",
+};
+
+function groupBy(arr, key) {
+  return arr.reduce((acc, x) => {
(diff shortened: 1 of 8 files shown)


static/js/components/CentroidPlot.jsx
```diff
@@ -38,351 +55,480 @@ const gcirc = (ra1, dec1, ra2, dec2) => {
...
+  points = points.map((p) => {
+    const newPoint = { ...p };
+    newPoint.flux = 10 ** (-0.4 * (p.mag - PHOT_ZP));
+    newPoint.fluxerr = (newPoint.magerr / (2.5 / Math.log(10))) * newPoint.flux;
+    newPoint.snr = newPoint.flux / newPoint.fluxerr;
+    return newPoint;
+  });
+  // remove those with a snr < 3.0 and valid flux and fluxerr
+  points = points.filter(
+    (p) => p.snr >= 3.0 && p.flux !== null && p.fluxerr !== null,
```
mcoughlin, line 191: Maybe make 3.0 a variable?

static/js/components/CentroidPlotPlugins.jsx
```diff
@@ -0,0 +1,71 @@
+import React from "react"; // eslint-disable-line no-unused-vars
+import PropTypes from "prop-types";
+import makeStyles from "@mui/styles/makeStyles";
+import Typography from "@mui/material/Typography"; // eslint-disable-line no-unused-vars
+
+// import * as archiveActions from "../ducks/archive";
+// the import above is needs to be implemented in the codebase where this plugin is used
```
mcoughlin, line 7: I don't understand this comment.

mcoughlin (approved):
Huge step forward.

static/js/components/CentroidPlot.jsx
Theodlz: smort. Let me do that!

static/js/components/CentroidPlotPlugins.jsx
Theodlz: argh sorry I typed that like a madlad

static/js/components/CentroidPlot.jsx
Theodlz: It is fine if I make it a constant at the top of the file for now, and add a slider/number input in a future PR? I would love to get this one in today.

mcoughlin: Of
Course

static/js/components/CentroidPlotPlugins.jsx
Theodlz: Done. Do you think it's clear enough now?

mcoughlin: Can remove is. Otherwise good

PR #4980, opened 2024-06-01
obspy 1.4.0

mcoughlin:
1.4.1 pins an older sqlalchemy.


mcoughlin:
This PR pins obspy 1.4.0.

PR #4993, opened 2024-06-02
classification test

mcoughlin:
This PR fixes the classification test.

### DIFF
--- a/skyportal/handlers/api/classification.py
+++ b/skyportal/handlers/api/classification.py
@@ -37,7 +37,9 @@ def post_classification(data, user_id, session):
     sources = session.scalars(Source.select(user).where(Source.obj_id == obj_id)).all()
     source_group_ids = [source.group_id for source in sources]
     user_group_ids = [g.id for g in user.accessible_groups]
-    group_ids = data.pop("group_ids", list(set(source_group_ids) & set(user_group_ids)))
+    group_ids = data.pop("group_ids", [])
+    if len(group_ids) == 0:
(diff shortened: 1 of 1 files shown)


mcoughlin:
@Theodlz thats fine, it was already there. I was just catching the test sending an empty list and handling it. We could have it error out instead, but it would be a change of behavior. On the plus side, it is only posted to the groups the source is in.

Theodlz:
> @Theodlz thats fine, it was already there. I was just catching the test sending an empty list and handling it. We could have it error out instead, but it would be a change of behavior. On the plus side, it is only posted to the groups the source is in.

Actually I am brining it up because the backend wasn't doing what the backend said before. Couple of people complained that it was posted to the groups of the source and not the public group as advertised, which caused confusion when a classification was saved, then the source saved to more groups and the folks from the new groups would not see them.

I will open another PR before deploy to only post to the public group when not specified (as per the frontend)

mcoughlin:
@Theodlz ok sounds good. Will you add the source to the site-wide group then?

Theodlz:
No! We don't need to add the source to a group to post a classification on it. Classification can just be made public and when accessing a source page (just like you would while scanning) you see it already before saving through source. Also that avoids folks from sending a request to say SEDM when it was already classified  by someone but not share to the public group 3(I saw that happen one or twice)

mcoughlin:
@Theodlz ok sounds good to me!

mcoughlin:
@Theodlz Did you push this change to the classification behavior?

PR #4979, opened 2024-06-01
auto send observation plan

mcoughlin:
This PR enables option for automatically sending observation plans.

PR #4994, opened 2024-06-03
moving objects

mcoughlin:
This PR adds the concept of moving objects.


Theodlz:
Can we discuss this at some point today or this week? I'm wondering if we could add some cool features to that thing to couple it with a MPC microservice or smth fancy like that (though probably in a following PR).

Theodlz:
Also wondering if we want to store mjd,ra,dec for these objects or some kind of orbital parameters when we have those (could be both)

mcoughlin:
@Theodlz Orbital parameters makes sense, although probably only if we think we will save them and then propagate the orbit forward ourselves.

Theodlz:
Definitely. Would it be fine to wait until tomorrow's Fritz meeting to discuss the objectives of that feature?

mcoughlin:
@Theodlz Of course. The objective I want to build to is being able to trigger ZTF on one of the localization regions. So we will need to add the ability to translate the RA+error, DEC+error quantities into a skymap (this should be easy enough), trigger a gwemopt schedule on this map (easy in principle but the plumbing will take a bit of work), and then trigger ZTF on that schedule.

Theodlz:
> @Theodlz Of course. The objective I want to build to is being able to trigger ZTF on one of the localization regions. So we will need to add the ability to translate the RA+error, DEC+error quantities into a skymap (this should be easy enough), trigger a gwemopt schedule on this map (easy in principle but the plumbing will take a bit of work), and then trigger ZTF on that schedule.

Are you talking about https://github.com/skyportal/skyportal/pull/4979 here, I was talking about this moving objects thing. But maybe I'm just missing smth

mcoughlin:
@Theodlz no I am just saying what we will build towards with this one. The long term vision has overlap with that PR.

Theodlz:
I know I've got some migrations to fix :) Will take care of it later today

skyportal/facility_apis/observation_plan.py
```diff
@@ -357,17 +371,18 @@ def submit(request_id, asynchronous=True):
...
+                        '*',
+                        "skyportal/REFRESH_GCNEVENT_OBSERVATION_PLAN_REQUESTS",
+                        payload={"gcnEvent_dateobs": request.gcnevent.dateobs},
+                    )
 
-                flow.push(
-                    '*',
-                    "skyportal/REFRESH_OBSERVATION_PLAN_NAMES",
-                )
+                    flow.push(
```
Theodlz, line 382: We don't use that anymore I'm pretty sure. Right before submitting I think that we check if the queue name used is valid. But fetching the names of all existing plans was really not a good idea. Aside from being slow and mostly useless, it's a lot of data sent for mostly no good reason to everyone using the app

Theodlz: You should be able to remove that

Theodlz: but worth double checking the frontend component on submit method (for gcn obs plans)

Theodlz (commented):
We already have clear performance challenges with the localizations. Accounting for that and just for the sake of clarity, shouldn't we have a dedicated table for moving object localizations or smth?

mcoughlin:
@Theodlz we should look at this again now that the automated triggering PR is in.

mcoughlin:
@Theodlz some suggestion from knowledgeable folks to use OpenOrb https://github.com/oorb/oorb/blob/master/classes/Orbit_class.f90 to track the orbits.

Theodlz:
@jinglin-l any updates?

Theodlz:
We have a number of others PR for that, so closing this one. We can always go back to this closed PR if needed anyways

PR #5017, opened 2024-06-04
einstein probe

mcoughlin:
This PR adds support for JSON GCN notices, including Einstein Probe and GUANO alerts.

PS: Due to a limitation of pygcn we are not saving the notices yet, but this is enough to create event pages with localizations to trigger observations of.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -701,6 +701,9 @@ gcn:
     - ICECUBE_ASTROTRACK_GOLD
     - ICECUBE_ASTROTRACK_BRONZE
     - MAXI_UNKNOWN
+  json_notice_types:
+    - einstein_probe.wxt.alert
+    - swift.bat.guano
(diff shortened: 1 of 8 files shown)


Theodlz:
Looks like we aren't storing notices for now. It's fine (minus notifications that won't work) for now, but we definitely want to be able to store those. I believe we need toupdate the NoticeTypes int-> str mapper, add a few extra optional columns that are JSON format specific, make the XML specific ones optional, add a format column to track which are XML and which are JSON, and handle all that in the backend and fr

Theodlz:
BUT, this PR is self-contained and we can start create events + localizations + tags even without the notices.

Theodlz:
@mcoughlin I fixed some issues in the GCN service and handlers/gcn.py + added support for posting those JSON notices from the API + a unit test using a EP JSON notice

PR #5018, opened 2024-06-05
get_observations method: remove heavy joinedload statements

Theodlz:
I noticed in prod really heavy queries hammering the DB on a daily basis around the same time that we ingest observations. I thought that was the issue but looking at the exact query logged by the gcloud console it looks more like something is fetching observations with a lot of nested joinedloads which are unnecessary during the query itself most of the time and where the relationships could just be lazy loaded when accessed later on (which is the default behavior).

All the joinedloads on the ExecutedObservations or QueuedObservations from the `get_observations` method are removed, and now only the fields are accessed after the query. Adding the instruments and their telescope to every single observations seems super overkill, especially given how easy it is to fetch those separately and just match by instrument_id later on.

### DIFF
--- a/skyportal/handlers/api/observation.py
+++ b/skyportal/handlers/api/observation.py
@@ -369,16 +369,10 @@ def get_observations(
             )
         )
     else:
-        obs_query = (
-            Observation.select(
-                session.user_or_token,
-                mode="read",
-            )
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #5019, opened 2024-06-05
get_sources: deprecate old version + change case sensitivity and tns_name matching

Theodlz:
* deprecate the old get_sources() method in favor of the much faster one we have implemented (and in use already from the frontend).
* our handlers that attempted to match sources without being case sensitive using func.lower(Obj.id) or Obj.id ILIKE ended up very slow. This is because DB indexes are not usable with these operators. On top of that, folks generally know how sources are named including the case sensitivity (ZTF20abcdefg, PGIR20abcedfg, 2020abc, ...). So here we keep using LIKE to look for partial matches where we can, but we remove the case insensitivity in favor of somewhat significant performance gains and reduced DB load.
* similarly, the same case insensitive logic had been implemented for tns_name matching. BUT, on top of that, since folks might look for AT2020abc, AT2020abc (notice the missing space here), or simply 2020abc, matching an input with a space or no space to our tns_name column which does include the space has been somewhat annoying. So in the backend, we try to detect if the provided sourceID passed to the `get_sources()` looks like a TNS name with or without designation. That way if it does match a TNS name convention, we check if objects match by TNS name and otherwise do not check that (which means better performance when not searching by TNS name). We also remove the designation when searching by TNS name so that we do not care about the spacing anymore + can return the object even if the designation is incorrect. Particularly useful if looking with the AT name but the source has been "promoted" to SN already.
* Minor changes are made to the QuickSearch component on the frontend so that the MUI autocomplete can still match the results (that might have a different TNS name than the one used as input) of the query to the text box input.


Theodlz:
bye bye 1200 lines of code, that's the kind of PR I like

mcoughlin (approved):
LGTM

PR #5020, opened 2024-06-06
Correction public source page

antoine-le-calloch:
Correction of some comment about the new public source page functionality.


skyportal/models/classification.py
```diff
@@ -93,7 +93,7 @@ class Classification(Base):
     def to_dict_public(self):
         return {
             'classification': self.classification,
-            'author_name': self.author_name,
+            'author_name': self.author.first_name + ' ' + self.author.last_name,
```
mcoughlin, line 96: Should this fallback to author_name if first and last name aren't defined?

mcoughlin (approved):
@Theodlz will need to take a look, but this looks good to me

skyportal/models/classification.py
Theodlz: It should yes 👍. @antoine-le-calloch if you look at the TNS components of the frontend there is a nice method nice that tries to have firstname + last name (affiliations) and uses fallback when needed. You could do the same. Using the affiliations if available could be a big plus too.

skyportal/handlers/api/public_pages/public_source_page.py
```diff
@@ -21,6 +21,10 @@ def calculate_hash(data):
 
 
 def process_thumbnails(thumbnails, ra, dec):
+    thumbnails = sorted(
+        thumbnails,
+        key=lambda x: ["new", "ref", "sub", "sdss", "ls", "ps1"].index(x["type"]),
```
Theodlz, line 26: This should source the enum found here, not redefine one: https://github.com/skyportal/skyportal/blob/main/skyportal/enum_types.py#L52

static/js/utils/photometry.js
```diff
@@ -1,80 +1,203 @@
...
+  if (isMobile !== window.isMobile) {
+    window.isMobile = isMobile;
+    Plotly.relayout(
+      document.getElementsByClassName("plotly")[0].parentElement,
+      getLayout(),
+    );
+  }
+}
+
+window.addEventListener("resize", adjustLegend);
```
Theodlz, line 172: I don't think we want an event listener added to a file that is important by the private page too, this will hurt performance for no good reason there as we don't use it. I think you could just have that listener created in your public page and pass it the callback

Theodlz (commented):
Some comments

Theodlz:
@antoine-le-calloch are you ok with me merging this and you can simply keep pushing changes and open another PR later?

Theodlz:
Ah nvm I see that some things are not committed yet. Any chance for us to merge by tonight?

antoine-le-calloch:
> @antoine-le-calloch are you ok with me merging this and you can simply keep pushing changes and open another PR later?

Yeah sure you can merge it !

antoine-le-calloch:
> Ah nvm I see that some things are not committed yet. Any chance for us to merge by tonight?

Humm not sure to understand, what did you mean ?

Theodlz:
> > Ah nvm I see that some things are not committed yet. Any chance for us to merge by tonight?
> 
> Humm not sure to understand, what did you mean ?

I'm still seeing some of the window listeners in the js/utils

PR #5021, opened 2024-06-06
json alerts part 2

mcoughlin:
This PR continues to work towards json alerts.


skyportal/utils/gcn.py
```diff
@@ -266,46 +280,61 @@ def get_skymap_url(root, notice_type, timeout=10):
...
-        retraction = int(retraction.attrib['value'])
-        if retraction == 1:
-            return True
+    if isinstance(root, dict):
+        retraction = root.get('retraction')
+    else:
+        retraction = root.find("./What/Param[@name='Retraction']")
+        if retraction is not None:
+            retraction = int(retraction.attrib['value'])
+            if retraction == 1:
```
Theodlz, line 289: Idk if it's git but it looks like in the JSON case you don't return True based on anything (the if statement looking at the retraction value is in the else that's above). Is that on purpose?

skyportal/utils/gcn.py
```diff
@@ -266,46 +280,61 @@ def get_skymap_url(root, notice_type, timeout=10):
...
+        if loc is None:
+            return ra, dec, error
+
+        ra = loc.find('./AstroCoords/Position2D/Value2/C1')
+        dec = loc.find('./AstroCoords/Position2D/Value2/C2')
+        error = loc.find('./AstroCoords/Position2D/Error2Radius')
 
-    if None in (ra, dec, error):
-        return ra, dec, error
+        if None in (ra, dec, error):
```
Theodlz, line 313: same comment for this method

skyportal/utils/notifications.py
```diff
@@ -65,24 +69,27 @@ def gcn_notification_content(target, session):
...
+            ra = notice_content.get("ra")
+            dec = notice_content.get("dec")
+            error = notice_content.get("ra_dec_error")
+        else:
+            loc = notice_content.find('./WhereWhen/ObsDataLocation/ObservationLocation')
+            ra = loc.find('./AstroCoords/Position2D/Value2/C1')
+            dec = loc.find('./AstroCoords/Position2D/Value2/C2')
+            error = loc.find('./AstroCoords/Position2D/Error2Radius')
+
+            if ra is not None:
```
Theodlz, line 82: same comment here. Any reason not to apply this dtype conversion to both cases?

alembic/versions/b854bf5d22b3_json_alerts.py
```diff
@@ -0,0 +1,335 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('gcnevents', sa.Column('gcn_type', sa.String(), nullable=True))
+    op.alter_column(
+        'gcnnotices',
+        'notice_type',
+        existing_type=postgresql.ENUM(
```
Theodlz, line 26: not familiar with this `existing_type` arg here. What is this supposed to do?

skyportal/handlers/api/gcn.py
```diff
@@ -171,7 +172,11 @@ def post_gcnevent_from_xml(
 
     if event is None:
         event = GcnEvent(
-            dateobs=dateobs, sent_by_id=user_id, trigger_id=trigger_id, aliases=aliases
+            dateobs=dateobs,
+            sent_by_id=user_id,
+            trigger_id=trigger_id,
+            aliases=aliases,
+            gcn_type="voevent",
```
Theodlz, line 179: Quick question. Why do we need to track this on the event? I mean to me JSON/VOEVENT is purely for the notice and shouldn't matter for this event itself.

Theodlz: We could have both voevent notices and JSON notices for the same event

Theodlz (commented):
Just a few questions

services/gcn_service/gcn_service.py
```diff
@@ -171,9 +172,6 @@ def poll_events(*args, **kwargs):
                                 notify=False,
                             )
                         elif alert_type == "json":
-                            # save the payload to a file
-                            with open(f'/tmp/{str(uuid.uuid4())}.json', 'w') as f:
```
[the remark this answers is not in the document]
mcoughlin: No worries.

skyportal/utils/gcn.py
mcoughlin: Fixed, I think.

skyportal/utils/gcn.py
mcoughlin: The function returns ra, dec, error at the end.

skyportal/utils/notifications.py
mcoughlin: same deal here.

skyportal/handlers/api/gcn.py
mcoughlin: Ok I moved it over to the notice.

Theodlz:
Thanks for addressing the comments @mcoughlin. WIll give another review soon

PR #5022, opened 2024-06-07
Mui v5

### LINKED ISSUE #4982: mui v5
https://next.mui.com/material-ui/migration/migrating-from-jss/

We are accruing technical debt with our mui situation. We should consider migrating. I tried the first one and it ran on most of the files, but slightly weird and some errors.

dj-256:
This applies codemods on the whole code base for MUI 5 migration (see https://next.mui.com/material-ui/migration/migration-v4). Specifically, it migrates styles from jss to [Emotion](https://emotion.sh/docs/@emotion/css). Manual fixes were required as the jss to emotion codemod had broken a lot of files.

closes #4982


dj-256:
There is one test failing in /tests/api/test_reminders.py:
```python
assert data[reminder_index]['next_reminder'] == next_reminder.strftime(
        "%Y-%m-%dT%H:%M:%S"
    )
```
It shows an offset of an hour between the expected and actual value:
```
  - 2024-06-11T00:28:42
  ?          ^
  + 2024-06-12T00:28:42
  ?          ^
  ```
  This should not have been affected as it is an API test

mcoughlin:
@dj-256 do you know where the tests went?

One note, we like to pin packages in package.json, so please remove the ^ introduced there.

Theodlz:
as @mcoughlin said, @dj-256 none of the CI ran here for some reason. It needs to be fixed before this is reviewed

dj-256:
@mcoughlin @Theodlz I skipped the ci because I needed to push some changes as I am working on two different computers but I know the tests will fail as the issue is not resolved yet. You can skip ci by adding "[skip ci]" at the end of your commit message

dj-256:
@mcoughlin Ok for the package.json but why do you do that though?

Theodlz:
without that you have the risk of - not for all package though I agree - to break a feature after building the docker image for production (if not sometimes breaking the docker build itself directly) , even if your CI here passed with flying colors (because a package was updated since then).

Also not pinning specific version for linting packages is super annoying. It means you and I could technically have different versions running, and each of our PRs modifying the same files will have this ping pong match of reformatting stuff all the time.

So for stability and practicality's sake, we pin the version of as many packages as possible. We run dependabot monthly (or more often when we feel like it instead) which is just the best way to do this without spending our time to fix minor errors brought by unexpected updates.

dj-256:
@Theodlz  Alright I see but you usually commit the `package-lock.json` for that though. Pinning versions in `package.json` does not necessarily ensure your versions stay consistants between different builds and your local version, it is the job of `package-lock.json`.

Theodlz:
> @Theodlz Alright I see but you usually commit the `package-lock.json` for that though. Pinning versions in `package.json` does not necessarily ensure your versions stay consistent between different builds and your local version, it is the job of `package-lock.json`.

package.json with a properly pinned version of the packages virtually does the same thing as package-lock.json. It is more than sufficient and avoids redundancy. So might as well pin the packages properly and just have the package.json which includes the scripts too.

Sure, you still encounter the risk of having a dependency of a dependency that did not pin their dependencies properly too. But the problem with adding package-lock.json is that dependabot would probably completely lose it as it will be looking at many more dependencies. Unless there is a way to only make it look at what's in package.json but reflect the changes in the lock file too.

If you want to test that on your fork (so you can run dependabot manually on the branch), that would be interesting.

dj-256:
I am using dependabot on several repositories with package-lock.json. I have not encountered any issue, it just checks for the package.json file and updates both. Sure, I can do it on my fork to try.

Theodlz:
If you can make it work that's great. Out of curiosity, how does your dependabot handle all the dependencies in the lock file? Does it only care about the ones that are in the package.json?

Theodlz:
In any case let's do this in a separate PR, this already has 8000+ lines changed

Theodlz:
and please let the CI run next time. It's really ok. At least that way me and michael can have an idea of how close this is to completion (and what fails). It's just nice for bookeeping to see how the commits make the CI progress. Also it's just on push so unless you push after every single commit, it's not an issue

dj-256:
@Theodlz I dont think I have ever had an update from the package-lock.json from dependabot. I looked into it and it seems that dependabot does read package-lock.json but only for security alerts on known vulnerabilities.

Theodlz:
closing for now, too many conflicts

PR #5023, opened 2024-06-08
Public page: Photometry plot improvement

antoine-le-calloch:
Improve the actual Photometry plot use on the public pages.
- Increase the number of decimal display for the Y/X axies.
- Change the plotly import to import the same version as the one on the private page.
- Update the photometry.js to display minor ticks and to only recompute the graph, and not the legend position, when reset button is press.


Theodlz (approved):
LGTM!

PR #5026, opened 2024-06-10
HST/Chandra

mcoughlin:
This PR enables HST/chandra cutouts.


Theodlz (commented):
So I'm obviously happy with the backend code. However, the frontend with what is essentially a third row of cutouts makes the source page and scanning kinda broken again, with clearly sub optimal UI.

I think we should split that PR in 2. First let the backend start processing things, while we work on a solid frontend. I think what we need is to have the new/ref/diff together, always shown when available, and for the other survey/archival thumbnails have some slideshow with a maximum of 3 of them visible to keep it looking like it does right now.

Theodlz:
Also on the demo data, HST and Chandra cutouts are systematically missing. The only HST one I see looks like noise:
image

mcoughlin:
@Theodlz Thanks. I added Cygnus X-1 to the demo data and removed the jsx changes.

Theodlz:
@mcoughlin actually... I got something decent working in JSX for the source page. I will just fork this branch and keep working on it while we get the backend here.

PR #5031, opened 2024-06-11
Search by comment: Part 1 (Sources only)

Theodlz:
* fixes the sources query by comment, which broke when the `get_sources` method was re-implemented.
* When using the search bar on the `SourceTable.jsx` component, add a MUI Select to pick whether you search by ID/IAU name, or by comment. 

In the following PR we'll want to redesign some of the `QuickSearch` components to allow folks to do the same. But I think we need to discuss the UI and backend functionality at the next Fritz meeting. Do we want to say "search for sources -> by name OR comment with a MUI Select" or "search comments -> on sources OR gcn event OR ...". Basically, are we looking for the comments or the object that the comment is one. One could argue that the former makes more sense, but the advantage of querying the comments themselves is that we can show a render of what the comment that contains the text you are typing looks like with a link to the resource (source, gcn, ...). I like the latter because aside from being able to show the comments that match the query which makes it so much easier to figure out if you found the right thing, we can render the markdown comments properly, show the attachments, ...


mcoughlin (approved):
This seems fine to me at least.

PR #5032, opened 2024-06-12
Search by comment: Part 2 (Sources & GCN, via QuickSearchBar)

Theodlz:
Part 2 of #5031, allows to query comments without specifying a resourceID (we make it optional now) but by specifying a text. Both options can be combined, but one cannot leave both unspecified.

We then make it possible to search for said objects using the quick search bar.

### DIFF
--- a/skyportal/handlers/api/comment.py
+++ b/skyportal/handlers/api/comment.py
@@ -31,6 +31,8 @@ from ...models import (
 
 log = make_log('api/comment')
 
+MAX_COMMENTS_NO_RESOURCE_ID = 1000
+
 
 def users_mentioned(text, session):
     punctuation = string.punctuation.replace("-", "").replace("@", "")
@@ -89,7 +91,7 @@ def instruments_mentioned(text, session):
 
 class CommentHandler(BaseHandler):
     @auth_or_token
-    def get(self, associated_resource_type, resource_id, comment_id=None):
+    def get(self, associated_resource_type, resource_id=None, comment_id=None):
         """
         ---
         single:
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/comment.py
```diff
@@ -174,69 +182,80 @@ def get(self, associated_resource_type, resource_id, comment_id=None):
...
                 else:
                     return self.error(
                         f'Unsupported associated resource type "{associated_resource_type}".'
                     )
 
-                query_output = [
-                    {**c.to_dict(), 'resourceType': associated_resource_type.lower()}
-                    for c in comments
-                ]
+                stmt = table.select(self.current_user)
```
mcoughlin, line 213: This should be associated_user_or_id or whatever we use in case of tokens I think.

skyportal/handlers/api/comment.py
```diff
@@ -174,69 +182,80 @@ def get(self, associated_resource_type, resource_id, comment_id=None):
...
+                    if pageNumber < 1:
+                        return self.error("Page number must be greater than 0.")
+                    numPerPage = 25 if numPerPage is None else int(numPerPage)
+                    if numPerPage < 1:
+                        return self.error("Number per page must be greater than 0.")
+                    if numPerPage > MAX_COMMENTS_NO_RESOURCE_ID:
+                        return self.error(
+                            f"Number per page must be less than {MAX_COMMENTS_NO_RESOURCE_ID}."
+                        )
+                    stmt = stmt.where(
```
mcoughlin, line 227: neat.

mcoughlin (approved):
Minor comment.

skyportal/handlers/api/comment.py
Theodlz: good catch, new code inherited from incorrect old code 😉

Theodlz: done, also updated it everywhere else in that file

mcoughlin: @Theodlz great, eventually we will have squashed them ;)

PR #5033, opened 2024-06-12
GroupSources frontend: minor fix

Theodlz:
when querying for group sources (so with group_ids length = 1), return the group_id used in the query response to read it from the redux and not show another group's sources when opening the group source page


mcoughlin (approved):
LGTM

PR #5034, opened 2024-06-12
Instrument API classes: render_frontend_info kwargs

Theodlz:
adds kwargs to the render_frontend_info method of all instrument API classes. That way, reusable quantities (like unique galaxy catalog names that some instruments take as a parameter) can be computed once before we loop on each instrument and build the rjsf form for each.

### DIFF
--- a/skyportal/handlers/api/internal/robotic_instruments.py
+++ b/skyportal/handlers/api/internal/robotic_instruments.py
@@ -18,7 +18,7 @@ class RoboticInstrumentsHandler(BaseHandler):
                         )
                     ).all()
                     retval = {
-                        i.id: i.api_class.frontend_render_info(i, self.current_user)
+                        i.id: i.api_class.frontend_render_info(i, session.user_or_token)
                         for i in instruments
                     }
                 elif apitype == "api_classname_obsplan":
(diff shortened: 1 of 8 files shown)


skyportal/handlers/api/internal/robotic_instruments.py
```diff
@@ -25,9 +27,18 @@ def get(self):
...
+                    # and pass them to the frontend_render_info method
+                    # to avoid having to run that query for each instrument
+                    galaxy_catalog_names = session.scalars(
+                        sa.select(Galaxy.catalog_name).distinct()
+                    ).all()
                     retval = {
                         i.id: i.api_class_obsplan.frontend_render_info(
-                            i, self.current_user
+                            i,
+                            self.current_user,
```
mcoughlin, line 40: Any reason this should also be user_or_token?

mcoughlin (approved):
Looks neat, thanks @Theodlz

skyportal/handlers/api/internal/robotic_instruments.py
Theodlz: Ok it should you are right, I didn't change anything here. We're just lagging behind but nice catch. Let me fix it here and in a couple more places

Theodlz: Done @mcoughlin

PR #5035, opened 2024-06-13
Public page : Improve public page security and style

antoine-le-calloch:
- Improve and correct some behavior of time ago display on public sources page.
- Improve see all button style.
- Add html escaping on data display on the public source template to avoid XSS attack.


mcoughlin (approved):
LGTM

Theodlz (approved):
LGTM!

PR #5037, opened 2024-06-15
only query necessary columns for phot stats

mcoughlin:
This PR makes a minor optimization to only query necessary columns for phot stats.


Theodlz (approved):
LGTM

PR #5038, opened 2024-06-15
notice type filtering

mcoughlin:
This PR tries to better handle notice type filtering.


Theodlz (approved):
LGTM

skyportal/facility_apis/ztf.py
```diff
@@ -1191,20 +1191,21 @@ def fetch_depot_observations(instrument_id, session, depot_url, jd_start, jd_end
...
-    obstable.rename(
-        columns={
-            'obsjd': 'obstime',
-            'maglim': 'limmag',
-            'fwhm': 'seeing',  # equivalent as plate scale is 1"/pixel
-        },
-        inplace=True,
-    )
-    obstable['target_name'] = None
+    if len(dfs) > 0:
```
Theodlz, line 1194: Someone has been looking at Sentry

alembic/versions/0f1f62adc91c_notice_types.py
```diff
@@ -0,0 +1,337 @@
...
+            'HAWC_BURST_MONITOR',
+            'AMON_NU_EM_COINC',
+            'ICECUBE_ASTROTRACK_GOLD',
+            'ICECUBE_ASTROTRACK_BRONZE',
+            'SK_SN',
+            'ICECUBE_CASCADE',
+            name='noticetype',
+        ),
+        type_=sa.String(),
+        existing_nullable=True,
```
Theodlz, line 174: What is this migration doing? I get from the models that you don't use the ENUM from pygcn anymore, so why does it not just drop the enum here? Just curious

skyportal/models/gcn.py
```diff
@@ -12,7 +12,6 @@
 import json
 import random
 
-import gcn
```
Theodlz, line 15: GOOD RIDDANCE.

Theodlz (commented):
Just one question about the migration

alembic/versions/0f1f62adc91c_notice_types.py
mcoughlin: But there is still a column. So it's an alter I guess.

PR #5043, opened 2024-06-17
Public page : thumbnail improvement

antoine-le-calloch:
Improve the thumbnail process for the public page :
 - filter dr8


skyportal/handlers/api/public_pages/public_source_page.py
```diff
@@ -22,19 +22,20 @@ def calculate_hash(data):
 
 
 def process_thumbnails(thumbnails, ra, dec):
+    has_ls = any("ls" in thumbnail["type"] for thumbnail in thumbnails)
```
mcoughlin, line 25: Can you add some comments on what this does? I guess the "dr8" is specifically to avoid showing in favor of dr9?

mcoughlin (approved):
Would benefit from a few comments.

mcoughlin (approved):
LGTM

skyportal/handlers/api/public_pages/public_source_page.py
```diff
@@ -22,19 +22,21 @@ def calculate_hash(data):
 
 
 def process_thumbnails(thumbnails, ra, dec):
+    # Sort thumbnails by type, and remove 'DR8' thumbnail if 'LS' (that corresponds to DR9) thumbnail are present
+    has_ls = any("ls" in thumbnail["type"] for thumbnail in thumbnails)
     thumbnails = sorted(
-        thumbnails,
+        [thumb for thumb in thumbnails if not has_ls or "dr8" not in thumb["type"]],
```
Theodlz, line 28: though this is completely correct, would you mind changing it to: `if not (thumb["type"] == "dr8" and has_ls). Just much more clear and readable that way (and not at the cost of performance clearly).

Thanks!

Theodlz (approved):
LGTM! One super quick comment

skyportal/handlers/api/public_pages/public_source_page.py
Theodlz: @antoine-le-calloch in case you missed this

Theodlz (approved):
LGTM

PR #5044, opened 2024-06-18
Source candidates history + default comment and classification group access

Theodlz:
This PR addresses a couple of issues:
- from a source page, show a list of all the times the source's candidates passed filters that the user has access to. Greatly facilitates debugging filters and figuring out when a source first and last passed a filter.

- a somewhat recent PR (won't say from who 😉) created a pretty messy situation when posting comments on top of the already messy situation when posting classifications, which clearly has started hurting performance quite a lot and needs to be addressed ASAP. Comments have recently been posted to ALL user-accessible groups, which includes single user groups AND the public group, potentially including 400+ groups in production for a single classification... This means that when fetching the source from the frontend end and its comments, we started accumulating quite a lot of latency... So in this PR, instead of sharing to EVERYTHING including the public group (which already makes it useless to share to any other group!!!), and as per user requests we simply share new comments and classifications to the public group by default UNLESS specified otherwise explicitly, for both comments and classifications.

- to avoid showing folks a lot of comments and classifications (now public by default) on objects that might not be saved as sources to the groups they have access to, in the newsfeed handler we make sure to only return comments and classifications that are on sources that the current user has access to, except if the user is a sysadmin (in which case they can see everything).

- we attempt to access the `comment.groups` when looping over a source's comments instead of including it in a joinedload, simply because the one-to-many relationship creates duplicated entry when joining (naturally) which forces the query to then have an awkward nested query thing. This might help to load the sources whose comments have been posted to a lot of groups in the past couple of weeks/months. If performance gets worse overall though, we will revert that change in a following PR after doing a thorough clean-up of the database instead.

TODO: likely fix the tests that will freakout.

Screenshots of the candidates history button (on the right of the source name):
image
and the dialog once open:
image


mcoughlin (approved):
Oops

PR #5048, opened 2024-06-18
GCN Tach: prepare support for EP events

Theodlz:
- let the TACH gcn features look for `EP` prefixes, though anyway TACH doesn't seem to have EP events (only their circulars) anyway.
- moves the NewGcnEvent.jsx component to the gcn directory


mcoughlin (approved):
LGTM

PR #5049, opened 2024-06-19
Add a GalaxyCatalog table + ObjPhotometryHandler GET speed up

Theodlz:
* add a GalaxyCatalog table with name, description, and URL. That way the Galaxy table references a catalog instead of containing the catalog_name itself. This is useful to avoid having to query the entire Galaxy table to find the list of catalogs we have in the DB, which is adding latency to different endpoints.
* use the `joinedload(<relationship>).load_only(<columns>)` and not just `joinedload(<relationship>)` in the photometry query from the ObjPhotometryHandler GET method. That way we have control over what columns from the relationship we are loading, so we can remove a large number of useless information and even personal information for the photometry owners. This speeds up the query significantly.
* Make it optional to load photometry annotations, a feature we are not using in production yet (the table is pretty much empty) meaning that we don't have a real reason to force joinedload on that table which slows the query a little bit.

With these changes, we reduce the time needed for the server to respond while hydrating a few things on the web app, but we also make the photometry load faster. From my own tests ran locally on `ZTFe028h94k` (variable source from the testing data that has a lot of data points), I load the photometry in ~300 ms instead of ~900 ms, meaning that it's around 3 times faster. Of course that will scale differently depending on the source and the data access enforced on the photometry (nb of groups, owner, and streams), but this is very encouraging.

In a future PR, we should consider adding similar-ish improvements to a few other handlers. We obviously send a lot of data in joinedloads that we could ignore, which should really be done everywhere we query for multiple rows of the DB. This will speed things up, especially on large tables and/or nested relationships.

The migration for the galaxies was a little tricky but it's doable (alembic is not smart enough to make it work by itself sadly). This migration script should definitely be used as a template for this kind of DB schema changes that we might want to do in the future.

EDIT: this PR also attempts to duplicate some of the DB queries that run when validating group_ids and stream_ids when posting photometry. This will likely speed up posting photometry a little bit.


mcoughlin (approved):
Wow.

PR #5057, opened 2024-06-20
public page : CSV table download improvement

antoine-le-calloch:
- Improve csv download for photometry on public page to display only 2 digits after decimal for all header except mjd.
- Move CSV and Photometry js file to the public_page folder.
- Add notes column on public page photometry csv download
- Increase photometry plot size
- Add style on public page button and correct a photometry data duplication behavior.


mcoughlin (approved):
LGTM

skyportal/handlers/api/public_pages/public_source_page.py
```diff
@@ -140,7 +140,8 @@ async def post(self, source_id):
                         )
                     )
                 data_to_publish["photometry"] = [
-                    photo.to_dict_public() for photo in session.scalars(query).all()
+                    photo.to_dict_public()
+                    for photo in session.scalars(query.distinct()).all()
```
Theodlz, line 144: very minor comment. But in most places what you call the `query` here is called `stmt` (short for "statement"). That is because `.query()` is also an actual method which makes things a little confusing and prone to errors. It's ok as is, but something to keep in mind next time you open a PR.

skyportal/models/photometry.py
```diff
@@ -449,18 +449,27 @@ def snr(self):
...
+            "mjd": serialize_photometry["mjd"],
+            "mag": serialize_photometry["mag"],
+            "magerr": serialize_photometry["magerr"],
+            "filter": serialize_photometry["filter"],
+            "limiting_mag": serialize_photometry["limiting_mag"],
+            "instrument_id": serialize_photometry["instrument_id"],
+            "instrument_name": serialize_photometry["instrument_name"],
+            "origin": serialize_photometry["origin"],
+            "notes": serialize_photometry["validations"][0].get("notes", "")
+            if serialize_photometry["validations"]
```
Theodlz, line 471: quick tip here to avoid errors later on. Instead of doing if `serialize_photometry["validations"]`. I suggest doing `if instance(serialize_photometry.get("validations"), <type_you_expect>)`. That's a little more robust.

static/public_pages/sources/source/source_template.css
```diff
@@ -114,6 +114,7 @@
       }
     }
     .photometryPlot {
+      height: 75vh;
```
Theodlz, line 117: thanks!

static/public_pages/style.css
```diff
@@ -120,16 +120,15 @@ button {
...
   border-radius: 4px;
   color: rgba(0, 0, 0, 0.87);
   background-color: #e0e0e0;
   font-family: "Roboto", "Helvetica", "Arial", sans-serif;
-  transition:
-    background-color 250ms cubic-bezier(0.4, 0, 0.2, 1) 0ms,
-    box-shadow 250ms cubic-bezier(0.4, 0, 0.2, 1) 0ms,
-    border-color 250ms cubic-bezier(0.4, 0, 0.2, 1) 0ms,
-    color 250ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
+  transition: background-color 0.3s;
```
Theodlz, line 128: someone gave up on animations hahaha

Theodlz (approved):
LGTM 🚀 . I left a few minor comments that are mostly coding-style suggestions for future PRs

static/public_pages/style.css
antoine-le-calloch: Ah ah yes didn't see the difference between the two.

skyportal/models/photometry.py
antoine-le-calloch: Yeah but it's because validations always in the serialize_photometry, just sometimes the array inside is empty. And the "if serialize_photometry["validations"]" test if the array is empty or not. @Theodlz

Theodlz (approved):
LGTM 🚀

PR #5058, opened 2024-06-22
Do not require specific permission to make an observation plan

mcoughlin:
This PR ensures that we do not require specific permission to make an observation plan.


Theodlz (approved):
LGTM

PR #5059, opened 2024-06-22
Comment Attachments: Preview Improvements

Theodlz:
* drop the react-file-preview dependency in favor of using the browser's native support of PDFs (like arxiv for example)
* support more files for preview (audio, video, image, text, ...). This makes the webpack a little lighter too.
* render skymap fits files too
* nicer frontend to render text, video, and images

Note: I noticed (though that was already an issue) that there are quite a bit of rerenders of the comments. We might want to open a dedicated PR to help prevent those with react.memo or something similar.

I suggest to the reviewer to just have fun uploading files - at least one of each time - to see how things look. Some video and audio formats are not supported (like .mpeg), but that's simply because the HTML audio and video tags (meaning the browser in general) don't support them anyway!


skyportal/handlers/api/comment.py
```diff
@@ -1235,139 +1280,117 @@ def get(self, associated_resource_type, resource_id, comment_id):
...
-                    CommentOnShift.select(session.user_or_token).where(
-                        CommentOnShift.id == comment_id
-                    )
-                ).first()
-                if comment is None:
-                    return self.error(
-                        'Could not find any accessible comments.', status=403
-                    )
-                comment_resource_id_str = str(comment.shift_id)
+        table, resource_id_col = None, None
```
mcoughlin, line 1288: Much neater.

mcoughlin (approved):
Thanks for fixing this.

Theodlz:
How did a broken mui icon import work for the past 4 years?????????? I am so confused

PR #5060, opened 2024-06-22
DNM: get_sources(): implement caching again

Theodlz:
When I rewrote the `get_sources()` method, given how fast it was I did not implement caching again.

This PR implements it here. I am not sure whether or not it will be of any use to be honest (given how fast the queries are right now). On queries with very little to no filters it might take longer to read from file than to run a query, but we'll see! Might as well have it and we can test it in production with lots of data.

⚠️ DO NOT MERGE: I see some changes that I did not make pop on git. Maybe git is just not matching the lines right but I need to look into it.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -515,6 +515,7 @@ services:
 misc:
   days_to_keep_unsaved_candidates: 7
   minutes_to_keep_candidate_query_cache: 60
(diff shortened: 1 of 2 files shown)


Theodlz:
How did a broken mui icon import work for the past 4 years??????????

PR #5061, opened 2024-06-23
Allocations which implement relevant APIs

mcoughlin:
This PR only return allocations which implement relevant APIs for the observations page.

### DIFF
--- a/skyportal/handlers/api/allocation.py
+++ b/skyportal/handlers/api/allocation.py
@@ -420,6 +420,25 @@ class AllocationHandler(BaseHandler):
             # order by allocation.instrument.telescope.name, then instrument.name, then pi
 
             apiImplements = self.get_query_argument('apiImplements', None)
+            if apiImplements not in [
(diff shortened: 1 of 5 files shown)


skyportal/handlers/api/allocation.py
```diff
@@ -418,6 +418,26 @@ def get(self, allocation_id=None):
 
             allocations = session.scalars(allocations).unique().all()
             # order by allocation.instrument.telescope.name, then instrument.name, then pi
+
+            apiImplements = self.get_query_argument('apiImplements', None)
```
Theodlz, line 422: Could we just enforce a set of valid values here?

Theodlz (approved):
Just one - minor - comment, but otherwise LGTM!

skyportal/handlers/api/allocation.py
mcoughlin: Done.

PR #5062, opened 2024-06-24
Websocket action: photometry refresh action name

Theodlz:
We had this weird situation where the `FETCH_SOURCE_PHOTOMETRY` emitted by the function to well fetch photometry, was also used as the action name for the websocket. It's confusing to read it (as it differs from everywhere else in the app) and it (though fortunately it looks like it doesn't) trigger some funky never ending refresh loop.

This PR updates `FETCH_SOURCE_PHOTOMETRY` to `REFRESH_SOURCE_PHOTOMETRY` for all the backend -> frontend updating logic with websockets.

### DIFF
--- a/skyportal/handlers/api/photometric_series.py
+++ b/skyportal/handlers/api/photometric_series.py
@@ -1886,7 +1886,7 @@ class PhotometricSeriesHandler(BaseHandler):
             session.commit()
 
             self.push_all(
-                action="skyportal/FETCH_SOURCE_PHOTOMETRY",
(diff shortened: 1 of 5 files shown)


mcoughlin (approved):
Small downside to this manual naming scheme for these ;)

Theodlz:
> Small downside to this manual naming scheme for these ;)

with my new IDE with the git blame on every line, I get a daily reminder that the real downside is just PEOPLE 👽

PR #5067, opened 2024-06-27
check for existing invite username

mcoughlin:
This PR exists for invite username.

### DIFF
--- a/skyportal/onboarding.py
+++ b/skyportal/onboarding.py
@@ -26,7 +26,9 @@ def create_user(strategy, details, backend, uid, user=None, *args, **kwargs):
     invite_token = strategy.session_get("invite_token")
 
     try:
-        existing_user = DBSession().query(User).filter(User.oauth_uid == uid).first()
+        existing_user = (
+            DBSession().scalars(sa.select(User).where(User.oauth_uid == uid)).first()
+        )
 
         if cfg["invitations.enabled"]:
             if existing_user is None and invite_token is None:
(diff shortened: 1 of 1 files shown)


skyportal/onboarding.py
```diff
@@ -26,7 +26,9 @@ def create_user(strategy, details, backend, uid, user=None, *args, **kwargs):
     invite_token = strategy.session_get("invite_token")
 
     try:
-        existing_user = DBSession().query(User).filter(User.oauth_uid == uid).first()
+        existing_user = (
+            DBSession().scalars(sa.select(User).where(User.oauth_uid == uid)).first()
```
Theodlz, line 30: please replace with .scalar(), rather than `scalars().first()`.

skyportal/onboarding.py
```diff
@@ -45,10 +47,10 @@ def create_user(strategy, details, backend, uid, user=None, *args, **kwargs):
 
             invitation = (
                 DBSession()
-                .query(Invitation)
-                .filter(Invitation.token == invite_token)
+                .scalars(sa.select(Invitation).where(Invitation.token == invite_token))
```
Theodlz, line 50: same comment here

skyportal/onboarding.py
```diff
@@ -62,6 +64,14 @@ def create_user(strategy, details, backend, uid, user=None, *args, **kwargs):
                     "Authentication Error: Invite token has already been used."
                 )
 
+            check_username = (
+                DBSession()
```
Theodlz, line 68: same comment here

Theodlz (approved):
LGTM, minor comments just to be consistent with the sqlalchemy 2 syntax from baselayer.

skyportal/onboarding.py
mcoughlin: Done.

skyportal/onboarding.py
mcoughlin: Done.

PR #5068, opened 2024-06-27
Public Page : Add summary

antoine-le-calloch:
Display the source summary on the public page if include in the option.
Display options field only if there are data to filter (photometry, classifications or summary)


Theodlz (approved):
LGTM 🚀

PR #5072, opened 2024-06-29
GCN: markdown summaries + primary/secondary display for ZTF obs planning

Theodlz:
* add a primary/secondary toggle when planning obs with ZTF
* make the fields slightly transparent to see the contour underneath the fields
* show the 50th and 90th contour, make them both colored (with 50th being darker)
* render the GCN summary tables in markdown and add styling to a few text elements when generating the summary's text.
* add a proper markdown editor (with text edit + markdown rendered) ala hackmd.io on the frontend.
* when scanning candidates for multiple groups at once, make the buttons of the "saving" dropdown trigger their action on click instead of requiring to first select the option and then click.
* add more options to this "saving" menu, with one button for each group.

TODO: let the CI run and have the reviewer try the markdown stuff :)


mcoughlin (approved):
Huge step forward here.

PR #5073, opened 2024-06-29
GCN Summary bugfix: NaN/None -> empty str 

Theodlz:
* make sure that the tns_name and redshift columns are empty and not showing `None/nan` if there is no value to display.
* add some padding to the tables when rendering the markdown


Theodlz:
self merging

PR #5095, opened 2024-07-01
py3-validate-email 1.0.12 -> 1.0.5.post2

Theodlz:
looks like 1.0.12 doesn't exist on pypi anymore?


mcoughlin:
@Theodlz Oh right, I read about this. We should really see if we can move off this packagge.

Theodlz:
yeah it's clearly a dead package...

PR #5096, opened 2024-07-01
ESLINT config update, CommentHandler PUT bugfix, DefaultObservationPlanTable.jsx bugfix, py3-validate-email -> email_validator

Theodlz:
a salad of bugfixes and updates:
- update the ESLint config after dependabot's last update
- Don't remove group access by accident when updating comments without specifying groups again
- don't let the allocation page crash if existing default obs plans don't have an auto_send parameter
- stop using py3-validate-email in favor of the better maintained (at least when it comes to pypi) email-validator


mcoughlin (approved):
LGTM

PR #5097, opened 2024-07-01
ObservationPage, InstrumentPage, AnalysisServicePage: V2

Theodlz:
Update the ObservationPage, InstrumentPage, and AnalysisServicePage to have this cleaner full width data table + add icon in the toolbar + edit button for each entry model, with the optional "tabs" to switch between tables of the same page.


mcoughlin (approved):
LGTM

PR #5101, opened 2024-07-03
Automated source names

mcoughlin:
This PR fixes the automated source names from GCN events.

### DIFF
--- a/services/gcn_service/gcn_service.py
+++ b/services/gcn_service/gcn_service.py
@@ -142,7 +142,8 @@ def poll_events(*args, **kwargs):
                         payload['notice_type']
                         == "gcn.notices.icecube.lvk_nu_track_search"
                     ):
-                        if payload.get("pval_bayesian", 1) > 0.05:
+                        # 3 sigma
+                        if payload.get("pval_bayesian", 1) > 0.003:
                             continue
 
                 tags_intersection = list(set(tags).intersection(set(reject_tags)))
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -384,8 +384,6 @@ def post_skymap_from_notice(
                     if name is not None:
                         name = name.text
 
-                print(ra, dec, error, SOURCE_RADIUS_THRESHOLD, tags)
-
                 if name is not None:
                     source = {
                         'id': name.replace(' ', ''),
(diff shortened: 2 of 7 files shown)


services/gcn_service/gcn_service.py
```diff
@@ -142,7 +142,8 @@ def poll_events(*args, **kwargs):
                         payload['notice_type']
                         == "gcn.notices.icecube.lvk_nu_track_search"
                     ):
-                        if payload.get("pval_bayesian", 1) > 0.05:
+                        # 3 sigma
+                        if payload.get("pval_bayesian", 1) > 0.003:
```
Theodlz, line 146: the exactly 0 is also there on purpose? Just checking.

Theodlz: went from .05 to .003

mcoughlin: @Theodlz yes 3 sigma :)

skyportal/handlers/api/gcn.py
```diff
@@ -390,29 +390,25 @@ def post_skymap_from_notice(
                         'ra': ra,
                         'dec': dec,
                     }
-                elif any(
-                    [True if 'GRB' in tag.text.upper() else False for tag in tags]
-                ):
+                elif any([True if 'GRB' in tag.upper() else False for tag in tags]):
```
Theodlz, line 393: why is `.text` not needed? Have the tags been converted to text somewhere above? Is that true for all notices (and not just JSON)?

skyportal/handlers/api/gcn.py
```diff
@@ -427,7 +423,7 @@ def post_skymap_from_notice(
 
                 if source.get('id', None) is not None:
                     existing_source = session.scalars(
-                        Source.select(user).where(Source.id == source['id'])
+                        Source.select(user).where(Source.obj_id == source['id'])
```
Theodlz, line 426: nice catch

skyportal/handlers/api/gcn.py
mcoughlin: yeah these tags are all just the plain text.

mcoughlin: This is only the case where its like gcnevent.tags.

Theodlz: got it, yeah it was just confusing that .text all of a sudden had to go away but nothing else upstream of that seems to have changed.

Theodlz (approved):
LGTM 🚀 -> EDIT: lots of test failing, to fix before merging

Theodlz (changes requested):
tests need to be looked at and fixed if necessary

mcoughlin:
Merged in the other PR

PR #5105, opened 2024-07-04
gcnevent advocate

mcoughlin:
This PR adds concept of event advocate. It's simply a flag at the top of the page to let folks know you got it, useful for cases where you want to know someone is paying attention (and can therefore go back to sleep).


Theodlz:
@mcoughlin I'll try to take a look today, but this would definitely be a good feature.

Theodlz (commented):
I made a few very minor changes + added some code of my own (when we autogenerate sources from well-localized events, adds an "origin" (e.g. Swift, Fermi, LVC, Einstein Probe) that we can use as a parameter for the automated follow-up request stuff (e.g. only trigger on Swift GRBs and not Fermi ones).

I'll wait for the tests to run and will fix the migration once I've merged Antoine's data release PR and his tests

PR #5107, opened 2024-07-05
user deletion icon

mcoughlin:
This PR adds a user deletion icon.


Theodlz (commented):
Why would we ever want to delete users? I feel like we would risk deleting many data products due to the relationships we have between tables.

Just like for sources, we should probably keep them but add a "active" column that we turn to false, same thing for allocations, instruments, telescopes, ... but that's particularly true for users here. That would delete stuff like classifications, comments, photometry they uploaded, ...

mcoughlin:
@Theodlz Yeah maybe we can retire some accounts? I agree that's better than deletion, although it would be nice for it to be a separate table in the user page.

PR #5108, opened 2024-07-06
filter expired accounts

mcoughlin:
This PR filters expired accounts in user table.


Theodlz (approved):
LGTM 🚀. What about deactivated account? I looked at the user model in baselayer and looks like we already do have an "active" column to mark accounts as deactivated and not deleted them (referring to your other PR).

PR #5109, opened 2024-07-06
Candidate & Source HEAD: raw sql for speed

Theodlz:
Same treatment as for the `get_sources` method, applied here to the HEAD endpoints for candidates and sources, hoping to speed-up these API queries that are run in large quantities when Kowalski is trying to send data to Fritz during the night.


Theodlz:
DNM! I want to assess performance by running these queries (and alternatives) in prod before merging.

mcoughlin (approved):
LGTM

Issue #5103, opened 2024-07-04
Tiles covers the center


santier14:
Please fill out relevant sections below; remove those that are unused.

**Describe the bug**
When there are tiles, the center is covered by the tiles (see some exemple)
![center_single](https://github.com/skyportal/skyportal/assets/16883193/189cb14a-e06c-4236-8404-ff92b12e7b3b)

**To Reproduce**

PR #5111, opened 2024-07-08
Obsplan List: tiles transparency

Theodlz:
fixes https://github.com/skyportal/skyportal/issues/5103, and reverses the primary and secondary field selector for ZTF (that were reversed).


mcoughlin (approved):
LGTM

PR #5113, opened 2024-07-09
Default observation plan filters: gcn & localization properties

Theodlz:
Adds filtering on gcn and localization properties, not just the tags and notice types.


mcoughlin (approved):
LGTM

PR #5114, opened 2024-07-09
transactions for follow-up requests

mcoughlin:
This PR shows transactions for follow-up requests.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -4,6 +4,7 @@ import io
 import operator  # noqa: F401
 import time
 import traceback
+import json
 from json.decoder import JSONDecodeError
 
 import re
@@ -230,7 +231,7 @@ async def get_source(
         raise ValueError("Source not found")
 
     source_info = s.to_dict()
-    source_info["followup_requests"] = (
+    followup_requests = (
         session.scalars(
             FollowupRequest.select(
                 user,
@@ -241,6 +242,7 @@ async def get_source(
                     joinedload(FollowupRequest.allocation).joinedload(Allocation.group),
                     joinedload(FollowupRequest.requester),
                     joinedload(FollowupRequest.watchers),
+                    joinedload(FollowupRequest.transactions),
                 ],
             )
             .where(FollowupRequest.obj_id == obj_id)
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/source.py
```diff
@@ -241,6 +242,7 @@ async def get_source(
                     joinedload(FollowupRequest.allocation).joinedload(Allocation.group),
                     joinedload(FollowupRequest.requester),
                     joinedload(FollowupRequest.watchers),
+                    joinedload(FollowupRequest.transactions),
```
Theodlz, line 245: You can specify the column too in the joinedload with `joinonly`. that way you don't query all the stuff you don't need (so you can get only the content that you use after).

skyportal/handlers/api/source.py
```diff
@@ -249,6 +251,22 @@ async def get_source(
         .unique()
         .all()
     )
+
+    followup_requests_data = []
+    for req in followup_requests:
+        req_data = req.to_dict()
```
Theodlz, line 257: no need for the to_dict() here, that serializes all the columns you don't use too.

Theodlz (changes requested):
small comments to avoid loading data we don't need.

skyportal/handlers/api/source.py
mcoughlin: These requests were already in the source, so I think we want this.

skyportal/handlers/api/source.py
mcoughlin: added.

skyportal/handlers/api/source.py
Theodlz: Oh right apologies I thought the to_dict ran on on the transaction, my bad!

Theodlz (approved):
LGTM. One thing I am worried about is that some instruments might leak secrets from the allocation in their API response. I vaguely remember this happening for one or two instruments already, but I guess we'll see :). If this is just for debug and not everyone needs one, we should consider only querying and returning the transactions if the user is an admin, that might make more sense.

mcoughlin:
@Theodlz great idea

PR #5115, opened 2024-07-10
User Profile: show oauth email

Theodlz:
An attempt to help people understand that there is a fixed email used for login, so that they do not think the `email` in the "contact" section is the same thing.


mcoughlin (approved):
BAN ALL EMAILS

PR #5116, opened 2024-07-10
AssignmentList (on SourcePage): edit

Theodlz:
let users edit the priority and comment for an assigned target, which I noticed was quite annoying as one had to delete and re-assign the target.


mcoughlin (approved):
LGTM

PR #5140, opened 2024-07-12
add parseFloat to period annotation

### LINKED ISSUE #5139: Manual saving of period annotation does not recognize Period as number
When trying to save a period (with many decimal point) as an annotation on the web ui the error " 'Period' must be a number" appears. The issue is resolved if I highlight and copy the period in section, copy it and re-paste it into the same box and resubmit.

**To Reproduce**

1. Go to a source page
2. Click on Period in the photometry section
3. Paste a long string into the period section (like 0.1491956943056191, I have not tried it with less decimals)
4. Try to save annotation with the save icon

**Expected behavior**
Annotation saves with no issue

**Screenshots**
Screen Shot 2024-07-12 at 2 06 32 PM

**Platform information:**
This is Fritz v0.9.dev0+git20240710.8a4f63a3+fritz.43bd82b.

mcoughlin:
This PR adds a missing parseFloat in period annotation.

Closes #5139.

### DIFF
--- a/static/js/components/plot/PhotometryPlot.jsx
+++ b/static/js/components/plot/PhotometryPlot.jsx
@@ -153,7 +153,6 @@ const PeriodAnnotationDialog = ({ obj_id, period }) => {
   };
 
   const validate = (formData, errors) => {
-    console.log("period", formData.period);
     if (formData.period <= 0) {
       errors.period.addError("Period must be greater than 0");
     }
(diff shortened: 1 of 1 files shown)


Theodlz (approved):
LGTM. I wonder if we should address address another issue in that PR. Say you are trying periods and picked something else than days as the unit (hour, minute, ...). Saving the period doesn't account for that, when it should convert it to days.

mcoughlin:
Great point @Theodlz. See new code.

PR #5149, opened 2024-07-16
Upload Single Photometry: simpler instrument select

Theodlz:
Don't show the full list of filters of an instrument in the dropdown menu, which breaks the display when the instrument has more than a few filters, and isn't that useful given that the filters are selectable in a dropdown in the form anyway.


mcoughlin (approved):
LGTM

PR #5152, opened 2024-07-18
observing run improvements

mcoughlin:
This PR makes some small improvements to the observing run page.

[ x ] It would be useful to have a visual indicator to more obviously distinguish observed, unobserved, and not-observed targets.   Maybe the "status" column entry could be colourized, or something like that.   (On the old Marshal a big green check mark icon would appear for observed target, and a big red X for not-observed targets.)   This would help observers check things off (or strike them out) as they go through a run.

[  ] The Lick/P200 telescope operators like having their own copy of the finding charts for acquisition.  Since they don't have Fritz access we've been linking them to a Google Drive folder for this, or sending an e-mail with attachments.  It would be easiest if we could just download every finder for a target on the run all at once as a bulk download, such that for runs with loads of targets (our last one had 50+) we don't have to go one by one and end up accidentally forgetting some, etc.   (Another solution would be a special public-access page that is autogenerated for every run, containing static finders.)

[ x ] A button on each observing run page to auto-reassign a target from the current run to a future run would be nice.

[ x ] It would be useful to have a single button on each observing run page that, if clicked, would change every "pending" target on that page to "not observed".  (Particularly for runs that are entirely clouded out.)   (Probably this should have a verification pop-up to make sure it's not clicked by accident.)

[ x ] It would be nice to be able to have the date added/requested as a column on the observing run page.  Sometimes people add new targets at the last minute, and for the observer it's hard to tell what's changed/new since the last time they prioritized/scheduled targets for that run.  It would also be good to have this info in the observing requests section of the Source page.

[ x ] Once an observing run is created, it seems there's no way to edit it - i.e. to add an observer or fix a mistake in the PI name or something.  It would be useful to have an option on the list of observing runs to change the properties of a run.

[ x ] Some observing runs consist of multiple nights in a row.  Usually these are planned all together, i.e. use a common target list / starlist.   Being able to explicitly specify a run lasts more than 1 night might be a useful feature, particularly since it would ensure that it doesn't become unavailable for assignment midway through.

[ x ]  Related: On the list of Observing Runs page, after midnight an ongoing run (even a 1-night run) disappears from the list of (upcoming) runs that appears there initially by default, and requires choosing the option to show every run and then finding/searching through that list.  Would be nice if it stayed until at least noon on the day the run eends.

 
[ (small improvement where you can specify the telescope) ] As I mentioned to you this afternoon, we had a case where the finding chart displayed the offset stars in a different order from the starlist, perhaps because the finding chart is always based on the Keck list while the starlist is customizable per telescope, or because the position of the transient shifted very slightly and the order of stars changed as a consequence.  This can lead to using the wrong offset and ending up pointing at blank sky.  It would be useful if the offset star "names" were always fixed across all options for finding charts, even if the transient "moves" or a different telescope is selected.


Theodlz:
@mcoughlin any chance to list the items from Dan's list that this addresses? Just to keep track of what needs to be done and what I should look for when using the frontend

mcoughlin:
@Theodlz edited the PR.

Theodlz:
Awesome, thanks! I'll review it this morning and make some edits before merging if needed. That way I can deployed this on Fritz this afternoon which would be nice (there is a run tonight).

Theodlz:
This is great @mcoughlin. I did notice one super weird bug. When editing a run to say move it to another date, it didn't edit it. Instead it looks like it created a new once. On top of that, the targets in the new run are not the right ones! It looks as if it editing another run, not the one I clicked on.

Same issue with deleting, it deleted the wrong run! That could create all sorts of issues. I think that we went to prevent deleting runs with pending/completed targets in general.

Let me edit and merge this tomorrow, I can't make these edits tonight.

I will deploy prod anyways though, but yeah let's wait before we merge this.

PR #5156, opened 2024-07-24
Correct thumbnails duplication

antoine-le-calloch:
- Add logic to retain only the last thumbnail of each type retrieved with the get_source method.
- add verification when generating a public page to avoid duplication of thumbnails already published.


mcoughlin (approved):
LGTM

skyportal/handlers/api/source.py
```diff
@@ -231,8 +231,23 @@ async def get_source(
     s = session.scalar(stmt)
     if s is None:
         raise ValueError("Source not found")
-
     source_info = s.to_dict()
+
+    # retain only the last thumbnail of each type
+    if include_thumbnails:
```
Theodlz, line 237: May I ask why this is also needed in the normal source handler (not the public pages)? It's fine, just wondering.

And if so, might want to just make the database query that gets the thumbnails a bit smarter instead of doing it later on in python. Might be quicker if the query is well written, though both are valid methods.

Theodlz (commented):
I had a quick question. Also, did you end up not putting code to prevent that from happening for already generated public pages?

antoine-le-calloch:
> I had a quick question. Also, did you end up not putting code to prevent that from happening for already generated public pages?

Yeah sure, i've had it to the public page generation to work also on the page already publish.

skyportal/handlers/api/source.py
antoine-le-calloch: Its because we send unused data on the front-end without this. It was to avoid this type of behavior. And I didn't do it in the query because to do it correctly, you have to check the other thumbnails and compare them to avoid duplication and that seemed complicated to me in a query. But I'll take a look.

Theodlz: Thanks! Yeah ideally we want to just not even get the data from the DB when we can, it's ideal. But depending on the query planning and performance we sometimes end up doing the filtering in python. It's not really about whether or not the query is complex, but it's performance (that can weirdly differ, sometimes complex queries run faster than simpler ones, depends of the index and the amount of data retrieved). Here my feeling is that it would be equivalent so we could also just keep it as is. If you can't figure out a query by 11AM (today's meeting), let's just keep the current pythonic version.

Theodlz (approved):
LGTM 🚀

PR #5158, opened 2024-07-24
ZTF Executed Observations: fallback

Theodlz:
Fallback to an alternative file on the IPAC depot when the usual one is missing.

TODO: figure out which column to use for the seeing. In the current file we use a `fwhm` column, but here we have `diffhwm` and `sciseeing`.


Theodlz:
any idea about that `seeing` vs `fwhm` thing? Here are all the columns from that other file:

       jd        | field | rcid |        ra0        |       dec0       | nalertpackets | programid |   expid   | fid | scimaglim | diffmaglim | sciinpseeing | difffwhm | exptime | subtractionstatus | jobexitcode

PR #5159, opened 2024-07-26
LT front-end fixes

mcoughlin:
This PR makes some front-end fixes for LT.


Theodlz (approved):
LGTM 🚀. Maybe throw a comprehensive error when the mode is not in the expected list of options? Also, should we wait for Dan to confirm that these values should be fixed?

Regardless, happy to merge this.

mcoughlin:
@Theodlz Yeah let's wait for Dan/Ola. I added the mode in the error string. Were there any changes you made while looking at the script earlier in the week to include here? Maybe a try except that would have thrown a better error?

PR #5161, opened 2024-07-27
 Public page: Fix some behavior api public source page

antoine-le-calloch:
Delete the bad use and useless of nbResults to filter the public source pages fetch number. On the back and front-end.
Correct a behavior with the get_source which rase an exception when source not found.


mcoughlin (approved):
LGTM

antoine-le-calloch:
Ready for review !

Theodlz (approved):
LGTM 🚀

PR #5162, opened 2024-07-30
Notes design improvement

antoine-le-calloch:
Add notes panel to keep track of the warning and info pop-up that come from baselayer.
Add a number in a note which count how many notes with the same content follow this one, to avoid duplication note pop-up.


Theodlz (approved):
LGTM! 🚀. Looks great, we might want to think about the feature a little bit more in the future, but this is super useful already.

PR #5163, opened 2024-08-01
Corrections from security audit

antoine-le-calloch:
Update yaml.load to yaml.safe_load to protect against code injection.


mcoughlin (approved):
LGTM

PR #5164, opened 2024-08-01
Correction make load demo data

antoine-le-calloch:
- Correct the load demo data problem by fixing the flux2mag function in photometry_series, which incorrectly used the where argument in log10().
- Correct some tests in flaky folder which are link to photometry.


mcoughlin (approved):
LGTM

PR #5188, opened 2024-08-05
WIP - FacilityApis: formSchemaAltdata + use it in NewAllocation.jsx

Theodlz:
For most FollowupAPI classes we implement in `skyportal/facility_apis`, this PR adds a `formSchemaAltdata` field which is essentially its own RJSF form schema. In the frontend, we concatenate that schema with the NewAllocation.jsx schema to create a new allocation, when the instrument is selected.

The idea here is to make it much easier for folks to set up allocations by:
- preventing having them input this somewhat awkward JSON string that needs double and not single quotes to work
- directly showing them all the possible fields so they don't need to figure out where to find them in the documentation.

Note: This is a WIP, I need to go over the code a couple of times and make sure that the altdata schemas are correct, and ideally set this up for the EditAllocation too. Also right now once the backend receives the dict we serialize it to string to keep using the schema validation from the `skyportal/models/allocation.py::Allocation` itself, but maybe we could validate the dict as is.


mcoughlin (approved):
I pulled it down and made some allocations, but didn't populate any of them with real data, but the overall experience is much nicer than it is now. Really good.

PR #5189, opened 2024-08-07
Add filter to notes

antoine-le-calloch:
Improve the notes panel by not displaying the notes that contain "No WebSocket connection".


Theodlz:
@antoine-le-calloch please clean your main branch, the PRs are starting to look really weird will all those merge commits. Your main branch should always be identical to main not just in content but in commit history.

PR #5192, opened 2024-08-07
GCN event ingestion: extra logging

Theodlz:
We're having some curious issues that happen somewhat randomly when processing some event's localizations, where the apps or gcn service just restart (meaning they just die all of a sudden) without throwing any error whatsoever.

This PR adds some extra logs and try/catch statements to try to avoid some errors and narrow that down.


mcoughlin (approved):
LGTM

PR #5195, opened 2024-08-09
Public page: Automatically publish source when created

antoine-le-calloch:
- Allow users to select the "automatically_publish" option for a release. When this option is enabled, any new or updated source that enters the same group as that release will be automatically published.
- create tests for the release API
- add tests for the automatically published process when create a source or update source groups.
- improve the test for public_source_page API


Theodlz (changes requested):
I know this might not be ready for review yet, but I gave it a go anyway. Some comments:
- Say my release have groups A and B. If I save to A, then later save to B before any data on the source has changed, I don't think we need to show the "there is already a release with the same ..." message.
- Linked to the issue above, but when there is a release-related error message pushed to the frontend, it doesn't update the source even though it has correctly been saved to the group!
- Same example: say I have a release with groups A and B. I would imagine that once a source is unsaved from A and B, any release that was autogenerated should be removed, or made invisible at least. That makes me think that a flag to track what releases were autogenerated and which were generated manually would be nice.
- I think that in general you can wrap your code that autogenerates/deletes public pages given a list of group ids (to which a source was saved to, and unsaved from) in a method, that you can then call with the `run_async` method from `skyportal/utils/asynchronous.py`. That way the handler to save sources can return safely undisturbed by that addition, and refresh the source page properly.

antoine-le-calloch:
> I know this might not be ready for review yet, but I gave it a go anyway. Some comments:
> 
>     * Say my release have groups A and B. If I save to A, then later save to B before any data on the source has changed, I don't think we need to show the "there is already a release with the same ..." message.
> 
>     * Linked to the issue above, but when there is a release-related error message pushed to the frontend, it doesn't update the source even though it has correctly been saved to the group!
> 
>     * Same example: say I have a release with groups A and B. I would imagine that once a source is unsaved from A and B, any release that was autogenerated should be removed, or made invisible at least. That makes me think that a flag to track what releases were autogenerated and which were generated manually would be nice.
> 
>     * I think that in general you can wrap your code that autogenerates/deletes public pages given a list of group ids (to which a source was saved to, and unsaved from) in a method, that you can then call with the `run_async` method from `skyportal/utils/asynchronous.py`. That way the handler to save sources can return safely undisturbed by that addition, and refresh the source page properly.

- Wrap the code with 'run_async'. Done ✅ !
- So don't display public page creation error when its automatic. Done ✅ !
- And allow update the source when there is page creation error. Done ✅ !
- Add flag, to track and delete page automatically publish. In process ⏳!

alembic/versions/d05256bc14ed_automatically_publish_source_when_.py
```diff
@@ -0,0 +1,37 @@
...
+revision = 'd05256bc14ed'
+down_revision = '3eb663b5c3e0'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        'publicreleases',
```
Theodlz, line 23: weird that the 2 columns have almost the same name but not really. Please make sure that they are identical if they mean the same thing

Theodlz: this comment applies to all the files that use these columns

skyportal/handlers/api/public_pages/public_release.py
```diff
@@ -182,6 +184,10 @@ def patch(self, release_id):
                 groups = session.scalars(
                     Group.select(session.user_or_token).where(Group.id.in_(group_ids))
                 ).all()
+            else:
+                return self.error("Invalid groups")
+            if not groups:
```
Theodlz, line 189: not sure if this is a git diff issue, but it looks like the `if not groups` is duplicated. 

Also if you have an if -> else, followed by another if, and the else and the second if return the same error, I wonder if this could be reformatted to avoid having the same 2 returns.

skyportal/handlers/api/source.py
```diff
@@ -939,6 +940,32 @@ def post_source(data, user_id, session, refresh_source=True):
...
 
+        # if there is releases with automatically_publish and one of the source groups,
+        # a public page is published
+        releases = session.scalars(
+            PublicRelease.select(session.user_or_token).where(
+                PublicRelease.groups.any(id=group.id),
+                PublicRelease.automatically_publish,
+            )
+        ).all()
+        if releases is not None and len(releases) > 0:
```
Theodlz, line 951: i don't think it could be None here right? I suppose that the ORM would always return lists. Worth checking to avoid unnecessary if statements

Theodlz: you could also imagine to do `if len(releases) == 0 -> return or whatever you need to do because returning` rather than `if len(releases) > 0 -> a ton of logic that needs to be indentated`. I like to follow this "never indent" principle where possible. Just an advice! Not a requirement at all :)

Theodlz: the less indentation, the easier to read

skyportal/handlers/api/source_groups.py
```diff
@@ -49,7 +52,11 @@ def post(self):
                       List of group IDs from which specified source is to be unsaved.
                 required:
                   - objId
-                  - inviteGroupIds
+                anyOf:
```
Theodlz, line 55: ah didn't know you could use an anyOf here, nice!

skyportal/handlers/api/source_groups.py
```diff
@@ -121,6 +128,22 @@ def post(self):
                 source.active = False
                 source.unsaved_at = datetime.datetime.utcnow()
 
+            if len(unsave_group_ids) > 0:
+                # Delete all this source auto-published pages, link to a release and sharing no groups with this source
```
Theodlz, line 132: i think the sentence sounds a little odd. the comment could be shortened if not removed

Theodlz (changes requested):
Hey @antoine-le-calloch. Thanks a lot for the changes and the clean up! I have a few minor comments, please take a look.

Theodlz (approved):
LGTM! Nice work Antoine. Let's get this merged so we can try it out soon and see if anything needs to change

Theodlz:
@antoine-le-calloch I just merged main to your branch to see if everything is still fine. When green (or if no related tests fail), I'll go ahead and merge it

Theodlz:
@antoine-le-calloch there is some conflit with what has been merged on main today (introduced by https://github.com/skyportal/skyportal/pull/5417), please take a look and let me know once resolved, and if it's all green I'll merge it

skyportal/handlers/api/public_pages/public_source_page.py
```diff
@@ -80,6 +81,17 @@ def get_photometry(source_id, group_ids, stream_ids, session):
     return [photo.to_dict_public() for photo in session.scalars(stmt).unique().all()]
 
 
+def get_phot_stats(source_id, phot_stats, session):
+    if phot_stats is not None:
+        return phot_stats[0]
+
+    return session.scalar(
```
Theodlz, line 88: could you just wrap that in a try except or something? The query could return None if the source has no phot stats yet. Unlikely but could be a point of failure that we want to avoid.

skyportal/handlers/api/public_pages/public_source_page.py
```diff
@@ -133,12 +145,15 @@ def post_public_source_page(options, source, release, is_auto_published, session
...
         "dec": safe_round(source.get("dec"), 6),
+        "peak_mag_per_filter": None,
+        "first_detected_mjd": None,
         "redshift_display": get_redshift_to_display(source),
         "gal_lon": safe_round(source.get("gal_lon"), 6),
         "gal_lat": safe_round(source.get("gal_lat"), 6),
         "ebv": safe_round(source.get("ebv"), 2),
         "dm": safe_round(source.get("dm"), 3),
         "dl": safe_round(source.get("luminosity_distance"), 2),
+        "t0": safe_round(source.get("t0"), 6),
```
Theodlz, line 156: does that safe round behave correctly when there is no t0? Just wondering

Theodlz (changes requested):
One change request regarding the safety of that query, and then let's merge

skyportal/handlers/api/public_pages/public_source_page.py
antoine-le-calloch: Yep, this is why I have create this safe_round function.

Theodlz (approved):
LGTM!

PR #5196, opened 2024-08-09
Follow-up request: SWIFT ToO mode

Theodlz:
* the notification we show when clicking "retrieve" on a follow-up request had its type set to "success" when it should have been set to "info" to appear green.
* adds the SWIFT ToO mode


mcoughlin (approved):
LGTM

PR #5197, opened 2024-08-15
Correct test_show_starlist

antoine-le-calloch:
- Correct the test_show_starlist which searching for _o1 instead of _k1.


mcoughlin (approved):
LGTM

mcoughlin:
This PR corrects the test_show_starlist which searching for _o1 instead of _k1.

PR #5199, opened 2024-08-16
GCN - Point source auto-generation: larger radius threshold

Theodlz:
* increases the radius threshold from 5 arcmin to 8 arcmin
* fixes a bug introduced end of last week, that prevented EP sources from being created


Theodlz:
self merging

PR #5200, opened 2024-08-16
ZTF Treasuremap filters

Theodlz:
* force the ztf filters to r,g,i when sending to treasuremap, as otherwise treasuremap matches the central wavelength + bandpass of ztf i-band to TESS.


mcoughlin (approved):
LGTM

PR #5206, opened 2024-08-19
Obj Redshift - use existing z origin

Theodlz:
We already have a concept of redshift origin but we've been ignoring it from the frontend. This PR let's user set it from the frontend if desired, and then shows it on the redshift history table.


mcoughlin (approved):
LGTM!

PR #5207, opened 2024-08-19
check group_ids when making a source

mcoughlin:
This PR checks that group_ids are accessible when making a source.

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -692,14 +692,19 @@ def post_source(data, user_id, session, refresh_source=True):
         raise AttributeError(
             "You must belong to one or more groups before you can add sources."
         )
-    try:
-        group_ids = [
-            int(id)
-            for id in data.pop('group_ids')
-            if int(id) in user_accessible_group_ids
-        ]
-    except KeyError:
+
+    group_ids = data.pop('group_ids')
+    if group_ids is None:
         group_ids = user_group_ids
+    else:
+        group_ids = []
+        for id in group_ids:
+            if int(id) in user_accessible_group_ids:
+                group_ids.append(int(id))
+            else:
+                raise ValueError(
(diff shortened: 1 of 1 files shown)


skyportal/handlers/api/source.py
```diff
@@ -692,14 +692,19 @@ def post_source(data, user_id, session, refresh_source=True):
...
-            for id in data.pop('group_ids')
-            if int(id) in user_accessible_group_ids
-        ]
-    except KeyError:
+
+    group_ids = data.pop('group_ids')
+    if group_ids is None:
         group_ids = user_group_ids
+    else:
+        group_ids = []
```
Theodlz, line 700: don't think that one works 😉 @mcoughlin

Theodlz: you overwrite it and loop over it, but well it will always be empty

mcoughlin: Thanks bud! Should be fixed.

skyportal/handlers/api/source.py
```diff
@@ -692,14 +692,20 @@ def post_source(data, user_id, session, refresh_source=True):
...
-            for id in data.pop('group_ids')
-            if int(id) in user_accessible_group_ids
-        ]
-    except KeyError:
+
+    group_ids = data.pop('group_ids')
+    if group_ids is None:
         group_ids = user_group_ids
+    else:
+        group_ids_tmp = []
```
Theodlz, line 700: Thanks for fixing that! Though I feel like you could even have made it happen without any extra list. All that this does is validate that the ids are accessible to the user and then make sure they are ints. You could have also just made them ints with a one liner and then loop over them and raise the error when one of them was invalid.

But that aside, this seems to work just fine! 

LGTM. Merge once green.

Theodlz (approved):
One small comment

PR #5208, opened 2024-08-20
Followup constraint: ignore_allocation_ids

Theodlz:
add a followup constraint named ignore_allocation_ids, to cancel posting follow-up requests if there are existing submitted/completed requests with any of the allocations specified in that list.

This is to help bots that run on one allocation that want to avoid triggering if there is a request for another allocation already.

There is a PR on Kowalski here: https://github.com/skyportal/kowalski/pull/412


mcoughlin (approved):
LGTM

PR #5209, opened 2024-08-21
date-time in form

mcoughlin:
This PR use strings rather than form (which has funky local time behavior)

### DIFF
--- a/skyportal/facility_apis/lt.py
+++ b/skyportal/facility_apis/lt.py
@@ -119,8 +119,9 @@ class LTRequest:
             etree.SubElement(photom_const, 'Clouds').text = 'light'
 
         date_const = etree.Element('DateTimeConstraint', type='include')
-        start = request.payload["start_date"].replace("Z", "+00:00")
-        end = request.payload["end_date"].replace("Z", "+00:00")
+
+        start = request.payload["start_date"] + "+00:00"
+        end = request.payload["end_date"] + "+00:00"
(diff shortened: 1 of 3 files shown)


Theodlz (approved):
Nice. Didn't you mention on zoom that you overwrote some frontend stuff? Just making sure that automated date reset thing that's on the frontend is working still.

Also one thing to keep in mind is that `date-time` rjsf format works poorly if not incorrectly on IE (now edge), safari, and sometimes firefox. For that alone, going back to a proper date time input (at least the one from rjsf and not a custom MUI-based one) will require more work then just fixing the UTC vs local thing.

LGTM

Theodlz:
> Nice. Didn't you mention on zoom that you overwrote some frontend stuff? Just making sure that automated date reset thing that's on the frontend is working still.
> 
> Also one thing to keep in mind is that `date-time` rjsf format works poorly if not incorrectly on IE (now edge), safari, and sometimes firefox. For that alone, going back to a proper date time input (at least the one from rjsf and not a custom MUI-based one) will require more work then just fixing the UTC vs local thing.
> 
> LGTM

They mention it here: [/docs/usage/widgets/, ](https://rjsf-team.github.io/react-jsonschema-form/docs/usage/widgets/), and I definitely had these issues too.

mcoughlin:
Yeah but the code there doesn't seem to work in the case of date-time, so switching it off still makes sense I think.

mcoughlin:
I will put the old behavior back, but we should look into this more.

PR #5211, opened 2024-08-23
do not split date if want full date

mcoughlin:
This PR only uses the split on the date if specifying date format (need full isot in some cases).

You might have a better way of doing this @Theodlz

### DIFF
--- a/static/js/components/followup_request/FollowupRequestForm.jsx
+++ b/static/js/components/followup_request/FollowupRequestForm.jsx
@@ -296,12 +296,16 @@ const FollowupRequestForm = ({
       const range = endDate - startDate;
       const newStartDate = new Date();
       const newEndDate = new Date(newStartDate.getTime() + range);
-      schema.properties.start_date.default = newStartDate
-        .toISOString()
-        .split("T")[0];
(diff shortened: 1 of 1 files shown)


Theodlz:
@mcoughlin so this piece of code is already in an if statement where we do things differently for followup requests and forced photometry right? If so, just leave the time for followup requests and only pop it out for the forced phot.

mcoughlin:
You do them for both forced photometry and normal requests @Theodlz

Theodlz:
> You do them for both forced photometry and normal requests @Theodlz

yes that's what I meant. Like in the existing if statement, you can just remove the split when its a followup request in general and the field is called start_date or end_date, like not sure we really need to check that its format === date

Theodlz:
just if follow-up request and not forced photometry, do the isostring but no split ?

Theodlz:
for forced phot we def not need the time its a large enough time range, but for the follow-up requests I feel like it's pretty much always useful

mcoughlin:
Any chance you can just fix it how you want it @Theodlz ?

Theodlz:
Sure, will do in a little bit.

PR #5213, opened 2024-08-25
Followup constraint schema: fix

Theodlz:
In a previous PR, and incorrect schema was added, missing some changes.

This PR fixes it. Hotfix in production already


Theodlz:
self merging

PR #5216, opened 2024-08-28
Comment Attachment Preview: simpler UI

Theodlz:
This PR makes a few minor changes to help use more of the screen to display attachments, with:
- smaller DialogTitle bar at the top of the dialog
- replace the "Comment attachment" by the filename (user knows they already are looking at the attachment they just clicked on), so we can remove the filename from where it was previously and regain some height.
- make the dialog a little higher and wider
- replace the "DOWNLOAD" text-based button by a download MUI icon, which we use everywhere else already anyway.


mcoughlin (approved):
LGTM

PR #5218, opened 2024-08-29
get_sources: phot series session.execute fix

Theodlz:
Fixes a call to session.execute() when fetching photometric series, which we never noticed because it is only called when there is no photometry.


Theodlz:
self merging

PR #5219, opened 2024-08-29
Autogenerated GCN sources: GRB SWIFT type

Theodlz:
Differentiate between XRT and BAT, so we can select what to auto trigger on with the default follow-up requests.


mcoughlin (approved):
LGTM

PR #5220, opened 2024-08-29
SDSS thumbnails: dr12 -> dr18

Theodlz:
Move to DR18 since some of DR12s services seem to be down when DR18's aren't.


mcoughlin (approved):
LGTM

stefanv (approved):

PR #5221, opened 2024-08-30
GCN fixtures

Theodlz:
For a long time now, we had a lot of flaky tests. A good chunk of those (for the backend/API tests) are caused by all of the GCN-related operations. When we post an event from the API, some things are async, like processing the localization that we need for some of said tests.

This PR adds GCN fixtures, so that when we need these events to test other features, we do not need to keep posting/deleting from the API, but rather load it when starting the tests (just like we do for pretty much everything else in the app).

Some of the data that is generated when posting events takes some time to be generated everytime, but is always identical. So, to speed up loading the GCN Localization fixtures, we load the values computed from the SkyPortal directly from disk. Generating those is very easy. One has to simply run the API tests that add this event to the DB, and dump the localization & localization tiles table to disk. The only thing I did here is then cleanup the CSVs to only include the columns we need, to keep it as light as possible. I also made them parquet files (brotli compressed) to make them as small as possible.

TODO:
 - [ ] Instruments have the same issue, where loading instrument tiles is async and can create some issues. We've trimmed down the number of tiles however, so it might be quick enough that we don't have that problem.
 - [ ] We also keep posting/deleting galaxy catalogs in a couple of tests. We could consider making those fixtures too.

*These TODOs can be addressed in a separate PR*


mcoughlin (approved):
Definitely a great initiative.

PR #5245, opened 2024-09-03
Swift error logs

Theodlz:
Here we remove the inconvenient `r.raise_for_status()` to properly log successful & failed requests as a transaction, + we add some backend logging and display the error on the front end to make it easier to track.


mcoughlin (approved):
LGTM

PR #5247, opened 2024-09-03
GitHub actions: dustmap & sncosmo filters caching

Theodlz:
This is an attempt at caching some of the stuff that we keep downloading when we start github actions.

Might have to commit a few times to see this in action.


mcoughlin (approved):
This is great.

PR #5249, opened 2024-09-04
Build & Deploy docs GitHub action: auto delete cancelled

Theodlz:
Based on a suggestion from @stefanv, this PR is an attempt at using the github action from https://github.com/MercuryTechnologies/delete-cancelled-runs.

The problem we are facing is that the github action we rely on to build&deploy the docs already cancels previous workflows that are still running, except that when workflow are just cancelled they make the overall test suite appear as failed. Deleting cancelled workflows makes a lot of sense here, and prevents the tests from appearing as failed.

Also, I made small changes to the `labeler.yaml` config, because pretty much all of our PRs ended up with this `needs-migrations?` label. I think that it needs merging for that behavior to stop.

### DIFF
--- a/.github/workflows/build-and-deploy-docs.yaml
+++ b/.github/workflows/build-and-deploy-docs.yaml
@@ -14,7 +14,7 @@ jobs:
   delete-cancelled-runs:
     runs-on: "ubuntu-latest"
     steps:
-      - uses: "MercuryTechnologies/delete-cancelled-runs@v1.0.0"
+      - uses: "MercuryTechnologies/delete-cancelled-runs@f5c9d322d9c7afbd41f779199818c1ffe381d34b" #v1.0.0
         with:
           workflow-file: "build-and-deploy-docs.yaml"
   build-and-deploy-docs:
--- a/.github/workflows/build-and-deploy-docs.yaml
+++ b/.github/workflows/build-and-deploy-docs.yaml
@@ -14,7 +14,7 @@ jobs:
   delete-cancelled-runs:
     runs-on: "ubuntu-latest"
     steps:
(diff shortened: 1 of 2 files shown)


Theodlz:
Btw this makes me wonder if we want to do the same for all the unit tests (auto cancel + maybe delete) within a single PR (and not on main).

Cancelling a PRs actions that are still running once we merge it (which happens with the dependabot stuff sometimes) would be great too.

.github/workflows/build-and-deploy-docs.yaml
```diff
@@ -6,7 +6,17 @@ on:
...
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
+  delete-cancelled-runs:
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: "MercuryTechnologies/delete-cancelled-runs@v1.0.0"
```
stefanv, line 17: Better to pin to the hash here. See https://scientific-python.org/specs/spec-0008/#pin-github-actions-release-workflows-to-their-full-release-commit-shas

stefanv (approved):
Approved, pending pinning to hashed action.

.github/workflows/build-and-deploy-docs.yaml
Theodlz: pinning any version failed, so yeah I'll try the hash anyway hahaha

Theodlz: @stefanv I suppose that we don't actually need a separate job like in their docs, but that the step could be the first step of our existing build&deploy docs job?

stefanv: Yes, I think that's fine.

PR #5251, opened 2024-09-05
pin fiona to 1.9.6 to fix geopandas

Theodlz:
the fiona package (that geopandas uses) released a new version that broke geopandas. Until geopandas makes a new release, we can simply pin fiona its previous version.

This impacts features like the earthquake stuff, which breaks make load_demo_data and the migration CI

### DIFF
--- a/.github/workflows/test_migrations.yaml
+++ b/.github/workflows/test_migrations.yaml
@@ -135,6 +135,7 @@ jobs:
           make dependencies
 
           pip install black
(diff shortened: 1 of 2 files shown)


Theodlz:
hmmm so the issue here is that the migration test runs on an older branch first, which of course causes that failure to keep happening. Maybe I should update the git hash of the down migration we run on for this github action.

PR #5258, opened 2024-09-10
Reminders: API & Frontend tests: less flaky

Theodlz:
This is an attempt at making some tests a little less flaky (if not, not flaky at all) by specifying timezones, and being a little more smart about the checks that we run to not be too sensible to the timing of things.


mcoughlin (approved):
Ah finally a fix for reminders!

Theodlz:
God damm CI

Theodlz:
A feature I definitely regret adding 😅

PR #5259, opened 2024-09-10
more careful source patching

mcoughlin:
This PR fixes issue if only patching RA or Dec.


Theodlz (approved):
LGTM. Can merge once green

PR #5260, opened 2024-09-10
Swift sources origin - no BAT/XRT

Theodlz:
Turns out we didn't really need that, so let's keep it simple :)


Theodlz:
self merging

PR #5261, opened 2024-09-11
TRT Facility API: various improvements

Theodlz:
TRT joins the club of APIs that give us a 200 when it actually fails, so this PR accounts for some of those failures, updates to SQL alchemy 2.0 syntax, adds more error handling + frontend notifications, and moves to the `run_async` decorator to call the `download_observations` method.

The reviewer should keep in mind that I do not have access to the machine where this is running, so hard to test until deployed. However, the current API class if faulty anyway, so it doesn't hurt to test that in prod (on ICARE).

### DIFF
--- a/skyportal/facility_apis/trt.py
+++ b/skyportal/facility_apis/trt.py
@@ -1,11 +1,8 @@
 import base64
-import functools
 import json
 import os
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM. Agree on testing in production.

PR #5264, opened 2024-09-16
Default Observation Plans: auto send - plan properties filter

Theodlz:
This PR addresses a couple of issues such as:
- associating that localization of the plan to the correct GCN notice
- looking at an incorrect key when filtering on gcn notices

Also, it let's the user - optionally - set a list of plan_properties, that the plan generated by gwemopt needs to meet before we can send a plan to the scheduler (when auto send is turned on).


mcoughlin (approved):
LGTM

PR #5269, opened 2024-09-18
WIP - API Docs V2: redoc -> scalar + better docstrings

Theodlz:
Work in progress.

Essentially this moves away from redoc in favor of a much nicer tool (scalar), and also adds a few things such as a "summary" key to the docstrings of all the handlers, so that a readable text shows up in the list of endpoints rater than an obscure endpoint regex.

The reviewer can just run `make api-docs` to generate it locally and then can open `openapi.html` to see them


Theodlz:
@stefanv I've deactivated the "try it" feature for now, until we fix the CORS. It's a nice feature but it's not critical, so I'd suggest doing that in a future PR.

tools/docs/patch-template.py
```diff
@@ -0,0 +1,39 @@
...
+with open('openapi.json') as f:
+    openapi_spec = json.load(f)
+
+# here, we could add a list of servers to the openapi_spec
+# for the scalar docs to let the users point to when testing an endpoint
+servers = cfg.get("docs.servers", [])
+# validate that servers have only 2 keys: url and description
+if not isinstance(servers, list) and servers is not None:
+    raise ValueError("API servers must be a list.")
+if servers is not None:
```
stefanv, line 20: Two checks that servers is not none; maybe handle that a different way on line 16

tools/docs/patch-template.py
```diff
@@ -0,0 +1,39 @@
...
+    for server in servers:
+        if not all(k in server for k in ("url", "description")):
+            raise ValueError("Each server must have 'url' and 'description' keys.")
```
stefanv, line 23:
```suggestion
    if not all(key in server for key in ("url", "desc") for server in servers)
        raise ValueError("Each server must have 'url' and 'description' keys.")
```

Makefile
```diff
@@ -36,13 +36,9 @@ doc_reqs:
 
 api-docs: FLAGS := $(if $(FLAGS),$(FLAGS),--config=config.yaml)
 api-docs: | doc_reqs
-	@PYTHONPATH=. python tools/openapi/build-spec.py $(FLAGS)
-	npx @redocly/cli build-docs openapi.json \
-          --title "SkyPortal API docs" \
-          --theme.openapi.logo.gutter 2rem
+	@PYTHONPATH=. python tools/docs/build-spec.py $(FLAGS)
+	@PYTHONPATH=. python tools/docs/patch-template.py $(FLAGS)
```
stefanv, line 40: Perhaps a more descriptive name? Something like `patch-api-doc-template.py`. And we could rename the other one to `build-api-spec.py`.

stefanv:
image

Some font size weirdness :/

stefanv (approved):

mcoughlin (approved):
Dis coooolll

tools/docs/patch-template.py
```diff
@@ -0,0 +1,39 @@
+import json
```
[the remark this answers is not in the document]
Theodlz: done, made it a function and added a docstring

tools/docs/patch-template.py
Theodlz: done

Makefile
Theodlz: done

Theodlz:
addressed all of your comments I believe @stefanv

Theodlz:
> image
> 
> Some font size weirdness :/ (Minor.)
> 
> Otherwise, looks great!

yeah... I'll open an issue on scalar's repo. I opened 2 yesterday and they responded super quickly (they are super active, it's awesome and very reassuring).

stefanv (approved):

stefanv:
Good stuff, thanks Theo!

PR #5270, opened 2024-09-18
Clean up API multi-page fetch example

stefanv:


Theodlz:
@stefanv should we consider merging this? Happy to review it if it can be moved out of draft mode

stefanv:
I had it in draft because I haven't tried it yet; but if the logic looks good to you!

Theodlz:
> I had it in draft because I haven't tried it yet; but if the logic looks good to you!

I think it looks great @stefanv, I especially like that retry. Just 3 minor comments:
- might want to just build the `headers` once at the beginning and then pass it to `requests.request()` in the loop, rather than add extra computational time to recreate that dict every time. It's so small it's negligible though, so can leave it as is.
- feels a little odd to me to have that `base_url` variable contain the full endpoint. Usually if I have a base_url it's just the protocol+host+port+api prefix, and then I might have a `url`/`endpoint` variable where I concat that with the endpoint in question.
- might want some error handling on the existence of totalMatches. I can see users just copy paste this blindly for every endpoint, and just knowing that the endpoint doesn't seem to return a totalMatches would be good, rather than throw that missing key error

Theodlz:
ah and also, the `page` and `num_per_page` in the `params` should be: `pageNumber` and `numPerPage` instead!

Theodlz:
This has already been addressed by https://github.com/skyportal/skyportal/pull/5304, closing!

stefanv:
I don't follow how my commits from here made it over there, but OK :upside_down_face:

PR #5273, opened 2024-09-19
OpenAPI spec - security/auth + ACLs

Theodlz:
This PR requires https://github.com/cesium-ml/baselayer/pull/387 to be merged first, and once that done the new version of baselayer should be pinned on this PR before merging

Essentially it:
- adds the authentication/security (that we already specified in the `components` of the openapi spec) to ALL endpoints by default
- based on what is added in that baselayer PR, is able to figure which endpoints do not use authentication, and makes an exception for those. That way the autogenerated API docs show headers with a token when needed, otherwise do not.
- fixes the endpoint displayed for a few handlers' methods that had unused arguments (often happened on post methods that had the resource_id just to comply with the endpoint's regex, but didn't need them) to avoid confusion for the end user.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -914,7 +914,7 @@ class GcnEventAliasesHandler(BaseHandler):
 
 class GcnEventTagsHandler(BaseHandler):
     @auth_or_token
-    async def get(self, dateobs=None, tag=None):
+    async def get(self, *ignored_args):
         """
         ---
         summary: Get all GCN Event tags
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -133,9 +133,9 @@ todo_include_todos = False
 # Per https://github.com/snide/sphinx_rtd_theme: specify theme if not on RTD
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 if not on_rtd:  # only import and set the theme if we're building docs locally
-    import furo  # noqa
+    import sphinx_book_theme  # noqa
 
-    html_theme = 'furo'
+    html_theme = 'sphinx_book_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/doc/openapi.html.template
+++ b/doc/openapi.html.template
@@ -1,14 +1,13 @@
 <!doctype html>
 <html>
   <head>
-    <title>Scalar API Reference</title>
+    <title>SkyPortal API Reference</title>
     <meta charset="utf-8" />
     <meta
       name="viewport"
       content="width=device-width, initial-scale=1" />
   </head>
   <body>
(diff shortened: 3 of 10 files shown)


Theodlz:
@stefanv I just pinned baselayer so let's wait till the CI is green (though it should be, as it was before already) before reviewing, but this should be ready to be merged soon enough.

skyportal/handlers/api/gcn.py
```diff
@@ -914,7 +914,7 @@ def delete(self, dateobs):
 
 class GcnEventTagsHandler(BaseHandler):
     @auth_or_token
-    async def get(self, dateobs=None, tag=None):
+    async def get(self, *ignored_args):
```
stefanv, line 917: This changes those args from keyword to positional?

stefanv (approved):
LGTM, just the one small question.

skyportal/handlers/api/gcn.py
Theodlz: so ew have a couple of endpoints where the GET, POST, DELETE take a different number of arguments (for example the GET, DELETE might take an id that the POST does not). To avoid duplicating the endpoint in the `app_server.py` the easiest way to fix this is to actually add the unused parameter to the handler, with a None value. However the issue with that is that these the documentation would still show that unused argument in the endpoint URL that it generates, as if it was required or smth.

Here all that I did is look for handlers with unused arguments, make those `*args` with a specific name (here, ignored_args) so that in `openapi.py` I ignore those specifically, and can end up with the correct & more concise endpoints).

Does that make sense? 

Oh and actually opening another small PR on baselayer to help display the permissions required to use an endpoint (if any), which should help remove a lot of the confusion our users have at the moment!

stefanv: Totally fine, was simply wondering whether you wanted `**ignored_args` instead?

Theodlz: ah well that way I had a single argument name I can look for an ignore when autogenerating the openapi specs, instead of having to maintain a list of args. But I mean I could simply rename all of those to idk, something like `unused_arg` and avoid those in a loop, but *args seemed appropriate as it automatically throws all of the unused arguments in a list that I can ignore all at once.

Theodlz:
Here's the ACL display additional btw @stefanv. It just reads that list, and adds it to the `description` of the endpoint, while formatting it with some very minimal HTML:

image

skyportal/handlers/api/gcn.py
stefanv: I'll ping you on Slack; I am really only wondering about the argument vs keyword distinction.

stefanv:
Nice, I like that a lot!

Theodlz:
@stefanv looks like tornado just deals with *args in the first place, then **kwargs. Makes sense, since the regex just spits out an ordered list of values and not a dictionary or similar structure with keys for each values.

So, this approach seems to be valid?

stefanv (approved):

stefanv:
Thanks for verifying, Theo. In it goes!

PR #5274, opened 2024-09-19
Python dependencies: loose version requirement for common packages

Theodlz:
@stefanv mentioned that it wasn't possible to run SkyPortal on Python 3.12 because of the strict requirement we had on a numpy version.

In general, pinning exact versions has proved to make deployment and the CI much nicer and more reliable for us, but it is true that it makes it a little restrictive when trying to just run SkyPortal on whatever environment one has already.

In this PR, I tried to be a bit more flexible version-wise, for a couple of very packages python packages that I expect people to already have installed, or that might have some Python-version requirement.

If the reviewer has python 3.12 on their system, it could be a good idea to give that a go and see if they can run SkyPortal without downgrading to 3.11

PS: I still specific some maximum major version for these packages, just in an effort to minimize our chances of SkyPortal just breaking when one of these dependencies makes a lot of changes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,31 +1,31 @@
 astroscrappy==1.2.0
-numpy==1.23.4
-scipy==1.13.1
-pandas==2.2.0
-dask==2024.5.2
+numpy>=1.23.4,<2.0.0
+scipy>=1.13.1, <2.0.0
+pandas>=2.2.0, <3.0.0
+dask>=2024.5.2
 geopandas==1.0.1
 fiona==1.9.6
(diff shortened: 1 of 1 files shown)


mcoughlin:
@Theodlz how will this play with dependabot? does the upper bound just grow then?

Theodlz:
> @Theodlz how will this play with dependabot? does the upper bound just grow then?

That's the idea. The end of the month is very soon, so it would be a good time to merge this now and try it out.

mcoughlin (approved):
LGTM!

PR #5277, opened 2024-09-28
logging for source creation

mcoughlin:
This PR adds more logging for source creation from GCN sources.


Theodlz (commented):
Might want to log the dateobs with the localization name? These are often the same, but would make it easier to track stuff if we had the event's datetime

mcoughlin:
Great idea @Theodlz, implemented.

Theodlz (approved):
Looking good! Merge once green. I'm back to work tomorrow so I'm hoping to redeploy with all this no later than Wednesday.

PR #5278, opened 2024-09-30
webpack -> rspack migration

Theodlz:
Bundling times for SkyPortal (in production mode for example) are cut in half using rspack, and it looks like the bundle sizes are also slightly reduced in production mode.

Overall rspack seems to offer us the same features as webpack, while speeding up things which is very handy while developing, and while running any CI/CD.

I'd be in favor of moving to rspack, especially given how minimal the changes needed are.

This PR depends on a baselayer PR: https://github.com/cesium-ml/baselayer/pull/388


mcoughlin:
@Theodlz agreed, I liked my experience with it as well when working on that initial PR. thanks for getting this done.

Theodlz:
good progress on the baselayer + template app side of thing, once these get merged this one can pin main.

Theodlz:
also have to wait for the dependabot stuff to calm down before the CI runs here

Theodlz:
@mcoughlin baselayer is now running rspack in place of baselayer, so once the tests are green this is mergeable.

mcoughlin (approved):
LGTM

PR #5299, opened 2024-10-01
PhotometryTable delete confirmation + @rjsf update

Theodlz:
This PR does 2 things:
- add a delete confirmation dialog when one attempts to delete a photometry point from the table we show on the source page
- update all the `@rjsf` (react-json-schema-form) dependencies to have all the versions coincide, and to - finally - have their bugfix that caused validation to freak out when lists were a little too long on Firefox (that prevented users of that browser to use a couple of features).


mcoughlin (approved):
LGTM

PR #5300, opened 2024-10-01
WIP - PR Labels: frontend-only

Theodlz:
The idea is to label PRs that only edit the frontend, to avoid running all the backend CI that wouldn't be affected by a pure frontend change anyway. This would help speed up these PRs, but also speed-up how quickly we run all of these CI jobs when dependabot does its monthly thing.

This PR is not complete yet, I'm adding a few things incrementally, but can't make it a draft otherwise things won't run


Theodlz:
closing for now

PR #5301, opened 2024-10-02
Gemini Facility API

Theodlz:
Adds the building blocks needed for us to trigger Gemini from SkyPortal, along with some metadata to test it.


Theodlz:
@mcoughlin having the form show the list of available template IDs isn't really an option except if I start sending the content of altdata back to the frontend which is not great. It's not impossible to make it work but requires more frontend and backend changes.

I already added a few things here, and let's keep in mind that just showing the template IDs isn't useful either, as the user need to know about the content of the templates to choose them. For example, you pick the ToO mode (regular or rapid) and the priority (low, medium, high) in these templates.

So, just because of how this is all designed, we anyway need the user to have access to the OT, where they get that info. Therefore having a text-like input is acceptable.

I added a few extra facility API-related things to this PR (like altdata validation on the server side when adding an allocation), and one can specify a list of template_ids in the altdata so that on submission the server can validate that the template_id from the request is in that list, and if not shows an error message specifying the list of options.

mcoughlin:
Thank @Theodlz !

PR #5302, opened 2024-10-05
sort source ids when doing pagination

mcoughlin:
This PR sorts the source ids when doing pagination, which matters when sources are being added to the app during the request.


Theodlz:
@mcoughlin I will close this one in favor of a PR I'll open today that adds a proper order by in the query for saveSummary, and implements caching for both normal & saveSummary queries

PR #5304, opened 2024-10-07
get_sources caching

Theodlz:
Users have been complaining about the consistency of sources (notice the S, it's not an issue when querying just one) queries with the API. This happens because these lists are dynamic, and can see items being added and/deleted while the user paginates through the results.

This becomes an issue because even if we implement pagination for this endpoint, we had no functional caching so far meaning that the query would rerun everytime you request a new page, meaning that if things are added/deleted at some random spot in the list, the results would be shifted as you paginate. That means potential duplication, missing some objects, etc.

This PR implements caching for that endpoint. That way, the database query runs ONLY ONCE when requesting the first page, and then the user can paginate with a statically defined list by simply specifying the queryID the server provided in the parameters sent to the server thereafter.

When using the saveSummary=true version of the endpoint (that directly pulls from the source table and not the obj table) we do not enforce any sorting (since that table doesn't have that field). Here, to at least make sure that the order is consistent, we sort the results alphabetically.

With these 2 changes, I can successfully retrieve a static list of sources by specifying useCache=true & the queryID (for pages > 1) and loop through the results without any duplication issues which is what was reported to us. Things look consistent too.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -523,6 +523,7 @@ services:
 misc:
   days_to_keep_unsaved_candidates: 7
   minutes_to_keep_candidate_query_cache: 60
+  minutes_to_keep_source_query_cache: 360
   minutes_to_keep_annotations_info_query_cache: 360
   minutes_to_keep_localization_instrument_query_cache: 1440
   max_items_in_localization_instrument_query_cache: 100
--- a/skyportal/api_description.md
+++ b/skyportal/api_description.md
@@ -118,45 +118,58 @@ for the failure given in `message`:
 Several API endpoints (notably the sources and candidates APIs) enforce
 pagination to limit the number of records that can be fetched per request.
 These APIs expose parameters that facilitate pagination (see the various
-API docs for details). A sample pagination script is included here:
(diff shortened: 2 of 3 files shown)


Theodlz:
Btw one thing I forgot to mention (and subject to change depending on how things to in production) is that I didn't use a randomly generated uuid as the caching IDs here, but rather a hash of the list that is cached. That way if user A queries the API with caching and user B does the same (and his query returns the same list of objects, in the same order), we don't need to cache things twice and we avoid using too much space on disk.

Only potential issue is reusing the same cache again and again, until we get too close to the cache's invalidation date and we break the consistency (though this is also true in general if looping over the results takes longer than we cache things on disk).

However since when user B queries the first page (i.e. they don't have a queryID yet) we fetch the results and write the cache to disk (to the same filename as user A's identical cache/results), that should reset the caching duration every time another user (or the same but that restarts his loop) queries sources to get the same results.

In a future PR it would be awesome if we could do that cache duration update every time we loop over a page, so that we are 100% sure that as long as someone queries sources in a loop, until they reach the last page the cache is guaranteed to be there and not invalidated, which in term guarantees consistency completely.

I mean this is the cherry on top, and quite frankly if someone's script takes more than an hour to loop over sources, maybe it should not run in the first place. But for any reasonable application, we are all good now.

PR #5305, opened 2024-10-07
ESLint fix

Theodlz:
It's very poorly documented but when specifying a path to eslint (say, what's in static/js, as added recently on main's recommit config) in the pre-commit config, it stops running it **only** on newly committed files, and instead keeps running it on everything in that folder, instead of the behavior I hoped it would have (which is running with just the same logic, but limited to the specified directory).

This PR fixes that, in an attempt to fix the CI. Also fixed a few jsx files.


mcoughlin (approved):
LGTM!

PR #5307, opened 2024-10-08
tdtax: update from 0.1.4 to 0.1.7

Theodlz:
This PR:

- pins the latest version of tdtax in the requirements.txt file
- in the POST taxonomy method, use the sitewide group if no groups are specified, instead of all the user groups (which already includes sitewide
- fixes the `isLatest` update of outdated taxonomies when uploading a new one, which did not work if more than one version of the updated taxonomy already existed.


profjsb (approved):
LGTM

PR #5311, opened 2024-10-22
Weather: API migration and frontend cache invalidation

Theodlz:
In this PR:
- invalidate the weather data that's in redux if it's older than 1 hour, which is also the default caching time we use in the config (we don't query openweathermap's API more than once every hour).
- migrate openweathermap's `onecall` API from 2.5 (discontinued) to 3.0

Note: Instances that use that feature might want to go to their openweathermap account, and subscribe to the onecall API service. It will ask for credit card info but we get 1000 free API calls a day, and a limit can be added on top of that to make sure the free tier becomes not so free anymore. Also since SkyPortal caches every hour (and per telescope), you would need people to query weather info for >41 distinct facilities (lon/lat coordinates) to get to the paid tier, which is VERY unlikely.

### DIFF
--- a/skyportal/handlers/api/weather.py
+++ b/skyportal/handlers/api/weather.py
@@ -113,7 +113,7 @@ class WeatherHandler(BaseHandler):
             message = ""
             if refresh:
                 response = get_url(
-                    "https://api.openweathermap.org/data/2.5/onecall?"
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #5314, opened 2024-10-29
WIP - migrate JS/Python package managers: npm -> bun, pip -> uv

Theodlz:
Work in progress, do not review.


Theodlz:
@stefanv any idea why the Docker build CI is not running here???

Theodlz:
Aside from the docker build I would really like to run when we edit the Dockerfile and I'm weirded out that it's not working, uv + bun is a great combo. I will keep adding documentation here and in baselayer before we even consider merging, but these seem to be a super nice options to run SkyPortal

stefanv:
It may be that, for CI, pixie is even faster than uv. We could explore generating our dependencies list from a pixie specification, and then having baselayer / our users use `uv`.

stefanv:
> @stefanv any idea why the Docker build CI is not running here???

```
on:
  push:
    branches:
      - main
      - test-docker
  pull_request:
    types: [opened, reopened, synchronize, ready_for_review, labeled]
```

I would expect this to trigger only on commits to `main` or `test-docker` branches.

Theodlz:
Thanks @stefanv! Should we consider first moving to uv + bun and if that seems to work well then integrate that with pixie?

Theodlz:
If feel like the CI specific stuff could be its own PR

mcoughlin:
Superseded by https://github.com/skyportal/skyportal/pull/6191

PR #5316, opened 2024-10-29
Frontend Photometry Single upload - date fix

Theodlz:
This PR:
- fixes the date to mjd conversion
- is more strict on the date format
- doesn't get rid of all the forms data when changing the instrument that is selected.


mcoughlin (approved):
LGTM

PR #5317, opened 2024-10-29
Follow-up requests constraints: not_if_tns_reported

Theodlz:
Add an optional `not_if_tns_reported` constraints that can be passed to the API when posting follow-up requests. This parameter is a float/int (in hours). If a source within the specified `radius` parameter (which defaults to 0.5 arcsec) has been reported on TNS more than `not_if_tns_reported` hours before the current time, we can avoid triggering.

Depends on: https://github.com/skyportal/kowalski/pull/444


mcoughlin (approved):
LGTM

PR #5338, opened 2024-11-02
log full traceback when posting autogenerated GCN sources fails

Theodlz:
This PR logs the full traceback when posting autogenerated GCN sources fails

@mcoughlin that's for Icare. The current log message doesn't show us where the error happens, so it's not sufficient for debugging.


mcoughlin (approved):
LGTM

PR #5339, opened 2024-11-03
Gwemopt - del unused skymap data

Theodlz:
Some keys from the `map_struct` we build with gwemopt's `read_skymap()` method are only necessary for plotting or running some efficiency calculations which we do not use. These use quite a bit of memory and can cause OOM issues. 

In this PR, we simply remove these unused variables to avoid OOM issues that are currently killing the observation plan python process.


mcoughlin (approved):
LGTM

PR #5344, opened 2024-11-07
Photometry form frontend test: fix

Theodlz:
Fixes a test broken in a previous PR that edited the form's layout and parameter names.


Theodlz:
self merging

PR #5345, opened 2024-11-07
WINTER API: various improvements

Theodlz:
This PR:

- catches errors appropriately rather than returning the request's URL which contains credentials of the allocation.
- fixes the re-submission of failed requests


mcoughlin (approved):
LGTM

PR #5349, opened 2024-11-10
next sunrise to next twilight

### LINKED ISSUE #5348: Default end time for MMA plans set to 12 degree twilight
**Feature Summary**
I am not sure exactly what the current definition is for default end date for the MMA observation plans (see attachment), but it seems to be set to exactly sunrise. It would be great if the default was something more reasonable, e.g 12 degree twilight, when you can still actually observe. Right now, generating a plan using the defaults will include extra fields that can never be observed.  

**Usage / behavior**
The default should be 12 degree twilight, so the corresponding plans are shorter and can then be fully executed.

**Alternative Solutions**
None come to mind

**Implementation details**
Set at 12 degree by default. could give people the option to change to e.g 6 degree by tickbox, or just leave 12 for everyone. I don't have a strong opinion.

**Additional context**
Screenshot 2024-11-09 at 08 28 27

mcoughlin:
This PR changes the default end of an observation plan from sunrise to 12 degree twilight. 

Closes #5348.

### DIFF
--- a/skyportal/facility_apis/observation_plan.py
+++ b/skyportal/facility_apis/observation_plan.py
@@ -439,7 +439,7 @@ class MMAAPI(FollowUpAPI):
             galaxy_catalogs = [
                 g for g, in DBSession().query(GalaxyCatalog.name).distinct().all()
             ]
-        end_date = instrument.telescope.next_sunrise()
+        end_date = instrument.telescope.next_twilight_morning_nautical()
         if end_date is None:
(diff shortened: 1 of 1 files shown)


Theodlz (approved):
LGTM 🚀. Should the start change too, or the end time is enough?

mcoughlin:
@Theodlz We start from the current time at the moment, which is fine during the night, and gwemopt should probably be setup to do something smarter in case its the day, but thats probably a gwemopt problem.

PR #5350, opened 2024-11-11
Analysis Services Page - details mixup bugfix

Theodlz:
This PR:
- fixes an issue on the analysis services page where the popup dialog that shows the details of a service aren't showing the right ones
- fixes the indent of the docstring of a few endpoints (POST & PATCH of the photometric series, GET of the GCN events), which prevented from them appearing on the documentation.
- Hides group users that can be added to a shift, if they are expired
- Show expired users again in the user management page, because otherwise it is impossible to extend their account's expiration date from the frontend.
- document the includeExpired flag of the users endpoint
- fixes the download of executed observations on the GCN page
- adds the `ports` section in the default config, present in baselayer but missing here in Fritz. Both need to be edited to change the port on which the app is running so the fact that it is missing makes it very hard to figure out for folks

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -522,6 +522,8 @@ server:
     google_oauth2_secret:
     google_oauth2_timeout: 15
 
+ports:
+  app: 5000 # This is the internal port
 services:
   dask: False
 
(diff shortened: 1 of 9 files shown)


mcoughlin (approved):
LGTM

mcoughlin:
@Theodlz yes after you pointed out on my PR ;) looks great

PR #5351, opened 2024-11-12
Return non-expired users for shifts

mcoughlin:
This PR ensures we only return non-expired users for shifts.


Theodlz:
Hmmm did you see my other PR? I should have added that already

Theodlz:
And this causes issue if there are users in existing shift that expire. In my PR that is accounted for

PR #5352, opened 2024-11-13
NGPS starlist: backend & frontend

Theodlz:
This PR is a first attempt at supporting the starlist format of the P200 for NGPS, which differs from other instruments that are on it.

One thing that isn't clear so far is how their new format figures out what's not a source but a standard or offset, since I haven't been made aware of any comments/columns that I should add for that. 

So opening as a WIP and once the NGPS folks have that figured out and communicated with me, I'll add that in before merging. That said the rest is ready.

It's a bit different from the other formats that all have the same columns but just different formatting, this one has its own specific columns so we just make an exception and enforce the appropriate schema/format.

Since it's more of a CSV, I have the frontend add the columns on top of the star list. Technically the backend can do that (with an argument called first_line that a starlist config can have), except that this would add it to every source (since we query them one by one on the observing run page), and the frontend would have to have extra code to deduplicate anyways, so might as well keep it simple and add the columns there, since this is something that folks need to see on the frontend specifically.


skyportal/utils/offset.py
```diff
@@ -228,6 +234,15 @@ def format_hmsdms(skycoord, coord_sep, col_sep):
...
         "first_line": None,
     },
+    'P200-NGPS': {
+        "coord_sep": ' ',
+        "col_sep": 'XXX',
+        "commentstr": "!",
+        "giveoffsets": False,
+        "maxname_size": 18,
+        "first_line": None,
+        "format": "csv",
```
mcoughlin, line 244: Does this line end up doing anything? Most of the PR is IF statements. Not a problem, just curious.

mcoughlin (approved):
Looks reasonable. Given the format, maybe a little bit different than the others as you said leading to some more custom code than might seem ideal.

skyportal/utils/offset.py
Theodlz: Oh no you are right this is an artifact of my attempt at making it work with the current config-like system and failing. I'll remove it. Thanks!

Theodlz:
I made a number of changes since the last review (as I got some more detailed instructions recently), so asking for another review. I suggest simply picking the "changes since last review" when taking a look

skyportal/utils/offset.py
```diff
@@ -50,6 +50,21 @@
     f':{cfg["server.port"]}' if cfg['server.port'] not in [80, 443] else ''
 )
 
+NGPS_TARGET_BANDS_TO_SNCOSMO = {
+    'G': ['ztfg', 'sdssg', 'lsstg'],
+    'R': ['ztfr', 'sdssr', 'bessellr', 'standard::r', 'lsstr'],
+    'I': ['ztfi', 'sdssi', 'besselli', 'standard::i', 'lssti'],
+    'U': ['sdssu', 'bessellux', 'standard::u', 'lsstu'],
+}
+SNCOSMO_BANDS_TO_NGPS_TARGET = {}
```
mcoughlin, line 59: add a comment that you are inverting this dictionary?

skyportal/utils/offset.py
```diff
@@ -797,6 +826,31 @@ def get_formatted_standards_list(
 
     if return_dataframe:
         return df
+    elif starlist_type == "P200-NGPS":
```
mcoughlin, line 829: a bit of a mess, but maybe hard to avoid.

skyportal/utils/offset.py
```diff
@@ -1133,15 +1203,46 @@ def get_nearby_offset_stars(
...
+            if magfilter is None:
+                raise ValueError(
+                    f"Cannot find corresponding NGPS filter for sncosmo filter {magfilter}"
+                )
+            try:
+                mag = round(float(source_mag), 2)
+            except ValueError:
+                raise ValueError(f"Cannot convert magnitude {source_mag} to float")
+
+            # check that there is a corresponding value in NGPS_TARGET_BANDS_TO_SNCOSMO
```
mcoughlin, line 1221: what is this comment for?

mcoughlin (approved):
Looks good

skyportal/utils/offset.py
Theodlz: I think that the whole offset stuff in general is a mess, which causes something like this to happen. Everything was designed to duplicate / generalize but only for a specific ascii like format where all facilities essentially have the same columns. I'd be willing to redo the whole thing at some point, but for now the P200 folks just need this to be up and running for their observing runs

skyportal/utils/offset.py
Theodlz: done

skyportal/utils/offset.py
Theodlz: removed

PR #5354, opened 2024-11-15
Docker build action rules - fix

Theodlz:
An attempt at fixing the docker build action which is often not running when it should on PRs.


Theodlz:
yeahhhhh it is running on this one

PR #5355, opened 2024-11-17
arcs only for soar calibration

mcoughlin:
This PR updates the SOAR configuration to be arcs only (this is in practice the desire).


Theodlz (approved):
LGTM

PR #5357, opened 2024-11-20
Observation plan auto trigger - end date check bugfix

Theodlz:
Looks like `arrow` creates a timezone aware date in new versions, which break this feature when comparing it to the vanilla `utcnow()` generated `datetime`. This PR is a quick fix, but overall we should migrate to the non `datetime.utcnow()` syntax that newer versions of python are deprecating.

Honestly it's such a stupid choice, everyone uses it and this syntax was perfectly intuitive...


mcoughlin (approved):
LGTM

PR #5358, opened 2024-11-21
Swift triggering - exposure time clarification

Theodlz:
It's not super clear from the follow-up request form on the frontend if the exposure time is a total, or per visit. the "number of visits" field is called "exposure counts" so that does help but it's still not intuitive.

This PR renames these fields to "Exposure time per visit" and "Number of visits" to make it as transparent as possible.


mcoughlin (approved):
LGTM

Theodlz:
DNM

Theodlz:
My office mate spotted when triggering that it didn't work, looks like the Swift API now requires the uvot_just for all modes but 0x9999, and not just ToO mode as we had it set up for

Theodlz:
fixing that rn

PR #5359, opened 2024-11-25
GCN Notice contents download + NGSF service bugfix

Theodlz:
This PR solves 2 problems:

**GCN Event Notices - heavy content:**

Today we insgested an event from BAT Guano, coincident with a BBH event from ligo. Issue is that the Guano events don't give you a link to the skymap, but instead put the skymap in the GCN event notice directly. This is not the best but not much a problem by itself, except that we return the content of GCN notices along with the events, meaning that it now takes multiple megabytes to load an event, causing timeouts and other issues. This PR adds an optional `excludeNoticeContent` argument to the GET endpoint of the GcnEventHandler. It defaults to false (still included, not changing the default behavior of the backend) and set to true by the frontend. To still allow notice contents to be downloaded but this time as proper files, we add a `GcnEventNoticeDownloadHandler` that makes it possible, and edit the frontend to use it.

**NGSF analysis service:**

NGSF uses an older version of numpy. Everything is compatible with new versions (like the one we use in Fritz), except one `np.float()` call (which now doesn't exist anymore) that needs to be replaced by a regular `float()`. This PR simply patches that after git cloning NGSF, but with less urgency we can just contribute to the NGSF repo.


mcoughlin (approved):
LGTM

PR #5360, opened 2024-11-26
Follow-up requests: allocation-wide notifications (by slack & email)

Theodlz:
This PR:

- refactor Generic API to move the slack & email notification logic into the notification utils, to be important by other facility APIs
- adds optional slack OR email notifications when submitting/updating follow-up requests (for winter, sedm, swift).
- Upgrade the format of Slack and email notifications, to look nicer and be more functional
- change Gemini auth parameter names
- move radec2lb from sources to utils

The idea here is that having every single user that has an allocation set up notifications for it is cumbersome, and often its the same group of user that has access to an allocation that could just easily get notifications sent to a slack channel for everyone to see, or to an email list. This PR enables that.

TODOs (in future PRs):
- when I'll redesign the notification framework (hopefully before the end of the year, otherwise beginning of the next), make use of these new methods to also send nice looking notifications (with this new format added by that PR) with the regular notification (user-specific) system.
- update the follow-up request documentation which is lagging behind the backend features overall, and mention the notification system there too. I'm planning to address that after Thanksgiving, and make a new release then.

### DIFF
--- a/skyportal/facility_apis/sedm.py
+++ b/skyportal/facility_apis/sedm.py
@@ -102,6 +102,8 @@ def convert_request_to_sedm(request, method_value='new'):
     }
 
     if rtype == "Mix 'n Match":
+        # we make a deepcopy so that the payload's observation_choices
+        # aren't edited when we edit choices
         choices = deepcopy(request.payload['observation_choices'])
         hasspec = 'IFU' in choices
         followup = 'IFU' if hasspec else ''
--- a/skyportal/facility_apis/sedm.py
+++ b/skyportal/facility_apis/sedm.py
@@ -1,6 +1,8 @@
 from datetime import datetime, timedelta
 import json
 import requests
+import traceback
+from copy import deepcopy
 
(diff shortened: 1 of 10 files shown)


skyportal/facility_apis/sedm.py
```diff
@@ -100,7 +102,7 @@ def convert_request_to_sedm(request, method_value='new'):
     }
 
     if rtype == "Mix 'n Match":
-        choices = request.payload['observation_choices']
+        choices = deepcopy(request.payload['observation_choices'])
```
mcoughlin, line 105: Maybe make a note why this was needed?

mcoughlin (approved):
Awesome!

skyportal/facility_apis/sedm.py
Theodlz: Good point, will do!

Yeah it's just that we remove the IFU from that variable few lines after, and since this editing the variable directly then the notification stuff doesn't have it anymore. Hence the deepcopy

PR #5362, opened 2024-11-27
Obsplan statistics - localization tiles partition selection

Theodlz:
In the `skyportal.utils.observation_plan.py`'s `generate_observation_plan_statistics()` method, we do not make use of the python-based selection of the database partition everywhere

The LocalizationTile table is partitioned, and we can either just query the table and let the DB select the partition, or bypass that and query the partition directly which we do pretty much everywhere already.

In this PR, we make sure that the pre-selected partition is used for all the queries in that method, for consistency but also as it seems to prevent index-selection issues on a production system (Icare) where the postgres migration created issues.


mcoughlin (approved):
LGTM

PR #5402, opened 2025-01-01
TNS API changes

Theodlz:
In this PR we adopt some of the changes recently added to the TNS API, that broke most (if not all!) the scripts that interact with it.

We had to fix a few endpoints, and directly look for the responses' data in `data`, and not `data.reply`.

PS: I already made these changes in production, on Fritz. I already reprocessed all of the failed submissions so we're ok.


Theodlz:
and the CI is broken OF COURSE

Theodlz:
ok issue fixed, we well calling npm with `sudo` which made it use an outdated version of node & npm that isn't the one installed with the `setup-node` action. The `sudo` isn't necessary to begin with, so I removed it and it can work again

mcoughlin (approved):
Oof.

PR #5404, opened 2025-01-07
check for spectrum units for synthetic photometry

mcoughlin:
This PR adds check for spectrum units for synthetic photometry.


Theodlz (approved):
LGTM

PR #5407, opened 2025-01-14
LT request - start and end date format

Theodlz:
I've been told that LT requires a T before the date and time (and not a space) when submitting requests. This PR adds the missing T. For now, informed users were adding it by hand.

Also, in this PR I reorganized some imports across the different files in `skyportal/facility`, to keep things a bit cleaner.


mcoughlin (approved):
LGTM

PR #5409, opened 2025-01-14
Refactor cache restoration into reusable workflow

stefanv:
Also move cache earlier, so that we don't overwrite npm installation


stefanv:
Closing to fix branch name.

PR #5410, opened 2025-01-14
Refactor caching into composite action

stefanv:


stefanv:
Yup, that was it :)

@Theodlz ready for review.

stefanv:
I think we can use this same pattern to do other things, e.g. build a "setup-skyportal" action.

Theodlz (approved):
LGTM, very nice @stefanv !!! Thanks a lot, the deduplication is very appreciated 🚀

PR #5412, opened 2025-01-15
Ruff

stefanv:
Note: this PR *must* be merged, *not* squash merged, to make `.git-blame-ignore-revs` work.


stefanv:
Migrations failure due to reliance on import order:

```
ALLOWED_API_CLASSNAMES = [
    k
    for k, v in facility_apis.__dict__.items()
    if inspect.isclass(v)
    and issubclass(v, facility_apis.FollowUpAPI)
    and v is not facility_apis.FollowUpAPI
]
```

Theodlz:
> Migrations failure due to reliance on import order, yikes:
> 
> ```python
> ALLOWED_API_CLASSNAMES = [
>     k
>     for k, v in facility_apis.__dict__.items()
>     if inspect.isclass(v)
>     and issubclass(v, facility_apis.FollowUpAPI)
>     and v is not facility_apis.FollowUpAPI
> ]
> ```
> 
> We can't sort these alphabetically, I think, since they have to be in order of addition to code?

It should have been an ordered dict maybe a while ago, but this is likely used to create a DB enum which now has an order so changing things might be an issue. Can we make an exception with certain files?

stefanv:
> It should have been an ordered dict maybe a while ago, but this is likely used to create a DB enum which now has an order so changing things might be an issue. Can we make an exception with certain files?

Yes, could definitely do that. For now, I opted to make the order explicit in the code, but could go the implicit route if preferred. Let's see if we can get the tests to pass :)

stefanv:
Let me know before merging, so that I can re-create the git-blame-ignores.

stefanv:
OK, I think this is ready to go. Please note again that this must *not* be squash merged.

Theodlz:
@stefanv when trying to do a regular merge, it says "not available for this repo". Might need to change some repo rules temporarily?

stefanv:
Try now?

Theodlz:
@stefanv done, did that work?

stefanv:
Yes, thanks Theo!

```
git blame --ignore-revs-file .git-blame-ignore-revs alembic/env.py
```

shows none of my changes. And [GH also supports this feature](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view).

PR #5415, opened 2025-01-17
Sleep services that repeatedly process queues

stefanv:
These two workers currently take up a lot of CPU, and they will flood the database with requests.

I wasn't sure what the longest appropriate wait is here; I'll defer to @mcoughlin.


stefanv:
The tests expect the reminder to be broadcast immediately, so would need to adjust that as well.

stefanv:
Good suggested approach. Thanks, @Theodlz, I'll close this PR.

PR #5416, opened 2025-01-17
Fix state update affecting all open tabs

antoine-le-calloch:
This PR resolves a bug where publishing a new public source triggered updates in all open tabs, even for sources that were unrelated.


stefanv (approved):
Nice catch.

Theodlz (changes requested):
Thanks Antoine! Could you just use the `?` operator to avoid yielding an error if a source isn't loaded into the state yet?

antoine-le-calloch:
> Thanks Antoine! Could you just use the `?` operator to avoid yielding an error if a source isn't loaded into the state yet?

Yes well seen.
Done ✅.

Theodlz (approved):
LGTM 🚀

PR #5418, opened 2025-01-20
Test config fix

jinglin-l:
There's a mismatch between how config files are loaded when running the app via Makefile vs. running tests directly using pytest ci. This can lead to issues like port assignment conflicts.

This is a PR to update the pytest configuration to always load test_config.yaml, when running tests. 

For (my own) future reference, this is the order that the config file hierarchy cascades (@Theodlz fact check this pls): 

1. baselayer/config.yaml.defaults (base configuration)
2. skyportal/config.yaml.defaults (skyportal-specific defaults)
3. config.yaml (local overrides)
4. test_config.yaml (when running tests)


Theodlz:
For context, the issue we have is that since running tests via pytest or other doesn't get the `--config=config.yaml` argument that we pass to everything in the Makefile, the tests only read the default configs and not the config.yaml or test_config.yaml.

What we expect is that just as running `make run_testing` ensures that you are using the `test_config.yaml`, running any tests would also always use that same config to stay in sync and avoid issues, namely port assignment issues.

Theodlz (approved):
LGTM! Will merge once green. Thanks a lot @jinglin-l

stefanv:
@Theodlz Did this work? I see that there's also a test config being loaded in fixtures. It feels like we should remove that duplication. Could be that the best place is in `conftest.py`.

It confused me to see that there's a `fixtures.py`, but fixtures are defined in `conftest.py` :thinking:

Theodlz:
@stefanv yes, it did work, as the config read in __init__.py is for the `api()` method that we call everywhere in the tests so we can have the host and port.

Basically this PR was written by @jinglin-l as we fixed it on her local installation while on a Zoom call. Before that edit, the API tests were trying to hit port 5000 (the default from the default configs), after that edit it correctly read the port assignment that she had added to her test_config.yaml

Theodlz:
I agree that the fixtures.py vs conftest.py thing is confusing...

stefanv:
Thanks @jinglin-l!

PR #5419, opened 2025-01-20
check for object before creating classification

mcoughlin:
This PR adds a check for an object before adding classification.


Theodlz (approved):
LGTM

PR #5420, opened 2025-01-21
Dockerfile - install rsync

Theodlz:
I've been asked to make a clone of Fritz on another machine, as a point in time backup for ZTF partners that should not access the live instance anymore, as they should not see any of the new data.

One problem I have is that copying the persistent data & thumbnails to another machine isn't very easy to do on gcp, and having access to `rsync` would make that much easier.

So, this PR adds it to the Docker image


mcoughlin (approved):
LGTM

PR #5421, opened 2025-01-21
Test Migration - caching issue

Theodlz:
In the migration test, we use the new caching composite action. Issue is we try to cache dependencies after checking out to the reference commit (down migration version of SkyPortal, from which we then try to migrate the db to the latest). However the composite caching does not exist in that down migration, creating a failure.

This PR removes the caching of the dependencies from the reference commit, which while it can help, creates an issue and prevents us from running the GA. We can think about how to reintroduce caching at that step at a later date.

### DIFF
--- a/.github/workflows/test_migrations.yaml
+++ b/.github/workflows/test_migrations.yaml
@@ -129,11 +129,6 @@ jobs:
           ref: ${{ env.MIGRATION_REFERENCE }}
           submodules: true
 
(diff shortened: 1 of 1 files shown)


Theodlz:
hmmm, the migration test don't run here

Theodlz:
ok back to running, I just made sure that editing the migration or caching GA runs the migration test

Theodlz:
self merging so we can get back to green on main

PR #5422, opened 2025-01-21
Queue-based services - add missing sleeps based on queue

Theodlz:
In this PR, we edit the recurring_apis & reminders services to sleep between each iteration. Essentially, if at the end of every processing loop we do not have anything to process in the next 60 seconds, we sleep 60 seconds. If we do have something to process soon, we only sleep until that time. And if the next entry to process should be processed now, then we do not sleep.

This is based on @stefanv's https://github.com/skyportal/skyportal/pull/5415


services/recurring_apis/recurring_apis.py
```diff
@@ -94,6 +96,29 @@ def perform_api_calls():
...
+                RecurringAPI.next_call <= now + timedelta(seconds=sleep_time),
+                RecurringAPI.active.is_(True),
+            )
+            .order_by(RecurringAPI.next_call)
+            .limit(1)
+        ).first()
+        if next_recurring_api is not None:
+            dt = (
+                next_recurring_api.next_call
+                - datetime.now(timezone.utc).replace(tzinfo=None)
```
stefanv, line 112: Can you re-use `now` here? Then you also shouldn't need the negativity check.

stefanv: In fact, you should be able to simply calculate sleep_time directly.

services/recurring_apis/recurring_apis.py
```diff
@@ -94,6 +92,23 @@ def perform_api_calls():
...
+                RecurringAPI.next_call <= now + timedelta(seconds=sleep_time),
+                RecurringAPI.active.is_(True),
+            )
+            .order_by(RecurringAPI.next_call)
+            .limit(1)
+        ).first()
+        if next_recurring_api is not None:
+            dt = (next_recurring_api.next_call - now).total_seconds()
+            if dt < sleep_time:
+                sleep_time = dt
```
stefanv, line 108:
```suggestion
            sleep_time = min(sleep_time, dt)
```

stefanv (approved):
Great, much simpler! Can make it two lines shorter with the recommendation, but up to you :)

Theodlz:
> Great, much simpler! Can make it two lines shorter with the recommendation, but up to you :)

hahaha I know, I just committed some changes but I'm still working on the tests at the moment, so I'll add the min() at the same time. I'll tag you for review once its ready!

stefanv:
At some point, we should perhaps have the microservices each define their own default configuration (in their own directory). Then we can build the default config from those snippets.

Theodlz:
Yes 100%! I had a PR on baselayer some month ago as a first attempt to make the microservices configurable and modular, and thar was the idea. But more config files can also be a pain, so we could also be able to define them somewhere in the main config tike, overwriting the individual services configs

stefanv (approved):

PR #5430, opened 2025-01-25
Data Sharing - frontend refresh fix

Theodlz:
The Sharing handler was sending the obj_id and not the obj_internal_key to the frontend (via websockets) to refresh the data displayed when updating the data access of spectra, which prevented it from being refreshed as expected.

This PR fixes this.

### DIFF
--- a/skyportal/handlers/api/sharing.py
+++ b/skyportal/handlers/api/sharing.py
@@ -113,12 +113,13 @@ class SharingHandler(BaseHandler):
                 for group in groups:
                     spec.groups.append(group)
                 # Grab obj_id for use in websocket message below
-                spec_obj_ids.append(spec.obj_id)
+                spec_obj_ids.append(spec.obj.internal_key)
 
         self.verify_and_commit()
 
-        spec_obj_ids = set(spec_obj_ids)
         phot_obj_ids = set(phot_obj_ids)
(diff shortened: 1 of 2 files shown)


skyportal/handlers/api/sharing.py
```diff
@@ -113,20 +113,22 @@ def post(self):
...
                     spec.groups.append(group)
                 # Grab obj_id for use in websocket message below
-                spec_obj_ids.append(spec.obj_id)
+                spec_obj_ids.append(spec.obj.internal_key)
 
         self.verify_and_commit()
 
-        spec_obj_ids = set(spec_obj_ids)
         phot_obj_ids = set(phot_obj_ids)
+        spec_obj_ids = set(spec_obj_ids)
```
mcoughlin, line 121: Should we note in the variable name that these are all internal?

mcoughlin (approved):
Just an idea that might be good to specify in the variable that these are internal keys.

skyportal/handlers/api/sharing.py
Theodlz: yes good point, let me do that

Theodlz: @mcoughlin done

PR #5431, opened 2025-01-25
Group admission - stream access error message clarification

Theodlz:
the stream access error triggered when adding a user with insufficient stream access to a group was not clear as we reported IDs rather than actual stream names. This PR hopefully makes that message clearer


stefanv (approved):
Looks good to me. Is it okay to make stream names known? This could be a "discovery attack", but not sure that matters.

Theodlz:
Good point, let's make this a draft and I'll bring it up on Tuesday if needed

Theodlz:
we discussed it at the weekly meeting, and displaying stream names is fine, merging

PR #5432, opened 2025-01-25
Issues 5424, 5423 : Display T0 on private and public source page

antoine-le-calloch:
- [x] Display T0 (Time of first detection) on source page.
- [x] Display T0 (Time of first detection) on public source page.
- [x] Improve subtitle style and code readability on public source page.
- [x] Allow edit and clear T0 from the source page.
- [x] Display the photometry plot using T0 (if set) as start range value.
- [x] Update photometry plot range if T0 is update


skyportal/models/obj.py
```diff
@@ -205,6 +205,7 @@ class Obj(Base, conesearch_alchemy.Point):
     offset = sa.Column(
         sa.Float, default=0.0, doc="Offset from nearest static object [arcsec]."
     )
+    t0 = sa.Column(sa.Float, nullable=True, index=True, doc="Time of first detection.")
```
mcoughlin, line 208: I think the description should be slightly different, as it really is just a reference time for this object, which could be a first detection or a GRB time or something else.

static/js/components/source/UpdateSourceT0.jsx
```diff
@@ -0,0 +1,134 @@
...
+        <DialogContent>
+          <div>
+            {(!t0.trim() || isNaN(t0)) && (
+              <FormValidationError message="Please enter a valid float" />
+            )}
+            <TextField
+              className={classes.formInput}
+              data-testid="updateT0Textfield"
+              size="small"
+              label="First detection time (mjd)"
```
mcoughlin, line 86: Same point here.

mcoughlin (approved):
I think we should be more generic in the description of this variable, see comments.

antoine-le-calloch:
> I think we should be more generic in the description of this variable, see comments.

Yeah sure, what should I put ?

mcoughlin:
Reference time or something

Theodlz:
@antoine-le-calloch I just don't think that we need any column in the database. My intuition is that when creating a public page for a source (in a release for example), one would pick if the T0 should be something different than the first detection time. So essentially in the form, you could do something like "display T0: True/Flase => if True => Origin: First detection / GCN Event => if GCN Event => dropdown to select event (or search bar)".

Any thoughts? @mcoughlin

mcoughlin:
@Theodlz that's alright for me.

Theodlz:
We can start simple, with T0 options (in the form) as:
- First detection
- Custom (then display a datetime input or text input, depending on what dateformat you think is best

Theodlz:
we can think about how to read that from a GCN event later if needed

antoine-le-calloch:
Yeah but the PR cover also the T0 edit options on the private source page like the one for redshift, how I can manage this without database entry. And I also use this information to start the plot range.

santier14:
hello @Theodlz my idea was to add T0 either with connection with the first time provided in GCN, or with a T0 of the choice of the user (since the first obs might be confusing if you add ATLAS data upper limit). It is for private page and public and with a different axe over the plot with T-T0 in log :)

Theodlz:
> hello @Theodlz my idea was to add T0 either with connection with the first time provided in GCN, or with a T0 of the choice of the user (since the first obs might be confusing if you add ATLAS data upper limit). It is for private page and public and with a different axe over the plot with T-T0 in log :)

thanks @santier14. We discussed it yesterday with Antoine and he's working on adding some changes

Theodlz (commented):
As discussed the other day @antoine-le-calloch, the plots can use quite a bit of memory for sources with a lot of data. I just want to make sure that's these changes to the plot don't trigger too many additional re-renders.

So my only question was: do you see any obvious performance hit using the source page with these changes? If not, then I think that we can merge this oon

static/js/components/plot/PhotometryPlot.jsx
```diff
@@ -391,9 +394,19 @@ const PhotometryPlot = ({
       return newPoint;
     });
 
+    // Set the range of the plot to be 2% larger than values or if t0 is set, start the range from t0
```
Theodlz, line 397: the 2% is for the mag, doesn't seem related to the time axis. This comment is a little bit confusing

antoine-le-calloch:
@Theodlz No obvious performance hit, the rerender only hit on T0 change or on any switch change, so nothing to worry about for me.

PR #5433, opened 2025-01-28
GCN autogenerated sources: only use dateobs for obj_id

Theodlz:
Currently, to create the obj_id of GCN autogenerated sources (when an event's localization is a small enough cone, we create a source associated with it) we first try to see if the GCN notice came with a pre-defined name for the event, and if not generate a name based on the dateobs of the event.

The issue is for the same event, some notices can have that name and some won't (e.g. SWIFT notices show this irregularity), this means we might end up with 2 sources for the same event: one named based on the notice content, one based on the dateobs. This creates confusion for the users.

This PR simplifies the source creation code and makes sure that we only use the dateobs to generate a source's name/id.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -30,9 +30,7 @@ from astropy.table import Table
 from astropy.time import Time
 from marshmallow import Schema, validate
 from marshmallow.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


Theodlz:
@santier this prevents the duplicate sources

mcoughlin (approved):
This was a good solution.

PR #5452, opened 2025-02-03
P200 NGPS Starlist: formatting changes

Theodlz:
In this PR we:
- change the ra, dec to hms, dms format
- if the associated assignment comment contains the column separator in it (a comma), we replace it with a space to avoid breaking the CSV


mcoughlin (approved):
LGTM

PR #5453, opened 2025-02-04
Observation plan queue - replace heavy dependency methods by lightweight code

Theodlz:
gwemopt's read_skymap function is giving us issues, and as I looked at what it did, it's pretty obvious that most of it doesn't run for us as we already have the skymap read, and only constraints such as confidence level and galactic latitude need to be enforced on the map we got out of the DB. Also, I noticed that the confidence level constraint could be ignored given that we enforce it when we query the instrument tiles, meaning that anyway we don't give gwemopt the tiles that overlap with parts of the map outside of the requested confidence level. That said it does set a bunch of skymap tiles to zero, so we might as well still do that in an abundance of caution when it comes to gwemopt's potential memory usage.

### DIFF
--- a/skyportal/utils/observation_plan.py
+++ b/skyportal/utils/observation_plan.py
@@ -1,12 +1,17 @@
 import time
 import traceback
 
+import astropy_healpix as ah
 import healpix_alchemy as ha
+import healpy as hp
 import humanize
 import numpy as np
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
Great idea. I hope this helps.

PR #5455, opened 2025-02-07
GCN JSON events - missing tags

Theodlz:
I noticed that some - not all - EP events wouldn't get an associated source, and I tracked it as a bug coming from a PR of mine https://github.com/skyportal/skyportal/commit/a2a929b2daccc33db5d8d80650db76a64c4a6270

This PR fixes that


mcoughlin (approved):
LGTM

PR #5459, opened 2025-02-12
Improve follow up information tarot API

antoine-le-calloch:
- [x] Improve the way the submitted status on the follow request are check to allow adding more information after it.
- [x] Improve the error and the submit message for the tarot API
- [x] Check if the requested scene have been observed on an other page from TAROT.
- [x] Display the "pending" status when no status have been found or requested yet on the TAROT Page.
- [x] Add timeout parameter to requests in TAROT API to manage cases where the server is not accessible.
- [x] Add except to catch when there is not the endpoint URL in the env or when the request is timeout.

### DIFF
--- a/services/notification_queue/notification_queue.py
+++ b/services/notification_queue/notification_queue.py
@@ -1068,7 +1068,7 @@ def api(queue):
                                             session.commit()
                                             queue.append(notification.id)
                                 elif is_followup_request:
-                                    if target_data["status"] == "submitted":
+                                    if "submitted" in target_data["status"]:
                                         continue
                                     allocation_id = target_data["allocation_id"]
                                     allocation = session.scalars(
--- a/services/notification_queue/notification_queue.py
+++ b/services/notification_queue/notification_queue.py
@@ -1068,7 +1068,7 @@ def api(queue):
                                             session.commit()
                                             queue.append(notification.id)
                                 elif is_followup_request:
-                                    if "submitted" in target_data["status"]:
+                                    if target_data["status"].startswith("submitted"):
                                         continue
                                     allocation_id = target_data["allocation_id"]
                                     allocation = session.scalars(
--- a/skyportal/facility_apis/tarot.py
+++ b/skyportal/facility_apis/tarot.py
@@ -266,6 +266,7 @@ def login_to_tarot(request, session, altdata):
         f"{cfg['app.tarot_endpoint']}/manage/manage/login.php",
         data=data,
         auth=(altdata["browser_username"], altdata["browser_password"]),
+        timeout=5.0,
     )
 
     if login_response.status_code == 200 and "hashuser" in login_response.text:
@@ -328,6 +329,7 @@ class TAROTAPI(FollowUpAPI):
             f"{cfg['app.tarot_endpoint']}/manage/manage/depot/depot-defaultshort.res.php?hashuser={hash_user}&idreq={altdata['request_id']}",
             data=payload,
             auth=(altdata["browser_username"], altdata["browser_password"]),
+            timeout=5.0,
         )
 
         if "New Scene Inserted" not in response.text:
@@ -393,6 +395,7 @@ class TAROTAPI(FollowUpAPI):
         response = requests.get(
             f"{cfg['app.tarot_endpoint']}/rejected{station_dict[request.payload['station_name']]['url_to_request']}.txt",
             auth=(altdata["browser_username"], altdata["browser_password"]),
+            timeout=5.0,
         )
 
         if response.status_code != 200:
@@ -437,7 +440,7 @@ class TAROTAPI(FollowUpAPI):
                     f"app.{station_dict[request.payload['station_name']]['endpoint']}"
                 ]
 
-                response = requests.get(f"{station_endpoint}/klotz/")
+                response = requests.get(f"{station_endpoint}/klotz/", timeout=5.0)
 
                 if response.status_code != 200:
                     raise ValueError("Error trying to get the observation log")
@@ -542,6 +545,7 @@ class TAROTAPI(FollowUpAPI):
                 f"{cfg['app.tarot_endpoint']}/manage/manage/liste_scene.php?hashuser={hash_user}&idreq={altdata['request_id']}",
                 data=data,
                 auth=(altdata["browser_username"], altdata["browser_password"]),
+                timeout=5.0,
             )
 
             if response.status_code != 200:
--- a/skyportal/facility_apis/tarot.py
+++ b/skyportal/facility_apis/tarot.py
@@ -345,6 +345,7 @@ class TAROTAPI(FollowUpAPI):
         )
 
         session.add(transaction)
+        session.commit()
 
         try:
             flow = Flow()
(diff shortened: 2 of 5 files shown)


skyportal/facility_apis/tarot.py
```diff
@@ -315,12 +330,15 @@ def submit(request, session, **kwargs):
...
+        if "New Scene Inserted" not in response.text:
+            error_response = f"rejected: status code = {response.status_code}. "
+            if response.status_code == 200:
+                error_response += "Scene not Inserted" + (
+                    " - Hashuser not valid" if "secu_erreur" in response.text else ""
+                )
+            request.status = error_response
         else:
-            request.status = "submitted"
+            request.status = "pending"
```
mcoughlin, line 341: Is this the right nomenclature now?

skyportal/facility_apis/tarot.py
```diff
@@ -330,19 +348,22 @@ def submit(request, session, **kwargs):
...
+                    "*",
+                    "skyportal/REFRESH_SOURCE",
+                    payload={"obj_key": request.obj.internal_key},
+                )
+            if kwargs.get("refresh_requests", False):
+                flow.push(
+                    request.last_modified_by_id, "skyportal/REFRESH_FOLLOWUP_REQUESTS"
+                )
+        except Exception as e:
+            log(f"Failed to send notification: {e}")
```
mcoughlin, line 366: Does this need str(e)?

skyportal/facility_apis/tarot.py
```diff
@@ -367,44 +388,76 @@ def get(request, session, **kwargs):
...
+                    "baselayer/SHOW_NOTIFICATION",
+                    payload={
+                        "note": "TAROT request status has not changed",
+                        "type": "warning",
+                    },
                 )
-        else:
-            request.status = f"rejected: Scene {manager_scene_id} for {request.obj.id} not found on TAROT manager"
+            except Exception as e:
+                log(f"Failed to send notification: {e}")
```
mcoughlin, line 457: Same question here

mcoughlin (approved):
Looks good, just a few questions.

skyportal/facility_apis/tarot.py
antoine-le-calloch: image
I have take actual code from other API like trt or swift for this part.

Theodlz: it's fine without it for most errors

skyportal/facility_apis/tarot.py
antoine-le-calloch: Because I implemented the 'get' function, I deleted the part that checks the status on the TAROT manager on submit. So even if the request has been added to the TAROT website, it doesn't mean that the status is 'planned'. There are still all these possible statuses:
            "1": "rejected: date is before the current/upcoming night.",
            "2": "submitted: planned for a future night.",
            "4": "rejected: over quota",
            "5": "submitted: planified",
            "6": "submitted: planified over",
That need to be check by clicking on the "retrieve" button. So change to pending force the user to click the retrieve button. It seems that the logic is correct, this is quite possible, @Theodlz ?

services/notification_queue/notification_queue.py
```diff
@@ -1068,7 +1068,7 @@ async def post(self):
                                             session.commit()
                                             queue.append(notification.id)
                                 elif is_followup_request:
-                                    if target_data["status"] == "submitted":
+                                    if "submitted" in target_data["status"]:
```
Theodlz, line 1071: could be an issue. For some apis we might have a failed status that could contain "submitted" in it.

Instead, please do `target_data["status"].startswith()`

skyportal/facility_apis/tarot.py
```diff
@@ -315,12 +330,15 @@ def submit(request, session, **kwargs):
...
+        if "New Scene Inserted" not in response.text:
+            error_response = f"rejected: status code = {response.status_code}. "
+            if response.status_code == 200:
+                error_response += "Scene not Inserted" + (
+                    " - Hashuser not valid" if "secu_erreur" in response.text else ""
+                )
+            request.status = error_response
         else:
-            request.status = "submitted"
+            request.status = "submitted: please retrieve status"
```
Theodlz, line 341: the "please" isn't really necessary, maybe just"submitted: use retrieve to validate submission" or something (not too long) that makes it clear that this is necessary to be sure that it was submitted.

static/js/components/allocation/AllocationSummary.jsx
```diff
@@ -105,7 +105,7 @@ const SimpleMenu = ({ request }) => {
         open={Boolean(anchorEl)}
         onClose={handleClose}
       >
-        {(request.status === "submitted" ||
+        {(request.status.includes("submitted") ||
```
Theodlz, line 108: same comment here, we want to check that it starts with submitted, not just that it's present somewhere (could be in an error msg)

static/js/components/allocation/AllocationSummary.jsx
```diff
@@ -116,7 +116,7 @@ const SimpleMenu = ({ request }) => {
             Mark Observed
           </MenuItem>
         )}
-        {(request.status === "submitted" ||
+        {(request.status.includes("submitted") ||
```
Theodlz, line 119: same comment here

static/js/components/followup_request/FollowupRequestLists.jsx
```diff
@@ -332,9 +332,9 @@ const FollowupRequestLists = ({
         const isDone =
           followupRequest.status === "Photometry committed to database";
 
-        const isSubmitted = ["submitted", "pending"].includes(
-          followupRequest.status,
-        );
+        const isSubmitted =
+          followupRequest.status === "pending" ||
+          followupRequest.status.includes("submitted");
```
Theodlz, line 337: same comment here

static/js/components/followup_request/FollowupRequestLists.jsx
```diff
@@ -332,9 +332,9 @@ const FollowupRequestLists = ({
         const isDone =
           followupRequest.status === "Photometry committed to database";
 
-        const isSubmitted = ["submitted", "pending"].includes(
-          followupRequest.status,
-        );
+        const isSubmitted =
+          followupRequest.status === "pending" ||
```
Theodlz, line 336: might as well treat the other statuses to the same standard (startswith)

skyportal/facility_apis/tarot.py
Theodlz, line 1: it would be very appreciated if you could add somethere at the beginning of that file some error handling in case the tarot URLs aren't in the config. If they aren't we don't want to stop SkyPortal from running, but we want it to - when submitting/retrieving/deleting" to not perform the requested action and instead to return a 400 (or at least set a status) to make it clear that it's not working because the instance isn't setup to use tarot.

Theodlz (changes requested):
minor changes requested

skyportal/facility_apis/tarot.py
antoine-le-calloch: if cfg["app.tarot_endpoint"] is None:
            raise ValueError("TAROT endpoint not configured")
            
I have this line already, is it not sufficient ?

skyportal/facility_apis/tarot.py
```diff
@@ -34,6 +35,20 @@
 }
 
 
+def catch_timeout_and_no_endpoint(func):
```
Theodlz, line 38: nice!

Theodlz (approved):
LGTM

PR #5460, opened 2025-02-13
Dockerfile: add top command

Theodlz:
This PR adds `top` to the Dockerfile, so we can easily monitor CPU and RAM usage in production


Theodlz:
self merging

PR #5463, opened 2025-02-15
Improve followup request

antoine-le-calloch:
Improve text display during creation and retrieve for a follow-up request to a more generic text.

Display a new field with the station name in the list of submitted request when the allocation have multiple station:
image

image


mcoughlin (approved):
LGTM

PR #5464, opened 2025-02-16
Starlist - download button

Theodlz:
Adds a download button on the starlist


Theodlz:
minor change, self merging

PR #5465, opened 2025-02-17
Moving object observation planning

Theodlz:
This PR adds the backend necessary to do some basic observation planning for moving objects. The idea is that given:
- an instrument ID
- start time
- end time
- exposure count
- exposure time
- the name of a moving object
- (and some optional constraints)

we use JPL Horizons' API to retrieve position+time at a small sampling rate, apply our constraints, and find the instrument's fields that overlap with these pointings. Once we have that, we find which field gives us the longest consecutive observations that fit the requested exposure count + time. Then, we can return a list of observations of an instrument's field that can be performed to obtain images of said moving objects.

TODOs:
- [x] add unit tests

The frontend will be added in a future PR. We also plan to - longer term - move to an implementation where the selected moving object to plan observations for will be in the database and will have dedicated frontend pages.


skyportal/handlers/api/moving_object.py
```diff
@@ -0,0 +1,175 @@
...
+                            instrument_id:
+                                type: integer
+                                description: ID of the instrument to use
+                            exposure_count:
+                                type: integer
+                                description: Number of exposures
+                            exposure_time:
+                                type: number
+                                description: Exposure time in seconds
+                            start_time:
```
mcoughlin, line 56: Maybe make it clear this is a time window that the code will attempt to fit the exposures (specified by exposure_time and exposure_count) in this window.

skyportal/handlers/api/moving_object.py
```diff
@@ -0,0 +1,175 @@
...
+                                type: string
+                                format: date-time
+                                description: Start time of the observations
+                            end_time:
+                                type: string
+                                format: date-time
+                                description: End time of the observations
```
mcoughlin, line 63: Perhaps observational window.

skyportal/handlers/api/moving_object.py
```diff
@@ -0,0 +1,175 @@
...
+                            filter:
+                                type: string
+                                description: Filter to use
+                            primary_only:
+                                type: boolean
+                                description: Only consider primary observations
```
mcoughlin, line 69: primary field grid

skyportal/utils/calculations.py
```diff
@@ -269,3 +294,67 @@ def radec2lb(ra, dec):
...
+    float
+        Decimal degrees
+    """
+    d, m, s = input.split(" ")
+    d, m, s = float(d), float(m), float(s)
+    sign = 1 if d >= 0 else -1
+    return sign * (abs(d) + m / 60 + s / 3600)
+
+
+def radec_to_healpix(row: dict | pd.Series) -> int:
```
mcoughlin, line 346: Any need to also vectorize this?

skyportal/utils/moving_objects.py
```diff
@@ -0,0 +1,505 @@
...
+cache_dir = "cache/moving_object_ephemeris"
+cache = Cache(
+    cache_dir=cache_dir,
+    max_items=100,
+)
+
+log = make_log("api/moving_object")
+
+# next we define some constants
+BASE_URL = "https://ssd.jpl.nasa.gov"
```
mcoughlin, line 36: Any need for this to be specified in the config instead?

skyportal/utils/moving_objects.py
```diff
@@ -0,0 +1,505 @@
...
+    "dec",
+    "airmass",
+    "mag_ex",
+    "mag_ap",
+    "drop1",
+    "drop2",
+]
+
+
+def find_obj(obj_name: str):
```
mcoughlin, line 51: Could name this like find_obj_horizons to disambiguate.

mcoughlin (approved):
This looks great. Just some small comments on the docs.

skyportal/utils/calculations.py
Theodlz: you can only vectorize (at leat with np.vectorize) pure numpy/python functions, won't work here at least not as expected I bet

skyportal/utils/moving_objects.py
Theodlz: I don't think so, since it's not like there are multiple copies of that API

skyportal/utils/moving_objects.py
Theodlz: good point

Theodlz:
I should mention that the frontend tests that's failing is an issue that's on main. Can't figure out, it passes consistently locally...

skyportal/tests/api/candidates_sources_events/test_default_observation_plan.py
```diff
@@ -3,64 +3,20 @@
...
-                    "limiting_magnitude": 20.3,
-                    "magsys": "ab",
-                    "exposure_time": 30,
-                    "zeropoint": 26.3,
-                }
-            },
-        },
-        token=super_admin_token,
+def test_default_observation_plan_tiling(super_admin_token, public_group):
+    telescope_id, instrument_id, _, _ = add_telescope_and_instrument(
```
mcoughlin, line 17: Great addition.

skyportal/tests/api/candidates_sources_events/test_observation.py
```diff
@@ -1,69 +1,24 @@
...
 from skyportal.tests import api
+from skyportal.tests.external.test_moving_objects import (
+    add_telescope_and_instrument,
+    remove_telescope_and_instrument,
+)
 
 
-@pytest.mark.flaky(reruns=2)
-def test_observation(super_admin_token, view_only_token, gcn_GW190425):
+# @pytest.mark.flaky(reruns=2)
```
mcoughlin, line 15: Can just delete if you do not need it.

skyportal/tests/external/test_moving_objects.py
```diff
@@ -0,0 +1,171 @@
...
+
+import arrow
+import pandas as pd
+import pytest
+from regions import Regions
+
+from skyportal.tests import api
+
+
+def add_telescope_and_instrument(api_class, super_admin_token, fields_ids=[]):
```
mcoughlin, line 13: I just wonder if this function belongs in a different file? One dedicated to adding and removing repeated uses.

skyportal/tests/utils/test_moving_objects.py
```diff
@@ -0,0 +1,237 @@
...
+import numpy as np
+import pandas as pd
+
+from skyportal.models import DBSession
+from skyportal.tests.external.test_moving_objects import add_telescope_and_instrument
+from skyportal.utils.calculations import radec_to_healpix
+from skyportal.utils.moving_objects import *
+
+
+def test_find_jplhorizon_obj():
```
mcoughlin, line 12: Any concerns about this needing its own place depending on an external service?

skyportal/tests/utils/test_moving_objects.py
```diff
@@ -0,0 +1,237 @@
...
+
+    with DBSession() as session:
+        dfs, _ = add_instrument_fields(
+            data, instrument_id, "ZTF", session, observer, primary_only=True
+        )
+        df = dfs[0]
+
+        df = count_consecutive(df)
+        start, end, field_id, count = find_longest_sequence(df)
+        print(start, end, field_id, count)
```
mcoughlin, line 165: Remove print

mcoughlin (approved):
Awesome

skyportal/tests/api/candidates_sources_events/test_observation.py
Theodlz: right, let me do that

skyportal/tests/external/test_moving_objects.py
Theodlz: So I tried to put it under API but then it creates conflicts when importing the API method from __init__.py. It think that deserves it's own refactor where we put `api()` somewhere else first

skyportal/tests/utils/test_moving_objects.py
Theodlz: The external folder is also green on failure like flaky (not for all the tests, so I'd rather keep this here and move it to external only if the JPL API happens to be flaky

skyportal/tests/utils/test_moving_objects.py
Theodlz: thanks

Theodlz: done

skyportal/tests/api/candidates_sources_events/test_observation.py
Theodlz: done!

PR #5466, opened 2025-02-19
Add switch for photometry plot axe change

antoine-le-calloch:
Add more switch behind the photometry plot:
- [x] One switch to display "t - t0" in seconds instead of "days ago"
- [x] One switch to display only the "t - t0" axis in a log way.

image
image
image


mcoughlin (approved):
Might be worth thinking about if seconds is the right unit (or it should still be days), but I am glad this concept is being added.

antoine-le-calloch:
> Might be worth thinking about if seconds is the right unit (or it should still be days), but I am glad this concept is being added.
it was at Sarah's request @mcoughlin

Theodlz:
> > Might be worth thinking about if seconds is the right unit (or it should still be days), but I am glad this concept is being added.
> 
> it was at Sarah's request @mcoughlin

@santier should it be in seconds? please confirm.

@antoine-le-calloch feels a bit odd to have the "use T0" aligned with the forced phot and non det, and then the other T0-related options in the middle. Maybe you can reposition all the T0-related sliders in the middle?

Theodlz:
Maybe we should simplify this even. I think that as soon as the T0 is selected we can change the X axis to be time since T0. that way you only have 2 sliders not 3.

antoine-le-calloch:
Done @Theodlz.

Theodlz:
Thanks a lot Antoine. I'll run this locally later today.

antoine-le-calloch:
@Theodlz, axis in days updated.

PR #5468, opened 2025-02-20
Triggering constrain: bugfix + not trigger if assigned

Theodlz:
This PR fixed the not_if_tns_reported constraint which was ignored, and adds a new "don't report if assigned to an observing run" constraint: not_if_assignment_exists".

This is required for btsbot


mcoughlin (approved):
LGTM

PR #5469, opened 2025-02-20
Bugfixes: Finding chart print & GCN automated source generation

Theodlz:
In this PR we:
- fix the finding chart print button, broken after the update of a dependency. We also refactor some of the code to use the already downloaded & rendered image rather than retrigger a download when hitting "Print"
- fix the GCN sources autogeneration, which was broken - yet again - by a previous PR of mind, where the "id" field was renamed to "name" by error.


mcoughlin (approved):
LGTM

PR #5472, opened 2025-02-21
GCN Ingestion - SVOM support

Theodlz:
Supporting SVOM notices requires to finally move away from the old GCN package enum based system, so this PR makes that transition to string-based notice_types event for VOEVENTs.

Overall we add support for SVOM (both for autoingestion and manually ingestion), adapt the notification queue system, handle a few errors in the automated source ingestion (+ the automated follow-up and source releases system), and some better error handling in the observation plan queue.


config.yaml.defaults
```diff
@@ -718,28 +718,31 @@ gcn:
...
-    # - AMON_ICECUBE_HESE
-    - ICECUBE_ASTROTRACK_GOLD
-    - ICECUBE_ASTROTRACK_BRONZE
-    - MAXI_UNKNOWN
-  json_notice_types:
-    - einstein_probe.wxt.alert
-    - swift.bat.guano
-    - icecube.lvk_nu_track_search
+    voevent: []
+      # - gcn.notices.svom.voevent.grm
```
mcoughlin, line 722: Should we document this more? Also do we want everything commented?

mcoughlin (approved):
LGTM

config.yaml.defaults
Theodlz: I feel like unless folks want to do MMA stuff, this isn't useful. Also you need to have gcn credentials setup so when someone does that they will pick the notices they want to listen too. I don't expect people to want to use our defaults.

Theodlz: But yes let's add a comment pointing to the page where these can be found.

Theodlz: added @mcoughlin

PR #5475, opened 2025-02-22
TNS submission validation - handle 400 errors

Theodlz:
handle 400 errors which can occur when the reports sent are not valid, or just rejected by TNS. This happens after the report is sent, when we check if it has been posted on TNS using the provided submission ID.


mcoughlin (approved):
LGTM

PR #5476, opened 2025-02-25
SpectrumHandler - filter on modified date

Theodlz:
allow users to filter on when a spec was modified (before/after) when retrieving data from the SpectrumHandler


mcoughlin (approved):
LGTM

PR #5477, opened 2025-02-25
GCN source creation - deduplication between API handler & GCN service

Theodlz:
it looks like the API had it's own source creation code that was very outdated and caused issues in production, especially as it saved the autogenerated source to all the groups of a user and not just the sitewide group.

This PR creates one method that can be reimported by both functions that need to create a source.


mcoughlin (approved):
LGTM

PR #5478, opened 2025-02-26
GCN cone-based localization: avoid zero radius

Theodlz:
Some notices come with a radius = 0, which of course makes no sense, and would lead to incorrect sky localizations and unpredictable errors (e.g. incorrect localization center, area calculation, ...).

If the radius is 0, in this PR we set it to a very small value instead (4 arcsec) to avoid said errors. We might have to increase that value if still causes issues.

EDIT:
We also now use the localization name as the center we report for a localization, rather than the ligo.skymap calculated ones which gives us weird results for those small maps

### DIFF
--- a/skyportal/models/localization.py
+++ b/skyportal/models/localization.py
@@ -16,6 +16,8 @@ import ligo.skymap.moc
 import ligo.skymap.postprocess
 import numpy as np
 import sqlalchemy as sa
+from astropy import units as u
+from astropy.coordinates import SkyCoord
 from astropy.table import Table
 from dateutil.relativedelta import relativedelta
 from dustmaps.config import config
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

Theodlz:
@mcoughlin I also just added code to use the localization name as the center we report for a localization rather than the ligo calculated one, which gives us weird results for those small maps

Theodlz:
@mcoughlin weird that the ligo stuff is so incorrect here, I wonder if it's just not robust for small enough maps, or if it's the map generation itself which is faulty here

Theodlz:
hmmmm I suspect that since we flatten the map to a fixed healpix resolution (which is not so high res), that could explain why for small localizations we can't get the center right using the healpix map

PR #5500, opened 2025-03-05
Fix search bar on public sources page not working

antoine-le-calloch:
Fix source link structure and search functionality in sources template to fit the releases template logic.


Theodlz (approved):
LGTM

PR #5502, opened 2025-03-08
SVOM Properties & Alias + UpdatePhotometryForm data validation

Theodlz:
This PR:
- reads additional properties from the SVOM notices (Timescale, event type, ...)
- reads the burst ID when looking for aliases for SVOM events
- make sure that either mag+magerr, or limiting magnitude are specified on the frontend form
- make sure that even when there is no data, mag fields (mag, magerr, limiting mag) are set to null and sent to the backend, which expects them

PS: This hasn't been tested yet, I want to try the frontend before merging


static/js/components/photometry/UpdatePhotometry.jsx
```diff
@@ -89,6 +86,40 @@ const UpdatePhotometry = ({ phot, magsys }) => {
...
+        newState[key] === undefined ||
+        newState[key] === ""
+      ) {
+        delete newState[key];
+      }
+    });
+
+    if (
+      newState?.limiting_mag === undefined &&
+      newState?.mag === undefined &&
```
mcoughlin, line 101: Does this actually require a magnitude and magnitude error? Should it?

mcoughlin (approved):
Just a question on the check on limiting magnitude and magnitude/error.

static/js/components/photometry/UpdatePhotometry.jsx
Theodlz: I mean I think this is the issue we are trying to fix right? We don't need a mag and magerr if we have a limiting mag, and vice versa. But we do need one of these 2 options (detection or non detection data) to be given.

mcoughlin: Ok great.

PR #5504, opened 2025-03-08
Improve allocation altdata management

antoine-le-calloch:
Enhance allocation creation by improving the code and implement the same display of the altdata on the allocation modification form without the "required".


mcoughlin:
@antoine-le-calloch @Theodlz I will leave Theo to review this one as I didn't work on the current implementation of altdata stuff.

static/js/components/allocation/NewAllocation.jsx
```diff
@@ -235,33 +197,34 @@ const NewAllocation = ({ onClose }) => {
...
+                  }
+                : {}),
+            },
+          };
+        }),
+      },
+    },
   };
 
-  if (
```
Theodlz, line 240: why is this deleted?

Theodlz (commented):
I'm not sure what was deleted here should have been

Theodlz:
also we need to be **really** careful about allocation edits. It should be clear if the newly inputted alt data replaces the existing one entirely, or if only edited fields replace the ones that are already in the DB. i.e. do we update the altdata entirely or do we update it's content. The latter seems preferable. I'm mentioning this because now that we make editing the altdata much easier using a form, we need to make sure that the behavior is explicit to the user, so that they don't edit one field of the altdata and see all existing data removed.

static/js/components/allocation/NewAllocation.jsx
antoine-le-calloch: Because not needed anymore everything is done on the schema creation using the altdata form. Way more clean this way so.

skyportal/handlers/api/allocation.py
```diff
@@ -640,19 +640,24 @@ def put(self, allocation_id):
 
             data = self.get_json()
             data["id"] = allocation_id
+
+            only_update_changed_fields = data.pop("only_update_changed_fields", False)
```
Theodlz, line 644:
```suggestion
            only_update_changed_fields = data.pop("only_update_modified_fields", False)
```

Let's try to keep using the same vocabulary as the rest of the app. One might wonder if it's not easier to just have a `replace_altdata` True/False here instead. That's just a suggestion / personal preference, doesn't mean that you should make that change

skyportal/handlers/api/allocation.py
```diff
@@ -640,21 +640,25 @@ def put(self, allocation_id):
...
                     for k, v in data["_altdata"].items()
                     if v is not None and str(v).strip() != ""
                 }
-                # if it ends up being a dictionnary with no keys, remove it
-                # otherwise convert it to a string
-                if not data["_altdata"] or not any(data["_altdata"]):
+
+                if not data["_altdata"]:
                     data.pop("_altdata")
+                elif allocation._altdata and replace_altdata:
```
Theodlz, line 655: hmmmm doesn't that feel somewhat reversed? Idk but to me it feels like "replace altdata" means fully replacing it, not updating its content with newly provided values if any. Idk I feel like the naming here is a bit confusing, any thoughts @mcoughlin?

Theodlz (approved):
LGTM

antoine-le-calloch:
> Oh one thing @antoine-le-calloch, git diff doesn't show me here but is the replace_altdata parameter documented in the docstring of the API?

Nope, i'm adding it.

antoine-le-calloch:
@Theodlz doc update, is it good for you ?

PR #5506, opened 2025-03-11
Moving Object Observation planning - minimal frontend

Theodlz:
This PR: 
* adds a minimal frontend to generate observation plans for moving objects
* adds a route and menu icon to that new page in the config
* handles some connection errors that can occur when querying the JPL horizon API

### DIFF
--- a/static/js/components/moving_object/MovingObjectObsPlanPage.jsx
+++ b/static/js/components/moving_object/MovingObjectObsPlanPage.jsx
@@ -225,7 +225,13 @@ const MovingObjectObsPlanPage = () => {
   };
 
   if (!instruments) {
-    return "Loading...";
+    return (
+      <Paper>
+        <div className={classes.paperContent}>
+          <CircularProgress />
+        </div>
+      </Paper>
+    );
   }
 
   return (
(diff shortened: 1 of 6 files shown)


static/js/components/moving_object/MovingObjectObsPlanPage.jsx
```diff
@@ -0,0 +1,293 @@
...
+    fontSize: "1.25rem",
+    textAlign: "center",
+  },
+}));
+
+const PlaceHolder = () => {
+  const classes = useStyles();
+  return (
+    <div className={classes.spinner}>
+      <TextLoop interval={1500}>
```
mcoughlin, line 45: We should use this more.

mcoughlin (approved):
LGTM

static/js/components/moving_object/MovingObjectObsPlanPage.jsx
Theodlz: yeah! I stumbled upon it when fixing some code in the finding chart stuff, it's nice to give the user a loading bar that gives them the illusion that something is happening, to capture their attention. Makes them forget that some things are actually quite slow 😉

PR #5507, opened 2025-03-14
Publish photometry to hermes

antoine-le-calloch:
- [x] Implementing an handler to publish photometry from an object to Hermes kafka topics.
- [x]  Allow user to publish data to Hermes from source page using the same logic as TNS.


skyportal/handlers/api/tns/bulk_tns.py
```diff
@@ -0,0 +1,118 @@
...
+                    tnsrobot.id,
+                    self.associated_user_object.id,
+                    include_photometry=include_photometry,
+                    include_spectra=include_spectra,
+                    start_date=start_date.isot,
+                    group_ids=group_ids,
+                ),
+            )
+
+            return self.success()
```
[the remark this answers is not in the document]
Theodlz: You can just go ahead and delete the whole handler.

Theodlz:
@antoine-le-calloch if this PR requires the code from the tns dedicated PR (so stuff is duplicated), for now please make this one point to the tns PR's branch instead so we only see the relevant diff

antoine-le-calloch:
> @antoine-le-calloch if this PR requires the code from the tns dedicated PR (so stuff is duplicated), for now please make this one point to the tns PR's branch instead so we only see the relevant diff

Yeah yeah for sure it was the idea.

Theodlz:
@antoine-le-calloch I merged the TNS stuff to main and merged main to your branch, looks nicer now :)

antoine-le-calloch:
@Theodlz  @mcoughlin 
Here's a first version of a new logic that allows easier management of data access (obj, photometry) when publishing a source. This logic can be applied to both TNS and Hermes submissions. I essentially reused the TNS code, refactored and simplified it, so the methods in the data_access file can be called by tns_submission_queue. However, for now, the main general function is only used by Hermes because I didn’t want to risk breaking TNS with it. It seems to work for Hermes based on my initial tests. I still need to adapt the tests, but I think this could already benefit from a first review.

The idea behind all of this—discussed with @Theodlz —is to eventually centralize all the data access logic related to public data submission, and make it no longer TNS-specific. So for Hermes, I’ve tried to start with a clean first implementation for now. Let me now. (A first PR with just the creation of some parsing utils function is also available and merge here #5570 https://github.com/skyportal/skyportal/pull/5570).

Theodlz:
Thanks a lot for all the hard work @antoine-le-calloch. I will do my best to review this before the beginning of the week

antoine-le-calloch:
> Thanks a lot for all the hard work @antoine-le-calloch. I will do my best to review this before the beginning of the week

No rush, no need to do it before the week.

Theodlz:
> > Thanks a lot for all the hard work @antoine-le-calloch. I will do my best to review this before the beginning of the week
> 
> No rush, no need to do it before the week.

says the guy pushing code after midnight on a Friday 😉

Theodlz:
Oh I see that a number of changes here are in another PR (the parsing stuff). Let's get the other one wrapped up and merged before I review this, to avoid the duplication.

antoine-le-calloch:
> Oh I see that a number of changes here are in another PR (the parsing stuff). Let's get the other one wrapped up and merged before I review this, to avoid the duplication.

@Theodlz  yeah its what I say on my previous comment.

Theodlz:
there are quite a bit of conflicts with main, @antoine-le-calloch if you can merge that branch with main and solve those locally that would make it easier to review. If you can't do that tomorrow let me know and I'll try to do it myself.

antoine-le-calloch:
> there are quite a bit of conflicts with main, @antoine-le-calloch if you can merge that branch with main and solve those locally that would make it easier to review. If you can't do that tomorrow let me know and I'll try to do it myself.

Done !

skyportal/handlers/api/hermes/hermes.py
```diff
@@ -0,0 +1,240 @@
...
+from baselayer.app.access import auth_or_token
+from baselayer.app.env import load_env
+from baselayer.app.flow import Flow
+from baselayer.app.handlers.base import BaseHandler
+from baselayer.log import make_log
+
+from ....models import TNSRobot
+from ....utils.data_access import get_publishable_obj_photometry
+
+_, cfg = load_env()
```
Theodlz, line 15: What I would suggest (we do that in some handlers already) is to check at startup (so anywhere in the top level of the file, say here after the imports) if the app is configured to use hermes. You just create a boolean that is set to True if there is all 3 config keys that are required for this to work. If not, then you can have a decorator or just simply an if statement in the handlers that require hermes to be configured to work, and return an error saying "this instance isn't configured to use Hermes, missing ...". That way it's super clear for the user why they can't use it and they can report to the admin, and you don't need to look at the config more than once.

Theodlz: Caus like for example I see that you check some error if the endpoint is missing from the config in your `atch_timeout_and_no_endpoint` decorator but that only happens after calling the handler. The config check should happen before calling the method.

skyportal/handlers/api/hermes/hermes.py
```diff
@@ -0,0 +1,240 @@
...
+                    "bandpass": p.filter,
+                    "brightness": p.mag,
+                    **({"brightness_error": p.e_mag} if p.e_mag else {}),
+                    "unit": "AB mag",
+                    **(
+                        {
+                            "limiting_brightness": p.original_user_data["limiting_mag"],
+                            "limiting_brightness_unit": "AB mag",
+                        }
+                        if p.original_user_data
```
Theodlz, line 75: I don't think that should be read from the original user data. The limiting magnitude (just like the mag and magerr) are computed on the fly using the flux/fluxerr that we keep in the DB. You don't need to look at the original user data, that's another thing entirely which we don't need for this purpose.

skyportal/handlers/api/hermes/hermes.py
```diff
@@ -0,0 +1,240 @@
...
+
+    payload = {
+        "topic": cfg["app.hermes.topic"],
+        "title": data["title"],
+        "submitter": data["submitter"],
+        "data": {
+            "targets": [
+                {
+                    "name": obj.id,
+                    "ra": obj.ra,
```
Theodlz, line 56: Not in this one but in a future PR for both Hermes and TNS we should make use to use a ra/dec computed from a flux weighted centroid, just like the `ObjPosition` handler does. That yields a position which is better than the plain ra/dec.

skyportal/handlers/api/hermes/hermes.py
```diff
@@ -0,0 +1,240 @@
...
+
+    return json.dumps(payload), header
+
+
+def validate_payload_and_header(payload, header):
+    response = requests.post(
+        f"{cfg['app.hermes.endpoint']}/submit_message/validate/",
+        data=payload,
+        headers=header,
+        timeout=5.0,
```
Theodlz, line 93: Just curious, is hermes really that slow at times? (5 sec seems like a lot for a kafka based service which goal is to be fast...)

skyportal/handlers/api/hermes/hermes.py
```diff
@@ -0,0 +1,240 @@
...
+            ).first()
+            if tns_robot is None:
+                return self.error(f"TNSRobot {tns_robot_id} not found")
+            if tns_robot.testing:
+                try:
+                    flow = Flow()
+                    flow.push(
+                        action_type="baselayer/SHOW_NOTIFICATION",
+                        payload={
+                            "note": "Successfully validate by the API (testing mode, not sent to Hermes)",
```
Theodlz, line 221:
```suggestion
                            "note": " Payload validated successfully (testing mode, not sent to Hermes)",
```

skyportal/utils/data_access.py
Theodlz, line 1: Very nice refactor. Do the methods do the same thing, or has the logic changed in some of them? Just making sure we don't break anything for TNS.

Theodlz: Also some of these methods take a third optional argument (with some existing list of ids (e.g. accessible_instrument_ids=None), which I don't see used in the handler where you removed these functions. Aren't they imported somewhere where these extra arguments are used, like the TNS service or some other handler)???

Theodlz: I guess my question is: are these extra arguments truly not used somewhere in which case it's just dead code?

Theodlz (commented):
Hey @antoine-le-calloch, I left a batch of comments on the current version of the code. Very nice overall, just a few comments / questions.

skyportal/handlers/api/hermes/hermes.py
antoine-le-calloch: Not that slow I just use the same time everywhere else. But I can lower it.

Theodlz: I was just curious about if you picked that time based on your testing or arbitrarily, you don't need to change it, you ultimate decide what that value should be anyways

antoine-le-calloch:
> @antoine-le-calloch anything else you're planning on adding? Just so I know if I can give a go at this locally now or if I should wait for some changes (before giving a final test + review).

Yes still need to made some change sorry.

Theodlz:
> > @antoine-le-calloch anything else you're planning on adding? Just so I know if I can give a go at this locally now or if I should wait for some changes (before giving a final test + review).
> 
> Yes still need to made some change sorry.

no worries and no rush, just wanted to make sure I didn't review at the wrong time!

antoine-le-calloch:
@Theodlz its good !

Theodlz:
> @Theodlz its good !

I noticed some issues like the tns robot list being infinitely fetched when opening the Hermes form, and how much space the extra button took on the source page header (it's already pretty busy in there and it's worse in Fritz/Icare due to the extra buttons) so I combine the TNS and Hermes ones in a dropdown menu. Please take a look and don't hesitate to make changes.

antoine-le-calloch:
> > @Theodlz its good !
> 
> I noticed some issues like the tns robot list being infinitely fetched when opening the Hermes form, and how much space the extra button took on the source page header (it's already pretty busy in there and it's worse in Fritz/Icare due to the extra buttons) so I combine the TNS and Hermes ones in a dropdown menu. Please take a look and don't hesitate to make changes.

Seems really nice !

Theodlz (approved):
LGTM, I realized that we can't actually see the hermes.test topic's content from the UI of Hermes so a little hard to check if things worked,  but that's not a major concern right now, we'll be talking to their team and we can bring that up.

mcoughlin:
@Theodlz @antoine-le-calloch Can you remind me what the status / next PR is?

Theodlz:
Slowly but surely merging it with the tns submission queue, so in general unifying the 2 implementations to not make things TNS specific anymore. That will require some careful thinking. We need to talk about it and open an epic

mcoughlin:
@Theodlz @antoine-le-calloch Could we get this epic opened to track progress?

PR #5508, opened 2025-03-17
Improve instrument creation and update

antoine-le-calloch:
Improve the instrument creation and modification component.
- Rename NewInstrument component to InstrumentForm
- Update its functionality to support both creating and modifying instruments. 
- Simplify the code.
- Display more info about the instrument being updated (name, api, api_obs).
- Change useless array fields to a string fields to avoid closing select drop down when choosing an option.
- Update the global style to target empty option from mui select Form and display them grayed out for a better user experience.
- Refactor SourcePublishRelease to simplify release property schema and use the new global style.

image
image


Theodlz:
Seems ok to me, thanks for the cleanup @antoine-le-calloch. May I ask if this was done based on a ticket? Just trying to figure out why this needed fixing.

antoine-le-calloch:
> Seems ok to me, thanks for the cleanup @antoine-le-calloch. May I ask if this was done based on a ticket? Just trying to figure out why this needed fixing. Just to I can know how urgent the review is or isn't.

Nothing urgent, just working on a improvement of the Tarot API and meet this code so need to correct this to work with it after.

Theodlz (approved):
LGTM

PR #5509, opened 2025-03-17
SVOM GCN properties - handle string types

Theodlz:
This PR fixes a bug introduces recently with the support of additional SVOM GCN notices properties, where some ended up being strings, but we always try to cast to float which creates errors.


Theodlz:
This is a pretty straightforward PR and we need this in prod, so self merging

PR #5510, opened 2025-03-18
Dockerfile: pin setuptools version

Theodlz:
pin a max version for setuptools, as the newly released one (76.1.0)  breaks the installer for a number of packages and causes issues with distutils


Theodlz:
This is breaking our docker deployment and therefore Fritz and other instances, self merging

stefanv:
Worth figuring out what's going on, unless it is something setuptools will fix, otherwise we're going to get bogged down on this old version.

PR #5513, opened 2025-03-19
Facility Transaction Service: DB rather than API-driven

Theodlz:
At the moment the Facility Transaction micro service fetched existing and un-processed requests from the DB only once when starting up, but then is fed new requests by an API endpoint that the different APIs can communicate with. This isn't ideal, and can cause issues.

In this PR we simplify that service's logic by simply letting it query requests to process (either new requests, or uncompleted requests which has been last processed more than N minutes ago) from the DB in it's main loop, and remove the API component entirely.

The ATLAS and ZTF forced photometry services (which are the only 2 using that microservice) are also edited to avoid trying to push data to that API.

One more thing that needs to be considered is how we keep triggering way too many notifications for the requests. I'm not sure what is happening, but it feels like we keep spamming users with notifications even if the status of the requests hasn't really changed. Using sqlalchemy's event listens can be useful but I wonder if it's not just making things much harder than they need to be in the context of follow-up requests. Perhaps we should explicity create these notifications in the POST/PATCH handlers of the follow-up requests and in this micro-service instead. This should be addressed in its own PR.

### DIFF
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -135,7 +135,7 @@ def service():
                             try:
                                 json_response = response.json()
                             except Exception:
-                                raise ("No JSON data returned in request")
+                                raise ValueError(f"No JSON data returned in request")
 
                             if json_response["finishtimestamp"]:
                                 followup_request.status = (
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -135,7 +135,7 @@ def service():
                             try:
                                 json_response = response.json()
                             except Exception:
-                                raise ValueError(f"No JSON data returned in request")
+                                raise ValueError("No JSON data returned in request")
 
                             if json_response["finishtimestamp"]:
                                 followup_request.status = (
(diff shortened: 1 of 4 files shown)


mcoughlin (approved):
LGTM

Theodlz:
@mcoughlin @antoine-le-calloch btw there is still some duplicated code in there that can be removed, planning to make that other improvement in a separate PR. Right now, I just want to get rid of this API-like format

services/facility_queue/facility_queue.py
antoine-le-calloch, line 1: @Theodlz The code you changed looks good. Just one point, but it's not about what you changed, I just found this in the file:
`raise ("No JSON data returned in request")
`
I think it would be better to raise an instance of Exception here, something like:
`raise Exception("No JSON data returned in request")`
or another similar instance.

Theodlz: Nice catch! Fixing it now

Theodlz: done

PR #5514, opened 2025-03-21
Improve instruments management and hydrate

antoine-le-calloch:
The main cause of this PR was related to the hydration behavior of different API forms. When an instrument or an allocation was updated or created in a way that should have made it appear or change in the list of possible follow-up request allocations, the update did not take effect. This happened because modifying or adding an instrument or allocation only refreshed the list of instruments or allocations but did not update the list of forms or the allocationListApiClassname, which is used for the follow-up request selection.

With Sarah, we had filtered the allocations and created one for TAROT, but the change did not appear and could not be selected. So I had already noticed this behavior but didn’t know its cause. While working on the TAROT API, I finally understood where the issue was coming from.

Also:
- Delete useless code
- Improve code

### DIFF
--- a/static/js/components/allocation/AllocationTable.jsx
+++ b/static/js/components/allocation/AllocationTable.jsx
@@ -80,15 +80,15 @@ const getMuiTheme = (theme) =>
   });
 
 const AllocationTable = ({
-  groups,
+  groups = [],
   allocations,
   telescopes,
   instruments,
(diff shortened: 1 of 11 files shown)


Theodlz (approved):
that's the kind of addition/deletion ratio I'm a huge fan of 😉.

LGTM

PR #5517, opened 2025-03-25
[WIP] Facility Queue + Notification service: dedup

Theodlz:
For some reason updating the status of facility transactions in the dedicated `facility queue` service seems to keep triggering notifications. What's odd is that we only want notifications on insert for facility transactions and we don't create any of those there, and we have if statements to avoid updating a follow-up requests again and again with the same status, which would cascade in having a lot of notifications being sent. However we still get a lot of notifications. I couldn't find the root cause of that issue, but at least in this PR we:

- avoid calling session.add() on facility transactions and followup requests in the microservice, which was unnecessary thanks to recent changes
- just like we already avoid sending notifications for submitted requests, we avoid sending notifications for requests when they just get their status updated to any kind of `In progress:`-like string which isn't interesting for the user.


Theodlz:
@antoine-le-calloch oups wait, this is showing more changes than it should. Please wait before you review

Theodlz:
Ok, should be good now!

mcoughlin (approved):
Looks good.

Theodlz:
I unfortunately need to re-add the session.add as it looks like some changes are ignored otherwise, maybe some issue due to the relationships of the follow-up request with the transactions that overwrites the changes we make to the transactions when not added to the session (that they are already in though).

PR #5519, opened 2025-03-25
Fix tarot small issues

antoine-le-calloch:
- Fix observing time logic to handle when the target set before the observation time but not the night. The observing time is then recalculate.
- Add altdata key verification on tarot
- Rename altdata stuff
- Fix status url


mcoughlin (approved):
LGTM

Theodlz (approved):
LGTM

Issue #5516, opened 2025-03-24
Classification duplicate on the API scanning return


antoine-le-calloch:
Candidates retrieve when scanning can have a multiple of duplicate classification. The duplication is not visible on the frontend but still take place in the redux.

PR #5520, opened 2025-03-26
Improve candidate handler, deprecate unused parameters

antoine-le-calloch:
- [x] Fix classification duplication when retrieve candidates #5516
- [x] Add more parameters to add flexible retrieving of more data about the object of each candidate
     - This feature will be use on the mobile platform to display more information about a candidate without the need of specific source page

**Also:**
- Delete some useless code see during the development
- Improve candidate handler readability and delete duplicate code


antoine-le-calloch:
There is this code in the candidate get function:

            if annotation_exclude_origin is not None:
                if annotation_exclude_date is None:
                    right = (
                        Obj.select(session.user_or_token, columns=[Obj.id])
                        .join(Annotation)
                        .where(Annotation.origin == annotation_exclude_origin)
                        .subquery()
                    )
                else:
                    expire_date = arrow.get(annotation_exclude_date).datetime
                    right = (
                        Obj.select(session.user_or_token, columns=[Obj.id])
                        .join(Annotation)
                        .where(
                            Annotation.origin == annotation_exclude_origin,
                            Annotation.modified >= expire_date,
                        )
                        .subquery()
                    )

                q = q.outerjoin(right, Obj.id == right.c.id).where(right.c.id.is_(None))
                
This code allow to filter candidate using the annotation_exclude_origin parameters (annotationExcludeOrigin from query), but this parameter is never used or set in the app. The only place I see it is in the test folder. Do we really need this ? Something to be add in a future PR maybe ? The code is 4 years old so maybe something not anymore needed.

skyportal/tests/frontend/sources_and_observingruns_etc/test_observing_planner.py
```diff
@@ -198,9 +198,9 @@ def test_add_run_to_observing_run_page(
 
     ActionChains(driver).click(pi_element).pause(1).send_keys(pi_name).pause(
         1
-    ).send_keys(Keys.TAB).send_keys(calendar_keys).pause(1).send_keys(Keys.TAB).pause(
+    ).send_keys(Keys.TAB).send_keys(Keys.BACKSPACE).send_keys(calendar_keys).pause(
```
Theodlz, line 201: thanks for the quick fix!

skyportal/handlers/api/candidate/candidate_filter.py
```diff
@@ -32,44 +33,47 @@ def get_subquery_for_saved_status(
...
             )
             not_in = True
         elif saved_status == "notSavedToAnySelected":
-            source_subquery = source_subquery.where(Source.group_id.in_(group_ids))
+            subquery = active_sources.where(Source.group_id.in_(group_ids))
             not_in = True
-        elif saved_status == "notSavedToAllSelected":
-            source_subquery = (
-                source_subquery.where(Source.group_id.in_(group_ids))
+        else:  # notSavedToAllSelected
```
Theodlz, line 62: let's avoid that, we don't want to trigger it if it's any other string right?

Theodlz: maybe we should throw and error if we get to an else actually, like if the saved_status value specified isn't valid.

skyportal/handlers/api/candidate/candidate_filter.py
```diff
@@ -32,44 +33,47 @@ def get_subquery_for_saved_status(
...
+            subquery = (
+                active_sources.where(Source.group_id.in_(group_ids))
                 .group_by(Source.obj_id)
                 .having(func.count(Source.group_id) == len(group_ids))
             )
             not_in = True
 
         return (
-            stmt.where(Candidate.obj_id.notin_(source_subquery))
+            stmt.where(Obj.id.notin_(subquery))
```
Theodlz, line 74: how does that impact the resulting query generated by the ORM? Isn't this subquery applied to the Candidate's obj_id? Why would we filter on the `Obj` class?

skyportal/handlers/api/candidate/candidate.py
```diff
@@ -231,13 +228,24 @@ def include_requested_obj_data(
 
     if get_query_argument("includeComments", False):
         candidate["comments"] = sorted(
-            fetch_obj_data(Comment, [], obj_id, session),
+            fetch_obj_data(Comment, [joinedload(Comment.author)], obj_id, session),
```
Theodlz, line 231: do we really need to return the author? I think you're doing that for the mobile app and it's probably fine. Just pointing out that adding data in data has been making the app slower in general, and that if you have the list of users somewhere else already you can easily match by ID. But if you don't, I'm fine adding it.

skyportal/handlers/api/candidate/candidate.py
```diff
@@ -231,13 +228,24 @@ def include_requested_obj_data(
...
             key=lambda x: x.created_at,
             reverse=True,
         )
     if get_query_argument("includeFollowupRequests", False):
         candidate["followup_requests"] = fetch_obj_data(
-            FollowupRequest, [], obj_id, session
+            FollowupRequest,
```
Theodlz, line 237: similar comment here, will you need all of this? Why not load the allocation and instrument data when hydrating just like we do for the we app? That prevents a lot of duplication.

skyportal/handlers/api/candidate/candidate.py
```diff
@@ -853,22 +861,26 @@ def get(self, obj_id=None):
             ]
 
             if isinstance(group_ids, str):
```
Theodlz, line 863: since filters are subsets of groups, shouldn't we first check if filter_ids are specified and if so use those, and THEN check if groups are specified and if so get all filters for those?

Theodlz: This is just a comment about the logic but let's keep it as is since that's how it's implemented. Maybe worth opening an issue for this one. Anyways let's just not make that change in that PR.

Theodlz (commented):
I left some comments, a bit worried about some changes. Would you mind copy/pasting the unit test fix to a separate PR so we can merge it ASAP while this one is being reviewed? thanks

skyportal/handlers/api/candidate/candidate_filter.py
antoine-le-calloch: No no its not possible, everyone of this else if are already in a if testing for valid options. So this else didn't change anything in the logic, its just to stop IDEs putting red everywhere ahah.

skyportal/handlers/api/candidate/candidate_filter.py
antoine-le-calloch: Just keep the logic as it was, the code in production use this for CandidateHandler and use Candidate.obj_id in CandidateFilterHandler. So I just choose to use the actual candidate code to avoid error. But for me yes Candidate.obj_id is the better one.

skyportal/handlers/api/candidate/candidate.py
antoine-le-calloch: Yep you right, lets delete this.

skyportal/handlers/api/candidate/candidate.py
antoine-le-calloch: yes

skyportal/handlers/api/candidate/candidate_filter.py
Theodlz: Actually in production it uses `Obj.id` for the `CandidateHandler`, and `Candidate.obj_id` for the `CandidateFilterHandler`. With your PR it uses `Obj.id` for both. The `CandidateHandler` is the important one so it's best this way, but just checking to see if we are sure that `CandidateFilterHandler` still works with that change. We should at some point try both `Obj.id` and `Candidate.obj_id` locally and print the query that the ORM generates. If it's identical, let's just go for the more elegant `Candidate.obj_id`.

skyportal/handlers/api/candidate/candidate_filter.py
Theodlz: Not sure I understand. The string is already validated before getting to the if/elseif/else statement? Is that what you are saying?

antoine-le-calloch: Yep with this:

            if saved_status in [
                "savedToAllSelected",
                "savedToAnySelected",
                "savedToAnyAccessible",
                "notSavedToAnyAccessible",
                "notSavedToAnySelected",
                "notSavedToAllSelected",
            ]:

Theodlz: Great, is this done in both handlers? I guess my point is that I don't see why handling the invalid string in there wouldn't be a good thing. Now that this method is reusable, it should be clear that an if statement needs to be used elsewhere if the method itself doesn't handle that failure mode.

antoine-le-calloch: Its already done:
            
```
           if q is None:
                return self.error(
                    f"Invalid savedStatus: {saved_status}. Must be one of the enumerated options."
                )
```

antoine-le-calloch: The "if saved_status in" is done in the function use in the two handlers and each handler handle a None return and throw an error.

Theodlz: ok that's fine then, I like having cheap but redundant error handling because then the method is valid everywhere for sure but this is more than enough already.

Theodlz (approved):
LGTM. I'll take another crack at using it from the frontend to see if I spot any issues (as you know candidate scanning is super important so I need to be very careful), and will merge today, ideally deploy tomorrow

antoine-le-calloch:
> LGTM. I'll take another crack at using it from the frontend to see if I spot any issues (as you know candidate scanning is super important so I need to be very careful), and will merge today, ideally deploy tomorrow

Yep sounds good.

PR #5524, opened 2025-03-28
GCN auto-generated sources - configurable SOURCE_RADIUS_THRESHOLD

Theodlz:
No one wants to use the same value for their skyportal and really isn't practical to have to open a PR every time this quantity should change, so instead this PR makes it configurable so it can be customized on a per-instance basis.

### DIFF
--- a/.github/workflows/test_migrations.yaml
+++ b/.github/workflows/test_migrations.yaml
@@ -82,7 +82,7 @@ jobs:
           sudo add-apt-repository ppa:mozillateam/ppa
           printf 'Package: *\nPin: release o=LP-PPA-mozillateam\nPin-Priority: 1001' | sudo tee /etc/apt/preferences.d/mozilla-firefox
 
-          sudo apt install -y wget unzip firefox nginx libcurl4-gnutls-dev libgnutls28-dev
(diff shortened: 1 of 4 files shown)


Theodlz:
> ## Pull Request Overview
> This PR makes the GCN source radius threshold configurable by reading its value from the configuration, allowing customization without code changes.
> 
> * Added comments to document the configuration usage
> * Updated the constant definition from a hardcoded value to a configurable setting
> 
> Files not reviewed (1)
> Comments suppressed due to low confidence (1)
> **skyportal/models/gcn.py:69**
> 
> * Consider validating that the value returned by cfg.get("gcn.source_radius_threshold", 8) is numeric to prevent potential runtime errors during the division.
> 
> ```
> SOURCE_RADIUS_THRESHOLD = cfg.get("gcn.source_radius_threshold", 8) / 60.0
> ```

BTW I do not verify that the value in the config is valid because it's fine to let it yield an error. We don't want the app to start if that value is incorrect anyways

antoine-le-calloch:
Nice, you are quick ahah ! I love our new colleague copilot 😆.

antoine-le-calloch (approved):

Theodlz:
> Nice, you are quick ahah ! I love our new colleague copilot 😆.

yeah! It's a great tool to first get a quick review as it's very good at spotting the obvious, and then getting a proper review from someone is still necessary

PR #5525, opened 2025-03-31
Small improve on scanning report and MMT api

antoine-le-calloch:
- Rename 'filter' to 'filters' in MMT API payload to fit others APIs payloads.
- Add gap between "edit report item" and source name and rename a title to improve clarity.


Theodlz (approved):
LGTM

Issue #5529, opened 2025-03-31
Troubleshooting documentation for failure to start due to sncosmo bandpass problems


rknop:
Please fill out relevant sections below; remove those that are unused.

**Feature Summary**
Based on experimentation between Theo and myself this morning, I suggest adding a troubleshooting section to the deployment docs re: sncosmo failing to download filters.

The cause, we speculate, is sncosmo getting a partial or empty-file download of bandpass files, and then later failing to read them.

The symptom is that various services fail to start, going into a crash-and-restart loop as sncosmo raises an exception trying to read the bad bandpass files.  You see repeated stack dumps in the `app*log` files that end in something like `ValueError: Could not get bandpass for f460m due to sncosmo error: zero-size array to reduction operation maximum which has no identity`, and there are processes on the machine using a lot of CPU as they rapidly go through the restart-and-fail loop.

(I will submit a PR shortly with proposed documentation.)

stefanv:
We can detect this scenario, so why not provide the user with sensible feedback in the log, or try and re-download or work around the issue in some other way?

Theodlz:
We should make it so that we initialize the filters before starting the apps (since concurrency is the issue here), or find a way to only let app 0 auto download sncosmo data.

PR #5530, opened 2025-03-31
Add troubleshooting docs to deployment

rknop:
Issue #5529

I also made some suggested changes earlier in the file -- you don't need to figure out the actual name that docker sees for the running container, as you can just run `docker compose exec`.


Theodlz (commented):
Thank you so much @rknop ! I left a couple of minor comments

PR #5537, opened 2025-03-31
Fix frontend test error

antoine-le-calloch:
Fix input handling in observing planner test for calendar keys.

[Fix sun time calculation errors, improve tarot API handler and fix calculations error and tests.


Theodlz:
thanks a lot @antoine-le-calloch!

Theodlz (approved):
LGTM, will merge once green

Theodlz:
I wonder if this is somehow just failing when we are the 31, i.e. the last day of the month.

antoine-le-calloch:
> I wonder if this is somehow just failing when we are the 31, i.e. the last day of the month.

humm maybe, I don't understand why, because we don't use the actual time.

PR #5558, opened 2025-04-01
Parameterize Dockerfile to allow specification of image name, uid, gid

rknop:
The uid/gid change is one I had to make to install on my kubernetes cluster to allow me to use bind-mounted directories.  The image name change was just for convenience (allows me to skip a `docker tag` step).

I broke the `useradd` out from the long `RUN` that included `apt-get` because making it a separate line doesn't appreciably affect size, but does allow skipping all the `apt-get` when rebuilding the dockerfile after changing those arguments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,12 @@ RUN apt-get update && \
     curl https://sh.rustup.rs -sSf | sh -s -- -y && \
     apt-get update && \
     apt-get install -y cargo nodejs nginx libnginx-mod-http-brotli-static libnginx-mod-http-brotli-filter && \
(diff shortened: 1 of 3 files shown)


Theodlz:
@rknop could these changes create issues with existing k8 deployments?

rknop:
@Theodlz I don't think so; it's set up so the defaults are as before.  To get changed behavior, you have to specify the new arguments.

Theodlz:
> @Theodlz I don't think so; it's set up so the defaults are as before. To get changed behavior, you have to specify the new arguments.

Perfect, i didn't know what were said defaults but if it's identical then this is perfect.

PR #5559, opened 2025-04-02
Remove multiple warnings

Theodlz:
In this PR, we:

- remove numerous warnings that show up as the apps are starting up, which makes it hard to read through the logs. Most of these (if not all), are sqlalchemy-related. Easily solved by removed unused and duplicated relationships, specifying that queries are subqueries, and avoiding having multiple python classes with the same name (even though they are renamed with their `__name__` attribute already).
- We address some pandas related warnings that show up in great numbers when adding photometry to the database.
- We avoid some logs that can be misleading or redundant.

I'm opening this PR as I noticed how confusing these warnings can be for folks that are trying to get SP up and running (recently Robert with the LS4 instance for example).


skyportal/handlers/api/photometry.py
```diff
@@ -193,6 +193,16 @@ def get_filters_mapper(photometry):
 }
 
 
+# this is a drop-in replacement for pandas pd.to_numeric
+# which now does not offer the errors='ignore' option anymore
+def to_numeric(col):
+    """Convert a column to numeric, ignoring errors."""
+    try:
+        return pd.to_numeric(col)
+    except Exception:
```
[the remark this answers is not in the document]
Theodlz: Ok I checked the pandas implementation and they do only ignore these errors, so that's a good idea.

services/thumbnail_queue/thumbnail_queue.py
```diff
@@ -72,7 +72,12 @@ def fetch_obj(session):
 
 @check_loaded(logger=log)
 def service(*args, **kwargs):
+    # start a timer we'll use to have a heartbeat every 60 seconds
+    heartbeat = time.time()
     while True:
+        if time.time() - heartbeat > 60:
+            heartbeat = time.time()
+            log("Thumbnail queue heartbeat.")
```
mcoughlin, line 80: To the beat of the drum

mcoughlin (approved):
LGTM

services/thumbnail_queue/thumbnail_queue.py
```diff
@@ -81,9 +86,6 @@ def service(*args, **kwargs):
                     log(f"Error fetching object with missing thumbnails: {str(err)}")
                     time.sleep(1)
                 elif obj is None:
-                    log(
-                        "No objects with missing thumbnails found, sleeping for 5 seconds."
-                    )
                     time.sleep(5)
```
antoine-le-calloch, line 89: Niceee, finally this log will be gone.

antoine-le-calloch (approved):

services/thumbnail_queue/thumbnail_queue.py
Theodlz: 🥁🥁🥁🥁🥁🥁🥁🥁🥁🥁

Theodlz:
@antoine-le-calloch @mcoughlin I had to use some dark magic to get the migrations to work just fine, it's a minor change but feel free to take a look to see if the comment I left there made sense. But basically with the previous code somehow that made it ignore the index specified on the created_at column by the `Base` class (and so when this code was added, I manually added my own index with a different name), and these changes make things work as expected, which is good except that now by default it tries to create the index which is basically a duplicate of the existing one but with another name. This minor change avoids it from using the index from the base class.

Theodlz:
Quite frankly one might even ask if we needed the base class. It's nice because it gives us the id column but it also gave us the created_at and modified columns that honestly are useless for that table, other than using disk space. We might want to drop those entirely in a future PR, but that's a low priority item.

antoine-le-calloch:
> Quite frankly one might even ask if we needed the base class. It's nice because it gives us the id column but it also gave us the created_at and modified columns that honestly are useless for that table, other than using disk space. We might want to drop those entirely in a future PR, but that's a low priority item.

Yep I agree.

PR #5562, opened 2025-04-04
Follow-up requests Frontend - fixes to API calls that return files

Theodlz:
In this PR we:
- add a API.DOWNLOAD method for the frontend, that downloads files if successful, and shows error message from structured response if status isn't 200
- add methods in the ducks for follow-up requests, to download schedule and allocation analysis report
- make use of these methods in the frontend component for the follow-up request page

### DIFF
--- a/static/js/API.js
+++ b/static/js/API.js
@@ -125,4 +125,55 @@ function DELETE(endpoint, actionType, payload) {
   return API(endpoint, actionType, "DELETE", payload);
 }
 
-export { GET, POST, PUT, PATCH, DELETE, API, API_CALL };
+function DOWNLOAD(endpoint, actionType, payload) {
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
Thanks for enabling this

Theodlz:
@mcoughlin one thing I removed here but could add back, is applying to the filters used when querying for follow-up requests in the frontend form. Right now it just gets the schedule for the instrument, but it doesn't apply any start date, end date, status, ... constraints. I feel like start and end dates could be relevant? But anyways I suppose that the schedule download should - if it doesn't - look for things where the end date and / start date aren't in the past. The start and end date parameters in the form are somewhat arbitrary value, so maybe not ideal to use them?

PR #5564, opened 2025-04-11
Fix ps1 legacy link

antoine-le-calloch:
Fix link placement in Thumbnail component to still work even when there is the crosshairs in front.


Theodlz (approved):
LGTM

PR #5565, opened 2025-04-13
Improvement on MMT API

antoine-le-calloch:
1. The Mask ID field should be automatically set based on the slit width and observation type. [See here for the mapping](https://github.com/SAGUARO-MMA/PyMMT/blob/a91f984ef9865f2ae20bfb7b1d00ef1863fcdc95/src/pymmt/pymmt.py#L145). This probably doesn’t need to be visible on the frontend, I have never needed to configure this and it’s not visible on the MMT front end.

2. The “Photometric” checkbox should instead be called “Require photometric conditions"

3. Because there are so many options for configuring Binospec, it would be great to have the default options set to what we (transient astronomers) typically use. If that’s okay, here are those values:
	Observation type: spectroscopy
	Slit Width: Longslit1
	Filter: LP3800
	Grating: 270
	Central wavelength: 6500

4. Your field called “Number of visits per night” should be a checkbox called “One visit per night” which is checked by default - and strictly speaking it only needs to be visible if visits is > 1, but that’s less important.


mcoughlin (approved):
Looking forward to seeing this in production!

skyportal/facility_apis/mmt/mmt_utils.py
```diff
@@ -81,30 +85,45 @@ def check_obj_for_mmt(obj):
         raise ValueError("Missing the 'magnitude' value on the object")
 
 
+def bool_to_int(value):
```
Theodlz, line 88: could be in your new parsing.py file no?

Theodlz (approved):
LGTM! Just one small comment

antoine-le-calloch:
@Theodlz Done !

antoine-le-calloch:
> > @Theodlz Done !
> 
> it looks a little weird, I see you moved it to another file but I don't see you importing it where you are calling it!

My bad.

Issue #5531, opened 2025-03-31
Source Tags


Theodlz:
**Feature Summary**
We've added the concept of tags to GCN events, and similarly it would be very useful if we could add tags to sources. Some instances have relied on classifications for that, but well tags do not have associated probability and can be free form text, whereas classifications require a fixed taxonomy and could confuse other systems that do rely on those.

**Usage / behavior**
We want to have a "tags" section in the source page header/info card, where we can add/edit/delete tags.

**Alternative Solutions**

**Implementation details**
This means first opening a PR to add an `ObjTag` table with `text, obj_id, author_id` columns, and a `ObjTagsHandler` (in the skyportal/handlers/api/obj.py or a dedicated file) that implements POST, PATCH, DELETE. We can simply retrieve source(s) tags along with the source(s) when using the `SourcesHandler` (that could be added in a separate PR). Last but not least we should edit the frontend for the source page to display said tags, with add/edit/delete options.

Theodlz:
Hi @douzetc! That cookbook is referring to GCN event tags, not tags on sources! This is precisely what this issue is about, we want to add tags to the sources too.

JulienPeloton:
Ah gotcha -- my bad, we read it quickly, but the issue is clear to me now 👍

Theodlz:
As discussed this morning with Camille @JulienPeloton, best would be to not just have a simple table with object id + tag as string, because in a large instance with lots of sources you'd have tons of them and retrieving the list of all unique tags would get expensive (and also it would quite a bit of space). Instead we suggest:
- A table called "ObjTagOptions", with an id, and a tag as string. Optional adding tag options to that list could be limited to users with a permission to manage tags. That way we avoid folks typing the same thing differently and such.
- A table called "ObjTags", which has object id + tag id (+ optionally author id if we need that), which is an N to N relationship between the Obj table and the ObjTagOptions table. That way we don't duplicate a ton of string but cheaper integer-like ids, and finding all unique tables is as simple as retrieving the ObjTagOptions and no query over millions of rows is necessary. But by using a table rather than an enum it can be fuly customized by instances and doesn't even require config changes.

mcoughlin:
Complete.

PR #5566, opened 2025-04-14
feat: adding source tags #5531 (backend)

camilledzt:
This scope of this pull request is :
* Adding two tables to the database to handle tags properly (meaning one table for the tag options and one table for the association tag/source).
* Adding two endpoints to the API in order to handle tags.


skyportal/models/tag.py
```diff
@@ -0,0 +1,18 @@
...
+import sqlalchemy as sa
+
+from baselayer.app.models import (
+    Base,
+    join_model,
+)
+from skyportal.models.source import Source
+
+
+class ObjTagOption(Base):
```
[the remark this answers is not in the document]
Theodlz: I think we discussed the possibility of adding an `author_id` here like we do for a number of tables. Any thoughts @mcoughlin, doy ou think keeping track of who added a tag to an object would be useful?

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,273 @@
+from baselayer.app.access import auth_or_token
+
+from ...models import ObjTagOption, ObjTags, Source
+from ..base import BaseHandler
+
+
+class ObjTagOptionHandler(BaseHandler):
+    @auth_or_token
+    def get(self, tag_identifier=None):
```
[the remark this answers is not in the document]
camilledzt: I wanted to separate the tag_id (in the table) and the concept of tag_identifier (to allow users to search tag with different type of information). I think this is linked to this conversation (https://github.com/skyportal/skyportal/pull/5566#discussion_r2042377395).

I will change it if you think we should remove research by id and names.

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,273 @@
...
+        with self.Session() as session:
+            if tag_identifier is None:
+                tags = session.scalars(ObjTagOption.select(session.user_or_token)).all()
+
+                return self.success(
+                    data=[{"id": tag.id, "tag_name": tag.tag_name} for tag in tags]
+                )
+
+            elif tag_identifier.isdigit():
```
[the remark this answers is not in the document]
camilledzt: My first reason to implement this is for testing purpose. For example, in my unit test when I test the delete function, then I submit a GET request to be sure that the tag doesn't exist anymore. Maybe I can simply do the request twice to verify this point (by getting an error 404 on the second request). 

My second thought was that it was maybe useful if I have a search bar in the tag section (to search a tag by name in this case).

What do you think ?

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,273 @@
...
+                return self.error("Tag already exists", status=404)
+
+            new_tag = ObjTagOption(tag_name=tag_text)
+            session.add(new_tag)
+            session.commit()
+
+            return self.success(data={"id": new_tag.id, "tag_name": new_tag.tag_name})
+
+    @auth_or_token
+    def put(self, tag_id):
```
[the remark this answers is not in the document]
camilledzt: Do you think I should implements a real put or is it ok like that ?

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,273 @@
...
+            return self.success(
+                data={
+                    "id": new_assoc.id,
+                    "objtagoption_id": new_assoc.objtagoption_id,
+                    "source_id": new_assoc.source_id,
+                }
+            )
+
+    @auth_or_token
+    def put(self, association_id):
```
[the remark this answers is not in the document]
camilledzt: I was thinking of the case that an admin wanted to modify tags in order to match a homogeneous way to create tags (maybe space instead of _ for example)

skyportal/models/tag.py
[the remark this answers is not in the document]
camilledzt: Reading the issue, I was thinking we wanted to associate tags to sources and not obj, I will change it, sorry

skyportal/tests/api/test_tag.py
```diff
@@ -0,0 +1,210 @@
...
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from skyportal.models import ObjTagOption
+from skyportal.tests import api
+
+
+# --- Testing ObjTagOption API
+def test_get_tag(super_admin_token):
+    tag_to_create = {"tag_name": f"test_get_{random.randint(0, 100000)}"}
```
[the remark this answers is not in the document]
Theodlz: https://github.com/skyportal/skyportal/blob/65183ff33bd3f5e2ad6dc5894effb98117208c57/skyportal/tests/api/test_shifts.py#L8

camilledzt:
#5531

skyportal/handlers/api/tag.py
```diff
@@ -51,13 +23,24 @@ def post(self):
         if not tag_text or not isinstance(tag_text, str):
             return self.error("`tag_name` must be provided as a non-empty string")
 
+        if not re.fullmatch(r"[A-Za-z0-9]+", tag_text):
```
Theodlz, line 26: Can't remember what we landed on but I think this is a bit different. What I suggested what to string the string of special characters, spaces, casing, ... and such **when checking for duplication**, not completely banning them.

It's better to be strict now and loosen it up later though. So let's keep this and not change anything, and we'll think about this more thoroughly for another PR. Sounds good?

skyportal/handlers/api/tag.py
```diff
@@ -51,13 +23,24 @@ def post(self):
...
+            return self.error(
+                "`tag_name` must contain only letters and numbers (no spaces, underscores, or special characters)",
```
Theodlz, line 28: I see the parameter is called `tag_name`, but in your code the variable is called `tag_text`. I believe that a comment left on the previous PR was that having `tag_...` as a naming scheme in the model was not the best since well it's on a tag-related table so we know it's a tag, we don't need that in the column names.

I don't think you changed that in the model or here, could you look into it? Just calling everything `text` seems more consistent and would make most sense for the data model.

skyportal/handlers/api/tag.py
```diff
@@ -167,34 +151,47 @@ def post(self):
...
+                user_id = self.current_user.id
 
-            new_assoc = ObjTags(objtagoption_id=objtagoption_id, source_id=source_id)
+            new_assoc = ObjTags(
+                objtagoption_id=objtagoption_id, obj_id=obj_id, user_id=user_id
+            )
             session.add(new_assoc)
             session.commit()
 
             return self.success(
```
Theodlz, line 171: I've put that comment in the previous review,  but I see you create the `data` in a lot of handlers when you could just send the python object from the db (e.g. `new_assoc`) which gets auto converted to dict/JSON.

skyportal/models/tag.py
```diff
@@ -14,5 +12,22 @@ class ObjTagOption(Base):
     tag_name = sa.Column(sa.String, nullable=False, doc="Tags list.", unique=True)
 
 
-ObjTags = join_model("obj_tags", ObjTagOption, Source)
+ObjTags = join_model("obj_tags", ObjTagOption, Obj)
```
Theodlz, line 15: please drop the `s`. As you noticed the table names do not have an `s` in their name. This is because the python name is lowercased and appended an `s` when autogenerating the postgres table names.

skyportal/models/tag.py
```diff
@@ -14,5 +12,22 @@ class ObjTagOption(Base):
...
+
+read = update = delete = AccessibleIfUserMatches("user")
+
+ObjTags.user_id = sa.Column(
```
Theodlz, line 19: I'm fine with that, though do keep in mind that for most tables where a user is associated to the record and it's somewhat text related (comment, classification, ...) we usually call it `author_id`. Consistency in coding is good in general because when working on any feature, if its inner workings and field names are similar to other similar concepts, it's easier for the developer.

skyportal/tests/api/test_tag.py
```diff
@@ -9,90 +9,102 @@
...
+        "POST",
+        "objtagoption",
+        data={"tag_name": f"tagadded{unique_id}"},
+        token=super_admin_token,
+    )
+    assert status == 409
 
-    status_by_name, _ = api("GET", f"objtagoption/bonjour", token=super_admin_token)
-    assert status_by_name == 404
+    # Verification that we can't create a tag with a space
```
Theodlz, line 35: Thanks for being so thorough! May I suggest a shorter way to write your tests though? If you have `N` number of characters you want to test, you might as well just have a list/array with those, and do your POST + assert in a loop over that array. That would make your test **much** easier to read and shorter, hence easier to maintain.

skyportal/tests/api/test_tag.py
```diff
@@ -0,0 +1,210 @@
...
+        data=update_data,
+        token=super_admin_token,
+    )
+    assert status == 409
+    assert "already exists" in data["message"]
+
+
+def test_delete_association(super_admin_token):
+    tag_data = {"tag_name": f"tag_delete_association_{random.randint(0, 100000)}"}
+    _, tag = api("POST", "objtagoption", data=tag_data, token=super_admin_token)
```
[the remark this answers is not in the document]
Theodlz: @camilledzt I still don't see the status check I believe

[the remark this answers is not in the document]
Theodlz: here and below

Theodlz (changes requested):
Hi @camilledzt, thanks for addressing some of the previous comments. I added comments about other elements, and also noted that a number of comments from the previous PR are not addressed yet. Please do make sure to either address comments or answer to them (if you believe they are not correct) before requested for another review. Or you can always request for one but then just put a comment tagging the reviewer and explaining that some things aren't implemented yet and that's to be expected. Otherwise, it's a bit confusing!

skyportal/models/tag.py
```diff
@@ -9,25 +9,25 @@
 class ObjTagOption(Base):
     """A record of a list of tags that are available."""
 
-    tag_name = sa.Column(sa.String, nullable=False, doc="Tags list.", unique=True)
+    name = sa.Column(sa.String, nullable=False, doc="Tags list.", unique=True)
 
 
-ObjTags = join_model("obj_tags", ObjTagOption, Obj)
+ObjTags = join_model("obj_tag", ObjTagOption, Obj)
```
Theodlz, line 15: Oh I meant the Python class mostly. It's possible that you need to write the name manually in that `join_model` method, so adding an `s` in there is fine in that case. I meant that the python class should not have an `s` as table names which are inferred from python classes use the class name + s. Not done anyway in your case since you specify the name explicitely in `join_model(...)` but even then let's not have the `s` in the python class name which would different from all other tables in the codebase = confusion.

skyportal/tests/api/test_tag.py
```diff
@@ -9,50 +10,32 @@
 
 # --- Testing ObjTagOption API
 def test_get_tag(super_admin_token):
-    tag_to_create = {"tag_name": f"TestTag{uuid.uuid4().hex}"}
+    tag_to_create = {"name": f"TestTag{uuid.uuid4().hex}"}
```
Theodlz, line 13: out of curiosity why do you convert to `.hex` rather than just use the uuid? Is that to not have the `-` in there?

skyportal/tests/api/test_tag.py
```diff
@@ -9,50 +10,32 @@
...
-    assert status == 409
-
-    # Verification that we can't create a tag with a space
-    status, _ = api(
-        "POST", "objtagoption", data={"tag_name": f"Tag added"}, token=super_admin_token
-    )
-    assert status == 400
-
-    # Verification that we can't create a tag with an underscore
+@pytest.mark.parametrize(
```
Theodlz, line 21: interesting, I was suggesting a simple loop but if that works, why not!

Theodlz (commented):
This is already much better @camilledzt, I quickly reviewed it now but will do a more extensive review tomorrow morning PST

skyportal/tests/api/test_tag.py
camilledzt: Yes it's to remove the ` - ` :)

skyportal/tests/api/test_tag.py
camilledzt: Yes, your suggestion reminds me this function and I thought that it fit the need well

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
...
+
+class ObjTagOption(Base):
+    """A record of a list of tags that are available."""
+
+    name = sa.Column(sa.String, nullable=False, doc="Tags list.", unique=True)
+
+
+ObjTag = join_model("obj_tag", ObjTagOption, Obj)
+
+read = update = delete = AccessibleIfUserMatches("user")
```
Theodlz, line 17: this doesn't do anything just written in the middle of a file. It's supposed to be fields of a python class you're applying this to. Right now it's just creating unused variables. Let's just remove it and worry about data access later.

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
...
+
+from baselayer.app.models import AccessibleIfUserMatches, Base, join_model
+from skyportal.models import User
+from skyportal.models.obj import Obj
+
+
+class ObjTagOption(Base):
+    """A record of a list of tags that are available."""
+
+    name = sa.Column(sa.String, nullable=False, doc="Tags list.", unique=True)
```
Theodlz, line 12: doc is incorrect, this is not a list

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
+import sqlalchemy as sa
+from sqlalchemy.orm import relationship
+
+from baselayer.app.models import AccessibleIfUserMatches, Base, join_model
+from skyportal.models import User
+from skyportal.models.obj import Obj
+
+
+class ObjTagOption(Base):
+    """A record of a list of tags that are available."""
```
Theodlz, line 10: same this here, not a list

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
...
+
+    name = sa.Column(sa.String, nullable=False, doc="Tags list.", unique=True)
+
+
+ObjTag = join_model("obj_tag", ObjTagOption, Obj)
```
Theodlz, line 15: did you have a chance to check in which columns the resulting columns are? not that it matters too much but having obj_id first and then the tag_id second might be more logical?

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
...
+
+read = update = delete = AccessibleIfUserMatches("user")
+
+ObjTag.author_id = sa.Column(
+    "author_id",
+    sa.ForeignKey("users.id", ondelete="SET NULL"),
```
Theodlz, line 21: why this specific on delete? I think that conflicts with your nullable=True.

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
...
+    nullable=True,
+    index=True,
```
Theodlz, line 23: I really don't think we need an index on that one, I doubt we'll ever query by who indexed something.

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
...
+    doc="ID of the user who created the tag association",
+)
+
+ObjTag.author = relationship(
+    User,
+    doc="The associated User",
+    foreign_keys=[ObjTag.author_id],
+)
+
+ObjTag.__table_args__ = (sa.Index("obj_tags", "name", "tag_id", unique=True),)
```
Theodlz, line 33: I don't think I understand this index here, I mean the `ObjTag` doens't even have a `name` field right? That's on the `ObjTagOption` table. Let's just remove this

skyportal/models/tag.py
```diff
@@ -0,0 +1,33 @@
...
+
+ObjTag.author_id = sa.Column(
+    "author_id",
+    sa.ForeignKey("users.id", ondelete="SET NULL"),
+    nullable=True,
+    index=True,
+    doc="ID of the user who created the tag association",
+)
+
+ObjTag.author = relationship(
```
Theodlz, line 27: This syntax looks a bit different from other tables. We usually don't need to specify any foreign key as they are inferred from the author_id field you created above. Is that some exception due to the fact that we are adding these *after* creating the class? Let's double check, if this different syntax isn't required let's keep it consistent with the rest, such as: https://github.com/skyportal/skyportal/blob/main/skyportal/models/classification.py#L57-L63

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,238 @@
...
+
+        with self.Session() as session:
+            existing_tag = session.scalars(
+                ObjTagOption.select(session.user_or_token).where(
+                    func.lower(ObjTagOption.name) == name.lower()
+                )
+            ).first()
+
+            if existing_tag:
+                return self.error(
```
Theodlz, line 40: maybe also return the existing tag's name so the user can understand why it says it exists (as in it's identical to another one and providing it is a bit clearer than only saying case insensitive match).

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,238 @@
...
+                )
+
+            new_tag = ObjTagOption(name=name)
+            session.add(new_tag)
+            session.commit()
+
+            return self.success(new_tag)
+
+    @auth_or_token
+    def patch(self, tag_id):
```
Theodlz, line 52: I suggest checking that you can cast the tag_id to an int, to make sure it's a valid ID. This or in `skyportal_app_server.py` ideally your regex would not just match anything like it does now, but also numbers. I suggest doing that actually, dealing with that in the route is best. That said still a good idea to cast to int here otherwise the queries might not work as expected.

skyportal/app_server.py
```diff
@@ -457,6 +459,8 @@ def log_request(self, handler):
         ObservationPlanFieldsHandler,
     ),
     (r"/api/objs(/[0-9A-Za-z-_\.\+]+)", ObjHandler),
+    (r"/api/objtagoption(/.*)?", ObjTagOptionHandler),
```
Theodlz, line 462: for routes that expect numbers only as parameters, we should enforce that so you don't get ids that make no sense. Could just be  something like`(/[0-9]+)?`.

skyportal/handlers/api/tag.py
Theodlz: that comment is valid for all the endpoints btw

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,238 @@
...
+            ).first():
+                return self.error("Specified tag does not exist", status=404)
+
+            # Verify obj exists
+            if not session.scalars(
+                Obj.select(session.user_or_token).where(Obj.id == obj_id)
+            ).first():
+                return self.error("Specified obj does not exist", status=404)
+
+            if hasattr(self.current_user, "created_by"):
```
Theodlz, line 150: I really don't understand that, it seems very convoluted. You can directly access the ID of the user creating the request.

You can use `self.associated_user_object.id`

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,238 @@
...
+            return self.success(new_assoc)
+
+    @auth_or_token
+    def patch(self, association_id):
+        """Update an existing tag-obj association"""
+        data = self.get_json()
+        new_tag_id = data.get("objtagoption_id")
+        new_obj_id = data.get("obj_id")
+
+        if hasattr(self.current_user, "created_by"):
```
Theodlz, line 170: same comment here about the user id. In fact that comment is valid all across that file.

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,238 @@
...
+            new_assoc = ObjTag(
+                objtagoption_id=objtagoption_id, obj_id=obj_id, author_id=author_id
+            )
+            session.add(new_assoc)
+            session.commit()
+
+            return self.success(new_assoc)
+
+    @auth_or_token
+    def patch(self, association_id):
```
Theodlz, line 164: so while I understand why we have a POST/PATCH/DELETE for the tag options (users want to add new tag options, delete them, or edit existing one), I really don't think we need any PATCH/PUT for the tags on an object. Since it's just about associating things anyways from the frontend there will never be a scenario where the user will want to click on a tag to replace it with another. It will always be as simple as removing an unwanted tag to then add another. So let's keep this simple, and just remove the `patch` method for the `ObjTagHandler`.

Theodlz: and just to be extra clear, we do want to keep it for `ObjTagOptionHandler` though

Theodlz (changes requested):
Hi @camilledzt, I left an additional batch of comments. Mostly things to remove actually, not that many things to edit left!

skyportal/models/tag.py
camilledzt: During my tests, I noticed that if I had the `ondelete = CASCADE` option, the association was created then deleted at the end of my tests because the user no longer exists.
I set it to `SET_NULL` so that the association is not deleted if a user is deleted. 

Would you prefer that I set `nullable = False`? This would allow the author_id to be kept in case of deletion. (Ex: the author toto has an id of 7, he is deleted. The record keeps the author_id at 7 and not Null. But we will no longer know who user 7 was so I don't know if this is of any interest?)

skyportal/models/tag.py
camilledzt: Here is an example of the objtag table view :
image

Do you want me to swap the colomns?

camilledzt: Btw, obj_id in the obj table makes reference to a string. Is it ok for you ?

Theodlz: ideally having obj_id first (which is just a matter of doing `join_model("obj_tag", Obj, ObjTagOption)` (see how the 2 classes swapped place) would be best yes.

And yeah the `obj_id` is a string and that's normal, the `id` of the `Obj` table is of type string. It basically represents the name of an object which is its unique identifier, which is why we made that the ID.

skyportal/models/tag.py
Theodlz: I will let you decide. In practice we do not delete users from the database so I'd simply suggest not having `ondelete="SET NULL"` and keeping the `nullable=False`. `ondelete="SET NULL"`  should be a discussion item we have one of these days, to see if we want to start using that pattern for some table. Lacking a user id could disrupt a lot of frontend code and backend logic so it's something to think carefully about.

camilledzt:
I just pushed the requested migration file.

alembic/versions/d2ab4a302b57_adding_new_table_for_the_source_tag_.py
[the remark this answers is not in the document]
Theodlz: So please delete it, and make sure you go through the steps from the documentation thoroughly

alembic/versions/9239216843f5_adding_new_table_for_the_source_tag_.py
Theodlz, line 1: I really suggest installing pre-commit (with `apt install pre-commit` on linux or `pip install pre-commit`) and then running pre-commit install. That ensures you can't even do a commit if stuff isn't formatted properly.

Theodlz:
@camilledzt could you kindly mrege with main and update the `down_revision` in your migration script (that's the ID of the previous revision on main)? Ihad to merge a PR ASAP yesterday due to a time constraint and that including a migration.

Or, usually what we do to keep these things easy is you can give me write access to your fork, and I can sync with main before merging

skyportal/handlers/api/tag.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class ObjTagOptionHandler(BaseHandler):
+    @auth_or_token
+    def get(self):
+        with self.Session() as session:
+            tags = session.scalars(ObjTagOption.select(session.user_or_token)).all()
+            return self.success(data=tags)
+
+    @auth_or_token
```
Theodlz, line 18: didn't you mention that only admins could post tags? Looks like there is no rules enforced here (if you look at other handlers, there is a `@permissions([...])` method that takes the allowed permissions to call a handler.

skyportal/tests/api/test_tag.py
Theodlz, line 1: The tests are super clean btw, very nice job!

skyportal/tests/api/test_tag.py
```diff
@@ -0,0 +1,135 @@
...
+from sqlalchemy.orm import Session
+
+from skyportal.models import ObjTagOption
+from skyportal.tests import api
+
+
+# --- Testing ObjTagOption API
+def test_get_tag(super_admin_token):
+    tag_to_create = {"name": f"TestTag{uuid.uuid4().hex}"}
+    status, _ = api("POST", "objtagoption", data=tag_to_create, token=super_admin_token)
```
Theodlz, line 14: maybe for at least one combination of POST/GET/PATCH/DELETE + endpoint + status) , you might want to check that the content of `data` is valid too, not just the status? doesn't have to be for everything that's really the cherry on top, but at least for the GETs, you should really really check that the content if correct, because well if you get a 200 but the list is empty when it shouldn't, that's an issue! Especially as we add data access and want to test who can see what, that will be absolutely necessary,

Theodlz (commented):
Some minor comments @camilledzt, I'll let you take a look!

skyportal/handlers/api/tag.py
```diff
@@ -15,7 +15,7 @@ def get(self):
             tags = session.scalars(ObjTagOption.select(session.user_or_token)).all()
             return self.success(data=tags)
 
-    @auth_or_token
+    @permissions(["Manage sources"])
```
Theodlz, line 18: That's not admin, but I'm fine with that too

Theodlz (approved):
LGTM 🚀

PR #5567, opened 2025-04-14
Prevent the creation of duplicate report

antoine-le-calloch:


skyportal/handlers/api/candidate/scan_report.py
```diff
@@ -124,6 +124,21 @@ def post(self):
             if not saved_range:
                 return self.error("No saved candidates range provided")
 
+            # Check if this report already exists
+            existing_report = session.scalars(
+                ScanReport.select(session.user_or_token).where(
+                    ScanReport.groups.any(Group.id.in_(group_ids)),
```
Theodlz, line 130: didn't know that `any` operator existing on relationships, that's interesting. I wonder what the resulting query looks like

skyportal/handlers/api/candidate/scan_report.py
```diff
@@ -124,6 +124,21 @@ def post(self):
...
+                    ScanReport.options["passed_filters_range"] == passed_filters_range,
```
Theodlz, line 131: I wonder if this would always work, might be better to access each elements in the ranges, cast to a datetime and check there? Anyways that's a very minor comment you can merge already

Theodlz (approved):
LGTM

skyportal/handlers/api/candidate/scan_report.py
antoine-le-calloch: I have test it and it seems to work well. This feature will no be use a lot because the time is set when opening the modal so its just when people don't close the page and maybe forget that they have already create it and try to do it again.

PR #5570, opened 2025-04-19
Create utils parse function

antoine-le-calloch:
- Add utility functions for parsing integers and pagination handling
- Refactor integer parsing across multiple files to use parsing integers and pagination handling utility function
- Refactor candidate and candidate_filter modules to utilize utility functions for integer parsing and pagination handling and move duplicate code to a new function.

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -144,3 +144,9 @@ It will print a command that you can use to SSH into the runner.
     - A brief summary of the changes, including any breaking changes
     - A list of the changes split in the relevant categories (e.g., new features, bug fixes, frontend, backend, etc.). For each change, include a list of the associated PR numbers, as well as the GitHub usernames of the contributors.
     - If applicable, a migration guide to help users and/or admins upgrade to the new version.
+
+## Developer instructions
+
+There are a few methods in `skyportal/utils/parse.py` that are used to help parse common types of input.
+- `get_int_list`: This method is used to parse a comma-separated string into a list of integers. It can also handle lists and will convert them to a list of integers.
+- `get_page_and_n_per_page`: This method is used to parse the page number and number of items per page used for pagination in multiple handlers. It takes the page number and number of items per page as input and returns them as integers. It also ensures that the number of items per page does not exceed a maximum value (default is 500).
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -148,5 +148,5 @@ It will print a command that you can use to SSH into the runner.
 ## Developer instructions
 
 There are a few methods in `skyportal/utils/parse.py` that are used to help parse common types of input.
-- `get_int_list`: This method is used to parse a comma-separated string into a list of integers. It can also handle lists and will convert them to a list of integers.
+- `get_list_typed`: This method is used to parse a comma-separated string into a list of a specified type. It can also handle a list and will convert it to a list of the specified type.
 - `get_page_and_n_per_page`: This method is used to parse the page number and number of items per page used for pagination in multiple handlers. It takes the page number and number of items per page as input and returns them as integers. It also ensures that the number of items per page does not exceed a maximum value (default is 500).
--- a/skyportal/handlers/api/candidate/candidate.py
+++ b/skyportal/handlers/api/candidate/candidate.py
@@ -845,7 +845,7 @@ class CandidateHandler(BaseHandler):
         with self.Session() as session:
             group_ids, filter_ids = get_user_accessible_group_and_filter_ids(
                 session,
-                self.current_user,
+                session.user_or_token,
                 group_ids,
                 filter_ids,
             )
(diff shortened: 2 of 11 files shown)


skyportal/handlers/api/candidate/candidate.py
```diff
@@ -972,7 +927,7 @@ def get(self, obj_id=None):
                 q,
                 saved_status,
                 group_ids,
-                user_accessible_group_ids,
+                self.current_user,
```
mcoughlin, line 930: Should this be the user_or_object or whatever?

skyportal/handlers/api/candidate/candidate_filter.py
```diff
@@ -99,92 +129,27 @@ def get(self):
...
-        except ValueError:
-            return self.error("Invalid page number value.")
-        try:
-            n_per_page = int(n_per_page)
-        except ValueError:
-            return self.error("Invalid numPerPage value.")
-        n_per_page = min(n_per_page, 500)
+            group_ids, filter_ids = get_user_accessible_group_and_filter_ids(
+                session,
+                self.current_user,
```
mcoughlin, line 134: Same deal here.

skyportal/handlers/api/candidate/candidate_filter.py
```diff
@@ -193,7 +158,7 @@ def get(self):
                 stmt,
                 saved_status,
                 group_ids,
-                user_accessible_group_ids,
+                self.current_user,
```
mcoughlin, line 161: Here too.

mcoughlin (approved):
Looks good to me. Just wondering if we should continue to replace current_user which as I recall behaves strangely with tokens?

antoine-le-calloch:
> Looks good to me. Just wondering if we should continue to replace current_user which as I recall behaves strangely with tokens?

Done !

skyportal/utils/parse.py
```diff
@@ -0,0 +1,26 @@
+def get_int_list(text, error_msg="Error parsing string to int list"):
+    """
+    Convert a comma-separated string to a list of integers.
+    If the input is already a list return it as a list of integers.
+    """
+    try:
+        if isinstance(text, str):
+            return [int(x.strip()) for x in text.split(",") if x.strip()]
+        if isinstance(text, list):
+            return [int(x) for x in text]
```
Theodlz, line 10: missing `strip()` in there?

Theodlz: I think that for a list unless you check the dtype you should cast the elements to str (they could be int already or other!), strip, then to int

Theodlz: I wonder if the method would be even more useful if it worked for lists in general, meaning you pass the expected dtype of the elements as a parameter?

Theodlz (commented):
Super nice. I suppose that this is a none exhaustive list of places that benefit from these new methods right? I think that this should be documented especially until every piece of code has been migrated to use this new syntax. 

Could you check if there is a place in the docs (maybe dev instructions) to document that we should use this method for pagination, and for reading list inputs?

skyportal/utils/parse.py
antoine-le-calloch: But what you want to do with the dtype pass in argument ? verify if type is right and then convert to int ? I don't understand quiet well.

antoine-le-calloch:
@Theodlz  I try to use the get_int_list in every piece of code that was doing the same thing but I didn't do the same for get_page_and_n_per_page (only change code that I see when coding).
I will take a look for the doc.

antoine-le-calloch:
@Theodlz in the https://skyportal.io/docs/adding_features.html page on the backend part will be good ?

Theodlz:
Yes that would be fine. We can revise that later

skyportal/utils/parse.py
Theodlz: No I meant we might want to have list of strings, int, floats... the parameter would be the targeted data type

Theodlz:
@antoine-le-calloch let me know when you want me to review this again

doc/dev.md
Theodlz, line 1: that's great, thanks!

Theodlz (approved):
Very nice! LGTM

skyportal/tests/utils/test_parse.py
```diff
@@ -0,0 +1,49 @@
+from skyportal.utils.parse import get_list_typed, get_page_and_n_per_page
+
+
+def test_get_list_typed():
+    assert get_list_typed("1,2,3", int) == [1, 2, 3]
+    try:
+        assert get_list_typed("1.0,2.0,3.0", int) == [1.0, 2.0, 3.0]
+    except ValueError as e:
+        assert str(e) == "Error parsing string to list of <class 'int'>"
```
Theodlz, line 9: `Error parsing string to list of in` would be nicer in my opinion. If we ever let these errors propagate back to the API response (which we might already do?), that doesn't look great for the client. I think you should be able to print data types by just their name.

Theodlz (commented):
Just one minor comment about the error formatting

skyportal/tests/utils/test_parse.py
antoine-le-calloch: yep for sure.

Theodlz (approved):
LGTM

Issue #5571, opened 2025-04-21
Swift UVOT Request Confirmation


wynnjacobson-galan:
Could we add in a confirmation page feature for Swift UVOT requests on Fritz so that the user can review their submission before it is sent to the Swift team?

Theodlz:
@wynnjacobson-galan thanks for that suggestion. What we'll do is that basically for all instrument APIs that do not implement any `delete` method (as in, once submitted you can't delete an accidental request, and that includes Swift), we'll show some popup to validate the submission after hitting `submit`.

Theodlz:
Addressed in https://github.com/skyportal/skyportal/pull/5572

Theodlz:
t

PR #5572, opened 2025-04-23
Follow-up request frontend: confirmation dialog if delete isn't implemented

Theodlz:
If an allocation's API does not implement a delete method, this PR adds a confirmation dialog that shows up after clicking `submit` on the form, where the user has to click `confirm` to go ahead and submit, or `cancel` to, well, cancel.

image

Addresses #5571


mcoughlin:
@Theodlz Thanks!!!

antoine-le-calloch (approved):

PR #5577, opened 2025-04-28
Photometry upload/update: less queries

Theodlz:
Right now in the `add_external_photometry` method which takes a dataframe of photometry rows as input, we run 2 queries:
- the first looks for which rows are new
- the second looks for which rows are duplicates (already exist)

when instead we could just look for which are duplicates, and get the new ones in python just by doing a simple diff between the idx of the duplicates and the new ones in the dataframe, indexes which we get back when we run the query.

I suspect that this can be applied in other places. It does speed things up quite a bit (between 25% and 50%) when it comes to adding the new rows in the DB. This is a simple but really appreciated speed boost when uploading data in bulk


mcoughlin (approved):
LGTM

PR #5579, opened 2025-04-28
Parsing utils: str to bool

Theodlz:
This PR adds a function to parse a string (or anything that can be converted to string) to a bool. If accepts a variety of string-like representations for True or False, and can be given a `default` (which is set to None if not specified in which case it yields an error if the input does not match a valid True/False value) to return if there is no match to True or False.


mcoughlin (approved):
Clearly should be in the standard python lib ;)

Theodlz:
@antoine-le-calloch tagging you too in case you want to take a look, since you added this parsing utils yourself recently.

PR #5580, opened 2025-04-28
Instrument field data post/put: delimiter check

Theodlz:
when looping over delimiters, don't try loading df with pandas if the delimiter does not even appear in the string. This is required because we try a comma and then a space, but even if there is no comma pandas will successfully read the df and the whole row (regardless of the number of columns) will just be assumed to be one col. 

This PR simply makes sure that we don't even attempt to read it with a certain delimiter if it's char isn't even found in the data string.


mcoughlin (approved):
LGTM

PR #5582, opened 2025-04-29
LocalizationTiles: additional partitions until end of 2027

Theodlz:
This PR migrates the DB to add new localizationtiles partitions to take us to `Dec 2027` (so last partition spans from 2027-12-01` to `2028-01-01`).

For reference: I simply looked at the additional migrations and extracted all the operations required to register a partition and unregister it (for both the update and downgrade scripts). Then instead of hardcoding everything like I did back then, I made it a loop so the migration is much easier to read.

Also, I added a `VACUUM ANALYZE instrumentfieltiles` query at the end of the `add_tiles` method we call when inserting new fields for an instrument. Without that, queries relies on that table might yield really bad query plans in PostgreSQL (queries will then be too slow to execute).


Theodlz:
oups I added some files by mistake! Let me remove that

mcoughlin (approved):
Hopefully this means the queries will work for a while longer ;)

Theodlz:
> Hopefully this means the queries will work for a while longer ;)

yep, and what's nice is that this PR can be used as a reference point to just add more migrations later (pretty much copy/pasted). I'll try to add a link to it in the docs at some point.

alembic/versions/5f949cc5cae9_localizationtiles_partitions_2027.py
```diff
@@ -0,0 +1,51 @@
...
+# revision identifiers, used by Alembic.
+revision = "5f949cc5cae9"
+down_revision = "7eab209f29d3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    for year in range(2025, 2028):
+        for month in range(1 if year != 2025 else 5, 13):
```
JulienPeloton, line 24: This line above is not correct if the patch is ran after April, with `localizationtiles_<date_after_april>` already existing. It will throw the error:

```bash
[11:18:51 app_8] Could not connect to migration manager on port [64501]
[11:18:51 app_8] Database not migrated, or could not verify; trying again in 30s
[11:18:51 app_12] Could not connect to migration manager on port [64501]
[11:18:51 migration_manager] INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.[11:18:51 app_12] Database not migrated, or could not verify; trying again in 30s

[11:18:51 migration_manager] INFO  [alembic.runtime.migration] Will assume transactional DDL.[11:18:51 app_9] Could not connect to migration manager on port [64501]

[11:18:51 migration_manager] INFO  [alembic.runtime.migration] Running upgrade 7eab209f29d3 -> 5f949cc5cae9, localizationtiles_partitions_2027[11:18:51 app_9] Database not migrated, or could not verify; trying again in 30s

[11:18:51 migration_manager] Traceback (most recent call last):
...
[11:18:51 migration_manager]     self.dialect.do_execute(
[11:18:51 migration_manager]   File "/pdisk/pyenv/versions/3.11.10/envs/skyportal-311/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 747, in do_execute
[11:18:51 migration_manager]     cursor.execute(statement, parameters)
[11:18:51 migration_manager] sqlalchemy.exc.IntegrityError: (psycopg2.errors.CheckViolation) updated partition constraint for default partition "localizationtiles_def" would be violated by some row
[11:18:51 migration_manager] 
[11:18:51 migration_manager] [SQL: CREATE TABLE localizationtiles_2025_05 PARTITION OF localizationtiles FOR VALUES FROM ('2025-05-01') TO ('2025-06-01')]
[11:18:51 migration_manager] (Background on this error at: https://sqlalche.me/e/20/gkpj)
```

as the partition would already exist. I propose to change to:

```python

curr = datetime.datetime.now()
next_month = (curr.replace(day=1) + datetime.timedelta(days=32)).replace(day=1)

def upgrade():
    for year in range(curr.year, 2028):
        for month in range(1 if year != curr.year else next_month.month, 13):
            ...
```

Theodlz: Hi Julien, are you saying that the DB auto created the next partition basically. And therefore the conflict when elwe try to add it?

Theodlz: No wait looking at your error, are you sure that partition already exists? Isn't it rather that there is data from May that ended up in the default partition when it should be in the new one, and it doesn't know how to deal with that?

Theodlz: There is no automated partition creation logic as far as I can tell, so I'm not sure how a May to June partition could auto spawn, since it has no way of knowing it should be a month or whatever like that

Theodlz: are you sure that it didn't create the new partition and then failed when attaching it because the real issue was data that should belong to that partition already existing in the localization_def table?

Theodlz: Yeah I really don't think this is happening because the new partition already exists. Otherwise our migration tests on main would fail (they checkout to an old branch, initialize the DB, checkout to a new one and migrate, just like you do).

Theodlz: in which case what needs to be done is to create a temporary table, move all the data with a dateobs >= 2025-05-01 to the temp table, add the partitions, add the data from the temp table to the right partitions, delete the temporary table?

PR #5604, opened 2025-05-01
InstrumentFieldTiles: VACUUM query fix

Theodlz:
Fixes an error encountered while running `VACUUM ANALYZE instrumentfieldtiles` after inserting new instrument field tiles, since `VACUUM` can't run in a pending transaction so now we run it in a separate session, async and with auto-commit which is required.

### DIFF
--- a/skyportal/handlers/api/instrument.py
+++ b/skyportal/handlers/api/instrument.py
@@ -1,4 +1,5 @@
 import ast
+import time
 from io import StringIO
 
 import arrow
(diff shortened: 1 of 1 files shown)


stefanv:
> ANALYZE: Collects up-to-date statistics about the distribution of data in each table and its columns. The PostgreSQL query planner uses these statistics to estimate row counts, value distributions, and selectivity, which are essential for choosing the most efficient query execution plan

stefanv (approved):

PR #5606, opened 2025-05-02
LS Thumbnails: DR9 -> DR10

Theodlz:
LS DR10 covers a lot more area in the sky than DR9 does. While it does also come with a few area where the band covered are sparse (and that results in images which colors aren't quite as nice), the difference in sky coverage is significant enough in my opinion to justify a migration. As we start having alerts from surveys in both the south and the north, that increased coverage in more than welcome.

| DR9 Sky Coverage | DR10 Sky Coverage |
| --- | --- |
| image | image |

### Notes:
- Migrating the existing thumbnails is more or less bonus, but I have a script I wrote and plan to run in Fritz which I think I could add somewhere in the codebase and link in this issue + the next release, any thoughts about that?
- BTW this was motivated by a discussion with Peter Nugent over email, that started using one of the new SkyPortal instances that are out there, and noticed that the thumbnails weren't running on the latest DR.


mcoughlin:
I think doing the migration of thumbnails is nice, and maybe a brief note in the docs how to do it?

Theodlz:
> I think doing the migration of thumbnails is nice, and maybe a brief note in the docs how to do it?

Sounds good, I'll find a place to add that script then, and will mention it in the docs

Theodlz:
not saying we couldn't come up with something more clever, but I don't think it would be that helpful nor worth the effort.

Theodlz:
@stefanv I was thinking about a simple script like:

```python
import sqlalchemy as sa

from baselayer.app.env import load_env
from baselayer.app.models import init_db
from skyportal.models import DBSession, Obj, Thumbnail

env, cfg = load_env()

init_db(**cfg["database"])

if __name__ == "__main__":
    with DBSession() as session:
        page, total_processed = 0, 0
        while True:
            obj_ids = session.scalars(
                sa.select(Obj.id).order_by(Obj.id).distinct().offset(page * 1000).limit(1000)
            ).all()
            if not obj_ids:
                break
            for obj_id in obj_ids:
                stmt = sa.select(Thumbnail).where(Thumbnail.obj_id == obj_id, Thumbnail.type == "ls")
                thumbnails = session.scalars(stmt).all()

                # if no thumbnail exists, create one
                if not thumbnails:
                    obj = session.scalar(sa.select(Obj).where(Obj.id == obj_id))
                    session.add(
                        Thumbnail(obj=obj, public_url=obj.legacysurvey_dr10_url, type="ls")
                    )
                    session.commit()
                    continue

                # delete duplicate thumbnails if any
                if len(thumbnails) > 1:
                    for thumbnail in thumbnails[1:]:
                        session.delete(thumbnail)
                    session.commit()

                # update the existing thumbnail, if necessary
                public_url = thumbnails[0].public_url
                if public_url is None:
                    print(f"Warning: LS thumbnail for obj_id {obj_id} is None, which is unexpected.")
                    continue
                if "layer=ls-dr10" and "bands=griz" in public_url:
                    continue
                public_url = public_url.replace("layer=ls-dr9", "layer=ls-dr10")
                public_url = public_url.replace("bands=grz", "bands=griz")
                thumbnails[0].public_url = public_url
                session.commit()
            
            page += 1
            total_processed += len(obj_ids)
            print(f"Processed page {page} (total={total_processed})")

        print(f"Total processed: {total_processed}")
```

Theodlz:
Just to be clear, the migration is 100% optional. Existing DR9 thumbnails are still working just fine. It's simply that new ones will be pointing to DR10.

stefanv:
That script looks great :+1:

Theodlz:
> That script looks great 👍

Thanks, I'll add it somewhere in the codebase (maybe in the tools or skyportal/utils folder?) + mention it somewhere in the docs, and of course will add a pointer to it in the next release.

Theodlz:
I added the script + documentation about thumbnails in general

mcoughlin (approved):
LGTM

PR #5607, opened 2025-05-03
Joblib update bugfix

Theodlz:
* specifies the bytes limit of the cache size after initiating the Memory() object, since that is now deprecated in joblib v1.5.0 (released on May 3rd 2025)


mcoughlin (approved):
LGTM

PR #5608, opened 2025-05-04
OpenAI + Pinecone: python dependencies and config default updates

Theodlz:
In this PR we:
- update all openai summary-related python dependencies
- update the default chat and embedding models in the config, which are very outdated
- update the pinecone default config, which now seems to be using aws for the free tiers.

What's not clear to me is how we migrate the existing indexes to the new embedding models. That said, I believe that the indexes we already had were turned off / deleted so might not be a problem anyways... As far as I can tell, we've been using the free tier of pinecone in production and we could just run a quick script to regenerate embeddings for all existing summaries, which isn't that many and given how short they are and how few we have, it will be super cheap too.
I say "we" but I should mean the Fritz team. I don't know of anyone using this feature in production (and quite frankly, even in Fritz no one really uses this but we can only hope that with the new models the outputs are more useful).


services/openai_analysis_service/app.py
```diff
@@ -310,10 +310,20 @@ def run_openai_summarization(data_dict):
...
+            e = client.embeddings.create(
+                input=openai_summary,
+                model=summarize_embedding_config.get("model", "text-embedding-3-small"),
+            )
+        except Exception as e:
+            log(f"OpenAI embedding failed {e}")
+            rez.update(
+                {
+                    "status": "failure",
+                    "message": f"OpenAI embedding failed {e}",
```
mcoughlin, line 323: Missing colon :)

mcoughlin (approved):
Good to get this repaired

Theodlz:
> Good to get this repaired

I'm mostly annoyed at the constant updates from dependabot that we can't really approve safely given how quickly some of these python APIs changes. That said looks like it all still worked here (though I don't have any credits so I can't really check the whole workflow) even after updating all the dependencies so that's good. Didn't work that smoothly last time!

services/openai_analysis_service/app.py
Theodlz: added

services/openai_analysis_service/app.py
```diff
@@ -310,10 +310,20 @@ def run_openai_summarization(data_dict):
...
 
     if USE_PINECONE:
-        e = client.embeddings.create(
-            input=openai_summary,
-            model=summarize_embedding_config.get("model", "text-embedding-ada-002"),
-        )
+        try:
+            e = client.embeddings.create(
+                input=openai_summary,
+                model=summarize_embedding_config.get("model", "text-embedding-3-small"),
```
profjsb, line 316: Can you pull the model name from the `embeddings_store` section of the config instead of hard-coding it?

profjsb (approved):
LGTM, small request for a change.

services/openai_analysis_service/app.py
Theodlz: @profjsb That's already what this does. It defaults to `text-embedding-3-small` if no model is specified in the config, but this line here does read from `summarize_embedding_config` which contains the config's keys.

Theodlz: @profjsb let me know when you have seen my answer, just so I know if you are fine with me merging as is

profjsb:
Go ahead and merge after tests pass.

PR #5609, opened 2025-05-04
Test Migration bugfix: joblib version

Theodlz:
* make sure joblib version is less than 1.5.0 in the migration test, as the migration reference commit would otherwise run on joblib latest which yields an error (see #5607)


mcoughlin (approved):
Merge once solved of course

PR #5610, opened 2025-05-04
CI: Use PostgreSQL v17

Theodlz:
This is an attempt at using postgresql version 17 instead of 15 for the CI, so we know if that version is suitable for SkyPortal


Theodlz:
@stefanv as per our discussion last week ⬆️

stefanv (approved):

PR #5616, opened 2025-05-09
feat: Source tags frontend

camilledzt:
The scope of this pull request is :
* Adding the frontend side of the tag feature on sources.
* Only users that can manage sources are allowed to add tags to sources.

Here is a preview of the feature:
image

image


static/js/components/ObjectTags.jsx
```diff
@@ -0,0 +1,331 @@
...
+          size="small"
+          className={styles.addButton}
+          onClick={handleOpenDialog}
+          data-testid="add-tag-button"
+        >
+          <AddIcon fontSize="small" />
+        </IconButton>
+      </Tooltip>
+
+      <Dialog
```
Theodlz, line 215: We usually avoid using custom forms when we can, to stick with forms from react json schema form. Any reason to use a custom form here? Not saying it's incorrect, just curious.

static/js/components/source/Source.jsx
```diff
@@ -591,6 +591,7 @@ const SourceContent = ({ source }) => {
               <div className={classes.header}>
                 <FavoritesButton sourceID={source.id} />
                 <h6 className={classes.name}>{source.id}</h6>
+                <ObjectTags source={source} />
```
Theodlz, line 594: Looking at your screenshots, the placements of these tags doesn't look ideal. I mean maybe it looks off because it's those long UUIDs from the tests, but wondering how well does that scale on a real source if folks put lots of tags. Basically it can be next to the source name on the right, or under it. Under it might avoid running into some issues.

Theodlz: Also from your screenshot (if you do keep them on the right of the name at least) there is not much spacing between the source name and the tags.

Theodlz: and maybe the text seems a little small when compared to the other text elements? (this is a minor comment)

static/js/ducks/objectTags.js
```diff
@@ -0,0 +1,60 @@
...
+
+export function addObjectTag(data) {
+  return API.POST("/api/objtag", ADD_OBJECT_TAG, data);
+}
+
+export function deleteObjectTag(data) {
+  return API.DELETE(`/api/objtag/${data.id}`, DELETE_OBJECT_TAG, data);
+}
+
+messageHandler.add((actionType, payload, dispatch) => {
```
Theodlz, line 38: but it doesn't look like you are even using the output of this right? If you want to update the object tags in place, might want to listen for an `FETCH_OBJECT_TAGS_OK` event in the `source.js` ducks to update a source's tags?

skyportal/tests/frontend/test_tag.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    assert status == 200
+
+    tag_id = data["data"]["id"]
+
+    driver.get(f"/source/{public_source.id}")
+    driver.wait_for_xpath(f'//h6[contains(text(), "{public_source.id}")]', timeout=20)
+
+    add_tag_button = driver.wait_for_xpath_to_be_clickable(
+        '//div[contains(@class, "chips")]/following-sibling::button[@data-testid="add-tag-button"] | //div[contains(@class, "chips")]/following-sibling::*//button[@data-testid="add-tag-button"]',
```
Theodlz, line 28: that path is super super specific, why not just look for `*//button[@data-testid="add-tag-button"]'`? That seems overkill and prone to easily breaking if we change the parent elements

skyportal/tests/frontend/test_tag.py
```diff
@@ -0,0 +1,100 @@
...
+        timeout=10,
+    )
+    driver.execute_script("arguments[0].click();", add_tag_button)
```
Theodlz, line 31: Not needed, we have a click action provided by the driver directly. Please look at how it's done in other tests

skyportal/tests/frontend/test_tag.py
```diff
@@ -0,0 +1,100 @@
...
+    driver.wait_for_xpath('//div[@data-testid="add-tag-dialog"]', timeout=20)
+    driver.wait_for_xpath_to_be_clickable('//div[@data-testid="tag-select"]').click()
+    tag_option = driver.wait_for_xpath(f'//li[@data-testid="tag-option-{tag_id}"]')
+    tag_option.click()
```
Theodlz, line 35: ah well it's just like this actually, you can do the same thing above instead of the `.execute_script(...)`

skyportal/tests/frontend/test_tag.py
```diff
@@ -0,0 +1,100 @@
...
+
+def test_create_new_tag(driver, user, public_source, super_admin_user):
+    """Test creating a new tag from the source page."""
+
+    driver.get(f"/become_user/{super_admin_user.id}")
+    driver.get(f"/source/{public_source.id}")
+
+    driver.wait_for_xpath(f'//h6[contains(text(), "{public_source.id}")]', timeout=20)
+
+    add_tag_button = driver.wait_for_xpath_to_be_clickable(
```
Theodlz, line 58: same comments as the previous test

skyportal/tests/frontend/test_tag.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def test_permission_for_tag_creation(driver, user, public_source):
+    """Test that normal users can't create tags (only admins or users with permission)."""
+    driver.get(f"/become_user/{user.id}")
+    driver.get(f"/source/{public_source.id}")
+
+    driver.wait_for_xpath(f'//h6[contains(text(), "{public_source.id}")]', timeout=20)
+
+    add_tag_button = driver.wait_for_xpath_to_be_clickable(
```
Theodlz, line 87: same comment as the first test

skyportal/tests/frontend/test_tag.py
```diff
@@ -0,0 +1,100 @@
...
+        '//div[contains(@class, "chips")]/following-sibling::button[@data-testid="add-tag-button"] | //div[contains(@class, "chips")]/following-sibling::*//button[@data-testid="add-tag-button"]',
+        timeout=10,
+    )
+    driver.execute_script("arguments[0].click();", add_tag_button)
+
+    driver.wait_for_xpath('//div[@data-testid="add-tag-dialog"]', timeout=20)
+
+    create_section_elements = driver.find_elements(
```
Theodlz, line 95: actually for this we usually use the baked-in driver's methods where we check that an xpath has disappeared. I'm fine with how you did it too, that's a minor suggestion

Theodlz (commented):
I have some questions/comments/suggestions @camilledzt I will let you take a look. I'm unsure some of the things you coded up in the ducks are working as expected

static/js/components/source/Source.jsx
camilledzt: I put the tags on the right to do the same as GCN events but I will put it under if you find it better (and also increase the font)

Theodlz: Ok, it's subjective anyways so unlike backend a lot of this is a matter of personal taste, but from my experience some source names can be very long and so could be the tags, so to the right the display could break easily.

static/js/ducks/objectTags.js
Theodlz: @camilledzt you still don't need/use this websocket thing right? Like never does your backend send a `skyportal/FETCH_OBJECT_TAGS` action am I correct? It only sends a `REFRESH_SOURCE` action I believe? If so, please remove this before we can merge

Theodlz (approved):
LGTM!

PR #5618, opened 2025-05-13
TRT docs

mcoughlin:
This PR adds TRT to docs (and orders alphabetically).


Theodlz (approved):
LGTM. We do show all those parameters in the allocation forms now, so I'm hoping this page becomes less and less useful so we don't need to maintain it!

PR #5619, opened 2025-05-14
Set MUI table header to fixed position

antoine-le-calloch:
- Improve the usability of certain MUI tables by making the header fixed during vertical scrolling.
- Remove the unused hideTitle variable from the codebase.


mcoughlin (approved):
I don't really remember the provenance of the hideTitle bit, e.g. why it was there originally. Obviously if we don't use it, fine for me to get rid of it.

Theodlz:
> I don't really remember the provenance of the hideTitle bit, e.g. why it was there originally. Obviously if we don't use it, fine for me to get rid of it.

@antoine-le-calloch @mcoughlin basically we used to use it because some tables were used on their own page vs embedded in other pages, and in some cases we wanted to have a title for the table shown in the table's header and in some cases we had a separate title on top of table. A while back I remember having a PR where I removed a lot of the separate titles (e.g. "Sources" being displayed as a h2 title on top of the source table for the source page) to just use the ones from mui datatable which show in the table's header, but I didn't drop the `hideTitle` (and might even have added some!) in case we were (or wanted to) re-use some tables in some pages but without re-displayed the title in the header.

In my opinion we can just show the header title everytime and hence don't need the `hideTitle` text, especially since most pages do not have a separate title anymore.

Theodlz:
@antoine-le-calloch I think you claimed that hideTitle wasn't used anywhere right? Because doing a quick search on main I see a ton of places where we use it on the component, it doesn't look unused to me... (e.g. https://github.com/skyportal/skyportal/blob/578e2335315751af486d3695553c94c73f81f04a/static/js/components/gcn/GcnSelectionForm.jsx#L296)

Theodlz:
@antoine-le-calloch thanks for the in-depth check Antoine. I'm fine removing unused parameters but quite frankly given how minor this is, I just wonder why bother removing it in the first place if we do indeed use it in some instances. It's just practical to have a title show by default (because that's what we want in most cases) and have the option to hide it (and yes in JS it's a standard for boolean parameters to be false by default on components, and then you don't need to set them to a value but just call them by name as the default value is true when that's what you do).

Theodlz (approved):
LGTM. One thing to note is that as you dezoom, the table height decreases so it's not 100% responsive (you could calculate the height as some view height - the size of the header which is fixed in pixel I think). It's a small detail and it already looks good so that could just be a separate PR (if its even desirable).

here's an example of what it currently looks like when I dezoom (and would look like on any screen larger than a laptop):
image

antoine-le-calloch:
> @antoine-le-calloch did you see these comments btw?

Yes yes, it's planned.

antoine-le-calloch:
@Theodlz Done!

Theodlz:
@antoine-le-calloch I'm fixing some issues that likely impact the startup time of the app (and hence the CI) in #5669, so I'll update your branches and merge stuff once it's all green

Issue #5622, opened 2025-05-19
Photometry table customization


camilledzt:
The photometry table required some customization to fit GRANDMA needs.

Removes the following default columns :
* Ra
* Dec
* Ra_unc
* Dec_unc 

Add these columns by default:
* SNR 
* Magnitude System (e.g., AB/Vega)

PR #5623, opened 2025-05-19
enh: change default columns in the photometric table

camilledzt:
Resolves issue #5622 
Removes the following default columns :
* Ra
* Dec
* Ra_unc
* Dec_unc

Add these columns by default:
* SNR
* Magnitude System (e.g., AB/Vega)

Move UTC column next to MJD column

### DIFF
--- a/static/js/components/photometry/PhotometryTable.jsx
+++ b/static/js/components/photometry/PhotometryTable.jsx
@@ -468,7 +468,6 @@ const PhotometryTable = ({ obj_id, open, onClose, magsys, setMagsys }) => {
                 return [
                   x.data[0], // id
                   x.data[1], // mjd
-                  x.data[17], // UTC date
                   x.data[2], // mag
                   x.data[3], // magerr
                   x.data[4], // limiting_mag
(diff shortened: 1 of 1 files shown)


static/js/components/photometry/PhotometryTable.jsx
```diff
@@ -201,7 +208,8 @@ const PhotometryTable = ({ obj_id, open, onClose, magsys, setMagsys }) => {
           </div>
         );
       };
-      columns.push({
+      const mjdIndex = columns.findIndex((col) => col.name === "mjd");
```
Theodlz, line 211: I'm a little confused by this line, what is it trying to do?

Theodlz: (that and the line under it)

Theodlz: Sounds good

Theodlz:
@camilledzt one thing to keep in mind is that the SNR will often be empty. This is because that table uses the SNR from the DB and well we don't really use that in practice. For the download and the photometry plot, I compute it for example. Might want to edit the backend to compute the SNR on the fly or something like that, so we don't need to recompute it in multiple places in the frontend?

PR #5624, opened 2025-05-19
Add delay on tarot request

antoine-le-calloch:
Add a 10-minute delay to TAROT's observing time when observing now to ensure the observation.


mcoughlin (approved):
Did you check with Sarah and Alain here? I am ok with it as long as it works of course.

antoine-le-calloch:
> Did you check with Sarah and Alain here? I am ok with it as long as it works of course. 

This is a request from Sarah, and I need to speak with Alain to find a better solution to Tarot being really strange ahah.

Theodlz (approved):
LGTM

PR #5625, opened 2025-05-20
add and delete source aliases

mcoughlin:
This PR manages source aliases.


Theodlz:
There is a minor css issue, see this large gap between the 2 rows:

image

Theodlz:
also the text elements aren't aligned:
image

Theodlz:
From what I can see, there is an extra 1rem margin on the whole component that shouldn't be there, and the `aliasItem` has incorrect padding.

mcoughlin:
@Theodlz I think I got it.

Theodlz (approved):
LGTM, unfortunately with that delete button that shows on hover (but still uses the width before showing) there is quite a bit of spaces between the aliases but it's still fine, could revisit that later if we want too, but absolutely not a blocker.

Merge once green

Issue #5429, opened 2025-01-23
Vega refresh problem


santier14:
Hello, 
When we turn into vega and modify the validation, then it comes back to AB and not vega and then it is impossible to change it :)

santier14:
@Theodlz ? :)

PR #5626, opened 2025-05-20
fix: magsys state is reinitialized after a validation in the photometry table

camilledzt:
Resolve issue #5429 

The goal of this PR is to add confidence to the magsys used. 
The magsys was set to "ab" automatically after a validation.
This PR fixes the issue by adding a magsys prop to the PhotometryValidation component.


Theodlz (approved):
LGTM

PR #5627, opened 2025-05-20
Colibri submission form

mcoughlin:
This PR updates the Colibri submission form to be in line with SEDM-KP style (with the correct filter set).


skyportal/facility_apis/colibri.py
```diff
@@ -181,33 +233,41 @@ def delete(request, session, **kwargs):
...
-        "properties": {
-            "username": {
+            "end_date": {
                 "type": "string",
-                "title": "Username",
+                "title": "End Date (UT)",
+                "default": (Time.now() + TimeDelta(7, format="jd")).isot,
             },
-            "password": {
+            "too": {
```
Theodlz, line 244: any reason why this isn't just a bool?

Theodlz (approved):
LGTM

PR #5628, opened 2025-05-20
GECam VOEvents

mcoughlin:
This PR adds GECam VOEvents by default.


Theodlz:
@mcoughlin nothing required in the code other than adding to the config? Do we support those notices just fine already?

mcoughlin:
I am not convinced anything is coming through these days: https://gcn.gsfc.nasa.gov/gecam_events.html

PR #5629, opened 2025-05-20
Fix priority handling in scanning report to handle None values

antoine-le-calloch:
Sometimes, follow-up requests may not include a priority value which create an error when creating scanning report. This case should be handled properly when sorting and selecting the follow-up request to display.


Theodlz (approved):
LGTM

PR #5632, opened 2025-05-22
Bugfixes: TNS reporter string deduplication + lalsuite<7.26

Theodlz:
Quick bugfix to avoid duplicating names in case multiple user accounts with the same name + affiliation are in a report. While its not impossible to have 2 people with the exact same first name + last name, + the same affiliation is rare enough that I think it's safe to deduplicate based on that.

This is based on an issue we had where one user was set as a coauthor with their real account, and then another account used as a bot (diff username but same name for TNS reporting purposes) created TNS reports where the same name showed up twice.

Also, latest `lalsuite` (release 24h ago) breaks the python dependencies. In an effort to get back to green on main for some folks to deploy the app safely, we can just pin it for now.


Theodlz:
necessary to redeploy, self merging

PR #5635, opened 2025-05-27
Fix tarot config key error

antoine-le-calloch:
fix tarot: defer config access to prevent early KeyError
- Move configuration access inside the error-handled function to avoid KeyError during initialization.

Link to this comment:
https://github.com/fritz-marshal/fritz/issues/562#issuecomment-2905559788


mcoughlin (approved):
LGTM

antoine-le-calloch:
@Theodlz I went this way since the feature still works (submit, delete, etc.) even if observation_url isn’t defined. But can switch to the flag if you prefer.

PR #5636, opened 2025-05-27
PhotSeries: stats bugfix

Theodlz:
Looks like the stats were hardcoded to mjd and not the field they should be looking at in a loop. This PR fixes that.

Unfortunately this is in the DB itself so we should add some endpoint to force recompute all photstats series or something like that, to make it easier for users that already rely on those to update them.

### DIFF
--- a/skyportal/models/photometric_series.py
+++ b/skyportal/models/photometric_series.py
@@ -550,10 +550,23 @@ class PhotometricSeries(conesearch_alchemy.Point, Base):
         self.maxima = {}
         self.stds = {}
         for key in self._data:
-            self.medians[key] = float(np.nanmedian(self.mjds))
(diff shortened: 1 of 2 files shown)


guynir42:
Wow this is really bad on my part! I'm glad you found this... 

There's a PhotStatUpdateHandler that you can use to force update the phot stats for a batch of sources.
https://github.com/skyportal/skyportal/blob/62bc55616c61a5b6a9a9c1f5e7f10db0b2feecd6/skyportal/handlers/api/phot_stat.py#L533

I remember running a bash script to go over all sources with this API endpoint whenever I needed to update the stats. It was pretty much query a page of sources, and then call this endpoint.

antoine-le-calloch (approved):

howardisaacson:
Thanks @Theodlz !

PR #5638, opened 2025-05-28
feat: Add source tags to "recent sources" widget and "source table"

camilledzt:
Source tags should be visible in the widget "Recently saved sources" and should be in the sources page in the table.


camilledzt:
UI example with the last modifications 

image

Theodlz:
@camilledzt thanks for adding this, some comments:
- About your latest screenshot:
  - First, once we have tags the elements on the right are now taller than the image. I wonder if we should edit the image size (which is fixed) to account for this potential extra height. To be clear, not saying it should be responsive, just saying we might want to make it bigger by default so it works fine with and without tags.
  - Also you have 2 tags than `+1`, but you have a lot more width you can use right? Maybe there is a clever way to calculate how much width is available for you to show tags before you start collapsing them? Could be a nice way to show as many tags physically possible?
- In your list of sources: Maybe it's just an impression but you are using a ton of width for that column. That + the classifications, it's starting to be a lot, and we can't see the buttons on the right anymore without scrolling. I am wondering if we should reduce the size of both columns (if its fixed), make the size responsive + some limit on nb of tags/characters, or just combine the classifications and tags column? Not sure which is best, but as is it's just hard to use the table nicely.

camilledzt:
>       * First, once we have tags the elements on the right are now taller than the image. I wonder if we should edit the image size (which is fixed) to account for this potential extra height. To be clear, not saying it should be responsive, just saying we might want to make it bigger by default so it works fine with and without tags.

Yes, I agree, do you want me to put the new image size in the TopSources widget, to have the same visual appearance?
Btw, do you want the tags in the TopSources widget ?

>       * Also you have 2 tags than `+1`, but you have a lot more width you can use right? Maybe there is a clever way to calculate how much width is available for you to show tags before you start collapsing them? Could be a nice way to show as many tags physically possible?

Yes I can try :)

>     * In your list of sources: Maybe it's just an impression but you are using a ton of width for that column. That + the classifications, it's starting to be a lot, and we can't see the buttons on the right anymore without scrolling. I am wondering if we should reduce the size of both columns (if its fixed), make the size responsive + some limit on nb of tags/characters, or just combine the classifications and tags column? Not sure which is best, but as is it's just hard to use the table nicely.

Yes, I'll check that too, I'll keep you posted.

Theodlz:
@camilledzt I'll review this again tomorrow, thanks for making the changes

Theodlz:
@camilledzt would you mind showing this at the next SkyPortal call? I'd like to get the opinion of other folks, see if they think this is the right approach and looks nice. I'll review it today anyways (and merge it if happy w/ it), but wouldn't mind getting some more eyes on it still.

camilledzt:
> @camilledzt would you mind showing this at the next SkyPortal call? I'd like to get the opinion of other folks, see if they think this is the right approach and looks nice. I'll review it today anyways (and merge it if happy w/ it), but wouldn't mind getting some more eyes on it still.

Yep no problem, which part are you unsure about?

Theodlz:
@camilledzt I pushed my frontend changes to your PR, please take a look and let me know what you think

Theodlz:
@mcoughlin @antoine-le-calloch what do you guys think, here's the before (prod on fritz):
image

and after (locally, demo data):

image

Basically images are larger for readability and to get some of the necessary space to show the tags, and each source is in its own container to have a more obvious separation

Theodlz:
> @antoine-le-calloch I think that's because the +N is in italic. Do you want that I remove that ?

The offset is barely noticable, let's just fix it in another PR if needed so we can go ahead and merge this one ASAP

skyportal/handlers/api/sources.py
```diff
@@ -2445,6 +2447,45 @@ async def get_sources(
...
+                FROM obj_tags
+                JOIN objtagoptions ON obj_tags.objtagoption_id = objtagoptions.id
+                WHERE obj_tags.obj_id IN {query_str}
+                """
+
+                tags_result = session.execute(text(stmt).bindparams(*bindparams))
+                tags_by_obj = {}
+
+                for row in tags_result:
+                    obj_id = row[0]
```
Theodlz, line 2467: Would be more elegant to deconstruct this with smth like:
```python
obj_id, tag_id, tag_name, objtagoption, _ = row
```
or similar, not that you should make any change here but just good to keep in mind for later

Theodlz (approved):
LGTM

PR #5639, opened 2025-05-31
Fix tarot time calculation

antoine-le-calloch:
- Set the 10 minutes delay before observation time calculation and improve error message.

- Use a 12-hour increment instead of 1 day when the target is not visible, to better avoid wasting all 7 loop attempts, since if T₀ is not visible, T₀ + 1 day likely won't be either.

### DIFF
--- a/skyportal/facility_apis/tarot.py
+++ b/skyportal/facility_apis/tarot.py
@@ -398,14 +398,14 @@ class TAROTAPI(FollowUpAPI):
         check_payload(request.payload, specific_config["station_name"])
 
         hash_user = login_to_tarot(request, session, altdata)
-        observing_time = get_observing_time(session, request)
 
         # Add 10 minutes delay to the observing time to avoid issues with the TAROT server
         # This code is a workaround and should be removed after finding a solution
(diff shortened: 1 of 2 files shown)


skyportal/facility_apis/tarot.py
```diff
@@ -596,7 +596,7 @@ def get(request, session, **kwargs):
                     initiator_id=request.last_modified_by_id,
                 )
                 session.add(transaction)
-                raise ValueError("Error trying to get the observation log")
+                raise ValueError("Observation log currently unavailable on TAROT")
```
mcoughlin, line 599: For TAROT

mcoughlin (approved):
LGTM

Theodlz (approved):
LGTM, makes a lot of sense to do 12 hours rather than a day.

skyportal/facility_apis/tarot.py
mcoughlin: I just meant for instead of on.

antoine-le-calloch: yes yes I understood.

PR #5661, opened 2025-06-04
Migration CI bugfix + python dependencies updates

Theodlz:
Trying to revive some of the CI, pinecone update caused some weird issues on main (runs locally for me but fails here...).

I took advantage of this PR to update some dependencies, and address a couple of deprecation warnings. I also opened a PR on baselayer to address more warnings there.

### DIFF
--- a/.github/workflows/test_migrations.yaml
+++ b/.github/workflows/test_migrations.yaml
@@ -191,6 +191,9 @@ jobs:
           # First make sure the dependencies are up-to-date
           npm i --legacy-peer-deps
           make dependencies
+          # Make sure to remove python packages from the reference commit that conflict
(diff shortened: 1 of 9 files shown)


.github/workflows/test_migrations.yaml
```diff
@@ -191,6 +191,9 @@ jobs:
           # First make sure the dependencies are up-to-date
           npm i --legacy-peer-deps
           make dependencies
+          # Make sure to remove python packages from the reference commit that conflict
```
mcoughlin, line 194: We could also bump the reference commit right?

Theodlz: Yes but then we would not test any migrations so it's not an option at the moment

stefanv (approved):

PR #5665, opened 2025-06-09
Add MUI table for telescopes display

antoine-le-calloch:
- Add MUI table for telescope list display to allow sorting
- Improve telescope list display
- Improve research
- improve adding new telescopes
- Delete useless code


Theodlz:
very nice @antoine-le-calloch, I'll start reviewing it now

Theodlz:
2 minor comments:
- the arrow left icon to switch between map and mui table looks a bit odd, there is some weird border + shadow like thing in my opinion. Maybe that's a quick fix? it feels a little out of place compared to the rest of the page.
- what about mobile? I think that before we showed the mui table but now it looks like you show the table we used to have on the right of the telescope map. Is that switch intentional?

PR #5668, opened 2025-06-11
Fix some frontend warning on profil page

antoine-le-calloch:
I was investigating an error we encountered when trying to open the QR code, but I saw that Théophile had already fixed it and it was just not in production yet. So I just pushed small warning fix.


Theodlz:
@antoine-le-calloch

Theodlz (approved):
LGTM, thanks for addressing these warnings

PR #5669, opened 2025-06-11
TAP Services: Gaia implementation

Theodlz:
Currently, in `skyportal/utils/offset.py` and `skyportal/handlers/api/annotation_services.py` we query the Gaia DR3 `gaia_source` catalog. We first try to use `astroquery`'s implement (class called `Gaia`), then if that fails to connect we then use `pyvo`'s `TapService` to connect to a backup service. However:
- the `astroquery.gaia` import causes a lot of issues, because in their code they instantiate their class which triggers a query. So the imports are slowed down and when servers are unstable/down like they are now, it just **prevents the app from even starting**.
- the `astroquery.gaia` implementation does not let us specify timeouts, so queries that take longer than expected just paralyze us and we can't do anything about it.
- out of these 2 options to query gaia, the query implementations are a little different, which leads to a number of if/else statements
- out of these 2 options to query Gaia (astroquery pointing to the main gaia server or pointing to a backup one using pyvo), we do no cache which one worked last. So if the main server is down, every time we want to query gaia we keep retrying it. When really it would be better to keep track of which server last answered in time and use that thereafter.

So in this PR:
- we completely drop astroquery.gaia in favor of a custom class that only uses `pyvo`'s `TapService` class to query either the main server (ESA's) or the backup one (AIP).
- the new custom class does not query the server at initialization but only at query time. If the query fails for any reason, we switch to the next server and try again. As soon as a server responds, we put that server at the top of our list of servers to try. Effectively caching which one last responded.
- we switch the main and backup servers, as the AIP has proved to be very stable (unlike the main ESA one that has been timing out most of the past 2 weeks).
- implement the `__enter__` and `__exit__` methods so this class can be used as a context manager, which ensures that the `requests` `session`'s attached to the server connections are closed when we are done querying (that's a little overkill and could be dropped).
- we add the new class to a dedicated `skyportal/utils/tap_services` folder, in `gaia.py`. Then we import it everywhere we need to query Gaia.

### DIFF
--- a/skyportal/handlers/api/annotation_services.py
+++ b/skyportal/handlers/api/annotation_services.py
@@ -18,7 +18,6 @@ from io import StringIO
 import numpy as np
 import pandas as pd
 import sqlalchemy as sa
-from astroquery.gaia import GaiaClass
 from astroquery.ipac.irsa import Irsa
 from astroquery.vizier import Vizier
 from sqlalchemy.exc import IntegrityError
@@ -26,6 +25,7 @@ from sqlalchemy.exc import IntegrityError
 from baselayer.app.access import auth_or_token
 from baselayer.app.env import load_env
 from skyportal.utils.calculations import great_circle_distance
+from skyportal.utils.tap_services import GaiaQuery
 
 from ...models import (
     Annotation,
@@ -37,6 +37,8 @@ from ..base import BaseHandler
 _, cfg = load_env()
 
 PS1_URL = cfg["app.ps1_endpoint"]
+
+gaia = GaiaQuery()
 
 
 class GaiaQueryHandler(BaseHandler):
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/annotation_services.py
```diff
@@ -122,35 +124,47 @@ def post(self, obj_id):
...
+                        {radius_degrees}))
+            """
+
+            with gaia as g:
+                try:
+                    df = g.query(query_string)
+                    df = df.to_pandas()
+                except Exception as e:
+                    return self.error(
+                        f"Error querying Gaia: {e}. Please try again later."
```
mcoughlin: Should we mention the object name or RA/DEC or something in the error string?

skyportal/utils/offset.py
```diff
@@ -995,15 +895,8 @@ def get_nearby_offset_stars(
...
-    r = None
-    for retry in range(2):
-        try:
-            g = GaiaQuery()
-            r = g.query(query_string)
-            break
-        except Exception as e:
-            log(f"Gaia query failed: {e}]")
-            time.sleep(1 + 2**retry)
+    with gaia as g:
```
mcoughlin, line 898: No need to wrap this in a try catch or similar?

skyportal/utils/tap_services/gaia.py
```diff
@@ -0,0 +1,181 @@
...
+DEFAULT_TIMEOUT = 10  # seconds
+SERVERS = [
+    # Leibniz-Institute for Astrophysics Potsdam (AIP)
+    "https://gaia.aip.de/tap",
+    # European Space Astronomy Centre (ESAC) of the European Space Agency (ESA)
+    "https://gea.esac.esa.int/tap-server/tap",
+]
+
+
+class TimeoutHTTPAdapter(HTTPAdapter):
```
mcoughlin, line 29: Should this be in a more generic utils.py function?

mcoughlin (approved):
A couple of things to consider.

skyportal/handlers/api/annotation_services.py
Theodlz: I'm not sure I see how that would help the user? But I can add that anyways

skyportal/utils/offset.py
Theodlz: the with itself doesn't need it (nothing in the `__enter__` could cause a failure) but maybe the query could, let me add that.

skyportal/handlers/api/annotation_services.py
mcoughlin: Good point. I was thinking just for logging, but you are right.

skyportal/utils/tap_services/gaia.py
Theodlz: Yes I think that's a good point, but could not find any other use for it. I'd suggest moving it to a dedicated and reusable module only if we find another use case for it.

skyportal/handlers/api/annotation_services.py
Theodlz: added it anyways.

skyportal/utils/offset.py
Theodlz: added

PR #5671, opened 2025-06-12
SEDM Allocation based unit tests failures: fix

Theodlz:
Most follow-up request test rely on the `sedm_allocation` test fixtures, which itself depends on the `sedm` fixture. Naturally, the `sedm` fixture uses the associated SEDMAPI. However for a lot of tests we just want to test more generic follow-up request features and absolutely do not need to do that w/ SEDM's API which points to telescope's server.

In this PR we introduce a `generic_instrument` and `generic_instrument_allocation` test fixtures that we can use in place of the `sedm` one in most tests that aren't SEDM-specific. Then we use said fixture in a handful of tests which have been failing today as the SEDM server seems to be down.

PS: In the future we may consider making that move to an SEDM-less allocation (one that in general does not ping a distant server) for most follow-up request-related tests, for consistency and reliability sake.

### DIFF
--- a/skyportal/tests/api/candidates_sources_events/test_followup_requests_api.py
+++ b/skyportal/tests/api/candidates_sources_events/test_followup_requests_api.py
@@ -36,19 +36,22 @@ def test_token_user_post_robotic_followup_request(
 
 
 def test_token_user_delete_owned_followup_request(
-    public_group_sedm_allocation, public_source, upload_data_token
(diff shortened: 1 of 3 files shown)


Theodlz:
tests failing on main are passing locally, it's a minor change and this is needed to have some working CI on main, so self merging. I'll make changes if needed, based on the review I'll get later.

mcoughlin (commented):
I did not see anything of issue, hopefully it's working better now.

PR #5673, opened 2025-06-13
ATLAS FP chi/N + baselayer update

Theodlz:
Collaborators suggested doing a cut on chi/N so that when it's > 10, datapoints are considered as non detections.

Also, I updated baselayer so we have some of Antoine's latest changes (notification styling).


Theodlz:
I also updated the default values for NGPS starlists and avoided having those defined in more than one place, for consistency

PR #5674, opened 2025-06-16
Add utc time header and shift calendar and fix PDF error

antoine-le-calloch:
- Fixes an issue preventing PDF downloads (observability and finder chart) from working reliably on the source page.
The error occurred most of the time due to the MenuItem's onClick event interfering with the anchor tag's default behavior, preventing the download from triggering properly.

- Add UTC time in the header and refresh time display every minute (My idea on this one, this is something I often need when requesting follow-ups, and I imagine other users might benefit from having both timezones visible as well.
image

- Add UTC time to shift calendar display. Not the prettiest solution visually, but it's the best I can do with the calendar component. This was requested by Christina and apparently several other users as well.
image

### DIFF
--- a/static/js/components/shift/ShiftCalendar.jsx
+++ b/static/js/components/shift/ShiftCalendar.jsx
@@ -376,6 +376,13 @@ function MyCalendar({ events, currentShift, setShow }) {
             components={{
               event: Event,
             }}
+            formats={{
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #5677, opened 2025-06-21
Fix small skyportal issues

antoine-le-calloch:
Fix:
- Improve error handling for field_fov_attributes input and improve description in InstrumentForm.
- When checking the existence of a source by position (api/source_exists), and if there is no source at the given ra/dec/radius, the returned information says “a source of that name does not exist.”
- Improve handling of source_exists on the frontend.
- Correct improper use of form_data during source creation which prevents using the form twice by updating the actual form data.

### DIFF
--- a/skyportal/handlers/api/instrument.py
+++ b/skyportal/handlers/api/instrument.py
@@ -152,7 +152,7 @@ class InstrumentHandler(BaseHandler):
                     else:
                         field_fov_attributes = [float(field_fov_attributes)]
                 except ValueError:
-                    raise ValueError(
+                    return self.error(
                         "field_fov_attributes must be a list of floats or a float"
                     )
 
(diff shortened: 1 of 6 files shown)


mcoughlin (approved):
LGTM

antoine-le-calloch:
> LGTM. The git diff doesn't make it so obvious so let me ask to double check: In [skyportal/handlers/api/source_exists.py](https://github.com/skyportal/skyportal/pull/5677/files#diff-82bbe0a52b4a8baea93c4e2ebb2eecd183faf460057f8dea800f65ce3e28617b), the logic of the query itself did not change at all right? It's just the error handling around it that did?

Not really before it was checking if the source exist and sending back error sometime or success but with the text (source exist) and always success with "A source of that name does not exist." when no source exist. So for our frontend the frontend need to check for success and for A source of that name does not exist.

PR #5678, opened 2025-06-23
Update tarot API to call Icare

antoine-le-calloch:
Update TAROT API integration to use Icare tarot proxy endpoints and improve error handling.

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -41,10 +41,7 @@ app:
   lco_archive_endpoint: https://archive-api.lco.global
 
   # Tarot API endpoints
-  tarot_endpoint: http://cador.tarotnet.org/ros
-  calern_endpoint: http://tca4.tarotnet.org/ros/klotz
-  chili_endpoint: http://tch4.tarotnet.org/ros/klotz
-  reunion_endpoint: http://tre4.tarotnet.org/ros/klotz
+  tarot_proxy_endpoint: https://skyportal-icare.ijclab.in2p3.fr/api/tarot_proxy
 
   # MMT API endpoint
   mmt_endpoint: https://scheduler.mmto.arizona.edu/APIv2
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
Looks good. Hopefully can coordinate merger and deployment with Icare instance.

skyportal/facility_apis/tarot.py
```diff
@@ -819,8 +828,11 @@ def custom_json_schema(instrument, user, **kwargs):
         "properties": {
             "request_id": {
                 "type": "string",
-                "title": "Request ID to add scene",
-                "default": "2282",
+                "title": "Request ID to add scene to (2282 for skyportal)",
+            },
+            "icare_token": {
```
Theodlz: what about `proxy_token`? In theory a machine other than icare that has its ports opened and access to the tarot network could run such a proxy, I'd suggest making it a bit more generic and adding a short `description` to the field here

config.yaml.defaults
```diff
@@ -41,10 +41,7 @@ app:
   lco_archive_endpoint: https://archive-api.lco.global
 
   # Tarot API endpoints
-  tarot_endpoint: http://cador.tarotnet.org/ros
-  calern_endpoint: http://tca4.tarotnet.org/ros/klotz
-  chili_endpoint: http://tch4.tarotnet.org/ros/klotz
-  reunion_endpoint: http://tre4.tarotnet.org/ros/klotz
+  tarot_proxy_endpoint: https://skyportal-icare.ijclab.in2p3.fr/api/tarot_proxy
```
Theodlz, line 47: what about addking a little bit of a multi-line comment here to explain how things work for tarot (i.e. it's on a private network, so we connect to a proxy instead)

Theodlz (approved):
LGTM, some minor comments to keep things clean, generic, and clear for the dev and user

Theodlz (approved):
LGTM

PR #5679, opened 2025-06-24
NGPS starlist: number formatting fix

Theodlz:
Handle not just TypeError but also ValueError, when formatting the default values + mag/filter for the NGPS starlist


Theodlz:
self merging

PR #5682, opened 2025-06-26
feat: Add color column to source tags

camilledzt:
This PR is adding customizable color support for object tags with a default gray (#dddfe2).  
Users can now create tags with custom colors using an HTML color picker.

_This is a subtask needed to remove RecentSources widget from Icare instance._


skyportal/models/tag.py
```diff
@@ -16,6 +16,13 @@ class ObjTagOption(Base):
         unique=True,
     )
 
+    color = sa.Column(
+        sa.String,
+        nullable=False,
+        default="#dddfe2",
```
Theodlz: wondering what's best, having a default in the DB or just null. I'm saying this because I suppose this grey matches existing elements on the frontend, but if we change the default color palette there, then this would look obsolete. I would suggest doing null by default and on the frontend use that grey as the backup if no color is provided (e.g. `tag.color || "#dddfe2"`)

camilledzt: Ok :+1:

Theodlz:
@camilledzt same comment about baselayer being incorrectly pinned here, please undo before I can review it and merge it.

Theodlz:
will merge once green

PR #5683, opened 2025-06-28
Fix public page groups filtering

antoine-le-calloch:
Fix wrong public source page filtering logic for groups and streams.

### DIFF
--- a/skyportal/handlers/api/public_pages/public_source_page.py
+++ b/skyportal/handlers/api/public_pages/public_source_page.py
@@ -72,10 +72,15 @@ def get_photometry(source_id, group_ids, stream_ids, session):
     stmt = Photometry.select(session.user_or_token, mode="read").where(
         Photometry.obj_id == source_id
     )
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

skyportal/handlers/api/public_pages/public_source_page.py
```diff
@@ -72,13 +72,10 @@ def get_photometry(source_id, group_ids, stream_ids, session):
...
         Photometry.obj_id == source_id
     )
-    if len(group_ids) > 0 and len(stream_ids) > 0:
-        stmt = stmt.where(
-            or_(
-                Photometry.groups.any(Group.id.in_(group_ids)),
-                Photometry.streams.any(Stream.id.in_(stream_ids)),
-            )
-        )
+    if len(group_ids) > 0:
```
Theodlz: @antoine-le-calloch I thought we had agree on an OR condition over zoom didn't we?

PR #5707, opened 2025-07-03
feat: adding a tag management tab

camilledzt:
This PR is adding a Tag management tab in "Other".
User with permission can manage, meaning edit them bu changing their names and the color.

**_This PR needs to be merged after the PR #5682_**


Theodlz:
@camilledzt same comment here about baselayer being incorrectly pinned to an old version, please undo that change and I'm happy to merge this one

PR #5731, opened 2025-07-07
Fix public page summary not display markdown

antoine-le-calloch:
Add markdown support for public page summary and sanitize output.

### DIFF
--- a/static/public_pages/sources/source/source_template.html
+++ b/static/public_pages/sources/source/source_template.html
@@ -27,6 +27,20 @@
       let subtitle_date = document.getElementById('subtitle_date')
       subtitle_date.innerHTML = "<b>" + radec_hhmmss(`{{ data['ra'] | e }}`,`{{ data['dec'] | e }}`) + "</b>" + subtitle_date.innerHTML
     </script>
+    <!-- Load Marked and DOMPurify and render summary as markdown if available -->
+    {% if data.get('summary') %}
+      <script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>
+      <script src="https://cdn.jsdelivr.net/npm/dompurify@3.0.5/dist/purify.min.js"></script>
+      <script type="module">
+        const summary = document.getElementById("summary");
+        if (summary) {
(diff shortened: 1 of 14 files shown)


mcoughlin (approved):
LGTM

Theodlz (commented):
I'm surprised that this can't just be done in the frontend of the public page (like it is done in the frontend of the private page) directly, rather than adding additional backend dependencies?

static/public_pages/sources/source/source_template.html
```diff
@@ -27,6 +27,20 @@
       let subtitle_date = document.getElementById('subtitle_date')
       subtitle_date.innerHTML = "<b>" + radec_hhmmss(`{{ data['ra'] | e }}`,`{{ data['dec'] | e }}`) + "</b>" + subtitle_date.innerHTML
     </script>
+    <!-- Load Marked and DOMPurify and render summary as markdown if available -->
+    {% if data.get('summary') %}
+      <script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>
+      <script src="https://cdn.jsdelivr.net/npm/dompurify@3.0.5/dist/purify.min.js"></script>
```
Theodlz, line 33: Out of curiosity, what happens if the markdown data is incorrect? Worth trying to see if the page dies or if it just does not render if nicely. Later is preferable.

antoine-le-calloch:
> > I'm surprised that this can't just be done in the frontend of the public page (like it is done in the frontend of the private page) directly, rather than adding additional backend dependencies?
> 
> Not that its a huge issue of course, but I'd rather let the frontend take care of it and avoid the extra python dependency load if we can. More dependencies = more problems

@Theodlz My bad, I thought backend processing was necessary to ensure security. But actually, there is a secure way to do it on the frontend (using DOMPurify and "| tojson")

static/public_pages/sources/source/source_template.html
antoine-le-calloch: Just not render nicely.

Theodlz (approved):
LGTM! Thanks for taking care of this @antoine-le-calloch

PR #5732, opened 2025-07-07
InstrumentHandler field query - cache bugfix

Theodlz:
We load field ids (queried earlier and cached to disk) for a query ran in the InstrumentHandler, but when we query the DB to retrieve the corresponding fields do not enforce a constraint on instrument.id (multiple instrument may - and likely - have same field_id). This means that one may have fields of the wrong instrument returned in the response. This short PR fixes that.


mcoughlin (approved):
LGTM

PR #5733, opened 2025-07-08
feat: Add color to tag in the recent sources widget and top sources widget

camilledzt:
This PR is adding color feature to the RecentSources widget and the TopSources widget


Theodlz (approved):
LGTM

PR #5734, opened 2025-07-09
ATLAS FP: filter out detections fainter than associated limiting mag

Theodlz:
Users pointed out that ATLAS data often comes back with detections (datapoints with a mag and mag error defined) fainter than the associated limiting magnitude. In theory, that should not happen. This PR filters those datapoints out as detections (effectively turning them into non detections).

This was suggested by @robertdstein


mcoughlin (approved):
LGTM

PR #5735, opened 2025-07-09
Update Vega dependencies

Theodlz:
This PR updates all Vega-related dependencies (and schema urls in all vega components, to match the version in package.json).

PS: It also fixes an issue querying for gaia annotations


mcoughlin (approved):
LGTM

PR #5737, opened 2025-07-13
Add log switch to public pages photometry plot

antoine-le-calloch:
Add toggle switches for X axis since t0 and X axis in log in the public page photometry plot.

### DIFF
--- a/static/public_pages/sources/source/source_template.css
+++ b/static/public_pages/sources/source/source_template.css
@@ -124,6 +124,64 @@ header {
       height: 75vh;
       margin: 1rem 0;
     }
+    .plotOptions {
+      padding: 1rem;
+      display: flex;
+      .switchContainer {
+        width: 50%;
+        gap: 0.5rem;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        &.hide {
+          display: none;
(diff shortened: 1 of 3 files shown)


antoine-le-calloch:
> LGTM. The toggles feel weirdly separated / spread appart on that screenshot given that they are 2 options that go together. Might want to have these 2 toggles grouped in a sub div, so they stay close together.

@Theodlz Yeah but put this two toggles on the left or on the right will be weirdly also no, given that there is no other toggle on the plot.

Theodlz:
> > LGTM. The toggles feel weirdly separated / spread appart on that screenshot given that they are 2 options that go together. Might want to have these 2 toggles grouped in a sub div, so they stay close together.
> 
> @Theodlz Yeah but put this two toggles on the left or on the right will be weirdly also no, given that there is no other toggle on the plot.

Does't shock me, it's common to align stuff to the left/right rather than center everything. It feels weird to see them so far from each on the screen if they are options that are meant to be used together. Any thoughts @mcoughlin?

antoine-le-calloch:
@Theodlz ?
Screenshot 2025-07-14 at 5 46 39 PM

PR #5739, opened 2025-07-16
Improve followup request allocations

antoine-le-calloch:
This pull request replaces the start_date/end_date fields with a more flexible validity_ranges structure for allocations, adding validation on the backend and improved display on the frontend. The allocation table was refactored to show active/inactive ranges, and new and update dialog on allocations page has been merge.


skyportal/facility_apis/atlas.py
```diff
@@ -408,7 +408,7 @@ def delete(request, session, **kwargs):
         "properties": {
             "api_token": {
                 "type": "string",
-                "title": "API Token (required)",
+                "title": "API Token",
```
mcoughlin, line 411: ATLAST doesn't require the API token?

data/db_demo.yaml
```diff
@@ -749,245 +749,278 @@ instrument/=SEDMv2/log:
 allocation:
   - pi: Robert Stein
     proposal_id: WINTER-001
-    start_date: "2020-02-12T00:00:00"
-    end_date: "2030-07-12T00:00:00"
+    validity_ranges:
```
mcoughlin, line 752: Maybe add an example with multiple windows? As I guess that is what is changing?

data/db_demo.yaml
```diff
@@ -1439,8 +1472,6 @@ default_gcn_tag:
     filters: {gcn_tags: ['Fermi']}
 
 gcn_event:
-#  - xml:
```
mcoughlin: Personally I like having this in there for ease of testing.

skyportal/handlers/api/candidate/candidate.py
```diff
@@ -850,7 +850,7 @@ def get(self, obj_id=None):
                 filter_ids,
             )
 
-        page_number, n_per_page = get_page_and_n_per_page(page_number, n_per_page, 500)
+        page_number, n_per_page = get_page_and_n_per_page(page_number, n_per_page)
```
mcoughlin, line 853: Is 500 already the default?

skyportal/handlers/api/candidate/candidate_filter.py
```diff
@@ -138,7 +138,7 @@ def get(self):
                 filter_ids,
             )
 
-        page_number, n_per_page = get_page_and_n_per_page(page_number, n_per_page, 500)
+        page_number, n_per_page = get_page_and_n_per_page(page_number, n_per_page)
```
mcoughlin, line 141: Same here.

skyportal/handlers/api/followup_request.py
```diff
@@ -587,6 +587,22 @@ def post_followup_request(
     if allocation is None:
         raise ValueError(f"Could not find allocation with ID {data['allocation_id']}.")
 
+    ranges = allocation.validity_ranges
+    if not isinstance(ranges, list) or not ranges:
+        raise ValueError("Allocation has no validity ranges defined.")
```
mcoughlin: Yeah just want to be sure existing allocations are given a default wide range so no disruptions.

mcoughlin (commented):
Mostly just some questions on setting the defaults. I am glad we are moving in this direction.

skyportal/facility_apis/atlas.py
Theodlz: It does for sure

Theodlz: but it's already marked as required by the schema I suppose, so adding it in text is not needed since it renders the `*` if marked as such?

data/db_demo.yaml
Theodlz: Yes examples are for sure necessary here

data/db_demo.yaml
Theodlz: yes, in general we comment things for a reason, for code because it may still be useful, and for data most likely because we do want to use it in some scenario.

skyportal/facility_apis/atlas.py
antoine-le-calloch: Yep exact.

data/db_demo.yaml
antoine-le-calloch: Okey

skyportal/handlers/api/candidate/candidate.py
antoine-le-calloch: Yes

skyportal/handlers/api/candidate/candidate_filter.py
antoine-le-calloch: Yep

skyportal/handlers/api/followup_request.py
Theodlz: Well best would be to define for which APIs a validity range is REQUIRED. In most cases we absolutely do not care about those, right @mcoughlin? I feel like it should be mandatory where applicable, otherwise completely optional

mcoughlin: I think optional is fine. The reality is that the vast majority of allocations are tied to semesters e.g. 2025B, 2026A, etc. so would be good if people actually put dates in. But we don't want an allocation to run dry when active of course.

antoine-le-calloch: I could set default values using the current date, but that might unintentionally reactivate an expired allocation.
What do you think, @Theodlz?

antoine-le-calloch: If its not required you can still put a big range like the things people do now, because for now frontend required start and end date.

Theodlz: @antoine-le-calloch let's call to discuss this, @mcoughlin feel free to join if you can cause I can't remember exactly how the start and end dates were meant to work here

data/db_demo.yaml
antoine-le-calloch: But the GW190425_initial.xml file don't exist. At least on the github. @Theodlz @mcoughlin

Theodlz: It does, it's right here: https://github.com/skyportal/skyportal/blob/main/skyportal/tests/data/GW190425_initial.xml

Theodlz: that's where it reads it from, if you check the path

Theodlz (approved):
LGTm! Looks great

PR #5743, opened 2025-07-21
enh: enable instrument edit in the photometry table

camilledzt:
The goal of this PR is to allow user to edit the instrument name in the Photometry table.

Before:  
image

After:  
image


Theodlz (commented):
Quick question but otherwise we can go ahead and merge: the backend already supports an instrument id change? Just wanted to make sure this worked as expected

camilledzt:
> the backend already supports an instrument id change? Just wanted to make sure this worked as expected

@Theodlz I think the backend handles this modification seamlessly as long as the field is defined in the schema, and with to the partial=True parameter in the photometry PATCH handler

PR #5744, opened 2025-07-21
fix: Fix mag system conversion with data with the swiftxrt filter

camilledzt:
The goal of this PR is to fix the conversion problem when a user is trying to convert to vega system some photometry points that are in the x-ray domain and cannot be converted. Photometry points are also hidden in the plot using vega system to be coherent.

image

The solution is to ignore photometry points with the filter swiftxrt during the conversion :

image

Note: x-ray data while valid in AB mag system, is not defined / can't be converted in Vega. So this isn't an issue with SkyPortal failing to convert from one system to another, it's just that x-ray data does not exist in Vega to begin with.

### DIFF
--- a/static/js/components/plot/VegaPhotometry.jsx
+++ b/static/js/components/plot/VegaPhotometry.jsx
@@ -193,8 +193,13 @@ const VegaPhotometry = (props) => {
           await dispatch(photometryActions.fetchSourcePhotometry(sourceId));
         }
         if (photometry && photometry?.length > 0 && filters.length === 0) {
+          // Filter out SwiftXRT points as they are not relevant for in the vega system
+          const photometryFiltered = photometry.filter(
+            (datum) => datum.filter !== "swiftxrt",
+          );
+
           const newFilters = [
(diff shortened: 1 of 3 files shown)


static/js/components/plot/VegaPhotometry.jsx
```diff
@@ -231,8 +236,15 @@ const VegaPhotometry = (props) => {
   }
 
   let photometryFiltered = photometry;
+
+  photometryFiltered = photometryFiltered.filter(
```
Theodlz, line 241: maybe the git diff is not helping, but in the VegaPhotometry component, are you **always** hiding the swiftxrt data, regardless of the magsys?

camilledzt: I was thinking that it was not useful because we are in the VegaPhotometry file, but I add the check just to clarify and make sure we are in the right magnitude system

Theodlz: oh @camilledzt vega in this context is just the name of the plotting library, has nothing to do with the magnitude system! This component here is completly agnostic to the magsys values

PR #5745, opened 2025-07-23
feat: enable nmma format option in the download form in the Photometry Table

camilledzt:
The goal of this PR is to enable users to download photometry table in the NMMA format.

This PR also fix the column order to fit user needs.

image

Result: .dat file content example  
```
"utc","filter","mag","magerr"
2020-07-27T10:23:50Z,ztfg,12.095780629486423,0.2
2020-07-27T10:23:50Z,ztfg,20.412458499097877,inf
2020-07-27T10:48:47Z,ztfr,20.638849949069094,inf
```

### DIFF
--- a/static/js/components/photometry/PhotometryDownload.jsx
+++ b/static/js/components/photometry/PhotometryDownload.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import PropTypes from "prop-types";
 import { useDispatch } from "react-redux";
 import Dialog from "@mui/material/Dialog";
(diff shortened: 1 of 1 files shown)


Theodlz:
Hi @camilledzt, I'm not sure I see why this is needed, where do folks need the data to be downloaded in such a specific format? It seems weirdly specific to support something called "NMMA" format. IMO if folks want data to be in a certain format for some kind of script/analysis pipeline, they should very easily be able to handle our existing format.

mcoughlin:
I agree that a page in the NMMA docs giving the lines of code to use would do it.

PR #5747, opened 2025-07-24
enh: refactoring column order

camilledzt:
The goal of this PR is to fix the order of the Photometry table when it's downloaded.

image


Theodlz (approved):
LGTM

PR #5748, opened 2025-07-26
Json file formatting for rubin ToOs

antoine-le-calloch:
For Rubin schedules generated on Fritz/Skyportal, modify the default JSON output file to a format that is compatible with the Rubin block scheduler. In case of a coarsely localized GW event for which Rubin-generated schedules are not optimal, schedules generated using Fritz/Skyportal for Rubin could be ingested manually into the Rubin block scheduler.

### DIFF
--- a/skyportal/handlers/api/gcn.py
+++ b/skyportal/handlers/api/gcn.py
@@ -31,6 +31,7 @@ from astropy.time import Time
 from marshmallow import Schema, validate
 from marshmallow.exceptions import ValidationError
 from marshmallow.fields import Integer
+from sqlalchemy import Integer as Alchemy_Integer
 from sqlalchemy import String, func
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import joinedload, scoped_session, sessionmaker
@@ -99,6 +100,7 @@ from ...utils.gcn import (
     has_skymap,
 )
 from ...utils.notifications import post_notification
+from ...utils.parse import get_page_and_n_per_page
 from ...utils.UTCTZnaiveDateTime import UTCTZnaiveDateTime
 from ..base import BaseHandler
 from .galaxy import MAX_GALAXIES, get_galaxies, get_galaxies_completeness
@@ -1261,7 +1263,6 @@ class GcnEventCatalogQueryHandler(BaseHandler):
               application/json:
                 schema: ArrayOfCatalogQuerys
         """
-
         try:
             gcnevent_id = int(gcnevent_id)
         except ValueError:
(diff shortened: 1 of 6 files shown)


static/js/components/gcn/GcnEventPage.jsx
```diff
@@ -122,20 +99,9 @@ const useStyles = makeStyles((theme) => ({
...
     marginBottom: theme.spacing(1),
   },
-  noSources: {
-    padding: theme.spacing(2),
-    display: "flex",
-    flexDirection: "row",
-  },
-  sourceList: {
-    padding: "0",
-  },
```
Theodlz, line 138: I suppose this means that these styles aren't used by the components we have on the gcn event page anymore (e.g. the comments table, list of sources, ...)? Asking because some of these may take styles as input parameters

skyportal/handlers/api/gcn.py
```diff
@@ -31,6 +31,7 @@
 from marshmallow import Schema, validate
 from marshmallow.exceptions import ValidationError
 from marshmallow.fields import Integer
+from sqlalchemy import Integer as Alchemy_Integer
```
Theodlz: if you want to keep things a little less obscure (without the need for renames), you could just do `import sqlalchemy as sa` and thereafter `sa.Integer`. That's what we usually do if you look at other files in the codebase.

skyportal/handlers/api/observation_plan.py
```diff
@@ -436,16 +437,14 @@ def post_survey_efficiency_analysis(
 
     instrument = allocation.instrument
 
-    observation_plan = observation_plan_request.observation_plans[0]
-    planned_observations = observation_plan.planned_observations
-    num_observations = len(observation_plan.planned_observations)
-    if num_observations == 0:
```
Theodlz, line 442: not really necessary but ok hahaha

static/js/components/gcn/GcnEventPage.jsx
antoine-le-calloch: Yep not use. I've check.

skyportal/handlers/api/gcn.py
antoine-le-calloch: Done.

Theodlz:
If you think the output format looks identical to the one expected (no extra fields that aren't in Shreya's example/docs), and that none of the existing features have been altered (like the auto triggering of plans and such, I saw that you made some changes there which I don't think have anything to do with the Rubin stuff, so just making sure none of the core logic has changed), then yes please go ahead.

antoine-le-calloch:
image

I have fix this error at the same time.

PR #5750, opened 2025-07-27
GCN sources - name vs origin

Theodlz:
Set auto-generated sources' origins independently of their ids/names


antoine-le-calloch (commented):
I've identified additional issues with TAROT API that only occurs with DefaultFollowupRequest. Exemple, the following code causes the process to break:

```
# Set the start date to be at least 10 minutes in the future to avoid issues with the TAROT server
# This is a temporary workaround and should be removed once a proper solution is found
minimum_observing_time = Time.now() + TimeDelta(600, format="sec")
if request.payload["start_date"] < minimum_observing_time:
    request.payload["start_date"] = minimum_observing_time

```
The issue is due to:

> TypeError: Object of type Time is not JSON serializable

I've just pushed a fix to handle this.

skyportal/models/followup_request.py
```diff
@@ -327,20 +328,20 @@ def receive_after_flush(session, context):
...
         # since it's a join table between Obj and Group
         if target_class_name == "GroupObj":
+            try:
+                obj_origin = session.scalar(
+                    sa.select(Obj).where(Obj.id == target.obj_id)
+                ).origin
+            except Exception:
+                obj_origin = None
+            target_data = {**target.to_dict(), "obj_origin": obj_origin}
+
```
antoine-le-calloch, line 344: Here I have also fix an error occurring because target.obj can be None, which caused obj_origin to be unset even if the origin was defined.

skyportal/utils/gcn.py
```diff
@@ -819,7 +819,6 @@ def get_xml_notice_type(root):
     # but it's unclear if these will be used for future notices.
     # So we will just look at the ivorn which is guaranteed to be there
     ivorn = str(root.attrib.get("ivorn", ""))
-    print(f"ivorn: {ivorn}")
```
antoine-le-calloch, line 822: I think this is a oversight but is it really ?

Theodlz: yeah let's remove that print

Theodlz: just a leftover debug print

static/js/components/followup_request/FollowupRequestLists.jsx
```diff
@@ -318,7 +318,17 @@ const FollowupRequestLists = ({
...
+      name: "status",
+      label: "Status",
+      options: {
+        setCellProps: () => ({
+          style: {
+            minWidth: "250px",
+          },
+        }),
+      },
+    });
```
Theodlz, line 331: what's the reason for that change @antoine-le-calloch?

static/js/ducks/gcnEvent.js
```diff
@@ -198,15 +194,6 @@ export const submitObservationPlanRequest = (params) => {
   );
 };
 
-export const editObservationPlanRequest = (params, requestID) => {
```
Theodlz, line 201: oh, these were not in use anymore???

static/js/ducks/gcnEvent.js
```diff
@@ -252,13 +239,6 @@ export function fetchObservationPlan(id) {
   );
 }
 
-export function getCommentOnGcnEventAttachment(gcnEventID, commentID) {
```
Theodlz, line 255: same comment here

static/js/ducks/gcnEvent.js
antoine-le-calloch: yep not use.

static/js/ducks/gcnEvent.js
antoine-le-calloch: yep

static/js/components/followup_request/FollowupRequestLists.jsx
antoine-le-calloch: To avoid veryyy bad display of long status like the status with error message display one word in each line.

Theodlz: Fair enough hahaha, I remember those now

PR #5751, opened 2025-07-29
fix: fix allocation unit test

camilledzt:
The test_allocation_comment_display Selenium test was failing intermittently with an ElementClickInterceptedException:

```
Element <span class=""> is not clickable at point (1898,489) because another element <div class="MuiDialog-container MuiDialog-scrollPaper css-hz1bth-MuiDialog-container"> obscures it
```

Added an explicit wait condition in the one_request_comment_process helper function to ensure the comment edit dialog fully closes to continue


Theodlz (approved):
LGTM

PR #5753, opened 2025-07-30
Gemini takes multiple template_ids

mcoughlin:
This PR updates Gemini to take multiple template_ids.

### DIFF
--- a/skyportal/facility_apis/gemini.py
+++ b/skyportal/facility_apis/gemini.py
@@ -126,28 +126,6 @@ class GeminiRequest:
         group = f"{allocation_group_name}_{allocation_group_pi}_{allocation_id}"
         group = str(group.replace(" ", "_")).strip()
 
-        obsid = request.payload.get("template_id")
-        # it needs to be castable to an int
-        try:
-            obsid = int(obsid)
-        except Exception:
-            raise ValueError("Invalid template ID")
-
-        if (
-            isinstance(altdata.get("template_ids"), str | list)
-            and len(altdata.get("template_ids")) > 0
-        ):
-            template_ids = get_list_typed(
-                altdata.get("template_ids"),
-                int,
-                "Invalid template IDs specified in altdata",
-            )
-            if len(template_ids) > 0 and obsid not in template_ids:
-                raise ValueError(
-                    f"Invalid template ID, must be one of: {str(template_ids)}"
-                )
-
-        obsnum = str(obsid).strip()
         target = request.obj.id
         ra, dec = request.obj.ra, request.obj.dec
         # the ra dec are in deg, we need them in hms dms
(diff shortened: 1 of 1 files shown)


skyportal/facility_apis/gemini.py
```diff
@@ -206,36 +184,63 @@ def _build_payload(self, request, session):
...
+                    int,
+                    "Invalid template IDs specified in altdata",
+                )
+                if len(template_ids) > 0 and obsid not in template_ids:
+                    raise ValueError(
+                        f"Invalid template ID, must be one of: {str(template_ids)}"
+                    )
+
+            obsnum = str(obsid).strip()
+
```
antoine-le-calloch, line 206: If the altdata["template_ids"] is not a valid type, the "Invalid template IDs specified in altdata" error will not be raised because the block will not be executed. Since get_list_typed() already checks for the correct type, you can just use that.

skyportal/facility_apis/gemini.py
```diff
@@ -348,10 +354,10 @@ def submit(request, session, **kwargs):
...
         "properties": {
-            "template_id": {
-                "title": "Template ID",
-                "type": "integer",
-                "description": "The template ID is found on the program's page on the OT",
+            "template_ids": {
+                "title": "Template IDs",
+                "type": "string",
+                "description": "The template IDs can be found on the program's page on the OT",
             },
```
antoine-le-calloch, line 369: Maybe add a small note explaining the expected format for template_ids (e.g., comma-separated).

skyportal/facility_apis/gemini.py
```diff
@@ -206,36 +184,63 @@ def _build_payload(self, request, session):
...
+
+            if (
+                isinstance(altdata.get("template_ids"), str | list)
+                and len(altdata.get("template_ids")) > 0
+            ):
+                template_ids = get_list_typed(
+                    altdata.get("template_ids"),
+                    int,
+                    "Invalid template IDs specified in altdata",
+                )
```
antoine-le-calloch, line 203: The altdata.get("template_ids") don't need to be process for each obsid. It can be retrieve before the loop. You can use `get_list_typed(request.payload.get("template_ids"), int)` to retrieve the obsids and then `get_list_typed(altdata.get("template_ids"),int)` to retrieve altdata template_ids.

skyportal/facility_apis/gemini.py
mcoughlin: ok I think I got this one.

skyportal/facility_apis/gemini.py
mcoughlin: added.

skyportal/facility_apis/gemini.py
mcoughlin: Here too.

skyportal/facility_apis/gemini.py
```diff
@@ -206,36 +184,58 @@ def _build_payload(self, request, session):
...
+        )
+
+        payloads = []
+        for obsid in obsids:
+            if len(template_ids) > 0 and obsid not in template_ids:
+                raise ValueError(
+                    f"Invalid template ID, must be one of: {str(template_ids)}"
+                )
+
+            obsnum = str(obsid).strip()
```
antoine-le-calloch: Since obsid is an integer, strip() isn't useful here, but it's not a big deal if it stays.

skyportal/facility_apis/gemini.py
```diff
@@ -300,7 +300,8 @@ def submit(request, session, **kwargs):
             log(traceback.format_exc())
             raise ValueError(f"Error building Gemini request: {e}")
 
-        r = requests.post(API_URL, verify=False, params=gemini_request.payload)
+        for payload in gemini_request.payload:
+            r = requests.post(API_URL, verify=False, params=payload)
 
         if r.status_code == 200:
             request.status = "submitted"
```
antoine-le-calloch, line 315: Here the request status will only be the status of the last post request. Maybe say something like submitted if at least 1 post work and like: "submitted for id X,Y,Z failure for id U,W", what do you think ?

antoine-le-calloch (commented):
One last comment (the strip() one don't count).

skyportal/facility_apis/gemini.py
mcoughlin: done

skyportal/facility_apis/gemini.py
mcoughlin: see new implementation

antoine-le-calloch (approved):

PR #5774, opened 2025-08-04
Docker build: rspack fix

Theodlz:
Fixes issue introduces by latest @mui/x-data-grid, that added some *.css file that rspack tried to load as a javascript one.


Theodlz:
self merging as this is an important fix we need for production

PR #5776, opened 2025-08-15
WIP - Finding Chart results cache + public cache access

Theodlz:
In this PR, we:
- add caching to finding chart results, optional and defaults to True
- add apublic handler (no auth required) to retrieve cached finding charts
- include public url to said handler in the response from the SourceFinderHandler
- add a "share link" button to the finder chart frontend to let users of SkyPortal share a link to the finding chart they just generated to external users. This link is available for as long as the cache is.
- the default cache validity window in the config is set to 30 days

Aside from the obvious benefit of caching finding charts to avoid doing the same work again and again, this lets folks share finding charts to people outside of the app, temporarily. This is very useful when sharing observation details to an observer which does not necessarily have access to the app.

In a following PR, the goal is to use this to send a link to a finding chart to Gemini observers (as part of the request payload), when triggering the instrument.

### DIFF
--- a/skyportal/handlers/public/finder.py
+++ b/skyportal/handlers/public/finder.py
@@ -1,0 +1,73 @@
+import io
+import traceback
+
+import numpy as np
+
+from baselayer.app.access import auth_or_token
+from baselayer.app.env import load_env
+from baselayer.log import make_log
+
+from ...utils.offset import finding_charts_cache
+from ..base import BaseHandler
+
+_, cfg = load_env()
+log = make_log("api/source")
+
+
+class CachedSourceFinderHandler(BaseHandler):
+    @auth_or_token
+    async def get(self, cache_key):
+        """
+        ---
+        summary: Retrieve a cached finding chart
+        description: Download a pre-generated PDF/PNG finding chart that has been cached
+        tags:
+          - sources
+          - finding chart
+          - public
+        parameters:
+          - in: path
+            name: cache_key
+            required: true
+            schema:
+              type: string
+        responses:
+          200:
+            description: A PDF/PNG finding chart file
+            content:
+              application/pdf:
+                schema:
+                  type: string
+                  format: binary
+              image/png:
+                schema:
+                  type: string
+                  format: binary
+          400:
+            content:
+              application/json:
+                schema: Error
+        """
+
+        try:
+            value = finding_charts_cache[cache_key]
+            if value is None:
+                return self.error("Finding chart not found in cache", status=404)
+
+            value = np.load(value, allow_pickle=True)
+            value = value.item()
+
+            filename = value["name"]
+            data = io.BytesIO(value["data"])
+            output_type = filename.split(".")[-1].lower()
+        except Exception as e:
+            # if its a value error with text "Source not found", we return a 404
+            if isinstance(e, ValueError) and str(e) == "Source not found":
+                return self.error("Source not found", status=404)
+
+            # otherwise, we log the error and return a 500
+            log(f"Error retrieving cached finding chart: {str(e)}")
+            traceback.print_exc()
+            return self.error(f"Error generating finding chart: {str(e)}")
+
+        await self.send_file(data, filename, output_type=output_type)
--- a/skyportal/utils/offset.py
+++ b/skyportal/utils/offset.py
@@ -4,6 +4,7 @@ import math
 import os
 import re
 import string
+import traceback
 import urllib
 import warnings
 from functools import wraps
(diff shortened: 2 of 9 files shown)


skyportal/facility_apis/gemini.py
```diff
@@ -173,22 +175,26 @@ def _build_payload(self, request, session):
         if notetitle:
             notetitle = str(notetitle).strip()
         if note:
-            note = str(note).strip()
+            note = str(note if note else "").strip()
+
+        note = note + f"(finder chart: {finding_chart_public_url})"
```
antoine-le-calloch: Small issue here, the `if note:` here seems to be a leftover.
A more compact idea can be:
```
note = request.payload.get("note") or ""
note = f"{str(note).strip()}(finder chart: {finding_chart_public_url})"
```

skyportal/handlers/public/finder.py
```diff
@@ -0,0 +1,73 @@
...
+            content:
+              application/json:
+                schema: Error
+        """
+
+        try:
+            value = finding_charts_cache[cache_key]
+            if value is None:
+                return self.error("Finding chart not found in cache", status=404)
+
```
antoine-le-calloch, line 56: To reach `if value is None:` we should use `finding_charts_cache.get(cache_key)` instead no ?

skyportal/utils/offset.py
```diff
@@ -1391,6 +1424,56 @@ def get_finding_chart(
...
+            fallback_image_source="ps1_cds",
+            zscale_contrast=0.045,
+            zscale_krej=2.5,
+            extra_display_string="",
+            **offset_star_kwargs,
+        )
+        value = finding_charts_cache[cache_key]
+        if value is not None:
+            try:
+                value = np.load(value, allow_pickle=True)
```
antoine-le-calloch, line 1447: Same here.

static/js/components/FindingChart.jsx
```diff
@@ -33,9 +37,6 @@ const useStyles = makeStyles((theme) => ({
     maxWidth: "100%",
     width: "95%",
   },
-  button: {
-    margin: theme.spacing(1),
-  },
   paper: {
     width: "100%",
     padding: theme.spacing(1),
```
antoine-le-calloch: Unused style:
  ```
  paper: {
    width: "100%",
    padding: theme.spacing(1),
    textAlign: "left",
    color: theme.palette.text.primary,
  },
  items: {
    width: "100%",
    minWidth: "100%",
    display: "flex",
    wrap: "nowrap",
  },
  ```

static/js/components/FindingChart.jsx
```diff
@@ -106,25 +108,31 @@ const FindingChart = () => {
   const [params, setParams] = useState({ ...initialFormState });
 
   const [image, setImage] = useState(null);
+  const [public_url, setPublicUrl] = useState(null);
```
antoine-le-calloch: Maybe use `publicUrl` instead ?

static/js/components/FindingChart.jsx
```diff
@@ -343,6 +351,23 @@ const FindingChart = () => {
                     >
                       Print
                     </Button>
+                    <Tooltip title="The public link is only valid temporarily.">
+                      <Button
+                        secondary
+                        className={classes.button}
+                        onClick={() => {
+                          navigator.clipboard.writeText(public_url);
```
antoine-le-calloch: `classes.button` have been delete.

skyportal/handlers/public/finder.py
Theodlz: `Cache` is not a dict but a custom class with a getter/setter, where the getter returns None when there are no values, so this is valid here

static/js/components/FindingChart.jsx
Theodlz: removed

static/js/components/FindingChart.jsx
Theodlz: removed

static/js/components/FindingChart.jsx
Theodlz: done! thx for the suggestion, it's more javascripty for sure

skyportal/facility_apis/gemini.py
antoine-le-calloch: @Theodlz I still see this issue, this can create a `None + str` if note is None.

Theodlz: Oh crap I missed this specific comment! Let me address it now. Sorry about that

Theodlz: added your suggestion, thanks!

PR #5777, opened 2025-08-18
fix: add a try catch to avoid error on the airmass_time calculation

camilledzt:
Fixing  the following sentry error : https://university-of-minnesota-gz.sentry.io/issues/6808630071/?alert_rule_id=15261570&alert_type=issue&notification_uuid=151021c4-8127-4a0f-9472-12eabea507ff&project=4507403282546688&referrer=slack

### DIFF
--- a/skyportal/handlers/api/instrument.py
+++ b/skyportal/handlers/api/instrument.py
@@ -414,13 +414,17 @@ class InstrumentHandler(BaseHandler):
             if localization_dateobs is not None:
                 try:
                     airmass_time = Time(arrow.get(localization_dateobs).datetime)
-                except:
-                    pass
+                except Exception as e:
+                    return self.error(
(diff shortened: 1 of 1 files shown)


skyportal/handlers/api/instrument.py
```diff
@@ -412,9 +412,15 @@ async def get(self, instrument_id=None):
 
         if airmass_time is None:
             if localization_dateobs is not None:
-                airmass_time = Time(arrow.get(localization_dateobs).datetime)
+                try:
+                    airmass_time = Time(arrow.get(localization_dateobs).datetime)
+                except:
+                    pass
```
mcoughlin: Shouldn't we return an error here? Rather than pass?

camilledzt: added :)

PR #5781, opened 2025-08-21
enh: improve TTT API for TTT-200

camilledzt:
This PR is adding camera choice for TTT-200 between iKon936 and QHY600M Pro.

image


mcoughlin (approved):
LGTM

PR #5782, opened 2025-08-21
Fix planned observations key error on load demo data

antoine-le-calloch:
- Resolve key error in post_survey_efficiency_analysis when loading demo data
- Delete forgotten console.log.

### DIFF
--- a/skyportal/handlers/api/observation.py
+++ b/skyportal/handlers/api/observation.py
@@ -2030,7 +2030,7 @@ def retrieve_observations_and_simsurvey(
     field = session.scalars(stmt).first()
     if field is None:
         raise ValueError(
-            'Missing field {obs_dict["field"]["id"]} required to estimate field size'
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

Theodlz (commented):
@antoine-le-calloch do you know that there is a difference between a field's id and a "field_id"? These are not necessarily the same thing, so just checking this isn't what you deleted here

Theodlz:
the git diff doesn't make it obvious if this is true here, but for example we have `InstrumentField` in the DB, these naturally have an `id` like any other tables, but they also have a `field_id` which is another number. For example when we add ZTF fields to the database, they all get their own unique `id` in postgres as usual, but they also have a `field_id` which is the id that comes from the input field grid, i.e. the "real" id of the fields and not some automated DB-generated one.

PR #5783, opened 2025-08-23
Enhance Upload Spectrum page

antoine-le-calloch:
- Enhance Upload Spectrum page style and code to try to avoid lagging.
- Create a custom Paper component with default 1rem padding.
   This custom component seems useless for now but there is many places where we duplicate that style just for this padding so this component can be use to replace this code everywhere.

### DIFF
--- a/static/js/components/spectrum/UploadSpectrum.jsx
+++ b/static/js/components/spectrum/UploadSpectrum.jsx
@@ -5,14 +5,12 @@ import { Link, useSearchParams } from "react-router-dom";
 import MUIDataTable from "mui-datatables";
 
 import Form from "@rjsf/mui";
-import validator from "@rjsf/validator-ajv6";
+import validator from "@rjsf/validator-ajv8";
 import { dataUriToBuffer } from "data-uri-to-buffer";
 import Typography from "@mui/material/Typography";
 import Accordion from "@mui/material/Accordion";
 import Grid from "@mui/material/Grid";
-import Paper from "@mui/material/Paper";
 import HelpOutlineIcon from "@mui/icons-material/HelpOutline";
-import makeStyles from "@mui/styles/makeStyles";
 import CircularProgress from "@mui/material/CircularProgress";
 import embed from "vega-embed";
 import dayjs from "dayjs";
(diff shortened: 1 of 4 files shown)


static/js/components/spectrum/UploadSpectrum.jsx
```diff
@@ -99,170 +67,150 @@ SpectrumPreview.displayName = "SpectrumPreview";
...
+    setUserEnumOptions({
+      enum: users.map((user) => user.id),
+      enumNames: users.map((user) => userLabel(user, true)),
+    });
+  }, [users]);
 
   // on page load or refresh, block until state.spectra.parsed is reset
   useEffect(() => {
-    const blockingFunc = async () => {
+    const unwrapQuotes = (str) =>
```
Theodlz: could maybe have these outside of the useeffect?

static/js/components/spectrum/UploadSpectrum.jsx
```diff
@@ -99,170 +67,150 @@ SpectrumPreview.displayName = "SpectrumPreview";
...
+      str && str.startsWith('"') && str.endsWith('"') ? str.slice(1, -1) : str;
+
+    const getIntList = (param) =>
```
Theodlz: same comment here

static/js/components/spectrum/UploadSpectrum.jsx
```diff
@@ -99,170 +67,150 @@ SpectrumPreview.displayName = "SpectrumPreview";
...
 
-      let file_url = searchParams.get("file_url");
-      if (file_url && file_url.startsWith('"') && file_url.endsWith('"')) {
-        file_url = file_url.slice(1, -1);
-      }
-      let file_name = searchParams.get("file_name");
-      if (file_name && file_name.startsWith('"') && file_name.endsWith('"')) {
-        file_name = file_name.slice(1, -1);
-      }
+      const [_, sourceResult] = await Promise.all([
```
Theodlz, line 127: Nice

Theodlz: we never use that pattern, but it's great

antoine-le-calloch:
@Theodlz Change done ✅

Theodlz:
> If you put the enumNames in the schema like in the example above, you will get a warning because this approach is deprecated. Now, you should put it in the uiSchema (since enum is data and enumNames is display, it makes more sense for them that way).

ahhhhh got it, yeah I suppose it does make sense that way too, maybe a little more work as one needs to go back and forth between the schema and ui schema, but that's alright. Thanks a lot for figuring this out!

Theodlz (approved):
LGTM, excellent work

PR #5785, opened 2025-08-25
Fix MMT parameter naming.

antoine-le-calloch:
[Fix validity range checks for start and end dates in follow-up requests. Check that request start and end date exist on the frontend and on the backend and take the current date if the request start date is not defined.

### DIFF
--- a/data/db_demo.yaml
+++ b/data/db_demo.yaml
@@ -750,8 +750,8 @@ allocation:
   - pi: Robert Stein
     proposal_id: WINTER-001
     validity_ranges:
-      - start_date: "2020-02-12T00:00:00"
-        end_date: "3030-07-12T00:00:00"
+      - start_date: "2020-02-12T00:00:00.000Z"
+        end_date: "3030-07-12T00:00:00.000Z"
       - start_date: "3030-08-01T00:00:00"
         end_date: "3030-12-31T00:00:00"
     hours_allocated: 100
@@ -761,8 +761,8 @@ allocation:
   - pi: Michael Coughlin
     proposal_id: SEDM-001
     validity_ranges:
-      - start_date: "2020-02-12T00:00:00"
-        end_date: "3030-07-12T00:00:00"
+      - start_date: "2020-02-12T00:00:00.000Z"
+        end_date: "3030-07-12T00:00:00.000Z"
     hours_allocated: 100
     group_id: =program_A
     instrument_id: =SEDM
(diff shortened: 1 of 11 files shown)


Theodlz (approved):
LGTM

skyportal/handlers/api/followup_request.py
```diff
@@ -590,17 +590,29 @@ def post_followup_request(
...
+                end_date is None
+                or arrow.get(range["start_date"])
+                <= arrow.get(end_date)
+                <= arrow.get(range["end_date"])
             ):
-                raise ValueError("Provided dates are outside allowed validity ranges.")
+                break
+        else:
+            raise ValueError(
+                f"{'Provided dates' if start_date else 'Current date'} are outside allowed validity ranges."
```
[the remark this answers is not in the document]
Theodlz: @antoine-le-calloch might want to have the `are` and `is` in the if statement options so it still works

static/js/components/followup_request/FollowupRequestForm.jsx
```diff
@@ -287,14 +287,21 @@ const FollowupRequestForm = ({
     if (
       !isSomeActiveRangeOrNoRange(
         allocationLookUp[selectedAllocationId].validity_ranges,
-        new Date(formData.start_date),
+        new Date(formData.start_date || Date.now()),
```
Theodlz, line 290: do you need to put Date.now() IN Date()??? Can't it just be outside it. Also isn't it an issue that formData.start_date is likely UTC time (or at least some string that's agnostic to timezones), whereas Date.now() is likely local time?

antoine-le-calloch:
```
export const isSomeActiveRangeOrNoRange = (ranges, date = new Date()) => {
  return !ranges?.length || ranges.some((range) => rangeIsActive(range, date));
};

export const rangeIsActive = (range, date = new Date()) =>
  range.start_date <= date.toISOString() &&
  range.end_date >= date.toISOString();
```

@Theodlz In the check I'm doing `toISOString` to avoid this kind of issue.

Theodlz: Got it

antoine-le-calloch: > do you need to put Date.now() IN Date()

This its because if `formData.start_date` is NULL new Date will break.

PR #5786, opened 2025-08-29
NEWFIRM

mcoughlin:
This PR adds the Blanco NEWFIRM instrument as an API option.


Theodlz (approved):
LGTM! Not a comment for this PR specifically but here is SO MUCH duplication between API classes, we should look for patterns to reduce these in the future. Would make the whole thing easier to maintain and expand with new instruments.

Merge once green

Theodlz:
@mcoughlin the migration is messed up :)

PR #5810, opened 2025-09-05
feat: adding a new ACL to manage photometry

camilledzt:
The goal of this PRT is to add a new ACL called `Manage Photometry`. 
Any user with the `Manage Photometry` right can edit and delete photometry points for any source belonging to the group they are a member of.

This was an issue because `Full User` with `Upload Data` right or a `Group Admin`  can only modify photometry points they are owning. One specific limitation was related to GRANDMA because photometry points are imported from STDpipe and the owner is always the same person.

### DIFF
--- a/static/js/components/gcn/GcnSelectionForm.jsx
+++ b/static/js/components/gcn/GcnSelectionForm.jsx
@@ -43,6 +43,7 @@ import SourceTable from "../source/SourceTable";
 import ProgressIndicator from "../ProgressIndicators";
 
 import * as localizationActions from "../../ducks/localization";
+import Spinner from "../Spinner";
 
 const GcnReport = React.lazy(() => import("./GcnReport"));
 const GcnSummary = React.lazy(() => import("./GcnSummary"));
@@ -51,31 +52,6 @@ dayjs.extend(relativeTime);
 dayjs.extend(utc);
 
 const useStyles = makeStyles(() => ({
-  select: {
-    width: "25%",
-  },
-  container: {
-    width: "99%",
-    marginBottom: "1rem",
-  },
-  selectItem: {
-    whiteSpace: "break-spaces",
-  },
-  localizationSelect: {
-    width: "100%",
-  },
-  localizationSelectItem: {
-    whiteSpace: "break-spaces",
-  },
-  instrumentSelect: {
-    width: "100%",
-  },
-  instrumentSelectItem: {
-    whiteSpace: "break-spaces",
-  },
-  form: {
-    marginBottom: "1rem",
-  },
   formGroup: {
     display: "grid",
     gridTemplateColumns: "repeat(auto-fit, minmax(8rem, 1fr))",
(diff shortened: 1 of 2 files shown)


skyportal/models/photometry.py
```diff
@@ -26,6 +30,22 @@
...
+        access_control = (
+            accessible_by_groups_members
+            | accessible_by_streams_members
+            | accessible_by_owner
+        )
+        return access_control.query_accessible_rows(cls, user_or_token)
+    else:
+        return accessible_by_owner.query_accessible_rows(cls, user_or_token)
+
+
```
Theodlz, line 50: I think this should be:
- if the user is an admin, just return the rows there is nothing to enforce
- if the user has "Manage Photometry" check if they the 3 rules in your above code are true
- otherwise just check if it's accessible by owner

What's really quite wrong here is that the `is_admin` results in permissions being enforced, when that's precisely when they should be skipped

skyportal/handlers/api/photometry.py
```diff
@@ -1431,6 +1432,13 @@ def put(self):
                                 added in request. Can be used to later delete all
                                 points in a single request.
         """
+        user_permissions = self.current_user.permissions
```
Theodlz: All `Full users` that the "Upload data" permission already, so why not just use the `permission` decorator as before?

Theodlz: It feels like Upload data should always be enforced for this endpoint, and that "Manage Photometry" is just useful at the database model level to allow users to edit photometry even when they aren't the owner. I'm not sure we need to have it here in the handler

camilledzt: I have replaced the `@permissions` decorator because otherwise if I put `@permissions(["Upload data", "Manage photometry"]),` then the user has to have both of them to edit or delete a point and in my case I want an OR. Should I implement an other decorator in baselayer to fit the needs ? Or is there a subtlety that I miss here ?

skyportal/models/photometry.py
camilledzt: Ok I will fix this

skyportal/handlers/api/photometry.py
Theodlz: I think just decorating with @permissions(["Upload data") is enough

Theodlz: I think that permission should always be needed, and all Full users have it anyways

Theodlz: the manage photometry is enforced in the data model, just as you did.

Theodlz: so users require upload photometry to use this endpoint (i.e., can they interact with photometry to edit/delete), and which data they can access is dependent on the manage photometry  (i.e. can see edit/delete THESE specific photometry points). One of them is about can they use the feature, the other is about what data points can they use the feature on.

camilledzt: I understand, I fixed it

skyportal/models/photometry.py
```diff
@@ -26,6 +31,24 @@
...
+    for objects accessible to their groups.
+    """
+    if user_or_token.is_admin:
+        return public.query_accessible_rows(cls, user_or_token)
+    elif "Manage photometry" in user_or_token.permissions:
+        access_control = (
+            accessible_by_groups_members
+            | accessible_by_streams_members
+            | accessible_by_owner
+        )
```
Theodlz: Are these scoped by function call? I guess I'm wondering if the `OR` statement (`access_control` variable) should be defined here every time the function is called, or above in the file and then simply called here. Would you mind giving that a go? Fetching photometry is a VERY common operation, and any redundant operation should be avoided.

camilledzt: `accessible_by_groups_members`, `accessible_by_streams_members` and `accessible_by_owner` are defined as constant, so I will do the same for `access_control` :)

skyportal/models/photometry.py
Theodlz: Awesome, thanks!

PR #5811, opened 2025-09-05
Remove useless fixed position dialog

antoine-le-calloch:
_Part of the major front-end improvements I made last month._

- Remove all the useless fixed position on the dialog component
- Delete forgotten empty hermes file.


mcoughlin (approved):
LGTM

Theodlz:
Why is "fixed" position useless? I thought it was there for a reason...

Theodlz:
according to the docs: `position: 'fixed' ensures the dialog's position is relative to the viewport and does not scroll with the page.`

antoine-le-calloch:
> Why is "fixed" position useless? I thought it was there for a reason...

The position of the Mui Dialog is already fixed by default.

PR #5812, opened 2025-09-12
Feat: add extinction calculation to the Photometry table

camilledzt:
The goal of this PR is to add some field to the photometry table to compute the extinction, the corrected magnitude and the corrected flux.

Each user can activate or not the extinction calculation using a button. Default is set to False.

The extinction law used is the G23 law. By default the Rv coefficient is set to 3.1

This PR is adding a new requirement :  dust_extinction
Dust maps will be download in the /tmp directory.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -57,6 +57,7 @@ module.exports = [
       "jsx-a11y/control-has-associated-label": 0,
       "react-hooks/rules-of-hooks": "error",
       "react-hooks/exhaustive-deps": "warn",
+      "react-hooks/set-state-in-effect": "off",
       "react/jsx-wrap-multilines": 0,
       "react/jsx-one-expression-per-line": 0,
       "react/jsx-props-no-spreading": 0,


Theodlz:
Looks like nothing more was added since I last looked at this, is this ready @camilledzt ?

camilledzt:
Yes it's ready, sorry for the delay

skyportal/handlers/api/photometry.py
```diff
@@ -1938,6 +1958,10 @@ def get(self, obj_id):
             include_annotation_info = True
         else:
             include_annotation_info = False
+        if str(include_extinction).lower() in ["true", "t", "1"]:
+            include_extinction = True
+        else:
+            include_extinction = False
```
Theodlz: @antoine-le-calloch introduced a helper function to standardize parsing boolean arguments, and updated multiple handlers to use it. Maybe this would be a good opportunity to update the photometry get handler to use that too like the other handlers?

skyportal/models/photometry.py
```diff
@@ -445,6 +446,31 @@ def snr(self):
             else_=None,
         )
 
+    @hybrid_property
+    def extinction(self):
+        """Galactic extinction A_lambda in magnitudes for this filter."""
+        if all(
+            coord is not None and not np.isnan(coord) for coord in [self.ra, self.dec]
+        ):
+            return calculate_extinction(self.ra, self.dec, str(self.filter))
```
Theodlz: I might benefit from someone's input (who's more of an astronomer than I am :sweat_smile:), but my understanding is that extinction/reddening isn't just a 2D thing. When you don't have a distance, best to use a 2D map, but when you do then 3D maps can be used. Would it be better to have some if else statement here, where if a distance is defined then we can make use of it in this calculation?

Theodlz: also unrelated, but is this `calculate_extinction` method performant/fast? Asking because this is ran in a loop over lots of photometry points

skyportal/utils/extinction.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    try:
+        # Get extinction coefficient A_λ/E(B-V)
+        coeff = get_extinction_coefficient(filter_name, Rv=Rv, Ebv=1.0)
+
+        # Get E(B-V) from SFD dust maps
+        ebv = _calculator.get_ebv(ra, dec)
+
+        # Calculate A_λ = (A_λ/E(B-V)) * E(B-V)
+        extinction = coeff * ebv
```
Theodlz, line 107: Just curious about this, as I'm not familiar with this process. It looks like you first compute the coeff by using ebv=1.0, then compute the ebv and multiply the coefficient with it. Any reason not to just compute the ebv first and feed that to the function computing the coeff?

Theodlz (commented):
Hey @camilledzt, a few questions/comments!

skyportal/handlers/api/photometry.py
camilledzt: done :)

skyportal/utils/extinction.py
camilledzt: I have followed the implementation Sergey has made for Fink. And then I asked the same question to Julien and we have looked the code. I think it is just a way to separate the shape and the amplitude of the extinction curve using the normalized curve once and then apply a coef.

skyportal/models/photometry.py
camilledzt: for your first question, I don't think I have the answer but yes I think it's not a 2D thing because we have to take in consideration the dust at a position ?

I don't know what is your expectation about performance but here some numbers:
The first call to calculate_extinction will be slow (~30 sec) because we have to download the dust maps. Then a call is approximately 20 ms. What do you think ?

JulienPeloton: Regarding the 2D/3D thing -- we mostly care about extra-galactic stuff, and extinction is mostly from our galaxy. In that case 2D is only what we want? (integration of all the galactic dust over the line of sight).

Theodlz: > for your first question, I don't think I have the answer but yes I think it's not a 2D thing because we have to take in consideration the dust at a position ?
> 
> I don't know what is your expectation about performance but here some numbers: The first call to calculate_extinction will be slow (~30 sec) because we have to download the dust maps. Then a call is approximately 20 ms. What do you think ?

Sorry is this 20ms PER photometry point being corrected, or is that the total time that the handler takes for some lightcurve to be returned with the correction. If the latter, than yes its fine.

Theodlz: > Regarding the 2D/3D thing -- we mostly care about extra-galactic stuff, and extinction is mostly from our galaxy. In that case 2D is only what we want? (integration of all the galactic dust over the line of sight).

I'm honestly not too sure myself, which is why I asked. In any case let's start with the 2D case and we can worry about something more complex later, if we want to.

Theodlz:
> * I was thinking that maybe it will compute the same value a large number of time ?

I think I'm worried about that statement in general, in this PR. Most of the detections will all have the same position, or the variation in position will be minimal. I guess that's not true for high proper motion stars or moving objects, but do we care about those here? Still probably ok for the former, and the latter idk if extinction is even relevant.

So I guess I'm wondering: should we use the RA/Dec of each photometry point (that will be null, except for alert-based photometry or the handful of pipelines that send to SkyPortal data back that includes positions), or just the RA/Dec of the source? If we use the RA/Dec of the source, then the correction factor needs to be computed once (and could be a parameter of the `Obj` class), and then the `Photometry` could have a function that takes that factor as input and corrects the magnitude?

That or a fallback, as you suggested. But yeah the same value will be computed many times, but isn't that already true with the current implementation and how little the position changes from one detection to another?

skyportal/handlers/api/photometry.py
```diff
@@ -380,14 +393,12 @@ def serialize(
                 "limiting_mag": maglimit_out,
             }
 
-            if extinction:
+            if extinction_dict is not None:
```
Theodlz, line 395: Since you now check if the dict is not None in all your "ifs" except the first one where you also check if `extinction` param is `True`, do we even need that extinction boolean parameter? Best to remove it if we anyways have extinction_dict as a proxy, that's already what you use most of the time

static/js/components/photometry/PhotometryExtinction.jsx
```diff
@@ -15,7 +17,17 @@ const PhotometryExtinction = ({ showExtinction, setShowExtinction }) => {
             data-testid="photometry_extinction_toggle"
           />
         }
-        label="Extinction"
+        label={
+          <span style={{ display: "flex", alignItems: "center", gap: "4px" }}>
+            Compute extinction
+            <Tooltip
+              title="The extinction law used is the G23 law. By default the Rv coefficient is set to 3.1"
```
Theodlz, line 24: Nice addition, folks were already asking about this so that's perfect

Theodlz (approved):
Just one change requested: I think we don't need both `extinction` (bool) and `extinction_dict` (dict or None) parameters. Once that change is made, this is ready to be merged

PR #5815, opened 2025-09-20
Fix "new gcn event" form issue

antoine-le-calloch:
- Refactor error handling in NewGcnEvent form validation to avoid error: `TypeError: can't access property "addError", errors.dateobs is undefined`.
- Push the error `Either (i) ra, dec, and error or (ii) polygon or (iii) skymap must be defined if not uploading VOEvent / JSON` to the base form instead of the skymaps field for consistency.
- Refactor NewGcnEvent component to simplify form data handling, improve readability and delete duplicate info.

### DIFF
--- a/static/js/components/gcn/NewGcnEvent.jsx
+++ b/static/js/components/gcn/NewGcnEvent.jsx
@@ -12,9 +12,7 @@ import * as gcnTagsActions from "../../ducks/gcnTags";
 
 const NewGcnEvent = ({ handleClose = null }) => {
   const dispatch = useDispatch();
-
-  let gcnTags = [];
-  gcnTags = gcnTags.concat(useSelector((state) => state.gcnTags));
+  const gcnTags = useSelector((state) => state.gcnTags) || [];
   gcnTags.sort();
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


static/js/components/gcn/NewGcnEvent.jsx
```diff
@@ -49,9 +43,7 @@ const NewGcnEvent = ({ handleClose = null }) => {
     const result = await dispatch(submitGcnEvent(formData));
     if (result.status === "success") {
       dispatch(showNotification("GCN Event saved"));
-      if (handleClose) {
-        handleClose();
-      }
+      handleClose?.(); // Call handleClose if it's provided
```
Theodlz, line 46: wow I didn't know that `?` worked with function calls, that's awesome!

static/js/components/gcn/NewGcnEvent.jsx
```diff
@@ -67,7 +59,7 @@ const NewGcnEvent = ({ handleClose = null }) => {
     }
     if (!(formData.xml || formData.json)) {
       if (!formData.dateobs) {
-        errors.dateobs.addError(
+        errors.addError(
```
Theodlz, line 62: I'm surprised by that, I thought the errors needed to be added on the associated field, but yeah global errors are fine too

antoine-le-calloch: The issue with that is that dateobs wasn't set by default in the errors object this is why the form was break.

PR #5836, opened 2025-09-26
enh: add documentation on the tag handlers

camilledzt:
The goal of this PR is to add documentation on the tag option and objets tag association API handlers.

### DIFF
--- a/skyportal/handlers/api/tag.py
+++ b/skyportal/handlers/api/tag.py
@@ -29,7 +29,7 @@ class ObjTagOptionHandler(BaseHandler):
         summary: Retrieve all tag options
         description: Retrieve all available tag options accessible to the user
         tags:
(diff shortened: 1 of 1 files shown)


Theodlz (commented):
Thanks @camilledzt. My only comment here will be that `-tags` tell openapi in what menu/subsection of the API docs yours endpoint will live. Here you use `object tags` and `tag options`, but I don't think this feature warrants adding 2 new menus of their own. I'd suggest using only one thing `object tags`, if not consider using an existing tag, e.g. `sources`.

What I would so, given how busy the `sources` section already is, is just use a tag `source tags` on all these endpoints. That way it is it's own menu and doesn't get lost in the current `sources` one, but is co-located with it which makes it easy enough to find.

PR #5837, opened 2025-09-30
Fix mjd key error

antoine-le-calloch:
- Fix "KeyError" in PhotometryHandler in sentry-backend.

- Fix `cannot find module 'versor/src/index'` error by downgrading `d3-geo-zoom` to "1.4.4"


antoine-le-calloch:
Your welcome !

PR #5838, opened 2025-10-01
Fix function nested too deeply in gcn event page

antoine-le-calloch:
- simplify GcnSelectionForm on the gcn event page
- delete useless code
- fix `function nested too deeply` error on Firefox and lagging on every other browser.


package.json
```diff
@@ -34,7 +34,7 @@
     "d3-array": "3.2.4",
     "d3-geo": "3.1.1",
     "d3-geo-projection": "4.0.0",
-    "d3-geo-zoom": "1.5.5",
+    "d3-geo-zoom": "1.4.4",
```
mcoughlin, line 37: Can you open an issue tracking this (linking to your PR on that repo)?

antoine-le-calloch: Done

PR #5858, opened 2025-10-01
Fix function nested too deeply in form

camilledzt:
The goal of this PR is to fix the error "function nested to deeply".

I revert my commit on the file  `static/js/components/gcn/GcnSelectionForm.jsx` as it is already fixed by @antoine-le-calloch on the PR https://github.com/skyportal/skyportal/pull/5838

### DIFF
--- a/static/js/components/gcn/GcnSelectionForm.jsx
+++ b/static/js/components/gcn/GcnSelectionForm.jsx
@@ -43,6 +43,7 @@ import SourceTable from "../source/SourceTable";
 import ProgressIndicator from "../ProgressIndicators";
 
 import * as localizationActions from "../../ducks/localization";
+import Spinner from "../Spinner";
 
 const GcnReport = React.lazy(() => import("./GcnReport"));
 const GcnSummary = React.lazy(() => import("./GcnSummary"));
@@ -51,31 +52,6 @@ dayjs.extend(relativeTime);
 dayjs.extend(utc);
 
 const useStyles = makeStyles(() => ({
-  select: {
-    width: "25%",
-  },
-  container: {
-    width: "99%",
-    marginBottom: "1rem",
-  },
-  selectItem: {
-    whiteSpace: "break-spaces",
-  },
-  localizationSelect: {
-    width: "100%",
-  },
-  localizationSelectItem: {
-    whiteSpace: "break-spaces",
-  },
-  instrumentSelect: {
-    width: "100%",
-  },
-  instrumentSelectItem: {
-    whiteSpace: "break-spaces",
-  },
-  form: {
-    marginBottom: "1rem",
-  },
   formGroup: {
     display: "grid",
     gridTemplateColumns: "repeat(auto-fit, minmax(8rem, 1fr))",
@@ -99,24 +75,10 @@ const useStyles = makeStyles(() => ({
   formContainer: {
     maxWidth: "95vw",
     width: "100%",
+    marginTop: "0.3rem",
   },
   formContainerItem: {
     maxWidth: "87vw",
-    width: "100%",
-  },
-  dateGroup: {
-    display: "grid",
-    gridGap: "1rem",
-    gridTemplateColumns: "repeat(auto-fit, minmax(40%, 1fr))",
-    justifyContent: "space-evenly",
-    alignItems: "center",
-    width: "100%",
-  },
-  localizationPlot: {
-    display: "flex",
-    flexDirection: "column",
-    justifyContent: "center",
-    alignItems: "center",
     width: "100%",
   },
   localizationPlotSmall: {
@@ -126,9 +88,6 @@ const useStyles = makeStyles(() => ({
     alignItems: "center",
     maxWidth: "90vw",
     width: "100%",
-  },
-  marginBottom: {
-    marginBottom: "1rem",
   },
   buttons: {
     display: "grid",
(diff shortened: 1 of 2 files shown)


static/js/components/photometry/NewPhotometry.jsx
```diff
@@ -204,6 +200,14 @@ const NewPhotometryForm = ({ obj_id }) => {
...
   };
 
+  const uiSchema = {
+    group_ids: {
+      items: {
+        "ui:enumNames": groups.map((group) => group.name),
+      },
+    },
+  };
+
```
antoine-le-calloch, line 210: Why using `items` here ?

static/js/components/release/ReleaseForm.jsx
```diff
@@ -91,6 +87,11 @@ const ReleaseForm = ({ release, setRelease, setOpenReleaseForm }) => {
             rows: 3,
           },
         },
+        group_ids: {
+          items: {
+            "ui:enumNames": groups.map((group) => group.name),
+          },
+        },
       }}
```
antoine-le-calloch, line 95: Same here.

static/js/components/sharing_service/SharingServicesForm.jsx
```diff
@@ -330,6 +318,26 @@ const SharingServicesDialog = ({ obj_id, dialogOpen, setDialogOpen }) => {
...
+        ),
+      },
+    },
+    stream_ids: {
+      items: {
+        "ui:enumNames": streams.map((stream) => stream.name),
+      },
+    },
+  };
+
```
antoine-le-calloch, line 340: Same.

static/js/components/spectrum/UploadSpectrum.jsx
```diff
@@ -526,6 +519,18 @@ const UploadSpectrumForm = ({ route }) => {
...
+        ? instruments.map(
+            (instrument) =>
+              `${
+                telescopes.find((t) => t.id === instrument.telescope_id)
+                  ?.nickname
+              } / ${instrument.name}`,
+          )
+        : [],
+      "ui:disabled": !instruments?.length,
+    },
```
antoine-le-calloch, line 530: Here Instruments cannot be null, so maybe you can simplify this.

static/js/components/sharing_service/SharingServicesPage.jsx
```diff
@@ -1355,6 +1343,21 @@ const SharingServicesPage = () => {
...
+              owner_group_ids: {
+                "ui:enumNames": (groups || [])
+                  .sort((a, b) => a?.name?.localeCompare(b?.name))
+                  .map((group) => group.name),
+              },
+              instrument_ids: {
+                "ui:enumNames": (allowedInstruments || []).map(
+                  (instrument) => instrument.name,
+                ),
+              },
```
antoine-le-calloch, line 1360: Here allowedInstruments cannot be null, and it seems to me that groups cannot either, but it might not be a bad idea to keep it just in case. I see in the code:
- Back end always send it:
```
info["user_accessible_groups"] = sorted(
  (
    g
    for g in self.current_user.accessible_groups
    if not g.single_user_group
  ),
  key=lambda g: g.name.lower(),
)
```
- In the reducer it is empty by default:
`function reducer(state = { user: [], userAccessible: [], all: null }, action) { `
- `const groups = useSelector((state) => state.groups.userAccessible);`

@Theodlz , could there be any side effects or coding practices that would justify using this regardless? Because I see many places where we don't check and many where we do check.

static/js/components/sharing_service/SharingServicesPage.jsx
```diff
@@ -1053,11 +1049,7 @@ const SharingServicesPage = () => {
           type: "array",
           items: {
             type: "integer",
-            anyOf: (allowedInstruments || []).map((instrument) => ({
-              enum: [instrument.id],
-              type: "integer",
-              title: instrument.name,
-            })),
+            enum: (allowedInstruments || []).map((instrument) => instrument.id),
```
antoine-le-calloch: Same here (cannot be null).

antoine-le-calloch (commented):
Some comment.

static/js/components/photometry/NewPhotometry.jsx
camilledzt: I see in the [doc](https://rjsf-team.github.io/react-jsonschema-form/docs/json-schema/arrays) that it was used like that. But if you think it's not needed I can remove it ?

antoine-le-calloch: Oh okay, I didn’t know that, let’s go with that then. But my comment was more about the fact that you used it in some places but not in others. Like in "GenerateReportForm.jsx" or in "ClassificationForm.jsx".

antoine-le-calloch: Can you fix the one I have add in the `UploadSpectrum.jsx‎` also please, if you have time.

static/js/components/sharing_service/SharingServicesPage.jsx
Theodlz: I mean it's always best to handle edge cases rather than assuming they won't happen, so I'm fine with Camille's approach. It's a bit more verbose sure, but I don't think that's a huge problem.

PR #5865, opened 2025-10-24
enh: add a way to remove expiration date to reactivate account and add a sorting system on the user management page

camilledzt:
The goal of this PR is to add:
* A button in the user management page to be able to remove an expiration date.
* A way to filter user by creation date
* A switch button to show or not expired users
* A new duck `users_management.js` to separate the list of the users used everywhere in the app and the list of users that can be filtered in the User management page.


Theodlz (approved):
Thanks for the added awaits @camilledzt. LGTM. I'll merge once the CI is green

PR #5866, opened 2025-10-28
enh: add filters to the options of the download of the photometry table

camilledzt:
This was requested by Marion Pillas to help GRANDMA to filter data they download


Theodlz (approved):
LGTM, thanks @camilledzt!

PR #5868, opened 2025-10-29
feat: add hermes listener external service

camilledzt:
The goal of this service is to synchronize data from a Hermes topic to a Skyportal instance.

  - Consumes source data from Hermes Kafka topics (via SCiMMA)
  - Creates or updates astronomical objects (Obj) and saves them to configured groups (Source)
  - Imports photometry from multiple instruments with deduplication
  - Uses TNS IDs for instrument identification to ensure compatibility across different SkyPortal instances
       - hermes submission script has been updated to integrate when possible, the tns id of the instrument to avoid doing the mapping with the instrument name but with the tns instrument id.
  
In ICARE, the tns_id colomn  on the instrument table is empty so we need to populate it, I added a script to populate the database in this PR.


services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,693 @@
...
+    """Main service entry point with app loading check."""
+    hermes_sync_cfg = cfg.get("app.hermes.sync", {})
+
+    # Kafka/SCiMMA configuration
+    username = hermes_sync_cfg.get("scimma_username")
+    password = hermes_sync_cfg.get("scimma_password")
+    server_url = hermes_sync_cfg.get("kafka_server", "kafka.scimma.org")
+
+    topic = cfg.get("app.hermes.topic", "skyportal.skyportal_test")
+
```
antoine-le-calloch, line 597: We don't want to hard code any topic name. You should raise an error if no topic is available.

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,693 @@
...
+    # Service configuration
+    from_start = hermes_sync_cfg.get("from_start", False)
+    max_age_days = hermes_sync_cfg.get("max_age_days")
+    bot_user_id = hermes_sync_cfg.get("bot_user_id", 1)
+    group_ids = hermes_sync_cfg.get("group_ids", [1])
```
antoine-le-calloch: Same here.

antoine-le-calloch (approved):
LGTM

antoine-le-calloch:
@Theodlz can you review/merge this, please? I've already reviewed it and it looks good on my side.

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+        Right ascension in degrees
+    dec : float
+        Declination in degrees
+    """
+    obj = session.scalar(sa.select(Obj).where(Obj.id == obj_id))
+
+    if obj is None:
+        obj = Obj(id=obj_id, ra=ra, dec=dec)
+        session.add(obj)
+        session.flush()
```
Theodlz: why do we need to flush?

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+                group_id=group_id,
+                saved_by_id=user_id,
+                active=True,
+            )
+            session.add(source)
+            sources_created += 1
+            log(f"Created Source for {obj_id} in group {group_id}")
+
+    if sources_created > 0:
+        session.flush()
```
Theodlz, line 140: same question about the flushing

Theodlz: though here perhaps it's to get the `id` from the DB. In which case I see why it's useful for groups since its an autoincremented integer we need the DB to assign, but then we don't need it for the `Obj` where the id is a string we set ourselves, right?

Theodlz: (see above comment about the Obj)

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+    existing_keys = {
+        (
+            safe_round(p.mjd, 7),
+            p.filter,
+            safe_round(p.flux, 5),
+            safe_round(nan_to_none(p.fluxerr), 5),
+        )
+        for p in existing_phot
+    }
+
```
Theodlz, line 204: would be good to explain what motivates the number of digits at which you are rounding

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+        origin = point.get("origin")
+
+        if date_obs is None or brightness is None or not bandpass:
+            continue
+
+        jd = float(date_obs)
+        mjd = safe_round(jd - 2400000.5, 7)
+
+        mag = float(brightness)
+        flux = PhotometricSeries.mag2flux(np.array([mag]))[0]
```
Theodlz: feels a little odd to involved photometricseries for this stuff? If we don't have helpers for mag -> flux conversions, maybe that's what should be added to the `utils` and then imported by the photometric series and this new script.

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+
+        if brightness_error is not None:
+            magerr = float(brightness_error)
+            fluxerr = PhotometricSeries.magerr2fluxerr(
+                np.array([mag]), np.array([magerr])
+            )[0]
```
Theodlz: same comment here about using a DB model we don't need other than for some helper function.

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+        else:
+            fluxerr = np.nan
+
+        # Check for duplicate : try to match mjd, filter, flux and fluxerr
+        duplicate_key = (
+            mjd,
+            bandpass,
+            safe_round(flux, 5),
+            safe_round(nan_to_none(fluxerr), 5),
+        )
```
Theodlz, line 235: why are we not checking for instrument id here? Can't we, in principle, have 2 instruments measure something at the same time? I agree that it is VERY unlikely though, so I'm fine deduplicating just based on these keys! I'm just curious about the choice that was made that's all.

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+        if duplicate_key in existing_keys:
+            duplicate_count += 1
+            continue
+
+        altdata = {
+            "imported_from": "hermes",
+        }
```
Theodlz, line 242: nice!

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+
+        for group in groups:
+            if group not in phot.groups:
+                phot.groups.append(group)
+
+        added_count += 1
+        existing_keys.add(duplicate_key)
+
+    if added_count > 0:
+        session.flush()
```
Theodlz: same q about flushing. I'm just surprised by how often it's done in this new service but we don't seem to do it in most other places in the codebase.

services/hermes_skyportal_sync/hermes_skyportal_sync.py
```diff
@@ -0,0 +1,686 @@
...
+
+    return added_count, duplicate_count
+
+
+class SourceProcessor:
```
Theodlz, line 267: nice!

tools/populate_tns_ids/populate_tns_ids.py
```diff
@@ -0,0 +1,181 @@
...
+import json
+import sys
+from pathlib import Path
+
+import sqlalchemy as sa
+
+from baselayer.app.env import load_env
+from skyportal.models import DBSession, Instrument, init_db
+
+DATA_FILE = Path(__file__).parent / "tns_instruments.json"
```
Theodlz, line 24: how was the JSON generated? by calling some API from TNS, or manually? If its calling some API, I'd suggesting letting the script do just that rather than hardcoding a JSON into the codebase. TNS sees new instruments added every once in a while, and letting the script build the JSON at runtime prevents you from having to open PRs every time someone adds an instrument to TNS.

config.yaml.defaults
```diff
@@ -105,6 +105,15 @@ app:
...
       test_topic: skyportal.skyportal_test
+      sync:
+        scimma_username: # SCiMMA/Kafka username for authentication
+        scimma_password: # SCiMMA/Kafka password for authentication
+        kafka_server: kafka.scimma.org
+        from_start: False  # Start from beginning of topic (true) or latest message (false)
+        max_age_days:  # Skip messages older than N days (null/empty = no limit)
+        bot_user_id:   # User ID of the bot account that will save sources and own photometry
+        group_ids: []  # List of group IDs to assign imported sources to
+        instrument_tns_ids: # List of TNS IDs of instruments for imported photometry
```
Theodlz: what is this meant to do? Is this some kind of "whitelist" for which photometry we want to import? If so, please comment that here.

config.yaml.defaults
```diff
@@ -105,6 +105,15 @@ app:
       token:
       topic: skyportal.skyportal
       test_topic: skyportal.skyportal_test
+      sync:
+        scimma_username: # SCiMMA/Kafka username for authentication
+        scimma_password: # SCiMMA/Kafka password for authentication
+        kafka_server: kafka.scimma.org
+        from_start: False  # Start from beginning of topic (true) or latest message (false)
```
Theodlz: true and false should be capitalized here

config.yaml.defaults
```diff
@@ -105,6 +105,15 @@ app:
...
+        max_age_days:  # Skip messages older than N days (null/empty = no limit)
+        bot_user_id:   # User ID of the bot account that will save sources and own photometry
```
Theodlz: what is the default if not specified?

Theodlz (commented):
I have some comments, but otherwise this looks great! I kind of wonder if the tns_id of the instrument table should be auto-populated by this service, but perhaps best to keep it as a separate script (like you did) for now.

services/hermes_skyportal_sync/hermes_skyportal_sync.py
camilledzt: I was having this reasoning with the filter, because instruments used different filters but since there are some filters that are the same from different instruments, i don't know ? should I replace filter by the instrument id ?

services/hermes_skyportal_sync/hermes_skyportal_sync.py
camilledzt: well, it's an old PR but I remember to have some trouble with the deduplication. I was testing multuple options. 7 digits for mjd is equivalent to round at the ms if I'm write. So I think that's a good value for an observation time.
For the flux, I chose the value arbitrarily, do you have an opinion on that ?

services/hermes_skyportal_sync/hermes_skyportal_sync.py
camilledzt: I removed it because yes I was for the id for the source but not needed for obj

services/hermes_skyportal_sync/hermes_skyportal_sync.py
camilledzt: I have created a util file with photometry conversion

config.yaml.defaults
camilledzt: done

config.yaml.defaults
camilledzt: Yes it is, I change the comment

config.yaml.defaults
camilledzt: The default behavior is that the service will not start without specifying bot_user_id

tools/populate_tns_ids/populate_tns_ids.py
camilledzt: I've added an option to do that

services/hermes_skyportal_sync/hermes_skyportal_sync.py
Theodlz: for the flux, that might be a good question for a real astronomer to chime in 😉 @mcoughlin ?

config.yaml.defaults
Theodlz: we should clearly have some documentation comment here! It's a new feature that not all skyportal maintainers will want to opt in.

Theodlz (commented):
Looks good overall! One request of mine: I think this is a cool feature but for folks to know about it and use it correctly, it would be great to document it. Right now the code for the microservice is there, a default config is there, but there isn't much of anything that tells a person setting up their skyportal what this is for and how to set it up (where to get hermes credentials, etc). I'd suggest adding a `.md` file to the docs for this, and documenting it a bit in the config.yaml.defaults (that could refer to the newly added docs). Does that sound good?

config.yaml.defaults
mcoughlin: Is the flux and flux error done as a log? Otherwise I worry about just having 0.0000 depending on the units.

camilledzt: the flux and flux error are linear here :)

yes, I've added some doc

mcoughlin: What happens if flux is 0.0000000145 then?

camilledzt: It would be treated as duplicates. What to you want the behavior to be ? @mcoughlin

services/hermes_skyportal_sync/hermes_skyportal_sync.py
Theodlz: I was not saying we need to use one over the other. I was saying why not use both. 

But honestly, given how tricky the syncing with an external data source is and the non zero probability that we already have data we'll get from Hermes but under a different instrument name, not a bad idea to not deduplicate on instrument id, so if an instrument already has the exact same datapoint we don't include it twice! 

You can keep your code as is

tools/populate_tns_ids/populate_tns_ids.py
Theodlz: thx. I might argue that you could fetch data from TNS and keep instrument TNS ids in the DB in sync everytime the microservice boots up, but maybe that's an optimization for another PR!

config.yaml.defaults
Theodlz: I mean the idea here is that data gets moved around between systems and so it will change (different systems will round or read the data on their CPU differently), so that level of precision is a no go anyways. Rounding to some reasonable level (btw camille is not rounding the data that goes in the db, just rounding when assessing duplication! right @camilledzt?) is the right move.

I was just wondering if some reasonable number of digits for flux (that folks would agree on is good enough) already existed out there @mcoughlin. If not we can just use whatever.

mcoughlin (approved):

Issue #5867, opened 2025-10-28
Error sending data with multiple instruments using SharingService


camilledzt:
**Describe the bug**
I have a source with photometry points with multiple instruments. An error is not catch :  ``Error processing sharing request: IN expression list, SELECT construct, or bound parameter object expected, got None.``. It is happening when a Sharing Service is not allowed to send photometry for all instruments.

**Screenshots**
Notification : ``Error processing sharing request: IN expression list, SELECT construct, or bound parameter object expected, got None.``

```
[17:38:01 flow] Pushing action baselayer/SHOW_NOTIFICATION to user 1
Traceback (most recent call last):
  File "/home/camilledouzet/workspace/dskyportal/services/sharing_service_submission_queue/sharing_service_submission_queue.py", line 327, in process_submission_requests
    process_submission_request(submission_request, session)
  File "/home/camilledouzet/workspace/dskyportal/services/sharing_service_submission_queue/sharing_service_submission_queue.py", line 219, in process_submission_request
    source, photometry, stream_ids = get_publishable_source_and_photometry(
  File "/home/camilledouzet/workspace/dskyportal/skyportal/utils/data_access.py", line 527, in get_publishable_source_and_photometry
    photometry = get_photometry_by_instruments_stream_and_options(
  File "/home/camilledouzet/workspace/dskyportal/skyportal/utils/data_access.py", line 384, in get_photometry_by_instruments_stream_and_options
    session.scalars(sa.select(Stream.name).where(Stream.id.in_(stream_ids)))
  File "/home/camilledouzet/workspace/dskyportal/venv/lib/python3.10/site-packages/sqlalchemy/sql/operators.py", line 787, in in_
    return self.operate(in_op, other)
  File "/home/camilledouzet/workspace/dskyportal/venv/lib/python3.10/site-packages/sqlalchemy/orm/attributes.py", line 449, in operate
    return op(self.comparator, *other, **kwargs)  # type: ignore[return-value]  # noqa: E501
  File "/home/camilledouzet/workspace/dskyportal/venv/lib/python3.10/site-packages/sqlalchemy/sql/operators.py", line 1963, in in_op
    return a.in_(b)
  File "/home/camilledouzet/workspace/dskyportal/venv/lib/python3.10/site-packages/sqlalchemy/sql/operators.py", line 787, in in_
    return self.operate(in_op, other)
[...]

PR #5890, opened 2025-11-02
Fix error sending data with sharing service

antoine-le-calloch:
Some fix (https://github.com/skyportal/skyportal/issues/5867) and improvement of the sharing service feature.
- handle empty stream_ids when sending data with no photometry point to avoid `Error processing sharing request: IN expression list, SELECT construct, or bound parameter object expected, got None`.
- Fix the Sharing Service form (both creation and edit) to prevent form data from being lost when toggling TNS service options.
- Improve the code and the style and remove useless code.

### DIFF
--- a/skyportal/tests/tools/test_offset_util.py
+++ b/skyportal/tests/tools/test_offset_util.py
@@ -264,9 +264,13 @@ def test_get_desi_finding_chart():
 
 # test for failure on a too-small image size
 def test_get_finding_chart():
-    rez = get_finding_chart(123.0, 33.3, "testSource", imsize=1.0, image_source="dss")
+    rez = get_finding_chart(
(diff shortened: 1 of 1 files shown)


skyportal/tests/tools/test_offset_util.py
```diff
@@ -264,9 +264,13 @@ def test_get_desi_finding_chart():
 
 # test for failure on a too-small image size
 def test_get_finding_chart():
-    rez = get_finding_chart(123.0, 33.3, "testSource", imsize=1.0, image_source="dss")
+    rez = get_finding_chart(
+        123.0, 33.3, "testSource", imsize=1.0, image_source="dss", use_cache=False
```
Theodlz, line 268: nice catch! Thx

PR #5894, opened 2025-11-14
Fix add annotation to invalid source

antoine-le-calloch:
Fix a bad error message saying `Annotation already exists` when the targeted source is not found.
Improve code readability.

### DIFF
--- a/skyportal/handlers/api/annotation.py
+++ b/skyportal/handlers/api/annotation.py
@@ -384,9 +384,7 @@ class AnnotationHandler(BaseHandler):
                 session.commit()
             except IntegrityError as e:
                 if 'is not present in table "objs"' in str(e).lower():
-                    return self.error(
-                        f"Could not access object {resource_id}.", status=403
-                    )
+                    return self.error(f"Obj {resource_id} not found", status=404)
                 return self.error(f"Annotation already exists: {str(e)}")
 
             if isinstance(
--- a/skyportal/handlers/api/source_groups.py
+++ b/skyportal/handlers/api/source_groups.py
@@ -66,7 +66,7 @@ class SourceGroupsHandler(BaseHandler):
                 Obj.select(session.user_or_token).where(Obj.id == obj_id)
             )
             if not obj:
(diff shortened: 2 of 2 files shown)


Theodlz (commented):
@antoine-le-calloch we post a lot of annotations during the night. In an effort to avoid queries that aren't needed, I wonder if we could simply better handle whatever error we get back from the DB upon insert (or wherever we currently get an error) to understand if it's because the obj doesn't exist or something else. That would avoid us running yet another query for something which is anyways enforced by the DB?

antoine-le-calloch:
@Theodlz, Done ✅

skyportal/handlers/api/annotation.py
```diff
@@ -385,6 +383,10 @@ def post(self, associated_resource_type, resource_id):
             try:
                 session.commit()
             except IntegrityError as e:
+                if 'is not present in table "objs"' in str(e).lower():
+                    return self.error(
+                        f"Could not access object {resource_id}.", status=403
```
Theodlz: is a 403 the standard for that kind of errors @antoine-le-calloch ?

skyportal/handlers/api/source_groups.py
```diff
@@ -62,11 +62,11 @@ def post(self):
...
         with self.Session() as session:
-            obj = session.scalars(
+            obj = session.scalar(
                 Obj.select(session.user_or_token).where(Obj.id == obj_id)
-            ).first()
-            if obj is None:
-                return self.error("Invalid objId")
+            )
+            if not obj:
+                return self.error(f"Could not access source {obj_id}.", status=403)
```
Theodlz: same question here.

skyportal/handlers/api/annotation.py
antoine-le-calloch: My bad — it actually makes sense when checking the existence of a source, since a 403 means we don’t have access to it. But here we’re checking the existence of an object that we always have access to, so it should be a 404 Not Found instead. I had implemented everything as if we were checking a source, not a regular object, so the logic was incorrect. I also need to update the error message accordingly.

PR #5900, opened 2025-11-29
Improve group pages

antoine-le-calloch:
- Rewrote Group.tsx with tab navigation (Members / Sources / Streams & Filters) instead of accordions.
- Refactored layout to MUI Box, removed custom makeStyles where possible, improved responsiveness and overall styling.
- Redesigned GroupList.tsx, and used it directly, deleted GroupManagement.tsx and NonMemberGroupList.tsx.
- Cleaned up Groups.tsx, GroupUsers.tsx, GroupSources.tsx, GroupFiltersStreams.tsx, GroupPageManageUserButtons.tsx, NewGroupForm.tsx, Filter.tsx
- AddUsersFromGroupForm.tsx → AddGroupOfUsersForm.tsx, plus a new AddUserFor
- Fix errors and clean up
- Fix the divergences between the two add User select.


mcoughlin (approved):
I agree that this makes things simpler. I assume you don't need the baselayer update?

antoine-le-calloch:
> I agree that this makes things simpler. I assume you don't need the baselayer update?

@mcoughlin You want me to remove the baselayer bump ?

mcoughlin:
@antoine-le-calloch Ah no that's fine, sorry.

PR #5922, opened 2025-12-10
Update the delay when triggering TAROT

antoine-le-calloch:
We have implemented a new feature on the TAROT server that sends scenes directly to each station as soon as they come from SkyPortal. The 10-minute workaround can now be removed. However, a delay is still required to account for the network time needed to push the scene into the station's database.

### DIFF
--- a/skyportal/facility_apis/tarot.py
+++ b/skyportal/facility_apis/tarot.py
@@ -26,14 +26,17 @@ station_dict = {
     "Tarot_Calern": {
         "filters": ["NoFilter", "g", "r", "i"],
         "status_url": 1,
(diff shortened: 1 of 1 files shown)


antoine-le-calloch:
> I understand that different telescopes take more or less time and that therefore we have a different delay for each, but does that delay need to be a customizable parameter in the schema also?

@Theodlz I don’t think so. Every SkyPortal instance and every request should use the same value, since I’ve already calculated it by measuring the latency. Nobody will know what value to enter anyway, so I don’t really see the point of adding a new field to the form. This value shouldn’t change unless there’s an internal network update on the telescope.

PR #5923, opened 2025-12-10
Upon starting container, user will not get added to group if already a member.

howardisaacson:
I run Skyportal as a submodule in our instance (called MNTLEAF). 

I am running the most recent Skyportal and baselayer commits. 

When starting the container, I am continually having a problem with the provisional admin user already being in the Statewide group. This SQL duplication error prevents the app from starting. 

This small (3line) change will only add the user if not already in the group.  

If this problem should be handled in a different way (maybe my provisional admin should never be in a group), please me know.


Theodlz (approved):
This is great, I know 2 other folks ran into that error recently.

Thanks!

Issue #5637, opened 2025-05-27
Photometric Series not uploading group_ids, causing permissions issues.


howardisaacson:
**Describe the bug**
When uploading a photometric series, via the API, the group_ids field is not being populated. This is causing permissions issues for users trying to download the data via API.

**To Reproduce**

1. With a mntleaf account go to ['LINK'](https://mntleaf.astro.berkeley.edu/api/photometric_series?objectID=PB250162)
2. Search for 'group_ids', you will see it has the value 'None', despite it being populated in the payload. 

          payload = {
                "obj_id": objectID,
                "instrument_id": str(instrument_id),
                "group_ids": [1,2,3],  # Statewide group is group 2. 'all' does not work. 
                "filter": inst_filter,
                "origin": origin,
                "limiting_mag": 20,
                "series_name": series_name,
                "series_obj_id": objectID,
                "exp_time": 60,
                "ra": coord_ra,
                "dec": coord_dec,
                "data": {
                    "mjd": chunk['mjd'].to_list(),
                    "mag": chunk['mag'].to_list(),
                    "magerr": chunk['magerr'].to_list()
                }
            }
**Expected behavior**
The payload['group_ids'] value should be populated. 

**Platform information:**

- SkyPortal version 1.4.0: (On MNTLEAF)
- Interface
  - [ ] I am using the API on a laptop, and looking at the API output on a browser. 

Other:
The general problem is that I can download the photometric_series via API, but another use, who seems to have the same permissions cannot down it.  That user can see the photometric_series data when typing the API endpoint in the browser, so this seems strange.

Example of the API output.
{
  "status": "success",
  "data": {
    "series": [
      {
        "origin": "PRIME",
        "mag_last_detected": 14.468053387505481,
        "limiting_mag": 20.0,
        "owner_id": 2,
        "maxima": {
          "mjd": 60818.19299502019,
          "mag": 60818.19299502019,
          "magerr": 60818.19299502019
        },
        "created_at": "2025-05-27T22:52:07.662282",
        "filename": "/skyportal/persistentdata/phot_series/series-1/series_PB250162_inst_112_PRIME.h5",
        "is_detected": true,
        "ref_flux": null,
        "mean_mag": 15.665885019009773,
        "minima": {
          "mjd": 60385.04613611009,
          "mag": 60385.04613611009,
          "magerr": 60385.04613611009
        },
[some content was removed for brevity]
        "id": 134067,
        "group_ids": null,     <---------------------------------
[...]

PR #5924, opened 2025-12-11
Patch phot series permissions

howardisaacson:
We use photometric_series in our Skyportal instance (MNTLEAF).  The issue with permissions is described in this issue: #5637 . 

This pull request should address those issues. I was unable to run the tests. I was unable to build the llvm module, despite some effort.

### DIFF
--- a/skyportal/models/photometric_series.py
+++ b/skyportal/models/photometric_series.py
@@ -23,7 +23,12 @@ from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import reconstructor, relationship
 
 from baselayer.app.env import load_env
-from baselayer.app.models import Base, CustomUserAccessControl, accessible_by_owner, public
+from baselayer.app.models import (
+    Base,
+    CustomUserAccessControl,
+    accessible_by_owner,
+    public,
+)
 
 from ..enum_types import allowed_bandpasses, time_stamp_alignment_types
 from ..utils.hdf5_files import dump_dataframe_to_bytestream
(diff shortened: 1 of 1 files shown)


skyportal/handlers/api/photometric_series.py
```diff
@@ -538,6 +538,15 @@ def post_photometric_series(json_data, data, attributes_metadata, user, session)
...
         session.commit()
 
+        # Assign groups and streams to the relationships
+        ps.groups = session.scalars(
+            sa.select(Group).where(Group.id.in_(group_ids))
+        ).all()
+        ps.streams = session.scalars(
+            sa.select(Stream).where(Stream.id.in_(stream_ids))
+        ).all()
+        session.commit()
```
Theodlz: I see 2 commit operations happening, which ideally should be avoided so we don't save an invalidate state in the DB.Best would be to add `ps` to the session, add the groups and streams for it, and then commit

skyportal/model_util.py
```diff
@@ -96,7 +96,9 @@ def add_user(username, roles=[], auth=False, first_name=None, last_name=None):
                 session.add(public_group)
                 session.flush()
 
-        user.groups.append(public_group)
+            # Only add user to group if not already a member
+            if public_group not in user.groups:
+                user.groups.append(public_group)
```
Theodlz, line 101: gentle reminder to merge main into that branch so that this change (that's now on main) doesn't show up as something specific to this PR anymore

howardisaacson: I believe I did this correctly...

skyportal/handlers/api/photometric_series.py
howardisaacson: I believe this is resolved now.

PR #5925, opened 2025-12-11
Spectra plot: refactor to accommodate changes in latest plotly

Theodlz:
In this PR, we refactor the spectra plot to address issues resetting issues introduced by the newer versions of plotly.js. Basically, newer plotly coupled with react-plotly seems to trigger full rerenders (and therefore resets of the zoom level) as soon as a state variable is updated in the same component where <Plot ... /> is called. That's obviously incorrect, but locking ourselves out of newer versions is worse. By using react's `useMemo` and `useCallback` hooks, here we prevent these changes from happening. The one issue left was the `redshiftInput` which needs to update the secondary X axis when it changes, which cascades into resetting the layout/zoom again. To get around that, we don't make the redshiftInput a dependency of the createLayout useMemo but instead directly tap into the plotly component (with a useRef hook) to call the internal method to update an axis value.

Then I tried to help a bit with the syntax/logic of the overall component. I should say that I got a lot of help from copilot on this one! Performance looks identical to what it was, and zoom level is now preserved when changing any of the plot parameters (which I don't think was even the case before), which folks should be very happy about.

We also update plotly.js to latest

### DIFF
--- a/static/js/components/plot/SpectraPlot.jsx
+++ b/static/js/components/plot/SpectraPlot.jsx
@@ -109,11 +109,13 @@ const SpectraPlot = ({ spectra, redshift, mode, plotStyle }) => {
 
   // Memoize user custom lines to avoid recreating on every render
   const userCustomLines = useMemo(() => {
-    return Object.keys(preferences?.spectroscopyButtons || {}).map((key) => ({
-      name: key,
-      color: preferences?.spectroscopyButtons[key].color,
-      x: preferences?.spectroscopyButtons[key].wavelengths,
-    }));
+    return Object.entries(preferences.spectroscopyButtons || {}).map(
+      ([name, { color, wavelengths }]) => ({
+        name,
+        color,
+        x: wavelengths,
+      }),
+    );
   }, [preferences?.spectroscopyButtons]);
 
   // Memoize the combined lines array to avoid recreating on every render
@@ -232,7 +234,7 @@ const SpectraPlot = ({ spectra, redshift, mode, plotStyle }) => {
           text,
         };
       })
-      .filter((spectrum) => spectrum !== null);
+      .filter(Boolean);
 
     // Finalize stats ranges
     spectrumTypes.forEach((type) => {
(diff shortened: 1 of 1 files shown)


static/js/components/plot/SpectraPlot.jsx
```diff
@@ -100,157 +107,134 @@ const SpectraPlot = ({ spectra, redshift, mode, plotStyle }) => {
...
-  }));
+  // Memoize user custom lines to avoid recreating on every render
+  const userCustomLines = useMemo(() => {
+    return Object.keys(preferences?.spectroscopyButtons || {}).map((key) => ({
+      name: key,
+      color: preferences?.spectroscopyButtons[key].color,
+      x: preferences?.spectroscopyButtons[key].wavelengths,
+    }));
+  }, [preferences?.spectroscopyButtons]);
+
```
antoine-le-calloch, line 120: Value can be destructured for clarity:
```
    .map([name, { color, wavelengths }]) => ({
      name,
      color,
      x: wavelengths,
    })
  );
```

static/js/components/plot/SpectraPlot.jsx
```diff
@@ -100,157 +107,134 @@ const SpectraPlot = ({ spectra, redshift, mode, plotStyle }) => {
...
-    // remove null values (spectra that had no valid fluxes or wavelengths)
-    newSpectra = newSpectra.filter((spectrum) => spectrum !== null);
+        return {
+          ...spectrum,
+          fluxes_normed: fluxes,
+          wavelengths,
+          text,
+        };
+      })
+      .filter((spectrum) => spectrum !== null);
```
antoine-le-calloch: Can also be: `.filter(Boolean);`

static/js/components/plot/SpectraPlot.jsx
```diff
@@ -395,162 +374,378 @@ const SpectraPlot = ({ spectra, redshift, mode, plotStyle }) => {
...
+  );
 
-  const handleChangeTab = (event, newValue) => {
-    setTabIndex(newValue);
-  };
+  // Memoize event handlers to prevent re-renders
+  const handleDoubleClick = useMemo(
+    () => () => setLayoutReset((prev) => prev + 1),
+    [],
+  );
```
antoine-le-calloch: Not sure about this one but useCallback would be better here no ?
```
const handleDoubleClick = useCallback(() => {
  setLayoutReset((prev) => prev + 1);
}, []);
```

Theodlz: good call, just changed it

static/js/components/plot/SpectraPlot.jsx
Theodlz: done! Thanks for the suggestion, definitely much easier to read that way!

Theodlz:
> Hard to fully verify the overall behavior, so lets validate it in production. Just a bit of minor cleanup as usual. 🙂

It should actually be pretty ok to check! All that matters is that when clicking on all the buttons under that plot to move and hide/show lines around, the zoom level of the plot doesn't get reset (that's the main goal of these changes).

PR #5950, opened 2026-01-14
enh: add extinction to the photometry plot

camilledzt:


Theodlz (commented):
That's a great addition Camille, thanks! As discussed on Slack, I moved the extinction on/off button up, in the Photometry plot header to avoid us from using more height in that plot which already often overflows on the screen.

Just one bug I noticed. When showing the plot in mag (the default behavior), after applying extinction correction I can still see non detections. When showing the plot in flux, the non detections are gone.

I guess the first question is, when showing a lightcurve with extinction correction does it still make sense to show non-detections (which i don't think are corrected)? The second question would be: if we do still show them why only on the mag plot?

camilledzt:
Thank you for the review, I removed non detection points when extinction option is on on the plot. The switch to show non detection is disable when extinction option is on and there is a tooltip explaining why.

PR #5951, opened 2026-01-16
enh: add a slider to only show validated points in the photometry plot

camilledzt:
The goal of this PR is to add a slider to only show validated points in the photometry plot.


Theodlz (approved):
LGTM 🚀

PR #5952, opened 2026-01-19
Fix Gemini API url

antoine-le-calloch:
Fix Gemini API URL configuration to use different URLs according to the telescope site (North or South).

### DIFF
--- a/skyportal/facility_apis/gemini.py
+++ b/skyportal/facility_apis/gemini.py
@@ -19,10 +19,12 @@ env, cfg = load_env()
 log = make_log("facility_apis/gemini")
 
 
-def get_api_url(is_south_site=True):
+def get_api_url(program_id):
+    if not any(program_id.startswith(x) for x in ["GN", "GS"]):
+        raise ValueError("Invalid program ID, must start with GN or GS")
     prefix = (
         cfg["app.gemini.south_prefix"]
-        if is_south_site
+        if program_id.startswith("GS")
         else cfg["app.gemini.north_prefix"]
     )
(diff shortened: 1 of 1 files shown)


skyportal/facility_apis/gemini.py
```diff
@@ -303,6 +308,10 @@ def submit(request, session, **kwargs):
         if not altdata:
             raise ValueError("Missing allocation information.")
 
+        if not any(altdata.get("programid", "").startswith(x) for x in ["GN", "GS"]):
+            raise ValueError("Invalid program ID, must start with GN or GS")
```
Theodlz: @antoine-le-calloch I think there is already an `if, else if, else if` statement that validates all this

Theodlz (approved):
Thanks Antoine. I think it could have been a little easier/cleaner to pass the programid to the function you added to generate the URL, and have that handle the north vs south logic. But all in all, this works and I'm fine merging it as is. 

Feel free to merge when you are happy with it too

skyportal/facility_apis/gemini.py
antoine-le-calloch: Yep, there is, but we check other things too here, and since it's not in the same function/block, I was okay with checking it again.

antoine-le-calloch:
> Thanks Antoine. I think it could have been a little easier/cleaner to pass the programid to the function you added to generate the URL, and have that handle the north vs south logic. But all in all, this works and I'm fine merging it as is.
> 
> Feel free to merge when you are happy with it too

Done ✅

PR #5955, opened 2026-01-20
Include arcs before and after observation in SOAR configurations

antoine-le-calloch:
Add pre and post-observation arcs to SOAR configurations (requested by one of Igor's graduate students).


Theodlz (approved):
LGTM

PR #5957, opened 2026-01-21
Enh: add new username creation method

camilledzt:
The goal of this PR is to add a method to automatically generate username from first initial + last name when `username_is_email` is set to `False` in the configuration. This last step depends on this PR : https://github.com/cesium-ml/baselayer/pull/409 

Previously, when USERNAME_IS_FULL_EMAIL=False, the username was taken directly from details["username"]. Now it generates a format using the first letter of the first name + last name (ie., "Camille Douzet" => "cdouzet").

Username generation priority:
  1. USERNAME_IS_FULL_EMAIL=True: Use email as username
  2. USERNAME_IS_FULL_EMAIL=False: Use first initial + lastname, with auto-incrementing number if already exists                                                   
  3. Fallback: Use details["username"] if first_name or last_name is unavailable


Theodlz:
@camilledzt. you might be aware of that already, but we are having difficulties getting the latest baselayer to work with skyportal. So, if this depends on a baselayer change, then it can't be merge until that happens. The PR where we are trying to get baselayer latest on skyportal is #5891. Just one broken test left to fix there...

doc/setup.md
```diff
@@ -291,6 +291,14 @@ SkyPortal, you should enable multi-user login by adding Google
 credentials to the `server:auth` section of the configuration file and
 setting `debug_login` to `False`.
 
+### Username generation
+
+When `server.auth.username_is_email` is set to `True` (the default), the user's email address is used as their username.
```
Theodlz: just one thing: the username is not just the email, right? isn't it some value generated FROM the email? I don't think new accounts have their exact email address as username currently. It's a bit of a nitpick, but good to make it more precise.

skyportal/onboarding.py
```diff
@@ -104,6 +104,18 @@ def create_user(strategy, details, backend, uid, user=None, *args, **kwargs):
...
+def get_unique_username(base_username):
+    """Generate a unique username by appending a number if needed."""
+    username = base_username
+    counter = 1
+    while (
+        DBSession().scalar(sa.select(User).where(User.username == username)) is not None
+    ):
+        username = f"{base_username}{counter}"
+        counter += 1
+    return username
```
Theodlz: Instead of doing multiple queries, why not run a query like "get me all the User.username (no need to get the full user btw) that start with <base_username>":
- if there are no results, all good
- if there are no exact matches, all good too
- otherwise, just enter a loop like the one you wrote above where you add an int and increment until you find a value that matches no existing usernames.

Theodlz: (I'm just saying this because it's always better to do one DB query and work in python, rather than multiple DB queries, especially where there is nothing but a small array of strings to return from the DB. Saves you all the roundtrips).

Theodlz: to be clear: this is nitpicking! You don't have to implement a change if you have more pressing matters to attend to.

skyportal/onboarding.py
```diff
@@ -116,7 +128,13 @@ def get_username(strategy, details, backend, uid, user=None, *args, **kwargs):
...
             username = details["email"]
         else:
-            username = details["username"]
+            first_name = details.get("first_name", "")
+            last_name = details.get("last_name", "")
+            if first_name and last_name:
+                base_username = (first_name[0] + last_name).lower()
+                username = get_unique_username(base_username)
+            else:
+                username = details["username"]
```
Theodlz: is there always a username?

Theodlz (approved):
just a few questions/nitpicks, but honestly this is great already. LGTM

skyportal/onboarding.py
Theodlz: or should we fallback to the email just in case?

doc/setup.md
camilledzt: I've added some details

skyportal/onboarding.py
camilledzt: I changed this

skyportal/onboarding.py
camilledzt: I've added an "or" condition to have a fallback

Theodlz (approved):

PR #5964, opened 2026-01-29
Refactor candidate handler

antoine-le-calloch:
- Replace ValueError raises with error returns in candidate.py to fix "ValueError: Localization XXX not found" and prevent other errors
- Improve the code by avoiding duplication, fixing a missing .all(), and changing all SQLAlchemy .filter() calls to .where()
- Refactor photometry annotations query to eliminate redundancy and avoid recreating the query at each iteration
- Fix list_name filter using a subquery join pattern. The previous implementation added a WHERE clause on Listing without an explicit join, causing an implicit cross join that returned all candidates if any listing matched, or none if no listing matched
- Test the number_of_detections param when require_detections is True.

### DIFF
--- a/skyportal/handlers/api/candidate/candidate.py
+++ b/skyportal/handlers/api/candidate/candidate.py
@@ -1708,14 +1708,14 @@ def grab_query_results(
                 query_id = str(uuid.uuid4())
                 all_ids = session.scalars(ordered_ids).unique().all()
                 cache[query_id] = array_to_bytes(all_ids)
-            totalMatches = len(all_ids)
+            total_matches = len(all_ids)
             obj_ids_in_page = all_ids[
                 ((page - 1) * n_items_per_page) : (page * n_items_per_page)
             ]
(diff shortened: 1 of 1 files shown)


skyportal/handlers/api/candidate/candidate.py
```diff
@@ -1145,27 +1135,16 @@ def get(self, obj_id=None):
...
                             )
 
-                else:
-                    photometry_annotations_query = create_photometry_annotations_query(
-                        session,
-                        photometry_annotations_filter_origin=photometry_annotations_filter_origin,
-                        photometry_annotations_filter_before=photometry_annotations_filter_before,
-                        photometry_annotations_filter_after=photometry_annotations_filter_after,
-                    )
-
```
antoine-le-calloch, line 1190: Here, we overwrite photometry_annotations_query at each iteration, which seems odd. Is this the intended behavior? @Theodlz @mcoughlin

PR #5965, opened 2026-01-31
docs: Expand Apple Silicon setup with troubleshooting guide

AnastasiaKWei:
Enhances the Apple Silicon setup documentation based on troubleshooting experience with an M4 MacBook.

Changes:
- Added detailed environment verification steps for x86_64 conda environments
- Added pip guardrails configuration to prevent ARM64 wheel issues
- Expanded port configuration documentation with test_config.yaml note
- Added troubleshooting section for architecture mismatch errors
- Included diagnostic commands and a step-by-step resolution process

Tested on: M4 MacBook Pro

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -150,12 +150,28 @@ alias python='python3'
 source ~/.zshrc
 ```
 ### Checking for Port Availability
-SkyPortal defaults to using port 5000. However, this port may already be in use by MacPorts or other services. To verify if port 5000 is available, use the `lsof` command in the terminal.
+SkyPortal defaults to using port 5000. However, this port may already be in use on MacOS (e.g., by AirPlay Receiver, MacPorts, or other services). To verify port availability, use the `lsof` command in the terminal:
 
 ```
 lsof -i :5000
 ```
-If the command outputs information about a service, it means that port 5000 is already in use. In this case, you may need to configure SkyPortal to use a different port (in `config.yaml`, and if you intend to run the unit tests, then also in `test_config.yaml`).
+If the command outputs information about a service, it means that port 5000 is already in use. In this case, you may need to configure SkyPortal to use a different port by updating the following lines in `config.yaml`:
+
+```yaml
+server:
+  port: 5001  # Change from 5000 to 5001
+
+ports:
+  app: 5001  # Change from 5000 to 5001
+
+docs:
+  servers:
+    - url: http://localhost:5001  # Change from 5000 to 5001
+```
+
+Before switching, verify that your chosen alternative port is itself available (e.g., `lsof -i :5001`).
+
+If you plan to run `make load_demo_data` or the unit tests, also update the port in `test_config.yaml` to match. Port mismatches between the two configs can cause 403 Forbidden errors during demo data loading.
 
 ## Installation: Debian-based Linux and WSL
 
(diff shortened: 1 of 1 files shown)


doc/setup.md
```diff
@@ -41,7 +39,21 @@ conda config --add channels conda-forge
 conda config --set channel_priority strict
 ```
 
+Verify your environment is correctly configured
+```
+# check which Python and pip you're using
+which python
+which pip
```
stefanv: I am busy porting package management over to uv, so hopefully a thing of the past soon.

doc/setup.md
```diff
@@ -41,7 +39,21 @@ conda config --add channels conda-forge
...
+
+# verify architecture and environment paths
+python -c "import platform,sys; print(platform.machine()); print(sys.executable)"
+python -m pip -V
+```
 
+Expected output:
+- `platform.machine()` should show `x86_64` (not `arm64`)
```
stefanv: But you'll see `arm64` on Mac machines, presumably?

doc/setup.md
```diff
@@ -103,12 +115,25 @@ After installing each package, Homebrew will print out the installation paths. Y
 	conda install pyproj numba Shapely
 	```
 
-5. (ARM M1/M2) Explicitly [install ligo.skymap using conda rather than pip](https://lscsoft.docs.ligo.org/ligo.skymap/quickstart/install.html#option-2-conda):
+6. (ARM M-series) Explicitly install packages with known ARM64 compatibility issues using conda rather than pip:
```
stefanv: Especially now that I'm moving us over to uv, this will be problematic. I could try the pixi.sh route, but I'm a bit surprised the niche packages we use are on conda-forge.

stefanv (commented):
Hi @AnastasiaKWei, thanks for this detailed feedback and sorry for all the trouble you had to go through to get SP running :/

I'm a bit confused by what you are seeing: your pip seems to be doing the right thing (installing arm64 wheels), but then you say you are running x86_64 via Rosetta—how do you tell Python to do this? How did you install your Python, also.

I am trying to figure out if we can do all this, consistently, via uv, which I'm working on in baselayer right now (actually just fixing up CI, and then I'll make a PR to SkyPortal as well).

doc/setup.md
```diff
@@ -30,7 +28,7 @@ source skyportal_env/bin/activate
 
 You can also use `conda` or `pipenv` to create your environment.
 
-If you developing on a Mac with an ARM (M1/M2) you might consider using a Rosetta-driven environment so that you more easily install dependencies (that tend to be x86-centric):
+If you are developing on a Mac with an ARM (M-series) you might consider using a Rosetta-driven environment so that you more easily install dependencies (that tend to be x86-centric). Packages known to have ARM64 compatibility issues (as of Jan 2026): `sncosmo`, `python-ligo-lw`, `igwn-ligolw`, `ligo-segments`, `ligo.skymap`.
```
Theodlz: @AnastasiaKWei could you, even if just on that issue and not in the docs themselves, explain these issues? All the dev team or most, are developing on arm64 M-series macs, and we never had to emulate an x86 environment. So this statement surprises me. In fact, we had to emulate x86 when M-series first came out, but everything has been well-supported and compatible since then. From what I can remember, the real issue was conda, where things were arm64, but it would work just fine with any other environments

doc/setup.md
Theodlz: Also @stefanv I just added that to a comment above, but me, antoine, thomas, ... we're all rocking M-series macs (so arm64) just fine. If it works natively, I just really don't think we should tell folks to bother emulating an old platform?

doc/setup.md
Theodlz: we should just delete section 6. It installs just fine with pip, and has for at least 3 years now

mcoughlin:
@AnastasiaKWei Did you decide what needed to be updated here?

AnastasiaKWei:
Hi @mcoughlin, sorry for the delay! Most of this PR was superseded by the uv migration, so I've rebased and just added some notes to the port-configuration docs (exact config.yaml keys to change, plus the test_config.yaml mismatch that causes 403 errors). These are mostly minor improvements, and it's ready for review.

mcoughlin (approved):
LGTM

PR #5979, opened 2026-02-02
enh: adding group access to source tags

camilledzt:
This PR adds group-based access control to object tags. Tags are now associated with specific groups allowing visibility control over users who can or not see/manage tags on sources.

System admins can see all tags for all groups, add any tag to any groups.
Fulluser + Manage sources can see all tags for the groups they belong to, add/remove tag for their groups.
Fulluser can see and add tags for their groups but only delete tags if they are the author.

GroupObjTag table has been created for many-to-many relationship between groups and tags.

Edit 03/05/25: 
* Existing tags are associated to the default public group in the alembic migration.
* If no group is selected when a user is adding a tag to a source, it will be set to the public group automatically. (a small text explains that on the frontend)


skyportal/handlers/api/internal/recent_sources.py
```diff
@@ -70,11 +70,30 @@ def get(self):
...
+                    **tag.to_dict(),
+                    "name": tag.objtagoption.name,
+                    "groups": [
+                        g.to_dict()
+                        for g in tag.groups
+                        if user_group_ids is None or g.id in user_group_ids
+                    ],
+                }
+                for tag in tags
+            ]
```
Theodlz: shouldn't the group access stuff already be handled by the query above? I'm actually not too sure... Worth checking, in case this additional access stuff isn't needed. Low priority comment, feel free to ignore it if you're too busy

skyportal/handlers/api/internal/recent_sources.py
```diff
@@ -70,11 +70,30 @@ def get(self):
                 self.current_user, session
             )
             tags = session.scalars(
-                ObjTag.select(session.user_or_token).where(
-                    ObjTag.obj_id.in_(list(set(query_results)))
-                )
+                ObjTag.select(session.user_or_token)
+                .where(ObjTag.obj_id.in_(list(set(query_results))))
+                .distinct()
```
Theodlz, line 75: the curse of data access 😉

skyportal/handlers/api/source.py
```diff
@@ -598,8 +598,25 @@ async def get_source(
...
+                **tag.to_dict(),
+                "name": tag.objtagoption.name,
+                "groups": [
+                    g.to_dict()
+                    for g in tag.groups
+                    if user_group_ids is None or g.id in user_group_ids
+                ],
+            }
+            for tag in tags
+        ]
```
Theodlz: this exact same patterns already shows up 3 times. Wondering if it wouldn't just deserve to be a function!!! Worth deduplicating to make it easier to maintain

skyportal/handlers/api/tag.py
```diff
@@ -359,43 +375,117 @@ def post(self):
...
         if not objtagoption_id or not obj_id:
             return self.error("Both `objtagoption_id` and `obj_id` must be provided")
 
         with self.Session() as session:
+            if not group_ids or not isinstance(group_ids, list) or len(group_ids) == 0:
+                public_group_id = session.scalar(
+                    sa.select(Group.id).where(
+                        Group.name == cfg["misc.public_group_name"]
+                    )
+                )
```
Theodlz: I think that here validation of a correct input is mixed with whether or not any input was provided. I would suggest to do in this order:
- don't default to `[]` for the group_ids parameter, so know for sure if its None that is was not provided.
- if its not None, check that it's a valid list and that its items can be cast to it. If not, throw an error.
- if its None you replace it with the public_group_id, else you verify that the user has access to all the groups they provided (except if they are an admin)

Theodlz: you already do some of that obvious, but I just propose to decouple input correctness validation vs no input is provided, for clarity

skyportal/handlers/api/tag.py
```diff
@@ -359,43 +375,117 @@ def post(self):
...
-                Obj.select(session.user_or_token).where(Obj.id == obj_id)
-            ).first():
-                return self.error("Specified obj does not exist", status=404)
+                groups_to_add = valid_group_ids - existing_group_ids
+
+                if not groups_to_add:
+                    return self.error(
+                        "This tag is already associated with all the selected groups",
+                        status=400,
+                    )
```
Theodlz: I'm wondering if the right option is to error out. If the user is trying to share a tag to groups that already have it, should we just return a 200 and do nothing? I feel like it's not a failure, it's just a useless action but there is nothing "wrong" with that?

skyportal/handlers/api/tag.py
```diff
@@ -429,15 +536,87 @@ def delete(self, association_id):
...
+
+                for group_id in groups_to_remove:
+                    group_obj_tag = session.scalars(
+                        sa.select(GroupObjTag).where(
+                            GroupObjTag.obj_tag_id == association_id,
+                            GroupObjTag.group_id == group_id,
+                        )
+                    ).first()
+                    if group_obj_tag:
+                        session.delete(group_obj_tag)
```
Theodlz: to avoid `N` queries, perhaps just run one with `group_id.in(your_list)` and then call delete on the results? That way you always do at most 1 query.

skyportal/handlers/api/tag.py
```diff
@@ -429,15 +536,87 @@ def delete(self, association_id):
...
+                ).first()
+                if group_obj_tag:
+                    session.delete(group_obj_tag)
+
+            remaining_count = len(current_tag_group_ids) - len(
+                groups_to_remove & current_tag_group_ids
+            )
+
+            if remaining_count == 0:
+                session.delete(obj_tag)
```
Theodlz, line 611: this logic is duplicated for the admin vs non admin. Maybe this can be simplified a lot more to avoid that?

skyportal/handlers/api/tag.py
```diff
@@ -429,15 +536,87 @@ def delete(self, association_id):
...
+                    status=403,
+                )
+
+            for group_id in groups_to_remove:
+                group_obj_tag = session.scalars(
+                    sa.select(GroupObjTag).where(
+                        GroupObjTag.obj_tag_id == association_id,
+                        GroupObjTag.group_id == group_id,
+                    )
+                ).first()
```
Theodlz: for the non admin case, I suggest not using sa.select but GroupObjTag.select(user, mode="delete"). The latter is the standard, that's how you enforce data access that may be written at the model level

skyportal/models/tag.py
```diff
@@ -46,6 +59,31 @@ class ObjTagOption(Base):
 
 ObjTag.author = relationship(User, doc="The associated User")
 
+# Define the join table for Group-ObjTag relationship
+GroupObjTag = join_model("group_obj_tags", Group, ObjTag)
+GroupObjTag.__doc__ = "Join table mapping Groups to ObjTags."
```
Theodlz, line 62: did you check what the auto generated __doc__ says? I think baselayer does that for us. Anyways just nitpick!

alembic/versions/bea830983575_add_group_access_to_source_tags.py
```diff
@@ -0,0 +1,103 @@
...
+    # from the app config; if it is not set or the group does not exist,
+    # the tags will simply be invisible
+    bind = op.get_bind()
+
+    try:
+        from baselayer.app.env import load_env
+
+        _, cfg = load_env()
+        public_group_name = cfg.get("misc.public_group_name")
+    except Exception:
```
Theodlz, line 63: May I suggest first checking if there are any tags? if there are none, they no need to bother running this and alerting the user that there is no public group?

alembic/versions/bea830983575_add_group_access_to_source_tags.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+        return
+
+    bind.execute(
+        sa.text(
+            "INSERT INTO group_obj_tags (group_id, obj_tag_id, created_at, modified) "
+            "SELECT :group_id, id, NOW(), NOW() FROM obj_tags"
+        ),
+        {"group_id": public_group_id},
+    )
```
Theodlz, line 97: nice!!! It's a very elegant way to do it

Theodlz (commented):
Hi @camilledzt, thanks for implementing all this, and sorry for the wait on this review!!!

I left a bunch of comments, mostly minor stuff, you'll see.

skyportal/handlers/api/internal/recent_sources.py
camilledzt: Yes, the query already handles access control. The group filtering is here to avoid giving access to the complete list of the groups that are associated to this tag in the API response. `groups` is also used in the frontend.

skyportal/handlers/api/source.py
camilledzt: done :)

skyportal/handlers/api/tag.py
camilledzt: done

skyportal/handlers/api/tag.py
camilledzt: yes I agree, I've changed it

skyportal/handlers/api/tag.py
camilledzt: done

skyportal/handlers/api/tag.py
camilledzt: Yes I also simplified that

skyportal/models/tag.py
camilledzt: Hum, I don't know what you expect here ? I also see the group_joins.py file. But when I try to implement it I have a problem with sqlalchemy.

skyportal/handlers/api/tag.py
```diff
@@ -552,63 +559,42 @@ def delete(self, association_id):
...
 
-            if not groups_to_remove:
-                return self.error(
-                    "You don't have any group associations with this tag to remove",
-                    status=403,
+            stmt = (
+                sa.select(GroupObjTag)
+                if is_system_admin
+                else GroupObjTag.select(session.user_or_token, mode="delete")
+            )
```
Theodlz: this `sa.select(...) if is_system_admin else ...` pattern is odd. I mean, isn't that precisely what the data access rules defined in the model are supposed to do for you? Typically, we wouldn't see this in the handlers since the data access has that logic built in.

Theodlz (commented):
just one comment!

static/js/components/EditTagGroups.jsx
```diff
@@ -0,0 +1,265 @@
...
+        {availableGroups.length > 0 ? (
+          <div className={classes.groupList}>
+            {availableGroups.map((group) => (
+              <FormControlLabel
+                key={group.id}
+                control={
+                  <Checkbox
+                    checked={selectedGroupIds.includes(group.id)}
+                    onChange={() => handleGroupToggle(group.id)}
+                    data-testid={`tag-group-checkbox-${group.id}`}
```
antoine-le-calloch: I don't see any test who need this `data-testid`.

static/js/components/EditTagGroups.jsx
```diff
@@ -0,0 +1,265 @@
...
+          >
+            Cancel
+          </Button>
+          <Button
+            onClick={handleSubmit}
+            primary
+            disabled={
+              isSubmitting || isDeleting || selectedGroupIds.length === 0
+            }
+            data-testid="save-tag-groups-button"
```
antoine-le-calloch: Same.

static/js/components/ObjectTags.jsx
```diff
@@ -171,30 +194,25 @@ const ObjectTags = ({ source }) => {
...
   const availableTags = tagOptions.filter(
     (tag) => !usedTagIds.includes(tag.id),
   );
 
+  const sourceGroupIds = source.groups?.map((g) => g.id) || [];
+  const availableGroups =
+    sourceGroupIds.length > 0
+      ? groups?.filter((g) => sourceGroupIds.includes(g.id)) || []
+      : groups || [];
+
```
antoine-le-calloch: You can use only one `|| []`

static/js/components/ObjectTags.jsx
```diff
@@ -236,7 +255,7 @@ const ObjectTags = ({ source }) => {
         data-testid="add-tag-dialog"
```
antoine-le-calloch, line 256: Same.

skyportal/handlers/api/tag.py
```diff
@@ -359,43 +375,124 @@ def post(self):
...
             return self.error("Both `objtagoption_id` and `obj_id` must be provided")
 
+        if group_ids is not None and len(group_ids) > 0:
+            if not isinstance(group_ids, list):
+                return self.error("`group_ids` must be a list of integers")
+            try:
+                group_ids = [int(gid) for gid in group_ids]
+            except (ValueError, TypeError):
+                return self.error("`group_ids` must be a list of integers")
+        else:
```
antoine-le-calloch: The `must be a list of integers` error message will never be reach since we already test the `len(group_ids)`, so this will raise the error.

static/js/components/ObjectTags.jsx
```diff
@@ -92,7 +100,10 @@ const ObjectTags = ({ source }) => {
 
   useEffect(() => {
     dispatch(objectTagsActions.fetchTagOptions());
-  }, [dispatch]);
+    if (!groups || groups.length === 0) {
+      dispatch(groupsActions.fetchGroups());
+    }
+  }, [dispatch, groups]);
```
antoine-le-calloch: The `fetchGroups` will Infinite loop if groups is empty. You don't need to run the use-effect on `groups` change: `}, [dispatch]);`

static/js/components/widget/DynamicTagDisplay.jsx
```diff
@@ -1,18 +1,47 @@
...
   const tagOptions = useSelector((state) => state.objectTags || []);
+  const userGroups = useSelector((state) => state.groups.userAccessible);
 
-  const measureTextWidth = (text) => {
+  useEffect(() => {
+    if (!userGroups || userGroups.length === 0) {
+      dispatch(groupsActions.fetchGroups());
+    }
+  }, [dispatch, userGroups]);
+
```
antoine-le-calloch, line 32: Same infinite loop here.

skyportal/handlers/api/tag.py
```diff
@@ -429,15 +543,66 @@ def delete(self, association_id):
...
-            session.delete(assoc)
+
+            obj_key = obj_tag.obj.internal_key
+            is_system_admin = self.current_user.is_system_admin
+            current_tag_group_ids = {g.id for g in obj_tag.groups}
+
+            if is_system_admin:
+                groups_to_remove = (
+                    set(requested_group_ids)
+                    if requested_group_ids
```
antoine-le-calloch, line 575: Can lead to a full removal if requested_group_ids is an empty array. Not very serious but good to know.

skyportal/handlers/api/tag.py
```diff
@@ -429,15 +543,66 @@ def delete(self, association_id):
...
+                if requested_group_ids:
+                    groups_to_remove = set(requested_group_ids) & user_group_ids
+                    if not groups_to_remove:
+                        return self.error(
+                            "You don't have access to any of the specified groups",
+                            status=403,
+                        )
+                else:
+                    groups_to_remove = current_tag_group_ids & user_group_ids
+
```
antoine-le-calloch, line 589: Same here requested_group_ids=[] remove all.

static/js/components/EditTagGroups.jsx
```diff
@@ -0,0 +1,265 @@
...
+        const deleteResult = await dispatch(
+          objectTagsActions.deleteObjectTag({
+            id: tag.id,
+            group_ids: groupsToRemove,
+          }),
+        );
+        if (deleteResult.status !== "success") {
+          throw new Error(deleteResult.message || "Failed to remove groups");
+        }
+      }
```
antoine-le-calloch: If someone remove all the current groups and add other groups. Then the deleteObjectTag will delete the tag and the add will create a new tag with the user as creator.

static/js/components/EditTagGroups.jsx
```diff
@@ -0,0 +1,265 @@
...
+        </Typography>
+
+        {availableGroups.length > 0 ? (
+          <div className={classes.groupList}>
+            {availableGroups.map((group) => (
+              <FormControlLabel
+                key={group.id}
+                control={
+                  <Checkbox
+                    checked={selectedGroupIds.includes(group.id)}
```
antoine-le-calloch, line 177: The groups selected for this tag, but which the user is not part of, will not be selected. As a result, the code will try to remove them when submitting the form. Instead, we should keep them selected but hide or disable them maybe?

static/js/components/widget/DynamicTagDisplay.jsx
```diff
@@ -1,18 +1,47 @@
...
+  // Filter tags to only show those the user has access to (via group membership)
+  const userGroupIds = useMemo(
+    () => new Set(userGroups?.map((g) => g.id) || []),
+    [userGroups],
+  );
+
+  const accessibleTags = useMemo(() => {
+    if (!source.tags) return [];
+    return source.tags.filter((tag) => {
+      if (!tag.groups || tag.groups.length === 0) return false;
```
antoine-le-calloch: `tag.groups.length` should never be 0 but if it happen we want to see it so we can remove it from the DB.

antoine-le-calloch (commented):
Some comments.

static/js/components/EditTagGroups.jsx
Theodlz: agreed, I don't think the frontend should take care of that kind of logic (backend should do the delete when no groups are left), and the frontend should be able to submit groups to add/delete in one call to the API

static/js/components/widget/DynamicTagDisplay.jsx
camilledzt: I change the return true so that we can still see the orphan tags, but if a public group is set on the skyportal instance, then it is handled by the alembic migration that associate all tags to the public group by default

static/js/components/EditTagGroups.jsx
camilledzt: I change the order to not create a new tag association

skyportal/handlers/api/tag.py
camilledzt: I add a guard in the code before to help with this issue

skyportal/handlers/api/tag.py
camilledzt: I add a guard in the code before to help with this issue

static/js/components/ObjectTags.jsx
camilledzt: this one is used

skyportal/handlers/api/tag.py
camilledzt: I've changed my if condition

static/js/components/EditTagGroups.jsx
camilledzt: Well it's not true because groups to delete are based on the groups of the user (lign 95). Also, the user can only add a tag to a group that he's in.

skyportal/handlers/api/tag.py
```diff
@@ -546,6 +547,9 @@ def delete(self, association_id):
         data = self.get_json() or {}
         requested_group_ids = data.get("group_ids")
 
+        if requested_group_ids is not None and len(requested_group_ids) == 0:
```
Theodlz: missing a type check, we need to verify its even a list before calling `len()`

static/js/components/widget/DynamicTagDisplay.jsx
```diff
@@ -22,10 +22,8 @@ const DynamicTagDisplay = ({ source, styles }) => {
   const userGroups = useSelector((state) => state.groups.userAccessible);
 
   useEffect(() => {
-    if (!userGroups || userGroups.length === 0) {
-      dispatch(groupsActions.fetchGroups());
-    }
-  }, [dispatch, userGroups]);
+    dispatch(groupsActions.fetchGroups());
+  }, [dispatch]);
```
Theodlz, line 31: groups are in principle hydrated when you open up a tab. I don't think we need to fetch them again everytime we want to display tags 😅. This should have some if statement to only run a `fetch` if we don't have groups loaded in yet.

static/js/components/ObjectTags.jsx
```diff
@@ -100,10 +100,8 @@ const ObjectTags = ({ source }) => {
 
   useEffect(() => {
     dispatch(objectTagsActions.fetchTagOptions());
-    if (!groups || groups.length === 0) {
-      dispatch(groupsActions.fetchGroups());
-    }
-  }, [dispatch, groups]);
+    dispatch(groupsActions.fetchGroups());
+  }, [dispatch]);
```
Theodlz, line 108: same comment here, no need to fetch data if its already there? 

Now obviously the question is: when new groups are added (or deleted/edited), and same for new tag options, do we have a websocket action that tells the users' frontend to refresh. If not, then yeah you have to fetch everytime. So that's what needs to be checked. The right pattern is the handler (on add/update/delete) tells all concerned users to re-fetch these resources, and that way no need to keep fetching them unnecessarily in the code. I don't think you need to worry about groups here (they have been here for a while already), but good to check what is done with tag options. If they are hydrated when the app is loaded and websockets force the client to stay up to date, then please remove this forced fetch.

Theodlz (commented):
Some minor comments, especially about how useEffects are handled to fetch data that might already be there. We should always be careful about that, to not send API calls unnecessarily!

skyportal/handlers/api/tag.py
camilledzt: I've added this check

static/js/components/widget/DynamicTagDisplay.jsx
camilledzt: Ok, I removed it, sorry!

static/js/components/ObjectTags.jsx
camilledzt: there is no websocket for tag options so I removed it for groups and keep it for tag otpions

static/js/components/widget/DynamicTagDisplay.jsx
Theodlz: you didn't have to remote it you know, it's not a bad idea to fetch if its not there. As my comment suggested, putting that behind an if (to only fetch if its null/undefined/length=0) is fine too!

static/js/components/ObjectTags.jsx
Theodlz: same comment as above. Also I would encourage you to just add that backend websocket stuff, it would be good so that when a user adds a new tag option the list is auto updated (without you explicitely making an API call) for all the users! That way you can put this call behind an if for efficiency, and you have the nice added effect to keep tag options in sync for all users?

camilledzt: Ok, I've added the websocket support and guard the fetch calls

static/js/components/widget/DynamicTagDisplay.jsx
camilledzt: Ok I understand, I've put back the if condition

static/js/components/widget/DynamicTagDisplay.jsx
```diff
@@ -1,18 +1,46 @@
...
-
+  const dispatch = useDispatch();
   const tagOptions = useSelector((state) => state.objectTags || []);
+  const userGroups = useSelector((state) => state.groups.userAccessible);
+
+  useEffect(() => {
+    if (!userGroups || userGroups.length === 0) {
+      dispatch(groupsActions.fetchGroups());
+    }
+  }, [dispatch, userGroups]);
```
Theodlz: maybe also a if guarded fetch of the tag options, since you grab them from the `useSelector` and new them to render tags properly? That way if for whatever reason they haven't been hydrated yet, you can fetch them (just like we do for the groups). Gotta make sure stuff is consistent! Here it somewhat arbitrarily queries one thing it needs and not the other.

Also, even if the linter says otherwise, please remove `userGroups` from the `[dispatch, userGroups]`. Since you fetch when `userGroups === 0`, if there are no groups to fetch then you enter an infinite loop. You should just try that fallback fetch (that only happens if data is not hydrated or there was no data at hydration time) once when the component renders, not every time that data from redux changes.

static/js/components/ObjectTags.jsx
```diff
@@ -82,17 +86,26 @@ const ObjectTags = ({ source }) => {
...
   useEffect(() => {
-    dispatch(objectTagsActions.fetchTagOptions());
-  }, [dispatch]);
+    if (!tagOptions || tagOptions.length === 0) {
+      dispatch(objectTagsActions.fetchTagOptions());
+    }
+    if (!groups || groups.length === 0) {
+      dispatch(groupsActions.fetchGroups());
+    }
+  }, [dispatch, tagOptions, groups]);
```
Theodlz: same comment as the one I left on DynamicTagDisplay, about not including `tagOptions` and `groups` in `[dispatch, tagOptions, groups]`, to avoid infinite fetch loops.

Theodlz: in theory you know, you can even just have `[]`

Theodlz (changes requested):
Some comments about the useEffects

static/js/components/ObjectTags.jsx
camilledzt: Ok I removed it

static/js/components/widget/DynamicTagDisplay.jsx
camilledzt: Ok, I've added the guarded fetch to be consistent

Issue #5980, opened 2026-02-05
Memory usage issue when scanning


Theodlz:
~Right now, as a user scans on the candidates page, we load the lightcurves for each object. Issue is, if you are scanning for hundreds (or worse) of candidates, all that photometry accumulates in the redux state.~

~Instead, I think that when we move from one page to another, once the candidates from the previous page show up we should empty the photometry cache.~

~Alternatively (and that could be combined with this), we could have a dedicated fetchphotometry function for candidate scanning, where we fetch the lightcurves without all the fields we don't care about (we can sanitize after loading, no need to necessarily make changes to the backend).~

~Voiding the cache on page change would be the easiest change here.~

Clearly something else is eating the RAM, since dumping that state to disk for a page that uses 5.3GB of RAM, only shows 100M. Will keep looking...

PR #5981, opened 2026-02-05
Various candidate scanning (+vega plots) optimization

Theodlz:
solves #5980 

For context: even though we use a `ViewportList` (a virtualized list, that only renders/loads in the DOM what needs to be on screen while scrolling) to not load ALL candidates but only those in view, we had some memory leaks:
- the most obvious one, is that once a candidate renders we fetch photometry for it and that stays in the redux forever, meaning if you go through a lot of candidates your RAM usage can only keep growing over time.
- the way less obvious one: `ViewportList` only controls which items are rendered in the DOM. It doesn’t manage or clean up "side effects" inside each item. Vega (which we use for the lightweight photometry plots) creates views, timers, subscriptions, event listeners, cached data, ... that `ViewportList` doesn't clean up (which is normal, it's not its purpose)! So, we need to clean these up ourselves.

So, in this PR we:
- add a photometry_minimal ducks, where we fetch and store in redux the bare minimum needed by Vega plots. we also add a method that can clear objects from the cache (specific ones, or all)
- update Vega plots to use the new `photometry_minimal`, and clean it up overall. We add a useEffect to control creation and teardown of the plots and associated "effects", so that when the DOM node they were attached to goes out of view, we clean up after Vega.
- clean the `photometry_minimal` cache when we change pages of the candidates page, to prevent redux from growing infinitely

Additionally, we update API.js to include API parameters in the `<...>_OK` actions (what we dispatch when an API calls finishes successfully, that the various ducks listen to in order to update their cached data), so the ducks can know what query parameters were used to retrieve the associated data. We use it to fix a known issue: when we get photometry back from the API we just get an array and the only way to know what object(s) it's for is to look at the datapoints. But when we have no data, well ... we can't set the array to empty since we don't know the associated `objectId`! But now that the query parameters are attached to the response, we use this to infer the `objectId` used to query photometry in `photometry_minimal.js`. This could also be done in the existing `photometry.js` (but may also require some updates to the components that use that data)

PS: Hard to know in a dev setup if this will work well in prod since we don't have many candidates loaded. I will try to load plenty of data locally, to see how it behaves then.

### DIFF
--- a/static/js/components/plot/VegaPhotometry.jsx
+++ b/static/js/components/plot/VegaPhotometry.jsx
@@ -192,7 +192,7 @@ const VegaPhotometry = (props) => {
     async function fetchPhotometry() {
       if (photometry === null || photometry === undefined) {
         return await dispatch(
-          photometryMinimalActions.fetchSourcePhotometryMini(sourceId),
+          photometryMinimalActions.fetchSourcePhotometryMinimal(sourceId),
         );
       }
       if (
--- a/static/js/ducks/photometry_minimal.js
+++ b/static/js/ducks/photometry_minimal.js
@@ -7,7 +7,7 @@ const FETCH_SOURCE_PHOTOMETRY_MINIMAL_OK =
   "skyportal/FETCH_SOURCE_PHOTOMETRY_MINIMAL_OK";
 const CLEAR_PHOTOMETRY_MINIMAL = "skyportal/CLEAR_PHOTOMETRY_MINIMAL";
 
-export function fetchSourcePhotometryMini(id) {
+export function fetchSourcePhotometryMinimal(id) {
   return API.GET(
     `/api/sources/${id}/photometry`,
     FETCH_SOURCE_PHOTOMETRY_MINIMAL,


Theodlz:
Ok I can confirm that this works super well. Memory profiling shows that on main, scanning through 3 pages of candidates (many of which, have lots of detections to plot) I am already at 2.5 GB of RAM used. A few more pages and it gets over 5GB used and the page crashes. With the changes in this PR, even after scanning the 10+ pages of candidates I have loaded, it never goes over 1 GB (and I'd say it averages around ~600MB).

What keeps growing - though absolutely not as much as before - isn't the RAM used to render the page, but the RAM used by the tab in general. That's because we use `redux-logger` to logsall the dispatched actions in the console, but unless we turn that off it's unavoidable. And even then, this is really minimal compared to the amount of RAM used by Vega plots when they accumulate (like they do on main), + it doesn't seem to impact rendering performance whatsoever. So, I'm really not worried about this. It's still an awesome debug tool that we should keep if we can afford to.

So overall, this will make scanning a lot smoother!!!

PR #5982, opened 2026-02-05
Filter model: altdata

Theodlz:
It is very useful to be able to add arbitrary metadata to a Filter, especially when want to create connections between skyportal and services that may run the filters on an actual data stream. For example, in our custom instance of SkyPortal that ties it together with BOOM, we store information about the filter (what's written from our custom UI, its ID in the broker ...)


mcoughlin (approved):
LGTM

PR #5986, opened 2026-02-06
enh: add t-t0 column in the photometry table

camilledzt:
As requested by Sarah and Marion, this PR is adding a t-t0 column in the Photometry Table. The column is only shown if the t0 is set on the object.


Theodlz (approved):
LGTM!

PR #5987, opened 2026-02-06
Fix user comments on gemini api

antoine-le-calloch:
- Fix `TypeError: CachedSourceFinderHandler.get() missing 1 required positional argument: 'cache_key'` by handling missing cache key on public finding_charts handler. 
- Cleanup the code.
- Update Gemini user authentication to use user_key instead of user_password. Add the migration for that.
- Fix warning message formatting and refactor parameters in finding charts function.
- Update Gemini API to allow user to choose a custom group_name or the default will use object ID.

_I updated the payload field from user_password to user_key, so I created a migration file. If you think this is too much for this, @Theodlz, we can keep user_password and just use User key as the title to avoid needing a migration._

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -657,6 +657,7 @@ email_service:
 invitations:
   enabled: False  # If debug_login=True above, invite tokens won't be used during auth
   days_until_expiry: 3
+  disable_emailing: False  # If True, the app will generate invite tokens but not send emails
   email_subject: "You've been invited to collaborate on SkyPortal"
   email_body_preamble: | # This can include HTML tags
     Welcome to <b>SkyPortal</b>!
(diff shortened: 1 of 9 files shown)


Theodlz (commented):
As you mentioned in your PR description, the migration stuff seems very overkill for this. I suggest the following: When you read the user key from the `altdata`, you could read from `user_key` with a fallback to `user_password`, that way you do write your code around the new field name but stay backwards compatible = no migration needed at all.

what do you think? A migration on a JSONB column just for this, feels overkill

antoine-le-calloch:
> As you mentioned in your PR description, the migration stuff seems very overkill for this. I suggest the following: When you read the user key from the `altdata`, you could read from `user_key` with a fallback to `user_password`, that way you do write your code around the new field name but stay backwards compatible = no migration needed at all.
> 
> what do you think? A migration on a JSONB column just for this, feels overkill

@Theodlz Very good idea. Done !

Theodlz (approved):
LGTM!

PR #5988, opened 2026-02-11
Update default docker configuration and default config.

antoine-le-calloch:
- Improve the default Docker configuration to support usage within the [skyportal-deploy](https://github.com/skyportal/skyportal-deploy) project.


config.yaml.defaults
```diff
@@ -501,26 +501,21 @@ database:
...
   # According to the Tornado docs at
   # https://www.tornadoweb.org/en/stable/guide/running.html#processes-and-ports:
-  #
-  # Due to the Python GIL (Global Interpreter Lock), it is necessary
-  # to run multiple Python processes to take full advantage of
-  # multi-CPU machines. Typically it is best to run one process per
-  # CPU.
+  #   Due to the Python GIL (Global Interpreter Lock), it is necessary
+  #   to run multiple Python processes to take full advantage of
+  #   multi-CPU machines. Typically, it is best to run one process per CPU.
```
stefanv: All above formatting changes: unless these are mandated by a pre-commit hook, I'd just leave it alone.

config.yaml.defaults
```diff
@@ -556,7 +551,8 @@ server:
     google_oauth2_timeout: 15
 
 services:
-  dask: False
+    disabled: # List of microservices to disable, EDIT THIS AS NEEDED
```
stefanv: Presumably, all configuration entries can be edited as needed.

config.yaml.defaults
```diff
@@ -586,7 +582,7 @@ misc:
   # consider a photometry point as a detection
   photometry_detection_threshold_nsigma: 3.0
 
-  # whether or not the frontend show photometry validation information
+  # whether the frontend show photometry validation information
```
stefanv:
```suggestion
  # whether the frontend should show photometry validation information
```

docker.yaml
```diff
@@ -1,3 +1,13 @@
+app:
+  title: SkyPortal
+  secret_key: abc01234  # This secret key can be any random string of
+                        # characters.
+                        #
+                        # You should re-generate this for your application
+                        # using:
```
stefanv:
```suggestion
                        # To ensure privacy of cookies and websocket communications, you
                        # should re-generate this token for your application using:
```

docker.yaml
```diff
@@ -6,7 +16,84 @@ database:
     password: password
 
 server:
-    port: 5000 # Internal port
```
stefanv: Need to decide on spacing: 2 vs 4, and keep it consistent throughout the file.

docker.yaml
```diff
@@ -6,7 +16,84 @@ database:
...
+  # From https://console.developers.google.com/
+  # - Create Client ID
+  # - Javascript origins: https://localhost:5000
+  # - Authorized redirect URLs: http://localhost:5000/complete/google-oauth2/
+  # You need to have Google+ API enabled; it takes a few minutes to activate.
+  protocol: http
+  host: localhost
+  port: 443 # This is the public-facing port (set to 80 if not using SSL)
```
stefanv: Hrm, presumably this will impact the auth port above?

docker.yaml
```diff
@@ -6,7 +16,84 @@ database:
     password: password
 
 server:
-    port: 5000 # Internal port
+  # From https://console.developers.google.com/
+  # - Create Client ID
+  # - Javascript origins: https://localhost:5000
+  # - Authorized redirect URLs: http://localhost:5000/complete/google-oauth2/
+  # You need to have Google+ API enabled; it takes a few minutes to activate.
+  protocol: http
```
stefanv: Is this correct if SSL is used?

docker.yaml
stefanv: And should probably be 5000 by default, not 443?

docker.yaml
```diff
@@ -6,7 +16,84 @@ database:
...
+weather:
+  # time in seconds to wait before fetching weather for a given telescope
+  refresh_time: 3600.0
+  # Get an API key at OpenWeatherMap https://openweathermap.org/price
+  # the free tier should be sufficient, as we cache the weather results
+  # for each telescope
+  openweather_api_key:
+
+twilio:
+  # Twilio Sendgrid API configs
```
stefanv:
```suggestion
  # Twilio Sendgrid API configs for sending email and text notifications
```

stefanv: We may want to move away from Sendgrid, if we haven't already, since they no longer have a free plan.

docker.yaml
```diff
@@ -6,7 +16,84 @@ database:
...
+  # This needs to be obtained via Sendgrid setup on their site
+  sendgrid_api_key:
+
+  # Twilio SMS API configs
+  # These values must be valid ones from the Twitter API console
+  from_number:
+  sms_account_sid:
+  sms_auth_token:
+
+smtp:
```
stefanv, line 84:
```suggestion
# See also `email_service` below
smtp:
```

docker.yaml
```diff
@@ -6,7 +16,84 @@ database:
...
+  from_email:  # This is both the email used to authenticate with smtplib and the sender address
+  password:
+  host: smtp.gmail.com  # Default Gmail settings; modify as needed
+  port: 587  # Default Gmail settings; modify as needed
+
+# This value needs to be either "sendgrid" or "smtp" (without quotes)
+email_service:
+
+invitations:
+  enabled: False  # If debug_login=True above, invite tokens won't be used during auth
```
stefanv: There's no debug_login in this file. Either add or adapt this comment.

docker.yaml
stefanv: If it is 443, shouldn't it be something like 5000:443, to expose the internal port?

stefanv (changes requested):

config.yaml.defaults
antoine-le-calloch: Because the configuration seems very huge to me, I was trying to make it as small as possible.

docker.yaml
antoine-le-calloch: `protocol` is used in some files, but the main URL is retrieved using SSL, so I will update the code to remove it.
```
def get_app_base_url():
    ports_to_ignore = {True: 443, False: 80}  # True/False <-> server.ssl=True/False
    return f"{'https' if cfg['server.ssl'] else 'http'}://{cfg['server.host']}" + (
        f":{cfg['server.port']}"
        if (
            cfg["server.port"] is not None
            and cfg["server.port"] != ports_to_ignore[cfg["server.ssl"]]
        )
        else ""
    )
```

docker.yaml
antoine-le-calloch: After reading the code I sent in the next comment, I see that this port is only used when it’s different from the SSL-calculated port, so only for customization. I have update it to 5000.

docker.yaml
antoine-le-calloch: Should I just remove the config part, or also the code that manages it? There’s a risk that someone running SkyPortal might be using it.

antoine-le-calloch:
@stefanv I’ve resolved most of the comments and replied to the others.
I found an issue with the pre-commit configuration that prevented it from running on docker.yaml (the regex was too specific, excluding `docker.yaml` because of `doc`).
I also overrode Prettier via .prettierrc so that config.yaml.default is treated as a YAML file and linted as expected.

.pre-commit-config.yaml
```diff
@@ -19,7 +19,7 @@ repos:
     hooks:
       - id: prettier
         pass_filenames: true
-        exclude: baselayer|node_modules|doc|data|^(.*_template\.html)$
+        exclude: (baselayer|node_modules|doc|data)/|^(.*_template\.html)$
```
Theodlz: what does this change? just curious

config.yaml.defaults
```diff
@@ -503,26 +501,20 @@ database:
 
 server:
   # From https://console.developers.google.com/
-  #
   # - Create Client ID
   # - Javascript origins: https://localhost:5000
   # - Authorized redirect URLs: http://localhost:5000/complete/google-oauth2/
-  #
   # You need to have Google+ API enabled; it takes a few minutes to activate.
-  protocol: http
```
Theodlz: i see you removed the protocol key. Is it not used anywhere in baselayer or skyportal?

config.yaml.defaults
Theodlz, line 1: very nice cleanup!

docker-compose.yaml
```diff
@@ -1,6 +1,16 @@
...
+    container_name: postgres17
+    environment:
+      PGDATA: /var/lib/postgresql/data/pgdata
+      POSTGRES_DB: skyportal
+      POSTGRES_PASSWORD: password
+      POSTGRES_USER: skyportal
+    restart: on-failure:3
+    volumes:
+      - ./data/dbdata:/var/lib/postgresql/data:z
+
```
Theodlz, line 38: I'm not a fan of putting the "external" images before what's actually specific. I'd rather have the web image/service (which we actually build from our code) first in this list

Theodlz: its a nitpick btw, not an important comment!

docker.yaml
```diff
@@ -1,12 +1,89 @@
+app:
+  title: SkyPortal
+
+  # This secret key can be any random string of characters.
+  # To ensure privacy of cookies and websocket communications, you
+  # should re-generate this token for your application using:
+  # base64.b64encode(os.urandom(50)).decode('ascii')
+  secret_key: abc01234
```
Theodlz, line 8: looks the same as the default config, so might as well remove this?

docker.yaml
Theodlz, line 1: the whole point of docker.yaml was to just change the bare minimum required for stuff to work in docker, but here I see a lot of copy paste from the default config. Why?

docker.yaml
```diff
@@ -1,12 +1,89 @@
...
+    - reminders
+    - slack
+    - sn_analysis_service
+    - spectral_cube_analysis_service
+    - sharing_service_submission_queue
+    - tns_retrieval_queue
+    - watch_list
+    - openai_analysis_service
+    - ngsf_analysis_service
+    - thumbnail_queue
```
Theodlz, line 70: i don't see why we would hardcode disable all this by default in docker.

docker.yaml
```diff
@@ -1,12 +1,89 @@
...
+
+# This value needs to be either "sendgrid" or "smtp" (without quotes)
+email_service:
+
+# See also `email_service` above.
+smtp:
+  from_email: # This is both the email used to authenticate with smtplib and the sender address
+  password:
+  host: smtp.gmail.com # Default Gmail settings; modify as needed
+  port: 587 # Default Gmail settings; modify as needed
```
Theodlz, line 88: why put this by default as gmail?

docker.yaml
```diff
@@ -1,12 +1,89 @@
...
+  # According to the Tornado docs at
+  # https://www.tornadoweb.org/en/stable/guide/running.html#processes-and-ports:
+  #   Due to the Python GIL (Global Interpreter Lock), it is necessary
+  #   to run multiple Python processes to take full advantage of
+  #   multi-CPU machines. Typically, it is best to run one process per CPU.
+  processes: 4
+  # How many of the above processes should be dedicated to
+  # frontend only (i.e., no token authorized API requests)
+  dedicated_frontend_processes: 2
+
```
Theodlz, line 49: same as default?

Theodlz (commented):
I have to admit that I am confused by some of the changes, that don't seem correct to me

antoine-le-calloch:
Fixed here https://github.com/skyportal/skyportal/pull/6690 and here https://github.com/skyportal/skyportal/pull/6689

Issue #5989, opened 2026-02-11
user/<id>/roles with { "roleIds": [ "Super admin" ] } fails, seems to be trying to change a group.


rknop:
I have a user:

```
skyportal=# SELECT id,username,contact_email FROM users;
 id |     username      | contact_email  
----+-------------------+----------------
  1 | provisioned-admin | 
  2 | raknop            | raknop@lbl.gov
(2 rows)
```

I made an API call:

`>>> res = requests.post( 'https://ls4-skyportal.lbl.gov/api/user/2/roles', json={ "roleIds": [ "Super admin" ] }, headers=headers )`

where `headers` has the necessary token authentication.  (I previously used the same authentication token to send the user invitation, which worked; I'm using `INITIAL_ADMIN` token from `/skyportal/.tokens.yaml` on the server.)

The response I get is an error:

```
>>> pprint( res.json() )
{'data': {},
 'message': '(psycopg2.errors.UniqueViolation) duplicate key value violates '
            'unique constraint "ix_groups_name"\n'
            'DETAIL:  Key (name)=(raknop) already exists.\n'
            '\n'
            '[SQL: UPDATE groups SET name=%(name)s, '
            'modified=timezone(%(timezone_1)s, CURRENT_TIMESTAMP) WHERE '
            'groups.id = %(groups_id)s]\n'
            "[parameters: {'name': 'raknop', 'timezone_1': 'UTC', 'groups_id': "
            '1}]\n'
            '(Background on this error at: https://sqlalche.me/e/20/gkpj)',
 'status': 'error',
 'version': '1.4.0'}
```

It looks like it's trying to change the name of the group with id 1 (which is the `provisioned-admin` group).  I suspect that that is not what this API endpoint is supposed to be doing?  In any event, I've done this in months (years?) past to set a super user admin, and from the API docs I believe it should work, but it's not.

I'm on the latest `main` branch (commit `12e67a35`)

PR #5990, opened 2026-02-12
Fix update roles fail

antoine-le-calloch:
Improve the update logic for single-user group names when a username changes. The single-user group name should only be updated when the username is modified, not on every user update (e.g., role changes). [issue 5989](https://github.com/skyportal/skyportal/issues/5989)


skyportal/models/user_token.py
```diff
@@ -468,16 +468,6 @@ def receive_after_flush(session, context):
             session.delete(new_single_user_group)
 
 
-@event.listens_for(User, "after_update")
```
Theodlz, line 471: Any specific reason to not do it here as an event.listen? The advantage of these is that if the username is edited anywhere, they will trigger (so you don't need to cppy/paste that logic everywhere a username change is possible). 

As far as I can tell, the only change to the username is in the patch profile, so I'm fine removing it from here anyways. But, I was just curious about what motivated that change.

antoine-le-calloch: Currently, every time a user is updated, we attempt to update the group table, even when the username has not changed. This was leading to errors in cases where there are inconsistent or unexpected user/group relationships, as seen in the linked issue, even when updating the roles associated with a user.

This change does not fix the root cause of the issue, since I still haven’t been able to inspect the actual values in the user/group tables when the error occurs. However, it does prevent an unnecessary and incorrect update operation.

We could filter the updates directly in the function itself, but I found it clearer and more logical to handle this at the endpoint level. Even after considering the possibility of making this change elsewhere, it was not clear to me in which situation we would want this behavior to occur from another endpoint.

skyportal/handlers/api/roles.py
```diff
@@ -74,19 +76,38 @@ def post(self, user_id, *ignored_args):
         new_role_ids = data.get("roleIds")
         if new_role_ids is None:
             return self.error("Missing required parameter roleIds")
-        if not isinstance(new_role_ids, list | tuple) or not all(
-            Role.query.get(r) for r in new_role_ids
+        if not isinstance(new_role_ids, list) or not all(
+            isinstance(role_id, str) for role_id in new_role_ids
         ):
```
Theodlz, line 81: now that you don't check if they are valid roles in the DB, should you make that check later as a separate validation step?

skyportal/handlers/api/internal/profile.py
```diff
@@ -186,16 +186,20 @@ def patch(self, user_id=None):
...
-                ).first()
-                if single_user_group is not None:
-                    single_user_group.name = slugify(username)
+                )
+                if single_user_group_id is not None:
+                    session.execute(
+                        sa.update(Group)
+                        .where(Group.id == single_user_group_id)
+                        .values(name=slugify(username))
+                    )
```
Theodlz, line 202: what does this change vs just in place chnge + commit later, which is what we usually do?

PR #6003, opened 2026-03-05
Enh : adding group source widget and refactor telescope map frontend

camilledzt:
The goal of the PR is to adding:
* a shortcut widget to group sources.
* a refactoring of the telescope map:

From:
Screenshot from 2026-02-25 16-51-53

To:
Screenshot from 2026-02-25 16-51-23

Labels are shown on hover and if the zoom is sufficient so that labels don't overlap with each other. There is also a reset button on the bottom right to reset the zoom to the default view.

### DIFF
--- a/static/js/components/widget/TelescopeMapWidget.jsx
+++ b/static/js/components/widget/TelescopeMapWidget.jsx
@@ -8,7 +8,7 @@ import { Tooltip, IconButton } from "@mui/material";
 import HelpOutlineOutlinedIcon from "@mui/icons-material/HelpOutlineOutlined";
 import ReplayIcon from "@mui/icons-material/Replay";
 
-const TelescopeMap = lazy(() => import("./TelescopeMap"));
+const TelescopeMap = lazy(() => import("../telescope/TelescopeMap"));
 
 const useStyles = makeStyles((theme) => ({
   mapContainer: {


Theodlz:
> Hello @Theodlz, do you think that I should move the [Telescope Map widget](https://github.com/camilledzt/icare/blob/enh_telescope_map/extensions/skyportal/static/js/components/telescope/TelescopeMapDashboard.jsx) from the ICARE repo to Skyportal? Or keep it as it is and create another PR for the widget on the ICARE repo?

I think you can move it here! Might as well maintain it in SkyPortal and anyways we can decide to enable it or not from the config.

Theodlz:
> Short video to show GroupSourceList in action :
> 
>  Screencast.from.03-16-2026.10.31.01.AM.webm

ok so basically it's an overwrite of the existing widget (that sends you to the group's page) where you send folks to the groups' source page instead? Since it's a copy/paste of an existing widget where the only change (at least that I can see) is where folks are sent on click, I guess I'm wondering if there isn't a better approach than duplicating. If the component was using the template system (that we use for a bunch of other components) then you should be able to make that configurable int he config without having 2 distinct widgets! Could you give that a go?

static/js/components/widget/TelescopeMapWidget.jsx
Theodlz, line 1: to fit in nicely with the existing widget, should this be renamed TelescopeMapWidget and moved to `static/js/components/widget` @camilledzt ?

camilledzt: Yes no problem !

PR #6008, opened 2026-03-09
Python package & env management: move to uv

Theodlz:
This PR aims to move away from the regular good-old `pip`-driven environment, to super-duper-fast `uv` (which is now what baselayer supports). Will take a few tries to get everything up and running CI wise I believe.

TODOs:
- [x] debug the CI
- [x] move to uv in the Dockerfile too
- [x] Update the docs to only rely on uv

Note: I wonder if `uv` "lazily" installs some packages. I noticed that while installing packages is a lot faster, initial statup time of the app (with a new venv) may be slower than before. I think that when you sum the time to create the venv + install packages + start the app `uv` is still the winner, but yeah just something I noticed.

### DIFF
--- a/.github/workflows/test_migrations.yaml
+++ b/.github/workflows/test_migrations.yaml
@@ -209,7 +209,7 @@ jobs:
           make dependencies
           # Make sure to remove python packages from the reference commit that conflict
           # with the current branch (e.g pinecone-client that has later been renamed to pinecone).
-          uv pip uninstall -y pinecone pinecone-plugin-assistant pinecone-plugin-interface pinecone-client
+          uv pip uninstall pinecone pinecone-plugin-assistant pinecone-plugin-interface pinecone-client
           # Make sure latest revision on main is included in Alembic history
           n_current_head=$( PYTHONPATH=. .venv/bin/python -m alembic -x config=config.yaml history | grep -c $CURRENT_HEAD )
           if [[ $n_current_head -eq 0 ]]


.github/workflows/test_api_and_frontend.yaml
```diff
@@ -149,34 +149,34 @@ jobs:
       - name: Run front-end tests part 1
         if: ${{ matrix.test_subset == 'frontend_pt1' }}
         run: |
-          PYTHONPATH=. python baselayer/tools/test_frontend.py --xml 'skyportal/tests/frontend/sources_and_observingruns_etc'
+          source .venv/bin/activate && PYTHONPATH=. python baselayer/tools/test_frontend.py --xml 'skyportal/tests/frontend/sources_and_observingruns_etc'
```
stefanv, line 152: Curious why we need to activate here, instead of `uv run`-ing.

.github/workflows/test_migrations.yaml
```diff
@@ -87,6 +94,7 @@ jobs:
 
           pip install pip==24.0
           pip install wheel numpy
+          pip install "setuptools<76.1.0"
```
stefanv, line 97: What's the problem with newer setuptools?

.github/workflows/test_migrations.yaml
```diff
@@ -164,9 +170,16 @@ jobs:
           psql -U skyportal -h localhost -c "GRANT ALL PRIVILEGES ON DATABASE skyportal_test TO skyportal;" skyportal_test
 
           export NPM_CONFIG_LEGACY_PEER_DEPS="true"
+
+          pip install black
+          pip install 'fiona==1.9.6' # 1.10.0 breaks geopandas
+          pip install 'joblib<1.5.0' # 1.5.0 breaks skyportal/utils/offset.py on reference commit
+          pip install 'lalsuite<7.26' # 7.26 breaks ligo.skymap
+          pip install 'python-ligo-lw==1.8.4' # 2.0.0 breaks ligo.skymap
+          pip install 'numexpr==2.13.1' # 2.14.0 breaks multiple packages
```
stefanv, line 179: What is happening here?

.github/workflows/test_migrations.yaml
```diff
@@ -196,9 +209,9 @@ jobs:
           make dependencies
           # Make sure to remove python packages from the reference commit that conflict
           # with the current branch (e.g pinecone-client that has later been renamed to pinecone).
-          pip uninstall -y pinecone pinecone-plugin-assistant pinecone-plugin-interface pinecone-client
+          uv pip uninstall pinecone pinecone-plugin-assistant pinecone-plugin-interface pinecone-client
           # Make sure latest revision on main is included in Alembic history
-          n_current_head=$( PYTHONPATH=. alembic history | grep -c $CURRENT_HEAD )
+          n_current_head=$( PYTHONPATH=. .venv/bin/python -m alembic -x config=config.yaml history | grep -c $CURRENT_HEAD )
```
stefanv, line 214: Should we invoke uv, not .venv/bin/python?

.github/workflows/test_models.yaml
```diff
@@ -137,18 +135,19 @@ jobs:
...
+          uvx pre-commit run --from-ref ${{ github.event.pull_request.base.sha }} --to-ref ${{ github.event.pull_request.head.sha }}
 
       - name: Test loading demo data
         run: |
           make run &
           sleep 180 && make load_demo_data
-          kill %1
+          make stop
+          kill %1 2>/dev/null || true
+          sleep 5
```
stefanv, line 146: What are we doing here?

Dockerfile
```diff
@@ -8,6 +8,10 @@ ENV NODE_MAJOR=20
 ENV NPM_CONFIG_LEGACY_PEER_DEPS=true
 ENV PATH="/root/.cargo/bin:${PATH}"
 ENV SNCOSMO_DATA_DIR=/skyportal/persistentdata/sncosmo
+ENV UV_NO_DEV=1
```
stefanv, line 11: Meaning?

.github/workflows/test_migrations.yaml
Theodlz: it breaks a bunch of dependencies (e.g. fiona)

.github/workflows/test_migrations.yaml
Theodlz: basically, we checkout to the old migration reference commit. Issue is that branch did not have a lock file and some dependencies are now broken (because folks didn't properly pin versions of their own dependencies). This fixes that by ensuring we run versions that don't break the environment.

Theodlz: it's either this or we update the migration reference commit. I'd rather not update it, since that maximizes how many migration scripts we cover.

.github/workflows/test_migrations.yaml
Theodlz: does `uv run` auto-magically pick the `.venv` ?

.github/workflows/test_api_and_frontend.yaml
Theodlz: does `uv run` auto-magically pick the `.venv` ?

.github/workflows/test_models.yaml
Theodlz: killing the main thread may not properly shut down the many processes handles by supervisor. I found that I run into less issues rerunning the CI locally multiple times in a row, if I called `make stop` that gracefully shuts down everything managed by supervisor. A short sleep helped a little as well but that I can probably remove.

Dockerfile
Theodlz: https://docs.astral.sh/uv/reference/environment/#uv_no_dev

`Equivalent to the --no-dev command-line argument. If set, uv will exclude development dependencies.`

we don't have a dev dependency group pinned but I felt it was a nice addition in case we do add these later, i.e. to make sure we only install was is required for production.

Theodlz: happy to remove if you think it's just noise

.github/workflows/test_api_and_frontend.yaml
Theodlz: what I do like - though I hope it won't happen anytime soon. Is that by sourcing the .venv we make this uv agnostic, meaning it will keep working as long as the venv name does not change (and the package/python manager doesn't matter).

.github/workflows/test_migrations.yaml
Theodlz: also same comment as above, about being uv-agnostic

.github/workflows/test_api_and_frontend.yaml
stefanv: Right, if that is a use-case we want to support, then we should probably make it explicit. I don't mind it!

`uv run` should run inside the environment, I thought that's its whole purpose.

.github/workflows/test_migrations.yaml
stefanv: Maybe we need instructions here on how & when to remove the workaround.

.github/workflows/test_migrations.yaml
stefanv: Makes sense!

PR #6022, opened 2026-03-16
Fix tns utils

antoine-le-calloch:
- Refactor TNS utils file, remove unused code, clean up, and fix logic issues (e.g., if photometry["limflux"] == "" not working when limflux is None).
- Update db_demo.yaml to restructure altdata so boom is used as the key, as expected by boom filters.
- Add new telescopes and instruments to db_demo.yaml required when querying TNS.
- Add the required telescopes and instruments to instruments.yaml and telescopes.yaml for TNS.
- Update db_demo.yaml to modify SOAR telescope details and add a new entry for the Southern Astrophysical Research Telescope.

### DIFF
--- a/skyportal/utils/tns.py
+++ b/skyportal/utils/tns.py
@@ -218,7 +218,7 @@ def get_recent_TNS(api_key, headers, public_timestamp, get_data=True):
 
     r = _tns_request(api_key, headers, req_data, retry_delay=30)
 
-    if r is None or r.status_code != 200:
+    if r.status_code != 200:
         try:
             content = r.json()
         except Exception:


skyportal/utils/tns.py
```diff
@@ -41,15 +39,18 @@
...
     "efosc2": 30,
     "gaia": 163,
     "goodman": 136,
     "goto": [218, 264, 265, 266],
+    "kast": 10,
+    "lsst": 287,
     "ps1": [98, 154, 155, 257],
     "sedm": [149, 225],
     "sprat": 156,
+    "wifes": 115,
```
Theodlz, line 53: you have more than one??? crazy

Issue #6023, opened 2026-03-17
Photometry plot legend shows 'undefined' for photometric_series.


howardisaacson:
We use photometric_series quite a lot, so I will work on this fix. Suggestions welcome.

---

**Describe the bug**
Photometry plots sometimes display `undefined` as the instrument name in the legend for data uploaded via `photometric_series`. This occurs because the frontend expects `instrument_name`, while photometric_series rows returned by the API may instead contain `instrument` or `telescope`.

**To Reproduce**

1. Upload photometry using the `/api/photometric_series` endpoint.
2. Open the corresponding source page in SkyPortal.
3. Scroll to the **Photometry** plot.
4. Observe legend entries beginning with `undefined`, e.g.:

```
undefined/standard::i
```

**Expected behavior**
The legend should display the instrument name and filter, e.g.:
```
KMTNet-SSO/standard::i
```
The frontend should resolve the instrument using a fallback chain such as:
```
instrument_name || instrument || telescope
```
---

**Platform information:**

* SkyPortal version: 1.4.0
* 
* Interface
  * [x] I am browsing on a desktop

**Additional context**

Regular photometry uploads include `instrument_name` and render correctly.
Photometric series responses sometimes include only:

```
instrument
telescope
```

which causes the legend builder to produce `undefined`.
A small frontend change in the photometry plot component to resolve the instrument via

```
instrument_name || instrument || telescope
```

would fix this while maintaining backward compatibility.

PR #6032, opened 2026-03-25
Removing 'undefined' from photometry legend.

howardisaacson:
This single file fix removed a persistent 'undefined' label in the photometry plot legend that comes from photometric_series. The current fix truncates the legend, preserving previous behavior. I tested it with a legend that allows more characters, which I can implement upon request.

Issue [#6023](https://github.com/skyportal/skyportal/issues/6023)

This screenshot shows that the fix removes  'undefined' from the legend.
Screenshot 2026-03-24 at 10 54 14 PM

### DIFF
--- a/static/js/components/plot/PhotometryPlot.jsx
+++ b/static/js/components/plot/PhotometryPlot.jsx
@@ -510,15 +510,29 @@ const PhotometryPlot = ({
     // we will use these values to set the range of the plot
 
     const groupedPhotometry = photometryData.reduce((acc, point) => {
-      let key = `${getPhotometryInstrumentLabel(point)}/${point.filter}`;
+      const instrumentLabel = getPhotometryInstrumentLabel(point);
+      const truncatedInstrument =
+        instrumentLabel.length > 10
+          ? `${instrumentLabel.substring(0, 10)}...`
+          : instrumentLabel;
+      let key = `${truncatedInstrument}/${point.filter}`;
       // if we are using duplicates, put the obj_id at the beginning of the key
       if (usingDuplicates) {
         key = `${point.obj_id}/${key}`;
       }
-      const MAX_LABEL_LENGTH = usingDuplicates ? 33 : 23;
-
-      if (key.length > MAX_LABEL_LENGTH) {
-        key = `${key.substring(0, MAX_LABEL_LENGTH - 3)}...`;
+      if (
+        point?.origin !== "None" &&
+        point.origin !== "" &&
+        point.origin !== null &&
+        point.origin !== undefined
+      ) {
+        // the origin is less relevant, so we crop it to not have more than 23 characters + 3 x ...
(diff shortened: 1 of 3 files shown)


static/js/components/plot/PhotometryPlot.jsx
```diff
@@ -507,23 +510,15 @@ const PhotometryPlot = ({
...
-        const remaining = (usingDuplicates ? 33 : 23) - key.length;
-        if (remaining < point.origin.length) {
-          key += `/${point.origin.substring(0, Math.max(remaining - 3, 3))}...`;
-        } else {
-          key += `/${point.origin}`;
-        }
+      const MAX_LABEL_LENGTH = usingDuplicates ? 33 : 23;
+
+      if (key.length > MAX_LABEL_LENGTH) {
+        key = `${key.substring(0, MAX_LABEL_LENGTH - 3)}...`;
```
antoine-le-calloch: We’d prefer not to truncate the all key, since the instrument_name/filter information is important to display. With longer instrument names, truncation can hide the filter (as shown in your example screenshot).
We’re only comfortable truncating the origin, as it’s less relevant.

static/js/components/plot/PhotometryPlot.jsx
```diff
@@ -507,23 +510,15 @@ const PhotometryPlot = ({
...
       }
```
antoine-le-calloch, line 947: Is there a specific reason for removing the origin from the key?

antoine-le-calloch (commented):
Thanks for the fix.
Some comments:

antoine-le-calloch:
> I made one more change. The order of the legend is currently Instrument:filter:origin. I truncated the instrument to 10 characters, to ensure the filter is visible. The overall truncation is the same.
> 
> What do you think of this? Screenshot 2026-03-26 at 3 12 15 PM

Sorry for the late reply. I think truncating the instruments can also be an issue since some instruments have the same prefix but not the same full name. I think what we already have is good because you are able to scroll horizontally to see all the text. You could also use tooltips if you want to make it smaller.

howardisaacson:
OK. I'll leave it how it was, but retain the change that fixes the 'undefined' for photometric_series.

mcoughlin:
@howardisaacson where does this stand? can you resolve the conflict?

howardisaacson:
I updated the base branch. Happy to resolve any other issues here.

antoine-le-calloch (changes requested):
I still see the truncation on the instrument name. As mentioned above https://github.com/skyportal/skyportal/pull/6032#issuecomment-4200313341, we'd rather not truncate it: two instruments sharing the same 10-char prefix would end up merged into a single trace.

howardisaacson:
Roger that. I removed the truncation.

antoine-le-calloch (approved):
LGTM. Thanks a lot for the fix.

PR #6063, opened 2026-04-04
Upgrade to numpy 2+

mcoughlin:
This PR attempts to upgrade SkyPortal to Numpy 2+.


skyportal/models/__init__.py
```diff
@@ -1,3 +1,24 @@
+# Monkey-patch matplotlib for simsurvey compatibility with matplotlib >= 3.8
+# simsurvey imports matplotlib.docstring which was renamed to matplotlib._docstring
+import matplotlib
+
+if not hasattr(matplotlib, "docstring"):
+    matplotlib.docstring = matplotlib._docstring
+
```
Theodlz: maybe this should be in skyportal's __init__.py, rather than in models?

skyportal/tests/frontend/sources_and_observingruns_etc/test_top_sources.py
```diff
@@ -48,7 +48,7 @@ def test_top_sources(driver, user, public_source, public_group, upload_data_toke
 
     # Test that front-end views register as source views
     driver.click_xpath(f'//a/span[contains(.,"{obj_id}")]')
-    driver.wait_for_xpath(f'//div[text()="{obj_id}"]')
+    driver.wait_for_xpath(f'//*[text()="{obj_id}"]')
```
Theodlz: Not a fan of that at all, because that could match to anything on that page. Not that its worse than the current `div`, that also isn't great. So since this isn't a regression I'm fine merging, but yeah let's just keep in mind that more specific paths are preferred, to not match with a random element on the page that looks similar to the one we are looking for.

pyproject.toml
```diff
@@ -124,6 +123,7 @@ docs = [
 [tool.uv]
 package = false
 workspace = { members = ["baselayer"] }
+no-binary-package = ["sncosmo"]
```
Theodlz: what does this do?

alembic/versions/2b8addba3e86_numpy2.py
Theodlz, line 1: it feels like this migration is a lot more complex than it needs to be. Can't we just update in place and push the new values where they need to be? I remember existing migrations doing just that.

Theodlz (commented):
Overall this seems like a nice reboot of #6016, that got more complicated than it needed to be! So, thanks a ton.

I do have some comments that I think need to be addressed before merging

skyportal/models/__init__.py
mcoughlin: Moved it.

skyportal/tests/frontend/sources_and_observingruns_etc/test_top_sources.py
mcoughlin: Changed to h6.

alembic/versions/2b8addba3e86_numpy2.py
mcoughlin: I tried like 10 things and they all failed.

pyproject.toml
mcoughlin: sncosmo doesn't have binaries for macos. I pinged the maintainer about it.

skyportal/handlers/api/photometry.py
```diff
@@ -62,6 +62,43 @@
...
+    """
+    if isinstance(value, np.integer):
+        return int(value)
+    if isinstance(value, np.floating):
+        return float(value)
+    if isinstance(value, np.bool_):
+        return bool(value)
+    return value
+
+
```
Theodlz, line 101: reading these 2 things (which btw this is great, that was the missing piece of the puzzle for my PR that got completely out of hands...), I wonder if these are needed in other places where we dump JSON data before/after inserting in the DB. Asking because it feels like they should be part of some utils in skyportal/utils that we import here, rather than some photometry specific thing

Theodlz (commented):
just one small comment, something that I missed before. I'm now playing with the migration stuff and I'll commit something here if I figure it out.

Theodlz:
> just one small comment, something that I missed before. I'm now playing with the migration stuff and I'll commit something here if I figure it out.

Ok migration stuff done and dusted. This new migration should avoid having to do an expensive scan over the whole photometry table (so many rows in prod, I'd be afraid of running anything over that).

mcoughlin:
Thanks @Theodlz ! That was giving me a real headache.

Theodlz:
> Thanks @Theodlz ! That was giving me a real headache.

For reference, I:
- checked out to main and built the database, got into the postgres shell and printed the enum to see what's in there and in what order
- did the same with your branch (wiping the DB clean in between)
- you're left with an ordered list of bandpasses before vs after the sncosmo migration, and with some manual inspection can check where new values should be inserted in a migration script so that the order of the migrated enum matches that of a new one.

BTW Claude gave me this trick to make it easier to compare the enums and make sure what I print is ordered:
```text
\copy (
  SELECT enumlabel, enumsortorder
  FROM pg_enum
  JOIN pg_type ON pg_enum.enumtypid = pg_type.oid
  WHERE pg_type.typname = 'bandpasses'
  ORDER BY enumsortorder
) TO '/tmp/bandpasses.csv' WITH CSV HEADER
```
that will create a CSV with the values as one column and their index as the other, makes it very easy to do comparisons.

PR #6072, opened 2026-04-16
Fix some hidden docs

antoine-le-calloch:
Add format_doc decorator to inject constants into handler docstrings, replacing broken f-string docstrings that apispec was hiding.

### DIFF
--- a/skyportal/handlers/base.py
+++ b/skyportal/handlers/base.py
@@ -8,6 +8,22 @@ from baselayer.app.handlers.base import BaseHandler as BaselayerHandler
 from .. import __version__
 
 
+def format_doc(**kwargs):
+    """Inject values into a handler method's docstring placeholders.
+
+    Use as a decorator on a handler method whose OpenAPI docstring contains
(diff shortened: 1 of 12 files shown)


skyportal/handlers/base.py
```diff
@@ -8,6 +8,33 @@
...
 
 
+def format_doc(**kwargs):
+    """Inject values into a handler method's docstring placeholders.
+
+    Use as a decorator on a handler method whose OpenAPI docstring contains
+    `{name}` placeholders that should be filled with module-level constants
+    (e.g. pagination limits) at import time.
+    """
+
```
Theodlz, line 21: My only negative comment is that to a random person looking at the code, nothing really tells them "why" this decorator system is needed, and not just an f-string. It would be good to document this here (i.e. what problem this is solving).

Theodlz (commented):
Just one comment

skyportal/handlers/base.py
antoine-le-calloch: Done!

Theodlz (approved):

PR #6075, opened 2026-04-20
fix: fix division by zero when computing object position with zero fluxerr

camilledzt:
This PR is fixing the ZeroDivisionError we have on a ICARE source (it comes from 3 phot points with fluxerr=0)

image

Photometry points with fluxerr=0 passed the nan/null pre-filter and caused a ZeroDivisionError line 272 in obj.py. I have added fluxerr != 0 check to exclude them before the division.

### DIFF
--- a/skyportal/handlers/api/obj.py
+++ b/skyportal/handlers/api/obj.py
@@ -282,6 +282,7 @@ class ObjPositionHandler(BaseHandler):
                     for p in photometry
                     if not np.isnan(p.flux)
                     and not np.isnan(p.fluxerr)
+                    and p.fluxerr != 0
                     and p.ra is not None
                     and not np.isnan(p.ra)
                     and p.dec is not None


mcoughlin:
Probably not for this PR, but perhaps we should also prevent inputting fluxerr==0?

Theodlz:
> Probably not for this PR, but perhaps we should also prevent inputting fluxerr==0?

Yes I was thinking the same thing. This PR prevents bad data from breaking a specific feature that uses it, but it would be good to prevent bad data to begin with. But yeah, doesn't need to be done here

Theodlz (approved):

antoine-le-calloch (approved):
Issue created https://github.com/skyportal/skyportal/issues/6076.

camilledzt:
Hello! Yes, I agree with you! Thanks for creating the issue @antoine-le-calloch . I think that GRANDMA people added points like this, but it should have been a limiting mag. The reviewer of the data asked me to remove these points this morning anyway.

PR #6080, opened 2026-04-24
fix: bugfix patch /api/photometry drops origin updates

camilledzt:
When we are sending a PATCH request to /api/photometry/<id> with a new origin value, it returns HTTP 200 but the field is never updated in the database.

Adding origin to the Photometry constructor in both parse_flux and parse_mag is resolving the issue.

This is needed by GRANDMA users to update origin of some points.

### DIFF
--- a/skyportal/models/schema.py
+++ b/skyportal/models/schema.py
@@ -871,6 +871,8 @@ class PhotometryFlux(_Schema, PhotBase):
         )
         if "alert_id" in data and data["alert_id"] is not None:
             p.alert_id = data["alert_id"]
+        if "origin" in data:
+            p.origin = data["origin"]
         return p
 
 
@@ -1062,6 +1064,8 @@ class PhotometryMag(_Schema, PhotBase):
         )
         if "alert_id" in data and data["alert_id"] is not None:
             p.alert_id = data["alert_id"]
+        if "origin" in data:
+            p.origin = data["origin"]
         return p
 
 
--- a/skyportal/models/schema.py
+++ b/skyportal/models/schema.py
@@ -871,7 +871,7 @@ class PhotometryFlux(_Schema, PhotBase):
         )
         if "alert_id" in data and data["alert_id"] is not None:
             p.alert_id = data["alert_id"]
-        if "origin" in data and data["origin"] is not None and data["origin"] != "":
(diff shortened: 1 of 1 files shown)


antoine-le-calloch (approved):

skyportal/models/schema.py
```diff
@@ -871,6 +871,8 @@ def parse_flux(self, data, **kwargs):
         )
         if "alert_id" in data and data["alert_id"] is not None:
             p.alert_id = data["alert_id"]
+        if "origin" in data:
```
Theodlz: are we making sure that origin is not None, empty string, .... something like that? A person could be sending an empty value and I doubt this means we should remove the current value. What do you think?

camilledzt: Done :)

Theodlz: Did you chat with the grandma folks about this? I wonder if they actually do want to remove rather than replace. Just curious!

camilledzt: It's to replace, because they upload data with the origin "final_dahlia" or whatever and it's not very clean like this.

skyportal/models/schema.py
```diff
@@ -871,6 +871,8 @@ def parse_flux(self, data, **kwargs):
         )
         if "alert_id" in data and data["alert_id"] is not None:
             p.alert_id = data["alert_id"]
+        if "origin" in data and data["origin"] is not None and data["origin"] != "":
```
Theodlz: I suggest just doing something like `if isinstance(data.get("origin"), str) and strip(data["origin"]) != "":` or similar.

Because right now, if someone sends an empty string that just contains spaces, it will still go through

skyportal/models/schema.py
```diff
@@ -1062,6 +1064,8 @@ def parse_mag(self, data, **kwargs):
...
```
Theodlz: same here perhaps

Theodlz (changes requested):

skyportal/models/schema.py
camilledzt: done

skyportal/models/schema.py
camilledzt: done

Theodlz (approved):
LGTM 🚀. Thanks Camille

PR #6081, opened 2026-04-24
SkyPortal __version__ import: don't require full venv

Theodlz:
I ran into issues getting Fritz up and running w/ the last version of skyportal, where we monkeypatch matplotlib and numpy, because Fritz reads `skyportal/__init__.py` to grab `__version__` but doesn't have matplotlib and numpy as this point. It doesn't make sense to require these dependencies just to grab a version number, and in general there is no need to monkeypatch a library that's not installed! So, this PR adds skip monkey-patching if the libraries aren't installed yet (so one can import skyportal.__version__ without having a full env going on).

### DIFF
--- a/skyportal/__init__.py
+++ b/skyportal/__init__.py
@@ -1,25 +1,27 @@
 __version__ = "1.4.0"
 
-# Monkey-patch matplotlib for simsurvey compatibility with matplotlib >= 3.8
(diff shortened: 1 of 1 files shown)


stefanv (approved):

mcoughlin (approved):
LGTM

PR #6082, opened 2026-04-26
Fix ps1 thumbnail display

antoine-le-calloch:
- Improve the display of `PanSTARRS DR2: Loading...` so that it is only shown when there is at least one item in the archivalThumbnails list (which usually means the thumbnails have not yet been processed, so PanSTARRS DR2 is not really loading).
- Improve the display of the StarList on the source page.
- Enhance thumbnail loading
- Enhance thumbnail error
- Add retry logic for thumbnail fetching to handle rate limits (usefull for "sdss") and improve error handling to display the real status of the archival thumbnail for "ls" and "sdss".
- Improve backToTop buttons style on the scanning page

I added some fetching logic using useEffect on the Thumbnail page (which might look a bit unusual), but it’s because I discovered several inconsistencies in thumbnail statuses in production:
- Legacy Survey was showing “Outside Survey Area” when an error occurred (usually a 429 Too Many Requests on the scanning page), but not when the alert was actually outside the survey area. In that case, they return a grey placeholder image instead of an error, so the code was just showing it.
- SDSS also return a placeholder image when outside the survey area that the code was showing, but also return a 404 with the message: `NOT FOUND. Requested (ra, dec) is outside the SDSS footprint.`. This can be used to properly update the status, but wasn’t handled before.

So I implemented a real retry mechanism that eventually resolves to a “Too Many Requests” status if it keeps failing.
And Improved the detection of “Outside Survey Area”:
- For SDSS: using the 404 error message.
- For Legacy Survey: using the size of the grey placeholder image.

_This isn’t perfect, but I couldn’t find a more reliable way to handle these cases._

### Before:
image

### After:
image

### Before:
image

### After:
image

### Before:
image

### After:
Screenshot 2026-04-27 at 12 30 37 AM


static/js/components/source/Source.jsx
```diff
@@ -1020,9 +1018,10 @@ const SourceContent = ({ source }) => {
                 }}
               />
             </div>
+            {showStarList && <StarList sourceId={source.id} />}
             {/* checking if the id exists is a way to know if the user profile is loaded or not */}
             {currentUser?.id &&
-              currentUser?.preferences?.hideSourceSummary !== true && (
+              !currentUser?.preferences?.hideSourceSummary && (
```
Theodlz: btw I think we had an `!== true` before to make sure that it only shows if we have "true" as a value (and not some random thing, here now it would work with a non boolean for example). Not a deal breaker, but just wanted to mention it.

Theodlz (commented):
Very nice! The UI is a LOT better now, so thumbs up for that. Lots of tests are failing (I suppose because of these changes), and I had one other question: I see you simplified our system by removing one component entirely and making another one more flexible. Did you get a chance to make sure:
- it renders well with the right panel both opened/closed, on the source page?
- that it still renders well when looking at a source detail (with the expand arrow) on a source list (e.g. group source page, ...)?

static/js/components/source/Source.jsx
antoine-le-calloch: `hideSourceSummary` can never be random truly value. But you’re right, we can never be too prudent.

antoine-le-calloch:
@Theodlz I added some fetching logic using useEffect on the Thumbnail page (which might look a bit unusual), but it’s because I discovered several inconsistencies in thumbnail statuses in production:
- Legacy Survey was showing “Outside Survey Area” when an error occurred (usually a 429 Too Many Requests on the scanning page), but not when the alert was actually outside the survey area. In that case, they return a grey placeholder image instead of an error, so the code was just showing it.
- SDSS also return a placeholder image when outside the survey area that the code was showing, but also return a 404 with the message: `NOT FOUND. Requested (ra, dec) is outside the SDSS footprint.`. This can be used to properly update the status, but wasn’t handled before.

So I implemented a real retry mechanism that eventually resolves to a “Too Many Requests” status if it keeps failing.
And Improved the detection of “Outside Survey Area”:
- For SDSS: using the 404 error message.
- For Legacy Survey: using the size of the grey placeholder image.

_This isn’t perfect, but I couldn’t find a more reliable way to handle these cases._

Theodlz (approved):

PR #6083, opened 2026-04-28
Fix unique constraint violation when sharing spectra with already-linked groups

antoine-le-calloch:
Fix the error that occurs when adding a list of groups to a spectrum with some groups are already linked.
```
(psycopg2.errors.Unique Violation) duplicate key value violates unique constraint "group_spectra_forward_ind" DETAIL: Key (group_id, spectr_id)=(1747, 52845) already exists. [SQL: INSERT INTO group_spectra (group_id, spectr_id, created_at, modified) VALUES (% (group_id)s, % (spectr_id)s, timezone(% (timezone_1)s, CURRENT_TIMESTAMP), timezone (%(timezone_1)s, CURRENT_TIMESTAMP)) RETURNING group_spectra.id] [parameters: {'group_id: 1747, 'spectr_id': 52845, 'timezone_1': 'UTC'}] (Background on this error at: https://sqlalche.me/e/20/gkpj)
```

- Fix unique constraint violation when sharing spectra and photometry with groups by adding `distinct()` when retrieving groups to avoid duplicates, and by checking for existing group IDs when updating photometry/spectroscopy.
- Move `self.push()` inside the session to avoid `sqlalchemy.orm.exc.DetachedInstanceError`.
- Improve Spectrum preview component positioning in UploadSpectrum for better visibility.
- Remove unnecessary key attributes from button components in various tables for cleaner code.
- Add back navigation link to source in Share, Upload Photometry, and Upload Spectrum components. Clean up code in ShareDataForm.jsx.

### DIFF
--- a/skyportal/handlers/api/sharing.py
+++ b/skyportal/handlers/api/sharing.py
@@ -78,7 +78,7 @@ class SharingHandler(BaseHandler):
 
             if phot_ids:
                 valid_phot = session.scalars(
-                    Photometry.select(session.user_or_token, mode="update").where(
+                    Photometry.select(session.user_or_token).where(
                         Photometry.id.in_(phot_ids)
                     )
                 ).all()
@@ -88,11 +88,17 @@ class SharingHandler(BaseHandler):
                 ]
 
                 if len(invalid_phot_ids) > 0:
-                    return self.error(
-                        f"Cannot share photometry IDs {invalid_phot_ids}: not found or you are not the owner."
-                    )
+                    return self.error(f"Invalid photometry IDs: {invalid_phot_ids}.")
 
                 for phot in valid_phot:
+                    if (
+                        phot.owner_id != self.associated_user_object.id
+                        and "System admin"
+                        not in self.associated_user_object.permissions
+                    ):
+                        return self.error(
+                            f"Cannot share photometry id {phot.id}: you are not the owner of this point."
+                        )
                     existing_group_ids = {g.id for g in phot.groups}
                     for group in groups:
                         if group.id not in existing_group_ids:
--- a/skyportal/handlers/api/sharing.py
+++ b/skyportal/handlers/api/sharing.py
@@ -62,7 +62,9 @@ class SharingHandler(BaseHandler):
 
         with self.Session() as session:
             valid_groups = session.scalars(
-                Group.select(self.associated_user_object).where(Group.id.in_(group_ids))
+                Group.select(session.user_or_token)
+                .where(Group.id.in_(group_ids))
+                .distinct()
             ).all()
             valid_group_ids = [g.id for g in valid_groups]
             invalid_group_ids = [gid for gid in group_ids if gid not in valid_group_ids]
@@ -75,9 +77,8 @@ class SharingHandler(BaseHandler):
             spec_obj_internal_keys = []
 
             if phot_ids:
-                # valid_phot = Photometry.query.filter(Photometry.id.in_(phot_ids))
                 valid_phot = session.scalars(
-                    Photometry.select(self.associated_user_object).where(
+                    Photometry.select(session.user_or_token, mode="update").where(
                         Photometry.id.in_(phot_ids)
                     )
                 ).all()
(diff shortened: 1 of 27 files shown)


skyportal/handlers/api/sharing.py
```diff
@@ -116,37 +117,32 @@ def post(self):
...
                 for spec in valid_spec:
-                    # Ensure user has access to data being shared
-                    if (
-                        spec.owner_id != self.associated_user_object.id
-                        and "System admin"
-                        not in self.associated_user_object.permissions
-                    ):
-                        return self.error(
-                            f"Cannot share spectrum id {spec.id}: you are not the owner of this spectrum."
-                        )
```
Theodlz, line 130: I suppose you removed this because `Spectrum.select(session.user_or_token, mode="update")` should take care of the data access already? If so, worth double-checking the data access in `models` matches this python logic.

static/js/components/gcn/DefaultGcnTagTable.jsx
```diff
@@ -111,13 +93,9 @@ const DefaultGcnTagTable = ({ default_gcn_tags }) => {
   const renderDelete = (dataIndex) => {
     const default_gcn_tag = default_gcn_tags[dataIndex];
     return (
-      <div>
+      <>
```
Theodlz, line 96: is that `<>` even doing anything? Seems like we have a single top level <Button> anyways

static/js/components/gcn/DefaultGcnTagTable.jsx
```diff
@@ -174,19 +152,17 @@ const DefaultGcnTagTable = ({ default_gcn_tags }) => {
...
-          <MUIDataTable
-            data={default_gcn_tags}
-            options={options}
-            columns={columns}
-          />
-        </ThemeProvider>
-      ) : (
-        <CircularProgress />
-      )}
+    <div style={{ width: "100%", minWidth: "40vw", overflow: "scroll" }}>
```
Theodlz, line 158: we prefer having styles defined at the top of components rather than embed styling in the html, but if that's the only `style` in the whole component I agree that defining it at the top is just cumbersome.

static/js/components/spectrum/UploadSpectrum.jsx
```diff
@@ -668,6 +669,17 @@ const UploadSpectrumForm = ({ route }) => {
...
+          <Link
+            to={`/source/${route.id}`}
+            style={{
+              display: "inline-flex",
+              alignItems: "center",
+              gap: "4px",
+              marginBottom: "0.5rem",
+            }}
+          >
+            <ArrowBackIcon fontSize="small" /> Back to source
```
Theodlz, line 681: Nice, I really like that you added these

Theodlz (approved):
LGTM, minor comments (more like questions)

skyportal/handlers/api/sharing.py
antoine-le-calloch: Yep the models data access is matching, but I can add it back if you prefer.

static/js/components/gcn/DefaultGcnTagTable.jsx
antoine-le-calloch: there is 2 div inside:
```
      <>
        <Button
          id="delete_button"
          onClick={() => openDialog(default_gcn_tag.id)}
        >
          <DeleteIcon />
        </Button>
        <ConfirmDeletionDialog
          deleteFunction={deleteDefaultGcnTag}
          dialogOpen={dialogOpen}
          closeDialog={closeDialog}
          resourceName="defaultGcnTag"
        />
      </>
```

skyportal/handlers/api/sharing.py
Theodlz: Nop! Not needed if it matches

static/js/components/gcn/DefaultGcnTagTable.jsx
Theodlz: Ah crap! Missed it on the git diff. Sorry

static/js/components/gcn/DefaultGcnTagTable.jsx
antoine-le-calloch: I prefer using inline when the style is only specific to this div and when the style is use multiple time I use makestyle to avoid duplication.
This style was the only one on all the file.

PR #6106, opened 2026-05-08
Add concurrency settings to cancel superseded CI runs for PRs.

antoine-le-calloch:


Theodlz (approved):
This is great, it will save us some compute and help avoid CI "downtimes" where one user commits a lot and now one else's PR can run their own CI jobs.

One comment: We have a similar mechanism for the documentation CI, but for main. It does: "cancel existing documentation deploy jobs if there is a new one", so basically what you do here except just on main for the docs deploy CI. Issue is that the cancelled runs are not deleted, and github treats them as failures when it comes to showing the ✅ or ❎ icons on the commits. Obviously that issue on main needs to be fixed but that's its own PR, but I guess I'm curious about what these new changes will do to PRs: will cancelled jobs be skipped or will make commits with cancelled jobs just look like failures?

@antoine-le-calloch

antoine-le-calloch:
> This is great, it will save us some compute and help avoid CI "downtimes" where one user commits a lot and now one else's PR can run their own CI jobs.
> 
> One comment: We have a similar mechanism for the documentation CI, but for main. It does: "cancel existing documentation deploy jobs if there is a new one", so basically what you do here except just on main for the docs deploy CI. Issue is that the cancelled runs are not deleted, and github treats them as failures when it comes to showing the ✅ or ❎ icons on the commits. Obviously that issue on main needs to be fixed but that's its own PR, but I guess I'm curious about what these new changes will do to PRs: will cancelled jobs be skipped or will make commits with cancelled jobs just look like failures?
> 
> @antoine-le-calloch

@Theodlz This seems to also be the case here, as we can see in my last two commits. It says 9 cancelled but still displays the red cross. I don't really think we can fix that.
[devactivity.com/insights/decoding-github-actions-why-cancelled-means-failure-in-your-software-project-tool/](devactivity.com/insights/decoding-github-actions-why-cancelled-means-failure-in-your-software-project-tool/)

Theodlz:
> @Theodlz This seems to also be the case here, as we can see in my last two commits. It says 9 cancelled but still displays the red cross. I don't really think we can fix that. [devactivity.com/insights/decoding-github-actions-why-cancelled-means-failure-in-your-software-project-tool/](devactivity.com/insights/decoding-github-actions-why-cancelled-means-failure-in-your-software-project-tool/)

what we tried to do for the docs CI (though we clearly failed, as it doesn't work as expected) is have some auto-delete of cancelled runs. That way, they don't count as failures. But anyways this is still an improvement! Feel free to merge

antoine-le-calloch:
@Theodlz I think I found something that works. The GitHub Action `delete-superseded-runs` needs to be pushed to `main` to work, but I tested it on my repo and it works perfectly fine.

PR #6110, opened 2026-05-13
async handlers

mcoughlin:
This PR implements async handlers.


Theodlz:
is this PR obsolete now that #6127 is open?

thomasculino:
@mcoughlin I see that some of the handlers are still sync. Do we want to go full async or keep some sync? Also, in `data_access.py`, there are some duplicates functions, with some being async, some not, is that meant to be?

mcoughlin:
@thomasculino yeah some of those functions touch eg external faculty APIs and I didn't want to tackle those this first round. We should be able to finish cleaning those up with a bit more careful testing (eg coordinating with folks to try their triggers post a deploy).

thomasculino:
I see, makes sense. In this case, it looks pretty good.

thomasculino (approved):

Issue #6114, opened 2026-05-14
Async DB layer


mcoughlin:
It would be good to migrate the driver from psycopg2 to psycopg v3, set up an async SQLAlchemy engine and async session factories, and move our heaviest endpoints to be async (and eventually all of them).

PR #6127, opened 2026-05-17
psycopg3

mcoughlin:
This PR updates from psycopg2->psycopg3.

### DIFF
--- a/skyportal/__init__.py
+++ b/skyportal/__init__.py
@@ -8,30 +8,33 @@ try:
     if not hasattr(matplotlib, "docstring"):
         matplotlib.docstring = matplotlib._docstring
 
-    # Register numpy types with psycopg2 for numpy 2.x compatibility
-    # numpy 2 scalars no longer inherit from Python builtins, so psycopg2
-    # needs explicit adapters to serialize them in SQL parameters
-    import math
-
+    # Register numpy types with psycopg (v3) for numpy 2.x compatibility.
+    # numpy 2 scalars no longer inherit from Python builtins, so psycopg's
+    # default lookup misses them. Each dumper coerces the numpy scalar to
+    # its native Python type and delegates to psycopg's built-in dumper,
+    # which already handles NaN/Inf correctly.
     import numpy as np
-    from psycopg2.extensions import AsIs, register_adapter
+    import psycopg
+    from psycopg.types.bool import BoolDumper
+    from psycopg.types.numeric import FloatDumper, Int8Dumper
 
-    def _adapt_numpy_float(val):
-        f = float(val)
-        if math.isnan(f):
-            return AsIs("'NaN'::float")
-        if math.isinf(f):
-            return AsIs("'%s'::float" % ("Infinity" if f > 0 else "-Infinity"))
-        return AsIs(repr(f))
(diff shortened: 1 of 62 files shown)


Theodlz:
just one question: this updates a ton of handlers, but not all of them. Might be good to have some TODO list somewhere, or at least an issue that says that not all handlers have been made async (I believe some had `async` in front of the handler get/post/delete but use purely blocking code in them too?). Just worried we open that one big PR and forget about the rest!

mcoughlin:
@Theodlz this is an intermediate PR that just handles the 2->3 transition. Luckily not all handlers needed variable updates. Once it is merged, I will start doing the async part in blocks of handlers at a time.

Theodlz:
> @Theodlz this is an intermediate PR that just handles the 2->3 transition. Luckily not all handlers needed variable updates. Once it is merged, I will start doing the async part in blocks of handlers at a time.

great! I was just suggesting to have an issue open about that, that we can keep alive until everything has been migrated. Just so we don't stop half way and forget about updating the remaining handlers :)

mcoughlin:
@Theodlz Ah yeah. I do have an issue open here: https://github.com/skyportal/skyportal/issues/6114 and can start using it to track progress once I get going.

Theodlz:
> @Theodlz Ah yeah. I do have an issue open here: #6114 and can start using it to track progress once I get going.

ah I missed it, my bad. But that's great. thx!

Theodlz (commented):
This definitely seems to work, just 2 worries of mine, and I was curious to hear what you think can be done about this:
- clearly our test coverage has declined over the years, as we prioritized shipping features and fixing bugs over spending the time to cover all new additions/changes. How did you decide what handles needing to be edited in this PR and which didn't? It is unit test driven? It's the right approach for sure, but given the lacking unit test coverage I do worry that some endpoints may break without us knowing (those not edited by this PR, or some optional features on the handlers you edited but that were left as is).
- there's a TON of boilerplate here, that's copy/pasted everywhere. For in path parameters for example, we often have IDs and I see the same few lines of code that validate them everywhere. I wonder if most of that can't be turned into a single decorator you can add on top of handlers to specify which in path params need to be validated, e.g.
    ```python
    @validate_path_params(
        obj_id=str,
        annotation_id=int,
        optional_id=(int,None) # or something like that to handle optional values and what their default should be
    )
    ```

mcoughlin:
So Claude went handler by handler fixing as needed and then did the unit tests, so I suspect we have the vast majority of them but a couple might remain which we should catch with sentry and I can clean those up quickly.

I like your idea a lot @Theodlz. Can we merge this and open an issue with your idea and assign it to me?

Theodlz:
> So Claude went handler by handler fixing as needed and then did the unit tests, so I suspect we have the vast majority of them but a couple might remain which we should catch with sentry and I can clean those up quickly.
> 
> I like your idea a lot @Theodlz. Can we merge this and open an issue with your idea and assign it to me?

yep! Fine with a follow-up PR

Theodlz (approved):
LGTM. Let's explore the deduplication of parameter validation in a dedicated PR, assigned to @mcoughlin

PR #6128, opened 2026-05-18
MUI v4 → v5 styling migration

mcoughlin:
This PR enables a MUI v4 → v5 styling migration.


Theodlz (approved):
It's one of those PRs again lol.

At first I what playing with the frontend and I thought this created some bugs, because the styling looked a bit off here and there. Then I realized Fritz had that too, meaning these problems are there on main already hahaha.

Thanks a lot for making that migration finally happen (and thank you Claude)!

package.json
```diff
@@ -96,13 +93,15 @@
     "redux-logger": "3.0.6",
     "redux-state-sync": "3.1.4",
     "redux-thunk": "2.4.2",
+    "regenerator-runtime": "0.14.1",
```
Theodlz, line 96: what is this for?

package.json
```diff
@@ -50,7 +48,6 @@
     "imports-loader": "5.0.0",
     "kapsule": "1.16.0",
     "material-ui-phone-number": "3.0.0",
-    "mathjax-full": "3.2.2",
```
Theodlz, line 53: I hope we don't need this in Fritz, I know Thomas' filter UI does a bunch of latex math display. But that's for us to figure out, better remove it from SkyPortal if its not used there!

package.json
mcoughlin: Replacement for depracated @babel/polyfill.

package.json
mcoughlin: We will find out ;)

PR #6129, opened 2026-05-18
remove table lock

mcoughlin:
This PR refactors the photometry dedup from LOCK TABLE to INSERT … ON CONFLICT


alembic/versions/6928ae655672_tablelock.py
```diff
@@ -0,0 +1,32 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(op.f("ix_photstats_obj_id"), table_name="photstats")
+    op.create_index(op.f("ix_photstats_obj_id"), "photstats", ["obj_id"], unique=True)
+    # ### end Alembic commands ###
+
```
Theodlz, line 38: since there is no unique index at present, we could have more than one entry per object. That would completely fail that migration which adds uniqueness. So if we really want to make this a unique obj_id thing, that needs to be handled (e.g. where we have more than 2 entries per obj_id keep the most recent one). I believe there are some scripts or handlers out there in the pipeline that recompute new photstats for an entry -> insert them -> then remove the old ones. We could just do it in transaction and then a unique index is great, but as is this change might break these scripts. Just thinking about this because maybe 2 years ago we added new fields to photstats and I had to recompute photstats for all sources, so I have a vague memory of such a thing. To verify!

skyportal/handlers/api/photometry.py
```diff
@@ -856,6 +857,130 @@ def get_values_table_and_condition(df, ignore_flux=False):
...
 
 
+_PHOTOMETRY_DEDUP_COLS = (
+    "obj_id",
+    "instrument_id",
+    "origin",
+    "mjd",
+    "fluxerr",
+    "flux",
+)
```
Theodlz: should this just be defined in the model, or inferred from it?

skyportal/handlers/api/photometry.py
```diff
@@ -856,6 +857,130 @@ def get_values_table_and_condition(df, ignore_flux=False):
...
+    else:
+        get = lambda k: getattr(row, k)  # noqa: E731
+    return (
+        get("obj_id"),
+        get("instrument_id"),
+        str(get("origin")),
+        float(get("mjd")),
+        float(get("fluxerr")),
+        float(get("flux")),
+    )
```
Theodlz: also feels like yet another copy of that list of columns, and if we could have all that defined in one place it would be less prone to incorrect edits made to the code. So I propose moving those "columns used for deduplication" definition to the model, if we can.

skyportal/handlers/api/photometry.py
```diff
@@ -927,15 +1026,31 @@ def insert_new_photometry_data(
...
+        # as JSON literals on the wire. Also normalize NaN→None (pandas leaks
+        # NaN for missing cells in object columns, and JSONB rejects NaN).
+        altdata_val = packet.get("altdata")
+        if isinstance(altdata_val, float) and np.isnan(altdata_val):
+            altdata_val = None
         phot = {
-            "id": packet["id"],
-            "original_user_data": json.dumps(original_user_data, cls=NumpyEncoder),
+            "original_user_data": (
+                json.loads(json.dumps(original_user_data, cls=NumpyEncoder))
```
Theodlz: so I see that before we had to dump the data. Now that we don't, do we really need/want to do a dump+load? Is that to sanitize the data somehow? if we can avoid the serialize+deserialize that would be better.

Theodlz (commented):
some minor comments and questions

alembic/versions/6928ae655672_tablelock.py
mcoughlin: I added the potential dedup to the migration. I did check fritz had no duplicates, and it did not, so seemed safe.

skyportal/handlers/api/photometry.py
mcoughlin: I now have a Photometry.DEDUP_COLUMNS for this.

skyportal/handlers/api/photometry.py
mcoughlin: Yeah see above.

skyportal/handlers/api/photometry.py
mcoughlin: ok we now use numpy_to_native.

alembic/versions/6928ae655672_tablelock.py
Theodlz: Right, I think it's great to handle it in migrations because we lack visibility on the other instances and we wouldn't want to break those.

Theodlz (commented):
LGTM! Did you have a chance to check that the features affected here are well covered by unit tests? Given how important handling photometry is (and how destructive it could get if it goes wrong), might be worth triple checking the unit test coverage of the related handlers.

Theodlz (approved):
LGTM! Did you have a chance to check that the features affected here are well covered by unit tests? Given how important handling photometry is (and how destructive it could get if it goes wrong), might be worth triple checking the unit test coverage of the related handlers.

mcoughlin:
@Theodlz yes I had a script shooting a lot of photometry through the handlers to be sure it behaved correctly, which definitely helped try to check for corner cases. I added a few tests to try to improve coverage a bit as well.

PR #6130, opened 2026-05-18
[WIP] multiple microservices

mcoughlin:
This PR makes it possible to run multiple instances of microservices (to enable scaling).

### DIFF
--- a/services/facility_queue/facility_queue.py
+++ b/services/facility_queue/facility_queue.py
@@ -99,13 +99,17 @@ def service():
             time.sleep(5)
             with DBSession() as session:
                 try:
+                    # FOR UPDATE SKIP LOCKED: concurrent replicas of this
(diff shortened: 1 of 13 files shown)


stefanv:
The PR seems to modify the microservices themselves, but does not run multiple instances—or perhaps I'm missing where that is happening? Could you explain the mechanism you are using, and the implications on the DB that are being addressed?

mcoughlin:
@stefanv My first goal is to figure out if/how one could run multiple instances of the various microservices safely. The queues rely on some kind of status (some of them would need it added) so that can be tracked. The periodic ones (reminders, observation plans) are weirder for this. I would probably focus on getting the busiest ones fully moved over. We know we will want multiple copies of the Kafka reader between BOOM/SkyPortal for example.

PR #6131, opened 2026-05-19
slimmer photometry for front-end

mcoughlin:
This PR makes a slim photometry serializer. The front-end did not return all of the columns, but they were still being queried in memory.

### DIFF
--- a/static/js/ducks/photometry_minimal.js
+++ b/static/js/ducks/photometry_minimal.js
@@ -12,16 +12,9 @@ export function fetchSourcePhotometryMinimal(id) {
     `/api/sources/${id}/photometry`,
     FETCH_SOURCE_PHOTOMETRY_MINIMAL,
     {
-      format: "mag",
+      format: "plot",
       magsys: "ab",
       individualOrSeries: "both",
-      includeOwnerInfo: false,
-      includeStreamInfo: false,
-      includeValidationInfo: false,
-      includeAnnotationInfo: false,
-      includeExtinctionInfo: false,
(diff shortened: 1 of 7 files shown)


static/js/ducks/photometry_minimal.js
```diff
@@ -12,16 +12,9 @@ export function fetchSourcePhotometryMinimal(id) {
     `/api/sources/${id}/photometry`,
     FETCH_SOURCE_PHOTOMETRY_MINIMAL,
     {
-      format: "mag",
+      format: "plot",
```
Theodlz, line 15: One might argue that "plot" being a format option like "mag" and "flux" feels a little confusing, but since "plot" is basically a subset of the "flux" format, why not!

Theodlz (commented):
Nice, just curious: did you manage to see a real performance gain (in speed and/or mem) by not loading the extra columns the frontend doesn't need? Are there many of them left when we already set all the `include` fields to False?

mcoughlin:
@Theodlz No this was more an idea from seeing some large calls in the nginx logs. It does get rid of automatically pulling in groups and instruments from the old call.

Theodlz:
> @Theodlz No this was more an idea from seeing some large calls in the nginx logs. It does get rid of automatically pulling in groups and instruments from the old call.

ohhh right I see, I didn't have the groups+instruments in mind, we definitely don't use those for the lightweight vega plots. So that's a great idea!

Theodlz (approved):
LGTM!

PR #6133, opened 2026-05-20
n+1 for user query

mcoughlin:
This PR fixes an n+1 user query issue.


Theodlz:
makes me wonder if we shouldn't just use `selectinload` everywhere rather than our current `joinedload` default behavior, which duplicates rows since it does a JOIN and that forces us to use `distinct()` everywhere and deal with N+1 issues. Something to consider (and check if there are no downsides)...

Theodlz (approved):
LGTM

Theodlz (approved):
LGTM

PR #6136, opened 2026-05-21
validate path params with decorator

### LINKED ISSUE #6135: decorator to validate path parameters
The psycopg2 -> 3 transition noted that throughout the app, there is a lot of lines like:

        try:
            filter_id = int(filter_id)
        except (TypeError, ValueError):
            return self.error(f"Invalid filter_id: {filter_id}")

that's copy/pasted everywhere. For in path parameters for example, we often have IDs and I see the same few lines of code that validate them everywhere. @Theodlz noted that maybe most of that can be turned into a single decorator you can add on top of handlers to specify which in path params need to be validated, e.g.
@validate_path_params(
    obj_id=str,
    annotation_id=int,
    optional_id=(int,None) # or something like that to handle optional values and what their default should be
)

mcoughlin:
This PR adds decorators for API path validation.

Closes https://github.com/skyportal/skyportal/issues/6135


skyportal/tests/utils/test_handlers_base_init_subclass.py
Theodlz, line 1: feels a little bit like this copy of the basehandler methods (like the _resolve_cast, _HANDLER_METHODS, ...) is a shame. Why not just import these here?

skyportal/tests/utils/test_handlers_base_init_subclass.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def test_invalid_int_returns_error():
+    class H(FakeBase):
+        def get(self, filter_id: int):
+            return filter_id
+
+    h = H()
+    result = _run(h.get("not-a-number"))
+    assert result == ("error", "Invalid filter_id: not-a-number")
```
Theodlz: lot's of these tests look very much alike. For better test coverage (of all dtypes) without writing a lot of distinct tests, maybe `@pytest.mark.parametrize` can come in handy?

skyportal/handlers/api/analysis.py
```diff
@@ -1017,7 +1012,7 @@ def delete(self, analysis_service_id):
 
 class AnalysisHandler(BaseHandler):
     @permissions(["Run Analyses"])
-    async def post(self, analysis_resource_type, resource_id, analysis_service_id):
+    async def post(self, analysis_resource_type, resource_id, analysis_service_id: int):
```
Theodlz: I noticed on a couple of handlers, that not all parameters get type hinted. Curious if there was a specific reason, other than that this PR focuses on adding type hints to the ID integers that really need validation for psycopg3? Anyways, just though that we could benefit from typehinting more stuff (even if psycopg3 doesn't care) to have stronger validation and type hinted variables for the IDE. That's a very minor comment though, can be addressed later.

Theodlz (approved):
LGTM! That's a really elegant solution, very happy with how it came out! I wonder if there is a cost performance wise (I mean there always is, but I wonder just how costly this is), my guess is that its negligible though. Plus if I understood how it works properly, it's a wrapper that runs at initialization time of the handlers to wrap the GET/POST/... methods with a cheap validation step, so the cost at API call time should be very very small, not too different from explicitly validating the fields in each handler.

I have a few minor comments/questions.

skyportal/tests/utils/test_handlers_base_init_subclass.py
mcoughlin: Good point, fixed

skyportal/tests/utils/test_handlers_base_init_subclass.py
mcoughlin: Yup!

skyportal/handlers/api/analysis.py
mcoughlin: Yup got more of these.

Theodlz (approved):
This looks great, awesome work @mcoughlin ! Please feel free to merge once green

PR #6137, opened 2026-05-22
add NEXP column to NGPS starlist

Theodlz:
As the title says :). Looks like whoever maintains the NGPS system added a new column and it broke stuff for us. I just wish they would coordinate a bit with us, since clearly a lot of the folks using NGPS create their starlist via Fritz...


mcoughlin (approved):
LGTM

mcoughlin:
@Theodlz We merge this?

Theodlz:
@mcoughlin would be good to merge and deploy this ASAP, so folks can use star lists from Fritz to trigger NGPS again (it's broken right now)

PR #6141, opened 2026-05-22
mitigate warnings (mostly datetime timezones)

mcoughlin:
This PR mitigates the various current Python warnings.


antoine-le-calloch (approved):
LGTM. I made some fix + clean up

PR #6144, opened 2026-05-29
test concurrent photometry writes

### LINKED ISSUE #6117: Photometry-table lock-removal test
We lock the table when writing photometry, originally added to prevent a class of race conditions during concurrent bulk-load and API-write. We should build a stress test that drives concurrent writers against a test DB, verify correctness invariants with and without the lock, and report, to see if this is still needed.

mcoughlin:
This PR tests concurrent writes to the photometry database in prep for the async capabilities.

Closes https://github.com/skyportal/skyportal/issues/6117


antoine-le-calloch:
@mcoughlin The timeout you added to the flaky test doesn't fix the issue. The problem occurs when the current day is greater than the maximum number of days in February. When the test updates the calendar, it first tries to set the month to 02. However, if the current date is, for example, the 29th, 30th, or 31st, the browser rejects the change because February does not contain that day. As a result, the input is not updated correctly, and the test later fails when trying to find the expected entry. This only happens when the current day is above 28. I ran into the exact same issue in your other PR and fixed it here:

https://github.com/skyportal/skyportal/pull/6141/changes/b025202b9ba093c653854cc30018c12fb98a1224

antoine-le-calloch:
When all the tests have passed, I will merge https://github.com/skyportal/skyportal/pull/6141. You will then be able to merge it into your PR to fix this issue.

mcoughlin:
Thanks @antoine-le-calloch !

antoine-le-calloch (approved):
LGTM

PR #6147, opened 2026-05-30
configurable micro-service IPs

mcoughlin:
This PR make micro-service IPs configurable (needed for any kind of multi-host / distributed deployment).

Based on https://github.com/skyportal/skyportal/pull/4628


stefanv (commented):
This is how I would have started too, and then we can use Kubernetes DNS to find the hosts we need.

The `notification_queue` is arguably the most important one. It may be worth renaming that one to `message_bus` in a follow-up PR to make it more obvious.

Another related refactoring is to ensure a different start-up part for app zero than for the others (initialize the enums etc.). In fact, it may make even more sense to factor out initialization to a separate micro-service, so that we do not keep having apps compete for that operation, and also don't keep redoing it.

stefanv (approved):

antoine-le-calloch (approved):

antoine-le-calloch:
@mcoughlin I made some quick fix. You can review my latest changes and merge if they look good to you.

PR #6148, opened 2026-05-30
pnpm

mcoughlin:
This PR moves to use pnpm.

Closes the spirit of https://github.com/skyportal/skyportal/pull/5314


stefanv:
Theo and I experimented with pnpm and both landed in favor of bun. Curious whether you've thought about that option.

PR #6152, opened 2026-05-31
sharing services n+1 query

mcoughlin:
This PR fixes an n+1 query identified by Sentry in the sharing services.


mcoughlin:
Trivial, self-merging.

PR #6154, opened 2026-05-31
original spectrum file must be explicitly queried for

mcoughlin:
This PR makes it such that the user must opt into returning the original spectrum file (as these queries correspond to some of our largest in the app) with includeOriginalFile=True.


mcoughlin:
I do not think this PR is controversial for developers, and we will manage user complaints if/as they come.

PR #6155, opened 2026-05-31
auto-trigger criteria

mcoughlin:
This PR prepares to move auto-triggering criteria from Kowalski into SkyPortal.

### DIFF
--- a/skyportal/handlers/api/followup_request.py
+++ b/skyportal/handlers/api/followup_request.py
@@ -35,7 +35,7 @@ from marshmallow.exceptions import ValidationError
 from scipy.stats import norm
 from sqlalchemy import func
 from sqlalchemy.dialects.postgresql import JSONB
-from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import joinedload, selectinload
 from sqlalchemy.sql.expression import cast
 from tornado.ioloop import IOLoop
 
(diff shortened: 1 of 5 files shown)


skyportal/handlers/api/followup_request.py
```diff
@@ -761,45 +894,182 @@ def post_default_followup_requests(obj_id, default_followup_requests, user_id):
...
+                    sa.select(FollowupRequest).where(
                         FollowupRequest.obj_id == obj_id,
                         FollowupRequest.allocation_id == allocation_id,
                         FollowupRequest.status != "deleted",
                     )
-                ).first()
-                if existing_request is not None:
-                    log(
-                        f"Skipping default followup request for {obj_id} with allocation ID {allocation_id} because one already exists."
+                ).all()
```
thomasculino, line 955: Doesn't this trigger a N+1 query when accessing `r.target_groups` (L954) in `_same_request`?

thomasculino (commented):
One comment about a N+1 query otherwise it looks good. We might want to add some tests someday but it's not blocking here.

skyportal/handlers/api/followup_request.py
mcoughlin: Good point. Fixed.

thomasculino (approved):
LGTM!

PR #6176, opened 2026-06-02
More typescript flags

mcoughlin:
This PR adds a few more typescript flags.


stefanv (approved):

stefanv (commented):
Lots of anys to get rid of in follow-ups :)

PR #6180, opened 2026-06-03
retry the sncosmo warm start

mcoughlin:
This PR retries the SVO caching.

### DIFF
--- a/.github/workflows/test_api_and_frontend.yaml
+++ b/.github/workflows/test_api_and_frontend.yaml
@@ -153,6 +153,19 @@ jobs:
         run: |
           sed -i 's/duration = 3000/duration = 15000/g' baselayer/static/js/components/Notifications.jsx
 
+      - name: Pre-warm sncosmo bandpass cache
+        # cache-dependencies restores ~/.astropy/cache/sncosmo from a prior
+        # run, but on a cold key (first run after a uv.lock bump, or after
+        # cache eviction) the app would otherwise discover missing
+        # bandpasses lazily at test time — each one blocks ~10s on
+        # astropy's urlopen and the cumulative wait pushes startup past
+        # test_frontend.py's window. Run the warm-up here so missing
+        # bandpasses are pulled in once, up front. Cached ones return
+        # instantly, so a warm cache hit is near free; failures are
+        # tolerated (script retries internally and `|| true` here).
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/test_api_and_frontend.yaml
```diff
@@ -153,6 +153,19 @@ jobs:
...
+      - name: Pre-warm sncosmo bandpass cache
+        # cache-dependencies restores ~/.astropy/cache/sncosmo from a prior
+        # run, but on a cold key (first run after a uv.lock bump, or after
+        # cache eviction) the app would otherwise discover missing
+        # bandpasses lazily at test time — each one blocks ~10s on
+        # astropy's urlopen and the cumulative wait pushes startup past
+        # test_frontend.py's window. Run the warm-up here so missing
+        # bandpasses are pulled in once, up front. Cached ones return
+        # instantly, so a warm cache hit is near free; failures are
+        # tolerated (script retries internally and `|| true` here).
```
stefanv, line 165: Man, the AIs are chatty.

stefanv: Should we commit the cache immediate after this step, so it gets saved even if tests fail?

tools/warm_sncosmo_cache.py
```diff
@@ -18,26 +18,42 @@
...
-    print(f"sncosmo warm-up done: {ok} cached, {fail} skipped")
+    all_names = [m["name"] for m in metadata if m.get("name")]
+    total = len(all_names)
+
+    # The Spanish Virtual Observatory host (svo2.cab.inta-csic.es) that hosts
+    # several bandpass families (skymapper, gaia, etc.) is flaky enough that
+    # a single sweep usually leaves ~20 of 183 bandpasses missing — and the
+    # set of failures rotates between runs. Retry a few times so the slower
+    # bandpasses get another shot. Each pass only re-tries the ones that
+    # are still missing; cached ones return instantly.
```
stefanv, line 47: Is there a way to get this behind a CDN? How often do the bandpasses change? I.e., are we reliant on them as the source?

mcoughlin: We could put them in the repo as really they dont change, just the number of filters we pull in increases (slowly in time).

PR #6181, opened 2026-06-03
Fix gcnevent dateobs type

thomasculino:
The `dateobs` FK column in `SourcesConfirmedInGCN` had no explicit type, causing
SQLAlchemy to return the value as a string. When those strings were passed to
`GcnEvent.dateobs.in_()`, PostgreSQL rejected the query with:

  operator does not exist: timestamp without time zone = character varying

Adding `sa.DateTime` fixes the deserialization so the ORM returns proper
`datetime` objects, matching the `timestamp` column type on `gcnevents.dateobs`.


mcoughlin:
Needs a migration? @thomasculino

thomasculino:
> Needs a migration? @thomasculino

I'm unsure. It just makes the type explicit for the column. Do we need a migration for that?

mcoughlin:
Looks like migrations pass so I guess not!

mcoughlin (approved):
Merge on green!

stefanv:
Quick review on that question from Claude:

```
 On the migration question: No migration is needed. The database column type is already timestamp (PostgreSQL infers it from the FK target gcnevents.dateobs). Adding sa.DateTime only makes the ORM type explicit — the schema doesn't change.

  The bigger issue — the fix is incomplete. There are 8 more dateobs FK columns missing sa.DateTime across the codebase:

  ┌─────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────┐
  │            File             │                                        Lines                                        │
  ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
  │ gcn.py                      │ GcnReport (141), GcnSummary (376), GcnNotice (445), GcnProperty (491), GcnTag (521) │
  ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
  │ observation_plan.py         │ 289, 401                                                                            │
  ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
  │ localization.py             │ 73                                                                                  │
  ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
  │ sources_confirmed_in_gcn.py │ 39 ← fixed by PR, but not yet in local repo                                         │
  └─────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────┘

  All of these are susceptible to the same string-deserialization bug.
```

thomasculino:
> Quick review on that question from Claude:
> 
> ```
>  On the migration question: No migration is needed. The database column type is already timestamp (PostgreSQL infers it from the FK target gcnevents.dateobs). Adding sa.DateTime only makes the ORM type explicit — the schema doesn't change.
> 
>   The bigger issue — the fix is incomplete. There are 8 more dateobs FK columns missing sa.DateTime across the codebase:
> 
>   ┌─────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────┐
>   │            File             │                                        Lines                                        │
>   ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
>   │ gcn.py                      │ GcnReport (141), GcnSummary (376), GcnNotice (445), GcnProperty (491), GcnTag (521) │
>   ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
>   │ observation_plan.py         │ 289, 401                                                                            │
>   ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
>   │ localization.py             │ 73                                                                                  │
>   ├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
>   │ sources_confirmed_in_gcn.py │ 39 ← fixed by PR, but not yet in local repo                                         │
>   └─────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────┘
> 
>   All of these are susceptible to the same string-deserialization bug.
> ```

That's what I thought, I just didn't know all of them. Let me add those in another PR

PR #6186, opened 2026-06-03
Fix delete superseded github action

antoine-le-calloch:
Fix `gh api` call by use explicitly GET method for fetching latest GitHub action run number. To fix `gh api` from using POST method when arguments are passed.


mcoughlin (approved):
LGTM

PR #6187, opened 2026-06-04
Fix gcnevent dateobs type

thomasculino:
Fix dateobs FK columns missing sa.DateTime


mcoughlin (approved):
Thanks for following up!

PR #6189, opened 2026-06-04
harden Selenium click helpers against layout-shift flakes

mcoughlin:
This PR tries to harden Selenium click helpers against layout-shift flakes.


stefanv (commented):
Worth a try. This stuff is super finicky. I think it will be better with playwright.

mcoughlin:
Good point @stefanv . Let's just do that.

PR #6190, opened 2026-06-05
Fix gitignore to work with typescript

antoine-le-calloch:
- Update template file extensions from .jsx to .tsx.
- Remove the .tsx generated files from GitHub since they should not be pushed.


antoine-le-calloch:
@mcoughlin Be careful not to push the generated files because they are specific to your configuration.

mcoughlin (approved):
Thanks for cleaning that up!

PR #6191, opened 2026-06-05
switch to bun as the JS package manager

mcoughlin:
This PR switches to bun as the JS package manager.

Compare to https://github.com/skyportal/skyportal/pull/6148


mcoughlin:
With votes from Stefan and Theo, I think it is fine to go with bun.

PR #6193, opened 2026-06-05
generate TypeScript types from OpenAPI spec

mcoughlin:
This PR generates TypeScript types from OpenAPI spec.


mcoughlin:
Mostly fixes some broken API specs and lays ground work for comparisons with marshmallow. Merging.

PR #6194, opened 2026-06-05
skyportal-data repo

mcoughlin:
This PR adds a skyportal-data repo to keep the external files that can be flaky.


mcoughlin:
As discussed on the SkyPortal call, this should help with the flakiness. Let's see.

PR #6195, opened 2026-06-05
Fix gitignore and template files

antoine-le-calloch:
Update .gitignore to include TypeScript template files, remove generated templates and delete unused dependencies from uv.lock.


mcoughlin (approved):
LGTM

PR #6198, opened 2026-06-05
docstring consistency with schema

mcoughlin:
This PR improves more docstrings to be consistent with schema.


stefanv:
Thanks for aligning these, Michael.

Can you explain the build route schema map script a bit?

mcoughlin:
@stefanv Yeah the idea is to read openapi.json and create static/js/types/routeSchemaMap.ts, which is a lookup table that RTK-Query uses (which will actually tie our API spec to our ducks).

stefanv:
This is currently non-existent, right? Just checking I didn't miss that development.

mcoughlin:
@stefanv Not yet, it's coming.

PR #6200, opened 2026-06-05
Fix deprecated input field props

antoine-le-calloch:
- Refactor input properties by using slotProps to replace deprecated InputProps.
- Refactor FilterCandidateList to improve state management and replace deprecated inputProps with slotProps.
- Cleanup.


mcoughlin (approved):
LGTM

PR #6201, opened 2026-06-06
playwright

mcoughlin:
This PR moves tests to playwright (which @stefanv thinks will improve on flakiness).


stefanv:
Please also see the patterns in the baselayer template app for reference.

PR #6203, opened 2026-06-06
agents.md

mcoughlin:
This PR adds a brief AGENTS.md file to SkyPortal.


stefanv (approved):

stefanv:
When exploring this for skimage, we needed to symlink this documents to three locations to make most CLIs pick up on it.

Which one are you using? Worth making sure it is covered, especially as we expand the instructions over time.

PR #6207, opened 2026-06-06
symlinks for Claude Code, Cursor, Copilot

mcoughlin:
This PR adds symlinks for Claude Code, Cursor, and Copilot (based on what scikit-image is doing).


stefanv (commented):
Added GEMINI.

PR #6210, opened 2026-06-07
Reduce bundle size

mcoughlin:
This PR trims our initial bundle from 14MB to ~7MB by requiring ES2020 and using vendors: initial (so that vendored packages only load when needed from a page).

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "openapi-typescript": "7.13.0",
     "typescript": "5.7.3",
     "typescript-eslint": "8.60.1",
-    "webpack-bundle-analyzer": "^5.3.0"
+    "webpack-bundle-analyzer": "5.3.0"
   },
   "overrides": {
     "yargs": "^17.7.2",
--- a/package.json
+++ b/package.json
@@ -138,7 +138,8 @@
     "json-loader": "0.5.7",
     "openapi-typescript": "7.13.0",
     "typescript": "5.7.3",
-    "typescript-eslint": "8.60.1"
+    "typescript-eslint": "8.60.1",
+    "webpack-bundle-analyzer": "^5.3.0"
   },
   "overrides": {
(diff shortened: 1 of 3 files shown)


leonitousconforti (approved):
Cool!

package.json
```diff
@@ -138,7 +138,8 @@
     "json-loader": "0.5.7",
     "openapi-typescript": "7.13.0",
     "typescript": "5.7.3",
-    "typescript-eslint": "8.60.1"
+    "typescript-eslint": "8.60.1",
+    "webpack-bundle-analyzer": "5.3.0"
```
stefanv, line 142: This should probably not be included unconditionally.

rspack.config.js
```diff
@@ -24,16 +20,23 @@ const config = (env, argv) => {
       splitChunks: {
         chunks: "all",
         cacheGroups: {
+          // `chunks: "initial"` keeps the big shared vendors chunk to only the
+          // libs reached on first paint. Modules used solely by lazy routes
```
stefanv, line 24: What are "lazy routes"?

stefanv (commented):
Great, thank you!

package.json
mcoughlin: It is a dev dependency, but I would be cool with commenting it out if we prefer. We could also do something like:

if (process.env.ANALYZE) {
  const { BundleAnalyzerPlugin } = require("webpack-bundle-analyzer");
  plugins.push(new BundleAnalyzerPlugin());
}
which is commented out now.

rspack.config.js
mcoughlin: The places we do: React.lazy(() => import(...)) will now only trigger loading that part of the bundle when triggered.

stefanv: Got it.

PR #6216, opened 2026-06-08
Extract unrealted changes from rtk branch

antoine-le-calloch:
Isolates the RTK-independent test & CI-infra changes from the rtk branch https://github.com/skyportal/skyportal/pull/6209

- Build the test frontend bundle in production mode
- PLAYWRIGHT_DEFAULT_TIMEOUT_MS 60s -> 90s, nginx proxy_read_timeout 60->120s

PR #6209, opened 2026-06-07
RTK Query

mcoughlin:
This PR continues the RTK Query port (the easy APIs are now done).

PR #6217, opened 2026-06-08
[WIP] Fix tests

antoine-le-calloch:
## Fix flaky/broken tests + underlying code

Fixes broken or flaky tests, along with the underlying code they exposed.

### Code
- Set Firefox launch options / pointer capabilities so MUI stops switching to mobile mode and breaking some tests (the calendar components were the main culprit).
- Fix the Widget dialog.
- Fix notification settings state updates, which were stale or triggering re-render loops.
- Guard against `undefined` values in `NewRecurringAPI` and `NewTokenForm` so they no longer crash when their data isn't loaded yet.
- Improve weather handling: stop returning an error to the frontend on every homepage load, which happened because the weather endpoint queried a hard-coded telescope id that didn't exist.
- Increase the default pagination values on some components to better fit the new, more compact tables.
- Fix forms and their `useEffect`s, plus general cleanup of the surrounding code.
- Refactor candidate re-search to always refetch, and force `seconds = 0` since the frontend can't select seconds, avoiding inconsistent behavior.

### Tests
- Improve date selection in tests.
- Clean up the test files.
- Isolate some tests so they no longer interfere with each other.
- Remove redundant / duplicate tests.


static/js/components/candidate/FilterCandidateList.tsx
```diff
@@ -542,12 +543,13 @@ const FilterCandidateList = ({
     // Save form-specific data, formatted for the API query
     dispatch(setFilterFormData(data));
 
-    // Trigger a new search (resets to page 1). The query result drives the
```
stefanv, line 545: Comment removed?

static/js/components/candidate/FilterCandidateList.tsx
```diff
@@ -542,12 +543,13 @@ const FilterCandidateList = ({
...
-    // loading state in the parent CandidateList.
+    const nextSearchNonce = searchNonce + 1;
+    setSearchNonce(nextSearchNonce);
```
stefanv: Nonce is a very technical term. Use something more understandable?

stefanv (commented):
Let's see if this helps! No problem with this PR structurally.

static/js/components/candidate/FilterCandidateList.tsx
```diff
@@ -515,7 +516,7 @@ const FilterCandidateList = ({
       data.sortByAnnotationOrder = formData.sortingOrder;
     } else {
       // Clear annotation sort params, if a default sort is not defined
-      await dispatch(setCandidatesAnnotationSortOptions(null));
```
stefanv, line 518: Why are we no longer awaiting the dispatch before moving on?

static/js/components/candidate/FilterCandidateList.tsx
antoine-le-calloch: This change was taken from https://github.com/skyportal/skyportal/pull/6209 but I can update this yes.

antoine-le-calloch:
> Let's see if this helps! No problem with this PR structurally.

This is a Work in Progress, I'm just starting and I have still a lot to fix there.

static/js/components/candidate/FilterCandidateList.tsx
antoine-le-calloch: Because setCandidatesAnnotationSortOptions is not async.

stefanv:
Sorry, missed the WIP tag :)

static/js/components/candidate/FilterCandidateList.tsx
stefanv: Ah, I thought I checked that! Thanks.

antoine-le-calloch:
@mcoughlin This PR should be merged first. Then I will update both https://github.com/skyportal/skyportal/pull/6216 and https://github.com/skyportal/skyportal/pull/6209.

This PR fixes a lot of bugs in the code and plenty of the tests. All the remaining failing tests are GCN-related that use external resources.

mcoughlin (approved):
Good progress, thanks!

PR #6218, opened 2026-06-09
Photometry editing permissions

mcoughlin:
This PR makes for clearer photometry permissions, including edit/delete icon access.


camilledzt (approved):
LGTM, thank you for this changes ! I think that it will help a lot ICARE users to see what points they are able to edit and delete

PR #6145, opened 2026-05-29
Remove depracated mui-datatables

mcoughlin:
This PR replaces mui-datatables with x-data-grid.

PR #6219, opened 2026-06-09
Add export feature to grid toolbars across multiple components

thomasculino:
Export used to be a default feature in mui-datatables which got deprecated. They got replaced by data grids in #6145 which do not have the export as a default feature. This PR aims to explicitly display the export button which got lost on the way.


mcoughlin (approved):
I am happy to see this go in now. The other PR does something similar.

thomasculino:
Actually I just looked at the other PR and I'm not sure if this should go in as the other one takes care of that already. Should I close and wait for the other one to go in?

PR #6227, opened 2026-06-15
Fix some sentry backend errors

antoine-le-calloch:
- Add instrumentID conversion and add error handling.
- clean up


skyportal/handlers/api/classification.py
```diff
@@ -558,6 +558,11 @@ async def delete(self, classification_id: int):
               application/json:
                 schema: Success
         """
+        try:
```
mcoughlin, line 561: This should be caught by the type checker, right?

mcoughlin (approved):
Just checking on one addition.

skyportal/handlers/api/classification.py
antoine-le-calloch: Not really, the int annotation is decorative. Tornado passes URL path args as str (or None when the optional (/[0-9]+)? group isn't matched), so the type checker never sees it.

mcoughlin: Ok!

Issue #6121, opened 2026-05-14
Lazy imports startup-time study


mcoughlin:
A chunk of cold-start time is the import graph itself as many service entry points pull in numpy, pandas, matplotlib, astropy, and SQLAlchemy even when those modules aren't used on the hot path. Python 3.13's importlib.util.LazyLoader could be used to identify the heaviest import trees and migrate the top three-to-five to lazy loading.

PR #6228, opened 2026-06-15
OpenTelemetry metrics endpoint

mcoughlin:
This PR adds an (optional) OpenTelemetry metrics endpoint.


skyportal/handlers/api/metrics.py
```diff
@@ -0,0 +1,38 @@
...
+        responses:
+          200:
+            content:
+              text/plain:
+                schema:
+                  type: string
+        """
+        # Imported lazily so the module loads even without the optional
+        # prometheus-client dependency installed.
+        from prometheus_client import CONTENT_TYPE_LATEST, generate_latest
```
leonitousconforti, line 35: Maybe we could use <https://docs.python.org/3.15/reference/simple_stmts.html#lazy-imports> once python3.15 is an option? Not sure if we would be interested in that though

mcoughlin: Definitely! https://github.com/skyportal/skyportal/issues/6121

skyportal/tests/api_tests/spectra_misc/test_observability.py
```diff
@@ -0,0 +1,76 @@
...
+        # Prometheus text exposition that MetricsHandler serves.
+        from opentelemetry import metrics
+        from prometheus_client import generate_latest
+
+        metrics.get_meter("test").create_counter("unit_demo").add(
+            2, {"route": "/api/sources"}
+        )
+        out = generate_latest().decode()
+        assert "unit_demo_total" in out
+        assert 'route="/api/sources"' in out
```
leonitousconforti: Is this testing skyportal metrics endpoint? It seems like it is testing the prometheus_client library to me which feels like not our concern / not something we should be testing specifically?

skyportal/handlers/api/metrics.py
leonitousconforti: Nice, didn't see that yet!

skyportal/tests/api_tests/spectra_misc/test_observability.py
mcoughlin: good point. I improved this.

leonitousconforti (approved):

PR #6233, opened 2026-06-16
asyncify facilities

mcoughlin:
This PR makes facility interactions async.


thomasculino (approved):
LGTM! Do we want to add more tests in the future? Not a blocker, just a question I had.

mcoughlin:
@thomasculino ideally yes. Early on we were a bit better about using vcrpy and such to capture facility responses and record them, but we got away from this a bit. It will come up again for example in the broker interactions PR.

PR #6238, opened 2026-06-20
remove notedown

mcoughlin:
This PR removes notedown (pulling in the depracated bleach).


thomasculino (approved):
We might want to remove it from baselayer as well. Otherwise LGTM!

PR #6245, opened 2026-06-22
Enh: resize table columns width + add t-t0 to the download options

camilledzt:
The goal of the PR is to:
* add the t-t0 column to the option available in the download form on the photometry table
* update the columns width of the share data page to be able to see the last column correctly (see screenshot)
* fix the "select all" button that was not working anymore
* change the rounding level of the column values (requested by @santier14)
  * limiting mag: 2 digits
  * mag and mag err: 4 digits
  * mjd: 3 digits

image


mcoughlin (approved):
thanks for fixing all that @camilledzt !

Issue #6247, opened 2026-06-23
ZTF-ref acquisition offsets drop offset-star proper motion (epoch-mismatched, breaks for high-PM stars)


profjsb:
**Describe the bug**

When ZTF reference positions are used for acquisition offset stars (the default path for ZTF targets), the offset star positions are **not** proper-motion–corrected to the current epoch. The resulting offset is computed as `star @ ~ZTF-ref-epoch (~2018) → target @ now`, silently dropping the star's proper motion over the intervening ~8 years.

This was confirmed in the code in `skyportal/utils/offset.py`.

**Where it happens**

In `get_nearby_offset_stars` there are two branches for computing an offset star's position:

- **Gaia branch** (`offset.py:1049-1053`, `use_original=True`): builds the star `SkyCoord` with PM at the Gaia `ref_epoch` and calls `c.apply_space_motion(new_obstime=source_obstime)`. Both star and target end up at the same (current) epoch in the same ICRS frame, so the differential offset is correct and PM-complete.

- **ZTF-ref branch** (`offset.py:1013-1041`, the default for ZTF targets): when a ZTF reference catalog match is found within 0.5″, it builds `cprime` directly from the ZTF catalog RA/Dec but stamps it with `obstime=source_obstime` (now) and applies **no** proper motion:

  ```python
  cprime = SkyCoord(
      ra=ztfcatalog[idx].ra.value,
      dec=ztfcatalog[idx].dec.value,
      unit=(u.degree, u.degree),
      frame="icrs",
      obstime=source_obstime,   # labeled "now", but the position is the ~2018 ZTF-ref position
  )
  dra, ddec = cprime.spherical_offsets_to(center)
  ```

  The ZTF position is at the ZTF reference epoch (~2018) but is treated as if it were the current position, while the target `center` is at the current epoch. This is an epoch-mismatched differential offset with the star's ~8 yr of proper motion dropped.

**Are the offsets the current offset?**

- Gaia branch: yes — star and target both at the current epoch, same ICRS frame → differential offset is current and correct.
- ZTF-ref branch (default): no, not exactly — offset is `star @ ~2018` → `target @ now`, missing the star's PM from the ZTF ref epoch to tonight (~PM × 8 yr).
  - Typical field star (PM 5–15 mas/yr): ~0.04–0.12″ — negligible in a 1–1.5″ slit.
  - High-PM star (≥100 mas/yr): ≳0.8″ — enough to push the target partly/fully off a narrow slit.

[...]

PR #6248, opened 2026-06-23
High-PM ZTF-ref offset stars: fall back to PM-corrected Gaia position (#6247, complements #6243)

profjsb:
## Summary

Addresses the high-proper-motion case of #6247, as a **complement to #6243** (rather than an overlapping alternative).

When `use_ztfref` is set, an offset star matched to the ZTF reference catalog adopts the ZTF position stamped at the observation epoch but with **no** proper-motion correction from the (~decade-old) reference epoch. For a high-PM star that offset is stale and can fall off a narrow slit.

This PR skips the ZTF-ref position for stars above `max_ztfref_pm` (50 mas/yr) and falls back to the **PM-corrected Gaia position**. Low-PM stars are unaffected (still use the ZTF-ref position), and high-PM stars are still selected — they are not dropped.

## Relationship to #6243

[#6243](https://github.com/skyportal/skyportal/pull/6243) is the primary fix: it extracts the ZTF ref-coadd epoch and PM-corrects the ZTF-ref position via `apply_space_motion` when the epoch is known. This PR is the safety net for the high-PM case — in particular #6243's `ztfref_epoch is None` fallback still uses the un-PM-corrected ZTF position, where a high-PM star would remain stale.

On merge these touch the same branch and should be reconciled. The intended end state is to fold this guard into #6243's epoch-unknown fallback — roughly:

```python
if ztfdist < 0.5 * u.arcsec:
    if ztfref_epoch is not None:
        ...  # #6243: PM-correct the ZTF position
    elif np.hypot(source["pmra"], source["pmdec"]) < max_ztfref_pm:
        ...  # static ZTF position only when PM is small
    # else: fall through to the PM-corrected Gaia position
```

## Test plan
- [x] `skyportal/tests/tools/test_offset_util.py::test_get_nearby_offset_stars_ztfref_high_pm_uses_gaia` — high-PM star uses the PM-corrected Gaia position (~1.5″ shift), low-PM star keeps the ZTF position (~0″), neither is dropped, and the high-PM star survives even with no ZTF catalog
- [x] `ruff check` / `ruff format --check` clean on changed files
- [ ] CI: full offset/tools test suite

### DIFF
--- a/skyportal/tests/tools/test_offset_util.py
+++ b/skyportal/tests/tools/test_offset_util.py
@@ -1,9 +1,11 @@
 import uuid
 
+import astropy.units as u
 import numpy as np
 import numpy.testing as npt
 import pytest
 import requests
+from astropy.table import Table
 from requests.exceptions import ConnectionError, HTTPError, MissingSchema, Timeout
 
 from skyportal.models import Photometry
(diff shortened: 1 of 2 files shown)


mcoughlin:
I think the important part of this were covered elsewhere at this point. Thanks @profjsb !

PR #6250, opened 2026-06-24
typechecking CI

mcoughlin:
This PR adds dedicated typechecking CI.


skyportal/handlers/api/photometry.py
```diff
@@ -2349,6 +2349,44 @@ def delete(self, photometry_id: int):
...
+            schema:
+              type: string
+            description: |
+              Magnitude or zeropoint system of the output (e.g. "ab" or "vega").
+              Defaults to "ab".
+        responses:
+          200:
+            content:
+              application/json:
+                schema: ArrayOfPhotometrys
```
thomasculino: Is it normal to have a different schema compared to the one line 2890?

thomasculino (approved):

skyportal/handlers/api/photometry.py
mcoughlin: Good call!

PR #6251, opened 2026-06-25
sentry query issues

mcoughlin:
This PR fixes two issues identified by sentry: (i) a slow instrument query and (ii) a stream permissions access error.

### DIFF
--- a/skyportal/handlers/api/group.py
+++ b/skyportal/handlers/api/group.py
@@ -572,9 +572,9 @@ class GroupUserHandler(BaseHandler):
 
         async with self.AsyncSession() as session:
             group = await session.scalar(
-                Group.select(session.user_or_token, mode="update")
-                .options(selectinload(Group.streams))
-                .where(Group.id == group_id)
+                Group.select(session.user_or_token, mode="update").where(
+                    Group.id == group_id
+                )
             )
             if group is None:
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/group.py
```diff
@@ -584,20 +584,37 @@ async def post(self, group_id: int, *ignored_args):
...
-                if stream.id not in user_streams:
-                    return self.error(
-                        f"User with ID {user_id} ({user.username}) does not have stream access with ID {stream.id} ({stream.name}). Please contact a super-admin to grant access.",
-                        status=403,
-                    )
+            # Validate the user has access to every stream of the group via SQL
+            # (avoid lazy `group.streams`/`user.streams`, whose access-controlled
+            # load raises AccessError for a requester lacking read access to a
+            # stream — same pattern as GroupStreamHandler.post below).
+            from ...models import StreamUser  # local import to avoid cycles
```
thomasculino: Is it cycling if we put it at the top?

thomasculino (approved):
LGTM! Just one small question on a comment, nothing blocking

skyportal/handlers/api/group.py
mcoughlin: Good point!

PR #6252, opened 2026-06-25
add follow-up form key

mcoughlin:
This PR adds a follow-up form key so form data refreshes.


thomasculino (approved):
Let's hope it fixes it

PR #6253, opened 2026-06-30
enh: enable motifications for all types of reminders and improve reminder for shifts

camilledzt:
The goal of this PR is:
* enable email notifications for shifts
* improve reminders for shifts (ie. to have reminders before a shift is starting and not from the date in the form)
* adding differents scale of reminders. default was days, adding minutes, hours and weeks
* adding notifications preferences for reminders


mcoughlin (approved):
LGTM

PR #6254, opened 2026-06-30
enh: enable email notifications for mentions in comments

camilledzt:
The goal of this PR is to enable mention notifications when it's set in the profile so that you can receive email when enabled

image


mcoughlin (approved):
LGTM

PR #6255, opened 2026-06-30
faster localization table writes

mcoughlin:
~This PR uses latest healpix-alchemy to do fast localization table writes.~

`add_tiles_and_properties_and_contour` built one `LocalizationTile` ORM object per sky map tile before inserting them, which is slow and holds the whole tile set in memory. Tiles are now streamed into PostgreSQL with `COPY`.

- Replace `session.add_all(tiles)` with `COPY localizationtiles ... FROM STDIN` over the localization's UNIQ array. Measured on 423,600 tiles: 29.7s down to 13.3s, with no row object built per tile.
- Convert each UNIQ index to its `int8range` tile through the `Tile` type already attached to the column, so the rows written are identical to the ones the ORM produced. Checked against the previous code path on a real localization: same `healpix`, `probdensity` and `dateobs`, row for row.
- Read the deferred `uniq` and `probdensity` columns before opening the COPY. Loading them inside it would emit a SELECT on a connection stuck in COPY and hang, which is reachable whenever the earlier default GCN tag commit expires the localization.

No dependency change: this stays on `healpix-alchemy==1.1.0` from PyPI. An earlier version of this PR pinned a git commit of `healpix-alchemy` for a vectorized UNIQ decoder, but it turned out to make no difference end to end (13.36s vs 13.30s) because the decoding overlaps with the COPY stream, so it was dropped.


antoine-le-calloch (approved):
LGTM

PR #6257, opened 2026-06-30
update galaxy catalog ingestion

mcoughlin:
This PR updates the functions for galaxy catalog ingestion.


skyportal/handlers/api/galaxy.py
```diff
@@ -1086,282 +1086,237 @@ async def post(self):
...
-            df = df.replace({np.nan: None})
-            positive_definite_parameters = [
-                "distmpc",
-                "distmpc_unc",
-                "redshift_error",
-            ]
-            # remove rows where any of the positive definite parameters are negative using pandas
-            for key in positive_definite_parameters:
-                df = df.drop(df.index[df[key] < 0])
-            # if ra, dec or name are missing, remove the row
```
thomasculino, line 1248: Is this positive check not needed anymore?

skyportal/handlers/api/galaxy.py
```diff
@@ -1086,282 +1086,237 @@ async def post(self):
...
+                length, blueshift_length = _copy_galaxies(df, catalog_id)
+                full_length += length
+                full_blueshift_length += blueshift_length
+                log(
+                    f"{label} - rows {start}-{start + len(sub)}: added {length} "
+                    f"galaxies (including {blueshift_length} with a negative "
+                    f"redshift) in {time.perf_counter() - start_timer:0.2f} s"
+                )
+            except Exception as e:
+                log(f"{label} - chunk at row {start}: Error: {e}")
```
thomasculino, line 1227: Maybe we should add a rollback in the Except block. If one chunk fails then the rest of the chunk will automatically fail as well.

mcoughlin: done

skyportal/handlers/api/galaxy.py
mcoughlin: I brought this back.

thomasculino (approved):
👌

PR #6280, opened 2026-07-01
mui v9

mcoughlin:
This PR bumps our dependencies to mui v9.


stefanv:
Too many to read through, but from a first glance these are mainly API translations. So, if the site renders OK for you, I'd say go for it and keep an eye out for any anomalies in the next few days.

PR #6282, opened 2026-07-02
Remove legacy peer deps flag as it doesn't exist for Bun

leonitousconforti:
The legacy peer deps flag doesn't exist for Bun so it isn't needed after switching from npm to bun

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -298,7 +298,7 @@ If you have trouble running `make run` and see this error:
 make[1]: *** [dependencies] Error 1
 make: *** [run] Error 2
 ```
-You may need to run `npm install --legacy-peer-deps` and then try `make run` again.
+You may need to run `bun install` and then try `make run` again.
 
 ---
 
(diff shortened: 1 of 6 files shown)


doc/setup.md
```diff
@@ -252,7 +253,7 @@ If you have trouble running `make run` and see this error:
 make[1]: *** [dependencies] Error 1
 make: *** [run] Error 2
 ```
-You may need to run `npm install --legacy-peer-deps` and then try `make run` again.
+You may need to run `bun install` and then try `make run` again.
```
stefanv: You may want to ensure that you lock dependencies into `bun.lock`. Those are then installed with `bun install --frozen-lockfile`.

mcoughlin:
@leonitousconforti rebase.

stefanv (approved):

stefanv:
Looks good to me overall; haven't dug into why CI is failing. If unrelated, please go ahead and merge.

leonitousconforti:
> Looks good to me overall; haven't dug into why CI is failing. If unrelated, please go ahead and merge.

Yeah CI is unrelated. I canceled a bunch or jobs after rebasing some prs because CI was taking forever to process jobs. Maybe GitHub was having an outage but also one PR consumes all the allotted runners for the free plan (which we are on) for 20-30 minutes so jobs were just taking a while to get an available runner

PR #6283, opened 2026-07-02
Fix async handler commit

antoine-le-calloch:
* Restrict GCN event updates to the original poster only.
* Prevent `MissingGreenlet` in `get_source` by attaching the eagerly fetched host so `host_offset` and `host_distance` don't lazy-load the relationship.
* Query thumbnail `file_uris` via the flush connection in the `before_delete` listener so it doesn't lazy-load `target.thumbnails` during an async commit's flush.
* Restore the psycopg3 `COPY` API in `add_glade` so GLADE ingestion uses `cursor.copy` instead of psycopg2's nonexistent `copy_from`.
* Map returned photometry IDs back by deduplication key in the error branch, since `insertmanyvalues` doesn't preserve input order under `ON CONFLICT`.


mcoughlin (approved):
Great. Is the baselayer bump needed?

antoine-le-calloch:
> Great. Is the baselayer bump needed?

Nop but still good to have it.

PR #6285, opened 2026-07-03
broker APIs

mcoughlin:
This PR provides a generic broker API interface for SkyPortal. The idea is to emulate what we do for facility APIs but do the same for brokers, providing both an Alert interface like we have in Fritz and a service plugin for stream ingestion (which we currently maintain separately in the plugins for each broker now).

@thomasculino @leonitousconforti @antoine-le-calloch @camilledzt 
I am interested in your feedback on going this direction.


skyportal/handlers/api/broker.py
thomasculino, line 1: Only `BrokerAlertsHandler` uses `broker.active`. All the handlers should add a `broker.active` check before actually doing anything.

thomasculino (approved):
One small comment but the rest looks exciting

PR #6286, opened 2026-07-04
Connection pooling

mcoughlin:
This PR introduces an optional transaction-mode connection pooler.

### DIFF
--- a/.github/workflows/test_api_and_frontend.yaml
+++ b/.github/workflows/test_api_and_frontend.yaml
@@ -31,6 +31,10 @@ jobs:
         env:
           POSTGRES_USER: skyportal
           POSTGRES_PASSWORD: anything
+          # Trust host connections too (already trust for local): the connection
+          # pooling test runs pgbouncer against this DB, and pgbouncer has no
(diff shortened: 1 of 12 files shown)


antoine-le-calloch (commented):
Looks good overall, one small comment, though it concerns the **Baselayer** changes rather than this PR.

In `_resolve_pooler`, the `connect_args` merge currently puts 
`prepare_threshold: None` before the caller's `connect_args`, meaning a caller-supplied `prepare_threshold` would override it:

```python
"connect_args": {
    "prepare_threshold": None,
    **(engine_args.get("connect_args") or {}),  # caller wins
},
```

Since disabling prepared statements is mandatory under transaction pooling, I'd reverse the order so `None` always wins:

```python
"connect_args": {
    **(engine_args.get("connect_args") or {}),
    "prepare_threshold": None,
},
```

Other than that, this looks good to me. It would just be good to test it in preview.

mcoughlin:
Thanks @antoine-le-calloch, see https://github.com/cesium-ml/baselayer/pull/434

PR #6292, opened 2026-07-05
intro tour

mcoughlin:
This PR adds an intro tour on the homepage.


leonitousconforti (approved):

leonitousconforti (approved):
Maybe some tests need to be updated?

PR #6293, opened 2026-07-05
sortable annotation and altdata columns in the sources table

### LINKED ISSUE #5631: Add columns to the Source view representing annotations
We are using skyportal as a submdule in our deployment called MNTLEAF. Our focus is microlensing.

**Feature Summary**
On the [Source page](https://mntleaf.astro.berkeley.edu/sources), the 24 columns that are visible do not include any annotations. The values that we find most useful are uploaded as annotations. The Einstein crossing time is an example. The first sort that we perform on our candidates is on this value, we want quickly view the longest events. 

**Usage / behavior**
From the source page, a user could modify the visible columns using the 'Show columns' button that currently exists, adding any given annotation as its own column. If these are numerical, they can then be easily sorted and the longest duration events identified. 

**Alternative Solutions**
In this screenshot of the 'Show Columns' dropdown, there are two empty slots. Even adding a single annotation would be useful. That annotation could be identified as having a specific origin/key pair that shows up by default in one of the empty slots. 

If the 'Filter sources' dropdown was a little intuitive (perhaps with mouse-overs with examples on how to formulate the conditionals, that would also be useful.)

Image

**Additional context**
One complication is that some annotations can have the same key and different origins, making the proposed drop down have too many options to choose from.

I tried to sort the source table with the 'Filter' option, but sorting the annotations was non-intuitive. I was able to do a few simple sorts by examining the API documentation, but the proposed feature would be much simpler.

mcoughlin:
This PR adds sortable annotation and altdata columns to the sources table.

Closes https://github.com/skyportal/skyportal/issues/5631

### DIFF
--- a/skyportal/handlers/api/sources.py
+++ b/skyportal/handlers/api/sources.py
@@ -1847,12 +1847,28 @@ async def get_sources(
 
                     if annotation_sort:
                         # Correlated subquery on the grouped obj id; -> keeps the
-                        # JSONB type so numeric values sort numerically.
+                        # JSONB type so numeric values sort numerically. Restrict
+                        # to annotations the user can access so sorting can't leak
+                        # values from groups they don't belong to.
+                        ann_access_clause = ""
+                        if not is_admin:
+                            ann_groups_str, ann_groups_bindparams = array2sql(
+                                user_accessible_group_ids,
+                                type=sa.Integer,
+                                prefix="sort_ann_group",
+                            )
+                            query_params.extend(ann_groups_bindparams)
+                            ann_access_clause = (
+                                "AND annotations_sort.id in (select annotation_id "
+                                "from group_annotations where group_id in "
+                                f"{ann_groups_str}) "
+                            )
                         statement += (
                             "ORDER BY (SELECT annotations_sort.data -> :sort_ann_key "
                             "FROM annotations AS annotations_sort "
                             "WHERE annotations_sort.obj_id = objs.id "
-                            "AND annotations_sort.origin = :sort_ann_origin LIMIT 1) "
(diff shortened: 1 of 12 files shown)


skyportal/handlers/api/sources.py
```diff
@@ -1834,7 +1845,21 @@ async def get_sources(
...
+                            "ORDER BY (SELECT annotations_sort.data -> :sort_ann_key "
+                            "FROM annotations AS annotations_sort "
+                            "WHERE annotations_sort.obj_id = objs.id "
+                            "AND annotations_sort.origin = :sort_ann_origin LIMIT 1) "
+                            f"{sort_order.upper()} NULLS LAST"
+                        )
+                        query_params.append(sa.bindparam("sort_ann_key", sort_ann_key))
+                        query_params.append(
+                            sa.bindparam("sort_ann_origin", sort_ann_origin)
+                        )
```
thomasculino, line 1877: We should only get the annotations the user has access to. Currently, all the annotations will be displayed, even the ones from a group the user doesn't have access to.

skyportal/handlers/api/sources.py
```diff
@@ -1903,7 +1928,21 @@ async def get_sources(
...
```
thomasculino, line 1976: Same comment

skyportal/handlers/api/sources.py
mcoughlin: Good point. Fixed!

skyportal/handlers/api/sources.py
mcoughlin: Fixed!

thomasculino (approved):

PR #6296, opened 2026-07-06
spectrum smoothing bug

mcoughlin:
This PR fixes a bug in spectrum smoothing (smoothing parameter was being passed as a string).


thomasculino (approved):
LGTM!

PR #6298, opened 2026-07-06
make filters in requests appear

mcoughlin:
This PR ensures the filters in an instrument request actually appear on the front-end.

### DIFF
--- a/static/js/components/followup_request/FollowupRequestLists.tsx
+++ b/static/js/components/followup_request/FollowupRequestLists.tsx
@@ -53,6 +53,9 @@ const displayedColumns = [
   "request",
   "filter",
   "filters",
+  // Filter/band keys for instruments that don't alias them (e.g. TTT, LCO, LT)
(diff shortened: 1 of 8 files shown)


static/js/components/followup_request/FollowupRequestLists.tsx
```diff
@@ -53,6 +53,9 @@ const displayedColumns = [
   "request",
   "filter",
   "filters",
+  // Filter/band keys for instruments that don't alias them (e.g. TTT, LCO, LT)
+  "observation_choice",
+  "observation_choices",
```
thomasculino, line 58: Don't we add the `observation_choices` alias for facility APIs inside the parenthesis? The comment is getting me confused. Is this edit for other instruments?

mcoughlin: Yeah good point, I am not sure why the filters didn't appear then :(

PR #6301, opened 2026-07-07
external thumbnail sizes

### LINKED ISSUE #5615: Survey Thumbnails: heterogeneous pixel scale?
I noticed while paying attention to the thumbnails recently that the objects on the images from LS, PS1, and SDSS aren't exactly at the same positions. Basically, the "zoom" level is different, meaning the parameters used to get thumbnails for the different surveys don't result in matching images.

We should do some testing to make sure that everything matches.

mcoughlin:
This PR makes all the external thumbnails 60".

Closes https://github.com/skyportal/skyportal/issues/5615

### DIFF
--- a/skyportal/models/obj.py
+++ b/skyportal/models/obj.py
@@ -13,6 +13,11 @@ import requests
 import sqlalchemy as sa
 from astropy import coordinates as ap_coord
 from astropy import units as u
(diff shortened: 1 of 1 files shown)


antoine-le-calloch (approved):
LGTM

PR #6302, opened 2026-07-07
bulk source properties page

mcoughlin:
This PR adds a source properties page where plotting photstats across sources is possible.


leonitousconforti (approved):
Does streaming the data using [json new line delineated streams](https://www.speakeasy.com/openapi/content/jsonl/) sound like a good or bad idea? I guess it could take resources away from other things?

mcoughlin:
@leonitousconforti opened an issue here: https://github.com/skyportal/skyportal/issues/6326

PR #6304, opened 2026-07-08
fix default follow-up request front-end

mcoughlin:
This PR fixes the default follow-up request front-end.


thomasculino (approved):
Good to merge after fixing conflicts

PR #6305, opened 2026-07-09
remove duplicate download buttons

thomasculino:
Some DataGrids had 2 download buttons. This PR keeps the `showExport` parameter when possible and the older download when it is custom.


mcoughlin (approved):
LGTM

PR #6306, opened 2026-07-09
Validate default follow-up source_filter regex

mcoughlin:
This PR validates the default follow-up source_filter regex.


leonitousconforti (approved):
nice

PR #6309, opened 2026-07-10
skymapper / NASA facility cutouts

mcoughlin:
This PR makes user requested skymapper / JWST / HST / Chandra cutouts possible


skyportal/handlers/api/source.py
```diff
@@ -3345,11 +3364,22 @@ async def post(self):
...
+            # view so the new thumbnails appear without a manual reload.
+            if requested_types:
+                flow = Flow()
+                for obj in objs:
+                    flow.push(
+                        "*",
+                        "skyportal/REFRESH_SOURCE",
+                        payload={"obj_key": obj.internal_key},
+                    )
+
```
leonitousconforti, line 3382: How is the `SurveyThumbnailHandler` endpoint hit from the frontend? Or I guess why does this need to be Multicast'ed using `Flow.push("*")`?

Are users really leaving pages open on sources or source list pages for long periods of time where they are expecting the thumbnails to dynamically change without a reload? Just curious

leonitousconforti (approved):

skyportal/handlers/api/source.py
mcoughlin: SurveyThumbnailHandler has a button, and the Flow.push("*") gives the completion signal for the async action the user just triggered.

PR #6319, opened 2026-07-11
Update NED api url to use latest version and get rid of legacy one

thomasculino:
NED queries are slow currently because we use legacy API


mcoughlin (approved):
LGTM

PR #6324, opened 2026-07-14
photometry id -> bigint

mcoughlin:
This PR moves the photometry table IDs to bigint (important as our photometry tables have passed a billion rows).


leonitousconforti (approved):
Looks like just missing `typegen `

PR #6331, opened 2026-07-15
Fix stream check issue

antoine-le-calloch:
Compare `stream.id` with `accessible_stream_ids` instead of comparing the `stream` object with `requesting_user.accessible_streams`. Since the async endpoints were introduced, the objects can come from different database sessions, so the object comparison is always false.


mcoughlin (approved):
LGTM

PR #6338, opened 2026-07-16
Use Python builtin logger?

leonitousconforti:
- Root logger sits at WARNING so third-party libraries (SQLAlchemy, urllib3, tornado) can't flood the logs even if they use the builtin logger and app loggers get INFO
- the ~70 `log.error(f"...{traceback.format_exc()}")` sites could become `log.exception(...)` in except blocks and get tracebacks for free for cleaner tracebacks


leonitousconforti:
Maybe will cause lots of conflicts (although I believe many of them should be able to be resolved automatically by git)

Maybe not even desired? I'm just usually in favor of using the built-in logger.

Thoughts?

PR #6340, opened 2026-07-16
Migrate from eslint to oxctools

leonitousconforti:
https://oxc.rs/

https://github.com/oxc-project/oxc

Thoughts?


mcoughlin:
@leonitousconforti I am supportive of this.

PR #6349, opened 2026-07-17
Filter rename: require group admin, add inline rename UI

leonitousconforti:
Ported from fritz, where this runs in production.

**Backend:** `PATCH /api/filters/{filter_id}` previously allowed any group member with row-level update access to rename a filter. It is now gated behind group admin / system admin and returns 403 otherwise. Tests cover group admin (allowed), plain member (403), and super admin (allowed).

**Frontend:** on the group page, group admins get an edit icon next to each filter that renames it inline (Enter saves, Escape cancels) via the existing PATCH endpoint.


mcoughlin:
I will take care of this one.

PR #6350, opened 2026-07-17
Modernize db_stats cron-job queries to session.scalars()

leonitousconforti:
`session.execute(select(CronJobRun.script)).all()` yields `Row` objects, so the later `CronJobRun.script == script` comparison binds a Row rather than the string. Switch both cron-job queries to `session.scalars()` and drop the manual row unpacking. No behavior change otherwise.

Extracted from fritz.


mcoughlin (approved):
LGTM

PR #6366, opened 2026-07-21
fix broken BOOM schema in filter front-end

mcoughlin:
This PR fixes the broken BOOM schema in filter front-end.


thomasculino (approved):
Looks good

PR #6372, opened 2026-07-22
filter candidates by filter within group

mcoughlin:
This PR enables filtering candidates by filter within group.


leonitousconforti (approved):
👍

PR #6375, opened 2026-07-22
Fix and improve data grid

antoine-le-calloch:
- Refactored the DataGrid components to improve readability, maintainability, and reduce code duplication.
- Improved table and Tabs display and performed general UI cleanup.
- Added TaxonomyList and enhanced TaxonomyTable for better taxonomy management.
- Fixed a permission check bug in the taxonomy access control logic.
- Improved CSV exports by adding missing valueGetters for group names, share groups, and allocation users.
- Enabled showExport by default to match the DataGrid default behavior.
- Added a link to the allocation page.
- Removed incorrect server-side sorting.
- Fixed a race condition in candidate.py that could cause psycopg.errors.UniqueViolation when multiple requests created the same object concurrently.


mcoughlin (approved):
Looks great

PR #6377, opened 2026-07-23
Nix flake maybe?

leonitousconforti:
I like nix flakes, if you don't like them we don't have to merge it but I find it really nice


mcoughlin (approved):
LGTM

stefanv:
Some docs around this would be helpful for those of us not familiar with Nix, what it provides, how to use it, etc.

PR #6382, opened 2026-07-23
Pydantic-typed endpoint validation + OpenAPI docs (prototype on annotation POST)

leonitousconforti:
Handlers call body = self.parse_body(Model) (new BaseHandler helper:
400 with field-level errors, aborts via tornado Finish), and
spec_from_handlers reads the same models from the method's type hints
(keyword-only body param annotation, return annotation) to generate
requestBody/response docs, so docs cannot drift from validation.

Pilot: annotation POST. Request models use extra="forbid", which
surfaced a real bug: PhotometryPlot sent group selection as 'groups'
(backend expects 'group_ids'), silently ignoring the chosen groups.

### DIFF
--- a/skyportal/handlers/api/annotation.py
+++ b/skyportal/handlers/api/annotation.py
@@ -1,8 +1,8 @@
-import re
 import time
-from collections.abc import Mapping
+from typing import Any
 
 from marshmallow.exceptions import ValidationError
+from pydantic import BaseModel, ConfigDict, Field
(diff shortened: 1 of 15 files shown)


mcoughlin:
@leonitousconforti I assume this is meant to replace all the instances where we otherwise do the docs by hand, and leaving marshmallow for the ones aligned with the database schema. If so, this is great.

PR #6388, opened 2026-07-23
show schema and dropdown correctly

thomasculino:
Schema was still not showing properly. Also, small screens were experiencing dropdowns collapsing instantly.


mcoughlin (approved):
LGTM

PR #6390, opened 2026-07-24
try to prevent group sources page crashes

mcoughlin:
This PR tries to prevent group sources page crashes (https://github.com/fritz-marshal/fritz/issues/635)


thomasculino (approved):
Looks promising

PR #6393, opened 2026-07-24
Pydantic-typed request bodies for group admission request endpoints

leonitousconforti:
Continues the typed-endpoint migration started in #6382, converting `GroupAdmissionRequestHandler` POST and PATCH to the `parse_body` pattern.

- `GroupAdmissionRequestPostBody` (`groupID`/`userID`) and `GroupAdmissionRequestPatchBody` (`status` as a `Literal` enum), both `extra="forbid"`, replace the hand-rolled presence/int-parse/enum checks; POST also gains a typed `GroupAdmissionRequestPostResponse` (`{id}`), so its `requestBody:`/`responses:` docstring YAML is now generated from the hints.
- Client sweep: the RTK mutations in `static/js/ducks/groupAdmissionRequests.ts` and all test call sites already send exactly these keys; no tests assert the old error strings, so no client changes were needed. (The DELETE tests send a redundant `status` body, but DELETE has no body model and ignores it.)

### DIFF
--- a/skyportal/handlers/api/group_admission_request.py
+++ b/skyportal/handlers/api/group_admission_request.py
@@ -1,3 +1,6 @@
+from typing import Literal
+
+from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy.orm import selectinload
 
 from baselayer.app import models as baselayer_models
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #6394, opened 2026-07-24
Pydantic-typed request bodies for filter endpoints

leonitousconforti:
Continues the typed-endpoint migration started in #6382, converting `FilterHandler` POST and PATCH to the `parse_body` pattern.

- `FilterPostBody` / `FilterPatchBody` (`extra="forbid"`) replace the marshmallow `Filter.__schema__().load(...)` calls, with explicit `name`/`stream_id`/`group_id`/`altdata` fields instead of the opaque `FilterNoID` docstring reference; POST gains a typed `FilterPostResponse` (`{id}`).
- PATCH keeps the old semantics: `group_id`/`stream_id` are rejected with "Cannot update group_id or stream_id." unless they match the current values (the marshmallow ModelSchema PK-merge made same-value no-ops pass before), and only `name`/`altdata` are applied.
- Client sweep: the RTK mutations in `static/js/ducks/filter.ts` send `{name, group_id, stream_id}` / `{name}`, and all test call sites (test_filters.py, test_groups.py, test_scanning_page.py) match; `public_filter` fixtures use `FilterFactory` directly. No client changes were needed.

### DIFF
--- a/skyportal/handlers/api/filter.py
+++ b/skyportal/handlers/api/filter.py
@@ -1,10 +1,54 @@
-from marshmallow.exceptions import ValidationError
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy.orm import joinedload
 
 from baselayer.app.access import auth_or_token, permissions
 
 from ...models import Filter
 from ..base import BaseHandler
+
+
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #6395, opened 2026-07-24
merge all xmatches catalogs and add DESI_DR1 & LS_DR10_PHOTOZ

thomasculino:
Adding DESI_DR1 & LS_DR10_PHOTOZ to the list of visible catalogs on the filter building UI. 

Each survey was using its own set of crossmatching catalogs. This PR also merges these different sets into one set of corssmatching catalogs used by all currently supported surveys. (Side note, those catalogs schemas are currently being hardcoded. It would nice to expose their schemas in BOOM.)

### DIFF
--- a/static/js/constants/crossmatch.js
+++ b/static/js/constants/crossmatch.js
@@ -608,28 +608,123 @@ const LSPSC = {
   default: null,
 };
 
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #6396, opened 2026-07-25
fix nested annotations

thomasculino:
Some nested annotations on the filter building UI were not properly projected all the way to the last project stage so they were showing up as null on the annotations table on the source page.


mcoughlin (approved):
LGTM

PR #6399, opened 2026-07-25
Pydantic-typed request body for data sharing endpoint

leonitousconforti:
Continues the typed-endpoint migration (#6382, #6392–#6394, #6397–#6398), converting `SharingHandler` POST (`/api/sharing`) to the `parse_body` pattern.

- `SharingPostBody` (`extra="forbid"`) types `groupIDs` (required, `min_length=1`, matching the old empty-list rejection) and optional `photometryIDs`/`spectrumIDs`; the cross-field "one of either photometryIDs or spectrumIDs" check stays in the handler with its original message.
- Client sweep: `shareData` in `static/js/ducks/source.ts` (used by `ShareDataForm.tsx`) sends exactly `{groupIDs, photometryIDs, spectrumIDs}`, and the test_sharing.py call sites match; error assertions there target handler-level messages that are unchanged.
- `static/js/types/api.ts` regenerated.

PR #6397, opened 2026-07-25
Pydantic-typed request body for user ACL endpoint

leonitousconforti:
Continues the typed-endpoint migration (#6382, #6392–#6394), converting `UserACLHandler` POST (`/api/user/{user_id}/acls`) to the `parse_body` pattern.

- `UserACLPostBody` (`aclIds: list[str]`, `extra="forbid"`) replaces the hand-rolled presence/isinstance checks; the unknown-ACL check (single-round-trip id validation) stays in the handler.
- Client sweep: `static/js/ducks/acls.ts` (used by `UserManagement.tsx`) and all test_acls.py call sites already send exactly `{aclIds}`; the shape-error test asserts only the 400 status, so no client changes were needed.
- `static/js/types/api.ts` regenerated.

PR #6392, opened 2026-07-24
Pydantic-typed request bodies for source label endpoints

leonitousconforti:
Continues the typed-endpoint migration started in #6382, converting `SourceLabelsHandler` POST and DELETE (`/api/sources/{obj_id}/labels`) to the `parse_body` pattern.

- `SourceLabelsPostBody` / `SourceLabelsDeleteBody` (`extra="forbid"`) replace the hand-rolled `groupIds` checks; the `requestBody:` docstring sections are now generated from the model hints.
- Client sweep: the only callers (`addSourceLabels`/`deleteSourceLabels` in `static/js/ducks/source.ts`, used by `SourceTable.tsx`, and two call sites in `test_sources.py`) already send exactly `{groupIds}`, so no client changes were needed.
- `static/js/types/api.ts` regenerated.

PR #6400, opened 2026-07-25
Pydantic-typed request bodies for stream endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382, #6392–#6394, #6397–#6399), converting `StreamHandler` POST/PATCH and `StreamUserHandler` POST to the `parse_body` pattern.

- `StreamPostBody`/`StreamPatchBody` (`extra="forbid"`) replace the `Stream.__schema__().load(...)` calls with explicit `name`/`altdata`/`auto_join` fields — the old docstrings omitted `auto_join` even though the marshmallow schema accepted it (and tests send it). PATCH keeps `name` required (the old non-partial `schema.load` required it) and uses `model_fields_set` so an explicit `altdata: null` still clears the field; its documented 200 response is corrected to plain `Success` (the handler never returned the stream). POST and `StreamUserHandler` POST gain typed responses.
- Client sweep: no frontend UI creates/updates streams; `addStreamUser` in `static/js/ducks/streams.ts` sends exactly `{user_id}`; all test call sites (test_streams.py, test_profile.py, test_groups.py) match; the demo-data loader posts `{name, altdata}` (`=id` aliases are stripped before POSTing). Error assertions target unchanged handler messages.


mcoughlin:
@leonitousconforti rebase

mcoughlin (approved):
LGTM

PR #6401, opened 2026-07-25
Pydantic-typed request bodies for team endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382, #6392–#6394, #6397–#6400), converting `TeamHandler` POST and PUT to the `parse_body` pattern.

- `TeamPostBody`/`TeamPutBody` (`extra="forbid"`) type the seven `EDITABLE_FIELDS` scalars plus `group_ids`; both endpoints gain typed `{id}` responses. PUT keeps its field-presence semantics via `model_fields_set` (only keys actually sent are applied; an explicit `group_ids: null` still clears the team's groups, and the `name` non-empty guard is preserved), and the whitespace-only-name rejection on POST stays in the handler.
- Client sweep: the team form in `Teams.tsx` sends exactly these keys (via `static/js/ducks/teams.ts`), and all test call sites match. One assertion updated: `test_cannot_create_team_without_name` now expects the pydantic `name: Field required` message instead of the old hand-rolled one.


mcoughlin (approved):
LGTM

PR #6402, opened 2026-07-25
Pydantic-typed request bodies for source groups endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `SourceGroupsHandler` POST and PATCH to the `parse_body` pattern.

- `SourceGroupsPostBody` (`objId` + `inviteGroupIds`/`unsaveGroupIds`, `extra="forbid"`) replaces the hand-rolled checks; pydantic's lax int coercion takes over the explicit string→int cast that was added for clients sending string ids. The "one of either" cross-field check stays in the handler with its original message.
- `SourceGroupsPatchBody` types `groupID`/`active`/`requested` as required, matching the documented contract (the old code silently tolerated a missing `active`, which would have crashed on commit).
- Client sweep: `updateSourceGroups` (used by `EditSourceGroups.tsx` and `SaveCandidateGroupsDialog.tsx`) and `acceptSaveRequest`/`declineSaveRequest` in `static/js/ducks/source.ts` send exactly these keys, as do all test call sites.
- `static/js/types/api.ts` regenerated.

### DIFF
--- a/skyportal/handlers/api/source_groups.py
+++ b/skyportal/handlers/api/source_groups.py
@@ -1,4 +1,5 @@
 import sqlalchemy as sa
+from pydantic import BaseModel, ConfigDict, Field
 
 from baselayer.app.access import permissions
 from baselayer.log import make_log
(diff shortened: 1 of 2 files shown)


mcoughlin (approved):
LGTM

PR #6403, opened 2026-07-25
Pydantic-typed request body for recurring API endpoint

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `RecurringAPIHandler` POST to the `parse_body` pattern.

- `RecurringAPIPostBody` (`extra="forbid"`) types the five required fields plus optional `number_of_retries` (`le=10`, replacing the manual cap check). The old docstring documented `payload` as an object, but the handler (and the only client) require a JSON **string** — the model now documents the real contract. The arrow `next_call` parse and the `payload must be a valid JSON string` check (exact message asserted in tests) stay in the handler, and the marshmallow load is kept for datetime construction.
- `method` is now required (it always was per the docstring and the RecurringAPI model; the old `if "method" in data` guard just deferred the failure to marshmallow). Also fixes the method-rejection message, which was missing its f-string prefix and printed the literal template.
- Client sweep: `NewRecurringAPI.tsx` (via `ducks/recurring_apis.ts`) sends exactly `{endpoint, method, next_call, call_delay, payload}`; the single test call site matches.

### DIFF
--- a/skyportal/handlers/api/recurring_api.py
+++ b/skyportal/handlers/api/recurring_api.py
@@ -2,6 +2,7 @@ import json
 
 import arrow
 from marshmallow.exceptions import ValidationError
+from pydantic import BaseModel, ConfigDict, Field
 
 from baselayer.app.access import auth_or_token, permissions
 from baselayer.app.env import load_env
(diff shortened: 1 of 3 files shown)


mcoughlin:
@leonitousconforti rebase

mcoughlin (approved):
LGTM

PR #6404, opened 2026-07-25
Pydantic-typed request bodies for listing endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `UserObjListHandler` POST and PATCH (`/api/listing`) to the `parse_body` pattern.

- `ListingPostBody` (`obj_id`/`list_name` required, optional `user_id`/`params`, `extra="forbid"`) replaces the marshmallow load; POST gains a typed `{id}` response. The `check_list_name` regex, admin-only-for-other-users, and all watchlist `params` checks stay in the handler with their original messages (tests assert them). A watchlist POST with an explicit `params: null` now gets the intended "must be provided" error instead of slipping through to a null-params listing.
- `ListingPatchBody` keeps the partial-update semantics (each of `user_id`/`obj_id`/`list_name` falls back to the existing listing value when omitted), which the tests exercise field-by-field.
- DELETE is intentionally untouched: its body is only read when no `listing_id` path param is given, which needs an optional-body convention before it can carry hints.
- Client sweep: `favorites.ts`/`rejected_candidates.ts` ducks send `{obj_id, list_name}`; all test call sites send `{user_id, obj_id, list_name}`. No client changes needed.

### DIFF
--- a/skyportal/handlers/api/user_obj_list.py
+++ b/skyportal/handlers/api/user_obj_list.py
@@ -1,6 +1,8 @@
 import re
+from typing import Any
 
 from marshmallow.exceptions import ValidationError
+from pydantic import BaseModel, ConfigDict, Field
 
 from baselayer.app.access import auth_or_token
 from baselayer.app.custom_exceptions import AccessError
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #6405, opened 2026-07-25
Pydantic-typed request bodies for taxonomy endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `TaxonomyHandler` POST and PUT to the `parse_body` pattern.

- `TaxonomyPostBody` (`extra="forbid"`) types the documented fields **plus** `hierarchy_file`, which the old docstring omitted even though the handler (and `TaxonomyForm.tsx`) rely on it — it's a YAML string parsed into `hierarchy`. The tdtax schema validation, the "A JSON of the taxonomy must be given" cross-field check (exact message asserted in tests), and the public-group fallback for missing/empty `group_ids` stay in the handler. POST gains a typed `{taxonomy_id}` response. Note the docstring previously claimed `group_ids` defaults to the user's groups; the handler actually falls back to the public group, and the model documents the real behavior.
- `TaxonomyPutBody` replaces the unvalidated `setattr`-any-key loop with explicit `name`/`version`/`provenance`/`isLatest`/`group_ids` applied via `model_fields_set`; `hierarchy` is a declared field so submitting it still returns the specific "Editing the hierarchy not allowed" error rather than a generic extra-key rejection.
- Client sweep: `TaxonomyForm.tsx` (via `ducks/taxonomies.ts`) sends `{name, version, provenance, hierarchy_file, isLatest, group_ids}` on create and `{name, version, provenance, group_ids}` on edit; all ~25 test call sites and `data/variable_taxonomy.yaml` seed data match.

### DIFF
--- a/skyportal/handlers/api/taxonomy.py
+++ b/skyportal/handlers/api/taxonomy.py
@@ -1,6 +1,9 @@
+from typing import Any
+
 import sqlalchemy as sa
 import yaml
 from jsonschema.exceptions import ValidationError as JSONValidationError
+from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy.orm import selectinload
 from tdtax import schema, validate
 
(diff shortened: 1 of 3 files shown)


mcoughlin:
@leonitousconforti fix test and rebase

mcoughlin (approved):
LGTM

PR #6406, opened 2026-07-25
Pydantic-typed request bodies for tag endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `ObjTagOptionHandler` POST/PATCH and `ObjTagHandler` POST to the `parse_body` pattern.

- `ObjTagOptionPostBody`/`ObjTagOptionPatchBody` (`name` required, `color` optional, `extra="forbid"`): the letters-and-numbers name regex and hex-color checks stay in the handler — their exact messages are asserted by tests, including the edge case that an empty-string `color` is accepted and stored as-is. The `data=""` tests still 500 in `get_json` before validation runs, so that path is unchanged.
- `ObjTagPostBody` documents `group_ids` as optional (the old docstring claimed it was required, but the handler falls back to the public group when it's missing or empty).
- Client fix: `updateTagOption` in `static/js/ducks/objectTags.ts` was sending the redundant `id` (already in the URL path) inside the PATCH body, which `extra="forbid"` now rejects — the duck strips it from the body.
- `ObjTagHandler` DELETE is left untouched: its body is optional (`get_json() or {}`), which needs an optional-body convention first.
- `static/js/types/api.ts` regenerated.

### DIFF
--- a/skyportal/handlers/api/tag.py
+++ b/skyportal/handlers/api/tag.py
@@ -8,6 +8,7 @@ This module provides REST API endpoints for:
 import re
 
 import sqlalchemy as sa
+from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy import func
 from sqlalchemy.orm import selectinload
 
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #6416, opened 2026-07-25
pydantic-typed query parameters for all API endpoints

leonitousconforti:
Extends the typed-endpoint migration (#6382 and follow-ups) from request bodies to query strings, covering **every public GET endpoint** plus the query parameters on non-GET methods. Builds on the invitations pilot that was the original scope of this PR.

## Mechanism

The keyword-only parameter *name* selects the role (`body:` vs `query:`), so a method can carry both.

- `self.parse_query(Model)` (`BaseHandler`) validates `request.query_arguments` and 400s with the standard `Invalid/missing parameters: ...` message. `query_dict_from` decodes tornado's `name → [bytes]`: `list[X]` fields accept both repeated params (`?ids=1&ids=2`) and comma-separated values (`?ids=1,2`), everything else keeps the last value.
- `spec_from_handlers` renders model fields as OpenAPI `parameters:`, and injection now happens *after* the `single:`/`multiple:` split, so it targets each subspec instead of the top level. That removes the documented blocker on the 36 `single:`/`multiple:` handler files, and those docstrings' `- in: query` entries are deleted in favour of the model.
- Query models are flat (scalars, lists, `Literal`s) and all use `extra="forbid"`.

955 query parameters are now generated from the models rather than hand-written YAML.

## Bugs fixed

Server:
- `source.py`: `hasTNSname`/`hasNoTNSname` read the raw string, so `hasTNSname=false` filtered *to* TNS-named sources. `maxAirmass` reached `AirmassConstraint()` as a string; an unknown `twilight` raised `AttributeError` (500) rather than 400.
- `thumbnail.py`: `types` came back as a string and was passed to `Thumbnail.type.in_()`, which iterated it character by character, so any explicit `types` value matched nothing (GET and PATCH).
- `obj.py`: `validate_list_parameter` returned `self.error(...)`, which returns `None`, so the handler wrote a 400 and then kept executing and wrote a second response.
- `candidate/scan_report.py`: `?page=-1` produced a negative OFFSET and `?numPerPage=-5` a negative LIMIT, both 500s from the driver.
- `photometric_series.py`: a non-numeric `numPerPage` raised an uncaught `ValueError` (500).
- `galaxy.py`: the `catalog_name` query arg unconditionally overwrote the documented `/api/galaxy_catalog/{catalog_name}` path value, so the path segment was ignored.
- `instrument_log.py`: unreachable date fallbacks, documented backwards.

Client:
- `API.ts`: `filterOutEmptyValues` tested `typeof key === "number"` (a key is always a string) instead of the value, so a numeric `0` was silently dropped from every query string despite the comment saying otherwise.
- `earthquake`: the status filter sent `tagKeep`/`tagRemove` while the server reads `statusKeep`/`statusRemove`, so it never filtered; the table also declared `sortingMode="server"` for an endpoint with no sorting, so clicking a header did nothing (now sorts client-side).
- `candidates`: `requireDetections: false` was stripped before it reached the server, making the checkbox a no-op.
- `QuickSearchBar`/`SourceIdAutocomplete` sent a `totalMatches` query param (a response field) to three routes; `GcnSelectionForm` forwarded galaxy-only fields to `/api/sources`; `sources.ts` sent a long-dead `includeLocalizationStatus`.

## Notes

- Deliberate tightening: garbage values now 400 instead of silently becoming the default (`?pageNumber=abc` used to mean 1), and unknown params 400 via `extra="forbid"`.
- `internal/` GETs are excluded: `spec_from_handlers` drops them from the spec, so there is no docs payoff to weigh against the added strictness.
- `followup_request`'s `refreshSource`/`refreshRequests` are also excluded: they take their default from the JSON body, so they belong with the request-body work rather than here.
- Merged `main` in and resolved the drift, including the query params upstream added meanwhile (`gcn` `groupIds`) and the new `BrokerFilterCatalogHandler` GET.


mcoughlin (approved):
Nice to see the various bugs caught along the way with this!

antoine-le-calloch (approved):
LGTM. I made a few quick changes to fix old and new issues.

PR #6418, opened 2026-07-25
Pydantic-typed request bodies for earthquake endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `EarthquakeHandler` POST and `EarthquakeMeasurementHandler` POST/PATCH to the `parse_body` pattern.

- `EarthquakePostBody` types the polymorphic ingest body: either a QuakeML `xml` string or the explicit `(event_id, date, latitude, longitude, depth, magnitude)` set; the either/or check stays in the handler with its original message. The old docstring pointed at the unrelated `EarthquakeEventNoID` marshmallow schema and documented a `gcnevent_id` response the handler never returned — the typed response is `{id}` (`str | int | null`: the xml cancel path returns the row PK or null).
- `EarthquakeMeasurementBody` (`rfamp: float`, `lockloss: int`, both optional, at-least-one check in the handler) is shared by measurement POST and PATCH. Their docstrings claimed `data: EarthquakeMeasured` responses but both return bare success — docs fixed to match.
- `EarthquakePredictionHandler` POST takes no body and is unchanged.
- Client sweep: `NewEarthquake.tsx` submits exactly the dict fields; nothing submits measurements from the frontend; `pdl_service` calls `post_earthquake_from_xml` directly as a function (signature unchanged); `db_demo.yaml` posts `{xml}`. Test payloads in `test_earthquake.py` all fit the models.


leonitousconforti:

PR #6419, opened 2026-07-25
Pydantic-typed request bodies for reminder endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `ReminderHandler` POST and PATCH to the `parse_body` pattern.

- `ReminderPostBody` documents the full wire contract the handler actually reads — the old docstring listed only `text` and `group_ids`, omitting `next_reminder`, `reminder_delay`, `number_of_reminders`, and `user_ids`. `next_reminder` is now a required string: the old code called `arrow.get(None)` when it was missing, which raises and surfaced as a 500 (the frontend form always sends it). The old documented response claimed `reminder_id`; the handler returns `reminder_ids` (one per user) — the typed response matches the handler.
- PATCH replaces the marshmallow `partial=True` merge with `ReminderPatchBody` and explicit field application (text, next_reminder, reminder_delay, number_of_reminders). The old merge also replaced the reminder's groups on every PATCH (defaulting to the requester's accessible groups, since the handler always injected `data["groups"]`) — that behavior is kept explicitly with a comment. The per-type `__schema__()` selection and the discarded `users` fetch go away; the `user_ids` accessibility check stays.
- Client sweep: the only client is `Reminders.tsx`, which POSTs `{text, next_reminder, number_of_reminders, reminder_delay}` (its `*_unit`/`lead_time` helper keys are deleted before submit); nothing in the frontend, tests, or services calls reminder PATCH.

### DIFF
--- a/skyportal/handlers/api/reminder.py
+++ b/skyportal/handlers/api/reminder.py
@@ -1,5 +1,5 @@
 import arrow
-from marshmallow.exceptions import ValidationError
+from pydantic import BaseModel, ConfigDict, Field
 
 from baselayer.app.access import auth_or_token, permissions
 from baselayer.app.custom_exceptions import AccessError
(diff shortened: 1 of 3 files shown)


leonitousconforti:

PR #6420, opened 2026-07-25
Pydantic-typed request bodies for user endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `UserHandler` POST and PATCH to the `parse_body` pattern.

- `UserPostBody` carries the documented fields; `groupIDsAndAdmin` is typed as a list of `[groupID, admin]` pairs (`list[tuple[int, bool]]`, rendered as a tuple type in the generated `api.ts`). Phone/email validation with exact messages stays in the handler; the `affiliations` isinstance check is replaced by `list[str]`. POST gains a typed `{id}` response. (The old docstring's `enum: {list(role_acls)}` was a never-interpolated f-string leftover.)
- `UserPatchBody` replaces a free-form `setattr` loop that applied arbitrary body keys to the User row. It also fixes a latent bug: the loop excluded `expiration_date` but the wire key is `expirationDate`, so the raw string was re-set as a stray instance attribute after the parsed value. The model now carries `expirationDate` (present-vs-null via `model_fields_set`, so explicit null/empty still clears the date — UserManagement uses that) plus the fields the UI actually patches via `UpdateUserParameter`: `username`, `first_name`, `last_name`, `contact_email`.
- Client sweep: `patchUser` callers send `{expirationDate}` (UserManagement) or a single `{[parameter]: value}` from the four above (UpdateUserParameter); no frontend POSTs to this route (users are created via invitations); `db_demo.yaml` posts `{username, roles, first_name, last_name}`. Test payloads fit the models.

### DIFF
--- a/skyportal/handlers/api/user.py
+++ b/skyportal/handlers/api/user.py
@@ -5,6 +5,7 @@ import phonenumbers
 import sqlalchemy as sa
 from email_validator import EmailNotValidError, validate_email
 from phonenumbers.phonenumberutil import NumberParseException
+from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy import func
 from sqlalchemy.orm import selectinload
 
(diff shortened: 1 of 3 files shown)


leonitousconforti:

PR #6421, opened 2026-07-27
Pydantic-typed request bodies for public release endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `PublicReleaseHandler` POST and PATCH to the `parse_body` pattern.

- `PublicReleasePostBody`/`PublicReleasePatchBody` keep all fields optional and leave the handler checks in place ("No data provided" via `model_fields_set`, "Name is required", "Link name is required", "Specify at least one group", link-name regex) because the tests assert those exact messages; pydantic adds shape validation and `extra="forbid"`. The models also document `auto_publish_enabled`, which the old docstring omitted, and POST gains a typed `{id}` response.
- The PATCH model deliberately has no `link_name`: the old handler silently ignored it (link names are immutable). With `extra="forbid"` that's now an explicit 400 — the test that verified silent-ignore now asserts the rejection.
- PATCH quirk preserved with a comment: the old cache-invalidation check read `is_visible` with a `False` default while the assignment used `True`, so an omitted `is_visible` cleared the page cache but left the release visible; replicated via `model_fields_set`.
- Client fix in `public_release.ts`: the edit form seeds its state from the full fetched release, so PATCH sent `id`/`link_name`/`created_at`/etc. The duck now picks only the accepted keys (same pattern as the telescopes duck fix in #6412).

### DIFF
--- a/skyportal/handlers/api/public_pages/public_release.py
+++ b/skyportal/handlers/api/public_pages/public_release.py
@@ -1,7 +1,9 @@
 import operator  # noqa: F401
 import re
+from typing import Any
 
 import sqlalchemy as sa
+from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy import func
 from sqlalchemy.orm import selectinload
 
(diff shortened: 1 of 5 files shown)


leonitousconforti:

PR #6423, opened 2026-07-27
Pydantic-typed request bodies for sources confirmed in GCN endpoints

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `SourcesConfirmedInGCNHandler` POST and PATCH to the `parse_body` pattern.

- `SourcesConfirmedInGCNPostBody` requires `source_id`, `localization_name`, `localization_cumprob`, `start_date`, `end_date` (matching what the marshmallow `Validator` enforced); `confirmed` stays optional-nullable — the old docstring claimed it required, but the handler never validated it and stores it raw. The docstring also mistyped `localization_cumprob` as a string and omitted `explanation`/`notes`, which the handler reads. Both endpoints gain a typed `{id}` response.
- The marshmallow `Validator` stays in place for what pydantic doesn't do here: converting the `dateobs` path parameter and `start_date`/`end_date` to naive UTC datetimes (`UTCTZnaiveDateTime`). Its truthy/falsy `confirmed` field was never actually applied to the body (the handlers read `confirmed` raw, bypassing the Validator), so typing it `bool` changes nothing on the wire.
- Client sweep: the only client is `ConfirmSourceInGCN.tsx`, whose POST/PATCH payloads match the models exactly; test payloads (`test_sources_confirmed_in_gcn.py`, `test_gcn.py`, `test_gcnevents.py`) all fit. DELETE has no body and is unchanged.


leonitousconforti:

PR #6424, opened 2026-07-27
Pydantic-typed request body for notification test endpoint

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `NotificationTestHandler` POST to the `parse_body` pattern.

- `NotificationTestPostBody` carries `notification_type` and `user_id`, both optional so the handler's exact "notification_type must be email or SMS" check (and the admin check for other users' IDs) keeps firing with its original message.
- The docstring's 200 response claimed `SingleTaxonomy` — a copy-paste leftover; the handler returns bare success. Fixed to `Success`. (The route is `/api/internal/`, so it's excluded from the OpenAPI spec — no `api.ts`/`routeSchemaMap.ts` changes; verified no drift.)
- Client sweep: the only callers are the email/SMS test buttons in `UpdateProfileForm.tsx`, which send `{notification_type}`; no test or service calls this route.

### DIFF
--- a/skyportal/handlers/api/internal/notifications_test.py
+++ b/skyportal/handlers/api/internal/notifications_test.py
@@ -1,4 +1,5 @@
 import sqlalchemy as sa
+from pydantic import BaseModel, ConfigDict, Field
 from twilio.rest import Client as TwilioClient
 
 from baselayer.app.access import auth_or_token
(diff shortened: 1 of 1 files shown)


leonitousconforti:

PR #6425, opened 2026-07-27
Pydantic-typed request body for frontend error log endpoint

leonitousconforti:
Continues the typed-endpoint migration (#6382 and follow-ups), converting `LogHandler` POST (the frontend crash reporter) to the `parse_body` pattern.

- `LogPostBody` requires `error` and makes `stack` optional: the old code indexed `data['stack']` directly, so a crash report without a stack raised a KeyError and the report was lost as a 500; now it still logs (with `None` in place of the stack).
- Client sweep: the only caller is `ErrorBoundary.tsx`, which sends `{error, stack}`; no tests or services hit this route. The route is `/api/internal/`, so it's excluded from the OpenAPI spec — no `api.ts`/`routeSchemaMap.ts` changes (verified no drift).

### DIFF
--- a/skyportal/handlers/api/internal/log.py
+++ b/skyportal/handlers/api/internal/log.py
@@ -2,6 +2,7 @@ import datetime
 import os
 
 import tornado.web
+from pydantic import BaseModel, ConfigDict, Field
 
 from baselayer.app.access import permissions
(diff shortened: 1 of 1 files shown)


leonitousconforti:

PR #6429, opened 2026-07-27
Delete pg bouncer tests

leonitousconforti:
Not in our critical path, I don't think we should be testing this?


mcoughlin:
@leonitousconforti We are pooling in production... why wouldn't we test?

PR #6459, opened 2026-08-01
[DNM] node 26

mcoughlin:
This PR bumps to Node 26.


leonitousconforti (approved):

camilledzt:
Hello, can we wait before merging it ? ICARE is still with node version 22.19.0. I will contact IJLCAB infrastructure team to update it. Thank you

mcoughlin:
@camilledzt Of course, set to DNM and let us know when an update would be ok.

camilledzt:
Thank you very much, I'll keep you updated

mcoughlin:
@camilledzt any news?

camilledzt:
not yet, I'm waiting for the meeting with Michel

camilledzt:
Hello @mcoughlin, following up on this discussion, I checked this morning with Michel, Gerard and Julien and it looks like Node version 26 is currently active, which is not an LTS release. Would it be possible to switch to version 24 instead, since it's the current LTS version? It would be easier for them to install, since it's available directly from the registry, and we'd avoid needing a custom installation process. Tell me what you think of that and if it's ok for you :)

mcoughlin:
We ended up doing 24 instead.

PR #6464, opened 2026-08-03
mark dark mode more usable

mcoughlin:
This PR tries to make dark mode a bit nicer.


antoine-le-calloch (approved):
I made a major cleanup of the profile page, improved the color mode switcher, and updated the secondary color that was being used as the primary color in the Dark/Blue themes (e.g. on the left panel and buttons) to make it look much better. There are quite a few changes, so I'll leave the merge up to you once you're happy with the PR.

There's still more to improve, like the `Shifts` page and a few buttons/text elements that sometimes appear dark on a dark background, as well as the table headers, which are still too dark. But since these aren't blocking issues, and it's already much better than before haha, let's merge this and continue improving it as we go.

PR #6466, opened 2026-08-03
Pydantic-typed request bodies for remaining API endpoints

leonitousconforti:
Consolidates the open typed-endpoint PRs plus a large batch of newly-migrated handlers into one PR, continuing the migration piloted in #6382. Every covered endpoint validates its JSON body against a module-level pydantic model (`extra="forbid"`) via `self.parse_body(Model)`, and the model drives the OpenAPI docs.

### Newly migrated in this PR
- **Small/self-contained:** sharing_service (PUT/POST/DELETE across 5 files), moving_object, mpc, webhook (kept `extra="allow"` — external analysis callback), summary_query, scan_report, scan_report_item, public_source_page, spatial_catalog, skymap_trigger, group (6 methods), shift (4), and the two deferred optional-body DELETEs (user_obj_list, tag).
- **High fan-out:** source + candidate (shared `_ObjBody` base modeling the runtime `Obj` schema; marshmallow load kept for deep validation), gcn (13 methods), followup_request (10; instrument `payload` stays a permissive dict), observation + observation_plan (incl. the dual single/batch plan POST), and the photometry family — photometry, photometric_series, and photometry_validation (the last replicates the marshmallow truthy/falsy `validated`/`rejected` bool contract via a pydantic validator).


mcoughlin:
One question is whether all of these Body contents should be put in their own file (kind of like how api.ts does) rather than in the individual handlers.

mcoughlin:
@leonitousconforti also I think the failing tests are related.

PR #6467, opened 2026-08-03
scanning report improvements

mcoughlin:
This PR makes a few scanning report improvements based on feedback from LIONS folks.

@kryanhinds

### DIFF
--- a/skyportal/handlers/api/candidate/scan_report.py
+++ b/skyportal/handlers/api/candidate/scan_report.py
@@ -39,7 +39,11 @@ async def get_sources_by_objs_in_range(
     try:
         # Coerce the ISO strings to datetimes so the DB compares timestamps, not text.
         def to_datetime(value):
-            return arrow.get(value).to("utc").datetime.replace(tzinfo=None) if value else None
+            return (
+                arrow.get(value).to("utc").datetime.replace(tzinfo=None)
+                if value
+                else None
+            )
 
         result = await session.execute(
             sa.select(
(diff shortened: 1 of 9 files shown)


skyportal/handlers/api/candidate/scan_report.py
thomasculino: This method created 2 queries/candidate, but it's now up to 6 queries/candidate, which might become a problem when there are a lot of candidates. How about we combine all the objs into one list so each data type (Obj, Source, FollowupRequest, ClassicalAssignment, Comment, Photometry) is fetched in a single batched query instead of once per candidate? That would bring it down to a constant 6 queries total, regardless of candidate count.

thomasculino (commented):
Just one comment about creating the scan report. The rest looks good

skyportal/handlers/api/candidate/scan_report.py
mcoughlin: Fixed!

skyportal/handlers/api/candidate/scan_report.py
```diff
@@ -33,6 +37,10 @@ async def get_sources_by_objs_in_range(
     list of tuples (obj_id, source_ids)
     """
     try:
+        # Coerce the ISO strings to datetimes so the DB compares timestamps, not text.
+        def to_datetime(value):
+            return arrow.get(value).datetime.replace(tzinfo=None) if value else None
```
thomasculino:
```suggestion
            return arrow.get(value).to("utc").datetime.replace(tzinfo=None) if value else None
```

thomasculino: We don't take timezone into consideration here so we will have weird results when applying the time range on the frontend.

thomasculino (approved):
LGTM

thomasculino:
Actually it needs linting but otherwise I'm good with the code

PR #6468, opened 2026-08-03
Pydantic-typed request bodies for remaining API endpoints (part 2)

leonitousconforti:
Completes the body-endpoint migration piloted in #6382 and continued in #6466, covering the handlers #6466 left out. Each endpoint validates its JSON body against a module-level pydantic model (`extra="forbid"`) via `self.parse_body(Model)`, which also drives the OpenAPI docs.

### Migrated
- **spectrum** (POST/PUT + ASCII-parse/upload + bulk + synthetic photometry), **annotation_services** (5 cross-service query POSTs), **survey_efficiency** (default-plan POST).
- **comment** (POST/PUT with a nested `CommentAttachment` model), **classification** (5 methods incl. the polymorphic single-vs-batch POST).
- **instrument**, **mmadetector** (3 handler classes), **allocation**, **catalog_services** (4).
- **analysis** (6 methods) and **broker** (9 — one more than first estimated).

### Approach / semantics
- Old wire semantics preserved exactly: `data.get(k, default)` → field default; `"k" in data` → `body.model_fields_set`; hard-indexed keys → required. Handler-level checks with test-asserted messages and cross-field checks stay in the handler; existing marshmallow `.load()` calls are kept and fed `body.model_dump(exclude_unset=True)`.
- **Null-in-list lesson (from #6466's CI fix):** array/vector fields whose elements can be null are typed `list[T | None]` (spectrum `wavelengths`/`fluxes`/`errors`), and pass-through blobs stay permissive (`dict[str, Any]`/`list`) so nothing the old code accepted is now rejected — the downstream validators still enforce the real per-element rules with their exact messages.
- Alias handling for leading-underscore wire keys (`_altdata`, `_authinfo`) via `Field(alias=...)` + `by_alias=True` dumps.

### Client fixes (required by `extra="forbid"` — latent bugs)
- Comment ducks (`source.ts`, `gcnEvent.ts`, `earthquake.ts`, `shifts.ts`) were spreading routing keys (`obj_id`/`spectrum_id`/`gcnevent_id`/`earthquake_id`/`shift_id`) into the request body — rebuilt each to the `{text, group_ids, attachment}` allowlist.
- `AnalysisForm.tsx` dropped two always-undefined fields (`photometry_filters`/`photometry_instruments`) that the server ignored.
- Test/demo payload cleanups (`test_comments.py`, `test_comments_on_spectrum.py`, `test_newsfeed.py`, `data/db_demo.yaml`) removing keys the old code silently ignored.

### Bug fixes surfaced
- survey_efficiency: a 403 path referenced a nonexistent key (`data['default_observation_plan_id']`) that would have raised KeyError.

### Not included
- **galaxy** — deferred. `#6257` ("update galaxy catalog ingestion") landed on main and substantially rewrote `galaxy.py` (handlers renamed/restructured around a new `_resolve_fits_source` helper) after this branch was cut, so this branch takes main's version unchanged; galaxy will be migrated cleanly against the new structure in a follow-up.
- **facility_listener** — intentionally skipped: a test posts a superfluous field and asserts a 500 from the downstream jsonschema validator, which `extra="forbid"` would turn into a 400.

### Verification
`ruff` (pinned 0.15.17) check + format, spec build, `api.ts`/`routeSchemaMap.ts` regenerated, eslint, prettier, and `bun run typecheck` all pass.


stefanv:
Is there a way that we can avoid duplication of the spec, i.e. have one source of truth from which we generate whatever other sources we need?

Perhaps I am misreading the patch, but it looks like there are now two definitions: a Pedantic one in code, and a YAML docstring for the API reference.

leonitousconforti:
> Is there a way that we can avoid duplication of the spec, i.e. have one source of truth from which we generate whatever other sources we need?
> 
> Perhaps I am misreading the patch, but it looks like there are now two definitions: a Pedantic one in code, and a YAML docstring for the API reference.

This becomes the single source of truth, we have deleted the response section from the docstring. This enforces the shape of the body at runtime, and provides nice intelisense + reusability for other endpoints (like mcp endpoints in the future)

PR #6470, opened 2026-08-04
Use new_session for database work outside a web request

### LINKED ISSUE #4670: DBSession is used only to get bind
We can add a more intuitive mechanism for grabbing the db engine, and get rid of all DBSession usage.

The API now is to use `self.Session` (verified commit, per handler) or just pure sqlalchemy Sessions, one per job thread.

mcoughlin:
Move to https://github.com/skyportal/skyportal/pull/6744.


stefanv:
Is this convenience function perhaps well suited to including in baselayer?

PR #6471, opened 2026-08-04
fix missing annotations

mcoughlin:
This PR fixes an issue with missing annotations coming from BOOM.

### DIFF
--- a/skyportal/broker_apis/_save.py
+++ b/skyportal/broker_apis/_save.py
@@ -200,6 +200,7 @@ async def _ingest_object(
     filter_ids=None,
     passing_alert_id=None,
     cutouts=None,
+    annotations_by_filter_id=None,
 ):
     """Create/refresh an Obj for ``data`` (a standard alert object), attach it to
     groups (as a Source) and/or filters (as a Candidate), and ingest its
@@ -220,9 +221,10 @@ async def _ingest_object(
         the provider supplies them, rendered into science/template/diff thumbnails.
     """
     import sqlalchemy as sa
+    from sqlalchemy.orm import joinedload
 
     from ..handlers.api.photometry import add_external_photometry
-    from ..models import Candidate, Group, Instrument, Obj, Source
+    from ..models import Annotation, Candidate, Filter, Group, Instrument, Obj, Source
     from ..utils.naive_datetime import utcnow_naive
 
     object_id = data["objectId"]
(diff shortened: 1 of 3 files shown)


skyportal/broker_apis/_save.py
```diff
@@ -200,6 +200,7 @@ async def _ingest_object(
     filter_ids=None,
     passing_alert_id=None,
     cutouts=None,
+    annotations_by_filter_id=None,
```
stefanv, line 203: Private function, but would still update the docstring.

skyportal/broker_apis/_save.py
```diff
@@ -296,6 +298,44 @@ async def _ingest_object(
...
+    # the filter's group so its scanners can see it. Upsert on the (obj_id, origin)
+    # unique key so a re-pass refreshes the data instead of raising.
+    annotated = {
+        fid: annotations_by_filter_id[fid]
+        for fid in filter_ids or []
+        if (annotations_by_filter_id or {}).get(fid)
+    }
+    if annotated:
+        filters = (
+            await session.scalars(
```
stefanv, line 327: Another pattern we see all over the place that we may want to provide a shorthan on our session for.

skyportal/broker_apis/_save.py
```diff
@@ -296,6 +298,44 @@ async def _ingest_object(
...
+                sa.select(Filter)
+                .options(joinedload(Filter.group))
+                .where(Filter.id.in_(annotated))
+            )
+        ).all()
+        for filt in filters:
+            group = filt.group
+            group_name = (group.nickname or group.name) if group else None
```
stefanv, line 337: Is this the standard way for accessing group name? If so, may wish to update the model to provide a canonical way of doing so.

skyportal/broker_apis/_save.py
```diff
@@ -296,6 +298,44 @@ async def _ingest_object(
     if created or filter_ids:
         await session.flush()
 
+    # Attach each passing filter's annotation (origin "{group}:{filter}"), scoped to
+    # the filter's group so its scanners can see it. Upsert on the (obj_id, origin)
+    # unique key so a re-pass refreshes the data instead of raising.
+    annotated = {
```
stefanv, line 320: Should this logic, for retrieving annotations, live in the models, or is it bespoke for this usage right here?

stefanv:
General thought: because Claude is not lazy to type, it may cause duplication; we should update the agent instructions to prefer refactoring into models whenever a query pattern occurs in multiple places.

skyportal/broker_apis/_save.py
mcoughlin: fixed

skyportal/broker_apis/_save.py
mcoughlin: custom here, matches how we manage these broker annotations for a while

skyportal/broker_apis/_save.py
mcoughlin: this is bespoke for annotations

skyportal/broker_apis/_save.py
mcoughlin: yeah I will open the baselayer pr

thomasculino (approved):

antoine-le-calloch (changes requested):
- json.loads can return a list, number or string that would land in Annotation.data, which is expected to be an object.

- This SELECT-then-INSERT races on the (obj_id, origin) unique constraint: two ingests of the same obj+filter both see existing is None and the second one raises an IntegrityError at the commit below, which drops the whole alert (candidate + photometry + thumbnails). Could we use on_conflict_do_update here.

- origin is user-writable through the annotation API, so an annotation created by a user with the same "{group}:{filter}" origin gets silently overwritten here, with no permission check.

antoine-le-calloch (approved):

PR #6475, opened 2026-08-05
fix onboarding username collisions and duplicate memberships

### LINKED ISSUE #3752: Move onboarding to session context managers
The onboarding.py file contains several uses of `DBSession()` that could be upgraded to a context manager (i.e., `with` statements). 
This needs to be done carefully since it causes the session to close and re-open for each function call, and in some cases the user object that is passed as an argument becomes detached.

mcoughlin:
This PR fixes onboarding username collisions and duplicate memberships

Closes https://github.com/skyportal/skyportal/issues/3752


thomasculino (approved):
LGTM

Issue #6076, opened 2026-04-20
Prevent photometry points with fluxerr equal to zero.


antoine-le-calloch:
We want to prevent input photometry points from having a flux error equal to zero.

PR #6476, opened 2026-08-05
Enh: Improving photometry workflow

camilledzt:
The goal of this PR is to improve the data management of photometry data. What has been done:
* `limiting_mag` is now optional in the photometry endpoint as sometimes in the GCN we don't have the associated limiting mag with the detections (so we don't have to put arbitrary values). Now, if mag AND limiting mag are null, then an error is thrown.
* Bugfix: when a user wants to change the MJD of a non-detection, it was throwing an error due to missing mag, this is fixed in the frontend.
* Photometry table
  * Make `origin` and `altdata` editable in the photometry table, by changing schema.py + component
  * Remove the duplicate icon of the download icon.
  * Pagination was broken, 250/500 was not working anymore, now we have 50/100/All options. Other options are not in the community edition.
  * Add the `t-t0` option in the download form if t0 is not null in the source page.
* issue #6076: when a user is uploading data, `fluxerr` should be strictly positive now (schema.py)


mcoughlin (approved):
LGTM

PR #6482, opened 2026-08-05
config level default share group

### LINKED ISSUE #4587: Default Share groups: should it be sitewide and not just the user?
Please fill out the relevant sections below; remove those that are unused.

**Feature Summary**
Most users of Fritz seem to complain that when uploading any data product without specifying groups to share the data with, the default is to only share to the current user, and not to either all the groups they have access to or sitewide. They assume that it does already, so they get confused.

**Alternative Solutions**
Should we consider making this configurable? Just like we can configure in `config.yaml` what permissions are given to new users, we could maybe configure what groups are used to share data when none are specified (which would be of course not used when groups to share the data with are specified).

**Implementation details**
It could be set in the `config.yaml` configuration file, grabbed in the handlers where needed, and maybe also sent to the frontend using the config handler to set the default groups there as well (though if we don't the handler will still take care of it, but it might be better to avoid confusion the other way around).

**Additional context**
I know the whole data access part of Fritz is a requirement and is important, but we should really consider defaulting to sharing data and not sharing only when specified, rather than the opposite (which is what we are doing now).

mcoughlin:
This PR makes it possible to set the default share group at the config level.

Closes https://github.com/skyportal/skyportal/issues/4587

### DIFF
--- a/config.yaml.defaults
+++ b/config.yaml.defaults
@@ -729,6 +729,12 @@ misc:
   max_seconds_to_sleep_reminders_service: 60
   max_seconds_to_sleep_recurring_apis_service: 60
   public_group_name: "Sitewide Group"
+  # When a data product (photometry, spectra, ...) is uploaded without specifying
+  # any groups, share it with the sitewide public group (public_group_name) in
+  # addition to the uploader's single-user group (which is always included). When
+  # false (default) an unspecified upload is shared only with the uploader. Groups
+  # given explicitly on the upload are always honored as-is.
+  share_data_with_public_group_by_default: false
   # External survey thumbnails fetched (outbound HTTP) when a source is saved.
   # Set to [] to disable, e.g. on instances without external network egress.
   external_thumbnail_types:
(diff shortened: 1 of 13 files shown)


thomasculino (changes requested):
If the flag to default share to Sitewide Group is on:
1. the photometry saved through BOOM would be visible by everyone
2. Users won't be able to publish private (only to themselves) photometry

To fix 1. we could add a parameter,  `true` by default, to `get_group_ids` and `add_external_photometry` in `photometry.py` and set it to `false` in `_save.py`. That would prevent sharing all ingested alerts with sitewide group.

mcoughlin:
Users won't be able to publish private (only to themselves) photometry
 -> They can still specify their own group... but maybe we could make that easier.

thomasculino:
> Users won't be able to publish private (only to themselves) photometry -> They can still specify their own group... but maybe we could make that easier.

I thought your own group was excluded from the list

mcoughlin:
@thomasculino Ah on the front-end! Ok let me add the user group to that form.

thomasculino (approved):

PR #6483, opened 2026-08-05
fix query results not showing when running a test query

thomasculino:
The results of the queries run on the filter page were not being displayed correctly.
image


mcoughlin (approved):
LGTM

PR #6484, opened 2026-08-06
fritz filter translation

mcoughlin:
This PR continues the fritz filter translation.


thomasculino (changes requested):
- We need to move the boom handlers as well
- We should align `FilterPlugins.tsx` with what we have on Fritz minus kowaslki. It will be easier to maintain if other brokers want to have a custom page as well.

thomasculino:
Let me commit changes and I'll ask for your review

thomasculino:
Ok I got really confused here but everything is okay in the end. I defaulted the filters which haven't been saved yet to the boom broker so the UI is available even for new filters. I also took advantage of that pr to fix a small bug causing the groups inside the `AutocompleteFields.tsx` to collapse when the list was too big.

thomasculino (approved):
Maybe you should take a quick look to what i changed but it looks good now

PR #6486, opened 2026-08-06
fix the P200-NGPS starlist

### LINKED ISSUE #6485: P200-NGPS starlist: MAGNITUDE blank when the newest positive-flux photometry point is not a detection
## Summary

For `starlist_type="P200-NGPS"`, the `MAGNITUDE` and `MAGFILTER` columns come back
empty for some sources even though the source has plenty of photometry — including
recent detections in NGPS-supported filters.

The magnitude is selected in
[`skyportal/handlers/api/source.py`](https://github.com/skyportal/skyportal/blob/main/skyportal/handlers/api/source.py)
(`ObjOffsetsHandler`, the `facility in ["P200-NGPS"]` branch):

```python
latest_photometry = (
    await session.scalars(
        Photometry.select(session.user_or_token)
        .where(
            Photometry.obj_id == obj_id,
            Photometry.flux.isnot(None),
            Photometry.flux > 0,
            Photometry.fluxerr.isnot(None),
            Photometry.filter.in_(ALL_NGPS_SNCOSMO_BANDS),
        )
        .order_by(Photometry.mjd.desc())
    )
).first()
if latest_photometry is not None:
    source_mag = latest_photometry.mag
    source_magfilter = latest_photometry.filter
```

**`flux > 0` is not the same as a detection.** The query takes the newest point with
positive flux, but `Photometry.mag` returns `None` when the point is below the
detection threshold. So when the most recent positive-flux point is a non-detection,
`source_mag` is `None`.

`skyportal/utils/offset.py` then blanks *both* fields:

```python
if not source_mag or not source_magfilter:
    mag, magfilter = "", ""
```

The source can have many detections; only the *newest positive-flux row* matters, and
if that one is sub-threshold the target is exported with no brightness at all.

## Reproduction

```
GET /api/sources/ZTF26ablhgaq/offsets?facility=P200-NGPS&observing_run_id=<run>&num_offset_stars=3
```

Parsing `data.starlist_info[].str` as CSV:

| NAME | MAGNITUDE | MAGFILTER |
|---|---|---|
| `ZTF26ablhgaq` | `''` | `''` |
| `ZTF26ablhgaq_o1` | `14.47` | `R` |

The **offset star is given a magnitude and the science target is not**, in the same
response. `GET /api/sources/ZTF26ablhgaq/photometry` returns 121 points with **12
detections**, the most recent ~4 days before the export:

```
mjd 61254.455  ztfg = 18.70 +- 0.07
mjd 61254.461  ztfi = 19.10 +- 0.13
```

Second case: `ZTF19acymzwg`, 2454 points and 163 detections, also exported blank.

Observed 2026-08-05 on Fritz production; 2 of 55 targets in one P200/NGPS run.

## What it is not

* Not a filter-mapping problem — `ztfg`/`ztfr`/`ztfi` are all in
  `NGPS_TARGET_BANDS_TO_SNCOSMO`.
* Not "no photometry" — 12 and 163 detections respectively.
* Not stale data — one of the two has a detection 4 days old.
* Not a general failure — 53 of 55 targets in the same run were fine, and offset stars
  always are.

## Impact

`MAGNITUDE` is the only brightness the P200/NGPS target list carries, and the exposure
time calculator cannot solve without it. In our pipeline a single unsolvable row
aborts the whole OTM run, so one blank target costs the night the exposure times of
every other target too. Recovering it client-side means pulling the full photometry
series (a few thousand points) to obtain one number.

## Suggested fix

Restrict the query to actual detections rather than positive flux — or skip rows whose
`.mag` is `None` and take the newest that has one:

```python
.where(
    ...
    Photometry.snr >= PHOT_DETECTION_THRESHOLD,   # or equivalent
)
```

A fallback to the last *detection* in any NGPS band would also fix it, and would match
what the offset-star path already manages to do.

## Aside

`GET /api/observing_run/{id}` carries no magnitude on `assignments[].obj`, and
`GET /api/sources/{id}?includePhotometryStats=true` returned no `photstats`. A
`last_detected_mag` on the assignment payload would let facility clients price a run
from the one call they already make.

mcoughlin:
This PR fixes the P200-NGPS starlist.

Closes https://github.com/skyportal/skyportal/issues/6485

### DIFF
--- a/skyportal/handlers/api/source.py
+++ b/skyportal/handlers/api/source.py
@@ -116,6 +116,7 @@ DEFAULT_SOURCES_PER_PAGE = 100
 MAX_SOURCES_PER_PAGE = 500
 MAX_NUM_DAYS_USING_LOCALIZATION = 31 * 12 * 10  # 10 years
 _, cfg = load_env()
+PHOT_DETECTION_THRESHOLD = cfg["misc.photometry_detection_threshold_nsigma"]
 log = make_log("api/source")
 
 MAX_LOCALIZATION_SOURCES = 50000
@@ -2809,20 +2810,30 @@ class SourceOffsetsHandler(BaseHandler):
             priority, comment = 1, "science"
             if facility in ["P200-NGPS"]:
                 # look for the latest photometry point
-                # in the filters supported by NGPS
+                # in the filters supported by NGPS.
+                # `flux > 0` also matches NaN flux (Postgres orders NaN above
+                # every number), and those points have no magnitude to export,
+                # so require one. Prefer a detection, falling back to the
+                # latest point that has a magnitude at all.
+                ngps_photometry = (
+                    Photometry.select(session.user_or_token)
+                    .where(
+                        Photometry.obj_id == obj_id,
+                        Photometry.mag.isnot(None),
+                        Photometry.fluxerr.isnot(None),
+                        Photometry.filter.in_(ALL_NGPS_SNCOSMO_BANDS),
+                    )
+                    .order_by(Photometry.mjd.desc())
+                )
                 latest_photometry = (
                     await session.scalars(
-                        Photometry.select(session.user_or_token)
-                        .where(
-                            Photometry.obj_id == obj_id,
-                            Photometry.flux.isnot(None),
-                            Photometry.flux > 0,
-                            Photometry.fluxerr.isnot(None),
-                            Photometry.filter.in_(ALL_NGPS_SNCOSMO_BANDS),
+                        ngps_photometry.where(
+                            Photometry.snr >= PHOT_DETECTION_THRESHOLD
                         )
-                        .order_by(Photometry.mjd.desc())
                     )
                 ).first()
+                if latest_photometry is None:
+                    latest_photometry = (await session.scalars(ngps_photometry)).first()
                 if latest_photometry is not None:
                     source_mag = latest_photometry.mag
                     source_magfilter = latest_photometry.filter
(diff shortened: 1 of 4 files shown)


skyportal/handlers/api/source.py
```diff
@@ -2812,20 +2810,30 @@ async def get(self, obj_id: str):
...
-                            Photometry.fluxerr.isnot(None),
-                            Photometry.filter.in_(ALL_NGPS_SNCOSMO_BANDS),
+                        ngps_photometry.where(
+                            Photometry.snr >= PHOT_DETECTION_THRESHOLD
                         )
-                        .order_by(Photometry.mjd.desc())
                     )
                 ).first()
+                if latest_photometry is None:
+                    latest_photometry = (await session.scalars(ngps_photometry)).first()
```
thomasculino:
```suggestion
latest_photometry = (
    await session.scalars(
        Photometry.select(session.user_or_token)
        .where(
            Photometry.obj_id == obj_id,
            Photometry.mag.isnot(None),
            Photometry.fluxerr.isnot(None),
            Photometry.filter.in_(ALL_NGPS_SNCOSMO_BANDS),
        )
        .order_by(
            (Photometry.snr > PHOT_DETECTION_THRESHOLD).desc(),
            Photometry.mjd.desc(),
        )
    )
).first()
```

Two things here: 
- the comparison between the `snr` and the `PHOT_DETECTION_THRESHOLD` is usually made by using > not >=
- We can avoid doing 2 queries

thomasculino (commented):
One fix, going from >= to > and one other comment on merging 2 queries into one

thomasculino (approved):

PR #6497, opened 2026-08-07
fix greenlet on observation plan

mcoughlin:
This PR fixes a sentry-identified observation plan async issue.


leonitousconforti (approved):

leonitousconforti:
Looks like base layer might need an update too with the new admin database user change?

PR #6498, opened 2026-08-07
Feat: adding `can_share` right at group level

camilledzt:
This PR is adding `can_share` right at group level. 
The current behavior to share data is:
* users can only share data they owned 
* `System admin` can share all data.

This PR in introducing a new option at the group level that lets non-owners the right to share photometry already in their group to other groups.


mcoughlin (approved):
LGTM

PR #6500, opened 2026-08-07
github action labels

### LINKED ISSUE #5252: GitHub Actions: Run tests based on PR labels
When we are merging something on main, we always want all the tests to run.

However when working on a PR, that is not completely true:
- If one is working on the schema, it makes sense to run the migrations test. Otherwise, it does not seem useful to me.
- When working on a frontend-only PR, there is no need to run anything other than frontend tests. This is also true for PRs that only update frontend dependencies.
- When updating backend dependencies or backend features however, I still think it is useful to run the frontend tests since any backend change could break something there.
- Similarly, we don't even need to start the docs-related Github action when working on anything that is not in the `skyportal/handlers` folder (we could add an API label when we do).

That list might not even be complete, but I think we should really consider conditioning what tests run and when (except when merging on main, where we want everything to be tested maybe?). This would speed up the development process for any feature that is not frontend+backend+db model+docs, which happens very very rarely.

Any thoughts or suggestions? @stefanv @mcoughlin

mcoughlin:
This PR adds logic for running specific unit tests depending on what code is touched.

Closes https://github.com/skyportal/skyportal/issues/5252

### DIFF
--- a/.github/workflows/build-and-deploy-docs.yaml
+++ b/.github/workflows/build-and-deploy-docs.yaml
@@ -12,11 +12,17 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  scope:
+    uses: ./.github/workflows/ci_scope.yaml
+
   build-and-deploy-docs:
+    needs: scope
     runs-on: ubuntu-latest
     timeout-minutes: 60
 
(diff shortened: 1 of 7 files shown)


thomasculino (approved):
✅

PR #6505, opened 2026-08-09
generic SearchableSelect component

### LINKED ISSUE #4817: TNS Robots Page - Add Coauthors: searchable list
On the TNS robots page, one can add co-authors for a given robot, which are basically just users of the app.

In production it can be hard to find users out of the 300+ users we have in Fritz for example, and having the list be searchable would be awesome.

In general we need searchable lists for everything, and a solid generic searchable list component would be super useful. Students worked on that in the past, but none of their PRs where merged (because incomplete or incorrect).

The idea here would be to look into the current frontend, identify the existing searchable lists, and come up with a generic component that can be reused everywhere. Ideally we want 2 versions of that component (or one version that does both), one that takes a full list of items to search from, and one that is able to search for the items using the API. A good example of the latter is the search bar on the top right of the web app.

mcoughlin:
This PR makes a generic SearchableSelect component.

Closes https://github.com/skyportal/skyportal/issues/4817


antoine-le-calloch (approved):
LGTM. I made some fix + clean up the code.

antoine-le-calloch:
@mcoughlin You should quickly review my changes, and then you can merge if you're okay with them.

PR #6506, opened 2026-08-09
correct standards for proper motion

mcoughlin:
This PR corrects standards for proper motion.


skyportal/handlers/api/internal/standards.py
```diff
@@ -90,6 +94,9 @@ async def get(self):
         dec_filter_range_str = self.get_query_argument("dec_filter_range", "[-90, 90]")
         ra_filter_range_str = self.get_query_argument("ra_filter_range", "[0, 360]")
         show_first_line = self.get_query_argument("show_first_line", False)
+        obstime = self.get_query_argument("obstime", utcnow_naive().isoformat())
+        if not isinstance(isoparse(obstime), datetime.datetime):
```
thomasculino: `isoparse(obstime)` either returns a `datetime.datetime` value or raises one of `ValueError` or `ParserError`. So the error message will never be displayed if `obstime` is incorrect.

thomasculino (approved):
One comment about the `obstime` error message but otherwise it's good.

skyportal/handlers/api/internal/standards.py
mcoughlin: Great catch. Was an issue in a couple of places.

PR #6509, opened 2026-08-10
Adjust permissions for telescope access control.

antoine-le-calloch:
Fix `Insufficient permissions for operation "User X create Telescope {X}".`


mcoughlin (approved):
LGTM

PR #6511, opened 2026-08-10
Pydantic-typed request bodies for the remaining API endpoints

leonitousconforti:
Converts the remaining body-bearing API endpoints to pydantic-typed request bodies, following the pattern already used elsewhere in the codebase: one model per request body drives both runtime validation and the generated OpenAPI schema.

- Each migrated handler gets a `<Thing><Verb>Body` model with `extra="forbid"`, parsed via `self.parse_body(...)` after auth.
- Field documentation moves out of the docstring `requestBody:` blocks into `Field(description=...)`.
- Frontend ducks and the test suite were swept for payload keys the models reject. That surfaced a few latent bugs, mostly ducks sending a URL path parameter in the body where the handler had always ignored it.
- `static/js/types/api.ts` and `static/js/types/routeSchemaMap.ts` regenerated.


leonitousconforti:
I kinda like having the body models next to the route handlers, that way you don't have to jump between files. But that is just my preference. I am open to putting them all in one file if that is others preference's

mcoughlin (approved):
I am happy to see this go in. @antoine-le-calloch any concerns?

leonitousconforti:
This pr got combined with everything finally, so here was my response to stephen https://github.com/skyportal/skyportal/pull/6468#issuecomment-5244770001

antoine-le-calloch (approved):
LGTM

PR #6514, opened 2026-08-11
fix: cannot find group X after group deletion

camilledzt:
The goal of the PR is to fix the error "Cannot fing Group with id XXX" after a group deletion by adding a flag in the frontend.


mcoughlin (approved):
LGTM

PR #6518, opened 2026-08-12
Add customizable NGPS assignment fields when requesting follow up

thomasculino:
This PR adds customizable NGPS assignment fields when requesting follow up on the source page. It also fixes a bug where react-jsonschema-form would convert decimals written a . to a , depending on the browser locale, making it impossible for the user to input a decimal value.


static/js/components/followup_request/FollowupRequestForm.tsx
```diff
@@ -474,6 +580,8 @@ const FollowupRequestForm = ({
             schema={schema as any}
             validator={validator}
             uiSchema={uiSchema}
+            templates={followupTemplates}
+            fields={followupFields}
```
mcoughlin: EditFollowupRequestDialog.tsx:146 and NewDefaultFollowupRequest.tsx:216 could both use this same addition I think.

mcoughlin (approved):
Just one comment to consider adding your comma fix in a few other spots.

static/js/components/followup_request/FollowupRequestForm.tsx
thomasculino: Done! I extracted the fix because i think we might have to apply that fix in a lot more places but I didn't want to do everything in this PR

mcoughlin: Yes sir. Merge when you are happy.

PR #6519, opened 2026-08-12
Show LSST alias and last detection in scanning report

thomasculino:
Adds one column to scanning report and the last detection by survey.

### DIFF
--- a/skyportal/handlers/api/candidate/scan_report_item.py
+++ b/skyportal/handlers/api/candidate/scan_report_item.py
@@ -37,6 +37,14 @@ def _survey_of(bandpass):
     return bandpass
 
 
+def _lsst_alias(aliases):
+    """Return the first LSST alias for this object, if any."""
+    for alias in aliases or []:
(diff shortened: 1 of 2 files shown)


mcoughlin:
@thomasculino I think the test failure might be related.

thomasculino:
> @thomasculino I think the test failure might be related.

@mcoughlin I tried rerunning it and I get a failure but a different error. In both cases, it looks like it's more due to a timeout. The `test_scan_reports.py` is the only test in this folder that uses the ObjFactory, which puts a load on the 2 shared vCPUs that the CI uses. It looks related to this comment:

`# The CI runner has only 2 vCPUs shared by Postgres, the app processes,`
`# the microservices and Firefox, so heavy (but progressing) requests can`
`# take >60s and return a 504. Give the upstream more headroom.`

The 504 is exactly the first error I had.

mcoughlin (approved):
You mentioned also wanting to show all aliases, so once that is done, feel free to merge!

PR #6525, opened 2026-08-12
public candidate page

mcoughlin:
This PR enables public candidate page (using the source workflow).


antoine-le-calloch:
@mcoughlin Why do you want a dedicated flow for candidates instead of saving them as sources and publishing the source?

mcoughlin:
@antoine-le-calloch I am thinking this could be the pathway to Zooniverse engagement (see the other PR). We can discuss this week on what this workflow might look like.

antoine-le-calloch:
Need to be public page (real React components) that don't use the "public page" feature.

PR #6526, opened 2026-08-12
Scanning report additional fields

thomasculino:
Adding more fields to the scanning reports


mcoughlin:
@thomasculino fix the model test. Otherwise should be good.

PR #6528, opened 2026-08-13
Show associated objs and previous mag in scanning reports

thomasculino:
Additional column previous mag in scanning reports.
Adding band and mjd to detections.
Use associated objects instead of aliases to show matches with other surveys and not names given by users. If those associated objects have aliases, these will be shown.


mcoughlin (approved):
LGTM

PR #6533, opened 2026-08-14
Add walkthrough for filter builder

thomasculino:
This pull request introduces a new interactive tour for the filter builder page, improves the reliability of tour triggers, and adds data attributes to key UI elements to support the new tour. The main changes are grouped as follows:

**New filter builder page tour:**

* Added a comprehensive "filter" tour to `PAGE_TOURS` in `PageTours.ts`, with steps covering blocks, AND/OR logic, conditions, adding elements, saving, and post-save actions.
* Updated the "Getting Started" checklist to include the new filter builder tour, linking users to their first filter for an interactive walkthrough. 

**Tour trigger logic and reliability improvements:**

* Refactored `PageTourProvider.tsx` to move the navigation state-clearing logic so that the tour trigger is only cleared after the tour actually starts or gives up, ensuring the tour runs reliably even if the destination page loads slowly. 
* Updated `BoomFilterPlugins.tsx` to detect the filter tour trigger from `location.state` at mount time, so the filter builder is ready before the tour starts polling for targets. 

**Data attributes for tour targeting:**

* Added `data-testid` attributes to key filter builder UI elements (`tour-filter-blocks`, `tour-filter-operator`, `tour-filter-condition`, `tour-filter-add`, `tour-filter-save`) to anchor tour steps accurately. 

**Minor UI and code improvements:**

* Improved formatting of the read-only pipeline alert in `FilterBuilderContent.tsx` for better readability.

These changes collectively make the filter builder more approachable for new users and ensure the tour system works reliably across navigation and slow-loading pages.

### DIFF
--- a/static/js/components/PageTourProvider.tsx
+++ b/static/js/components/PageTourProvider.tsx
@@ -38,19 +38,20 @@ const PageTourProvider = () => {
           (step) => !step.acl || permissions.includes(step.acl),
         ),
       );
-      navigate(location.pathname + location.search + location.hash, {
-        replace: true,
-        state: {},
-      });
     }
     // permissions is derived from the cached profile (stable ref); intentionally
     // not a dep — we filter with whatever ACLs are loaded when the tour launches.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [requested, location.pathname, location.search, location.hash, navigate]);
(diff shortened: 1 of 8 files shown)


mcoughlin (approved):
LGTM

PR #6537, opened 2026-08-14
Drop unrecognized filter rows in ATLAS forced photometry ingestion

thomasculino:
ATLAS FP results should only contain filter codes "c"/"o", but a stray value (e.g. "t") in the response was passed through unmapped and made the whole photometry commit fail validation. Drop such rows and log them instead of aborting the entire batch.

This was made assuming we don't want to keep filter code 't', maybe we do?

### DIFF
--- a/skyportal/facility_apis/atlas.py
+++ b/skyportal/facility_apis/atlas.py
@@ -175,6 +175,19 @@ def commit_photometry(
         cyan = df["filter"] == "c"
         orange = df["filter"] == "o"
 
+        # ATLAS forced photometry should only report "c" or "o"; drop any
+        # other rows instead of failing the whole batch on ingestion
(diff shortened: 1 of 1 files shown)


mcoughlin:
They didn't add a new passband did they? :) @thomasculino

thomasculino:
> They didn't add a new passband did they? :) @thomasculino

@mcoughlin it doesn't seem like it. Their doc says `F: the ATLAS filter (typically c for cyan, or o for orange)`. The error come from us receiving a 't' filter.

mcoughlin:
Alright let's see what we get then @thomasculino !

thomasculino:
@mcoughlin good to merge?

mcoughlin (approved):
LGTM

PR #6542, opened 2026-08-15
finish anyOf → enum transition

### LINKED ISSUE #5784: Use enum instead of anyOf in forms
### Issue:
Many forms currently use anyOf instead of enum.
This causes “functions nested too deeply” errors on Firefox (while Chrome is more permissive), because anyOf builds a much deeper validation tree than enum (and is therefore slower).

### Proposal:
- Replace anyOf with enum where applicable.
- Move enumNames into the uiSchema via "ui:enumNames", since putting it directly in the schema is deprecated.

This issue follows the discussion in [PR #5783](https://github.com/skyportal/skyportal/pull/5783)

mcoughlin:
This PR finishes the anyOf → enum transition.

Closes https://github.com/skyportal/skyportal/issues/5784


antoine-le-calloch (approved):
LGTM.

PR #6544, opened 2026-08-15
Add one-time feature announcement popups.

antoine-le-calloch:
- Add feature announcement provider and styles for guided tours.
- FeatureAnnouncementProvider plays the first announcement a user hasn't seen, then records it in preferences.featureAnnouncementsSeen
- Extract the shared Joyride look into useTourStyles
- Clean up some comments.


mcoughlin (approved):
Great addition.

PR #6549, opened 2026-08-17
Make tests reliable and faster

antoine-le-calloch:
The crossmatch tests were slow and intermittently failing.

- Disable astropy IERS auto-download in tests.
- Poll (with a timeout) instead of sleeping a fixed time while waiting for
  background tasks.
- Delete the crossmatch test event after each test so `run_cycle` keeps
  walking a single event.


mcoughlin (approved):
LGTM

PR #6551, opened 2026-08-17
Type path parameters from handler signatures

leonitousconforti:
Removes `install_path_param_validation`, which wrapped every handler method from `BaseHandler.__init_subclass__` to coerce path arguments.

Tornado sets `path_args` before calling the handler method, and baselayer's `prepare()` already normalizes the captured strings (including stripping the leading slash of `(/[0-9]+)` patterns), so the wrapper was redundant. `BaseHandler.prepare` now coerces `path_args` in place, which also stops the wrapper from shadowing handler signatures.

Coercion goes through pydantic `TypeAdapter`, matching `parse_body`, so `Literal[...]` and `Annotated[T, Field(...)]` work instead of being skipped like the old `resolve_cast` unions. The error response is unchanged.

## Path parameters in the spec

The same annotations now generate the `in: path` entries, so the ~390 hand-written docstring blocks are gone. Generating them from the rendered path keeps the documented parameters and the URL template in lockstep, which fixed 84 pre-existing inconsistencies:

- 51 path parameters that appeared in a URL but were documented nowhere
- 33 documented under names the URL does not have, e.g. `observation_plan_id` on a route that captures `observation_plan_request_id`, and one entry whose `name` was `"Instrument ID"`
- 4 typed `dateobs`, 1 typed `str`, 1 with `description`/`required` misnested inside `schema`

Descriptions worth keeping moved to `Annotated[T, Field(description=...)]`, aliased at module level where several methods share one. 30 parameters the docstrings typed as integers are now annotated as such; these already called `int()` internally, so a bad value returns a 400 instead of a 500.

## Verification

- Deleting the 390 docstring blocks produces a byte-identical spec, since generation ignores hand-written path entries.
- The 580 query parameters are unchanged, and the spec with `parameters` stripped is identical to before, so nothing outside path parameters moved.
- 23 unit tests, 3 of which drive a real tornado request to confirm the coercion lands between URL capture and the method call.
- `bun run typecheck` passes on the regenerated `api.ts`; `routeSchemaMap.ts` is unchanged.

## Follow-ups

- 13 `associated_resource_type` parameters lost an `enum` in the spec. `Literal` would restore it and validate, but the handlers lowercase the value before dispatch, so it would reject case variants that work today. The docstring enums also disagreed with each other and with the code. Worth doing as a separate behavioral change.
- The redundant inline `int()` calls in `broker.py` are now no-ops and could be dropped.

Conflicts with #6416 in `openapi.py`, where both add injection into the path-registration loop.


skyportal/handlers/api/analysis.py
```diff
@@ -2328,11 +2239,6 @@ async def get(self, analysis_service_id: int, default_analysis_id: int):
           description: Retrieve all default analyses
```
mcoughlin, line 2239:
```suggestion
          description: Retrieve all default analyses for the given analysis service
```

mcoughlin:
@leonitousconforti small suggestion.

PR #6553, opened 2026-08-17
Fix object query on broker alerts page

antoine-le-calloch:
- Fix LSST object search on the broker alerts page.
- Infer the survey from the objectId (with a manual selector when a broker has several), match LSST ids stored as int64, and scope LSST alerts by stream instead of programid.
- Fix the fetch of LSST cutouts
- Add tooltip to survey selection for improved user guidance in Broker component.


mcoughlin (approved):
LGTM

PR #6555, opened 2026-08-18
boom survey tooltip

mcoughlin:
This PR fixes the BOOM survey tooltip location (so it doesn't affect the dropdown).


antoine-le-calloch (approved):
LGTM

PR #6557, opened 2026-08-18
Refactor sidebar component for improved drawer behavior and styling.

### LINKED ISSUE #4831: Navbar: only open when clicking the menu button
Instead of expanding the menu after the user has hovered it for more than 500ms, we might want to simply show tooltips with the menu name when hovering an icon, and only expand the menu when the user clicks the "menu" icon in the top left hand corner of the app.

antoine-le-calloch:


mcoughlin (approved):
LGTM

PR #6559, opened 2026-08-18
Fix lost update on profile preferences

antoine-le-calloch:
Fix lost updates when saving profile preferences concurrently. Lock the user row and re-read preferences in the transaction: the JSONB. blob is merged in Python, so racing PATCHes clobbered each other.


mcoughlin (approved):
LGTM

PR #6560, opened 2026-08-18
Add links formatting

### LINKED ISSUE #4822: Source Page - New Comment: auto-format links before submitting.
A little bit of a tricky issue (easy to implement, but might be hard to figure out all of the edge-cases that might make it fail I guess)!

I noticed a lot of users adding links to their comments, but these are not clickable!

We use a markdown text renderer for the comments, so for links to be clickable, they need to be markdown compliant. The idea here would be that on submit of a comment, if there are any link(s) in the comment being posted, we detect that, put them in markdown format so that they can be clickable. Either just do it and post the comment, or if we make any modification, show a popup window to the user telling them what changed so they can validate/reject this modified version of their comment before we post it to the DB.

Low priority, but a cool life quality improvement.

antoine-le-calloch:
Enhance comment rendering with markdown links.


mcoughlin (approved):
LGTM

PR #6562, opened 2026-08-18
Improve user management page

antoine-le-calloch:
- Clean up user management and source statistics pages, and make the users table client-side.
- Display the edit/delete buttons on hover only to save space.
- Fix table column display. Fix some frontend bugs.


mcoughlin (approved):
Much needed. Thanks @antoine-le-calloch.

PR #6563, opened 2026-08-18
Dogfood skyportal-py in the API tests

leonitousconforti:
We have a python api client that has strict models defined for the api responses: https://github.com/leonitousconforti/skyportal-py

All the api inputs have models in here now (path params, query arguments, and bodies) and this api client defines response models

My idea is to start dogfooding this throughout the tests, so that we can know if the hand-derived response models are correct. My end vision is maybe some sort of monorepo? So this skyportal client would live in this repo too. Right now playing around with [this structure](https://github.com/skyportal/skyportal/tree/monorepo/projects) but maybe something like [namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) is more what we want. Can't do that right now though because there is the root `skyportal` folder

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,6 +131,7 @@ dev = [
     "pytest-rerunfailures>=14.0",
     "playwright==1.58.0",
     "pytest-playwright==0.8.0",
+    "skyportal-py==0.2.0",
 ]
 docs = [
     "sphinx>=9.1.0; python_version >= '3.12'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ dev = [
     "pytest-rerunfailures>=14.0",
     "playwright==1.58.0",
     "pytest-playwright==0.8.0",
-    "skyportal-py==0.2.0",
+    "skyportal-py==0.3.0",
 ]
 docs = [
     "sphinx>=9.1.0; python_version >= '3.12'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,7 +131,7 @@ dev = [
     "pytest-rerunfailures>=14.0",
     "playwright==1.58.0",
     "pytest-playwright==0.8.0",
-    "skyportal-py==0.3.0",
+    "skyportal-py==0.3.1",
 ]
 docs = [
     "sphinx>=9.1.0; python_version >= '3.12'",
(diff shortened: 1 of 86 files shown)


mcoughlin:
@leonitousconforti Can we move skyportal-py into the skyportal org please?

leonitousconforti:
> @leonitousconforti Can we move skyportal-py into the skyportal org please?

Of course, Ideally I think it lives in this repo since it shares models. Right now the models are duplicated so that is the next step I think

mcoughlin:
Personally I don't like needing to update the client every time I want to update an endpoint, although I understand why this has value. Maybe the client can have its own set of tests that emulate these? What do you think @antoine-le-calloch ?

leonitousconforti:
> Personally I don't like needing to update the client every time I want to update an endpoint, although I understand why this has value. Maybe the client can have its own set of tests that emulate these? What do you think @antoine-le-calloch ?

I don't think that anything in the client on our end would really need to be updated though when you update an endpoint. There isn't much maintenance burden for us. There would be like three or four folders

* skyportal-models (shared pydantic models for query params, bodies, and repsponses)
* skyportal (the endpoints)
* skyportal-py (the python client)
* mcp-tools (maybe a folder for mcp tools)

In my head, all of these must agree on what they accept and return. If you want to change what an endpoint accepts, you start my changing it's body model, then you go to the endpoint and update the logic there. Since the client imports the same models, it already has the new body shape and is correctly typed.

Yes, if you are a skyportal-py consumer and we add a new required body field to an endpoint, you are required to update. We made a breaking change to the endpoint. If you want backwards compatibility where the consumer doesn't have to update the skyportal client, then we need to make the field optional, that's something we need to think about.

Just my feelings

leonitousconforti:
If we are worried about desync between people installing different versions of the client package and not knowing which version they need for their server instance, then maybe we can use something like [CalVer](https://peps.python.org/pep-2026/) and only deploy the tags.

We have been tagging weekly on Thursdays, so like `2026-08-20`. In the deployment repos, the submodule gets pinned by a tag instead of by a branch and we only use tags for that going forwards. The PyPi package can be published as `2026.08.20`, so you know that if your server instance is on version `2026-08-20` you need client at least `2026.08.20`. When breaking changes are introduced we communicate that but consumers don't need to update until their instance updates and they would know what version they need to go to.

Issue #5576, opened 2025-04-26
Add Phot/Spec distinction in scanning report item info


antoine-le-calloch:
Make a distinction between PHOT / Spec requests info In the scanning report item. This one is trickier as it will require exceptions based on the API class and for those that support both, exceptions based on the payload.

mcoughlin:
Tagging @thomasculino since he has been working closely with LIONS folks on this lately.

PR #6566, opened 2026-08-19
add groups saved to and fix other surveys detections

thomasculino:
Add a columns groups which shows groups the obj is currently saved to.
Edits the way associated objs are displayed in the detections column.
Adds the last detection in the detections column as well. 

Make a distinction between PHOT / Spec requests info In the scanning report item. (#5576)

### DIFF
--- a/skyportal/handlers/api/candidate/scan_report_item.py
+++ b/skyportal/handlers/api/candidate/scan_report_item.py
@@ -40,6 +40,39 @@ log = make_log("api/scan_report_item")
 # named "desi_*" (e.g. "desi_dr1").
 DESI_ORIGIN_PREFIX = "desi_"
 
+# FollowUpAPI classes (Instrument.api_classname) whose requests are always
+# photometric or always spectroscopic -- instrument.type is often too coarse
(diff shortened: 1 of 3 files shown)


skyportal/handlers/api/candidate/scan_report_item.py
```diff
@@ -40,6 +40,39 @@
 # named "desi_*" (e.g. "desi_dr1").
 DESI_ORIGIN_PREFIX = "desi_"
 
+# FollowUpAPI classes (Instrument.api_classname) whose requests are always
```
mcoughlin, line 43: Maybe this classification can become part of the API classes themselves?

mcoughlin (commented):
Suggestion to move the API type classification to the API itself

skyportal/handlers/api/candidate/scan_report_item.py
thomasculino: Yes, I fixed that. One thing to keep in mind is that some classes cannot have a proper class type because the check needs to be done at payload level.

mcoughlin (approved):
LGTM

PR #6567, opened 2026-08-19
Fix ztf thumbnail dedup

thomasculino:
This pull request addresses an issue with duplicated thumbnail entries for the same object/type/survey combination, particularly affecting ZTF data. It ensures that only one thumbnail per object/type/survey exists, both by enforcing a unique constraint at the database level and by updating the application logic to update existing thumbnails instead of creating duplicates. The changes also include a migration to backfill and deduplicate existing data.

**Database migration and data cleanup:**

* Adds an Alembic migration (`02dff366befe_dedupe_thumbnail_survey.py`) that:
  - Backfills the `survey` field for legacy thumbnail rows where a sibling row with the correct survey exists.
  - Deletes duplicate thumbnail rows, keeping only the most recent per `(obj_id, type, survey)`.
  - Enforces a unique constraint on the `thumbnails` table for `(obj_id, type, survey)`.

**Database schema enforcement:**

* Adds a `UniqueConstraint` to the `Thumbnail` model to ensure only one thumbnail per object/type/survey combination can exist going forward.

**Application logic updates:**

* Updates the thumbnail upload handler to:
  - Check for an existing thumbnail with the same `obj_id`, `type`, and `survey` before inserting.
  - Update the existing row (including file and metadata) if it exists, rather than inserting a duplicate.
  - Ensures the `is_grayscale` field is updated on overwrite.
* Imports the `image_is_grayscale` utility to support the above logic.


antoine-le-calloch (changes requested):
- Legacy `survey=NULL` rows still produce duplicate tiles. The migration only backfills a NULL row that already has a non-NULL sibling, so a legacy object not yet reprocessed keeps `survey=NULL`; the next ZTF alert then inserts a second row and the source page renders both (`ThumbnailList.tsx:106` keys on `survey ?? ""`). That is exactly the population this PR targets. `post_thumbnail` should adopt the NULL-survey row when a survey is supplied.

- `raise StatementError(f"...")` raises `TypeError`. SQLAlchemy 2.0 requires `message, statement, params, orig`, and `IntegrityError` subclasses `StatementError`, so a unique violation at commit lands in this `except` and surfaces as "missing 3 required positional arguments" instead of the DB message.

- SELECT-then-INSERT is not atomic. On a concurrent ingest the losing insert raises `IntegrityError` at commit, which deactivates the transaction and makes the final `session.commit()` in `_save.py:399` fail, losing the whole object ingest rather than just the thumbnail. `pg_insert(...).on_conflict_do_update(...)` is already the pattern here (`_save.py:365`).

- The migration locks the table. A non-concurrent `ADD CONSTRAINT` plus an unbatched full-table self-join DELETE, while `3d9f7a1c2b45` already builds indexes on this same table with `CREATE INDEX CONCURRENTLY` inside an `autocommit_block()`. Batch the DELETE and use `CREATE UNIQUE INDEX CONCURRENTLY` + `ADD CONSTRAINT ... USING INDEX`.

- The survey value is not normalized. `recreate_thumbnails_from_broker` (`thumbnail.py:628`) passes `data.get("survey")` verbatim, and BOOM emits title-case names ("Ztf"), which is why `_record_survey` uppercases. `Ztf` and `ZTF` both satisfy the constraint and render as two tiles pointing at the same PNG.

- The migration can abort the deploy. The dedup DELETE is restricted to `('new','ref','sub')` but the constraint covers every type, so a pre-existing duplicate on another type with a non-NULL survey (the POST body accepts any `ttype`, e.g. `new_gz`) makes `create_unique_constraint` fail and rolls the migration back. Widen the DELETE or use a partial unique index.

- Two surveys share one PNG on disk. `file_uri`/`public_url` are keyed on `obj_id + type` while the constraint is keyed on `obj_id + type + survey`, so the second image overwrites the first and deleting either row fires the `after_delete` listener that removes the file the surviving row points at.

antoine-le-calloch (approved):
LGTM

PR #6571, opened 2026-08-19
Fix query param parsing for autosaveGroupIds and hasTNSname

leonitousconforti:
Two query param parsing bugs found while auditing the API surface:

- `autosaveGroupIds` (candidates scanner autosave) was handed to `post_source_async` as the raw comma-separated string. Its per-element `int()` then ran on single characters, so a multi-id list always failed and a multi-digit group id silently saved the source to the wrong groups (e.g. `"12"` saved to groups 1 and 2). It is now parsed into a list of ints at the read site, with a 400 on invalid values.
- `hasTNSname` / `hasNoTNSname` are read with a `None` default, which skips the bool coercion `get_query_argument` applies when the default is a bool, so any non-empty value, including `"false"`, enabled the filter. Both are now normalized with `str_to_bool`.

Includes regression tests for both: a two-group autosave via the candidates endpoint, and `hasTNSname=false` no longer filtering.


mcoughlin (approved):
LGTM (need to fix merge conflicts).

PR #6572, opened 2026-08-19
Fix boom mongo dialog rawpipeline

thomasculino:
For filters generated to support the kowalski to boom transition, the pipeline was not properly sent to the MongoQueryDialog, which would result in bad request when trying to run the filter. 

Also changed the default time range on the dialog from `today -> tomorrow `to `yesterday -> today`


mcoughlin (approved):
LGTM

PR #6579, opened 2026-08-19
Avoid 2h ci timeout

antoine-le-calloch:
Fail fast on a stuck apt in CI with retries and a 10 min step timeout, instead of hitting the 2h job timeout.


mcoughlin (approved):
YES PLEASE!

thomasculino (approved):
We really need this

PR #6580, opened 2026-08-20
optional terms of service popup

mcoughlin:
This PR enables an optional terms of service popup.


antoine-le-calloch:
@mcoughlin can you add screenshot to the description of the PR pls.

mcoughlin:
Screenshot 2026-08-19 at 10 50 11 PM

antoine-le-calloch (changes requested):
Need to change the code to add backend verification to ensure that only users who have accepted the popup can access the web app information. The implementation approach needs to be carefully thought through and discussed, as this is not trivial, especially regarding the security checks for the handlers. We should make sure users aren't blocked in case of backend issues and avoid adding too many back-and-forth requests with the backend before page load.

mcoughlin:
Thanks @antoine-le-calloch . Are you planning on trying this part or should I?

antoine-le-calloch:
> Thanks @antoine-le-calloch . Are you planning on trying this part or should I?

I will, I just added this so I can remember because I will do it after the more important thing I have to do.

antoine-le-calloch:
@mcoughlin You can check my new changes and merge if its good for you.

antoine-le-calloch (approved):
LGTM

PR #6585, opened 2026-08-22
tarot responses as Latin-1

mcoughlin:
This PR decodes TAROT responses as Latin-1 so an accent can't fail a submission.


camilledzt (approved):
LGTM, thank you for adding that

PR #6591, opened 2026-08-24
Public profil page

### LINKED ISSUE #4826: Public User Profile
Currently, each user has access to it's own profile where he can see all his info and preferences, and edit them.

It could be very useful for normal users (and not just admins through the admin user management page) to have access to some of the public information about a user, like first and last name, affiliations, email, phone number... For the email and the phone number we could add an option on an individual profile page where a user can make them publicly visible or not (private by default).

On that same public user profile page, we could show very useful information about user activity, like their recent comments, sources they've saved recently (limited to the groups that the the user looking at another user's profile have in common), ...l
Almost like a simplified version of what we have on the dashboard, but focusing on one user.

The reason why I would like this to exist is for book keeping, and to facilitate the job of new users. We have a bunch of experienced users that perform 80% of the actions that all users take, and being able to look at what they are doing on the data you have access too would be a great way to learn and understand what one can do with Fritz.

Maybe we can even imagine a chat, basically a variant of the comments where we have sender_user_id and receiver_user_id + the user comment and attachment. This way users of Fritz could talk to each other about operations without having to exit the app.

Users could even link papers and documents in a newly added "user description" column on users, which would render as markdown, effectively allowing users to customize their user profile pretty extensively without us having to develop a ton of individual features!

antoine-le-calloch:
- Add a public profile page at `/user/:id`, backed by a new `GET /api/user/{id}/profile` endpoint, replacing the old `UserInfo` page.
- Let users pick, field by field, what their profile shares through a visibility toggle next to each field in the settings; the public page reflects the change immediately.
- Merge the profile settings into a single editable card that mirrors the public profile layout, with navigation between the two pages, replacing `UpdateProfileForm` and the `PublicProfilePreferences` panel.
- Return only the fields a user chose to share from `GET /api/user` and `GET /api/user/{id}` for requesters without the `Manage users` ACL, and restrict the `email` filter to that ACL.
- Link avatars and usernames to their public profile in comments, the news feed and top savers.
- Save the profile before sending a test notification instead of submitting the form in parallel.
- Restyle the profile dropdown and the theme toggle.
- Make MUI links follow the theme instead of the global gray anchor style, and paint the theme swatches with their paper background.
- Accept a `Z` UTC suffix in the git log timestamp parser.
- Announce the public profile on the dashboard and in the profile settings.

Screenshot 2026-08-26 at 11 54 01 AM
Screenshot 2026-08-26 at 11 54 18 AM
Screenshot 2026-08-26 at 11 54 27 AM
Screenshot 2026-08-26 at 11 54 38 AM
Screenshot 2026-08-26 at 11 54 53 AM


mcoughlin (approved):
LGTM

PR #6601, opened 2026-08-26
Add stateless MCP endpoint (protocol 2026-07-28) for sources, photometry and spectra

leonitousconforti:
Adds a Model Context Protocol endpoint at `/mcp` so AI assistants can read and write sources, photometry and spectra through the existing REST API and its permission checks.

- Implements MCP revision 2026-07-28 only (Streamable HTTP, per-request `_meta`, `Mcp-Method`/`Mcp-Name`/`MCP-Protocol-Version` header validation, `server/discover`, `resultType`, `ttlMs`/`cacheScope`, spec error codes). No sessions, SSE, or `initialize` handshake, so requests can be served by any app process.
- Six tools: `get_sources`, `post_source`, `get_photometry`, `post_photometry`, `get_spectra`, `post_spectrum`. Each maps its arguments to the corresponding REST call, made with the caller's own token. Arguments are validated against the tool's input schema with `jsonschema`.
- Auth reuses SkyPortal API tokens (`Authorization: Bearer <token>` or `token <token>`). `Origin` is validated against the configured app URL.
- No new dependencies. The endpoint is excluded from the OpenAPI spec.
- Tests in `skyportal/tests/api_tests/test_mcp.py` (protocol rules, header/meta validation, end-to-end round trip); docs in `doc/mcp.md`.


mcoughlin:
@leonitousconforti skyportal-mcp manages things like analyze_light_curve and get_tns_summary effectively do multiple API calls and do things with the returned data. Can we make it such that the async function may use the REST helper zero or more times and returns its own content? This should make it generic enough, I would think.

mcoughlin:
@leonitousconforti another thought: as we add these capabilities, I wonder if putting the logic into shared functions e.g. fetch_sources(session, user_or_token, filters) which can be called by both SourceHandler.get and the MCP tool, will be the way to go so that we do not have duplication (while still being able to do more complicated analyses that we want the MCP server to handle).

leonitousconforti:
A tool is now `async def tool(handler, args)`. It can call `handler.api(method, path, query=..., body=...)` zero or more times (each call runs with the caller's token and returns the endpoint's `data`, raising `APIError` on failure) and returns its own content: a JSON value (sent as text and `structuredContent`) or a plain string. Raising `ToolError` reports a tool execution error. So something like `analyze_light_curve` is just a function that fetches photometry, computes, and returns a summary.

leonitousconforti:
Added `analyze_light_curve` as a first composite tool on the new contract. It fetches the source's photometry once (`format=mag`) and reports, per band: first/last/peak detections, rise and fade times and rates, status (`rising`, `fading`, or `complete` once fainter than the peak by `baseline_threshold`, default 0.3 mag), duration, pre-peak brightening events and RMS, and the last upper limit before the first detection. Optional `filters` restricts the bands; bands with fewer than two detections are listed in `skipped_filters`. Returns JSON plus a one-line summary per band. The per-band analysis is a pure function with unit tests, so it can move to a shared helper alongside the handler logic later.

Writing that test also caught that `post_photometry`'s schema rejected `null` in `mag`/`magerr`/`flux`/`fluxerr` lists, which is how non-detections are expressed; fixed.

doc/mcp.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Methods
+
+| Method            | Purpose                                                   |
+| ----------------- | --------------------------------------------------------- |
+| `server/discover` | Supported protocol versions, capabilities, server identity |
+| `tools/list`      | Tool definitions with input schemas                       |
+| `tools/call`      | Invoke a tool                                             |
+
+## Tools
```
mcoughlin: Maybe these can be auto-produced as part of our API spec or something? Maintaining it seems like it is going to be a pain.

mcoughlin (commented):
This looks great. One question about how we will try to keep the docs up-to-date.

mcoughlin (approved):
LGTM

PR #6604, opened 2026-08-26
Seed the undetected photometric series fixture to stop the test failing ~3% of the time.

antoine-le-calloch:


mcoughlin (approved):
LGTM

skyportal/tests/conftest.py
```diff
@@ -2502,7 +2502,7 @@ def photometric_series_undetected(
     user, public_source, public_group, public_group2, ztf_camera, phot_series_maker
 ):
     df = phot_series_maker(number=100, use_mags=False, format="pandas")
-    df["flux"] = np.random.normal(-50, 50, 100)
+    df["flux"] = np.random.default_rng(1).normal(-50, 50, 100)
```
stefanv, line 2505: Modern NumPy API is:

```
rng = np.random.default_rng(1)
rng.normal(...)
```

antoine-le-calloch: Both forms use the modern Generator API. Binding it to a variable is useful when generating multiple values, while for a single draw both approaches are equivalent.

PR #6615, opened 2026-08-31
Improve test workflow

antoine-le-calloch:
1. Move follow-up request browser tests into the frontend suite and remove accordion clicks now that it defaults to expanded.
2. Fix CI test collection: a significant number of tests were not running at all due to a mismatch between the test folder structure and the CI job paths. Move those tests into folders actually covered by CI and separate backend tests from the frontend suite.
3. Refactor test/CI configuration, remove unused fixtures, and add/update test subsets and parameters.
4. Improve galaxy and GCN test reliability and performance by handling ingestion delays and reducing retries.
5. Fix test isolation and disabled-button issues found by CI.
6. Fix recurring API handling for invalid `call_delay` and exhausted retries.
7. Fix thumbnail tests to count only thumbnails created by the test.


mcoughlin (approved):
LGTM

PR #6619, opened 2026-08-31
diff filter versions

mcoughlin:
This PR enables seeing the difference between filter versions.


leonitousconforti (approved):
Cool 😎

This is awesome

PR #6621, opened 2026-08-31
Add the shared pydantic response models

leonitousconforti:
Adds `projects/api-models`, a `skyportal-py-models` distribution describing the payload of each API endpoint with a pydantic model, imported from the [skyportal-py](https://github.com/skyportal/skyportal-py) client and packaged so the server and any client can describe a payload with the same class. This is the last of the pydantic models to add.

Nothing uses them yet. No handler, docstring, generated type or spec changes. The only edits outside the new package are the workspace member, the dependency and the lock entry that make it build, so `static/js/types/api.ts` and the OpenAPI spec are untouched.

### Mutually-referential models

Models that reference each other across resource modules are defined together in `_cyclic.py` and re-exported from the module they belong to, so they are still imported as `skyportal_py_models.<resource>`. That removes the import cycle that would otherwise force 14 relationships to be `dict[str, Any]`.

`PhotometryPointResponse` is the one model that allows extra fields: `GET /api/sources/{obj_id}/photometry` returns photometric series rows alongside individual points, and a series row carries whatever columns the uploaded data file had.


mcoughlin (changes requested):
Looks great @leonitousconforti 

Reviewing with Claude, then should be good to go in.

Worth fixing before merge

photometry.py:16 — PhotometryValidationResponse collides with an existing, differently-shaped model of the same name in skyportal/handlers/api/photometry_validation.py:58 (that one is just {id}). It is the only name shared between this package and the handler-local models, so it will silently shadow whichever is imported second once adoption starts. Cheapest fix now, while nothing imports the package.

uv.lock:2474 — the lock also bumps m4opt 9ca530e → d02e893 and drops platform markers from secretstorage's dependencies, beyond what the PR describes. In fairness, main's lock was already out of sync with its own pyproject pin, so the bump is arguably a fix, but it belongs in its own commit.

Cosmetic

_cyclic.py:517 — garbled comment: two truncated half-sentences spliced together, repeated verbatim at 537-538.
_cyclic.py:246 — the AllocationResponse docstring inverts the allocation_users shape. telescope.py:275-277 substitutes plain users exactly as the allocation endpoints do, so the AllocationUserResponse union arm is never produced by any handler.
shifts.py:36 — ShiftCommentAuthorResponse redeclares gravatar_url, which UserResponse already defines identically. The subclass is a no-op despite its docstring claiming otherwise.
sources.py:1 — __all__ appears in only 21 of 58 modules, and the largest ones (sources, spectra, photometry, candidates, shifts, brokers) omit it. import * therefore leaks models belonging to other resources, and the package has no consistent export rule.

leonitousconforti:
Thanks for the review, all six points addressed (and rebased onto current main):

- `PhotometryValidationResponse` renamed to `PhotometryPointValidationResponse`, so it no longer collides with the handler-local model in `photometry_validation.py`.
- The m4opt lock bump is now its own commit (main's lock was out of sync with the pyproject pin from #6759). The secretstorage marker churn is gone: the markers are kept as they are on main, and `uv lock --check` still passes, so the lock diff in the main commit is now purely the skyportal-py-models entries.
- Fixed the spliced comment in `_cyclic.py` (both copies).
- `AllocationResponse.allocation_users` is now `list[UserResponse]` and the docstring says so; `AllocationUserResponse` is deleted since no handler ever produces the join-row shape.
- `ShiftCommentAuthorResponse` removed; comment authors are plain `UserResponse`, which already has `gravatar_url`.
- Every resource module now defines `__all__` listing its own models (plus its `_cyclic` re-exports), so `import *` no longer leaks cross-resource names.

mcoughlin:
@leonitousconforti is the idea that a future PR will look to share these between the API client and the handlers?

leonitousconforti:
> @leonitousconforti is the idea that a future PR will look to share these between the API client and the handlers?

yup

mcoughlin (approved):
LGTM

PR #6672, opened 2026-09-01
Move the remaining pydantic models into the shared package

leonitousconforti:
Stacked on #6621, which adds the package.

Completes `skyportal-py-models` so it holds every pydantic model describing the API, in two commits.

**Add the request payload models** brings across the 75 payload models from the same skyportal-py release that the response import left behind, into the resource module each belongs to. A `GroupPost` is the body of `POST /api/groups`, where `GroupResponse` is what that endpoint answers with. Purely additive.

**Move the handlers' models** relocates the 341 models the handlers defined themselves, the bodies and query parameters they accept and some payloads they answer with, and points the handlers at the package. 359 definitions move in total: the models plus the literal constants they read, so a page-size default stays beside the field that uses it. `ALLOWED_MAGSYSTEMS` and `MMA_DETECTOR_TYPES` are spelled out instead, since the package cannot import skyportal.

Five models stay in the handlers because they read something the package has no access to:

- `GaiaQueryBody` takes a field default from the config
- `PhotometryValidationPostBody` uses a validator bound to a handler helper
- `SourceOffsetsGetQuery`, `SourceFinderGetQuery` and `UnsourcedFinderGetQuery` enumerate the facilities in `utils.offset`

### Verification

The second commit only relocates definitions, so the generated OpenAPI spec is **identical byte for byte** before and after, which is how the move was checked; `static/js/types/api.ts` is untouched for the same reason. The app imports with all 265 routes, the package still imports on its own with no skyportal or baselayer dependency, and the 275 imports the move orphaned were pruned after diffing against the base so nothing pre-existing was touched.


mcoughlin (approved):
LGTM

PR #6676, opened 2026-09-01
Fix swifttools bump

antoine-le-calloch:
Route Swift ToO submissions through the configured host again. swifttools 4.x posts the request itself and defaults to the production API, so dropping API_URL would have sent test submissions to the real observatory.
The test server route goes away too: with no route it answers 500, swifttools
degrades to `Rejected`, and the test passes with no network at all.


mcoughlin (approved):
LGTM

PR #6680, opened 2026-09-02
Fix reminder

camilledzt:
This PR is fixing reminders. Before, the "Reminder" permission was required but since it's not listed on the file `model_util.py`, only Super Admin can create reminders. I choose to remove this permission type because I think that anyone should be able to create personal reminders on shift, sources etc.


mcoughlin (approved):
LGTM

PR #6681, opened 2026-09-02
Add the skyportal-py client

leonitousconforti:
Stacked on #6672, which finishes the models package.

Adds `projects/python-client`, the first-party Python client, vendored from the standalone repository now that the models it needs live in the workspace.

One function per endpoint, **553 of them**, each taking an `httpx.Client` and returning the model that endpoint answers with. Every function is also bound as a method on the `SkyPortal` object `create_client` returns, so both styles work:

```python
from skyportal_py import create_client
from skyportal_py.roles import fetch_roles

client = create_client("https://your.instance", token=TOKEN)
roles = fetch_roles(client)     # -> list[RoleResponse]
roles = client.fetch_roles()    # the same function, bound
```

The modules carry no model definitions of their own: each imports from `skyportal-py-models` and re-exports what it uses, so `skyportal_py.groups.GroupResponse` and `skyportal_py_models.groups.GroupResponse` name the same class. Where the vendored client referred to a model by its old name or its old module, the import was repointed at wherever the shared package actually keeps it.

The client is a **dev dependency of the app, not a runtime one**. Nothing in `skyportal` imports it; it is a workspace member so the dev shell can exercise the API the way a caller would, and so it releases from this repository.

It keeps the stricter ruff rule set it had as a standalone repository, which ruff applies from the package's own pyproject rather than inheriting the app's.

`GcnTag` is not re-exported: no endpoint answers with one, which is why the models package does not describe it.

### Verification

The client imports with all 553 functions and every return annotation resolves against the shared package. Its own 8 tests pass (respx, no server needed), and a mocked request round-trips end to end into a `RoleResponse`. On the app side nothing changes: the generated OpenAPI spec is identical, and `skyportal.app_server` imports with all 265 routes.


mcoughlin:
@leonitousconforti Is there no way to auto-generate the client based on the Pydantic models? What about httpx vs requests?

leonitousconforti:
> @leonitousconforti Is there no way to auto-generate the client based on the Pydantic models? What about httpx vs requests?

Generating from the models alone isn't possible because the models only describe payload shapes. They don't know the route, the verb, the path/query parameters, or which model each endpoint answers with; that mapping lives in the handlers (yet). A generator would need the OpenAPI spec as an intermediate, and since the spec is assembled from handler docstrings it isn't complete or accurate enough to generate 553 functions from without hand-correcting the output I feel like.

The design here single-sources the part that actually drifts: every function imports its request/response models from `skyportal-py-models`, so a change to a shape shows up in the client with no regeneration step. What's written by hand is just the URL, verb, and envelope unwrap which is the glue codegen tools tend to get wrong. If we eventually put route metadata on every handler, generating that glue becomes would then become possible.

On httpx vs requests: requests is in maintenance mode while httpx has a near-identical API plus timeouts by default, http2 support, one codebase for sync and async so an async client is possible later without a rewrite, and a first-class mocking story via respx. `httpx.Client` also carries base_url/auth headers, so no wrapper class is needed.

If the maintenance burden feels too high though, I can explore more alternatives

mcoughlin:
@leonitousconforti No this is fine. Can you start working on moving the full app to httpx?

mcoughlin (approved):
LGTM

PR #6687, opened 2026-09-03
Part of improve_config_yaml PR #5988

antoine-le-calloch:
- Fix bad pre-commit configuration to make it running on docker.yaml.
- Override Prettier via .prettierrc so that config.yaml.default is treated as a YAML file.


mcoughlin (approved):
LGTM

PR #6690, opened 2026-09-03
Improve prettier

antoine-le-calloch:
- Make prettier and ruff cover the files they were silently skipping.
- Fix all the skipped files with ruff and prettier.
- Add some manual Cleanup up.


mcoughlin (approved):
LGTM

PR #6693, opened 2026-09-03
Fix docs warning

antoine-le-calloch:
- Fix the 173 Sphinx warnings in the docs build. Make -W take effect, since doc/Makefile was overriding it.
- Collapse over-long column types, since instruments.filters alone was 46% of the diagram width, and show the values it replaces in an expandable panel.
- Switch the graphviz layout to top-to-bottom and drop the 57 identical localizationtiles partitions, which stops the diagram growing by one box every month.
- Prune doc/conf.py from 347 to 85 lines, dropping the 2015 sphinx-quickstart boilerplate plus the latex, man, texinfo and intersphinx config, none of which affected the built HTML.
- Correlate the subquery in GcnEvent.retracted, which matched any event's tags rather than the event's own.
- Fix the entity relationship diagram failing to display on the docs site ("The image 'https://skyportal.io/docs/_images/erd.png' cannot be displayed because it contains errors") by rendering it as SVG instead of a 24 MB PNG that cairo was downscaling.
- Turn the diagram into an interactive viewer: zoom, pan, fullscreen, table search ranked on exact match, and a focus mode that highlights a table's relations and lists them as clickable chips.

**Before:**
Screenshot 2026-09-03 at 3 10 24 PM

**After:**
Screenshot 2026-09-03 at 3 09 49 PM


doc/conf.py
```diff
@@ -1,73 +1,48 @@
...
 
-# The encoding of source files.
-# source_encoding = 'utf-8-sig'
-
-# The master toctree document.
 master_doc = "index"
 
-# General information about the project.
 project = "skyportal"
 copyright = "2020–2023, The SkyPortal Team"
```
mcoughlin: Update the copyright to the current year?

mcoughlin (commented):
Just noting it would be good to make it 2026 in the year.

doc/conf.py
antoine-le-calloch: Done !

PR #6694, opened 2026-09-03
Reset the analysis parameter form when the service changes

mcoughlin:
This PR resets the analysis parameter form when the service changes.


mcoughlin:
Trivial bugfix. Self merging.

PR #6697, opened 2026-09-04
Fix async photometry patch

antoine-le-calloch:
- Persist the `PhotStat` created when patching photometry: it was instantiated but never added to the session, so the recomputed stats were silently dropped.
- Flush the pending deletes before recomputing photometry stats, in the single, per object and bulk delete handlers: the async session has autoflush off, so the stats were rebuilt from a set that still contained the deleted points.
- Guard those two bulk deletes against a missing `PhotStat`: deleting the photometry of an object that had no stats row raised a 500.
- Eager load the instrument and telescope of an object's photometric series, which `get_data_with_extra_columns()` reads: since the handler became async, requesting the photometry of any source having a series returned a 500.
- Wait for `currentUser` before rendering the group page instead of rendering with a null user.
- Bump baselayer to pick up cesium-ml/baselayer#464, which fixes the commit time access check on join tables with a composite primary key.
- Add a test patching the same photometry point twice with a stream, so the `StreamPhotometry` row is loaded back and that composite primary key check is exercised.

### DIFF
--- a/skyportal/handlers/api/photometry.py
+++ b/skyportal/handlers/api/photometry.py
@@ -2674,6 +2674,7 @@ class PhotometryHandler(BaseHandler):
             ).first()
             if phot_stat is None:
                 phot_stat = PhotStat(obj_id=photometry.obj_id)
+                session.add(phot_stat)
 
             # The edit above is still pending, so flush it or the stats are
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM

PR #6705, opened 2026-09-07
Fix invalid sx on the GCN event tabs

antoine-le-calloch:
Remove useless left arrow displayed by scrollable Tabs when there is no overflow

image


mcoughlin (approved):
LGTM

PR #6706, opened 2026-09-07
Fix scoped session leak

antoine-le-calloch:
Release the scoped session after each unit of background work in run_async, the scout consumer and the Hermes consumer.


mcoughlin (approved):
LGTM

PR #6708, opened 2026-09-07
Fix followup request retrieve handler

camilledzt:
The followup request handler was raising this error "expected ORM mapped attribute for loader strategy argument".
FollowupRequest.instrument is a @property delegating to allocation.instrument not an ORM mapped relationship.

+ removing usage of run_sync since get endpoint is async


mcoughlin (approved):
LGTM

PR #6712, opened 2026-09-08
Remove kowalski references

antoine-le-calloch:
Drop the stale Kowalski references from the broker lists and comments


mcoughlin (approved):
RIP

antoine-le-calloch:
> RIP

@mcoughlin Are you really sad about that one 😄

mcoughlin:
Hahaha well a bit ;)

PR #6714, opened 2026-09-08
Remove dead alert hook

### LINKED ISSUE #6713: Nothing processes `watchlist` listings anymore
`services/watch_list/` only does work when `skyportal/handlers/api/alert.py` is overlaid by an extension (`alert_available`). Nothing overlays it anymore, so the service is a loop that sleeps.

The `watchlist` list_name is still accepted by `POST /api/listing` (and requires `params`), documented on the `Listing` model, and referenced by `notification_queue`.

Should we delete that ? (https://github.com/skyportal/skyportal/pull/6714)

antoine-le-calloch:
- Drop the alert.py stub (alert_available, post_alert, get_alerts_by_position)
- Remove the watch_list service and its supervisor entry, which could not do anything without that hook.
- Remove the "new activity on a favorite source" notification, which was only ever sent by the watch list service.
- Stop validating listing params against the "watchlist" list name, as they are now free-form.
- Drop the leftover watchlist entries from the demo data and the test config service list.


mcoughlin (approved):
LGTM

PR #6717, opened 2026-09-08
Add TESS sector footprints as instrument fields

mcoughlin:
This PR adds TESS sector footprints as instrument fields.


leonitousconforti (approved):

leonitousconforti:
API test is broken on main, not here

PR #6718, opened 2026-09-08
Auto refresh source photometry

### LINKED ISSUE #5738: Fritz Alert Page/Handler: Photometry update + forced phot import
**Feature Summary**
Some filtering (e.g TDEs) is done outside skyportal. If you save a source externally, the photometry will not update. Also, it will not necessarily import the forced photometry. It would be great to have an "update photometry" button that would pull all photometry for a source from kowalski.

**Usage / behavior**
Simple button that pulls photometry, as displayed on the alert page, onto the source page.

Example: https://fritz.science/alerts/ztf/ZTF25abatudj

### LINKED ISSUE #6118: Cache-and-display photometry
One of our biggest slowdowns is needing to write photometry to the database (and it is also confusing as photometry only gets written if an alert passes a filter so it does not always get updated). An alternative is to cache-and-display photometry where we fetch photometry on demand from the broker, hold it in a Valkey read-through cache keyed by (obj_id, requester_scope), and skip the Postgres write entirely.

antoine-le-calloch:
- Pick the broker that serves the source page's photometry with a `default_photometry` flag on the broker, replacing the `photometry_display_endpoint` config override, and refuse an unreachable broker as default.
- Let a provider opt out of the passthrough (ANTARES too slow for a page load, Pitt-Google bills a BigQuery job per view).
- Serve the same fields as the source photometry endpoint, check obj access, and scope the points to the requester's programids.
- Serve every obj of a SuperObj, each fetched under its own survey, so a ZTF+LSST pair is complete on both halves.
- Never answer the source page's photometry request with an error: log the broker failure and serve the saved photometry instead.
- Read the broker photometry through a disk cache and bound the fetch with a timeout, so a slow or down broker cannot stall the source page.
- Make the brokers page defaults clearable and explain each one in a header tooltip.


mcoughlin (approved):
LGTM!

PR #6721, opened 2026-09-09
Find the thumbnail folder without relying on the repository folder name

antoine-le-calloch:
`post_thumbnail` looked for the `static/thumbnails` folder by checking whether the path ended with `skyportal/skyportal`, which is only true when the clone is named `skyportal`. Any other folder name (a second clone, a rename, some deployment layouts) failed the check, and thumbnails were written to `<repo>/skyportal/static/thumbnails/`, a folder the server never reads. The images then came back as 404 everywhere: source pages, dashboard widgets, finding charts.

The base folder is now taken from the file position, so the name of the clone no longer matters. This matches how `skyportal/utils/tess.py` finds the `data` folder and where `tools/clear_data.py` looks for the thumbnails.


mcoughlin (approved):
LGTM

PR #6723, opened 2026-09-09
make the autosave column the only store for filter auto-save

antoine-le-calloch:
The broker filter UI read altdata['autoSave'] while ingestion read the autosave column, and set_autosave() kept the two in step only for requests that carried the autosave key. A PATCH on /api/filters rewriting altdata left the column untouched, so a filter could auto-save every passing alert while its toggle showed off.

Drop the mirror: the UI reads the autosave column, which the filter GET already returns, and nothing writes autoSave into altdata any more. The PATCH parameter keeps its autoSave name.


mcoughlin (approved):
This PR creates a patch on /api/filters rewriting altdata left the column untouched, so a filter could auto-save every passing alert while its toggle showed off.

PR #6724, opened 2026-09-09
Improve loading

antoine-le-calloch:
- The source page placeholder now has the same shape as the loaded page: same two columns, same sections, same plot heights, and the same stacked order on narrow screens.
- The star next to the source name works while the page loads.
- The photometry plot draws a skeleton instead of a small spinner.
- Skeletons sweep instead of fading in and out, set once in the theme.
- The "This is taking longer than usual" message is now a card in the middle of a dimmed page, closed by clicking outside. The sidebar and top bar stay usable.

Screenshot 2026-09-09 at 12 15 40 PM
Screenshot 2026-09-09 at 12 13 41 PM


mcoughlin (approved):
Yay for responsive app!

antoine-le-calloch:
> Yay for responsive app!

@mcoughlin responsive is also working.

mcoughlin:
Yay for a working app!

PR #6728, opened 2026-09-10
Fix gcn localization ingest and skymap leak

antoine-le-calloch:
Fixes the two issues Tomas Ahumada reported on https://fritz.science/gcn_events/2026-09-10T01:23:49:

- "Ingest Localization" returned a 500 (`greenlet_spawn has not been called`): `LocalizationNoticeHandler` read the deferred `GcnNotice.content` lazily from the async session. The column is now loaded in the initial query.
- The sky map went blank ("Too many active WebGL contexts", shader errors): `LocalizationPlot` never released its Aladin WebGL context on unmount, so remounts leaked one context each until Chrome hit its limit of 16. The context is now released through `WEBGL_lose_context` on unmount.


mcoughlin (approved):
LGTM

PR #6731, opened 2026-09-10
Lasair kafka

mcoughlin:
This PR enables for the ingestion of Kafka streams from Lasair.


camilledzt:
Hello, I have a question relative to the broker creation. In the case of Lasair, some users can have private filters that need tokens / credentials to have access to them. Since we are limiting the broker creation to admins, it means that users will have to send us their tokens, but that's not something that we should do. Should we decouple the broker creation and the token association, or allow people to create broker instances?

mcoughlin:
@camilledzt I added a broker credential.

@antoine-le-calloch at this point, I think you will want to also look at this (since it affects brokers broadly).

antoine-le-calloch (approved):
LGTM

PR #6735, opened 2026-09-11
Improve logging

antoine-le-calloch:
- Log the number of custom bandpasses instead of the full list, which was printed once per service at startup.
- Log the cosmology name instead of its full repr.
- Probe the app workers directly for readiness, instead of going through nginx where a failed probe marks a worker down for `server.fail_timeout` seconds.
- Replace the raw exception print in the health monitor with a single readable line.
- Drop the "no valid pinecone configuration" line, which fired on every start of a normal deployment.


mcoughlin (approved):
LGTM

PR #6736, opened 2026-09-11
Bump swifttools

antoine-le-calloch:
Bump swifttools to 4.0.5, which stops it showing every library's deprecation warnings: https://github.com/Swift-SOT/swifttools/pull/40.


mcoughlin (approved):
LGTM

PR #6740, opened 2026-09-11
Remove the Query-based access control that baselayer no longer calls

antoine-le-calloch:
Bumps baselayer to the current main and removes the access-control code that
comes with it.

- cesium-ml/baselayer#433 removed `UserAccessControl.query_accessible_rows` and
  `query_records_accessible_by`. `skyportal/models/group.py` still carried two
  overrides of `query_accessible_rows`, one of them calling
  `super().query_accessible_rows(...)` on a parent that no longer has it.
- Nothing calls them any more. `bulk_verify` goes through the `Select` API now,
  so `select_accessible_rows` is the only path left. The two overrides were dead
  code that would raise `AttributeError` if anything reached them.
- Also fixes the comment on `select_accessible_rows`, which still said the sync
  `bulk_verify` runs on the legacy Query API.

The bump also brings cesium-ml/baselayer#449 (`new_session` and `db_engine`) and
cesium-ml/baselayer#460, neither of which needs anything on this side.

### DIFF
--- a/skyportal/models/group.py
+++ b/skyportal/models/group.py
@@ -13,11 +13,9 @@ from baselayer.app.models import (
     AccessibleIfUserMatches,
     Base,
     CustomUserAccessControl,
-    DBSession,
     User,
     UserAccessControl,
     join_model,
-    public,
     safe_aliased,
 )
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #6741, opened 2026-09-11
Expect 403 when a permission is refused

antoine-le-calloch:
Follows cesium-ml/baselayer#468, which changes what baselayer answers when it refuses a request. Self-contained: it carries the baselayer bump itself, so it can go in before #6470.

baselayer already made the right distinction almost everywhere: 401 when the caller is not identified, 403 when they are identified and refused. Two places did not follow it, and both now answer 403:

- a caller holding a valid token that lacks the endpoint's ACL, previously 401
- `AccessError`, raised when the commit-time check finds a record the user may not touch, previously 400

401 is left for the cases it describes: no `Authorization` header, an unknown token, and a cookie session that resolves to no user.

So 49 of the 52 assertions on 401 move to 403, and the 6 assertions on an "Unauthorized" message become "Forbidden". The three assertions that stay on 401 are the real missing-credential cases: `test_mcp.py` calling with no token and with `"not-a-token"`, and `test_config_handler.py` calling with no token.

The rule used to sort them: after this change no path answers 401 to a request carrying a valid token, so every assertion reached with a real token fixture becomes 403, whether the refusal came from the `permissions` decorator or from an uncaught `AccessError`.

The bump also brings the `_resolve_pooler` rename, which the pooler tests follow here, and drops `verify_and_commit`, which `doc/permissions.md` no longer mentions.

Nothing in the client reads either status code, so no behaviour there changes.


stefanv:
>   Expect 403 when a permission is refused, 401 stays for missing credentials

This is correct.

mcoughlin (approved):
That's a nice cleanup.

PR #6742, opened 2026-09-12
make PS1_DR2 and LSDR10 visible on the filter building UI

thomasculino:
Makes the 2 new catalogs visible on the frontend and removes LS_DR10_PHOTOZ


mcoughlin (approved):
LGTM

PR #6743, opened 2026-09-12
Sort All observing runs tab

thomasculino:
Right now, on the observing runs page, if you look at the `All` tab, you will see oldest runs first (6yo). This PR sorts the runs the other way so upcoming runs appear first on the list.


mcoughlin (approved):
LGTM

PR #6745, opened 2026-09-14
Bump baselayer and skyportal-data

antoine-le-calloch:
- Bump baselayer and skyportal-data
- Fix warning on invalid escape sequence in ATLAS photometry parsing.
- Cover the access control path that cesium-ml/baselayer#432 just fixed. Nothing caught that hole: `is_accessible_by`, and so the whole permission suite, never asks a policy for columns only, which is exactly the path `bulk_verify` takes on every commit and where the cross join granted the whole table.

### DIFF
--- a/skyportal/facility_apis/atlas.py
+++ b/skyportal/facility_apis/atlas.py
@@ -133,7 +133,7 @@ def commit_photometry(
          """
 
         try:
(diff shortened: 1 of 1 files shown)


mcoughlin:
@antoine-le-calloch i think a test might need updating?

PR #6746, opened 2026-09-14
Check that column-filtered access control never builds a cross join

antoine-le-calloch:
Follows cesium-ml/baselayer#432. Nothing caught that hole: `is_accessible_by`, and so the whole permission suite, never asks a policy for columns only, which is exactly the path `bulk_verify` takes on every commit and where the cross join granted the whole table.

These tests cover that path, on every column-filtered policy (88 model and mode pairs).


antoine-le-calloch:
Moved into #6745.

PR #6748, opened 2026-09-14
Escape backslash in ATLAS facility API regex

MBK-fr:
### Problem

`skyportal/facility_apis/atlas.py` has a `\s` in a non-raw string literal, so Python 3.12+ raises `SyntaxWarning: invalid escape sequence '\s'` on import.

### Why it matters for SkyPortal

SkyPortal pins `requires-python >=3.12,<3.13` and runs Python 3.12 in CI, so the warning appears on every run.

### Fix

Escape the backslash; the string value is byte-for-byte identical (this is what ruff's W605 rule does). No behaviour change.

### Verification

`ast.dump` unchanged; `python3 -W error` py_compile passes after, failed before.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014HBQNzAf5C2E3MiJC48HQw


mcoughlin (approved):
LGTM

PR #6750, opened 2026-09-14
update publishers when selecting a different sharing service

thomasculino:
When sending to TNS, publishers' list is not updated when you select a sharing service that is not the default one.


antoine-le-calloch (approved):
LGTM. I also have a PR open, https://github.com/skyportal/skyportal/pull/6751, to fix that along with some other code and style issues. We can merge that one, but we should wait for mine to be redeployed, since reloading the form will not trigger a reload of all the fields depending on certain edge cases.

thomasculino:
> LGTM. I also have a PR open, #6751, to fix that along with some other code and style issues. We can merge that one, but we should wait for mine to be redeployed, since reloading the form will not trigger a reload of all the fields depending on certain edge cases.

@antoine-le-calloch We can close this one and focus on your PR, it might be cleaner. I also wanted to change the way co-authors are displayed on the Sharing Services page. Currently, they are ordered alphabetically which doesn't make sense because they are not ordered this way before being sent to TNS. Additionally, users don't want to have reporters/publishers displayed alphabetically but ordered by contribution. And the current display on the Sharing Services page makes it very confusing.

antoine-le-calloch:
Move to https://github.com/skyportal/skyportal/pull/6751

PR #6751, opened 2026-09-14
Fix and improve sharing services

antoine-le-calloch:
- fix stale form defaults when switching sharing service in the send-to dialog.
- Order sharing service coauthors explicitly and reorder them by drag and drop
- add an `order` column on SharingServiceCoauthor, sorted on by the relationship
  and the submission queue, with an endpoint to reorder the coauthors
- display them numbered in the published order, draggable to reorder, and closed
  by the acknowledgments as when publishing
- show instruments and streams as chips and tighten the columns
- improve sharing service and sharing service submissions table style
- clean up

image

## Before:
Screenshot 2026-09-15 at 10 53 12 AM

## After:
Screenshot 2026-09-15 at 10 47 16 AM

## Before:
Screenshot 2026-09-15 at 10 53 19 AM

## After:
Screenshot 2026-09-15 at 10 47 09 AM


mcoughlin (approved):
Looks awesome.

PR #6752, opened 2026-09-15
summary embeddings

mcoughlin:
This PR lets the summary embeddings come from pgvector rather than pinecone (which we have long stopped using).


antoine-le-calloch (approved):
LGTM

PR #6753, opened 2026-09-15
Keep the local RTK Query cache when a new tab inherits another tab's redux state

antoine-le-calloch:
Reloading a tab did not always show fresh data. If another SkyPortal tab was open, the reloaded tab took its data from that other tab instead of asking the server, so you kept seeing the old content even after a hard refresh. The only way out was to close every other tab first.

This happened because a new tab copies the whole redux state from an already open tab, API cache included.

With this fix a tab no longer inherits another tab's cached API data, so reloading always fetches from the server.


mcoughlin (approved):
Ah yes! LGTM!

PR #6757, opened 2026-09-15
account request page

mcoughlin:
This PR adds an account request page.

The idea is to (optionally) add a form where users can request an account. As part of this, you can (optionally) allow a peer endorsement (used by GCN, arxiv, etc.) where the user can specify a known user of the app to endorse them and receive an invitation. Otherwise it is super admins who must approve.


antoine-le-calloch:
@mcoughlin Could you add a bit more to the description? The current one doesn’t really give any information. Nothing too extensive, just enough so that by opening the PR, we can understand what it does.

antoine-le-calloch (approved):
LGTM

PR #6758, opened 2026-09-15
Persist tab search

thomasculino:
Allow shared filtering across tabs for groups and observations. You can now search for a group in the default tab, my groups, and switch to the all groups tab without having to type your search again.


antoine-le-calloch (approved):
LGTM

PR #6775, opened 2026-09-19
Fix config

antoine-le-calloch:
- Remove the unused `dask`/`distributed` dependencies and disable the `dask` service by default.
- Add `doc/services.md` documenting every service and when it is safe to disable it, linked from the doc index.
- Replace `services.dask: False` in `config.yaml.defaults` with a commented `services.disabled` list.
- Fix two stale service names in `test_config.yaml`: drop `gcnserver`, rename `external_publishing_submission_queue` to `sharing_service_submission_queue`.
- Use `cross_match.gaia.catalog` (now `gaiadr3.gaia_source`) as the real default everywhere, and stop appending `.gaia_source` to a catalog name that already has it.
- Render the finding chart over a blank frame when no survey image is available, with a test.
- Fix `get_ps1_url` logging `e.message` and the NGPS filter error message printing the wrong variable.
- Clean up.


mcoughlin (approved):
LGTM

PR #6793, opened 2026-09-20
Display the JWST cutout on the thumbnail list

antoine-le-calloch:
`jwst` was missing from `ARCHIVAL_THUMBNAIL_TYPES`, so a JWST cutout was never rendered: the "Request more thumbnails" button generated it, it was stored, and the list ignored it. For the same reason it did not count as an existing on-demand cutout, so the button stayed available and kept regenerating it.

Adding `jwst` to the displayed types fixes both.


mcoughlin (approved):
LGTM

PR #6795, opened 2026-09-20
Keep a saved homepage layout usable when a release adds a widget

antoine-le-calloch:
The homepage layout comes from the user's saved preferences when there is one, and the grid saves those preferences on every layout change, including the one it fires on mount. So almost every user ends up with a saved layout, and a saved layout only lists the widgets that existed when it was saved: widgets added by a later release are dropped wherever they fit, overlapping the others.

- A saved layout is merged with the current widget set, so a new widget takes its configured position and the user keeps theirs.
- The layout is saved only when its geometry actually changed, so opening the page no longer persists a layout nobody chose.


mcoughlin (approved):
LGTM

PR #6796, opened 2026-09-20
Skip disabled services before the app-readiness wait

antoine-le-calloch:
- `ep_service` and `gcn_crossmatch` checked their config inside `service()`, which
  `check_loaded` guards, so a disabled service polled the app for up to 100s before
  logging "disabled, skipping". The check now runs before the wait, like neofixer
  and tess_sector already do.
- `ensure_vector_extension` reported a missing pgvector as a privilege problem and
  told the admin to run `CREATE EXTENSION vector`, which fails the same way. It now
  tells the two apart on SQLSTATE and says to install pgvector first.
- Dropped the comments and docstrings that restate the line below them, and the
  intermediates they went with.


mcoughlin (approved):
LGTM

PR #6797, opened 2026-09-20
Clean up, improvement and fix - 1

antoine-le-calloch:
- Reworked the account applications page: fixed the broken toolbar, reused the shared deletion dialog, and tidied the dialogs and the table.
- Starting the server with the default `app.secret_key` now fails outside debug mode instead of only printing a warning.
- `debug_login` now defaults to `False`, and the setup documentation describes OAuth configuration and the development-only role of `debug_login`.
- Bumped baselayer
- Source ID and TNS name search is now case-insensitive, so `ztf` finds `ZTF26abc`.
- The source table search field keeps the focus while you type and queries once the text settles, instead of firing one request per letter.
- Improve source table style, detail panel the filter/new-source dialogs are loaded on demand (~20% less JavaScript on the first open of the sources page), and the page no longer blocks behind a full-page spinner.

## Before
Screenshot 2026-09-20 at 9 33 03 PM

## After
Screenshot 2026-09-20 at 9 31 13 PM
Screenshot 2026-09-20 at 9 31 19 PM
Screenshot 2026-09-20 at 10 04 39 PM
Screenshot 2026-09-20 at 10 04 44 PM


mcoughlin (approved):
LGTM

PR #6798, opened 2026-09-21
Tell the user when the assistant endpoint refuses the configuration

antoine-le-calloch:
When a question fails, the assistant answers "Something went wrong while looking that up.", whether the endpoint is briefly unreachable or the deployment has no valid key, no such model or a wrong URL. In the second case the user keeps asking and no one is told to fix the configuration.

A 401, 403 or 404 from the endpoint now answers that the assistant is not set up correctly and that an administrator should be told. Every other failure keeps the existing message.


mcoughlin (approved):
LGTM

PR #6799, opened 2026-09-21
Add a delete button to the filter list on the brokers page

antoine-le-calloch:
The filter list on the brokers page had no way to delete a filter: it had to be done from the owning group page. The group column was also plain text, with no link to the group, and a filter created from a group page was left attached to no broker.

- Add a delete button to each row, shown only to group admins (and to users with `System admin` or `Manage groups`), like on the group page. The catalog endpoint now reports whether the user administers the filter's group.
- Render the group name as a chip linking to its group page.
- Replace the `Attach to` column with an `Actions` column holding both the attach controls and the delete button.
- Ask which broker will run a new filter created from a group page, so it no longer starts out unattached.

Filters tab of the brokers page

Create a new filter dialog on a group page


mcoughlin (approved):
LGTM

PR #6800, opened 2026-09-21
Select TESS candidates with an anti-join, not NOT IN

mcoughlin:
This PR selects TESS candidates with an anti-join, not NOT IN.


leonitousconforti (approved):

leonitousconforti (commented):
Looks like some sql query is unhappy? Maybe just a flaky test though?

PR #6802, opened 2026-09-22
Leave a crashed ingestion alone until the broker record changes

antoine-le-calloch:
An ingestion that crashes is started again at the next rescan, sixty seconds later, and every rescan after that. A broker whose configuration the provider refuses, a Fink record with no Kafka servers for instance, therefore fails in the same way forever, and each attempt leaves memory behind: the service grew by several hundred megabytes an hour until the pod was evicted.

A crashed ingestion is now left alone until the broker record changes, since an edit is the only thing that can make the next attempt end differently.


mcoughlin (approved):
LGTM

PR #6805, opened 2026-09-22
Build Einstein Probe aliases from WXT GCN JSON notices

antoine-le-calloch:
Einstein Probe events ingested from the GCN Kafka JSON notices get no alias at all, while the VOEvent path builds FERMI#, LVC# and SVOM# ones and the EP data center service posts EP#<name>. Consumers that identify an event by its alias therefore skip every WXT event, even though the notice carries the EP trigger id.

- build EP#<id> from the WXT notice id in post_gcnevent_from_json
- set the event trigger id from that same id, and match on it before dateobs, so a notice with a revised trigger time updates the existing event instead of creating a new one


mcoughlin (approved):
LGTM

PR #6806, opened 2026-09-22
Separate ingesting from a broker from having it active

antoine-le-calloch:
Making a broker `active` does two unrelated things: it lets the instance call that broker on demand, and it subscribes the instance to its stream, since `broker_ingest` runs every active broker whose provider implements `run_ingestion`. Activating ALeRCE to search its alerts therefore starts consuming it in the background and writing objects into the database.

- `ingest` now says whether the ingestion service consumes a broker, separately from `active`. Existing rows are backfilled from `active`, so ingestions that run today keep running.
- The brokers page offers the toggle, disabled for providers that do not implement `run_ingestion`.


mcoughlin (approved):
LGTM

PR #6807, opened 2026-09-22
Generate an app.secret_key for the Docker image, shared by its workers

mcoughlin:
Test Docker Build has been failing on main since #6681:

```
RuntimeError: app.secret_key is the one shipped in config.yaml.defaults,
so session cookies and the credentials encrypted in the database are
readable by anyone. Set it to a random string.
```

All four app workers exit 1 in a restart loop and the health check
times out. The guard is right; the image just has no way to satisfy it.
`Dockerfile` does `cp docker.yaml config.yaml` at build time, and
`docker.yaml` sets no `secret_key`, so the container inherits the one
from `config.yaml.defaults` — the exact key the app refuses to run on.

Putting a key in `docker.yaml` would only move the problem: it would be
published in the repo and baked into every image built from it.

So the container generates one on first start, keeps it in the
`persistentdata` volume, and leaves alone any key someone set
themselves. A restart reuses the stored key, so sessions and the
credentials encrypted under it survive; a fresh volume means a fresh
key, which is the right trade for a key nobody chose.

Replicas each generate their own and a session is only valid on the one
that issued it, so anything past a single container should set
`app.secret_key` explicitly. The module docstring says so.

## Shape

The logic is in `skyportal/utils/secret_key.py` and `tools/` holds a
three-line entry point, so the tests can import it without executing a
script.

Six tests: the shipped key replaced, a missing key added to the `app`
block, **a config with no `app` block at all getting one** (writing the
key to the volume and not to the config would leave the app on the key
it refuses to start with, silently — the first version of this did
exactly that), a chosen key left alone, the key surviving a restart, and
neighbouring settings undisturbed.


leonitousconforti (approved):

stefanv:
Wow, that's quite the essay Claude.

The title of the PR says "give each container its own app.secret_key". Presumably, the idea is to give *the* container its own secret key. I'm asking, because apps must share a secret key, otherwise nginx will direct you to a different worker and your websocket auth will be void.

Dunno, the patch looks wrong to me, but maybe it's fine; flagging it so at least it is considered.

mcoughlin:
@stefanv an essay indeed. This is resolving the Docker secret stuff, which isn't used directly by us but does take some working around.

stefanv:
Not sure the persistentdata is going to save you on Kubernetes; the key is generated upon first runtime, so what happens when you fire up a bunch of pods from different nodes?

PR #6808, opened 2026-09-22
Release the Python packages with Knope

leonitousconforti:
`skyportal-py` and `skyportal-py-models` live in `projects/` but have no way to get to PyPI. This wires up [Knope](https://knope.tech)

The two packages are versioned and released together, as one Knope package named `clients`, so the app keeps the bare `vX.Y.Z` tags for itself and these are tagged `clients/vX.Y.Z`. Version bumps come from changesets: a small Markdown file under `.changeset/`, required only on PRs that touch `projects/`. Conventional commits are deliberately ignored, because Knope applies an unscoped commit to every package, so a plain `fix:` on the app would bump the clients.

Three workflows:

- `clients_changeset.yaml` asks for a changeset on PRs that change the clients,   with a `skip-changelog` label as the escape hatch.
- `clients_prepare_release.yaml` opens or updates a `release/clients` PR on every push to main, which consumes the pending changesets, bumps both `pyproject.toml` files and writes `projects/CHANGELOG.md`.
- `clients_publish.yaml` tags the release and uploads both distributions when that PR is merged.


mcoughlin:
@antoine-le-calloch @stefanv can you take this one on?

PR #6809, opened 2026-09-22
Run the Docker build check on pull requests that touch skyportal/

leonitousconforti:
The workflow's `pull_request` filter only lists the four docker files and itself, but `ADD . /skyportal` bakes the whole repo into the image, so an app-side change can break the container while the check stays skipped. It then fails on the push to `main`, after the PR is already merged.

That is what happened with #6797: it added a startup guard in `skyportal/app_server.py` that refuses to boot when `app.secret_key` is still the default from `config.yaml.defaults`. `docker.yaml` does not set one, so every worker raises and `.tokens.yaml` is never written. #6797 touched no docker path, so the check never ran on it, and `main` has been red since (#6807 is the fix in flight).

Adding `skyportal/**` covers that class of break: the guard lives under `skyportal/`, so this filter would have run the build on #6797 and caught it at review time rather than post-merge.


mcoughlin:
@leonitousconforti we can't be any more specific than that? With this, we will generate a Docker container basically every PR. I would almost prefer it breaks than do this.

PR #6810, opened 2026-09-22
Fix list condition compilation in the filter block builder

antoine-le-calloch:
The block builder drops part of what the list-condition dialog was set to when it compiles the expression, so a filter saved from the UI does not do what it shows.

- The root block's and/or selector was not passed to the compiler, so an `Or` always came out as `$and`.
- `$allElementsTrue` was matched and emitted as `$allElementTrue`, so an all-elements-true condition fell through to the bare array instead of a boolean.
- Array subfields were looked up only in the top-level field options, so they resolved to no type and defaulted to string, and `$eq`/`$ne` turned a numeric value into a string: `distance_kpc = -1` compiled to `{"$eq": ["$$this.distance_kpc", "-1"]}`, which never matches a double.


mcoughlin (approved):
LGTM

PR #6811, opened 2026-09-22
Keep forced photometry when merging broker and database points

antoine-le-calloch:
Forced photometry never reached the source lightcurve. Both dedup keys used to
merge a broker alert into an object's photometry left out `origin`, while
`Photometry.DEDUP_COLUMNS` includes it, so a forced measurement was dropped
whenever an alert detection or upper limit already covered its epoch. ZTF
`fp_hists` covers the very exposures `prv_nondetections` reports a limit for, so
in practice all of it disappeared: on ZTF26abvtmma the endpoint returned 29
points and none of BOOM's 25 forced measurements.

- `merge_photometry_points` now keys on `origin` too, so a broker forced point is
  kept next to the saved alert point at the same epoch.
- `build_photometry_groups` dedups epochs per origin, so `fp_hists` no longer
  loses the epochs `prv_candidates` already carries.


mcoughlin (approved):
LGTM

